PR #447, opened 2015-09-09
Enable jenkins builds.

tolbertam:
Adds build.yaml for jenkins builds.  Creates a job in jenkins for each branch with build.yaml present that does the following:

For each jdk (openjdk6, oraclejdk7, oraclejdk8) and cassandra (latest major of 1.2, 2.0 and 2.1) combination runs 'long' integration tests and records the test results.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1,0 +1,21 @@
+java:
+  - openjdk6
+  - oraclejdk7
+  - oraclejdk8
+os:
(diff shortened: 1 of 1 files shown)


tolbertam:
Just made a slight update to use maven 3.2.5 in the event that a newer version of maven is installed (maven 3.3+ requires JDK7).  This should be ready to merge.

PR #451, opened 2015-09-14
Move call to connection.release() out of cancelHandler (JAVA-901)

tommystendahl:


olim7t:
Hi @tommystendahl, sorry I had already started working on this on a private branch, I just opened #499. The fixes are very similar, but I didn't call `connection.release()` in `onTimeoutTask()`, as that would still release twice if the caller did it (like in RequestHandler.onTimeout). Also there was a workaround in Connection.setKeyspace that can be removed now that this is fixed.

PR #452, opened 2015-09-15
JAVA-913: Redesign IndexMetadata API.

adutra:
See CASSANDRA-10216 for details.

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/IndexMetadataTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/IndexMetadataTest.java
@@ -20,7 +20,9 @@ import java.util.Collection;
 import java.util.List;
 
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.primitives.Ints;
+import org.testng.annotations.BeforeClass;
 import org.testng.annotations.Test;
 
 import com.datastax.driver.core.ColumnMetadata.*;
@@ -30,15 +32,17 @@ import com.datastax.driver.core.utils.CassandraVersion;
 import static com.datastax.driver.core.Assertions.assertThat;
 import static com.datastax.driver.core.ColumnMetadata.*;
 import static com.datastax.driver.core.DataType.cint;
+import static com.datastax.driver.core.DataType.map;
 import static com.datastax.driver.core.DataType.text;
+import static com.datastax.driver.core.IndexMetadata.Kind.*;
 
 @CassandraVersion(major = 1.2)
 public class IndexMetadataTest extends CCMBridge.PerClassSingleNodeCluster {
 
     /**
-     * Column definitions for schema_columns table.
+     * Column definitions for schema_columns table (legacy pre-3.0 layout).
      */
-    private static final ColumnDefinitions defs = new ColumnDefinitions(new ColumnDefinitions.Definition[]{
+    private static final ColumnDefinitions legacyColumnDefs = new ColumnDefinitions(new ColumnDefinitions.Definition[]{
         definition(COLUMN_NAME, text()),
         definition(COMPONENT_INDEX, cint()),
         definition(KIND_V2, text()),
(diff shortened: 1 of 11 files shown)


driver-core/src/test/java/com/datastax/driver/core/IndexMetadataTest.java
```diff
@@ -73,167 +95,187 @@
...
+            .hasTarget("text_column")
             .asCqlQuery(createValuesIndex);
-        assertThat(index).isSameAs(column.getIndex("text_column_index"));
-        assertThat(column).hasSingleIndex(index);
-        assertThat(column).isSameAs(index.getColumn("text_column"));
+        assertThat(column.getTable()).hasIndex(index);
+    }
+
+    @Test(groups = "short")
+    public void should_create_metadata_for_simple_index_on_mixed_case_column() {
```
tolbertam: This one is failing on <= 2.1 for some reason, though looks like it should work...looking into it.

tolbertam: Just needed to selectively create the index based on major version much like should_create_metadata_for_index_on_map_values, i.e.: https://gist.github.com/tolbertam/4be3efec48975fe4b4d4

tolbertam:
Added a version requirement check to one of the tests (feel free to squash that when merging).  This looks good to me :+1:

tolbertam:
Added another test (for testing with multiple indexes on same column - not really necessary but figured would be good to add) and also added 'kind' validation.   PR still looks good to me :+1:

tolbertam:
I noticed via [cassandra-dtests](https://github.com/riptano/cassandra-dtest/blob/540b754108fd0ca74aa0adc088c486d84b71cd1f/materialized_views_test.py#L408-L426) that it was possible to add secondary indexes to materialized views.   I noticed that the driver didn't support that yet, so i added it here in 92dabb3 since it seemed pretty straightforward, feel free to back out/change/etc. if it isn't an appropriate change.

Secondary indexes are semi-busted on C\* trunk right now (added comment to [CASSANDRA-9921](https://issues.apache.org/jira/browse/CASSANDRA-9921)), so the test I added currently fails (unless you put a breakpoint before the execute statement and add the index manually twice), but should pass once fixed in C*.

adutra:
lgtm :+1: thanks for finding out that 2is on mvs are allowed!

adutra:
@tolbertam I just pushed a commit reverting indexes on materialized views following the decision in CASSANDRA-9921. Could you kindly review it again?

tolbertam:
Sounds good, taking a look :)

driver-core/src/test/java/com/datastax/driver/core/IndexMetadataTest.java
```diff
@@ -48,192 +52,304 @@
...
+            .asCqlQuery(createKeysIndex);
+
+        assertThat(getIndex("map_all_values_index"))
+            .hasParent(table)
+            .asCqlQuery(createValuesIndex);
+    }
+
+    @Test(groups = "short", enabled = false)
+    @CassandraVersion(major = 3.0)
+    public void should_create_metadata_for_index_on_materialized_view() {
```
tolbertam: It'd probably be fine to just delete this test and if support gets added back we can readd it then

driver-core/src/test/java/com/datastax/driver/core/IndexMetadataTest.java
```diff
@@ -48,192 +52,304 @@
...
+                    ?
+                    ", map_full frozen<map<text, int>>,"
+                        + "set_full frozen<set<text>>,"
+                        + "list_full frozen<list<text>>,"
+                    :
+                    "")
+                + "PRIMARY KEY (id, id2));"
+        );
+
+        if(cassandraVersion.compareTo(VersionNumber.parse("3.0.0-rc1")) >= 0) {
```
tolbertam: This is no longer needed if MVs dont support indexes.

adutra: Thanks for both comments, I updated the PR.

tolbertam:
looks good to me :+1:

PR #454, opened 2015-09-15
Support cassandra.directory and cassandra.branch props for CCMBridge.

tolbertam:
As I was testing #453, I enhanced CCMBridge in such a way that I could use a custom install, figured I'd create a PR for this.

Adds optional cassandra.directory and cassandra.branch properties as
a means of running cassandra with CCM beyond deployed cassandra
versions.

The order of precedence for what is used is:
1. cassandra.directory
2. cassandra.branch
3. cassandra.version

cassandra.version is still required for `@CassandraVersion` annotation
checks for tests, but won't be used to launch CCM if either
cassandra.directory or cassandra.branch is present.

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
@@ -53,9 +53,9 @@ public class CCMBridge {
 
     public static final String IP_PREFIX;
 
-    private static final String CASSANDRA_VERSION_REGEXP = "\\d\\.\\d\\.\\d+(-\\w+)?";
-    static final File CASSANDRA_DIR;
     static final String CASSANDRA_VERSION;
+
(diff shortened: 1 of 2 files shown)


adutra:
That's very useful! So far I've been using symlinks, e.g.:

``` shell
ln -s ~/path/to/cassandra ~/.ccm/repository/3.0.0-rc1
```

Where `~/path/to/cassandra` is a local git branch checked out and compiled with `ant clean build`.

olim7t:
Nice. You could already specify a local directory before, but I like having an explicit property for it. Maybe we should also support the `-v github:` syntax.

tolbertam:
Ah, I wasn't even aware you could do that, nice!  Could extend cassandra.branch to support 'github:' syntax.  I'll open up a separate PR for that.

PR #455, opened 2015-09-16
JAVA-912: Use native CQL names to store data types in schema 'columns' table.

adutra:
This commit also incorporates the new "clustering_order" column in schema 'columns' table. 

Note that this is a work in progress. In particular UDTs do not work yet.


adutra:
Closing for now as native CQL types in schema 'columns' table has been reverted.

PR #456, opened 2015-09-16
JAVA-918: Allow MV's select statements to be more complex.

adutra:
See https://issues.apache.org/jira/browse/CASSANDRA-9664 for details.

Working branch:
https://github.com/thobbs/cassandra/tree/CASSANDRA-9664-rebase-3

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/MaterializedViewMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/MaterializedViewMetadata.java
@@ -18,13 +18,10 @@ package com.datastax.driver.core;
 
 import java.util.*;
 
-import com.google.common.base.Function;
-import com.google.common.base.Joiner;
 import com.google.common.base.Objects;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import static com.google.common.collect.Iterables.transform;
 
 /**
  *  An immutable representation of a materialized view.
@@ -38,6 +35,8 @@ public class MaterializedViewMetadata extends TableOrView {
 
     private final boolean includeAllColumns;
 
+    private final String whereClause;
+
     private MaterializedViewMetadata(
         KeyspaceMetadata keyspace,
         TableMetadata baseTable,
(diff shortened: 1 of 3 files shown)


driver-core/src/main/java/com/datastax/driver/core/MaterializedViewMetadata.java
```diff
@@ -178,21 +172,12 @@ public String apply(ColumnMetadata input) {
...
-        Iterator<ColumnMetadata> it = getPrimaryKey().iterator();
-        while(it.hasNext()) {
-            ColumnMetadata column = it.next();
-            sb.append(spaces(4, formatted)).append(Metadata.escapeId(column.getName()));
-            sb.append(" IS NOT NULL");
-            if(it.hasNext()) sb.append(" AND");
-            sb.append(" ");
-            newLine(sb, formatted);
-        }
+        newLine(sb.append("WHERE ").append(whereClause).append(' '), formatted);
```
tolbertam: Are where clauses required?   The cql grammar seems to allow you to create MVs without it, but cassandra doesn't allow it.  It also seems that previously you could create MVs without it (for example from [this blog post](http://christopher-batey.blogspot.com/2015/07/cassandra-30-materialised-views-in.html) but that could be outdated).   If they are not required we should have some criteria here to omit this if the whereClause is empty or null.

I posted a question on [CASSANDRA-9664](https://issues.apache.org/jira/browse/CASSANDRA-9664).

tolbertam: Got an [explanation](https://issues.apache.org/jira/browse/CASSANDRA-6477?focusedCommentId=14627173&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14627173) that for now the where clause will be required, but it does seem possible in the future that this could be optional, so I think it would still be good to handle this case.

tolbertam:
I added some tests for Accessors just to make sure the driver was able to properly map a `@Table` annotated pojo for a different able to a view, and also to exercise some basic MV functionality.   Everything looks ready to me other than my comment about the where clause (which will probably be dependent on the response in CASSANDRA-9664).  :+1:.  I am :+1: on merging this if no change is needed for that.

driver-core/src/test/java/com/datastax/driver/core/MaterializedViewMetadataTest.java
```diff
@@ -75,27 +81,33 @@ public void should_create_view_metadata() {
         assertThat(mv.getColumns().get(3)).isNotNull().hasName("score").isClusteringColumn().hasClusteringOrder(DESC);
         assertThat(mv.getColumns().get(4)).isNotNull().hasName("user").isClusteringColumn();
         assertThat(mv.getColumns().get(5)).isNotNull().hasName("day").isClusteringColumn();
-
-        assertThat(mv.asCQLQuery(false)).contains(createMV);
-
+        //whereClause currently puts quoted identifiers around column names so this assertion would fail, see:
```
tolbertam: This can be uncomment or removed based on the response in CASSANDRA-9664.

adutra: I re-enabled the assertion, knowing that it will probably fail until Sylvain's suggestion is implemented.

driver-core/src/main/java/com/datastax/driver/core/MaterializedViewMetadata.java
adutra: Added a test for missing WHERE clauses to be on the safe side.

PR #457, opened 2015-09-16
Fix CustomPayloadTest failing when run in suite.

tolbertam:
CustomPayloadTest fails when running in the full suite.  It turns out
this is caused by PreparedStatementTest restarting the cassandra node
and not propagating the CustomPayloadMirroringQueryHandler system
property that gets set when the single node cluster is initialized.

Changed ccmBridge to private and added 'stop' and 'start' methods on
PerClassSingleNodeCluster for stopping/starting the node while
preserving jvm arguments.

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
@@ -402,10 +402,11 @@ public class CCMBridge {
     // One cluster for the whole test class
     public static abstract class PerClassSingleNodeCluster {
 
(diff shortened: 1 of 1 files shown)


tolbertam:
@olim7t mind merging this in before merging 2.2 into the 3.0 branch?  Without this there are 14 test failures for each 2.2+ configuration in our test suite.

tolbertam:
This isn't as important now because those PreparedStatementTests were removed in favor of QueryOptionsTest.  I'll rebase but this doesn't need to be done right away anymore.

PR #459, opened 2015-09-17
JAVA-911: Process materialized views events.

adutra:
One test failing: `com.datastax.driver.core.SchemaChangesTest#should_notify_of_view_update`. It seems that the view is not being altered at all.


tolbertam:
looks good to me :+1:

PR #443, opened 2015-09-04
Update tests to use Scassandra 0.10.0, tests for JAVA-544, other test cleanup.

tolbertam:

PR #462, opened 2015-09-29
JAVA-929 Fix equals method for index_interval columns.

tolbertam:
Fix for [JAVA-929](https://datastax-oss.atlassian.net/browse/JAVA-929).

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/TableMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/TableMetadata.java
@@ -842,7 +842,11 @@ public class TableMetadata {
                 return false;
             if (defaultTTL != options.defaultTTL)
                 return false;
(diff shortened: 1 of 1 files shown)


tolbertam:
Previously failing test `SchemaChangesTest#should_notify_of_table_update` on C\* 2.0.   Would consider table as changed when it didn't since index_interval was never considered as equal.   TableMetadataTest could use some better tests (some good tests on upstream branches), will work that on another ticket.

tolbertam:
travis failing on should_stop_if_cancelled_between_attempts(com.datastax.driver.core.AbstractReconnectionHandlerTest) which is fixed on #443

olim7t:
Merged. I amended your commit to also update hashCode.

tolbertam:
Ah, great catch!  Didn't even think about the hashcode, glad you caught that :)

PR #463, opened 2015-09-30
Make TableOrView public to allow TableOrView.Option to be referenced outside package.

adutra:
Initially reported by DevCenter team.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -9,6 +9,7 @@
 - [bug] Fix Configuration builder to allow disabled metrics (JAVA-900)
 - [new feature] Prepare API for async query trace (JAVA-902)
 - [new feature] Add BoundStatement#unset (JAVA-930)
+- [bug] Make table metadata options class visible (JAVA-946)
 
 
 ### 3.0.0-alpha3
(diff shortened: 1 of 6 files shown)


olim7t:
I understand the rationale but making it public means we commit to its public/protected API forever. This raises a couple of issues:
- the name is not great
- could be an issue if we need to refactor in the future (e.g. add intermediary abstract classes in the hierarchy...)

Another way to approach this would be to make `Options` a top-level class (and rename it maybe), would that work?

On a side note, `nullInitializedList` can be replaced by `Collections.fill`.

adutra:
Ok for making Options a top-level class, but do you have any suggestion for its name? I can't think of anything suitable :)

adutra:
Turns out that `Collections.fill` is not suitable, but `Collections.nCopies` works.

olim7t:
Made the class top-level and renamed it to `TableOptionsMetadata`.

I accidentally rebased my local branch (sorry about that), so I ended up squashing my changes to the initial commit.

tolbertam:
I rebased on 3.0 (was some merge conflicts due to changes for #478).  This looks good to me :+1:

PR #464, opened 2015-09-30
JAVA-922: TypeCodec returns unmodifiable collections.

adutra:


olim7t:
Looks good to me. We can probably squash all commits together.
One small remark: could you describe what the commit does instead of what the problem was, i.e. "TypeCodec returns unmodifiable collections" => "Make TypeCodec return mutable collections".

adutra:
Rebased and squashed. I also followed your suggestion concerning the Row class and changed the commit message as per your suggestion.

tolbertam:
looks good to me as well, :+1:

olim7t:

PR #465, opened 2015-10-03
update method link in PoolingOptions class 

chriswebster:
This change updates the javadoc link to reference the setMaxConnectionsPerHost to match the textual description.


olim7t:
Good catch, thanks.

PR #467, opened 2015-10-06
JAVA-936: Adapt schema metadata parsing logic to new storage format of CQL types in C* 3.0

adutra:
NOTE: can now be tested against the newly-released C\* 3.0.0.

Majors changes:

Several columns in the schema metadata tables used to contain internal cassandra types. They now contain a CQL representation of the type, instead of internal Cassandra names. `DataTypeParser` is now used to parse them (instead of `CassandraTypeParser`).

This poses a problem for UDTs: they are now simply represented by their names (whereas before the column would contain enough information to build the UserType from scratch). The problem is even worse for nested UDTs.

From now on, this problem is solved by two techniques:
- Topological sort of UDTs: when the entire schema (or keyspace) is rebuilt, we have enough information to be able to build the dependency graph for UDTs. This allows us to process them in topological sort order, and all UDTs are resolved (note that nested udts now reference the same instance, whereas before we would construct a different, but equivalent instance).
- during a partial refresh, if a UDT definition references a nested UDT u2, and u2 is not part of the refresh, we look for the definition of u2 in the "old" metadata as it was before the start of the refresh. Thanks to the topological sort, we're sure that we won't reference a stale definition that would be updated later in the refresh.

The only case were UDT resolution should fail is if Cassandra sends us notifications out of order. In this case we abort the whole refresh and log a warning that the metadata will be out of date.

The other noticeable difference is initconds. They used to be stored as blobs; now they are stored as CQL literals.

See CASSANDRA-10365 and CASSANDRA-10650 for details.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/AggregateMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/AggregateMetadata.java
@@ -58,6 +58,7 @@ public class AggregateMetadata {
         this.stateTypeCodec = stateTypeCodec;
     }
 
+    // Cassandra < 3.0:
     // CREATE TABLE system.schema_aggregates (
     //     keyspace_name text,
     //     aggregate_name text,
@@ -70,42 +71,74 @@ public class AggregateMetadata {
     //     state_type text,
     //     PRIMARY KEY (keyspace_name, aggregate_name, signature)
     // ) WITH CLUSTERING ORDER BY (aggregate_name ASC, signature ASC)
-    static AggregateMetadata build(KeyspaceMetadata ksm, Row row, ProtocolVersion protocolVersion, CodecRegistry codecRegistry) {
+    //
+    // Cassandra >= 3.0:
+    // CREATE TABLE system.schema_aggregates (
+    //     keyspace_name text,
+    //     aggregate_name text,
+    //     argument_types frozen<list<text>>,
+    //     final_func text,
+    //     initcond text,
+    //     return_type text,
+    //     state_func text,
+    //     state_type text,
+    //     PRIMARY KEY (keyspace_name, aggregate_name, argument_types)
+    // ) WITH CLUSTERING ORDER BY (aggregate_name ASC, argument_types ASC)
+    static AggregateMetadata build(KeyspaceMetadata ksm, Row row, VersionNumber version, Cluster cluster) {
+        CodecRegistry codecRegistry = cluster.getConfiguration().getCodecRegistry();
+        ProtocolVersion protocolVersion = cluster.getConfiguration().getProtocolOptions().getProtocolVersion();
         String simpleName = row.getString("aggregate_name");
-        List<String> signature = row.getList("signature", String.class);
-        String fullName = Metadata.fullFunctionName(simpleName, signature);
-        List<DataType> argumentTypes = parseTypes(row.getList("argument_types", String.class), protocolVersion, codecRegistry);
+        List<DataType> argumentTypes = parseTypes(ksm, row.getList("argument_types", String.class), version, cluster);
+        String fullName = Metadata.fullFunctionName(simpleName, argumentTypes);
         String finalFuncSimpleName = row.getString("final_func");
-        DataType returnType = CassandraTypeParser.parseOne(row.getString("return_type"), protocolVersion, codecRegistry);
+        DataType returnType;
+        if(version.getMajor() >= 3) {
+            returnType = DataTypeCqlNameParser.parse(row.getString("return_type"), cluster, ksm.getName(), ksm.userTypes, null, false, false);
+        } else {
+            returnType = DataTypeClassNameParser.parseOne(row.getString("return_type"), protocolVersion, codecRegistry);
+        }
         String stateFuncSimpleName = row.getString("state_func");
         String stateTypeName = row.getString("state_type");
-        DataType stateType = CassandraTypeParser.parseOne(stateTypeName, protocolVersion, codecRegistry);
-        ByteBuffer rawInitCond = row.getBytes("initcond");
-        Object initCond = rawInitCond == null ? null : codecRegistry.codecFor(stateType).deserialize(rawInitCond, protocolVersion);
+        DataType stateType;
+        Object initCond;
+        if(version.getMajor() >= 3) {
+            stateType = DataTypeCqlNameParser.parse(stateTypeName, cluster, ksm.getName(), ksm.userTypes, null, false, false);
+            String rawInitCond = row.getString("initcond");
+            initCond = rawInitCond == null ? null : codecRegistry.codecFor(stateType).parse(rawInitCond);
+        } else {
+            stateType = DataTypeClassNameParser.parseOne(stateTypeName, protocolVersion, codecRegistry);
+            ByteBuffer rawInitCond = row.getBytes("initcond");
+            initCond = rawInitCond == null ? null : codecRegistry.codecFor(stateType).deserialize(rawInitCond, protocolVersion);
+        }
 
         String finalFuncFullName = finalFuncSimpleName == null ? null : String.format("%s(%s)", finalFuncSimpleName, stateType);
-        String stateFuncFullName = makeStateFuncFullName(stateFuncSimpleName, stateType.toString(), signature);
+        String stateFuncFullName = makeStateFuncFullName(stateFuncSimpleName, stateType, argumentTypes);
 
-        AggregateMetadata aggregate = new AggregateMetadata(ksm, fullName, simpleName, argumentTypes,
+        return new AggregateMetadata(ksm, fullName, simpleName, argumentTypes,
             finalFuncSimpleName, finalFuncFullName, initCond, returnType, stateFuncSimpleName,
             stateFuncFullName, stateType, codecRegistry.codecFor(stateType));
-        ksm.add(aggregate);
-        return aggregate;
     }
 
-    private static String makeStateFuncFullName(String stateFuncSimpleName, String stateTypeName, List<String> typeNames) {
-        List<String> args = Lists.newArrayList(stateTypeName);
-        args.addAll(typeNames);
+    private static String makeStateFuncFullName(String stateFuncSimpleName, DataType stateType, List<DataType> argumentTypes) {
+        List<DataType> args = Lists.newArrayList(stateType);
+        args.addAll(argumentTypes);
         return Metadata.fullFunctionName(stateFuncSimpleName, args);
     }
 
-    private static List<DataType> parseTypes(List<String> names, ProtocolVersion protocolVersion, CodecRegistry codecRegistry) {
-        if (names.isEmpty())
+    private static List<DataType> parseTypes(KeyspaceMetadata ksm, List<String> types, VersionNumber version, Cluster cluster) {
+        if (types.isEmpty())
             return Collections.emptyList();
 
+        CodecRegistry codecRegistry = cluster.getConfiguration().getCodecRegistry();
+        ProtocolVersion protocolVersion = cluster.getConfiguration().getProtocolOptions().getProtocolVersion();
         ImmutableList.Builder<DataType> builder = ImmutableList.builder();
-        for (String name : names) {
-            DataType type = CassandraTypeParser.parseOne(name, protocolVersion, codecRegistry);
+        for (String name : types) {
+            DataType type;
+            if (version.getMajor() >= 3) {
+                type = DataTypeCqlNameParser.parse(name, cluster, ksm.getName(), ksm.userTypes, null, false, false);
+            } else {
+                type = DataTypeClassNameParser.parseOne(name, protocolVersion, codecRegistry);
+            }
             builder.add(type);
         }
         return builder.build();
--- a/driver-core/src/main/java/com/datastax/driver/core/ColumnMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ColumnMetadata.java
@@ -56,8 +56,8 @@ public class ColumnMetadata {
         this.isStatic = isStatic;
     }
 
-    static ColumnMetadata fromRaw(TableOrView tm, Raw raw) {
-        return new ColumnMetadata(tm, raw.name, raw.dataType, raw.kind == Raw.Kind.STATIC);
+    static ColumnMetadata fromRaw(TableOrView tm, Raw raw, DataType dataType) {
+        return new ColumnMetadata(tm, raw.name, dataType, raw.kind == Raw.Kind.STATIC);
     }
 
     static ColumnMetadata forAlias(TableMetadata tm, String name, DataType type) {
@@ -165,12 +165,12 @@ public class ColumnMetadata {
         public final String name;
         public Kind kind;
         public final int position;
-        public final DataType dataType;
+        public final String dataType;
         public final boolean isReversed;
 
         public final Map<String, String> indexColumns = new HashMap<String, String>();
 
-        Raw(String name, Kind kind, int position, DataType dataType, boolean isReversed) {
+        Raw(String name, Kind kind, int position, String dataType, boolean isReversed) {
             this.name = name;
             this.kind = kind;
             this.position = position;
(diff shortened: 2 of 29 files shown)


driver-core/pom.xml
```diff
@@ -51,6 +51,12 @@
       <version>${metrics.version}</version>
     </dependency>
 
+    <dependency>
```
adutra: This is to be able to build DAGs (directed acyclic graphs) for inter-UDT dependencies. See SchemaParser and UserType.

driver-core/src/main/java/com/datastax/driver/core/AggregateMetadata.java
```diff
@@ -70,42 +71,75 @@ private AggregateMetadata(KeyspaceMetadata keyspace, String fullName, String sim
...
         String stateFuncSimpleName = row.getString("state_func");
         String stateTypeName = row.getString("state_type");
-        DataType stateType = CassandraTypeParser.parseOne(stateTypeName, protocolVersion, codecRegistry);
-        ByteBuffer rawInitCond = row.getBytes("initcond");
-        Object initCond = rawInitCond == null ? null : codecRegistry.codecFor(stateType).deserialize(rawInitCond, protocolVersion);
+        DataType stateType;
+        Object initCond;
+        if(version.getMajor() >= 3) {
+            stateType = DataTypeParser.parse(stateTypeName, cluster.getMetadata(), ksm.userTypes, false);
+            ByteBuffer rawInitCond = row.getBytes("initcond");
```
adutra: This field is back again to its original type, blob.

driver-core/src/main/java/com/datastax/driver/core/MaterializedViewMetadata.java
```diff
@@ -101,7 +101,15 @@ static MaterializedViewMetadata build(KeyspaceMetadata keyspace, Row row, Map<St
         // 'SELECT * FROM ...'
         Set<ColumnMetadata> otherColumns = new TreeSet<ColumnMetadata>(columnMetadataComparator);
         for (ColumnMetadata.Raw rawCol : rawCols.values()) {
-            ColumnMetadata col = ColumnMetadata.fromRaw(view, rawCol);
+            DataType dataType;
+            if(cassandraVersion.getMajor() >= 3) {
+                dataType = DataTypeParser.parse(rawCol.dataType, cluster.getMetadata(), keyspace.userTypes, false);
```
adutra: Data types are now resolved later in the process, because we need user types to be resolved first.

driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
```diff
@@ -182,26 +183,32 @@ void refresh(Metadata metadata,
...
+    private Map<String, UserType> buildUserTypes(KeyspaceMetadata keyspace, List<Row> udtRows, VersionNumber cassandraVersion, Cluster cluster) {
         Map<String, UserType> userTypes = new LinkedHashMap<String, UserType>();
         if (udtRows != null) {
-            ProtocolVersion protocolVersion = metadata.cluster.protocolVersion();
-            CodecRegistry codecRegistry = metadata.cluster.configuration.getCodecRegistry();
-            for (Row udtRow : udtRows) {
-                UserType type = UserType.build(udtRow, protocolVersion, codecRegistry);
-                userTypes.put(type.getTypeName(), type);
+            if(cassandraVersion.getMajor() >= 3 && udtRows.size() > 1) {
+                UserTypeDependencyGraph graph = new UserTypeDependencyGraph(udtRows, cluster.getMetadata());
```
adutra: Sort UDTs by topological order.

driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
```diff
@@ -641,12 +646,13 @@ private String whereClause(SchemaElement targetType, String targetKeyspace, Stri
                     whereClause += " AND table_name = '" + targetName + '\'';
                 else if (targetType == VIEW)
                     whereClause += " AND view_name = '" + targetName + '\'';
-                else if (targetType == TYPE)
+                else if (targetType == TYPE && cassandraVersion.getMajor() < 3)
+                    // C* 3+: we need all user types to be able to resolve nested UDTs
```
adutra: This might have performance impacts...

driver-core/src/main/java/com/datastax/driver/core/TableMetadata.java
```diff
@@ -180,7 +183,13 @@ else if (cassandraVersion.getMajor() > 2)
         }
 
         for (ColumnMetadata.Raw rawCol : rawCols.values()) {
-            ColumnMetadata col = ColumnMetadata.fromRaw(tm, rawCol);
+            DataType dataType;
```
adutra, line 186: Again, data types resolved later than before.

driver-core/src/main/java/com/datastax/driver/core/UserType.java
```diff
@@ -70,9 +138,15 @@ static UserType build(Row row, ProtocolVersion protocolVersion, CodecRegistry co
         List<String> fieldTypes = row.getList(COLS_TYPES, String.class);
 
         List<Field> fields = new ArrayList<Field>(fieldNames.size());
-        for (int i = 0; i < fieldNames.size(); i++)
-            fields.add(new Field(fieldNames.get(i), CassandraTypeParser.parseOne(fieldTypes.get(i), protocolVersion, codecRegistry)));
-
+        for (int i = 0; i < fieldNames.size(); i++) {
+            DataType fieldType;
```
adutra, line 77: Same here: data type resolution being delayed.

driver-core/src/main/java/com/datastax/driver/core/ColumnMetadata.java
```diff
@@ -198,26 +198,26 @@ static Raw fromRow(Row row, VersionNumber version, ProtocolVersion protocolVersi
                 position = row.isNull(COMPONENT_INDEX) ? 0 : row.getInt(COMPONENT_INDEX);
             }
 
-            String dataTypeStr;
+            String dataType;
```
adutra, line 201: Until now, data type resolution was done here. But now this is too early, so we only store the raw string for now, and we parse it later on (in TableMetadata, MVMetadata and UserType).

driver-core/src/main/java/com/datastax/driver/core/AggregateMetadata.java
```diff
@@ -70,42 +71,75 @@ private AggregateMetadata(KeyspaceMetadata keyspace, String fullName, String sim
...
         String stateFuncSimpleName = row.getString("state_func");
         String stateTypeName = row.getString("state_type");
-        DataType stateType = CassandraTypeParser.parseOne(stateTypeName, protocolVersion, codecRegistry);
-        ByteBuffer rawInitCond = row.getBytes("initcond");
-        Object initCond = rawInitCond == null ? null : codecRegistry.codecFor(stateType).deserialize(rawInitCond, protocolVersion);
+        DataType stateType;
+        Object initCond;
+        if(version.getMajor() >= 3) {
+            stateType = DataTypeParser.parse(stateTypeName, cluster, ksm, ksm.userTypes, false);
+            String rawInitCond = row.getString("initcond");
```
adutra, line 106: initconds are back to text type thanks to CASSANDRA-10650.

driver-core/src/main/java/com/datastax/driver/core/AggregateMetadata.java
```diff
@@ -70,42 +71,74 @@ private AggregateMetadata(KeyspaceMetadata keyspace, String fullName, String sim
...
         String stateTypeName = row.getString("state_type");
-        DataType stateType = CassandraTypeParser.parseOne(stateTypeName, protocolVersion, codecRegistry);
-        ByteBuffer rawInitCond = row.getBytes("initcond");
-        Object initCond = rawInitCond == null ? null : codecRegistry.codecFor(stateType).deserialize(rawInitCond, protocolVersion);
+        DataType stateType;
+        Object initCond;
+        if(version.getMajor() >= 3) {
+            stateType = DataTypeParser.parse(stateTypeName, cluster, ksm.getName(), ksm.userTypes, false);
+            String rawInitCond = row.getString("initcond");
+            initCond = rawInitCond == null ? null : codecRegistry.codecFor(stateType).parse(rawInitCond);
```
olim7t, line 107: Should we handle the case where stateType is an `UnresolvedUserType`?

driver-core/src/main/java/com/datastax/driver/core/DataTypeParser.java
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        // return a custom type for the special empty type
+        // so that it gets detected later on, see TableMetadata
+        if(type.equalsIgnoreCase(EMPTY))
+            return custom(type);
+
+        // Try UDTs
+        if (userTypes != null) {
+            UserType userType = userTypes.get(Metadata.handleId(type));
```
olim7t: I was surprised by the call to `handleId` here, but Cassandra does indeed quote mixed-case names in its type columns:

```
cqlsh:test> create type "MixedCaseType"(i int);
cqlsh:test> create table foo(k int primary key, t frozen<"MixedCaseType">);
cqlsh:test> select type from system_schema.columns where keyspace_name = 'test' and table_name = 'foo' and column_name = 't';

 type
-------------------------
 frozen<"MixedCaseType">
```

driver-core/src/main/java/com/datastax/driver/core/UserType.java
```diff
@@ -25,12 +25,118 @@
...
+     * A dependency graph to resolve dependencies between
+     * user-defined types.
+     * For C* 3+, user-defined type resolution must be done in proper order
+     * to guarantee that nested UDTs get resolved;
+     * this class provides topological sort for a given set
+     * of UDTs, implementing Kahn's algorithm.
+     *
+     * @see <a href="https://en.wikipedia.org/wiki/Topological_sorting">Topological sorting</a>
+     */
+    static class UserTypeDependencyGraph implements Iterable<Row> {
```
olim7t: Can we make this a top-level class?

driver-core/src/main/java/com/datastax/driver/core/DataTypeParser.java
tolbertam: It does seem like an inconsistency, but I'm guessing it needs to be done otherwise someone could throw a '>' in a quoted identifier to throw the parsing off.

For example:

```
cqlsh:test> create type "Mixe>dCase>Type"(i int);
cqlsh:test> create table foo3 ("Key" int primary key, v list<frozen<"Mixe>dCase>Type">>);
cqlsh:test> select * from system_schema.columns where keyspace_name='test' and table_name='foo3';

 keyspace_name | table_name | column_name | clustering_order | column_name_bytes | kind          | position | type
---------------+------------+-------------+------------------+-------------------+---------------+----------+---------------------------------
          test |       foo3 |         Key |             none |          0x4b6579 | partition_key |        0 |                             int
          test |       foo3 |           v |             none |              0x76 |       regular |       -1 | list<frozen<"Mixe>dCase>Type">>
```

This typename seems to throw off of the DataTypeParser due the appearance of '>' in the typename:

```
    657    [main] ERROR com.datastax.driver.core.SchemaParser  - Error parsing schema for table test.foo3: Cluster.getMetadata().getKeyspace("test").getTable("foo3") will be missing or incomplete
com.datastax.driver.core.exceptions.DriverInternalError: Excepting single parameter for list, got [frozen<"Mixe>, dCase]
    at com.datastax.driver.core.DataTypeParser.parse(DataTypeParser.java:84)
    at com.datastax.driver.core.TableMetadata.build(TableMetadata.java:188)
    at com.datastax.driver.core.SchemaParser.buildTables(SchemaParser.java:171)
    at com.datastax.driver.core.SchemaParser.buildKeyspaces(SchemaParser.java:123)
    at com.datastax.driver.core.SchemaParser.refresh(SchemaParser.java:64)
    at com.datastax.driver.core.ControlConnection.refreshSchema(ControlConnection.java:329)
    at com.datastax.driver.core.ControlConnection.tryConnect(ControlConnection.java:260)
    at com.datastax.driver.core.ControlConnection.reconnectInternal(ControlConnection.java:187)
    at com.datastax.driver.core.ControlConnection.connect(ControlConnection.java:79)
    at com.datastax.driver.core.Cluster$Manager.init(Cluster.java:1398)
    at com.datastax.driver.core.Cluster.init(Cluster.java:167)
    at com.datastax.driver.core.Cluster.connectAsync(Cluster.java:346)
    at com.datastax.driver.core.Cluster.connectAsync(Cluster.java:319)
    at com.datastax.driver.core.Cluster.connect(Cluster.java:257)
```

driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
```diff
@@ -182,26 +181,30 @@ void refresh(Metadata metadata,
...
+                UserTypeDependencyGraph graph = new UserTypeDependencyGraph(udtRows, cluster, keyspace.getName());
+                for (Row udtRow: graph) {
+                    UserType type = UserType.build(keyspace, udtRow, cassandraVersion, cluster, userTypes);
+                    userTypes.put(type.getTypeName(), type);
+                }
+            } else {
+                for (Row udtRow : udtRows) {
+                    UserType type = UserType.build(keyspace, udtRow, cassandraVersion, cluster, userTypes);
+                    userTypes.put(type.getTypeName(), type);
+                }
```
olim7t: This can be simplified to avoid duplicating the `for` loop.

driver-core/src/main/java/com/datastax/driver/core/DataTypeParser.java
```diff
@@ -0,0 +1,255 @@
...
+
+/*
+ * Helps transforming CQL types (as read in the schema tables) to
+ * DataTypes, starting from Cassandra 3.0.
+ *
+ * Note that these methods all throw DriverInternalError when there is a parsing
+ * problem because in theory we'll only parse class names coming from Cassandra and
+ * so there shouldn't be anything wrong with them.
+ */
+class DataTypeParser {
```
olim7t: We should rename `DataTypeParser` and `CassandraTypeParser`, the current names don't tell what's different between the two classes.
I suggest `CassandraTypeCqlNameParser` and `CassandraTypeClassNameParser`.

driver-core/src/main/java/com/datastax/driver/core/UnresolvedUserType.java
```diff
@@ -0,0 +1,130 @@
...
+        return resolved.get().getFieldIndicesByName();
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (!(o instanceof UserType))
+            return false;
+        resolve();
+        return resolved.get().equals(o);
+    }
```
olim7t: There are a few issues here:
1. an instance is not equal to itself
2. an instance is not equal to another unresolved type referencing the same UDT
3. equality is not symetric (the actual UserType is not equal to an unresolved type that references it)
4. equal objects can have different hashcodes

1 and 2 are easily solved, but 3 and 4 are more complex. I'm not really sure if an unresolved type should be equal to the type it references.

driver-core/src/test/java/com/datastax/driver/core/UnresolvedUserTypeTest.java
```diff
@@ -0,0 +1,126 @@
...
+        assertThat(b).hasField("f1", set(d));
+        assertThat(c).hasField("f1", map(e, d));
+        assertThat(d).hasField("f1", metadata.newTupleType(f, g, h));
+        assertThat(e).hasField("f1", list(g));
+        assertThat(f).hasField("f1", h);
+        assertThat(g).hasField("f1", cint());
+        assertThat(h).hasField("f1", cint());
+
+        // each CREATE TYPE statement has triggered a partial schema refresh,
+        // so nested UDTs are all unresolved
```
olim7t: I agree that it's how it works currently, but we should be able to do better than that. If `g` already exists in the metadata and we create `E` that references `g`, then a partial refresh that only targets `E` should not result in an unresolved type.
We could do this by allowing `buildUserTypes` to reference "old" types, only in the case where it's been triggered by a partial refresh.

olim7t:
Pushed changes in UDT resolution as discussed privately with Alexandre. I've updated the PR description.

tolbertam:
Looks good to me!  :+1:

olim7t:
I'm :+1: with @tolbertam's latest change.
Once JAVA-988 is on 3.0, we can rebase this. Then we just need to remove `sanitizedType` in `DataTypeCqlNameParser` and it will be good to go.

olim7t:
Force-pushed to rebase on current 3.0 and squash the commits.

PR #468, opened 2015-10-06
JAVA-937 declare codec singleton at TypeCodec level

arnaudroger:


adutra:
Thank you @arnaudroger for your contribution! I was wondering: why do you need codecs declared at TypeCodec level, is it because your code makes direct use of codecs and you couldn't access them because they were declared in package-private classes?
In this case, have you considered upgrading to 3.0? Since JAVA-721, the codecs subsystem has been entirely refactored and not only you can [access the driver's internal codecs](https://github.com/datastax/java-driver/blob/3.0/driver-core/src/main/java/com/datastax/driver/core/TypeCodec.java#L124) from within application code, but you can also [create your own codecs](https://github.com/datastax/java-driver/tree/3.0/features/custom_codecs).

driver-core/src/main/java/com/datastax/driver/core/AbstractAddressableByIndexData.java
```diff
@@ -44,42 +44,42 @@ protected ByteBuffer getValue(int i) {
...
-        return setValue(i, TypeCodec.DoubleCodec.instance.serializeNoBoxing(v));
+        return setValue(i, TypeCodec.doubleCodec.serializeNoBoxing(v));
     }
 
     public T setString(int i, String v) {
         DataType.Name type = checkType(i, DataType.Name.VARCHAR, DataType.Name.TEXT, DataType.Name.ASCII);
         switch (type) {
             case ASCII:
-                return setValue(i, v == null ? null : TypeCodec.StringCodec.asciiInstance.serialize(v));
+                return setValue(i, v == null ? null : TypeCodec.StringCodec.asciiStringCodec.serialize(v));
```
adutra, line 79: You shouldn't access a static field in the parent class through one of its subclasses.

arnaudroger:
the problem that occurs is linked to the fact that the instance are declared at the sublevel and some circular dependency on the initialisation of classes.

In some cases the the instance static variable in the subtype is not initialised when the Map is build. that lead to the map containing a null reference instance of the actual instance. The problem happened in some VM but not other when staring cassandra for a unit test. Right now I sorted the issue - trigger on openjdk8 in travis ci- by using reflection and setting the correct value if the map contains a null. 

Moving singleton from the subtype to the enclosing class solve the dependency initialisation issue.

The problem does not happen in the version 3 of the driver.

adutra:
Your proposed fix seems the right way to go, however I would keep the instances package-private - otherwise I think we could get some false positives with Maven clirr plugin.

adutra:
Closing in favor of #534 as I had to modify a few items but your commit has been preserved. Thank you again for helping us improve the driver!

PR #469, opened 2015-10-06
JAVA-933: Consider allowing users to override default codecs.

adutra:
Overview of changes:
- The overriding of a single codec takes place in `CodecRegistry.register()` : if the exact same CQL-Java mapping already exists, the existing codec is replaced by the new one (note the additional log warning).
- The _on the fly_  overriding of collection codecs is more complex: there is now a new `CodecFactory` class that can be subclassed for this purpose. 
- Behavior in case of _overlapping_ codecs didn't change: it is still possible to register more than one codec for the same CQL type.

Note: the CodecFactory class allow users to override the on-the-fly instantiation mechanism, but we could live without it. If the user knows which _exact_ list types he needs to handle, the following code:

``` java
CodecFactory factory = new DefaultCodecFactory() {
    protected <T> TypeCodec<List<T>> newListCodec(TypeCodec<T> eltCodec) {
        return new MyListCodec<T>(eltCodec);
    }
};
CodecRegistry registry = new CodecRegistry(factory);
```

Could easily be replaced with:

``` java
CodecRegistry registry = new CodecRegistry();
registry.register(new MyListCodec(IntCodec.instance));
registry.register(new MyListCodec(VarcharCodec.instance));
// etc. register all list types required by the application
```

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/CodecFactory.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/CodecFactory.java
@@ -1,0 +1,250 @@
+/*
+ *      Copyright (C) 2012-2015 DataStax Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+package com.datastax.driver.core;
+
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import com.google.common.reflect.TypeToken;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+import com.datastax.driver.core.TypeCodec.*;
+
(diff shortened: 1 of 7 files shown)


driver-core/src/main/java/com/datastax/driver/core/CodecFactory.java
```diff
@@ -0,0 +1,250 @@
...
+import com.datastax.driver.core.TypeCodec.*;
+
+import static com.datastax.driver.core.DataType.Name.*;
+
+/**
+ * A factory for {@link TypeCodec} instances.
+ * Instances of this interface are used by the {@link CodecRegistry} to create
+ * codecs on-the-fly.
+ */
+public interface CodecFactory {
```
adutra, line 39: This is basically created out of the `maybeCreateCodec` methods in `CodecRegistry`.

driver-core/src/main/java/com/datastax/driver/core/CodecFactory.java
```diff
@@ -0,0 +1,250 @@
...
+            return null;
+
+        }
+
+        /**
+         * Hook for subclasses wishing to customize codecs for CQL {@code list} types.
+         * @param eltCodec The element codec.
+         * @return A suitable codec.
+         */
+        protected <T> TypeCodec<List<T>> newListCodec(TypeCodec<T> eltCodec) {
```
adutra, line 226: This would be the method to override if someone wants to create list codecs on the fly.

driver-core/src/main/java/com/datastax/driver/core/CodecRegistry.java
```diff
@@ -355,9 +408,7 @@ public CodecRegistry register(Iterable<? extends TypeCodec<?>> codecs) {
      * @throws CodecNotFoundException if a suitable codec cannot be found.
      */
     public <T> TypeCodec<T> codecFor(T value) {
-        checkNotNull(value, "Parameter value cannot be null");
```
adutra, line 358: Had to get rid of a couple of (excessive) checks as the factory now needs to call some of these methods.

adutra:
Closing as the corresponding ticket has been closed as Won't Fix.

PR #471, opened 2015-10-06
JAVA-952 Tests for validating equals/hashCode implementations and include MaterializedViewMetadata#whereClause in equals/hashCode

tolbertam:
There have been a few occasions now where we've added fields to classes and missed updating equals/hashCode implementations.  I thought it would be good to use [equalsverifier](http://www.jqno.nl/equalsverifier/) to validate that equals and hashCode implementations are correct where applicable.  Having tests for classes that depend on equals/hashCode will 'future-proof' any changes to these classes by detecting missing field use in these methods.

I didn't implement this comprehensively for all classes overriding equals/hashCode in the driver, just the ones that were related to schema.

I chose to make this change on 3.0 since it adds some API restrictions (marking methods final that previously weren't).  We could make this worker on older branches but not sure if it's worth it since schema changes seem less likely.

I found a variety of issues which were since fixed.  The only remaining issue I've identifier that should probably be definitively fixed is:
- MaterializedViewMetadata was not including whereClause in equals/hashcode.

There are a few cases where its debatable where certain fields need to be included in equals/hashCode as they may not factor in what we use these methods for (i.e. keyspace/table parentage of the type).  I've added comments in the test classes where this is applicable.

Source-Level Changes:
- Add test level dependency for equalsverifier to device-core/pom.xml
- In General
  - Use Objects.equal in favor of .equals() for non-primitives not already using it.
  - Null-checking (Maybe we should use Nonnull annotation instead?)
  - Made equals and hashCode final.  Did this instead of making whole class final (as can't mock classes that are final).

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/AggregateMetadataTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/AggregateMetadataTest.java
@@ -20,6 +20,7 @@ import java.util.Collections;
 import java.util.List;
 
 import com.google.common.collect.ImmutableList;
+import nl.jqno.equalsverifier.EqualsVerifier;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
@@ -40,6 +41,13 @@ public class AggregateMetadataTest {
     @BeforeMethod(groups = "unit")
     public void setup() {
         keyspace = new KeyspaceMetadata("ks", false, Collections.<String, String>emptyMap());
+    }
+
+    @Test(groups = "unit")
+    public void should_use_all_fields_in_equals_and_hashCode() {
+        EqualsVerifier.forClass(AggregateMetadata.class)
+            .allFieldsShouldBeUsedExcept("simpleName", "finalFuncSimpleName", "stateFuncSimpleName", "stateTypeCodec")
+            .verify();
     }
 
     @Test(groups = "unit")
--- a/driver-core/src/test/java/com/datastax/driver/core/ColumnMetadataTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/ColumnMetadataTest.java
@@ -1,0 +1,29 @@
+/*
+ *      Copyright (C) 2012-2015 DataStax Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+package com.datastax.driver.core;
+
+import nl.jqno.equalsverifier.EqualsVerifier;
+import org.testng.annotations.Test;
+
+public class ColumnMetadataTest {
+
+    @Test(groups = "unit")
+    public void should_use_all_fields_in_equals_and_hashCode() {
+        EqualsVerifier.forClass(ColumnMetadata.class)
+            .allFieldsShouldBeUsedExcept("parent")
+            .verify();
+    }
+}
--- a/driver-core/src/test/java/com/datastax/driver/core/FunctionMetadataTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/FunctionMetadataTest.java
@@ -19,6 +19,7 @@ import java.util.Collections;
 import java.util.List;
 
 import com.google.common.collect.ImmutableList;
+import nl.jqno.equalsverifier.EqualsVerifier;
 import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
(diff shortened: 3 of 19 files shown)


driver-core/src/test/java/com/datastax/driver/core/KeyspaceMetadataTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+import nl.jqno.equalsverifier.EqualsVerifier;
+import org.testng.annotations.Test;
+
+public class KeyspaceMetadataTest {
+
+    @Test(groups = "unit")
+    public void should_use_all_fields_in_equals_and_hashCode() {
+        EqualsVerifier.forClass(KeyspaceMetadata.class)
+            .allFieldsShouldBeUsedExcept("views", "userTypes", "functions", "aggregates")
```
tolbertam, line 26: I noticed that 'tables' is included in equals/hashCode, but not views/userTypes/functions/aggregates.  Should that be considered a bug?

tolbertam:
Just realized that some of the updates i've made to equals/hashCode should also be applied on older branches, so would be good to mark this as a 'conceptual' PR (work in progress).  Would like feedback on the changes / approach and then can open PRs for equals/hashcode fixes as applicable on older branches first.

driver-core/src/test/java/com/datastax/driver/core/AggregateMetadataTest.java
```diff
@@ -43,6 +44,13 @@ public void setup() {
     }
 
     @Test(groups = "unit")
+    public void should_use_all_fields_in_equals_and_hashCode() {
+        EqualsVerifier.forClass(AggregateMetadata.class)
+            .allFieldsShouldBeUsedExcept("simpleName", "finalFuncSimpleName", "stateFuncSimpleName", "stateTypeCodec")
```
tolbertam, line 49: simple names are derived from other fields so don't need to be included, codec doesn't seem relevant either.

driver-core/src/test/java/com/datastax/driver/core/ColumnMetadataTest.java
```diff
@@ -0,0 +1,29 @@
...
+
+import nl.jqno.equalsverifier.EqualsVerifier;
+import org.testng.annotations.Test;
+
+public class ColumnMetadataTest {
+
+    @Test(groups = "unit")
+    public void should_use_all_fields_in_equals_and_hashCode() {
+        EqualsVerifier.forClass(ColumnMetadata.class)
+            .allFieldsShouldBeUsedExcept("parent")
```
tolbertam, line 26: The parent table/keyspace names are not considered in equals/hashCode.  Since this comparison is only done between Column changes in the same table i'm guessing this is not needed?

driver-core/src/test/java/com/datastax/driver/core/FunctionMetadataTest.java
```diff
@@ -38,6 +39,13 @@ public void setup() {
     }
 
     @Test(groups = "unit")
+    public void should_use_all_fields_in_equals_and_hashCode() {
+        EqualsVerifier.forClass(FunctionMetadata.class)
+            .allFieldsShouldBeUsedExcept("simpleName")
```
tolbertam, line 44: simpleName is derived.

driver-core/src/test/java/com/datastax/driver/core/IndexMetadataTest.java
```diff
@@ -102,6 +103,13 @@ public void beforeClass() {
         return builder.build();
     }
 
+    @Test(groups = "unit")
+    public void should_use_all_fields_in_equals_and_hashCode() {
+        EqualsVerifier.forClass(IndexMetadata.class)
+            .allFieldsShouldBeUsedExcept("table")
```
tolbertam, line 109: The parent table name is not considered in equals/hashCode.  Since this comparison is only done between Index changes in the same table i'm guessing this is not needed?

driver-core/src/test/java/com/datastax/driver/core/MaterializedViewMetadataTest.java
```diff
@@ -29,6 +30,13 @@
 @CassandraVersion(major = 3)
 public class MaterializedViewMetadataTest extends CCMBridge.PerClassSingleNodeCluster {
 
+    @Test(groups = "unit")
+    public void should_use_all_fields_in_equals_and_hashCode() {
+        EqualsVerifier.forClass(MaterializedViewMetadata.class)
+            .allFieldsShouldBeUsedExcept("cassandraVersion", "keyspace")
```
tolbertam, line 36: cassandraVersion is likely not considered since we don't expect it to be different between MVs.  Compared within MVs in same keyspace so keyspace doesn't factor in?

driver-core/src/test/java/com/datastax/driver/core/TableMetadataTest.java
```diff
@@ -67,6 +68,13 @@ public void should_parse_table_without_clustering_columns() {
         );
     }
 
+    @Test(groups = "unit")
+    public void should_use_all_fields_in_equals_and_hashCode() {
+        EqualsVerifier.forClass(TableMetadata.class)
+            .allFieldsShouldBeUsedExcept("cassandraVersion", "keyspace")
```
tolbertam, line 74: same as MaterializeViewMetadata

driver-core/src/test/java/com/datastax/driver/core/UserTypeTest.java
```diff
@@ -0,0 +1,28 @@
...
+package com.datastax.driver.core;
+
+import nl.jqno.equalsverifier.EqualsVerifier;
+import org.testng.annotations.Test;
+
+public class UserTypeTest {
+    @Test(groups = "unit")
+    public void should_use_all_fields_in_equals_and_hashCode() {
+        EqualsVerifier.forClass(UserType.class)
+            .allFieldsShouldBeUsedExcept("protocolVersion", "codecRegistry", "byName")
```
tolbertam, line 25: protocolVersion/codecRegistry will be Cluster specific (probably not needed), byName is a convenience map.

tolbertam:
rebased on 3.0 the only change that wasn't already in 3.0 was MaterializedViewMetadata equals/hashCode missing whereClause.

tolbertam:
Closing this as it needs to be rebased/revisited, will open up a new PR later.

PR #474, opened 2015-10-07
JAVA-932: Remove default codecs from public API.

adutra:
This pretty much removes everything public in TypeCodec:
- default codecs
- extra convenience codecs
- equals and hashcode 

Note that the `PrimitiveXCodec` interfaces remain public. I think legit that users could retrieve a codec from CodecRegistry and test if they implement them to avoid the boxing/unboxing overhead.

New static methods allow to retrieve the default codecs for every CQL type. I don't think it is very useful for the average user (CodecRegistry should be the way to go in most cases), but they could prove useful in some situations:
- if we allow users to override default codecs (JAVA-933)
- if one really doesn't have access to a CodecRegistry

I did not include a `DataType.getDefaultCodec()` instance method. Do you think it would be useful?


olim7t:
Added a commit to get rid of `setObject`. There is one failing test that should succeed once this is rebased with JAVA-940.

olim7t:

PR #475, opened 2015-10-08
JAVA-942: Incorrect implementation of UserType.hashCode()

adutra:


tolbertam:
Looks good to me :+1: (other than small comment about imports)!  When implementing tests using equalsverifier in #471, it also caught this by detecting Arrays.hashCode was not used on byIdxl.  I attempted a fix, but I like your approach better :).

tolbertam:
Just realized, should this be fixed on 2.1 as well?

tolbertam:
Another thing I noticed was that 'name' is not included in equals, but is in hashCode.  Do you think equals should be updated as well to incorporate that?

adutra:
Thank you for all the (very good) remarks - I squashed (sorry) my previous commit.
As for this being applied to 2.1, I think it should in theory, but it is more urgent for 2.2; and I'm not sure we have enough time to commit to 2.1 and merge upstream. Maybe could we cherry-pick this commit into 2.1?

olim7t:
Closing to reopen against 3.0 (#483)

PR #476, opened 2015-10-08
JAVA-940: GettableData "Old API" methods should not check for CQL type.

adutra:
Highlights:
- It is now possible to call `getInt()` even if the underlying CQL type is not `INT` - there must be however a custom codec previously registered to handle the conversion. Same for all similar methods.
- Collection setters (`setList`, `setSet`, `setMap`): there is no more runtime vodoo magic to try to guess the best codec; the algorithm relies solely on the underlying CQL type. So if the collection has an exotic element type, one of the overloaded methods should be used instead.
- Javadocs updated (lots of minor incoherences); `Row` class entirely pruned.
- The test `instanceof PrimitiveXCodec` remains: it is now really useful (while previously it was always true)


olim7t:

PR #477, opened 2015-10-08
Warn if a custom codec collides with an existing one (JAVA-934).

olim7t:


driver-core/src/test/java/com/datastax/driver/core/CodecRegistryTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.testng.annotations.Test;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import static com.datastax.driver.core.Assertions.assertThat;
+
+public class CodecRegistryTest {
+    TypeToken<List<Integer>> LIST_OF_INT_TOKEN = new TypeToken<List<Integer>>() {
+    };
```
adutra, line 32: I usually tend to avoid the newline after opening curly braces when using the type token idiom.

adutra:
lgtm

driver-core/src/test/java/com/datastax/driver/core/CodecRegistryTest.java
olim7t, line 32: Intellij's autoformat adds the line break, I'm not sure if that can be configured otherwise.

olim7t:

PR #478, opened 2015-10-09
JAVA-938 Parse extensions into TableOrView Options.

tolbertam:
[JAVA-938](https://datastax-oss.atlassian.net/browse/JAVA-938).

Work in progress, need a mechanism to test this.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/TableOrView.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/TableOrView.java
@@ -16,6 +16,7 @@
 package com.datastax.driver.core;
 
 
+import java.nio.ByteBuffer;
 import java.util.*;
 
 import com.google.common.base.Objects;
@@ -371,6 +372,7 @@ abstract class TableOrView {
         private static final String MIN_INDEX_INTERVAL          = "min_index_interval";
         private static final String MAX_INDEX_INTERVAL          = "max_index_interval";
         private static final String CRC_CHECK_CHANCE            = "crc_check_chance";
+        private static final String EXTENSIONS                  = "extensions";
 
         private static final boolean DEFAULT_REPLICATE_ON_WRITE = true;
         private static final double DEFAULT_BF_FP_CHANCE = 0.01;
@@ -402,6 +404,7 @@ abstract class TableOrView {
         private final Map<String, String> compaction;
         private final Map<String, String> compression;
         private final Double crcCheckChance;
+        private final Map<String, ByteBuffer> extensions;
 
         Options(Row row, boolean isCompactStorage, VersionNumber version) {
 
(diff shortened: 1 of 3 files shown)


tolbertam:
Added test that fakes out a populated extensions column from existing real table metadata.  Should be good to go.

driver-core/src/test/java/com/datastax/driver/core/TableMetadataTest.java
```diff
@@ -269,6 +273,7 @@ public void should_parse_table_options() {
             assertThat(table.getOptions().getMaxIndexInterval()).isEqualTo(2048);
             assertThat(table.getOptions().getReplicateOnWrite()).isTrue(); // default
             assertThat(table.getOptions().getCrcCheckChance()).isEqualTo(0.5);
+            assertThat(table.getOptions().getExtensions()).isEqualTo(ImmutableMap.of()); // default
```
olim7t: Maybe isEmpty() would be simpler?

driver-core/src/test/java/com/datastax/driver/core/TableMetadataTest.java
```diff
@@ -474,4 +482,53 @@ public void should_not_mix_indexes_from_different_tables() {
         assertThat(table_cd.getIndexes().size()).isEqualTo(1);
         assertThat(table_cd.getIndexes().get(0).getName()).isEqualTo("test_d");
     }
+
+    /**
+     * Validates that the 'extensions' option is properly parsed when set.  This value is currently not
+     * modifiable via CQL so we fake out a row containing populated extensions by retrieving
+     * the current metadata and changing the extensions column.  The row is passed to the Options
+     * constructor and we validate the extensions map is matches what was set.
```
olim7t: Another possible approach is to mutate the row in the system table with a CQL query:

```
update system_schema.tables set extensions = {'foo': 0x00}
where keyspace_name = 'test' and table_name = 'foo';
```

I'm not sure if it's simpler because we then need to trigger a reload (I don't think the above query triggers a notification). We could add a column to the table.

driver-core/src/main/java/com/datastax/driver/core/TableOrView.java
```diff
@@ -675,6 +683,17 @@ public Double getCrcCheckChance() {
             return compression;
         }
 
+        /**
+         * Returns the extension options for this table.
+         * <p>
+         * For Cassandra versions prior to 3.0.0, this method always returns {@code null}.
```
olim7t: An empty map might be better.

driver-core/src/test/java/com/datastax/driver/core/TableMetadataTest.java
tolbertam: :+1: will change that.

driver-core/src/test/java/com/datastax/driver/core/TableMetadataTest.java
tolbertam: That does seem like a better way to test this.  I'll change it.

driver-core/src/main/java/com/datastax/driver/core/TableOrView.java
tolbertam: :+1: will change.

driver-core/src/test/java/com/datastax/driver/core/TableMetadataTest.java
tolbertam: altering the table after changing the extensions column sets it back to an empty map, so I explicitly call refreshSchema instead.

PR #479, opened 2015-10-11
Add BoundStatement#unset (JAVA-930).

olim7t:


tolbertam:
I added a statement to the prepared statement docs since they were slightly stale (claimed unbound values were sent as null, where now we throw exceptions).  I pretty much copied the docs from BoundStatement.java as is.  This looks good to me :+1:, though I think it may be nice to consider making `BoundStatement.UNSET` public and moving it elsewhere (like `Statement.UNSET`)

PR #481, opened 2015-10-11
Fix Configuration builder to allow disabled metrics (JAVA-900).

olim7t:


tolbertam:
Looks good to me!  Added some tests for validating metric enablement / disablement.

PR #482, opened 2015-10-11
JAVA-877: Don't wait NEW_NODE_DELAY_SECONDS if protocol version = V4.

olim7t:


tolbertam:
This looks good, would be interesting to see how this effects some of our integration tests.  :+1: on merging it in.

PR #483, opened 2015-10-11
JAVA-942: Fix implementation of UserType.hashCode()

olim7t:


tolbertam:
:+1:, from the JIRA we should also make sure to backport this onto 2.1

olim7t:
Backported to 2.1 as 0f9d734.

PR #484, opened 2015-10-11
Change default consistency level to LOCAL_QUORUM (JAVA-926).

olim7t:


tolbertam:
:+1:, probably goes without saying that we will need to make this very explicit in the release announcement :smile:

PR #485, opened 2015-10-11
TypeCodec improvements (JAVA-931)

olim7t:


driver-core/src/main/java/com/datastax/driver/core/AbstractAddressableByIndexData.java
```diff
@@ -256,22 +209,14 @@ public T setInet(int i, InetAddress v) {
...
         return setValue(i, codecFor(i, UDTValue.class).serialize(v, protocolVersion));
     }
 
     public T setTupleValue(int i, TupleValue v) {
-        checkType(i, TUPLE);
         return setValue(i, codecFor(i, TupleValue.class).serialize(v, protocolVersion));
     }
 
     @Override
-    public <V> T setObject(int i, V v) {
```
tolbertam, line 269: Any concerns about a loss of utility here?   I think depending on class/typetoken/typecodec in set() is likely a good move in that the user needs to specify what targetClass they intend to be communicating here.  I do see someone wanting the use case of "Here's an object, I have no idea what this is, but let the driver figure out what to do with it", but that on its face sounds like it could have unreliable/unpredictable results.

tolbertam:
rebased on 3.0

driver-core/src/main/java/com/datastax/driver/core/CodecRegistry.java
```diff
@@ -31,29 +32,28 @@
...
 import com.datastax.driver.core.exceptions.CodecNotFoundException;
 
-import static com.datastax.driver.core.DataType.Name.*;
+import static com.datastax.driver.core.DataType.Name.LIST;
+import static com.datastax.driver.core.DataType.Name.MAP;
+import static com.datastax.driver.core.DataType.Name.SET;
 
 /**
- * A registry for {@link TypeCodec}s. When the driver
+ * A registry for {@link TypeCodec codec}s. When the driver
```
tolbertam, line 42: The 'codec' part here looks accidental since it breaks the link.

tolbertam, line 42: nevermind, misinterpretation on my part.

driver-core/src/main/java/com/datastax/driver/core/CodecRegistry.java
```diff
@@ -279,32 +302,50 @@ public void onRemoval(RemovalNotification<CacheKey, TypeCodec<?>> notification)
...
-            .build(
-                new CacheLoader<CacheKey, TypeCodec<?>>() {
-                    public TypeCodec<?> load(CacheKey cacheKey) {
-                        return findCodec(cacheKey.cqlType, cacheKey.javaType);
-                    }
-                });
+            .maximumWeight(1000);
+        if (logger.isTraceEnabled())
+            // do not bother adding a listener if it will be ineffective
+            builder = builder.removalListener(new TypeCodecRemovalListener());
```
tolbertam, line 317: It's possible that log levels can be manipulated dynamically, but that seems like such a thin use case that we don't need to worry about it.

tolbertam:
One thing I noticed that there are no `setList`/`setMap`/`setSet` methods for explicitly providing the `TypeCodec` implementation to use to serialize/deserialize the element values like there are for the `get`/`set` methods.  Do you think we should add these for convenience of being able to explicitly use a codec without needing to register with `CodecRegistry`?

tolbertam:
Looking good so far!  I'm working on some tests that ensure bind() and getX() methods work when going between different types with a codec registered (i.e. getString(0) on an int column will work if you have the appropriate codec registered), which I should finish up tomorrow.

driver-core/src/main/java/com/datastax/driver/core/AbstractAddressableByIndexData.java
olim7t, line 269: > "Here's an object, I have no idea what this is, but let the driver figure out what to do with it"

I can't think of a use case where someone would want to throw random objects at the driver. And even so, they can still obtain the codec manually with `CodecRegistry#codecFor(DataType, T)` and then call `set(int, V, TypeCodec<V>)`.

The only reason we initially added setObject was to have a counterpart to getObject. But while getObject has a valid use case (stuff like a generic row logger), setObject does not.

olim7t:
> One thing I noticed that there are no setList/setMap/setSet methods for explicitly providing the TypeCodec implementation 

You would use the standard `set(...)`, passing in a list codec. If you only have the element codec, it's easy to build the list one with `TypeCodec.list(elementCodec)`.

driver-core/src/main/java/com/datastax/driver/core/AbstractAddressableByIndexData.java
tolbertam, line 269: Thanks for the explanation.  Using `CodecRegistry#codecFor(DataType, T)` sounds good, in this case the user needs to know the `DataType`, but they should probably know this anyways, and if not they could use `TableMetadata` to get at that.

tolbertam:
> You would use the standard set(...), passing in a list codec. If you only have the element codec, it's easy to build the list one with TypeCodec.list(elementCodec).

That makes sense, it was not immediately apparent to me, but I don't think we should overload that API any more anyways, especially if that works as an alternative.

tolbertam:
Added a test and completed review, looks good to me :+1:

PR #486, opened 2015-10-14
Fix Cluster.connect with a case-sensitive keyspace (JAVA-950).

olim7t:
Go back to the previous strategy of first initializing a session with no
keyspace and then issuing a USE statement. This guarantees that
SessionManager.poolsState.keyspace always contains a keyspace in the
"internal" format (unquoted and in the correct case), and keeps the
implementation simple.
The downside is that if the keyspace is wrong, we'll open all
connections before realizing the error, but that's an edge case.


driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -333,11 +334,28 @@ public Session connect(String keyspace) {
...
      * Cluster.
      */
-    public ListenableFuture<Session> connectAsync(String keyspace) {
+    public ListenableFuture<Session> connectAsync(final String keyspace) {
         checkNotClosed(manager);
         init();
-        Session session = manager.newSession(keyspace);
-        return session.initAsync();
+        final Session session = manager.newSession();
+        ListenableFuture<Session> sessionInitialized = session.initAsync();
```
tolbertam, line 343: I like this because it now becomes fairly simple that if setting a keyspace fails, the future fails and the session is closed.  This could lead to session initialization that fails for other reasons, i.e. if you hit a client timeout here the session can't be established...but that seems fair anyways..

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
```diff
@@ -126,7 +124,7 @@ public void run() {
             }
             reusedConnection = null;
             connections.add(connection);
-            connectionFutures.add(handleErrors(setKeyspaceAsync(connectionFuture, connection, keyspace), initExecutor));
```
tolbertam, line 129: I did like this the way it was because we know that the keyspace is properly set before using a core connection, but there was still always the reality that any newly spawned connection would need to have it's keyspace set in a blocking fashion which was a problem anyways.  [JAVA-893](https://datastax-oss.atlassian.net/browse/JAVA-893) will solve that though.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -333,11 +334,28 @@ public Session connect(String keyspace) {
      * {@link #closeAsync()}), or as a result of an error while initializing the
      * Cluster.
      */
-    public ListenableFuture<Session> connectAsync(String keyspace) {
+    public ListenableFuture<Session> connectAsync(final String keyspace) {
```
tolbertam, line 339: Would be nice to update the connect / connectAsync javadoc to include a `@throws` doc for InvalidQueryException or something if setting the keyspace fails.

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
tolbertam, line 129: Just did a little testing and it seems that if you have an EventLoopGroup with a thread count < 2_hosts and you can get into a state where a thread is tied up for readTimeout and then it tries to borrow a connection and set a keyspace again...eventually it evens out.  If my ELG is at least 2_hosts threads the problem doesn't manifest.  As mentioned in previous comment we knew this was a problem with non-core connections (or if user changed the keyspace on us), its just more visible now since it affects core connections.

tolbertam:
This implementation seems much more simple/straightforward, so a big :+1: from me!

driver-core/src/test/java/com/datastax/driver/core/AsyncQueryTest.java
```diff
@@ -76,19 +90,20 @@ public void should_init_cluster_and_session_if_needed() throws Exception {
...
                 cluster2.close();
         }
     }
 
-    @Test(groups = "short")
-    public void should_chain_query_on_async_session_init() throws Exception {
-        ListenableFuture<Integer> resultFuture = connectAndQuery(this.keyspace);
+    @Test(groups = "short", dataProvider = "keyspace", enabled = false,
+        description = "disabled because the blocking USE call in the current pool implementation makes it deadlock")
+    public void should_chain_query_on_async_session_init(String keyspace) throws Exception {
```
tolbertam: Won't executing the transform function on a separate Executor solve this problem?  I made this change and it looks like it works ([diff](https://gist.github.com/tolbertam/54df36e66b331eb1a403)) since the function is now executed on a separate executor thread instead of the same thread, so the code proceeds.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
olim7t, line 343: Agreed, and that's the way it used to work in 2.0.10.

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
olim7t, line 129: Yes, setting the keyspace at pool construction was an attempt to work around the blocking USE call in borrow, but as we later found out it could still happen even with a fixed pool size, like when a connection gets replaced because of stream id depletion. [The doc](http://datastax.github.io/java-driver/2.0.11/features/async/#known-limitations) mentions that full async only works with a session with no keyspace set.
I agree that JAVA-893 will be the real fix for this.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
olim7t, line 339: Added it for connect. For connectAsync it will be wrapped in the future.

driver-core/src/test/java/com/datastax/driver/core/AsyncQueryTest.java
olim7t: Yes it is. There's still value in testing with the same executor, but we can do both.

tolbertam:
changes look good to me :+1:

PR #487, opened 2015-10-15
Fix MaterializedViewMetadata when base table name is case sensitive (…

olim7t:
…JAVA-953).


tolbertam:
This looks good to me :+1:

PR #488, opened 2015-10-16
Prevent reverse DNS lookup in HostMock.toString()

tolbertam:
`InetSocketAddress.getHostName()` does a reverse DNS lookup, this is costly and unreliable.   Noticed some of the TopologyStrategyTests taking an unreasonable amount of time in a Windows Openstack environment.  Using `InetSocketAddress.toString()` like `Host` does fixes this.


adutra:
:+1:

tolbertam:
Merging since trivial and speeds up a currently slow test.

PR #489, opened 2015-10-18
Update OSGi example guava versions.  Refine bundle version detection.

tolbertam:
Since Guava 16.0.1 is now required for the driver, updated the OSGi example configurations to reflect this.   Also refined the bundle version detection as it was not working with rc/alpha/beta versions (wasn't replacing .rcX with -rcX for example).

### DIFF
--- a/driver-examples/osgi/src/test/java/com/datastax/driver/osgi/VersionProvider.java
+++ b/driver-examples/osgi/src/test/java/com/datastax/driver/osgi/VersionProvider.java
@@ -15,6 +15,9 @@
  */
 package com.datastax.driver.osgi;
 
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
 import com.jcabi.manifests.Manifests;
 
 /**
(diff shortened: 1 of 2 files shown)


driver-examples/osgi/src/test/java/com/datastax/driver/osgi/VersionProvider.java
```diff
@@ -25,11 +28,23 @@
...
+            String bundleVersion = Manifests.read("Bundle-Version");
+            Matcher matcher = versionPattern.matcher(bundleVersion);
+            if(matcher.matches() && matcher.groupCount() == 2) {
+                String majorVersion = matcher.group(1);
+                // Replace all instances of '.' after the main version with '-' to properly
+                // resolve the correct version.
+                String rest = matcher.group(2).replaceAll("\\.", "-");
+                PROJECT_VERSION = majorVersion + rest;
+            } else {
+                PROJECT_VERSION = Manifests.read("Bundle-Version");
```
adutra, line 46: could be simplified to `PROJECT_VERSION = bundleVersion;`

driver-examples/osgi/src/test/java/com/datastax/driver/osgi/VersionProvider.java
```diff
@@ -25,11 +28,23 @@
...
+
     private static String PROJECT_VERSION;
     static {
         String bundleName = Manifests.read("Bundle-SymbolicName");
         if (bundleName.equals("com.datastax.driver.osgi")) {
-            PROJECT_VERSION = Manifests.read("Bundle-Version").replaceAll("\\.SNAPSHOT", "-SNAPSHOT");
+            String bundleVersion = Manifests.read("Bundle-Version");
+            Matcher matcher = versionPattern.matcher(bundleVersion);
+            if(matcher.matches() && matcher.groupCount() == 2) {
+                String majorVersion = matcher.group(1);
```
adutra, line 40: It seems that we always enter the if block, even for simple versions such as `1.2.3` because in these cases the capturing group `(.*)` captures an empty string.

driver-examples/osgi/src/test/java/com/datastax/driver/osgi/VersionProvider.java
tolbertam, line 46: The problem is that the `Bundle-Version` attribute replaces all dashes (-) with dots (.).  This reverts that so the driver jar can be resolved, unless I'm misunderstanding something.

tolbertam, line 46: Ah sorry, I see what you mean now.  When I was looking at this from the main PR page the comment appeared higher, i'll fix this, thanks!

driver-examples/osgi/src/test/java/com/datastax/driver/osgi/VersionProvider.java
tolbertam, line 40: Good point,  It's possible we could change our version string format entirely, and in that case which seems unlikely, this would break anyways so I agree we don't really need to have conditional logic here.

tolbertam:
Just realized that I should probably rebase this against 3.0 instead of 2.2.  Will close this and open up a new PR

PR #492, opened 2015-10-22
JAVA-965 Improve error handling of non-type 1 UUID used for timeuuid.

tolbertam:
[JAVA-965](https://datastax-oss.atlassian.net/browse/JAVA-965)

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -2,8 +2,9 @@
 
 ### 3.0.0-beta1 (in progress)
 
- - [improvement] JAVA-958: Make TableOrView.Order visible.
- - [improvement] JAVA-968: Update metrics to the latest version.
+- [improvement] Make TableOrView.Order visible (JAVA-958)
+- [improvement] Update metrics to the latest version (JAVA-968)
+- [improvement] Improve error handling for when a non-type 1 UUID is given to bind() on a timeuuid column (JAVA-965)
 
 ### 3.0.0-alpha4
 
--- a/driver-core/src/main/java/com/datastax/driver/core/TypeCodec.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/TypeCodec.java
@@ -860,10 +860,20 @@ public abstract class TypeCodec<T> {
         }
 
         @Override
-        public boolean accepts(Object value) {
-            return value instanceof String && ASCII_PATTERN.matcher((String)value).matches();
+        public ByteBuffer serialize(String value, ProtocolVersion protocolVersion) {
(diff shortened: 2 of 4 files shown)


adutra:
lgtm :+1: could you just add an entry to the changelog?

tolbertam:
Added changeling and squashed.

olim7t:
:+1:

adutra:
I fixed a small typo in the changelog file and force-pushed, I hope you don't mind.
I was wondering if we should amend the `accepts(Object)` method for AsciiCodec: it does a similar check to determine whether the string is pure ASCII or not. Should we let it go instead and throw an error in `serialize()`?

tolbertam:
Good catch on asciicodec.  My opinion is that it should probably behave the same as the change we've made for longcodec here as the error will be more explicit.  What do you think @olim7t ?

adutra:
I am even wondering if we are solving the problem in the right way.
The javadocs of `accepts(Object)` state:

> Similarly, codecs that only accept a partial subset of all possible values, such as TimeUUIDCodec or AsciiCodec, must override this method and manually inspect the object to check if it complies or not with the codec's limitations.

So I think users could legitimately override this method so that a custom codec would refuse objects in a wrong state, but they will never know _why_ it's being rejected.

What if the right solution would be to have the accepts() method return or log the reason why it is not accepting the object?

tolbertam:
> So I think users could legitimately override this method so that a custom codec would refuse objects in a wrong state, but they will never know why it's being rejected.

That's a good point, although if a user is implementing a custom codec with a selective accepts like this, they will probably be aware of what is happening.  Although I think the problem comes when the user who runs into this problem is not the implementor of the codec.  I.e. if a user is using a library providing codecs and runs into those errors.

> What if the right solution would be to have the accepts() method return or log the reason why it is not accepting the object?

I could be somewhat on board with that.  The only issue I see there is if you have an alternate codec registered that works, is it possible that every time you evaluate a value that doesn't satisfy this codec first a message would be logged, then it'd continue on to find a candidate codec?  I can imagine that creating a log of noise.  Although that is still a pretty thin use case. (I.e. it would be weird to write a String <-> ascii codec that deals with non-ascii characters or a timeuuid <-> UUID codec that deals with non-type 1 UUIDs).

tolbertam:
Based on our conversation earlier today we've decided the current behavior in this PR is fine for now and should also be applied to ascii codec.  I'll do that now.

tolbertam:
Added change for AsciiCodec and cleaned up test for timeuuid codec.  Should be good to go, feel free to squash and merge if this looks ok.

PR #494, opened 2015-10-23
Update OSGi example guava versions. Refine bundle version detection.

tolbertam:
Since Guava 16.0.1 is now required for the driver, updated the OSGi example configurations to reflect this. Also refined the bundle version detection as it was not working with rc/alpha/beta versions (wasn't replacing .rcX with -rcX for example).

Original PR: #489


adutra:
lgtm :+1:

PR #496, opened 2015-10-26
JAVA-958: Make TableOrView.Order visible.

adutra:
This commit promotes the enum Order to a public top level class and renames it to ClusteringOrder.


tolbertam:
Looks good to me, :+1:

olim7t:
Same here, good to go.

PR #498, opened 2015-10-26
JAVA-970 Don't complete Future in maybeAddPool until onDown completed.

tolbertam:
For [JAVA-970](https://datastax-oss.atlassian.net/browse/JAVA-970)

Avoids a race in `updateCreatedPools` by not completing `Future` returned in `maybeAddPool` on failure until the `Future` returned from `triggerOnDown` completes.

Note: Did not add a changelog for this yet as wasn't sure if this would be a candidate for 2.0.11.1/2.0.12.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
@@ -354,20 +354,31 @@ class SessionManager extends AbstractSession {
 
                     @Override
                     public void onFailure(Throwable t) {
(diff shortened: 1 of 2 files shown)


tolbertam:
I think this may not completely solve the problem.  If the pool fails to initialize and all connections go down, a `triggerOnDown` is called in signalConnectionFailure, so could possibly have a problem in this case.

tolbertam:
Closing this since it's against an old branch and the fix probably isn't ideal.  I'll leave the branch on [my personal fork](https://github.com/tolbertam/java-driver/tree/java970).

PR #501, opened 2015-10-27
JAVA-846: Provide custom codecs library as an extra module

adutra:
Also includes the following tickets:
- JAVA-742 (JSON codecs)
- JAVA-606 (Java 8 codecs)
- JAVA-565 (Array codecs)
- JAVA-605 (Enum codecs and mapper support)


driver-extras/src/main/java/com/datastax/driver/extras/codecs/ArrayCodecs.java
```diff
@@ -0,0 +1,583 @@
...
+ *         <td>{@code double[]}</td>
+ *     </tr>
+ * </table>
+ *
+ * Codecs in this class can either be registered individually, or,
+ * or they may all
+ * be registered via {@link #withPrimitiveArrayCodecs(CodecRegistry)}.
+ *
+ */
+public abstract class ArrayCodecs {
```
olim7t: We could use sub-packages instead of nesting stuff into abstract classes.

driver-extras/src/main/java/com/datastax/driver/extras/codecs/ArrayCodecs.java
```diff
@@ -0,0 +1,583 @@
...
+    /**
+     * Registers {@link TypeCodec} instances in this class
+     * that handle primitive arrays with the given {@link CodecRegistry}.
+     *
+     * @param registry registry to add codecs to.
+     * @return updated registry with codecs.
+     */
+    public static CodecRegistry withPrimitiveArrayCodecs(CodecRegistry registry) {
+        return registry.register(IntArrayCodec.instance, LongArrayCodec.instance, FloatArrayCodec.instance, DoubleArrayCodec.instance);
+    }
```
olim7t: I'm not a big fan of this, because the client-side syntax is inconsistent:

``` java
registry.register(new MyCustomCodec());
ArrayCodecs.withPrimitiveArrayCodecs(registry);
```

I'd rather have a method that returns an iterable, that we can pass to register:

``` java
registry.register(new MyCustomCodec());
registry.register(ArrayCodecs.all());
```

.travis.yml
```diff
@@ -1,5 +1,8 @@
 language: java
 jdk:
   - openjdk6
+    script: mvn test -DlegacyJdk=true
```
tolbertam: I think you will also want to do:

``` yaml
install: true
```

to bypass mvn install (which mvn test will do anyways).

driver-extras/src/main/java/com/datastax/driver/extras/codecs/ArrayCodecs.java
adutra: After splitting the classes into separate files, it seemed to me that it is not necessary to have such shortcuts.

driver-extras/src/main/java/com/datastax/driver/extras/codecs/ArrayCodecs.java
adutra: Very good idea, done.

olim7t:
`CodecUtils` has two distinct responsibilities: ser/deser stuff, and type token helpers. How about we extract the token methods in a separate class, for instance `TypeTokens`?

olim7t:
Pushed a new commit to fix JAVA-884 (also makes `MapperUDTCodec` extend `AbstractUDTCodec`). Note that this requires JAVA-936 to work with Cassandra 3.0.0 GA, otherwise it should be tested against rc2.

driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/InstantCodec.java
```diff
@@ -0,0 +1,134 @@
...
+            return "NULL";
+        return quote(FORMATTER.format(value));
+    }
+
+    @Override
+    public Instant parse(String value) {
+        if (value == null || value.isEmpty() || value.equalsIgnoreCase("NULL"))
+            return null;
+        // strip enclosing single quotes, if any
+        if (value.charAt(0) == '\'' && value.charAt(value.length() - 1) == '\'')
```
tolbertam: ParseUtils.isQuoted would be good for this.

tolbertam: Ah nevermind, these are single quotes.

driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/LocalDateCodec.java
```diff
@@ -0,0 +1,103 @@
...
+        int unsigned = cint().deserializeNoBoxing(bytes, protocolVersion);
+        int signed = fromUnsignedToSignedInt(unsigned);
+        return EPOCH.plusDays(signed);
+    }
+
+    @Override
+    public String format(LocalDate value) {
+        if (value == null)
+            return "NULL";
+        return "'" + ISO_LOCAL_DATE.format(value) + "'";
```
tolbertam: PraseUtils.quote would be good for this

driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/InstantCodec.java
adutra: Good catch thanks!

driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/LocalDateCodec.java
adutra: same here :)

PR #504, opened 2015-10-28
JAVA-971: Make type codecs invariant.

adutra:
This commit makes TypeCodec.accepts(TypeToken) invariant with respect to the Java type passed as argument.
TypeCodec.accepts(Object) remains covariant with respect to the runtime type of the passed object.
This commit also updates the documentation on custom codecs.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/TypeCodec.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/TypeCodec.java
@@ -508,6 +508,15 @@ public abstract class TypeCodec<T> {
     /**
      * Return {@code true} if this codec is capable of serializing
      * the given {@code javaType}.
+     * <p>
+     * The implementation is <em>invariant</em> with respect to the passed
+     * argument (through the usage of {@link TypeToken#equals(Object)}
+     * and <em>it's strongly recommended not to modify this behavior</em>.
+     * This means that a codec will only ever accept the
+     * <em>exact</em> Java type that it has been created for.
+     * <p>
+     * If the argument represents a Java primitive type, its wrapper type
+     * is considered instead.
      *
      * @param javaType The Java type this codec should serialize from and deserialize to; cannot be {@code null}.
      * @return {@code true} if the codec is capable of serializing
@@ -519,7 +528,7 @@ public abstract class TypeCodec<T> {
         if (javaType.isPrimitive()) {
             javaType = primitiveToWrapperMap.get(javaType);
         }
-        return this.javaType.isAssignableFrom(javaType);
+        return this.javaType.equals(javaType);
     }
 
     /**
@@ -548,13 +557,17 @@ public abstract class TypeCodec<T> {
      * <p>
      * Implementation notes:
      * <ol>
+     * <li>The default implementation is <em>covariant</em> with respect to the passed
+     * argument (through the usage of {@link TypeToken#isAssignableFrom(TypeToken)}
+     * and <em>it's strongly recommended not to modify this behavior</em>.
+     * This means that, by default, a codec will accept
+     * <em>any subtype</em> of the Java type that it has been created for.</li>
      * <li>The base implementation provided here can only handle non-parameterized types;
      * codecs handling parameterized types, such as collection types, must override
      * this method and perform some sort of "manual"
      * inspection of the actual type parameters.</li>
-     * <li>Similarly, codecs that only accept a partial subset of all possible values,
-     * such as {@link TimeUUIDCodec} or {@link AsciiCodec}, must override
-     * this method and manually inspect the object to check if it
+     * <li>Similarly, codecs that only accept a partial subset of all possible values
+     * must override this method and manually inspect the object to check if it
      * complies or not with the codec's limitations.</li>
      * </ol>
      *
@@ -565,7 +578,7 @@ public abstract class TypeCodec<T> {
      */
     public boolean accepts(Object value) {
         checkNotNull(value);
-        return accepts(TypeToken.of(value.getClass()));
+        return this.javaType.isAssignableFrom(TypeToken.of(value.getClass()));
     }
 
     @Override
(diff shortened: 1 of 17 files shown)


driver-core/src/main/java/com/datastax/driver/core/TypeCodec.java
```diff
@@ -508,6 +508,14 @@ public DataType getCqlType() {
     /**
      * Return {@code true} if this codec is capable of serializing
      * the given {@code javaType}.
+     * <p>
+     * The default implementation is <em>invariant</em> with respect to the passed
+     * argument (through the usage of {@link TypeToken#equals(Object)}.
+     * This means that, by default, a codec will only ever accept the
+     * <em>exact</em> Java type that it has been created for.
```
olim7t, line 516: "by default" implies that you can override it. I can't think of any legitimate case where you would do something different in a subclass (without wreaking havoc), so maybe this should be more like "this is invariant and it's strongly recommended not to override it". Or maybe the method should even be final...

features/custom_codecs/README.md
```diff
@@ -374,6 +374,212 @@ Foo<Bar> foo = row.get(0, new TypeToken<Foo<Bar>>(){})
 Foo<Bar> foo = row.get(0, Foo.class);
 ```
 
+### Support for subtype polymorphism
+
+Suppose the following class hierarchy:
+
+```java
+class A {}
+class B extends A {}
```
olim7t: Would be nice to have more expressive class names here as well.

features/custom_codecs/README.md
```diff
@@ -374,6 +374,212 @@ Foo<Bar> foo = row.get(0, new TypeToken<Foo<Bar>>(){})
...
+Another anti-pattern that is strongly discouraged is to register a codec for
+a Java type and another codec for one of its superclasses or superinterfaces.
+In this situation, the registry will choose the first registered one,
+*even if the other one is a better match*.
+
+For example, if there are registered codecs for both `A` and `B`
+(i.e. `ACodec extends TypeCodec<A>` and `BCodec extends TypeCodec<B>`),
+when looking for a codec accepting `A` instances, which codec the registry will choose
+will depend on their registration order. This can lead to unpredictable results
+and runtime errors, and thus should always be avoided.
```
olim7t: Well that's only true if you search for a codec for a value, right? Now that the type-based methods are invariant, B.class will return the BCodec and A.class the ACodec, whatever the order.

features/custom_codecs/README.md
```diff
@@ -374,6 +374,212 @@ Foo<Bar> foo = row.get(0, new TypeToken<Foo<Bar>>(){})
...
+
+#### Strategies for mapping polymorphic data models
```
olim7t: Frankly I would leave out this section altogether. That's a lot to digest for someone reading the doc for the first time, I don't think we should actively encourage polymorphism and if users want to get into it the "one codec per class hierarchy" should be intuitive enough (and the other approach is not really polymorphism, it's just different codecs for different columns).

driver-core/src/main/java/com/datastax/driver/core/TypeCodec.java
adutra, line 516: Good call, I rephrased the javadocs as you suggested.

features/custom_codecs/README.md
adutra: Now using the ubiquitous Animal / Cat example.

features/custom_codecs/README.md
adutra: I removed these last 2 paragraphs altogether.

features/custom_codecs/README.md
adutra: Section removed.

driver-core/src/main/java/com/datastax/driver/core/TypeCodec.java
```diff
@@ -548,13 +557,17 @@ public boolean accepts(DataType cqlType) {
...
+     * This means that, by default, a codec will accept
+     * <em>any subtype</em> of the Java type that it has been created for.</li>
      * <li>The base implementation provided here can only handle non-parameterized types;
      * codecs handling parameterized types, such as collection types, must override
      * this method and perform some sort of "manual"
      * inspection of the actual type parameters.</li>
-     * <li>Similarly, codecs that only accept a partial subset of all possible values,
-     * such as {@link TimeUUIDCodec} or {@link AsciiCodec}, must override
-     * this method and manually inspect the object to check if it
+     * <li>Similarly, codecs that only accept a partial subset of all possible values
```
adutra, line 569: This is actually an oversight from JAVA-965, we shouldn't be mentioning AsciiCodec and TimeUUIDCodec as examples of codecs accepting only certain values, as it is not the case anymore.

features/custom_codecs/README.md
```diff
@@ -374,6 +374,45 @@ Foo<Bar> foo = row.get(0, new TypeToken<Foo<Bar>>(){})
 Foo<Bar> foo = row.get(0, Foo.class);
 ```
 
+### Support for subtype polymorphism
```
tolbertam, line 377: I really like the explanation here.  It is very concise and helps me understand how this works.

features/object_mapper/creating/README.md
```diff
@@ -43,9 +43,9 @@ The class must provide a default constructor, and all fields (except the
 ones annotated `@Transient`) must have corresponding setters and
 getters.
 
-[table]:http://docs.datastax.com/en/drivers/java/2.2/com/datastax/driver/mapping/annotations/Table.html
-[case-sensitive]:http://docs.datastax.com/en/cql/3.2/cql/cql_reference/ucase-lcase_r.html
-[consistency level]:http://docs.datastax.com/en/drivers/java/2.2/com/datastax/driver/core/ConsistencyLevel.html
+[table]:http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/mapping/annotations/Table.html
+[case-sensitive]:http://docs.datastax.com/en/cql/3.1/cql/cql_reference/ucase-lcase_r.html
```
tolbertam: You'll probably encounter this in the rebase, but this has been updated to 3.3 from the upstream branches (2.0/2.1) and that seems to work good too.

tolbertam:
Looks good to me :+1:

PR #507, opened 2015-10-29
Avoid rejected tasks at shutdown (JAVA-914).

olim7t:


driver-core/src/main/java/com/datastax/driver/core/NettyOptions.java
```diff
@@ -183,22 +183,17 @@ public void afterChannelInitialized(SocketChannel channel) throws Exception {
...
      *
      * @param eventLoopGroup the event loop group used by the cluster being closed
      */
     public void onClusterClose(EventLoopGroup eventLoopGroup) {
-        // shutdownGracefully with default parameters employs a quiet period of 2 seconds
-        // where in pre-Netty4 versions, closing a cluster instance was very quick (milliseconds).
-        // Since we close the channels before shutting down the eventLoopGroup,
-        // it is safe to reduce the quiet period to 0 seconds
-        eventLoopGroup.shutdownGracefully(0, 15, SECONDS).syncUninterruptibly();
+        eventLoopGroup.shutdownGracefully().syncUninterruptibly();
```
adutra, line 196: Couldn't this have a negative impact on test durations? I recall having set this to zero because of that, after a brief discussion with @tolbertam.

olim7t, line 196: We could override it in tests.

tolbertam, line 196: It does indeed, here are the noticeable ones I found:
- `SessionTest#closeDuringClusterInitTest` went from ~13 seconds to 20 minutes, 
- `ClusterStressTest#clusters_should_not_leak_connections` from 1minute46 seconds to 38 minutes.
- `ControlConnectionTest# should_randomize_contact_points_when_determining_control_connection` from 1.7 sec to 3 min 42 sec

I'll create a NettyOptions in TestUtils called 'nonGracefulShutdownNettyOptions' (or something like that) and apply it to these three classes in particular.

Each Cluster instance we create and close outside of PerClassSingleNodeCluster will add an extra 2 seconds to each test, but I think that is ok.

tolbertam, line 196: On another note I noticed this AssertionError during closeDuringClusterInitTest:

```
16:03:40     7709692 [pool-44726-thread-1] INFO  com.datastax.driver.core.Cluster  - New Cassandra host /127.0.1.1:11002 added
16:03:42 Exception in thread "globalEventExecutor-1-2598" java.lang.AssertionError
16:03:42    at io.netty.util.concurrent.AbstractScheduledEventExecutor.pollScheduledTask(AbstractScheduledEventExecutor.java:83)
16:03:42    at io.netty.util.concurrent.GlobalEventExecutor.fetchFromScheduledTaskQueue(GlobalEventExecutor.java:110)
16:03:42    at io.netty.util.concurrent.GlobalEventExecutor.takeTask(GlobalEventExecutor.java:95)
16:03:42    at io.netty.util.concurrent.GlobalEventExecutor$TaskRunner.run(GlobalEventExecutor.java:226)
16:03:42    at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:137)
16:03:42    at java.lang.Thread.run(Thread.java:701)
```

This caused the test to hang (presumably in Cluster.close, I think?)

I've never seen this before, but looks like it's a known issue (netty/netty#4357) and is fixed in netty 4.0.33.

tolbertam:
Updated these tests to use custom NettyOptions with 0 second quiet period.  Will see how CI ran Friday morning.

driver-core/src/main/java/com/datastax/driver/core/NettyOptions.java
tolbertam, line 196: On further investigation, this doesn't seem to be tied to this change, i'm encountering the error on separate branches that don't have this.  Seems to have started when upgrading from JDKu60->65. Given when the fix was in netty that doesn't seem to corraborate, but i'm going to revert to see if that makes any difference.

PR #508, opened 2015-10-29
Refactor SSLOptions API (JAVA-841, JAVA-948).

olim7t:
Make SSLOptions more generic to allow use of Netty implementation, or SSLEngine customization when default JDK implementation is used.
By default, don't restrict the ciphers presented by the client.
Add tests with Netty implementation backed by OpenSSL.

Upgraded Netty to 4.0.32 to use the new SslContextBuilder.


driver-core/src/main/java/com/datastax/driver/core/NettySSLOptions.java
```diff
@@ -0,0 +1,43 @@
...
+ * <p>
+ * Netty has the ability to use OpenSSL if available, instead of the JDK's built-in engine. This yields better performance.
+ */
+public class NettySSLOptions implements SSLOptions {
+    private final SslContext context;
+
+    /**
+     * Create a new instance from a given context.
+     *
+     * @param context the Netty context. See the static {@code SslContext.newClientContext()} methods to build an instance.
```
olim7t: Need to fix this to refer to builder instead.

driver-core/src/main/java/com/datastax/driver/core/NettySSLOptions.java
```diff
@@ -0,0 +1,43 @@
...
+ * <p>
+ * Netty has the ability to use OpenSSL if available, instead of the JDK's built-in engine. This yields better performance.
+ */
+public class NettySSLOptions implements SSLOptions {
+    private final SslContext context;
+
+    /**
+     * Create a new instance from a given context.
+     *
+     * @param context the Netty context. {@code SslContextBuilder.forClient()} provides a fluent API to build it.
```
adutra, line 33: Nit: use `@link` instead of `@code` so that users can navigate to that methd.

driver-core/src/test/java/com/datastax/driver/core/SSLTestBase.java
```diff
@@ -74,35 +76,61 @@ protected void connectWithSSLOptions(SSLOptions sslOptions) throws Exception {
...
-        return new SSLOptions(sslContext, SSLOptions.DEFAULT_SSL_CIPHER_SUITES);
+                SSLContext sslContext = SSLContext.getInstance("TLS");
+                sslContext.init(kmf != null ? kmf.getKeyManagers() : null, tmf != null ? tmf.getTrustManagers() : null, new SecureRandom());
+
+                return JdkSSLOptions.builder().withSSLContext(sslContext).build();
+
+            case NETTY_OPENSSL:
+                SslContextBuilder builder = SslContextBuilder
+                    .forClient()
+                    .sslProvider(OPENSSL)
```
adutra, line 123: Just wondering: does the test risk to fail if openssl is not installed/available?

driver-core/src/main/java/com/datastax/driver/core/JdkSSLOptions.java
```diff
@@ -0,0 +1,115 @@
...
+        this.cipherSuites = cipherSuites;
+    }
+
+    @Override
+    public SslHandler newSSLHandler(SocketChannel channel) {
+        SSLEngine engine = newSSLEngine();
+        return new SslHandler(engine);
+    }
+
+    protected SSLEngine newSSLEngine() {
```
adutra: Since this method is protected, it will appear in the javadocs, so it would be better if there was a short description of situations where users could override it (e.g. hostname verification, see JAVA-716).

features/ssl/README.md
```diff
@@ -0,0 +1,180 @@
...
+  .withSSLContext(context)
+  .build();
+
+Cluster cluster = Cluster.builder()
+  .addContactPoint("127.0.0.1")
+  .withSSL(sslOptions)
+  .build();
+```
+
+Note that you can also extend the class and override `newSSLEngine()` if
```
adutra: Maybe add a link to the method javadoc?

changelog/README.md
```diff
@@ -2,8 +2,12 @@
 
 ### 3.0.0-beta1 (in progress)
 
- - [improvement] JAVA-958: Make TableOrView.Order visible.
- - [improvement] JAVA-968: Update metrics to the latest version.
+- [improvement] JAVA-958: Make TableOrView.Order visible.
+- [improvement] JAVA-968: Update metrics to the latest version.
+- [improvement] JAVA-841: Refactor SSLOptions API.
+- [improvement] JAVA-948: Don't limit cipher suites by default.
+- [improvement] JAVA-917: Document SSL configuration.
```
adutra, line 15: I think that this resolves JAVA-716 too.

features/ssl/README.md
```diff
@@ -0,0 +1,180 @@
...
+you need specific configuration on the `SSLEngine` (for example hostname
+verification).
+
+#### Netty
+
+[NettySSLOptions] allows you to use Netty's `SslContext` instead of
+the JDK directly. The advantage is that Netty can use OpenSSL if
+available, which provides better performance.
+
+##### Converting your client certificates for OpenSSL
```
adutra, line 131: This section is a bit out of scope from the driver's perspective imho, but it's short, so let's leave it.

adutra:
:+1: apart from very minor remarks (none of which being a blocker).

driver-core/src/main/java/com/datastax/driver/core/NettySSLOptions.java
olim7t, line 33: In fact that only works in some IDEs. We don't link our Javadoc to external libraries, so external links cause a warning during generation and are output as simple `code` blocks.

driver-core/src/test/java/com/datastax/driver/core/SSLTestBase.java
olim7t, line 123: Netty throws an `Error` if it's not available ([source](https://github.com/netty/netty/blob/master/handler/src/main/java/io/netty/handler/ssl/OpenSslEngine.java#L208)), although I've not tested this.

driver-core/src/main/java/com/datastax/driver/core/JdkSSLOptions.java
olim7t: Good point. I also realized that this method should take the channel so that you can extract the remote address and port.

changelog/README.md
olim7t, line 15: I've marked 716 as duplicate to simplify things.

olim7t:
I think I've addressed all feedback so this should be ready to go. Let me know before merging so that I can squash the commits appropriately.

olim7t:
Force-pushed to clean up history and rebase on current 3.0.

boliza:
class JdkSSLOptions method : newSSLEngine(SocketChannel channel)

this parameter : channel never used

driver-core/src/main/java/com/datastax/driver/core/JdkSSLOptions.java
```diff
@@ -0,0 +1,131 @@
...
+     * Creates an SSL engine each time a connection is established.
+     *
+     * <p>
+     * You might want to override this if you need to fine-tune the engine's configuration
+     * (for example enabling hostname verification).
+     *
+     * @param channel the Netty channel for that connection.
+     * @return the engine.
+     */
+    protected SSLEngine newSSLEngine(SocketChannel channel) {
```
boliza, line 69: this parameter : channel never used

olim7t, line 69: It's not used in this default implementation, but it can be useful if you override the method to do something else. In particular, that's how you would get access to the remote address and port.

boliza, line 69: Got it.thank you

PR #510, opened 2015-11-02
JAVA-917: Document SSL configuration.

olim7t:
Planning this for 2.0.12 since it required little effort to backport to older branches.
I'll do the 3.0-specific documentation on #508.

### DIFF
--- a/features/ssl/README.md
+++ b/features/ssl/README.md
@@ -1,0 +1,180 @@
+## SSL
+
+You can secure traffic between the driver and Cassandra with SSL. There
+are two aspects to that:
+
+* **client-to-node encryption**, where the traffic is encrypted, and the
+  client verifies the identity of the Cassandra nodes it connects to;
+* optionally, **client certificate authentication**, where Cassandra
+  nodes also verify the identity of the client.
+
+This section describes the driver-side configuration; it assumes that
+you've already configured SSL in Cassandra:
+
+* [the Cassandra documentation](http://docs.datastax.com/en/cassandra/2.0/cassandra/security/secureSSLClientToNode_t.html)
+  covers a basic approach with self-signed certificates, which is fine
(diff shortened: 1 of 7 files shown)


features/ssl/README.md
```diff
@@ -0,0 +1,158 @@
...
+[configured SSL in Cassandra](http://docs.datastax.com/en/cassandra/2.0/cassandra/security/secureSSLClientToNode_t.html).
+
+### Preparing the certificates
+
+#### Client truststore
+
+This allows the client to verify the identity of all the Cassandra nodes
+it connects to.
+
+As part of configuring Cassandra, you should already have
```
tolbertam: I think it would be good to either change this sentence to also indicate that you could just trust the CA that signed the certificates on the nodes or reduce the terminology that each nodes certificate would be trusted.  I don't think we need to cover end to end how to configure the C\* side of things (so just linking to the docs as done here is fine).

Unfortunately the datastax docs demonstrate creating a self-signed certificate on each node, but that does not work well as each node needs to be trusted individually and is hard to manage, especially in the client case (i.e. you'd have to reload your truststore every time you add a new node).   [These slides](http://www.slideshare.net/zznate/hardening-cassandra-for-compliance-or-paranoia/31) and accompanying [blog post](http://thelastpickle.com/blog/2015/09/30/hardening-cassandra-step-by-step-part-1-server-to-server.html) define a good approach for that (only focuses on internode encryption though).

tolbertam:
Added some tests for connecting with System properties, that fix seems to work great :+1:

features/ssl/README.md
olim7t: Great point, updated the doc.

tolbertam:
Looks great, big :+1: on the changes in the last commit :)

PR #514, opened 2015-11-05
JAVA-978: Remove double quote - return NULL when entityMapper.getTableMetaSpace()

regis-leray:
```
MappingManager manager = new MappingManager(session);
Mapper<Post> m = manager.mapper(Post.class);

m.getTableMetadata() // is returning NULL
```

We are double quoting the name of the table, and we are not able to find in the table map of the Keyspace, which is a bug

### DIFF
--- a/driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
+++ b/driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
@@ -75,7 +75,7 @@ public class Mapper<T> {
         this.mapper = mapper;
 
         KeyspaceMetadata keyspace = session().getCluster().getMetadata().getKeyspace(mapper.getKeyspace());
-        this.tableMetadata = keyspace == null ? null : keyspace.getTable(Metadata.quote(mapper.getTable()));
+        this.tableMetadata = keyspace == null ? null : keyspace.getTable(mapper.getTable());
 
         this.protocolVersion = manager.getSession().getCluster().getConfiguration().getProtocolOptions().getProtocolVersionEnum();
         this.mapOneFunction = new Function<ResultSet, T>() {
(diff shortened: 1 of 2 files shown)


driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
```diff
@@ -75,7 +75,7 @@
         this.mapper = mapper;
 
         KeyspaceMetadata keyspace = session().getCluster().getMetadata().getKeyspace(mapper.getKeyspace());
-        this.tableMetadata = keyspace == null ? null : keyspace.getTable(Metadata.quote(mapper.getTable()));
+        this.tableMetadata = keyspace == null ? null : keyspace.getTable(mapper.getTable());
```
regis-leray, line 78: `mapper.getTable()` is already quoting the name

olim7t:
Good catch, your fix looks good to me. Some minor remarks:
- `should_return_table_metadata` would be a better name for the test
- the commit message could be improved, mentions "getMetaSpace" and it's not clear whether "throw NPE" is the fix or the initial problem. Suggestion: "Fix quoting issue that caused Mapper.getTableMetadata() to return null".

Could you do these changes and rebase on the current 2.1? We're planning to release 2.1.9 soon and I'd like to include this, so if you don't have time today I'll do the changes myself (keeping proper attribution of the commit).

olim7t:
Cherry-picked as 97363ce, thanks.

regis-leray:
@olim7t  Thanks. i was planning to do it.
I would like to know if this fix will be included in the branch 2.2 ?

olim7t:
Yes, the fix will be merged to more recent branches.
However, we plan to retire the 2.2 branch in favor of 3.0, I'll send an announcement about that on the mailing list next week.

tolbertam:
Created [JAVA-978](https://datastax-oss.atlassian.net/browse/JAVA-978) for this for tracking.

PR #515, opened 2015-11-07
Add isolated test group and profile for tests requiring their own JVM.

tolbertam:
This is useful for tests that touch static members or require a 'clean' environment.

Applied to currently failing SSL tests that rely on SSLContext.getDefault which depends on System properties and is only ever loaded once.

Also configure failure parameters such that OSGi integration tests may run despite any test failures.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -10,12 +10,24 @@ cassandra:
 build:
   - type: maven
     version: 3.2.5
-    goals: clean compile test --fail-never -Plong
+    goals: verify --fail-never -Plong
     properties: |
       com.datastax.driver.TEST_BASE_NODE_WAIT=120
       com.datastax.driver.NEW_NODE_DELAY_SECONDS=100
       cassandra.version=$CCM_CASSANDRA_VERSION
       ccm.java.home=$CCM_JAVA_HOME
+      failIfNoTests=false
+      maven.test.failure.ignore=true
+  - type: maven
+    version: 3.2.5
+    goals: verify --fail-never -Pisolated
+    properties: |
+      com.datastax.driver.TEST_BASE_NODE_WAIT=120
+      com.datastax.driver.NEW_NODE_DELAY_SECONDS=100
+      cassandra.version=$CCM_CASSANDRA_VERSION
+      ccm.java.home=$CCM_JAVA_HOME
+      failIfNoTests=false
+      maven.test.failure.ignore=true
   - xunit:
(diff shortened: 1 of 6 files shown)


tolbertam:
Waiting to see if this works well in CI before removing wip tag.

build.yaml
```diff
@@ -10,12 +10,24 @@ cassandra:
...
   - type: maven
     version: 3.2.5
-    goals: clean compile test --fail-never -Plong
+    goals: clean compile test integration-test --fail-never -Plong
     properties: |
       com.datastax.driver.TEST_BASE_NODE_WAIT=120
       com.datastax.driver.NEW_NODE_DELAY_SECONDS=100
       cassandra.version=$CCM_CASSANDRA_VERSION
       ccm.java.home=$CCM_JAVA_HOME
+      failIfNoTests=false
```
tolbertam, line 19: The only reasons I have 'failIfNoTests' and 'maven.test.failure.ignore' present here is so the build can continue and run other tests.  For example, if the 'driver-core' build fails tests, the artifacts aren't created and thus integration tests fail.  We'd normally not want to use these flags for build/releases but for the intent of CI testing it works.

build.yaml
```diff
@@ -10,12 +10,24 @@ cassandra:
...
+      maven.test.failure.ignore=true
+  - type: maven
```
tolbertam, line 21: Not too happy about having to duplicate some of the configuration here.  The main reason is that while you can run multiple profiles in a single maven run, if goals share the same execution id only 1 is ran.  This keeps that less complex.

tolbertam:
This seems to be working good, think it's good to go unless any reservations on the approach.

build.yaml
```diff
@@ -10,12 +10,24 @@ cassandra:
 build:
   - type: maven
     version: 3.2.5
-    goals: clean compile test --fail-never -Plong
+    goals: clean compile test integration-test --fail-never -Plong
```
adutra: I think it is recommended to invoke the verify phase instead, like in `clean verify --fail-never -Plong`

adutra:
:+1:

build.yaml
tolbertam: Cool, so that'd run both test and integration-test phase (as well as any other checks?).  I'll give that a try, thanks!

tolbertam:
Looks like verify works great, thanks for the tip :).  I think this is now good to go.

PR #270, opened 2015-01-14
BoundStatment.toString (JAVA-115)

emilyselwood:
Add a toString method to BoundStatement.

BuiltStatement has one but BoundStatement doesn't and falls back to Objects.toString() which is not very helpful when logging errors.

Only thing I am not sure about is the use of StandardCharsets. Do we need to support versions of java less than 1.7?

PR #516, opened 2015-11-09
JAVA-855: Make RegularStatement implement GettableByData and SettableByData

adutra:
Note that this PR depends on JAVA-983 (#517) to make QueryBuilder tests pass. 

While JAVA-936 is not merged, please test this PR against C\* commit b761b86.

RegularStatement also exposes some more methods inspired by BoundStatement:
- bind
- isSet, isUnset, unset
- setToken, setPartitionKeyToken

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/QueryLogger.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/QueryLogger.java
@@ -15,6 +15,7 @@
  */
 package com.datastax.driver.core;
 
+import java.io.Serializable;
 import java.nio.ByteBuffer;
 import java.util.List;
 
@@ -22,8 +23,11 @@ import static java.util.concurrent.TimeUnit.NANOSECONDS;
 
 import com.google.common.annotations.Beta;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Objects;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.datastax.driver.core.RegularStatement.ValueDefinition;
 
 /**
  * A configurable {@link LatencyTracker} that logs all executed statements.
@@ -647,7 +651,9 @@ public abstract class QueryLogger implements LatencyTracker {
         boolean showParameterValues = logger.isTraceEnabled();
         if (showParameterValues) {
             StringBuilder params = new StringBuilder();
-            if (statement instanceof BoundStatement) {
+            if (statement instanceof RegularStatement) {
+                appendParameters((RegularStatement)statement, params, maxLoggedParameters);
+            } else if (statement instanceof BoundStatement) {
                 appendParameters((BoundStatement)statement, params, maxLoggedParameters);
             } else if (statement instanceof BatchStatement) {
                 BatchStatement batchStatement = (BatchStatement)statement;
@@ -655,6 +661,8 @@ public abstract class QueryLogger implements LatencyTracker {
                 for (Statement inner : batchStatement.getStatements()) {
                     if (inner instanceof BoundStatement) {
                         remaining = appendParameters((BoundStatement)inner, params, remaining);
+                    } else if (inner instanceof RegularStatement) {
+                        remaining = appendParameters((RegularStatement)inner, params, remaining);
                     }
                 }
             }
@@ -676,6 +684,10 @@ public abstract class QueryLogger implements LatencyTracker {
         } else if (statement instanceof BoundStatement) {
             int boundValues = ((BoundStatement)statement).wrapper.values.length;
             sb.append("[" + boundValues + " bound values] ");
+        } else if (statement instanceof RegularStatement) {
+            int values = ((RegularStatement)statement).valuesCount();
+            if (values > 0)
+                sb.append("[" + values + " bound values] ");
         }
 
         append(statement, sb, maxQueryStringLength);
@@ -702,7 +714,7 @@ public abstract class QueryLogger implements LatencyTracker {
             if (remaining == -1) {
                 numberOfLoggedParameters = numberOfParameters;
             } else {
-                numberOfLoggedParameters = remaining > numberOfParameters ? numberOfParameters : remaining;
+                numberOfLoggedParameters = Math.min(remaining, numberOfParameters);
                 remaining -= numberOfLoggedParameters;
             }
             for (int i = 0; i < numberOfLoggedParameters; i++) {
@@ -710,7 +722,9 @@ public abstract class QueryLogger implements LatencyTracker {
                     buffer.append(" [");
                 else
                     buffer.append(", ");
-                buffer.append(String.format("%s:%s", metadata.getName(i), parameterValueAsString(definitions.get(i), statement.wrapper.values[i])));
+                buffer.append(String.format("%s:%s",
+                    metadata.getName(i),
+                    parameterValueAsString(definitions.get(i).getType(), statement.wrapper.values[i])));
             }
             if (numberOfLoggedParameters < numberOfParameters) {
                 buffer.append(FURTHER_PARAMS_OMITTED);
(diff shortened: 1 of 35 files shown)


driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
```diff
@@ -43,31 +111,926 @@ protected RegularStatement() {}
...
+    @Override
+    public <V> RegularStatement set(String name, V v, TypeToken<V> targetType) {
+        // we must treat tuples and UDTs separately
+        // as they convey their own specific CQL types as well
+        DataType cqlType = null;
+        if (v instanceof TupleValue)
+            cqlType = ((TupleValue)v).getType();
+        if (v instanceof UDTValue)
+            cqlType = ((UDTValue)v).getType();
+        return set(name, v, codecRegistry.codecFor(cqlType, targetType));
```
olim7t, line 930: It's too bad that we forget `targetType` after that point. We could propagate it all the way to store it in the `Value` instance. That would even make the type check in `as()` (that I suggested in my other comment) much better. A raw value would be one that doesn't have a target type.

olim7t:
Pushed a new commit that only uses original values for `getObject`, and stores the inferred CQL type. This allows getting a value with a different type than the one it was set with (provided that there is a corresponding codec), which is more consistent with BoundStatement.
And yes I also made Value non-generic ;-)

Consider this a proposal, feedback is welcome.

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
```diff
@@ -42,32 +70,1143 @@ protected RegularStatement() {}
...
+            case POSITIONAL:
+                if (values.isEmpty())
+                    return Collections.emptyList();
+
+                // If there are gaps in the indices, we need to mark them unset
+                int last = (Integer)(values.lastEntry().getKey());
+                bbs = newArrayListWithCapacity(last + 1);
+                for (int i = 0; i <= last; i++) {
+                    Value value = values.get(i);
+                    bbs.add(value == null ? BoundStatement.UNSET : value.bytes);
```
adutra: UNSET should maybe be declared at RegularStatement level now.

driver-core/src/main/java/com/datastax/driver/core/QueryLogger.java
```diff
@@ -647,7 +651,9 @@ protected void logQuery(Statement statement, Exception exception, Logger logger,
         boolean showParameterValues = logger.isTraceEnabled();
```
adutra, line 650: I was wondering if we could extract `logQuery` and its dependent methods to another place and make it public (maybe `RegularStatement.toString()`? Or some helper class?). I can imagine users benefitting from this when they need to log statements in their own code. Logging parameters manually is tedious and can be dangerous if there are too many of them etc. and this method takes care of all this. E.g. in pull request #270 (JAVA-115) the author tries to do exactly that, but without all the guards and checks we have here.

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
```diff
@@ -42,32 +70,1143 @@ protected RegularStatement() {}
...
+            assert thatIndex >= 0; // from usesPositionalValues check above
+            int thisIndex = offset + thatIndex;
+            setInternal(thisIndex, new Value(new ValueDefinition(thisIndex, definition.getType()), that.getBytesUnsafe(thatIndex)));
+        }
+    }
+
+    /**
+     * Refreshes the statement's values.
+     * <p>
+     * This method is intended implementations that generate the values dynamically. It's called at the beginning of each method that
```
adutra: typo: "is intended _for_ implementations".

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
```diff
@@ -42,32 +70,1143 @@ protected RegularStatement() {}
...
+     * Both statements must use positional values (or no values at all). If this statement already contains values, the new
+     * values will be inserted starting at the index following the current max index.
+     *
+     * @param that the other statement.
+     */
+    protected void copyValues(RegularStatement that) {
+        if (!that.hasValues())
+            return;
+
+        Preconditions.checkArgument(parameterMode != ParameterMode.NAMED && that.usesPositionalValues());
```
adutra, line 1050: Why can't named parameters be copied?

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
```diff
@@ -42,32 +70,1143 @@ protected RegularStatement() {}
...
+        // The serialized form that will be sent to Cassandra alongside the query
+        private final ByteBuffer bytes;
+
+        private Value(ValueDefinition definition, ByteBuffer bytes) {
+            this.definition = definition;
+            this.bytes = bytes;
+        }
+
+        private <V> V as(TypeToken<V> javaType) {
+            if (OBJECT_TYPE.equals(javaType)) { // we're handling a getObject() call
```
adutra: `OBJECT_TYPE == javaType` would be safer I think. Otherwise you cannot tell if the user has called `get(..., Object.class)` (which, granted, would result in a CNFE).

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
```diff
@@ -42,32 +70,1143 @@ protected RegularStatement() {}
...
+                return with(codecFor(javaType));
+            }
+        }
+
+        private <V> V with(TypeCodec<V> codec) {
+            return codec.deserialize(bytes, protocolVersion);
+        }
+
+        private boolean isNull() {
+            return bytes == null || bytes.remaining() == 0;
```
adutra: That's not necessarily true for all CQL types, e.g. for varchars, `bytes == null`-> `null` but `bytes.remaining() == 0` -> empty string. I think it would be safer to store a flag to tell if the value was null or not, or better yet, let the codec decide: `return  with(codecFor(definition.type)) == null`. Let's keep in mind that a codec can choose never to return nulls, e.g. `OptionalCodec`.

driver-core/src/test/java/com/datastax/driver/core/querybuilder/QueryBuilderTest.java
```diff
@@ -865,24 +864,65 @@ public void should_quote_column_names_with_escaped_quotes() {
...
+    @CassandraVersion(major = 2.1)
+    public void should_not_attempt_to_serialize_function_calls_in_collections() {
+        BuiltStatement query = builder.insertInto("foo").value("v", Sets.newHashSet(fcall("func", 1)));
+        assertThat(query.getQueryString()).isEqualTo("INSERT INTO foo (v) VALUES ({func(1)});");
+        assertThat(query.getValueDefinitions()).isNullOrEmpty();
+    }
+
+    @Test(groups = "unit")
+    @CassandraVersion(major = 2.1)
+    public void should_not_attempt_to_serialize_bind_markers_in_collections() {
```
adutra: I realized recently when merging JAVA-983 that the query generated by this test is invalid, because bind markers are not allowed in collection literals: you need to put one single bind marker for the whole collection. This test has thus been deleted.

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
olim7t: Moving it to CBUtil.

driver-core/src/main/java/com/datastax/driver/core/QueryLogger.java
olim7t, line 650: Why not. That would be another ticket though.

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
olim7t, line 1050: Because of name collisions.

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
olim7t: That's a good point, but my plan was to move OBJECT_TYPE to TypeTokens and make it public, which in theory allows users to pass it to get(). So it might be better to use `null` to indicate getObject (as you initially did).

olim7t: Actually we can call asObject directly, which eliminates the problem.

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
olim7t: In the context of asObject(), I agree: we should let the codec decide how to deserialize the null.

Value.isNull is also called from RegularStatement.isNull, but in that case I think what we want is just to test for `bytes == null`, which means "a NULL will be sent at the protocol level" (that's what BoundStatement does).

olim7t:
Force-pushed to rebase on current 3.0 and clean up history. All the comments so far have been adressed.

olim7t:
Sorry the first commit's message is wrong and I should have kept Alexandre as the main contributor. Will try to fix that and push again.

olim7t:
We've decided not to fix this ticket.

Copied from JIRA:

> After careful consideration, we've decided not to merge this feature:
> - it adds significant complexity to `RegularStatement`;
> - it makes simple statement look more sophisticated than they really are (one-off queries that shouldn't be used in most cases). We believe that blurring the line between simple and prepared statement will be confusing to new users.
> 
> The benefits (named values, being able to set values after the statement is created, and retrieve them with typed getters) are minimal in comparison.

PR #517, opened 2015-11-10
JAVA-983: [3.0] QueryBuilder cannot handle collections containing function calls

adutra:
I'm opening this PR against 3.0 but it should be backported to 2.1 as well.


driver-core/src/main/java/com/datastax/driver/core/querybuilder/Insert.java
```diff
@@ -56,7 +56,7 @@ StringBuilder buildQueryString(List<Object> variables) {
         if (keyspace != null)
             Utils.appendName(keyspace, builder).append('.');
         Utils.appendName(table, builder);
-        builder.append('(');
+        builder.append(" (");
```
adutra, line 59: Not required, but imho the resulting string looks better with this extra whitespace between the table name and the columns list.

tolbertam:
Shouldn't there also be some kind of handling for Tuples and UDTs?

adutra:
I don't think so; TupleValue and UDTValue instances cannot contain special values such as function calls; their API is that of SettableData and as such setting a UDT or tuple field to a function call is not possible.

tolbertam:
Ah I see, thanks for clearing that up.  For both UDTValue and TupleValue the types must be expressed beforehand, so there would be no way to say 'fcall("minTimeUuuid", "2013-02-03 10:00+0000")' in a UDTValue or TupleValue containing a timeuuid.

tolbertam:
Looks good to me :+1:

driver-core/src/main/java/com/datastax/driver/core/querybuilder/Insert.java
olim7t, line 59: Agreed, that's also how insert statements are formatted in the CQL reference doc.
IMO this should stay as a separate commit when we eventually merge the PR.

olim7t:
:+1:
I'm just wondering if we shouldn't fix this first on 2.1 and then merge downstreams, backporting commits is always a bit clunky.
The same applies to #523.

adutra:
Squashed and rebased on top of current 3.0.

PR #519, opened 2015-11-12
JAVA-982: Add isSerial() to distinguish between query and serial CLs.

tolbertam:
For [JAVA-982](https://datastax-oss.atlassian.net/browse/JAVA-982).  This just adds a convenience method `isSerial` to `ConsistencyLevel`.   In the general case this isn't particularly useful except perhaps for programatic and library use.

Not an urgent request, doesn't need to be in upcoming release(s).


olim7t:
Closing since this was covered in #588.

PR #520, opened 2015-11-12
Refactor Load Balancing Policy Tests to use Scassandra.

tolbertam:
- Pull tests from LoadBalancingPolicyTest into individual test classes
  for each policy being tested.
- Rename SCassandraCluster -> ScassandraCluster.
- Add capability to prime keyspace metadata to ScassandraCluster.
- Update DCAwareRoundRobinPolicy tests to use builder based construction
  and HostFilterPolicy.
- Update more tests to use TestUtils#nonQuietClusterCloseOptions.

Cuts down integration suite by ~30 minutes.


tolbertam:
There's a few more tests that i'd like to refactor (ConsistencyTest primarily), but will do as a separate PR.

tolbertam:
Adding wip tag as there is sometimes a timing issue in CI waiting for nodes to go down for some reason (maybe something to do with debouncing)

tolbertam:
Removing [wip] tag.  This is running dependably now and it cuts 30-50 minutes off of a full integration test run while adding more tests.  There are a few rare cases where some minor issues ([JAVA-994](https://datastax-oss.atlassian.net/browse/JAVA-994), [JAVA-995](https://datastax-oss.atlassian.net/browse/JAVA-995) and [JAVA-997](https://datastax-oss.atlassian.net/browse/JAVA-997)) cause tests to fail, but it's pretty difficult to reproduce (like < 1% in the jenkins environment).  I've prioritized stopping Hosts that have the control connection last in tests to minimize those issues.

The main tests that inconsistently fail that need to be addressed now are ConsistencyTest and RecommissionedNodeTest which I will cover in the future in another PR.

PR #521, opened 2015-11-12
JAVA-444: Add process information to UUID LSB

pmaedel:
[Java-444](https://datastax-oss.atlassian.net/browse/JAVA-444)
For relevant discussion see also respective ticket for C*: [CASSANDRA-7925](https://issues.apache.org/jira/browse/CASSANDRA-7925)

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/utils/UUIDs.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/utils/UUIDs.java
@@ -91,6 +91,7 @@ public final class UUIDs {
             update(digest, props.getProperty("os.arch"));
             update(digest, props.getProperty("os.name"));
             update(digest, props.getProperty("os.version"));
+            update(digest, getProcessPiece());
 
             byte[] hash = digest.digest();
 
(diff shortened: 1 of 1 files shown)


olim7t:
Yes, I agree that we need to address this.

Since using getRuntimeMXBean is not an "official" way to get the PID, I'd like to add the following:
- extract the actual number from the bean name, and log it at DEBUG level
- if the bean name has an unexpected format or an exception is thrown, fallback to the random number, but also log a WARN message
- allow the pid to be overridden by a system property (e.g. `com.datastax.driver.PID`). It would probably be a good idea to also mention that property in the warn message above.

@pmaedel, do you have time to take a look at this?

pmaedel:
Well there is no single official way to get the PID, up until [Java 9](http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8003488)
With that in mind, I think we should go for the same solution as C*
@olim7t, could you add your proposal to [CASSANDRA-7925](https://issues.apache.org/jira/browse/CASSANDRA-7925) as well?

olim7t:
C\* uses [Sigar](https://github.com/hyperic/sigar) to get the PID, amongst other things. That works well for them because Cassandra is a standalone program, they just need to include all Sigar files in the binary distribution and it's easy to reference them.

In our case it's a bit more complicated because the driver uses Maven for dependencies. Sigar is [present](http://search.maven.org/#artifactdetails%7Corg.fusesource%7Csigar%7C1.6.4%7Cjar) in Maven central, but apparently it was deployed by a third party and I see a number of problems with it:
- not all platform-specific binaries are present (compare the list to what's present [in Cassandra](https://github.com/apache/cassandra/tree/trunk/lib/sigar-bin))
- some files in the repo don't follow Maven naming conventions, so I don't see how they could be referenced from our POM. For example: `http://repo1.maven.org/maven2/org/fusesource/sigar/1.6.4/libsigar-universal64-macosx-1.6.4.dylib`
- there's also a `sigar-native.jar` file containing all the platform-specific binaries, but it's not usable as-is, i.e. we can't just include it in our POM and expect Sigar to load from it. Apparently you're supposed to extract the files in a directory and pass the path as a system property, again something that's fine for a standalone program but not for a library like the driver.

Long story short, I think Sigar is too complicated for the driver, the getRuntimeMXBean solution should be good enough for most cases, provided we take the few precautions I outlined in my previous message.

pmaedel:
@olim7t Thank you for elaborating. I agree with your reasoning and will look into the proposed adjustments asap.

adutra:
Closing in favor of #593.

PR #522, opened 2015-11-12
JAVA-851: Added utility methods for generating UUIDs for specifiable timestamps

pmaedel:
[JAVA-851](https://datastax-oss.atlassian.net/browse/JAVA-851)
Provided fix might not be the best solution performance-wise but does satisfy the regular use case of being able to create UUIDs with user provided timestamps (presumably ones from the past)


olim7t:
Closing since we decided improving the documentation would be sufficient to solve JAVA-851.

PR #523, opened 2015-11-12
JAVA-819: Driver shouldn't retry on client timeout if statement is not idempotent.

adutra:
This commit prevents a retry from happening in the event of a client timeout if the statement is not idempotent.


tolbertam:
If a Connection gets defuncted (i.e. Channel associated with Connection closes).   closeAsync().force() is called which should call `errorOutAllHandler` in `CloseConnectionFuture#force`.

errorOutHandler cancels all the timeouts and invokes onException with a TransportException on any inflight request handler.   SpeculativeExecution will retry on the next host on this exception.  Shouldn't we handle this case as well?

This isn't completely related to JAVA-819, but thought it would be worth handling since there is a possibility that the request was written and being handled when the connection was severed

adutra:
Closing as this ticket needs to be addressed first in 2.1, and also because we are investigating using/extending `RetryPolicy` to achieve this goal in a more elegant way.

adutra:
Opened #542.

PR #524, opened 2015-11-13
JAVA-988: Metadata.handleId should handle escaped double quotes.

olim7t:


tolbertam:
looks good to me :+1:

PR #525, opened 2015-11-14
JAVA-989: Include keyspace name when invalid replication found.

tolbertam:
For [JAVA-989](https://datastax-oss.atlassian.net/browse/JAVA-989).  Now logs the following when invalid replication is found on a keyspace:

> Error while computing token map for keyspace X with datacenter Y: could not achieve replication factor 3 (found 2 replicas only), check your keyspace replication settings.

The one thing i'm not so sure of here is passing the keyspace name into `computeTokenToReplicaMap`.  It is only used for logging, so seems somewhat extraneous, but since this is an internal API maybe this is not too much of a problem.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/ReplicationStategy.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ReplicationStategy.java
@@ -60,7 +60,7 @@ abstract class ReplicationStrategy {
         }
     }
 
-    abstract Map<Token, Set<Host>> computeTokenToReplicaMap(Map<Token, Host> tokenToPrimary, List<Token> ring);
+    abstract Map<Token, Set<Host>> computeTokenToReplicaMap(String keyspaceName, Map<Token, Host> tokenToPrimary, List<Token> ring);
 
     private static Token getTokenWrapping(int i, List<Token> ring) {
         return ring.get(i % ring.size());
@@ -74,7 +74,7 @@ abstract class ReplicationStrategy {
             this.replicationFactor = replicationFactor;
         }
 
-        Map<Token, Set<Host>> computeTokenToReplicaMap(Map<Token, Host> tokenToPrimary, List<Token> ring) {
+        Map<Token, Set<Host>> computeTokenToReplicaMap(String keyspaceName, Map<Token, Host> tokenToPrimary, List<Token> ring) {
 
             int rf = Math.min(replicationFactor, ring.size());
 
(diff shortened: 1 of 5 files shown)


adutra:
Rebased on current 2.1 and added patch from @snazy for JAVA-664.
Note that the patch did not apply cleanly, so I had to apply manually.
Four tests are failing after this patch:
- SimpleStrategyTest.simpleStrategyExampleTopologyMapTest
- SimpleStrategyTest.simpleStrategyExampleTopologyTooManyReplicasTest
- NetworkTopologyStrategyTest.networkTopologyStrategyExampleTopologyTest
- NetworkTopologyStrategyTest.networkTopologyStrategyExampleTopologyTooManyReplicasTest

driver-core/src/main/java/com/datastax/driver/core/ReplicationStategy.java
```diff
@@ -74,17 +74,25 @@ private SimpleStrategy(int replicationFactor) {
...
+                Set<Host> replicas = hostSetMap.get(primaryHost);
+                if (replicas == null) {
+                    // Consecutive sections of the ring can assigned to the same host
+                    replicas = new LinkedHashSet<Host>();
+                    replicas.add(primaryHost);
+                    for (int j = 1; j < ring.size() && replicas.size() < rf; j++)
+                        replicas.add(tokenToPrimary.get(getTokenWrapping(i + j, ring)));
+                    hostSetMap.put(primaryHost, replicas = ImmutableSet.copyOf(replicas));
+                }
+                replicaMap.put(ring.get(i), replicas);
```
olim7t: We can't memoize `replicas` based on `primaryHost`. What determines the contents of `replicas` is the section of the ring after token `i`; but multiple tokens can be assigned to the same host, so for different start positions you might get the same `primaryHost`, but what you encounter next on the ring will not be the same. That's why the tests fail.

driver-core/src/main/java/com/datastax/driver/core/ReplicationStategy.java
```diff
@@ -117,10 +125,11 @@ private NetworkTopologyStrategy(Map<String, Integer> replicationFactors) {
             this.replicationFactors = replicationFactors;
         }
 
-        Map<Token, Set<Host>> computeTokenToReplicaMap(Map<Token, Host> tokenToPrimary, List<Token> ring) {
+        Map<Token, Set<Host>> computeTokenToReplicaMap(String keyspaceName, Map<Token, Host> tokenToPrimary, List<Token> ring) {
 
              // This is essentially a copy of org.apache.cassandra.locator.NetworkTopologyStrategy
             Map<String, Set<String>> racks = getRacksInDcs(tokenToPrimary.values());
+            Map<Host, Set<Host>> hostSetMap = new HashMap<Host, Set<Host>>();
```
olim7t: Same for NetworkTopologyStrategy obviously.

olim7t:
:+1:

tolbertam:
big :+1: on the JAVA-664 change

PR #526, opened 2015-11-16
JAVA-983: [2.1] QueryBuilder cannot handle collections containing function calls.

adutra:
This is the same PR as #517 but targeting branch 2.1.

This commit also incidentally formats INSERT statements with a whitestpace between table name and column list, so that they do not look like a function call.


tolbertam:
looks good to me :+1:

PR #527, opened 2015-11-16
Update tests to only use netty openssl on JDK 1.7+ and add note to docs.

tolbertam:
For [JAVA-841](https://datastax-oss.atlassian.net/browse/JAVA-841), I noticed that netty-tcnative requires JDK 1.7.  Updated the tests to only run with Netty OpenSSL if the JDK version is not 1.6.  Also added a note in the docs that using netty-tcnative requires 1.7 or above.

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/SSLAuthenticatedEncryptionTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/SSLAuthenticatedEncryptionTest.java
@@ -15,20 +15,11 @@
  */
 package com.datastax.driver.core;
 
-import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
 import com.datastax.driver.core.exceptions.NoHostAvailableException;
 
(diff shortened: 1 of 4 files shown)


driver-core/src/test/java/com/datastax/driver/core/SSLAuthenticatedEncryptionTest.java
```diff
@@ -59,7 +50,7 @@ public void should_connect_with_ssl_with_client_auth_and_node_requires_auth(SslI
      * @test_category connection:ssl, authentication
      * @expected_result Connection is not established.
      */
-    @Test(groups="short", dataProvider = "sslImplementation", expectedExceptions={NoHostAvailableException.class})
+    @Test(groups="short", dataProvider = "sslImplementation", dataProviderClass = SSLTestBase.class, expectedExceptions={NoHostAvailableException.class})
```
olim7t, line 53: Nice, I hadn't noticed that you can put provider methods in another class.

olim7t:
:+1:

PR #533, opened 2015-11-19
JAVA-993: Support for "custom" types after CASSANDRA-10365.

adutra:


driver-core/src/main/java/com/datastax/driver/core/DataTypeCqlNameParser.java
```diff
@@ -79,6 +79,9 @@
      */
     static DataType parse(String toParse, Cluster cluster, String currentKeyspaceName, Map<String, UserType> currentUserTypes, Map<String, UserType> oldUserTypes, boolean frozen, boolean shallowUserTypes) {
 
+        if(toParse.startsWith("'"))
```
adutra, line 82: "Custom" types (i.e. types that do not have a CQL representation) appear in schema metadata in the "old" format, i.e., with cassandra internal class names, but surrounded by single quotes. If we don't do the check here, the parsed identifier will be empty.

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -116,7 +116,7 @@ static String handleId(String id) {
             return id.toLowerCase();
 
         // Check if it's enclosed in quotes. If it is, remove them and unescape internal double quotes
-        if (id.charAt(0) == '"' && id.charAt(id.length() - 1) == '"')
+        if (!id.isEmpty() && id.charAt(0) == '"' && id.charAt(id.length() - 1) == '"')
```
adutra, line 119: Noticed a `StringIndexOutOfBoundsException` being raised here if the identifier is the empty string. With the fix for JAVA-993, empty identifiers shouldn't arrive anymore, but I think it doesn't hurt to check the string length here.

driver-core/src/test/java/com/datastax/driver/core/ColumnMetadataAssert.java
```diff
@@ -36,22 +36,22 @@ public ColumnMetadataAssert hasName(String name) {
     }
 
     public ColumnMetadataAssert isPrimaryKey() {
-        assertThat(actual.getParent().getPrimaryKey().contains(actual));
+        assertThat(actual.getParent().getPrimaryKey().contains(actual)).as("Expecting %s to be part of the primary key, but it was not", actual).isTrue();
```
adutra, line 39: Noticed that these assertions weren't actually asserting anything.

driver-core/src/test/java/com/datastax/driver/core/CustomTypeTest.java
```diff
@@ -22,26 +22,85 @@
...
-            + "    k int,"
-            + "    c 'DynamicCompositeType(s => UTF8Type, i => Int32Type)',"
-            + "    v int,"
-            + "    PRIMARY KEY (k, c)"
-            + ") WITH COMPACT STORAGE"
+            "CREATE TABLE test ("
+                + "    k int,"
+                + "    c1 'DynamicCompositeType(s => UTF8Type, i => Int32Type)',"
+                + "    c2 'ReversedType(CompositeType(UTF8Type, Int32Type))'," // reversed translates to CLUSTERING ORDER BY DESC
+                + "    c3 'ListType(Int32Type)'," // translates to list<int>
```
tolbertam: It looks like cassandra 1.2.19 can't handle this:

```
ERROR [Native-Transport-Requests:78] 2015-11-19 07:26:14,539 ErrorMessage.java (line 210) Unexpected exception during request
java.lang.IllegalStateException: Composite column is already fully constructed
        at org.apache.cassandra.db.marshal.CompositeType$Builder.add(CompositeType.java:300)
        at org.apache.cassandra.db.marshal.CompositeType$Builder.add(CompositeType.java:332)
        at org.apache.cassandra.db.marshal.CompositeType$Builder.add(CompositeType.java:269)
        at org.apache.cassandra.cql3.Lists$Appender.doAppend(Lists.java:356)
        at org.apache.cassandra.cql3.Lists$Setter.execute(Lists.java:271)
        at org.apache.cassandra.cql3.statements.UpdateStatement.mutationForKey(UpdateStatement.java:265)
        at org.apache.cassandra.cql3.statements.UpdateStatement.getMutations(UpdateStatement.java:140)
        at org.apache.cassandra.cql3.statements.ModificationStatement.execute(ModificationStatement.java:118)
        at org.apache.cassandra.cql3.QueryProcessor.processStatement(QueryProcessor.java:103)
        at org.apache.cassandra.cql3.QueryProcessor.process(QueryProcessor.java:113)
        at org.apache.cassandra.transport.messages.QueryMessage.execute(QueryMessage.java:87)
        at org.apache.cassandra.transport.Message$Dispatcher.messageReceived(Message.java:287)
        at org.jboss.netty.channel.SimpleChannelUpstreamHandler.handleUpstream(SimpleChannelUpstreamHandler.java:70)
        at org.jboss.netty.channel.DefaultChannelPipeline.sendUpstream(DefaultChannelPipeline.java:564)
        at org.jboss.netty.channel.DefaultChannelPipeline$DefaultChannelHandlerContext.sendUpstream(DefaultChannelPipeline.java:791)
        at org.jboss.netty.handler.execution.ChannelUpstreamEventRunnable.doRun(ChannelUpstreamEventRunnable.java:43)
        at org.jboss.netty.handler.execution.ChannelEventRunnable.run(ChannelEventRunnable.java:67)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
```

This happens on first insert of row c3.  It looks like it _should_ work.  I think it's the definition of c1 and/or c2 that is really the problem.  c1 and c2 also don't appear in system.schema_columns:

```
cqlsh> select keyspace_name, columnfamily_name, column_name, validator from system.schema_columns where keyspace_name='ks_1';

 keyspace_name | columnfamily_name | column_name | validator
---------------+-------------------+-------------+-------------------------------------------------------------------------------------
          ks_1 |              test |          c3 | org.apache.cassandra.db.marshal.ListType(org.apache.cassandra.db.marshal.Int32Type
```

tolbertam: ah, I forgot that the primary key components are part of system.schema_columnfamilies, not columns, so that wasn't a problem.

adutra: Weird error. Anyway it goes away if we change the column type to `Int32Type`, so I did that.

tolbertam: That works for me :+1:.  It turns out the problem wasn't ListType being used with c1 and c2.  It's just list type in general with 1.2.  When I tested it alone I still got the same failure, it seems C\* 1.2 can't figure out what to do with the custom type on the table declaration, but it works fine as 'list<int>' (as we'd expect).

I made a [change](https://gist.github.com/tolbertam/26c0c4f28f782ef61d1a) locally to use a different type for 1.2, but I think your solution is more straightforward and elegant.

tolbertam:
Looks good to me :+1:.   I also added a test for UDTs with fields having custom types.

olim7t:
:+1:

PR #534, opened 2015-11-19
JAVA-937: TypeCodec static initializers not always correctly executed.

adutra:


adutra:
Rebased on top of current 2.1.

PR #535, opened 2015-11-20
JAVA-996: Make CodecRegistry available to ProtocolDecoder.

adutra:


driver-core/src/main/java/com/datastax/driver/core/Message.java
```diff
@@ -240,7 +243,10 @@ protected void decode(ChannelHandlerContext ctx, Frame frame, List<Object> out)
             List<String> warnings = hasWarnings ? CBUtil.readStringList(frame.body) : Collections.<String>emptyList();
 
             try {
-                Response response = Response.Type.fromOpcode(frame.header.opcode).decoder.decode(frame.body, frame.header.version);
+                CodecRegistry codecRegistry = ctx.channel().attr(CODEC_REGISTRY_ATTRIBUTE_KEY).get();
+                if (codecRegistry == null)
```
tolbertam: Can codecRegistry realistically ever be null?  It may be good to just make this an assert instead of falling back on DEFAULT_INSTANCE.

tolbertam:
big :+1: from me on this!

driver-core/src/main/java/com/datastax/driver/core/Message.java
adutra: You're right, I was being lenient here but there is no way it can be null - I will put an assert instead.

olim7t:
:+1: this makes the code much cleaner.

PR #536, opened 2015-11-20
JAVA-998: SchemaChangeListener not invoked for Functions or Aggregates having UDT arguments.

adutra:


driver-core/src/main/java/com/datastax/driver/core/AggregateMetadata.java
```diff
@@ -225,15 +227,17 @@ public KeyspaceMetadata getKeyspace() {
...
+     * Returns the CQL signature of this aggregate.
      * <p>
      * This is the name of the aggregate, followed by the names of the argument types between parentheses,
      * like it was specified in the {@code CREATE AGGREGATE...} statement, for example {@code sum(int)}.
      *
      * @return the full name.
      */
-    public String getFullName() {
-        return fullName;
+    public String getSignature() {
```
adutra, line 235: I'm removing the old "getFullName()" method as discussed previously, but providing an alternate "getSignature()" method. The goal is to provide here a valid CQL string representing the signature of the function/aggregate. This can be used by client code as an unique identifier for the function (although we do not use it for this purpose, see later). Note that for it to be valid, function names and parameters names/types are all quoted if necessary. And lastly, note that signatures must not contain the "frozen" keyword as the outermost element (i.e. `frozen<foo>` is invalid, but `list<frozen<set<text>>>` is valid).

driver-core/src/main/java/com/datastax/driver/core/FunctionMetadata.java
```diff
@@ -211,6 +204,13 @@ private String asCQLQuery(boolean formatted) {
         return sb.toString();
     }
 
+    // signatures types must not contain the keyword frozen
+    static String unfreeze(String typeStr) {
```
adutra: This is a bit clumsy but works; the goal is to remove the frozen keyword when it is the outermost piece of the whole CQL type literal.

driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
```diff
@@ -353,11 +353,13 @@ void add(MaterializedViewMetadata view) {
     }
 
     void add(FunctionMetadata function) {
-        functions.put(function.getFullName(), function);
+        String functionName = Metadata.fullFunctionName(function.getSimpleName(), function.getArguments().values());
```
adutra, line 356: Our internal unique identifiers are actually constructed by `Metadata.fullFunctionName`.

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -134,15 +134,39 @@ static String escapeId(String ident) {
...
+     * to the function/aggregate signature.
+     * This is only used to generate keys for internal metadata maps (KeyspaceMetadata.functions and.
+     * KeyspaceMetadata.aggregates).
+     * Note that if simpleName comes from the user, the caller must call handleId on it before passing it to this method.
+     * Note that this method does not necessarily generates a valid CQL function signature.
+     * Note that argumentTypes can be either a list of strings (schema change events)
+     * or a list of DataTypes (function lookup from client code).
+     * This method must ensure that both cases produce the same identifier.
+     */
     static String fullFunctionName(String simpleName, Collection<?> argumentTypes) {
```
adutra, line 148: The reason why we have this method is that we need an unique identifier that is easily created either with the list of parameter types, or a list of Strings representing the parameter types, as provided by a schema change event. So the algorithm here differs a bit from the one used to create function signatures:
- The function name is not quoted
- user types are not qualified with the keyspace name
- the frozen keyword is allowed

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -160,7 +184,7 @@ static String fullFunctionName(String simpleName, Collection<?> argumentTypes) {
      * or even {@link Cluster#connect(String)}.
      */
     public static String quote(String id) {
-        return '"' + id + '"';
+        return '"' + id.replace("\"", "\"\"") + '"';
```
adutra, line 193: This is unrelated, but I think it's a bug we never came across so far. An identifier such as `foo " bar` would not be correctly transformed into `foo "" bar`. I modified a few tests in UserTypesTest to show evidence of it.

driver-core/src/test/java/com/datastax/driver/core/DataTypeCqlNameParserTest.java
```diff
@@ -88,8 +89,8 @@ public void should_parse_nested_collection_types() {
         assertThat(parse("list<list<int>>", cluster, null, null, null, false, false)).isEqualTo(list(list(cint())));
         assertThat(parse("set<list<frozen<map<bigint,varchar>>>>", cluster, null, null, null, false, false)).isEqualTo(set(list(map(bigint(), varchar(), true))));
 
-        UserType keyType = keyspaceMetadata.getUserType("Incr,edibly\" EvilTy<>><<><p\"e");
-        UserType valueType = keyspaceMetadata.getUserType("\"A\"");
+        UserType keyType = keyspaceMetadata.getUserType(quote("Incr,edibly\" EvilTy<>><<><p\"e"));
```
adutra, line 92: This shows that `quote` method is now behaving well. It's odd that it also works without calling `quote` and simply passing the identifier, but there is a comment about it in `handleId` so I guess it's a known corner case.

driver-core/src/main/java/com/datastax/driver/core/AggregateMetadata.java
```diff
@@ -122,7 +123,7 @@ static AggregateMetadata build(KeyspaceMetadata ksm, Row row, VersionNumber vers
     private static String makeStateFuncFullName(String stateFuncSimpleName, DataType stateType, List<DataType> argumentTypes) {
         List<DataType> args = Lists.newArrayList(stateType);
         args.addAll(argumentTypes);
-        return Metadata.fullFunctionName(stateFuncSimpleName, args);
+        return fullFunctionName(stateFuncSimpleName, args);
```
olim7t: In cases like that I find static imports make the code harder to read, especially outside of an IDE.

adutra:
Squashed commits; all remarks have been addressed as discussed.

PR #538, opened 2015-11-23
JAVA-819: [2.1] Driver shouldn't retry on client timeout if statement is not idempotent

adutra:


driver-core/src/main/java/com/datastax/driver/core/policies/ClientFailureAwareRetryPolicy.java
```diff
@@ -0,0 +1,60 @@
...
+import com.datastax.driver.core.Statement;
+import com.datastax.driver.core.exceptions.DriverException;
+
+/**
+ * A policy that defines a default behavior to adopt when a request returns
+ * a TimeoutException or an UnavailableException.
+ *
+ * Such policy allows to centralize the handling of query retries, allowing to
+ * minimize the need for exception catching/handling in business code.
+ */
```
olim7t: We should also mention that this interface exists for binary compatibility, and will be merged in the parent interface in 3.0.

adutra: Good point. I was thinking about renaming it to `ExtendedRetryPolicy` - I don't like much the current name.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -502,33 +526,45 @@ public void onSet(Connection connection, Message.Response response, long latency
                                 connection.release();
                                 // Try another node
                                 logger.warn("Host {} is overloaded, trying next host.", connection.address);
-                                DriverException overloaded = new DriverException("Host overloaded");
-                                logError(connection.address, overloaded);
                                 if (metricsEnabled())
                                     metrics().getErrorMetrics().getOthers().inc();
-                                retry(false, null);
-                                return;
+                                if (retryPolicy instanceof ClientFailureAwareRetryPolicy)
```
newkek: Seems like this code is duplicated in cases OVERLOADED, SERVER_ERROR and IS_BOOTSTRAPPING, couldn't we put it in a separate function or add it in `processRetryDecision()` to avoid duplication? Same for the metrics calls

newkek: And the `connection.release()`

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -634,21 +651,36 @@ public void onSet(Connection connection, Message.Response response, long latency
                     connection.release();
 
                     // TODO should we check the response ?
```
newkek: I guess this TODO can be removed?

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -502,33 +526,45 @@ public void onSet(Connection connection, Message.Response response, long latency
                                 connection.release();
                                 // Try another node
                                 logger.warn("Host {} is overloaded, trying next host.", connection.address);
```
newkek: Seems like logging "trying next host" is not really consistent anymore since the decision is taken later by the RetryPolicy.

newkek: In fact it is still valid but only for the else case line 536 for example

adutra: Good point, I will rephrase the sentences.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
adutra: Hmm the code is similar but rarely identical. I will investigate if some refactoring is possible.

adutra:
Pushed a substantial commit that renames `ClientFailureAwareRetryPolicy` to `ExtendedRetryPolicy`.
Also added a new method `onConnectionError`, because it's easy to isolate these errors from the others, which makes `onUnexpectedError` less of a catchall method.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -664,16 +712,7 @@ public void onException(Connection connection, Exception exception, long latency
 
                 @Override
                 public boolean onTimeout(Connection connection, long latency, int retryCount) {
-                    QueryState queryState = queryStateRef.get();
```
adutra, line 667: This piece seemed very similar to the root level `onTimeout()` - removing it did not affect any of our tests.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -692,13 +731,22 @@ public void onException(Connection connection, Exception exception, long latency
...
+                        metrics().getErrorMetrics().getConnectionErrors().inc();
+                        if (decision.getType() == Type.RETRY)
+                            metrics().getErrorMetrics().getRetriesOnConnectionError().inc();
+                        if (decision.getType() == Type.IGNORE)
+                            metrics().getErrorMetrics().getIgnoresOnConnectionError().inc();
+                    }
+                    processRetryDecision(decision, connection, exception);
                     return;
                 }
                 setFinalException(connection, exception);
```
adutra, line 742: I wonder what to do here. Should we leave it as is? Or pass the exception to the retry policy, even if it's not a `ConnectionException`?

driver-core/src/main/java/com/datastax/driver/core/exceptions/ConnectionException.java
```diff
@@ -13,29 +13,43 @@
...
  *   limitations under the License.
  */
-package com.datastax.driver.core;
+package com.datastax.driver.core.exceptions;
 
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 
-class ConnectionException extends Exception {
+public class ConnectionException extends DriverException implements CoordinatorException {
```
adutra: Had to make this exception and its subclasses public because they can now be exposed to user code.
An alternate solution would be to never expose these exceptions, but create equivalent `DriverException`s and expose these to the retry policy; a little like it was done for errors like SERVER_ERROR, OVERLOADED, etc.

driver-core/src/main/java/com/datastax/driver/core/policies/FallthroughRetryPolicy.java
```diff
@@ -86,4 +89,67 @@ public RetryDecision onWriteTimeout(Statement statement, ConsistencyLevel cl, Wr
...
+     * be achieved.
+     * @param cl the original consistency level for the operation.
+     * @param nbRetry the number of retry already performed for this operation.
+     * @return the retry decision. If {@code RetryDecision.RETHROW} is returned,
+     * an {@link OperationTimedOutException} will
+     * be thrown for the operation.
+     */
+    @Override
+    public RetryDecision onClientTimeout(Statement statement, ConsistencyLevel cl, int nbRetry) {
+        return RetryDecision.tryNextHost(cl);
```
adutra: I wonder if this policy should return RETHROW here. It would be more consistent with the class purpose, but would break the current behavior.

driver-core/src/test/java/com/datastax/driver/core/QueryLoggerErrorsTest.java
```diff
@@ -174,14 +174,9 @@ public void should_log_exception_from_the_given_result(PrimingRequest.Result res
         } catch(Exception e) {
             if(e instanceof NoHostAvailableException) {
                 assertThat(expectedException).isEqualTo(NoHostAvailableException.class);
-                // ok
                 Throwable error = ((NoHostAvailableException)e).getErrors().get(hostAddress);
                 assertThat(error).isNotNull();
-                if(loggedException.getSuperclass() == DriverInternalError.class) {
```
adutra, line 180: As discussed, the exceptions reported now are the very same that we decode from the response. I think that the code initially tried to avoid exposing those because they inherit from `DriverInternalError`, which, in its author's initial conception, is a synonym of "bug in the driver". But since the Query Logger and JAVA-720, exceptions like `ServerError` or `OverloadedError` are first-class citizens and do not mean that the driver is buggy; so I guess we can also pass them to the retry policy. The major mistake is to have created such exceptions as subclasses of `DriverInternalError` - but this cannot be changed, at least not in 2.1.
I don't have any strong opinion here; if someone disagrees about exposing these classes, I can revert this.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -150,13 +150,6 @@ private void cancelPendingExecutions(SpeculativeExecution ignore) {
             execution.cancel();
     }
 
-    private void logError(InetSocketAddress address, Throwable exception) {
```
adutra, line 153: Moved inside SpeculativeExecution to have the spec exec id logged as well. Fixed the check-then-act idiom.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
newkek: The code 

``` java
connection.release();
if (retryPolicy instanceof ExtendedRetryPolicy)
    retry = ((ExtendedRetryPolicy)retryPolicy).onUnexpectedError(statement,
        request().consistency(),
        (DriverException)exceptionToReport,
        retriesByPolicy);
else
    retry = RetryPolicy.RetryDecision.tryNextHost(null);
if (metricsEnabled()) {
    metrics().getErrorMetrics().getOthers().inc();
    if (retry.getType() == Type.RETRY)
        metrics().getErrorMetrics().getRetriesOnUnexpectedError().inc();
    if (retry.getType() == Type.IGNORE)
        metrics().getErrorMetrics().getIgnoresOnUnexpectedError().inc();
}
```

Is the same in each one of the new Exceptions exposed in the retry policy (overloaded/bootstrapped/server_error). I though it would be clearer to group this in a function because now the behaviour is the same everywhere, except for `SERVER_ERROR` where we also defunct the connection.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -726,8 +774,20 @@ public boolean onTimeout(Connection connection, long latency, int retryCount) {
             try {
                 connection.release();
 
-                logError(connection.address, timeoutException);
-                retry(false, null);
+                RetryPolicy retryPolicy = retryPolicy();
+                RetryPolicy.RetryDecision decision;
+                if (retryPolicy instanceof ExtendedRetryPolicy)
```
newkek, line 775: Same as for the other comments about duplication, would be nice to have a function like `raiseClientTimenoutException()` to do it I think, but if you don't see the utility of it that's ok. It is just to make sure if we change the behaviour at some point about client timeout in the future, doing it in the function would make sure it is done for all places where we call `RetryPolicy.onClientTimeout()`

driver-core/src/test/java/com/datastax/driver/core/QueryLoggerErrorsTest.java
```diff
@@ -131,7 +131,7 @@ public void should_log_timed_out_queries() throws Exception {
             .contains(ip)
             .contains(Integer.toString(scassandra.getBinaryPort()))
             .contains(query)
-            .contains(OperationTimedOutException.class.getName());
+            .contains("Timed out waiting for server response");
```
adutra, line 134: OperationTimedOutException now replaced with a generic DriverException (pre-JAVA-698 behavior).

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -281,6 +280,8 @@ void sendRequest() {
         }
 
         private boolean query(final Host host) {
+            // set this flag to false at the beginning of each attempt to send the query
+            requestSent = false;
```
adutra: Flag used only for ConnectionExceptions, see onException() later on.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -572,26 +615,7 @@ public void onSet(Connection connection, Message.Response response, long latency
                         if (retry == null)
```
adutra, line 616: I wonder if we should do something more in the default case of the switch block above: currently we only increment a counter and release the connection.

driver-core/src/main/java/com/datastax/driver/core/policies/FallthroughRetryPolicy.java
```diff
@@ -86,4 +86,17 @@ public RetryDecision onWriteTimeout(Statement statement, ConsistencyLevel cl, Wr
...
+    /**
+     * {@inheritDoc}
+     * <p>
+     * This implementation triggers a retry on the next host in the query plan,
+     * regardless of the consistency level, the number of retries or
+     * the possibility that the mutation has been applied server-side.
+     */
+    @Override
+    public RetryDecision onUnexpectedError(Statement statement, ConsistencyLevel cl, int nbRetry, boolean mightHaveBeenApplied) {
+        return RetryDecision.tryNextHost(cl);
```
adutra, line 99: tryNextHost(cl) is consistent with current behavior, but rethrow() might be more appropriate here?

driver-core/src/main/java/com/datastax/driver/core/policies/ExtendedRetryPolicy.java
```diff
@@ -0,0 +1,66 @@
...
+     *
+     *
+     * @param statement the original query that failed.
+     * @param cl the original consistency level for the operation.
+     * @param nbRetry the number of retries already performed for this operation.
+     * @param mightHaveBeenApplied whether the mutation is likely to have been applied server-side.
+     * @return the retry decision. If {@code RetryDecision.RETHROW} is returned,
+     * a {@link com.datastax.driver.core.exceptions.DriverException DriverException} will be thrown for the operation.
+     */
+    RetryDecision onUnexpectedError(Statement statement, ConsistencyLevel cl, int nbRetry, boolean mightHaveBeenApplied);
```
adutra: I feel this could be too vague for some users. E.g. in the event of a client timeout, certain users might have enough knowledge of their network infrastructure to be able to determine if the request has reached the coordinator or not, and thus take the appropriate measure. But here they don't even know if the failure was due to a client timeout, or something else.

newkek: Why not instead of the boolean, put like an Enum, where values could be each kind of exception we handle internally (CLIENT_TIMEOUT/OVERLOADED/etc..), that may give more information without having the previous problems we had (exposing ConnectionException, etc...) ?

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
olim7t, line 616: That case is for an unexpected error code; it will cause the error to be reported directly to the user (`DefaultResultSetFuture.onSet`), which sounds reasonable.

driver-core/src/main/java/com/datastax/driver/core/Metrics.java
```diff
@@ -273,6 +273,7 @@ void shutdown() {
         private final Counter writeTimeouts = registry.counter("write-timeouts");
         private final Counter readTimeouts = registry.counter("read-timeouts");
         private final Counter unavailables = registry.counter("unavailables");
+        private final Counter clientTimeouts = registry.counter("client-timeouts");
```
tolbertam, line 276: This metric will be super useful I think.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -633,22 +657,47 @@ public void onSet(Connection connection, Message.Response response, long latency
...
+                                else
+                                    decision = RetryPolicy.RetryDecision.tryNextHost(null);
+                                if (metricsEnabled()) {
+                                    metrics().getErrorMetrics().getOthers().inc();
+                                    if (decision.getType() == Type.RETRY)
+                                        metrics().getErrorMetrics().getRetriesOnUnexpectedError().inc();
+                                    if (decision.getType() == Type.IGNORE)
+                                        metrics().getErrorMetrics().getIgnoresOnUnexpectedError().inc();
+                                }
+                                processRetryDecision(decision, connection, driverException);
```
olim7t: I wonder if it's worth using the policy here (and in case of error below). Preparing a query is idempotent by nature, so there should be no harm in retrying each time.

driver-core/src/main/java/com/datastax/driver/core/Metrics.java
```diff
@@ -273,6 +273,7 @@ void shutdown() {
         private final Counter writeTimeouts = registry.counter("write-timeouts");
```
tolbertam, line 273: Weird/slightly offtopic topic question, but what is the policy on adding methods to classes that aren't meant to be extended/used otherwise?  Can we add methods to those (i.e. Metrics) without breaking API, i.e. will clirr care (guessing it doesn't because we can build)?  Should we make Metrics final in 3.0 just to avoid this possibility (can't remotely imagine this ever being a problem anyways).

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -664,16 +713,7 @@ public void onException(Connection connection, Exception exception, long latency
...
-                        !queryStateRef.compareAndSet(queryState, queryState.complete())) {
-                        logger.debug("onTimeout triggered but the response was completed by another thread, cancelling (retryCount = {}, queryState = {}, queryStateRef = {})",
-                            retryCount, queryState, queryStateRef.get());
-                        return false;
-                    }
-                    connection.release();
-                    logError(connection.address, new DriverException("Timeout waiting for response to prepare message"));
-                    retry(false, null);
+                    SpeculativeExecution.this.onTimeout(connection, latency, retryCount);
                     return true;
```
olim7t, line 706: My comment about PREPARE being idempotent also applies here, so we might want to keep a separate implementation that does not use the retry policy.

If on the contrary we keep calling the method of the enclosing class, we want to return its result instead of always returning `true`.

driver-core/src/main/java/com/datastax/driver/core/Metrics.java
tolbertam, line 273: ah, maybe this is not a problem since Errors is a non-static nested class, so couldn't be extended anyways.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
olim7t, line 742: Same remark as in onSet: if it's a completely unexpected exception, it sounds reasonable to fail fast and report it back to the user through the future.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -58,8 +59,7 @@
     private final io.netty.util.Timer scheduler;
 
     private volatile List<Host> triedHosts;
-
-    private volatile Map<InetSocketAddress, Throwable> errors;
```
tolbertam, line 62: Any significant introduced overhead of creating a CHM for each RequestHandler?  Previously it would only be constructed on error.  I think in any case it would be nice to reduce the initial capacity/load factor/concurrency.

driver-core/src/main/java/com/datastax/driver/core/policies/ExtendedRetryPolicy.java
olim7t: From the client's point of view, I'm not sure if the type of error makes any difference. If an update fails, the client decides not to retry, and it's not sure whether the query was applied or not, the fallback scenario will probably involve a read with the adequate consistency level to find out what the server-side situation is. Whether there was a client timeout or a dropped connection does not influence this scenario IMO.

driver-core/src/main/java/com/datastax/driver/core/policies/FallthroughRetryPolicy.java
olim7t, line 99: At least for 2.1 we should not change the behavior.

driver-core/src/main/java/com/datastax/driver/core/Metrics.java
olim7t, line 273: If you add methods to a class that someone added to their own sub-implementation, this could lead to unexpected behavior because the methods of the subclass now override the ones you added; but strictly speaking it's still binary compatible.
In that case I think it's reasonable to assume that nobody extended the class to add those methods.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
adutra: I just changed that (reverted to initial behavior) thinking exactly the same.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -694,11 +722,20 @@ public void onException(Connection connection, Exception exception, long latency
                 connection.release();
 
                 if (exception instanceof ConnectionException) {
-                    if (metricsEnabled())
+                    RetryPolicy retryPolicy = retryPolicy();
+                    RetryPolicy.RetryDecision decision;
+                    if (retryPolicy instanceof ExtendedRetryPolicy)
+                        decision = ((ExtendedRetryPolicy)retryPolicy).onUnexpectedError(statement, request().consistency(), retriesByPolicy, requestSent);
+                    else
+                        decision = RetryPolicy.RetryDecision.tryNextHost(null);
```
tolbertam: The documentation for RetryDecision#getRetryConsistencyLevel indicates that it returns null for RETRY or RETHROW cases.  We should update that doc as it now seems stale as in the case retry CL is null, it simply reuses the statement's CL.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -341,6 +342,8 @@ private void write(Connection connection, Connection.ResponseCallback responseCa
                     break;
             }
 
+            // set this flag to true before writing the request
+            requestSent = true;
```
olim7t: This is a bit too coarse-grained IMO. For exceptions thrown directly from `Connection#write(ResponseCallback, boolean)` and also if we detect an error in `Connection#writeHandler`, we can be sure that the request has not been sent. Thinking of a way to handle that.

driver-core/src/main/java/com/datastax/driver/core/Metrics.java
adutra, line 273: Also, adding a new method to a class does not break binary compatibility as long as the client doesn't need to implement it (e.g. if it's abstract). See [here](https://wiki.eclipse.org/Evolving_Java-based_APIs_2).

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -722,12 +759,30 @@ public boolean onTimeout(Connection connection, long latency, int retryCount) {
             }
 
             Host queriedHost = current;
-            OperationTimedOutException timeoutException = new OperationTimedOutException(connection.address);
+
+            // This cannot be OperationTimedOutException because
```
tolbertam, line 764: Not sure I understand from this comment the reason behind the change, can you provide some more info around this change?  Just for my own benefit :)

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
olim7t: Actually, exceptions thrown directly from `Connection#write` are caught in `query` which ignores the host and tries the next one directly, the retry policy is not involved there. So the only case where `requestSent` could be `false` would be from inside `writeHandler`.

driver-core/src/main/java/com/datastax/driver/core/policies/DefaultRetryPolicy.java
```diff
@@ -135,4 +135,17 @@ public RetryDecision onUnavailable(Statement statement, ConsistencyLevel cl, int
...
+
+    /**
+     * {@inheritDoc}
+     * <p>
+     * This implementation triggers a retry on the next host in the query plan,
+     * regardless of the consistency level, the number of retries or
+     * the possibility that the mutation has been applied server-side.
+     */
+    @Override
+    public RetryDecision onUnexpectedError(Statement statement, ConsistencyLevel cl, int nbRetry, boolean mightHaveBeenApplied) {
```
tolbertam: Sorry if I just missed the conversation, but why not also provide the Exception?  Or is that considered too specific/too much information to make a decision here?

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
adutra, line 62: Good point, but the ideal capacity / concurrency is related to the query plan size and the speculative execution policy... difficult to guess sensible values here. Maybe use an atomic reference? or a volatile + double checked locking?

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
adutra, line 764: Previously (before JAVA-698) it used to be a DriverException. The newbie I was back then (maybe I'm still am) found it weird because one would create an OperationTimedOutException instead in some other similar places (Connection.Future.onTimeout and DefaultResultSetFuture.onTimeout).
But I realized since then that the difference is that RequestHandler logs those errors and adds them to the final list of errors in NHAE. So it strives to hide non-public exceptions.
Things became complex with the QueryLogger, because some "internal" exceptions like this one would appear in client code for the first time. And now with the ExtendedRetryPolicy, users can even catch this exception if the policy says RETHROW.
(Curiously, the same logic would apply to ConnectionException, but oddly this one has never been hidden and appears bare in NHAE, Query Logger, etc. – see RequestHandler.onException.)
All of his is a bit flawed, but hopefully we can address this in a better way with JAVA-1006. I think that now that these exceptions bubble up into client code, they deserve to become public.

tolbertam, line 764: Thanks for the explanation.  So in general, should OperationTimedOutException should only be raised internally for things like preparing statements, checking cluster name, etc.?  I wonder if we should make that clear in the documentation, as it is currently somewhat unclear with regards to that.

Also since we have specific exceptions that inherit DriverException, would it make sense to create a specific implemetation for this case so users can properly check for it?

adutra, line 764: >  I wonder if we should make that clear in the documentation, as it is currently somewhat unclear with regards to that.

Yes I will try to put some comments in the code.

> would it make sense to create a specific implemetation for this case so users can properly check for it?

Either that, or promote ConnectionException to public. I will try to address that in JAVA-1006.

driver-core/src/main/java/com/datastax/driver/core/policies/DefaultRetryPolicy.java
olim7t: Yes, too much information, and also the fact that some of the exceptions are not public so we would need to change that or rewrap.
On a side note, I'm not a huge fan of the method name, all errors are kind of unexpected; how about `onRequestError` instead?

tolbertam: The only issue i have with not having the exception is you don't have any idea why there was an error, so retrying may not be the right behavior depending on the criticality of the error.  Some delineation of why there was an error might be nice.   I just noticed [this discussion](https://github.com/datastax/java-driver/pull/538/files#r45959031) so will carry on there with my input.

I agree 'unexpected' might be perceived as too generic.  +1 for `onRequestError`.  Straight up `onError` might be good as well, since we can assume that there was an error with the request.

driver-core/src/main/java/com/datastax/driver/core/policies/ExtendedRetryPolicy.java
tolbertam: I agree that this is currently too vague:

I can see three different error types i may want to handle differently:
- Client timeouts - I may just accept the fact that I have occasionally slow queries and treat every incidence of timeout the same or do it on a per statement basis.
- Connection/Transport errors - exceptions thrown while attempting the write/waiting for response.   I think the general case I'd always want to retry (unless non-idempotent)
- Server errors - Errors returned from the server (Protocol, Bootstrapping, Server error).  These are typically exceptional cases.

I think in all cases but the third, there is no need to get any more specific about the exception.  I propose the following:

``` java
onClientTimeout(Statement statement, ConsistencyLevel cl, int nbRetry);
onConnectionError(Statement statement, ConsistencyLevel cl, int nbRetry, boolean mightHaveBeenApplied);
onServerError(Statement statement, ConsistencyLevel cl, int nbRetry, DriverInternalError exception)
```

Thoughts on this approach:

**Pros**:
- Clear (enough) definition of what the error is about without giving too much information (like exposing internal exceptions).
- It is easy for me as a user to understand why the method is called and what to do based on the method called.

**Cons**:
- The user has to implement three separate methods instead of one.  
- There may be cases where neither of these three methods cover an error scenario i'm not thinking of maybe?

Thoughts?

driver-core/src/main/java/com/datastax/driver/core/policies/IdempotenceAwareRetryPolicy.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * A retry policy that avoids retrying non-idempotent statements.
+ * <p>
+ * In case of write timeouts or unexpected errors, this policy will always return {@link RetryDecision#rethrow()}
+ * if the statement is deemed non-idempotent (see {@link #isIdempotent(Statement)}).
+ * <p>
+ * For all other cases, this policy delegates the decision to the child policy.
+ */
+public class IdempotenceAwareRetryPolicy implements ExtendedRetryPolicy {
```
tolbertam, line 31: Nice, i like the idea of wrapping Retry Policies :+1:.

driver-core/src/main/java/com/datastax/driver/core/policies/ExtendedRetryPolicy.java
adutra: I like your proposal; it has also the advantage of solving the naming problem by creating more meaningful names than just `onUnpexpectedError`.
However, we just decided with @olim7t to drop the boolean `mightHaveBeenApplied` - it would never be `false` in fact.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
adutra: Thanks, I updated the javadocs.

adutra:
Squashed and rebased on top of 2.1.
The method name has been renamed to `onRequestError`.

driver-core/src/test/java/com/datastax/driver/core/policies/DefaultRetryPolicyIntegrationTest.java
```diff
@@ -146,4 +149,83 @@ public void should_rethrow_unavailable_in_no_host_available_exception() {
...
+            scassandras
+                .node(3).primingClient().prime(PrimingRequest.queryBuilder()
+                    .withQuery("mock query")
+                    .withThen(then().withFixedDelay(1000L).withRows(row("result", "result3")))
+                    .build());
+            try {
+                query();
+                fail("expected a NoHostAvailableException");
+            } catch (NoHostAvailableException e) {
+                assertThat(e.getErrors().keySet()).hasSize(3).containsExactly(
```
tolbertam: The ordering of keySet() is not going to be determinate since it depends on internal map implementation which may vary from JDK implementation,  containsOnly should be adequate here (I think?) so i'll change it.

tolbertam: Pushed fix and also made appropriate fix for 3.0 version as well.

driver-core/src/main/java/com/datastax/driver/core/policies/ExtendedRetryPolicy.java
tolbertam: :+1: to single method.  On the 3.0 PR [#542] onRequestError you've included DriverException which I think should be adequate.  If the user needs more information on driver 2.1, we can give them the option of upgrading :)

adutra:
Closing in favor of #587.

PR #540, opened 2015-11-24
JAVA-420: + recreate BoundStatement for obsolete prepared statement

vkuptcov:


driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -638,6 +659,14 @@ public void onSet(Connection connection, Message.Response response, long latency
                         case RESULT:
                             if (((Responses.Result)response).kind == Responses.Result.Kind.PREPARED) {
                                 logger.debug("Scheduling retry now that query is prepared");
+                                Responses.Result.Prepared prepared = (Responses.Result.Prepared) response;
```
vkuptcov: It isn't necessary to retry the statement, because we could get the result from the previous execution and just provide a new metadata in the fresh PreparedStatement.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -58,6 +59,7 @@
     private final io.netty.util.Timer scheduler;
 
     private volatile List<Host> triedHosts;
+    private volatile Statement actualStatement;
```
vkuptcov: `actualStatement` could be removed after https://github.com/datastax/java-driver/pull/540/files#r45767990

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -2116,6 +2116,18 @@ public PreparedStatement addPrepared(PreparedStatement stmt) {
             return stmt;
         }
 
+        public PreparedStatement replacePrepared(PreparedStatement stmt) {
+            PreparedStatement previous = preparedQueries.replace(stmt.getPreparedId().id, stmt);
+            if (previous != null) {
+                logger.debug("Replace already prepared query {}.", stmt.getQueryString());
+
+                // The one object in the cache will get GCed once it's not referenced by the client anymore since we use a weak reference.
+                // So we need to make sure that the instance we do return to the user is the one that is in the cache.
```
olim7t: This comment was copied from `addPrepared`, but I find it a bit confusing here, because client code doesn't use `previous` at all. In fact this method could be void.

maqdev:
+1

vkuptcov:
@olim7t  I updated the PR, could you please check it?

adutra:
@vkuptcov I am closing this PR as the definitive fix will come with #794 (JAVA-1196). Thanks for your contribution!

PR #541, opened 2015-11-26
JAVA-1003: Change default consistency level to LOCAL_ONE (amends JAVA-926).

adutra:


newkek:
I remember last time with the question on the mailing list, where LOCAL_ONE returns 2 replicas if a write is a BATCH_LOG, couldn't we add a comment about that somewhere?

newkek:
After code investigations and tests, couldn't reproduce an issue where a DC with RF>1, batch writes with CL=LOCAL_ONE return an error with 'required replicas' > 1, because it would require the Batchlog write process to fail. It looks like a corner case. We don't need to comment it.

olim7t:
:+1:

PR #542, opened 2015-12-01
JAVA-819: [3.0] Driver shouldn't retry on client timeout if statement is not idempotent.

adutra:
This is the same as #538 but `ExtendedRetryPolicy` is merged into `RetryPolicy`.


driver-core/src/main/java/com/datastax/driver/core/policies/FallthroughRetryPolicy.java
```diff
@@ -88,6 +88,19 @@ public RetryDecision onUnavailable(Statement statement, ConsistencyLevel cl, int
...
+    /**
+     * {@inheritDoc}
+     * <p>
+     * This implementation triggers a retry on the next host in the query plan,
+     * regardless of the consistency level, the number of retries or
+     * the possibility that the mutation has been applied server-side.
+     */
+    @Override
+    public RetryDecision onUnexpectedError(Statement statement, ConsistencyLevel cl, int nbRetry) {
+        return RetryDecision.rethrow();
```
adutra, line 99: I am modifying the decision here as compared to the 2.1 branch, which returns `tryNextHost(cl)` to mimic the driver's behavior prior to JAVA-819.

adutra:
Squashed and rebased on top of JAVA-1006.

tolbertam:
What do you think about making the default retry policy idempotence aware in 3.0?  it seems like good practice in general to not retry on a write timeout for an non-idempotent write.

adutra:
The problem I see is that by default statements are non-idempotent, which means that in practice, no retry would ever be made on a write timeout unless the user modifies his configuration.

tolbertam:
> The problem I see is that by default statements are non-idempotent, which means that in practice, no retry would ever be made on a write timeout unless the user modifies his configuration.

Great point, I agree with not making the retry policy idempotence aware based on that.  In the general case a query will be idempotent with the only exceptions being the ones listed in [the documentation](http://datastax.github.io/java-driver/features/speculative_execution/).

It is unfortunate in the case where a non-idempotent write that times out is retried, because it will likely happen rarely and be hard to detect, and when detected it may not be immediately obvious to the user that this is the cause.  However I still agree with the current behavior, but I will pursue if there could be means in C\* that can be added to help us out in these scenarios.

adutra:
Another thing to keep in mind is that the default retry policy already kind of assumes that statements could be non-idempotent and consequently is very restrictive; it only retries if the write failed when writing to the batch log. For all the rest it rethrows. Same for DowngradingConsistencyRetryPolicy: it's very permissive for read timeouts and unavailables but it actually avoids retrying on write timeouts.

tolbertam:
> Another thing to keep in mind is that the default retry policy already kind of assumes that statements could be non-idempotent and consequently is very restrictive; it only retries if the write failed when writing to the batch log. For all the rest it rethrows. Same for DowngradingConsistencyRetryPolicy: it's very permissive for read timeouts and unavailables but it actually avoids retrying on write timeouts.

That makes sense and does put me more at ease, thanks :)

driver-core/src/main/java/com/datastax/driver/core/policies/DefaultRetryPolicy.java
```diff
@@ -137,6 +138,18 @@ public RetryDecision onUnavailable(Statement statement, ConsistencyLevel cl, int
...
 
+    /**
+     * {@inheritDoc}
+     * <p>
+     * This implementation triggers a retry on the next host in the query plan,
+     * regardless of the consistency level, the number of retries or
+     * the possibility that the mutation has been applied server-side.
+     */
+    @Override
+    public RetryDecision onRequestError(Statement statement, ConsistencyLevel cl, DriverException e, int nbRetry) {
```
tolbertam, line 151: There is somewhat of a contradiction with this behavior and the RetryPolicy#onRequestError javadoc:

``` java
+     * Note that when this method is invoked, <em>the driver cannot guarantee that the mutation has
+     * been effectively applied server-side</em>; a retry should only be attempted if the request
+     * is known to be idempotent.
```

Here we are always retrying regardless of whether or not the query is idempotent, shouldn't the default behavior match that of the parent java doc?

tolbertam, line 151: Just saw [the conversation](https://github.com/datastax/java-driver/pull/538#discussion-diff-45864310) on the 2.1 PR (#538) that we did not change the existing behavior, but for 3.0 it seems more appropriate to rethrow here as you indicated.

driver-core/src/main/java/com/datastax/driver/core/policies/LoggingRetryPolicy.java
```diff
@@ -33,6 +35,31 @@
...
+    @VisibleForTesting static final String RETRYING_ON_WRITE_TIMEOUT =
+        "Retrying on write timeout on {} at consistency {} (initial consistency: {}, write type: {}, required acknowledgments: {}, received acknowledgments: {}, retries: {})";
+
+    @VisibleForTesting static final String IGNORING_UNAVAILABLE =
+        "Ignoring unavailable exception (initial consistency: {}, required replica: {}, alive replica: {}, retries: {})";
+
+    @VisibleForTesting static final String RETRYING_ON_UNAVAILABLE =
+        "Retrying on unavailable exception on {} at consistency {} (initial consistency: {}, required replica: {}, alive replica: {}, retries: {})";
+
+    @VisibleForTesting static final String IGNORING_UNEXPECTED_ERROR =
```
tolbertam: I think these log messages should be updated to "Ignoring request error" / "Retrying on request error" to match the method name.

driver-core/src/main/java/com/datastax/driver/core/policies/DefaultRetryPolicy.java
adutra, line 151: The implementation here simply mimics the legacy behavior, which was to retry on next host. Changing this to rethrow would be a significant behavioral change as users would start getting exceptions where previously the operation would be retried; so I'm not sure if we should change it, but I'm open to discussion. BTW in FallthroughRetryPolicy I did exactly that (i.e. it retries on next host in 2.1, but rethrows in 3.0).

tolbertam, line 151: > The implementation here simply mimics the legacy behavior, which was to retry on next host.

But there is a contradiction or at least its somewhat misleading in that the RetryPolicy#onRequestError doc states _a retry should only be attempted if the request is known to be idempotent_ and DefaultRetryPolicy always blindly retrying regardless of idempotence.  A user could think the guidelines in the interface javadoc could reflect the default behavior but in this case that is not true, just wondering if that's worth clarifying anywhere.

adutra, line 151: I amended the docs to clarify that the implementation breaks the contract.

tolbertam, line 151: Excellent, thanks :+1:

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -344,6 +349,7 @@ private void write(Connection connection, Connection.ResponseCallback responseCa
                     break;
             }
 
+            // set this flag to true before writing the request
```
tolbertam: Is this comment a relic of a previous commit?

driver-core/src/main/java/com/datastax/driver/core/policies/FallthroughRetryPolicy.java
```diff
@@ -88,6 +89,19 @@ public RetryDecision onUnavailable(Statement statement, ConsistencyLevel cl, int
         return RetryDecision.rethrow();
     }
 
+    /**
+     * {@inheritDoc}
+     * <p>
+     * This implementation triggers a retry on the next host in the query plan,
```
tolbertam: This doc is wrong since it doesn't trigger a retry, will update it.

tolbertam:
Pushed some slight javadoc changes (hope this is ok; feel free to squash them in).   I've created some more tests, but they can't be merged in quite yet since they require some functionality in Scassandra that  haven't been released yet.  These are:
- Refactor DowngradingConsistencyRetryPolicyIntegrationTest to use Scassandra.
- Add tests for retry on connection error.

Working branch for that can be found [here](https://github.com/tolbertam/java-driver/commits/java819-3.0-connectionErrors).  There should be an oncoming release of scassandra soon, I'll open up a separate PR as soon as it is released to incorporate those tests as I don't think that needs to hold anything up.

Other then my note on an extra comment which I don't think is needed (see [this](https://github.com/datastax/java-driver/pull/542#discussion_r46607156)), I am :+1: on merging this.

adutra:
Removed the spurious comment and squashed again. :100:  for your branch with tests for connection errors!

PR #543, opened 2015-12-02
JAVA-1006: Reorganize exceptions hierarchy.

adutra:
The first commit reorganizes the current hierarchy:
- `OverloadedException` and `BootstrappingException` inherit from `QueryExecutionException`;
- `UnpreparedException` inherits from `QueryValidationException`;
- `UnresolvedUserTypeException` doesn't inherit from `InvalidTypeException` (it's current usage does not have anything to do with codec failures).

The second commit promotes package-private exceptions to public. The exceptions in this group are those that can appear in client code, either via `NoHostAvailableException` or through latency trackers (e.g. Query Logger).
- `ConnectionException`, `TransportException` and `OperationTimedOutException`: all can appear in NHAE and latency trackers;
- `BusyConnectionException` (which is not checked anymore - can appear in NHAE, see [here](https://github.com/datastax/java-driver/blob/cdd32afd5e1ac9c9dfec0bb0a19c7aa0632e00f6/driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java#L316-L316)) - could maybe be renamed to `NoMoreStreamIdAvailableException`?
- `UnsupportedProtocolVersionException`: appears in NHAE (see [here](https://github.com/datastax/java-driver/blob/cdd32afd5e1ac9c9dfec0bb0a19c7aa0632e00f6/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java#L215-L215))

The only non-public exception that hasn't changed is `ClusterNameMismatchException` (never appears in client code).


driver-core/src/main/java/com/datastax/driver/core/StreamIdGenerator.java
```diff
@@ -95,7 +95,7 @@ public int next() throws BusyConnectionException {
             if (id >= 0)
                 return id + (64 * j);
         }
-        throw new BusyConnectionException();
+        return -1;
```
adutra, line 98: This is the only functional change in this pr; I had to throw BusyConnectionException outside of this class in order to have the host address.

PR #544, opened 2015-12-02
JAVA-1005: DowngradingConsistencyRetryPolicy does not work with EACH_QUORUM when 1 DC is down.

adutra:
I'm proposing to downgrade to `ONE` if `aliveReplicas == 0`. This is a bit like picking a number blindly, but with `EACH_QUORUM` the number of alive replicas cannot be trusted.


tolbertam:
looks good :+1:, but it would be nice to merge this one after #587 is merged and I can rewrite tests to use Scassandra.

tolbertam:
Now that #587 is merged, i'll rebase and update the tests.

tolbertam:
Rebased on 2.1 and updated the tests to work with the new way Retry Policies are tested.   Retained old branch [here](https://github.com/tolbertam/java-driver/tree/java1005_pre21).  Only changes were to the tests and the changelog.

PR #545, opened 2015-12-02
Code quality - Exception handlers should preserve the original exception

faisal-hameed:
This pull request is focused on resolving occurrences of Sonar rule 
squid:S1166 - (Exception handlers should preserve the original exception) 
When handling a caught exception, the original exception's message and stack trace should be logged or passed forward

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid%3AS1166

Please let me know if you have any questions.

Faisal


olim7t:
Hi Faysal,

Sorry but we don't agree with enforcing that as an automatic rule. For example, in cases such as this one including the original exception brings little value and only makes the stack trace longer:

``` java
} catch (IndexOutOfBoundsException e) {
    throw new DriverInternalError("Not enough bytes to read an UTF8 serialized string preceded by it's 2 bytes length.");
```

So we prefer handling this on a case-by-case basis.

PR #546, opened 2015-12-02
Code quality fix - Constructors should only call non-overridable methods

faisal-hameed:
This pull request is focused on resolving occurrences of Sonar rule 
squid:S1699 - (Constructors should only call non-overridable methods) 
Calling an overridable method from a constructor could result in failures or strange behaviors when instantiating a subclass which overrides the method.

You can find more information about the issue here: https://dev.eclipse.org/sonar/coding_rules#q=squid%3AS1699

Please let me know if you have any questions.

Faisal

### DIFF
--- a/clirr-ignores.xml
+++ b/clirr-ignores.xml
@@ -11,6 +11,41 @@
 <differences>
 
   <difference>
+    <differenceType>7014</differenceType> <!-- method now final -->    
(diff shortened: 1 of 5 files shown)


adutra:
Declining as this PR introduces breaking changes.

PR #549, opened 2015-12-04
JAVA-1007: Make SimpleStatement and QueryBuilder "detached" again.

olim7t:
Signature changes in statement classes:
- `Statement#getRoutingKey(ProtocolVersion, CodecRegistry)`
- `RegularStatement#getValues(ProtocolVersion, CodecRegistry)`
- `RegularStatement#getQueryString(CodecRegistry)`, with a no-arg overload that uses the default instance.
- `RegularStatement#hasValues(CodecRegistry)`, with a no-arg overload that uses the default instance.

See javadocs for when to use each method. I added a paragraph in `BuiltStatement` javadoc to explain why it needs the codec registry.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/ArrayBackedResultSet.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ArrayBackedResultSet.java
@@ -45,10 +45,12 @@ abstract class ArrayBackedResultSet implements ResultSet {
     private final boolean wasApplied;
 
     protected final ProtocolVersion protocolVersion;
+    protected final CodecRegistry codecRegistry;
 
-    private ArrayBackedResultSet(ColumnDefinitions metadata, Token.Factory tokenFactory, List<ByteBuffer> firstRow, ProtocolVersion protocolVersion) {
+    private ArrayBackedResultSet(ColumnDefinitions metadata, Token.Factory tokenFactory, List<ByteBuffer> firstRow, ProtocolVersion protocolVersion, CodecRegistry codecRegistry) {
         this.metadata = metadata;
         this.protocolVersion = protocolVersion;
+        this.codecRegistry = codecRegistry;
         this.tokenFactory = tokenFactory;
         this.wasApplied = checkWasApplied(firstRow, metadata, protocolVersion);
     }
@@ -79,8 +81,8 @@ abstract class ArrayBackedResultSet implements ResultSet {
                 assert r.metadata.pagingState == null || info != null;
 
                 return r.metadata.pagingState == null
-                    ? new SinglePage(columnDefs, tokenFactory, protocolVersion, r.data, info)
-                    : new MultiPage(columnDefs, tokenFactory, protocolVersion, r.data, info, r.metadata.pagingState, session, statement);
+                    ? new SinglePage(columnDefs, tokenFactory, protocolVersion, columnDefs.codecRegistry, r.data, info)
+                    : new MultiPage(columnDefs, tokenFactory, protocolVersion, columnDefs.codecRegistry, r.data, info, r.metadata.pagingState, session, statement);
 
             case SET_KEYSPACE:
             case SCHEMA_CHANGE:
@@ -105,7 +107,7 @@ abstract class ArrayBackedResultSet implements ResultSet {
 
     private static ArrayBackedResultSet empty(ExecutionInfo info) {
         // We could pass the protocol version but we know we won't need it so passing a bogus value (null)
-        return new SinglePage(ColumnDefinitions.EMPTY, null,  null, EMPTY_QUEUE, info);
+        return new SinglePage(ColumnDefinitions.EMPTY, null,  null, null, EMPTY_QUEUE, info);
     }
 
     public ColumnDefinitions getColumnDefinitions() {
@@ -166,9 +168,10 @@ abstract class ArrayBackedResultSet implements ResultSet {
         private SinglePage(ColumnDefinitions metadata,
                            Token.Factory tokenFactory,
                            ProtocolVersion protocolVersion,
+                           CodecRegistry codecRegistry,
                            Queue<List<ByteBuffer>> rows,
                            ExecutionInfo info) {
-            super(metadata, tokenFactory, rows.peek(), protocolVersion);
+            super(metadata, tokenFactory, rows.peek(), protocolVersion, codecRegistry);
             this.info = info;
             this.rows = rows;
         }
@@ -230,6 +233,7 @@ abstract class ArrayBackedResultSet implements ResultSet {
         private MultiPage(ColumnDefinitions metadata,
                           Token.Factory tokenFactory,
                           ProtocolVersion protocolVersion,
+                          CodecRegistry codecRegistry,
                           Queue<List<ByteBuffer>> rows,
                           ExecutionInfo info,
                           ByteBuffer pagingState,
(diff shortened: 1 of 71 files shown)


olim7t:
Force-pushed to squash and rebase on current 3.0.

driver-core/src/main/java/com/datastax/driver/core/ArrayBackedResultSet.java
```diff
@@ -78,9 +80,14 @@ static ArrayBackedResultSet fromMessage(Responses.Result msg, SessionManager ses
                 // this explicitly because MultiPage implementation don't support info == null.
                 assert r.metadata.pagingState == null || info != null;
 
+                // If session is null, this is an internal call so we know it won't use custom types. It doesn't matter which registry is used.
+                CodecRegistry codecRegistry = (session == null)
```
adutra: After JAVA-996, this can be replaced with `columnDefs.codecRegistry` - it is never null.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
```diff
@@ -235,17 +286,18 @@ public Boolean isIdempotent() {
...
-            return getQueryString();
+        try {
+            if (forceNoValues) {
+                return getQueryString();
+            }
+            StringBuilder queryString = buildQueryString(null, CodecRegistry.DEFAULT_INSTANCE);
+            return maybeAddSemicolon(queryString).toString();
+        } catch (CodecNotFoundException e) {
+            // Ugly but we have absolutely no context to get the registry from
+            return String.format("built query (could not generate with default codec registry: %s)", e.getMessage());
```
adutra, line 299: I think we can safely propose something better here:

``` java
StringBuilder queryString = buildQueryString(new ArrayList<Object>(), CodecRegistry.DEFAULT_INSTANCE);
return maybeAddSemicolon(queryString).toString();
```

When the parameter `variables` is non null, in `Utils.appendValue`line 94 the only case the codec registry could get accessed is if `Utils.isSerializable(value)` returns false – which can only happen for a `Number` or a collection thereof. So for an unknown object, a bind marker would be inserted instead. E.g.:

``` java
System.out.println(insertInto("foo").value("a", new Foo()));
```

This would print:

```
INSERT INTO foo (a) VALUES (?);
```

driver-core/src/main/java/com/datastax/driver/core/BoundStatement.java
```diff
@@ -222,9 +222,12 @@ public BoundStatement bind(Object... values) {
      * precedence. If the routing key has been set through {@link PreparedStatement#setRoutingKey} then that is used
      * next. If neither of those are set then it is computed.
      *
+     * @param protocolVersion unused by this implementation (no internal serialization is required to compute the key).
+     * @param codecRegistry unused by this implementation (no internal serialization is required to compute the key).
+     *
      * @return the routing key for this statement or {@code null}.
      */
-    public ByteBuffer getRoutingKey() {
+    public ByteBuffer getRoutingKey(ProtocolVersion protocolVersion, CodecRegistry codecRegistry) {
```
adutra, line 231: Missing `@Override`. We also could propose an overloaded version with no arguments.

driver-core/src/main/java/com/datastax/driver/core/SimpleStatement.java
```diff
@@ -22,49 +22,89 @@
...
             throw new IllegalArgumentException("Too many values, the maximum allowed is 65535");
         this.query = query;
         this.values = values;
-        this.cluster = cluster;
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * Note that this implementation doesn't actually use the protocol version and codec registry.
```
adutra: Nit: no need to mention "protocol version" here.

driver-core/src/main/java/com/datastax/driver/core/SimpleStatement.java
```diff
@@ -107,13 +153,16 @@ public Object getObject(int i) {
...
+     * @param codecRegistry unused by this implementation (no internal serialization is required to compute the key).
+     *
      * @return the routing key set through {@link #setRoutingKey} if such a key
      * was set, {@code null} otherwise.
      *
      * @see Statement#getRoutingKey
      */
     @Override
-    public ByteBuffer getRoutingKey() {
+    public ByteBuffer getRoutingKey(ProtocolVersion protocolVersion, CodecRegistry codecRegistry) {
```
adutra, line 153: Maybe keep the old method as an overloaded version with no args?

driver-core/src/main/java/com/datastax/driver/core/ArrayBackedResultSet.java
olim7t: Done, thanks.

driver-core/src/main/java/com/datastax/driver/core/Statement.java
```diff
@@ -24,6 +24,7 @@
 import com.datastax.driver.core.exceptions.UnsupportedProtocolVersionException;
 import com.datastax.driver.core.policies.RetryPolicy;
 import com.datastax.driver.core.querybuilder.BuiltStatement;
+import com.datastax.driver.core.schemabuilder.SchemaStatement;
```
adutra: Unused import.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
olim7t, line 299: Isn't there also the issue where the value is a list containing both a non-serializable thing (e.g. function call) _and_ a custom object?
I agree that what you propose would have less chance of failing, but it would also never inline default types (that's arguably a fair tradeoff).

driver-core/src/main/java/com/datastax/driver/core/BoundStatement.java
olim7t, line 231: The reason I didn't add a no-arg alternative for this (and for `getValues`), is that I think it's typically not called from client code, only internally by the driver.

driver-core/src/main/java/com/datastax/driver/core/schemabuilder/SchemaStatement.java
```diff
@@ -41,8 +43,14 @@
 
     abstract String buildInternal();
 
+    /**
+     * {@inheritDoc}
+     *
+     * Note that this implementation doesn't actually use the protocol version and codec registry.
```
adutra: No need to mention "protocol version" here.

driver-core/src/main/java/com/datastax/driver/core/schemabuilder/SchemaStatement.java
```diff
@@ -50,13 +58,19 @@ public String getQueryString() {
...
-    public ByteBuffer[] getValues() {
+    public ByteBuffer[] getValues(ProtocolVersion protocolVersion, CodecRegistry codecRegistry) {
         // DDL statements never have values
         return new ByteBuffer[0];
     }
 
+    /**
+     * {@inheritDoc}
+     *
+     * Note that this implementation doesn't actually use the protocol version and codec registry.
```
adutra: Again, no need to mention "protocol version".

driver-core/src/main/java/com/datastax/driver/core/SimpleStatement.java
olim7t: Good point, I forgot that from a previous version.
Actually I'm going to remove the javadoc altogether, because the one in the parent class is detailed enough.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
adutra, line 299: > Isn't there also the issue where the value is a list containing both a non-serializable thing (e.g. function call) and a custom object?

I don't think so because that would have been caught by line 85 in `Utils.appendValue`, and the entire list would have been inlined.

olim7t, line 299: Yes, so there is still a case where we could get a CodecNotFoundException. I'm going to leave the catch block, but I'll change the inside to use what you suggest, it's better to have it work in as many cases as possible, even it that means not inlining values.

adutra, line 299: No, you are right; I just tested my supposition and it does not work because `null` is passed to `appendList`. Please disregard :)

PR #550, opened 2015-12-05
Fetch the whole peers tables if not found specific host

wy96f:
We deploy cassandra in two datacenters(multi region). We use private ip as listen_address(also rpc_address) and public ip as broadcast_address. The public ip will change if instance restarts so we have to modify the broadcast_address. Cassandra cluster will handle it correctly but the driver will ignore the host on reconnection as no row is found in peer's table by using old public ip. In this case we can query the whole peers table as a fallback.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
@@ -450,7 +450,10 @@ class ControlConnection implements Host.StateListener, Connection.Owner {
                 ? new DefaultResultSetFuture(null, cluster.protocolVersion(), new Requests.Query(SELECT_LOCAL))
                 : new DefaultResultSetFuture(null, cluster.protocolVersion(), new Requests.Query(SELECT_PEERS + " WHERE peer='" + host.listenAddress.getHostAddress() + '\''));
             c.write(future);
-            return future.get().one();
+            Row row = future.get().one();
+            if (row != null) {
+                return row;
+            }
         }
 
         // We have to fetch the whole peers table and find the host we're looking for


olim7t:
> the driver will ignore the host on reconnection

I don't understand how the reconnection happens in the first place if the public IP has changed. `fetchNodeInfo` can be called either by the driver's scheduled reconnection attempts, or after receiving a server-side event to indicate that the node is up or added. But both of these mechanism rely on the broadcast_rpc_address, which in your case is the public IP if I understand correctly. If a public IP is gone, then the driver should never try to fetch the node info for that host again.

Besides, the address we use to connect to a node (`Host.getSocketAddress()`) is currently immutable. So even if we manage to trigger `fetchNodeInfo` for that host, the driver will still use the old address to connect.

Have you considered using elastic IPs?

wy96f:
Thanks for the reply. Out app just access the local datacenter, so the broadcast_rpc_address is internal ip, that's to say host.getSocketAddress() is too. After reconnecting successfully to the machine where public ip has changed, fetchNodeInfo can't find info with old public ip(the key of peer table is broadcast_address, given we deploy in two datacenters, this one is public ip). To address this problem we can fetch whole peers table and get info from the row in which host.getSocketAddress() is equivalent to rpc_address column since the broadcast_rpc_address of the machine doesn't change(it's still internal ip).

olim7t:
Thanks for clarifying, I get it now. I've created [JAVA-1038](https://datastax-oss.atlassian.net/browse/JAVA-1038) to keep track of this issue.

adutra:
Closing in favor of #596. Thanks for your contribution!

olim7t:
PS- we cherry-picked your commit in the other PR, so it's still properly attributed :)

PR #551, opened 2015-12-06
added DCAwareFailoverRoundRobinPolicy which allows fine control of DC failover scenarios

adejanovski:
Built on top of the DCAwareRoundRobinPolicy, this policy uses a one way switch in case a defined number of nodes are down in the local DC. 
As stated, the policy by default prevents switching back to the local DC in order to prevent inconsistencies and give ops teams the ability to repair the local DC before switching back manually.
A safe auto switch back to local DC is possible after downtime*switchBackDelayFactor to give enough time for hints to be executed, and prevents switching back if downtime > noSwitchBackDowntimeDelay, if both parameters are specified.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/policies/DCAwareFailoverRoundRobinPolicy.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/policies/DCAwareFailoverRoundRobinPolicy.java
@@ -109,8 +109,7 @@ public class DCAwareFailoverRoundRobinPolicy implements LoadBalancingPolicy,
 	private Date switchedToBackupDcAt;
 
 	/**
-	 * Automatically switching back to local DC is possible after : downtime*
-	 * {@code switchBackDelayFactor}
+	 * Automatically switching back to local DC is possible after : downtime*{@code switchBackDelayFactor}
 	 */
 	private Float switchBackDelayFactor=(float)1000;
 
--- a/driver-core/src/main/java/com/datastax/driver/core/policies/DCAwareFailoverRoundRobinPolicy.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/policies/DCAwareFailoverRoundRobinPolicy.java
@@ -1,0 +1,609 @@
+/*
+ *      Copyright (C) 2012-2015 DataStax Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+package com.datastax.driver.core.policies;
+
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import com.google.common.base.Joiner;
+import com.google.common.base.Preconditions;
+import com.google.common.base.Strings;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.AbstractIterator;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.datastax.driver.core.Cluster;
+import com.datastax.driver.core.Configuration;
+import com.datastax.driver.core.Host;
+import com.datastax.driver.core.HostDistance;
+import com.datastax.driver.core.Statement;
+import com.datastax.driver.core.policies.CloseableLoadBalancingPolicy;
+import com.datastax.driver.core.policies.LoadBalancingPolicy;
+import com.datastax.driver.core.policies.DCAwareRoundRobinPolicy.Builder;
+
+/**
+ * A data-center aware Round-robin load balancing policy with DC failover
+ * support.
+ * <p>
+ * This policy provides round-robin queries over the node of the local data
+ * center. It also includes in the query plans returned a configurable number of
+ * hosts in the remote data centers, but those are always tried after the local
+ * nodes. In other words, this policy guarantees that no host in a remote data
+ * center will be queried unless no host in the local data center can be
+ * reached.
+ * <p>
+ * If used with a single data center, this policy is equivalent to the
+ * {@code LoadBalancingPolicy.RoundRobin} policy, but its DC awareness incurs a
+ * slight overhead so the {@code LoadBalancingPolicy.RoundRobin} policy could be
+ * preferred to this policy in that case.
+ * <p>
+ * On top of the DCAwareRoundRobinPolicy, this policy uses a one way switch in
+ * case a defined number of nodes are down in the local DC. As stated, the
+ * policy never switches back to the local DC in order to prevent
+ * inconsistencies and give ops teams the ability to repair the local DC before
+ * switching back manually.
+ */
+public class DCAwareFailoverRoundRobinPolicy implements LoadBalancingPolicy,
+		CloseableLoadBalancingPolicy {
+
+	private static final Logger logger = LoggerFactory
+			.getLogger(DCAwareFailoverRoundRobinPolicy.class);
+
+	/**
+     * Returns a builder to create a new instance.
+     *
+     * @return the builder.
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+	
+	private final String UNSET = "";
+
+	private final ConcurrentMap<String, CopyOnWriteArrayList<Host>> perDcLiveHosts = new ConcurrentHashMap<String, CopyOnWriteArrayList<Host>>();
+	private final AtomicInteger index = new AtomicInteger();
+
+	@VisibleForTesting
+	volatile String localDc;
+	@VisibleForTesting
+	volatile String backupDc;
+
+	/**
+	 * Current value of the switch threshold. if {@code hostDownSwitchThreshold}
+	 * <= 0 then we must switch.
+	 */
+	private AtomicInteger hostDownSwitchThreshold = new AtomicInteger();
+	
+	/**
+	 * Initial value of the switch threshold
+	 */
+	private final int initHostDownSwitchThreshold;
+
+	/**
+	 * flag to test if the switch as occurred
+	 */
+	private AtomicBoolean switchedToBackupDc = new AtomicBoolean(false);
+	
+	/**
+	 * Time at which the switch occurred
+	 */
+	private Date switchedToBackupDcAt;
+
+	/**
+	 * Automatically switching back to local DC is possible after : downtime*
+	 * {@code switchBackDelayFactor}
+	 */
+	private Float switchBackDelayFactor=(float)1000;
+
+	/**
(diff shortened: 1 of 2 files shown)


adejanovski:
A bit of explanation on what motivates building this policy and how it should be used : 

# DC Aware Failover Round Robin Policy for Datastax Java Driver

Properly handling failover in multi DC Cassandra clusters with LOCAL_\* CL usually require writing specific code to properly process failover scenarios.

The Cassandra setups that fail delivering automatic DC failure handling are those with 2 DCs or those with 1+ real time DCs and 1+ analytical/search DC.

Having a 2 DC cluster and using CL.QUORUM will lead to systematic failure when a DC goes down. 
Using CL.ONE will lead to inconsistencies if a DC goes down and comes back up after a while.
Using CL.LOCAL_QUORUM with DCAwareRoundRobinPolicy will segregate the traffic to a single DC, but the load balancing policy will prevent switching to the other DC in case of failure (in its basic configuration).
This is a conservative behavior, because if done differently, when the primary DC comes back up it will probably not be up to date, and it could even be out of the hinted handoff window (which will require a full repair).
In case you have an analytical or a search DC, you must segregate traffic in order to guarantee low latencies, which forces the use of CL.LOCAL_\* consistencies. The behavior in case of failure is the same as the one above.

Handling those failure scenarios has to be implemented in client applications, by doing the following :
- detect how many nodes are down in the local DC
- if too many nodes are down to allow queries to succeed (depends on the replication factor), the app has to switch to a backup dc (either having two connections and switching between them, either recreating a new connection with a different LB policy conf).
- Switching automatically back at any time to the local DC must be prevented because repair might be necessary in order to regain consistency

The DCAwareFailoverRoundRobinPolicy is based on the DCAwareRoundRobinPolicy and handles the above one way switch without having to code it. It takes the following parameters : 
- localDc : the primary DC of the app
- backupDc : the DC to switch to in case the primary DC loses a defined number of nodes
- hostDownSwitchThreshold : the number of lost nodes that will trigger the switch
- switchBackDelayFactor : The connection can switch back if uptime >= downtime*switchBackDelayFactor (gives time for hinted handoff to complete)
- noSwitchBackDowntimeDelay : Switch back cannot happen if downtime > noSwitchBackDowntimeDelay (in seconds) - usually the hinted handoff window (default in Cassandra is 3h)

This policy does a one way switch to the backup DC to prevent inconsistencies, and the connection must be recreated in order to switch back to the local DC (which should require either rebooting the app, or recreating the Cluster and Session objects), unless auto back switch conditions are fulfilled : 
- Local DC has regained enough nodes
- uptime is superior to downtime*switchBackDelayFactor
- downtime has been lower than noSwitchBackDowntimeDelay

# Usage

Include this project in your classpath and use it as any other load balancing policy when creating the cluster object:

``` java
Cluster cluster = Cluster.builder()
                     .addContactPoints("127.0.0.1","127.0.0.2","127.0.0.3","127.0.0.4","127.0.0.5")
                     .withLoadBalancingPolicy(new DCAwareFailoverRoundRobinPolicy("primaryDc","backupDc",2))                    
                     .build();
```

The above will segregate coordination traffic to DC 'primaryDc' and if at least 2 nodes are lost in this DC, switch to 'backupDc' and never go back.

To authorize auto back switch you can add the switchBackDelayFactor and noSwitchBackDowntimeDelay : 

``` java
Cluster cluster = Cluster.builder()
                     .addContactPoints("127.0.0.1","127.0.0.2","127.0.0.3","127.0.0.4","127.0.0.5")
                     .withLoadBalancingPolicy(new DCAwareFailoverRoundRobinPolicy("primaryDc","backupDc", 2, 1.5, 10800))                   
                     .build();
```

The above will allow automatically switching back if conditions are fulfilled : 
- downtime didn't last more than 3 hours (10800 seconds)
- uptime since local DC recovery has lasted more than downtime*1.5 (if downtime was 30 mn, back switch will be prevented for 45 mn after local DC recovery)

doanduyhai:
I am +1 on this PR. Alexander already deployed this policy in production so it's kind of battlefield-tested. I'll give a more in-depth code review in the next days

olim7t:
I'm not questioning the validity of the use case, nor the implementation, but this seems a bit too specific to be shipped with the driver by default.
One could imagine other ways to handle failover (for example do it at the application level with a load balancer). The goal of the driver is not to cover every use case out of the box, but provide enough flexibility to allow you to do it yourself, which is why those policies are pluggable.
Custom implementations can still be shared as standalone projects.

zznate:
> The goal of the driver is not to cover every use case out of the box

@olim7t I agree completely with the sentiment, but this is a common use case from what we've seen with our clients. I would go so far as to call this functionality most folks might expect to be there when using terms like "DC Aware" and "Load Balancing." Specifically, that the traffic is managed by the policy based on the status of the datacenters.

adejanovski:
@olim7t I think you are wrong to consider this as a specific use case, because as Cassandra spreads in the legacy IT industry, it is getting considered as the common use case.
Cassandra pretends to be the "always on" database, which supposes failover is handled naturally and developers have limited efforts to do in order to achieve resiliency.
In practice though, you cannot achieve multi DC failover (which is a common use case even at Netflix) without handling the switch at the app / load balancing level. 
Currently the drivers DCAwareLBPolicy will allow working with a DC that has 99% of its node down without a problem although it will drop most of the requests.

While my approach is a bit naïve (on big clusters losing 3 nodes doesn't mean you lost a token range at RF=3), it provides enough automation to handle most failover scenarios while protecting consistency.
An improvement that should come quickly is to store which token ranges are lost when a node goes down so that we know when quorum or one is 100% achievable or not (this way you'll cover the big cluster case and will switch only when you're sure the DC has a problem).

Without it, you're left with : "Hey, Cassandra is always on but gather centralized metrics to see if too many queries are failing at the app level then switch yourself by any mean, and you'd better have a microservices architecture with ZUUL or whatever service load balancer, otherwise you should only use a single DC... or Oracle..."
Being "always on", means automatically handle as many problems and start failing when all hope is lost.

You must ease the use of Cassandra in companies that are not Apple, Netflix or Spotify, which may have more limited resources and a less shiny up to date architecture. 

Monoliths are everywhere and still deserve a redundant database, don't you think ?

And sure, I can share the policy myself but let's be honest, distribution, knowledge and support will be much much better if you include it in the core driver. 

I'm also sure you'll improve it a lot more than I can.

doanduyhai:
I suggest to put this policy in the **driver-extras** module, along side with the codecs package. This is the perfect place to put community contributions like extra codecs or policies

rsds143:
-1. This is very frequently requested, but I have personally yet to see a failure condition it would have helped (not saying I'm authoritative on this, but I haven't seen it, so I'm interested in the details of a use case where this really helped).  Most DC failures that I've seen are:

1) system problems that affect the app servers as well, IE network traffic. Where this will have no positive effect.
2) When Cassandra specific, the issue either takes down the whole cluster (IE giant batches that should have never have been pushed to prod or really slow cluster scan queries before paging was in the driver), or a couple of replicas fall over (fat partition). In either case failing over to another DC would not help the issue but only spread the contagion, and add extra latency to the request for fun (assuming the other DC is not located close by), while eventually overwhelming the 'backup dc'.
3) Total disaster, where the DC isn't coming back up. Again this isn't helping.

The only valid use case I can come up with is on a per replicaset type basis, and even then it'd have to be with a low latency backup DC link, and the load balancing would have to be smart enough to be only routing a given token range to the backup dc. For example: 

What happens when nodes A, B are down in DC1 but nodes C and D are down in DC2? Does the driver split traffic between the up replicas? Or do you just chain faillover back and forth?

This just strikes me as an extra tarpit to catch new users in, and the cure maybe worse than the disease.

adejanovski:
@rssvihla my company is using a cluster with 3 DCs : 2 operational DCs and 1 analytical DC.
The 2 operational DCs are not far from each other so apps located in both can access either one with acceptable latency (we're not on AWS or any other cloud provider, and operate our own hardware).
We cannot use QUORUM CL because it would involve the analytical DC, and even without an analytical DC, losing a DC would mean losing QUORUM.
Moreover, each DC uses a single physical rack for its Cassandra servers (we have other servers in other racks), and although it would have been better to use 3 distinct racks in each DC, it was just not possible for various reasons (mostly time and money).
To avoid having a backup DC that sleeps waiting to be used, each app/service using the C\* cluster has one of the DCs as primary and the other as backup, which allows using LOCAL_QUORUM and splitting (although unevenly) the load between the 2 DCs. 
Some time after setting this all up, our ops team wanted to shutdown DC1 because the rack needed an electrical intervention (which would not impact the apps accessing DC1) and given how the DC Aware LB policy works, with a planned 8 hours shutdown, apps that would've used DC2 were going to switch back to DC1 as soon as it got restarted, despite the fact that it wouldn't be up to date before we ran a full repair (which can take a few days).

To handle this "planned" intervention, we would have had to reconfigure/redeploy all apps using DC1 to use DC2, and do the opposite when DC1 would have been up to date after repair.

Failure can occur for many different reasons and affect only a subset of servers in a DC, like losing 2 out of 3 availability zones in a AWS DC (because you know, sh... happens), or problems with a specific network router, etc... I see many cases where the apps could be unaffected by a system/network problem or maintenance that would only affect a C\* DC. 
We also use an ESB (and a JMS bus) that has limited capability to switch database connections and route traffic to given DCs, which would make it very hard to colocate apps with the C\* DCs. Even harder would be to detect the fact that we should switch the app layer because the C\* DC is malfunctioning : it requires collecting metrics that help identifying that failures are related to Cassandra (for whatever reason). 
That showed that in case of unplanned outage, we would have had to develop routines to detect and assert that the C\* DC was not working properly. Although it was feasible, it would require a lot of work and understanding from all teams to get it right.

Working at the driver level has been way simpler, and makes it very easy to reason for developers (learning how to work properly with eventual consistency is already hard for most...), and doesn't require any specific development, just configuring the LB policy properly.

You mentioned that switching DCs could spread the disease to the other DC but that would happen in any failover scenario where the outage is caused by a few big batches or reads with too many tombstones. I guess any company with more than one C\* DC has to implement a failover scenario (may it be at the service routing level or at the app level), and they would all end up spreading diseases the same way. 

Routing traffic for given token ranges would be far too complicated (and not feasible without systematic prepared statements) for what it would bring. My approach is way more binary : either the DC is working (ie you can get quorum for any token range) either it is not and then you should switch.

The current implementation is not bulletproof though and should be improved in the future, mostly by not relying on how many nodes are down, but instead on how many token ranges lost quorum.
If you lost 2 nodes in DC1 and switch to DC2 and then lose two nodes there too, the policy won't switch back to DC1. The back switch is optional and will only happen if the 2 nodes are back up in DC1, and enough time has passed since they came back up.
I see a scenario where it would be possible to switch back and forth : when the outage is caused by queries overloading some specific nodes (due to batches or tombstones), and switching to DC2 would overload it, letting DC1 come back up properly, then it switches back to DC1 and so on.
Maybe we could add a condition on the number of switches in a certain time interval that would block switching to prevent it from happening again and again.

rsds143:
@adejanovski lot to respond to, but I don't think we're going to be on the same page so I'll keep it brief:

What you call a "C\* DC problem" I call a "replica problem" that was my point. There isn't a lot of binary situations where the whole C\* DC is down and the app servers in that DC are up without it being a code issue (and then in general the problem spreads as you point out, but you're missing failover leads to load bursts making already bad worse, and that it's still not helping a total failure situation. Failover is a "deck chair arrangement" at best in that scenario). 

Even in the some of the examples you give part of a DC is down rather than "all of the DC" being down and trying to detect a DC as "up" or "down" ends up being a continuum rather than a binary thing (ie how many hosts are up). Getting that "right" would completely vary from team to team and even project to project.

There is lots else in that response I can't get on board with, but ultimately, If you make this token range failover specific it would be a lot more consistent with actual Cassandra failure modes and still accomplish your goals.

adutra:
I'm closing this long-lasting PR since we still believe that cross DC failover should be done in the application level. See JAVA-2041, we are even going to deprecate all cross-DC failover capabilities in `DCAwareRoundRobinPolicy` in the next minor (3.7), and such features will disappear in the next major (4.0).

PR #552, opened 2015-12-07
JAVA-739: Expose private listen_address of each Cassandra node.

adutra:
This commit exposes both `broadcast_address` and `listen_address`.
However `listen_address` is not present in `system.peers`, so in my opinion the driver is still not able to provide the info required by the Spark connector team.


adutra:

adutra:
Closing in favor of #570.

PR #553, opened 2015-12-08
JAVA-1013: TimestampCodec should format dates in human-readable format.

adutra:


olim7t:
LGTM.
UTC is how a date would have to be formatted if it was "inlined" in the query string instead of sent along as a value, so it's the right timezone to use.

tolbertam:
:+1:

PR #554, opened 2015-12-08
wip; Graph Adaptation for the Java Driver

newkek:
This is a wip but I'm opening a PR so we can directly centralize comments and remarks here.

The current _graph_ branch, at the time this PR is open, is based on the Java Driver 3.0.0-alpha4 version, and needs to be rebased on the latest 3.x version available.


driver-dse/src/main/java/com/datastax/driver/graph/AbstractGraphStatement.java
```diff
@@ -0,0 +1,141 @@
...
+     * As of right now, mandatory fields for DSE Graph are :
+     * - graph-keyspace : Must be set by users
+     * - graph-source : default is "default"
+     * - graph-language : default is "gremlin-groovy"
+     *
+     */
+    static boolean checkStatement(AbstractGraphStatement graphStatement) {
+        return graphStatement.getGraphOptions().containsKey(GraphSession.GRAPH_SOURCE_KEY)
+            && graphStatement.getGraphOptions().containsKey(GraphSession.GRAPH_LANGUAGE_KEY)
+            && graphStatement.getGraphOptions().containsKey(GraphSession.GRAPH_SOURCE_KEY);
```
olim7t: Duplicate key GRAPH_SOURCE_KEY, should probably be GRAPH_KEYPSACE_KEY (which has a typo btw).

newkek: I thought I had fixed that, done.

mfiguiere:
I guess you guys have thought about it, but is there no way to use `*GraphStatement`s with the execute methods of the `Session`, rather than using a `GraphSession` wrapper? It's a pity to have this `GraphSession` class adapting most of the `Session` methods and adding only a few other configuration methods.

newkek:
We had discussed it internally indeed and it resulted that the goal of the integration of graph is to get these new Graph features without impacting the core components of the drivers. There are certainly more elegant ways to do it when modifying core components like `Session`, but this solution was chosen so we could have all the Graph components in a separate module (i.e. for Java _driver-dse_ module) and have no impact on the core of the drivers.

newkek:
I've just noticed that on Github the commits history is completely wrong.

driver-dse/src/main/java/com/datastax/driver/graph/AbstractGraphStatement.java
```diff
@@ -0,0 +1,141 @@
...
+     * Set the Graph keyspace to use for this statement.
+     *
+     * @param graphKeyspace The keyspace to set. Throws an exception if the parameter is null or empty since this parameter is mandatory.
+     * @return This {@link com.datastax.driver.graph.AbstractGraphStatement} instance to allow chaining call.
+     */
+    public AbstractGraphStatement setGraphKeyspace(String graphKeyspace) {
+        if (graphKeyspace == null || graphKeyspace.isEmpty()) {
+            throw new InvalidQueryException("You cannot set null value or empty string to the keyspace for the Graph, this field is mandatory.");
+        }
+        this.payload.put(GraphSession.GRAPH_LANGUAGE_KEY, ByteBuffer.wrap(graphKeyspace.getBytes()));
```
tolbertam: shouldn't this be `GraphSession.GRAPH_KEYSPACE_KEY`?

driver-dse/src/main/java/com/datastax/driver/graph/GraphTraversalResult.java
```diff
@@ -0,0 +1,100 @@
...
+        return this.jsonString;
+    }
+
+    /**
+     * Get the {@link com.datastax.driver.graph.GraphData} object for the specified key.
+     *
+     * @param keyOrIndex Can be either a String or a int/Integer. Must not be null, or an exception will be thrown.
+     * @return A GraphData instance containing the value for the Json entity required.
+     */
+    public GraphData get(Object keyOrIndex) {
```
tolbertam: what do you think about also adding a `has` method?  Since `get` delegates to `rootNode.get(x)` which throws an Exception when no key/index exists, a `has` would be a nice check-before construct.

driver-dse/src/main/java/com/datastax/driver/graph/GraphSession.java
```diff
@@ -0,0 +1,225 @@
...
+    final static String GRAPH_KEYPSACE_KEY;
+    final static String GRAPH_REBINDING_KEY;
+
+    // Add static DefaultPayload values for Graph
+    final static String DEFAULT_GRAPH_LANGUAGE;
+    final static String DEFAULT_GRAPH_SOURCE;
+    final static Map<String, ByteBuffer> DEFAULT_GRAPH_PAYLOAD;
+
+    static {
+        GRAPH_SOURCE_KEY        = "graph-source";
```
tolbertam: for convenience you could just assign these when you declare them above.

driver-dse/src/main/java/com/datastax/driver/graph/AbstractGraphStatement.java
newkek: Damn it I need to add unit tests for that stuff, thanks.

driver-dse/src/main/java/com/datastax/driver/graph/BoundGraphStatement.java
```diff
@@ -0,0 +1,116 @@
...
+    void copyConfigFromStatement(GraphStatement gst) {
+        this.wrappedStatement.setOutgoingPayload(gst.getGraphOptions() == null ? this.session.getDefaultGraphOptions() : gst.getGraphOptions());
+    }
+
+    // Bind variables in the PreparedStatement
+    @Override
+    BoundStatement configureAndGetWrappedStatement() {
+        JsonNodeFactory factory = new JsonNodeFactory(false);
+        JsonFactory jsonFactory = new JsonFactory();
+        ObjectMapper objectMapper = new ObjectMapper();
```
adutra: ObjectMapper instances are thread-safe and should be shared; they can even be declared as singletons.

driver-dse/src/main/java/com/datastax/driver/graph/AbstractGraphStatement.java
```diff
@@ -0,0 +1,141 @@
...
+import com.google.common.collect.Maps;
+
+import com.datastax.driver.core.Statement;
+import com.datastax.driver.core.exceptions.InvalidQueryException;
+
+abstract class AbstractGraphStatement<T extends Statement> {
+
+    protected final Map<String, ByteBuffer> payload;
+    protected T wrappedStatement;
+    protected final GraphSession session;
```
adutra: I'm not sure it's really required to keep a reference to a session; this "ties" the statement to a session, so what happens if it gets executed by a different `Session` instance?
At first glance it seems possible to get rid of this reference:
- It's mainly used to propagate `session.getDefaultGraphOptions()` from prepared to bound statements;
- `getSession()` is never used;
- `GraphStatement` needs the session to instantiate `SimpleStatement`s – but that should change once we rebase on top of current 3.0.

driver-dse/src/main/java/com/datastax/driver/graph/AbstractGraphStatement.java
```diff
@@ -0,0 +1,141 @@
...
+     *
+     */
+    static boolean checkStatement(AbstractGraphStatement graphStatement) {
+        return graphStatement.getGraphOptions().containsKey(GraphSession.GRAPH_SOURCE_KEY)
+            && graphStatement.getGraphOptions().containsKey(GraphSession.GRAPH_LANGUAGE_KEY)
+            && graphStatement.getGraphOptions().containsKey(GraphSession.GRAPH_KEYSPACE_KEY);
+
+    }
+
+    static void copyConfiguration(AbstractGraphStatement from, AbstractGraphStatement to) {
```
adutra: This method is never used.

driver-dse/src/main/java/com/datastax/driver/graph/BoundGraphStatement.java
```diff
@@ -0,0 +1,116 @@
...
+import com.datastax.driver.core.BoundStatement;
+import com.datastax.driver.core.exceptions.DriverException;
+
+/**
+ * Bound graph statement produced from a prepared statement.
+ */
+public class BoundGraphStatement extends AbstractGraphStatement<BoundStatement> {
+
+    //  "name", "value"
+    Map<String, Object> valuesMap;
```
adutra: Should be private final.

driver-dse/src/main/java/com/datastax/driver/graph/BoundGraphStatement.java
```diff
@@ -0,0 +1,116 @@
...
+        this.valuesMap = new HashMap<String, Object>();
+
+        /*
+        Need to keep the configuration from the Prepared statement to
+        the created BoundStatement, so users don't have to re configure the payload and such.
+         */
+        copyConfigFromStatement(gst);
+    }
+
+    void copyConfigFromStatement(GraphStatement gst) {
```
adutra: Several remarks:
- This method is only called by the constructor and could be inlined;
- Shouldn't the graph options be copied to `this.payload` instead, and let it be propagated to the wrapped statement when `configure()` is called?
- I would rather have the bound statement inherit its graph options from its prepared statement, whatever they are; the bound statement shouldn't directly interrogate the session.

driver-dse/src/main/java/com/datastax/driver/graph/BoundGraphStatement.java
```diff
@@ -0,0 +1,116 @@
...
+                    parameter.put("name", name);
+                    parameter.put("value", (Boolean)value);
+                } else if (value instanceof Long) {
+                    parameter.put("name", name);
+                    parameter.put("value", (Long)value);
+                } else {
+                    throw new DriverException("Parameter : " + value + ", is not in a valid format to be sent as Gremlin parameter.");
+                }
+                objectMapper.writeTree(generator, parameter);
+                this.wrappedStatement.setString(paramNumber++, stringWriter.toString());
```
adutra, line 91: `valuesMap` is a `HashMap` and iteration order is not deterministic, so it seems a bit strange to send positional parameters this way.
Moreover, the "gremlin-driver-integration" document states:

> To pass parameters, Gremlin really just expects a Map<String,Object> where the key becomes the name of the variable in the script to bind to and the value of the Map represents the value to apply to the script for that parameter name [...] It should be noted that Gremlin doesn't support positional parameters.

So shouldn't we simply add named parameters by calling `this.wrappedStatement.setXXX(name, value)`? Why do we need to convert each variable to JSON?

driver-dse/src/main/java/com/datastax/driver/graph/AbstractGraphStatement.java
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * Set manually the custom payload to use for this statement.
+     * Please see {@link com.datastax.driver.core.Statement#setOutgoingPayload(java.util.Map)} for more
+     * information.
+     *
+     * @param payload The payload to set. Note that this will override the default values set in the {@link com.datastax.driver.graph.GraphSession}.
+     * @return This {@link com.datastax.driver.graph.AbstractGraphStatement} instance to allow chaining call.
+     */
+    public AbstractGraphStatement setOutgoingPayload(Map<String, ByteBuffer> payload) {
+        this.wrappedStatement.setOutgoingPayload(payload);
```
adutra: Shouldn't this be: `this.payload.putAll(payload)` ? It seems that payloads are not consistently kept in sync between `his.payload` and `wrappedStatement`'s payload.

driver-dse/src/main/java/com/datastax/driver/graph/GraphStatement.java
```diff
@@ -0,0 +1,135 @@
...
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import com.datastax.driver.core.SimpleStatement;
+import com.datastax.driver.core.exceptions.DriverException;
+
+/**
+ * A simple graph statement implementation.
+ */
+public class GraphStatement extends AbstractGraphStatement<SimpleStatement> {
```
adutra, line 36: I would suggest renaming to `SimpleGraphStatement` for clarity.

driver-dse/src/main/java/com/datastax/driver/graph/GraphStatement.java
```diff
@@ -0,0 +1,135 @@
...
+    /**
+     * Set a parameter value for the statement.
+     *
+     * Values can be any type supported in JSON.
+     * @param name Name of the value, defined in the query. Parameters in Gremlin are named as variables, no
+     *             need for a CQL syntax like the bind marker "?" or the identifier ":" in front of a parameter.
+     *             Please refer to Gremlin's documentation for more information.
+     * @param value Any object serializable in JSON. The type will be detected automatically at statement's execution.
+     */
+    public void set(String name, Object value) {
```
tolbertam, line 130: For convenience it would be nice if this returned itself, so users could chain set, i.e.:

``` java
stmt.set("i", x).set("j", y).set("k", z);
```

tolbertam, line 130: Is it possible to provide a Vertex or Edge here?  Would be nice if I could use a previously returned Vertex / Edge here.

driver-dse/src/main/java/com/datastax/driver/graph/GraphStatement.java
```diff
@@ -0,0 +1,135 @@
...
+import com.datastax.driver.core.exceptions.DriverException;
+
+/**
+ * A simple graph statement implementation.
+ */
+public class GraphStatement extends AbstractGraphStatement<SimpleStatement> {
+
+    private final String query;
+    private final Map<String, Object> valuesMap;
+    private List<String> JsonParams;
```
tolbertam: Should rename this to 'jsonParams'

driver-dse/src/main/java/com/datastax/driver/graph/GraphStatement.java
```diff
@@ -0,0 +1,135 @@
...
+     */
+    private void processValues() {
+        JsonNodeFactory factory = new JsonNodeFactory(false);
+        JsonFactory jsonFactory = new JsonFactory();
+        ObjectMapper objectMapper = new ObjectMapper();
+        if (this.paramsHash == this.valuesMap.hashCode()) {
+            // Avoids regenerating the Json params if the params haven't changed.
+            return;
+        }
+        this.JsonParams.clear();
```
tolbertam: Since jsonParams is only ever used in configureAndGetWrappedStatement, it could just be something that is returned by processValues so it doesn't need to be an instance member.

driver-dse/src/main/java/com/datastax/driver/graph/GraphStatement.java
tolbertam, line 130: It looks like you can express a Vertex as a String, i.e.:

```
g.V('{member_id=0, community_id=800066, ~label=comic, group_id=1}')
```

driver-dse/src/main/java/com/datastax/driver/graph/Edge.java
```diff
@@ -0,0 +1,137 @@
...
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+
+/**
+ * A default representation of an Edge in DSE Graph.
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonDeserialize(using = DefaultEdgeDeserializer.class)
+public class Edge {
```
tolbertam, line 28: Since edge and vertex share some characteristics (id, label, type, properties), should we have a parent class like gremlin has with [Element](http://tinkerpop.incubator.apache.org/javadocs/3.1.1-SNAPSHOT/full/org/apache/tinkerpop/gremlin/structure/Element.html)?  The other thing I wonder is are we going to end up duplicating a lot of the gremlin java API, should we just use the gremlin API proper, or is that bringing in too much (Would add gremlin as a dependency)?  (apologies if this has already been discussed).

driver-dse/src/main/java/com/datastax/driver/graph/GraphSession.java
```diff
@@ -0,0 +1,225 @@
...
+    /**
+     * Execute a graph statement.
+     *
+     * @param statement Any inherited class of {@link com.datastax.driver.graph.AbstractGraphStatement}. For now,
+     *                  this covers execution of simple and bound graph statements. It is extensible in case
+     *                  users would need to implement their own way {@link com.datastax.driver.graph.AbstractGraphStatement}.
+     * @return A {@link GraphResultSet} object if the execution of the query was successful. This method can throw an exception
+     * in case the execution did non complete successfully.
+     */
+    public GraphResultSet execute(AbstractGraphStatement statement) {
```
tolbertam, line 81: I think we need an interface above `AbstractGraphStatement` or we need to make it public since this is a public method, there could be weird classloading issues (BootstrapMethodError) if someone tries to use this as a method reference in java 8.

driver-dse/src/main/java/com/datastax/driver/graph/GraphStatement.java
tolbertam: Ah, nevermind, I just noticed that it's maintained so `processValues` doesn't create a new map every time its called.

driver-dse/src/main/java/com/datastax/driver/graph/GraphStatement.java
olim7t, line 36: :+1:

driver-dse/src/main/java/com/datastax/driver/graph/GraphStatement.java
olim7t: But `processValues` is only ever called when we execute the statement, so that can be done at the last moment (could even be done in the GraphSession code). I agree that this should not be an instance member.

driver-dse/src/main/java/com/datastax/driver/graph/AbstractGraphStatement.java
newkek: This was part of the changes to adapt for the 3.0 way of creating simple statements. I am correcting this today.

driver-dse/src/main/java/com/datastax/driver/graph/AbstractGraphStatement.java
newkek: Yes, it should be `this.payload.putAll(payload)`, I'll correct it.
Users should only be dealing with the `AbstractGraphStatement#payload`, I was thinking about renaming it to `AbstractGraphStatement#graphOptions` to make this distinction and take a `Map<String, String>` in this method to make it simpler as well.
Indeed they are not kept synced, `AbstractGraphStatement#payload` is applied to the `wrappedStatement`'s payload only at the statement execution time.

driver-dse/src/main/java/com/datastax/driver/graph/BoundGraphStatement.java
newkek, line 91: This was made to follow the [specs](https://github.com/riptano/bdp/blob/DSP-5175/dse-graph/docs/dev/server.adoc#dse-graph-server-extension), parameters are sent as unnamed parameters, each containing `{name:'parameter_name', value:JSON_serialised_value}`. 
I think this is related to the way the tinkerpop's gremlin server takes parameters, and since the DSE server will embed a gremlin-server, they chose this format to avoid modifying the embedded gremlin server.
And the order should not matter as the gremlin server will know how to decode the values since each name is specified in the value itself.

driver-dse/src/main/java/com/datastax/driver/graph/Edge.java
newkek, line 28: My goal initially was to avoid having a dependency on the gremlin-driver, but I think I am going to have to reconsider this..

driver-dse/src/main/java/com/datastax/driver/graph/GraphStatement.java
newkek, line 36: Ok, I thought that `SimpleGraphStatement` was a bit long for a simple statement..

driver-dse/src/main/java/com/datastax/driver/graph/GraphStatement.java
newkek, line 130: `'{member_id=0, community_id=800066, ~label=comic, group_id=1}'` is a Vertex **id**, so yes we could use `set("i", vertex.getId().toString())` from a previous result.
Considering that now I made ids a GraphData in Vertex and Edge.

driver-dse/src/main/java/com/datastax/driver/graph/GraphData.java
```diff
@@ -0,0 +1,215 @@
...
+        JsonNode jsN;
+        if (keyOrIndex instanceof Integer) {
+            jsN = this.jsonNode.get((Integer)keyOrIndex);
+        } else {
+            jsN = this.jsonNode.get((String)keyOrIndex);
+        }
+        return new GraphData(keyOrIndex, jsN, this.objectMapper);
+    }
+
+    public GraphData get(String key){
```
adutra: One should clarify that get(String) applies to JSON objects and get(int) to JSON arrays - calling the wrong one returns null (took me a while to figure out).

driver-dse/src/main/java/com/datastax/driver/graph/GraphData.java
```diff
@@ -0,0 +1,215 @@
...
+    public GraphData get(int index){
+        return get((Object)index);
+    }
+
+    /**
+     * Get the raw enclosed JSON object.
+     *
+     * @return The enclosed JSON object.
+     */
+    public JsonNode getJsonObject() {
```
adutra, line 94: If we are exposing Jackson API through our own API, I don't really see the point of wrapping a Jackson JsonNode object in a GraphData object.

driver-dse/src/main/java/com/datastax/driver/graph/GraphData.java
```diff
@@ -0,0 +1,215 @@
...
+/**
+ * A result entity containing a graph query's result, wrapping a Json result.
+ */
+public class GraphData {
+
+    private JsonNode jsonNode;
+
+    private ObjectMapper objectMapper;
+
+    private Object key;
```
adutra, line 30: This seems unused.

driver-dse/src/main/java/com/datastax/driver/graph/GraphStatement.java
tolbertam, line 130: I tried that and it does not work because `getId().toString()` returns the json payload (i.e.: `{member_id: "0", community_id: "800066", ~label: "comic", group_id: 
"1"}`) which gremlin doesn't like for some reason (maybe that is something that should work on the server side?)

driver-dse/src/main/java/com/datastax/driver/graph/GraphStatement.java
newkek: The goal here was to avoid regenerating the `List<String> jsonParams` each time `processValues()` is called.
By not making `jsonParams` an instance member, if we call `processValues()` twice in a row and the parameters (`valuesMap.hashCode()`) haven't changed, we could avoid regenerating the `List<String> jsonParams`. If `jsonParams` becomes a returned value, we would be forced to regenerate it each time because it wouldn't have been saved anywhere.

driver-dse/src/main/java/com/datastax/driver/graph/GraphData.java
newkek, line 94: I was thinking of leaving the possibility for more advanced users to do more advanced stuff we would not want to provide in the drivers because it would not fit most use cases, I don't have a clear example right now but I guess it could happen

newkek:
Latest update on this is : 
- Removed all new instantiations of a ObjectMapper, kept only one in GraphSession 9a50dd1
- Rebased on current 3.0. 3947753 : 
- GraphStatement constructor public.
- No Session instance in the AbstractGraphStatement and its child classes.
- Prepared and BoundGraphStatements structure improved and tested (unit tests to insure graph options are propagated and parameter values are bound correctly)
- Optimised imports and reformatted on default java scheme. e244966
- Various minor changes from reviews' comments. f8aea24

driver-dse/src/main/java/com/datastax/driver/graph/Edge.java
olim7t, line 28: Reusing classes from gremlin-driver definitely sounds interesting.

Their ResultSet class is probably too specific for us, but our ResultSet could be an iterable of [Result](http://tinkerpop.incubator.apache.org/javadocs/3.1.1-SNAPSHOT/full/org/apache/tinkerpop/gremlin/driver/Result.html). There's also a [GraphSONMapper](http://tinkerpop.incubator.apache.org/javadocs/3.1.1-SNAPSHOT/full/org/apache/tinkerpop/gremlin/structure/io/graphson/GraphSONMapper.html) that seems to provide out-of-the-box deserialization of classes like Vertex and Edge.

One hurdle is that gremlin-driver compiles to Java 8, so we'd have to raise the requirement on our dse module.

adutra, line 28: +1 on this.

> One hurdle is that gremlin-driver compiles to Java 8, so we'd have to raise the requirement on our dse module.

We already do this for driver-extras so it's technically easy; however that might put some DSE users out of the game.

adutra:
Closing because this code is being moved out of the java-driver repo.

PR #555, opened 2015-12-10
JAVA-833: Improve message when a nested type can't be serialized.

olim7t:
Throw an exception instead of propagating null back. I also changed the base message to avoid repeating the same thing twice when the wrong type is the root type.

This does not need to be merged in 3.0.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/TypeCodec.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/TypeCodec.java
@@ -159,7 +159,7 @@ abstract class TypeCodec<T> {
     }
 
     /* This is ugly, but not sure how we can do much better/faster
-     * Returns null if it's doesn't correspond to a known type.
+     * Throws IllegalArgumentException if it's doesn't correspond to a known type.
      *
      * Also, note that this only a dataType that is fit for the value,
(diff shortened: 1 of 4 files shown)


adutra:
:+1: Very elegant solution to get a proper error message for collections of unknown elements.

driver-core/src/main/java/com/datastax/driver/core/TypeCodec.java
```diff
@@ -159,7 +159,7 @@ static TupleCodec tupleOf(TupleType type) {
     }
 
     /* This is ugly, but not sure how we can do much better/faster
-     * Returns null if it's doesn't correspond to a known type.
+     * Throws IllegalArgumentException if it's doesn't correspond to a known type.
```
tolbertam, line 162: Would be good to use `@throws` annotation here since this is an unchecked exception.

olim7t, line 162: That's a regular comment though, not a javadoc (the method is package-private).

tolbertam, line 162: ah didn't notice that, in that case it definitely isn't needed then :)

PR #556, opened 2015-12-10
JAVA-890: fix mapper for case-sensitive UDT.

olim7t:


tolbertam:
looks good to me, :+1:

PR #557, opened 2015-12-10
Added CodecRegistry.codecFor(TypeToken)

antoinebaudoux:
Sometimes there is a need to know which CQL datatype is needed to store a given JAVA type, for example when generating Create Table statements to map a JAVA class to a CQL table.
Therefore I propose to add a `CodecRegistry.codecFor(TypeToken)` method.
Since there is already a `CodecRegistry.codecFor(T object)` method, it seems logical to have a method that takes a Java Type in parameter also.
Thanks, 
Antoine

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/CodecRegistry.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/CodecRegistry.java
@@ -34,6 +34,7 @@ import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutionException;
 
 import static com.datastax.driver.core.DataType.Name.*;
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
@@ -404,6 +405,27 @@ public final class CodecRegistry {
     }
 
     /**
+     * Returns a {@link TypeCodec codec} that accepts the given Java type.
+     * <p/>
+     * This method takes an arbitrary {@link TypeToken Java type} and tries to locate a suitable codec for it.
+     * This method returns the first matching codec, regardless of its accepted CQL type.
+     * It should be reserved for situations where the target CQL type is not available or unknown.
+     * This method can be used, for example, to determine the {@link DataType CQL type} corresponding to a
+     * particular {@link TypeToken Java type}, in order to generate a corresponding {@link com.datastax.driver.core.schemabuilder.Create}
+     * statement.
+     * <p/>
+     * Codecs returned by this method are <em>NOT</em> cached (see the {@link CodecRegistry top-level documentation}
+     * of this class for more explanations about caching).
+     *
+     * @param javaType The {@link TypeToken Java type} the codec should accept; must not be {@code null}.
+     * @return A suitable codec.
+     * @throws CodecNotFoundException if a suitable codec cannot be found.
+     */
+    public <T> TypeCodec<T> codecFor(TypeToken<T> javaType) {
+        return findCodec(null, javaType);
+    }
+
+    /**
      * Returns a {@link TypeCodec codec} that accepts the given {@link DataType CQL type}.
      * <p/>
      * This method returns the first matching codec, regardless of its accepted Java type.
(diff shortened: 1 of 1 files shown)


adutra:
Thanks for your proposal. [Something similar](https://github.com/datastax/java-driver/blob/618a988fd0ddd87f7fb27c379fd9e63212527995/driver-core/src/main/java/com/datastax/driver/core/CodecRegistry.java) has been proposed in JAVA-855, but the ticket eventually got canceled.
If I understand correctly, you have a top-down approach where you want to generate a CQL schema from a Java model, and you need a way to map your Java types to CQL types, is that accurate?

antoinebaudoux:
Thanks for your reply. Yes, we would like to generate the CQL tables from Java classes. We would register all the necessary TypeCodec in the CodecRegistry, then either ask the CodecRegistry which TypeCodec should handle a given Java class, or manually specify a custom TypeCodec if needed.

driver-core/src/main/java/com/datastax/driver/core/CodecRegistry.java
```diff
@@ -404,6 +405,27 @@ public CodecRegistry register(Iterable<? extends TypeCodec<?>> codecs) {
...
+     * statement.
+     * <p/>
+     * Codecs returned by this method are <em>NOT</em> cached (see the {@link CodecRegistry top-level documentation}
+     * of this class for more explanations about caching).
+     *
+     * @param javaType The {@link TypeToken Java type} the codec should accept; must not be {@code null}.
+     * @return A suitable codec.
+     * @throws CodecNotFoundException if a suitable codec cannot be found.
+     */
+    public <T> TypeCodec<T> codecFor(TypeToken<T> javaType) {
```
adutra, line 424: A method `codecFor(Class<T> javaClass)` should also be proposed.

driver-core/src/main/java/com/datastax/driver/core/CodecRegistry.java
```diff
@@ -504,14 +526,17 @@ public CodecRegistry register(Iterable<? extends TypeCodec<?>> codecs) {
...
+        checkArgument(cqlType != null || javaType != null, "Parameter cqlType and javaType cannot be both null");
+        logger.trace("Looking for codec [{} <-> {}]", cqlType == null? "ANY" : cqlType, javaType == null ? "ANY" : javaType);
         for (TypeCodec<?> codec : codecs) {
-            if (codec.accepts(cqlType) && (javaType == null || codec.accepts(javaType))) {
+            if ((cqlType == null || codec.accepts(cqlType)) && (javaType == null || codec.accepts(javaType))) {
                 logger.trace("Codec found: {}", codec);
                 return (TypeCodec<T>) codec;
             }
         }
+        if (cqlType == null) {
```
adutra, line 537: You are returning too soon; this will prevent the registry to create a codec for, say, `[list<int> <-> List<Foo>]` given an existing codec `[int <-> Foo]`.

driver-core/src/main/java/com/datastax/driver/core/CodecRegistry.java
```diff
@@ -404,6 +405,27 @@ public CodecRegistry register(Iterable<? extends TypeCodec<?>> codecs) {
...
+     * <p/>
+     * Codecs returned by this method are <em>NOT</em> cached (see the {@link CodecRegistry top-level documentation}
+     * of this class for more explanations about caching).
+     *
+     * @param javaType The {@link TypeToken Java type} the codec should accept; must not be {@code null}.
+     * @return A suitable codec.
+     * @throws CodecNotFoundException if a suitable codec cannot be found.
+     */
+    public <T> TypeCodec<T> codecFor(TypeToken<T> javaType) {
+        return findCodec(null, javaType);
```
adutra, line 425: You should cache results by calling `lookupCodec` instead, although you would have to modify `TypeCodecWeigher` to accept null CQL types.

driver-core/src/main/java/com/datastax/driver/core/CodecRegistry.java
antoinebaudoux, line 537: Okay, you prefer that I adapt `maybeCreateCodec(DataType cqlType, TypeToken<T> javaType)` to handle the case where `cqlType==null` or I create a new similar method `maybeCreateCodec(TypeToken<T> javaType)` where `cqlType`is implicitly unspecified?

adutra:
Before you go any further, I will review your use case with the driver team. This seems a valid use case, but a bit particular; we need to weigh the pros and cons of adding methods to CodecRegistry. In any case, I feel that if we were to add this new method, the implementation would have to be identical to what has been proposed in JAVA-855.

driver-core/src/main/java/com/datastax/driver/core/CodecRegistry.java
antoinebaudoux, line 425: What about 

``` java
@Override
        public int weigh(CacheKey key, TypeCodec<?> value) {
            return codecs.contains(value) ? 0 : weigh(value.getCqlType(), 0);
        }
```

since the cqlType of the TypeCodec itself is never null?

antoinebaudoux:
Okay. Thanks for your time.

driver-core/src/main/java/com/datastax/driver/core/CodecRegistry.java
adutra, line 425: Very clever idea, I didn't think of it :)

adutra:
@ab-taktik Thank you very much for bringing this use case to us, I created [JAVA-1015](https://datastax-oss.atlassian.net/browse/JAVA-1015) to track progress on this. I am however closing this pull request as we wanted to seize this opportunity to improve test coverage of `CodecRegistry` and perform a little cleanup in `TypeCodec`. I opened #559 for it.

antoinebaudoux:
@adutra Thanks for your quick reply on this, I will follow the progress of this. In the meantime I've created my own registry where I define my Java Class -> TypeCodec mapping. Its a bit redundant with CodecRegistry but it does its work.

PR #559, opened 2015-12-14
JAVA-1015: Support for codec lookups by Java type only in CodecRegistry

adutra:
This PR also introduces two improvements in `TypeCodec`:
- A new overloaded method `accepts(Class<?> javaType)`: the motivation behind it is that oftentimes caller code calls by mistake e.g. `accepts(Integer.class)` which will actually resolve to `accepts(Object value)` leading to unexpected results.
- Removal of field `primitiveToWrapperMap`, replaced with `TypeToken.wrap()`.

This PR also improves test coverage for `CodecRegistry`.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -2,6 +2,7 @@
 
 ### 3.0.0-rc1 (in progress)
 
+- [new feature]: JAVA-1015: Support for codec lookups by Java type only in CodecRegistry.
 
 ### 3.0.0-beta1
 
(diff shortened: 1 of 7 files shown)


adutra:
We decided not to introduce the new methods in CodecRegistry so I'm closing this PR. Sorry @ab-taktik :(

antoinebaudoux:
Okay @adutra , thanks a lot for the feedback, much appreciated.

What do you propose as an alternative to implement the top-down approach?

adutra:
Well, that was the main reason why we decided not to fix JAVA-1015: the driver cannot itself create a "bullet-proof" mapping between Java classes and CQL types. E.g. one could register two codecs for the same Java type, and in this case, only the first one would be returned by the registry, but the data model might actually need the second one. So I guess it is up to third-party applications to implement this mapping in the way that best suits their needs.

PR #560, opened 2015-12-14
Retry Policy Test Enhancements 3.0

tolbertam:
This PR does the following:
- Upgrade to scassandra 0.11.0
- Add tests for connection errors (connection being closed during request) for Retry Policy tests.
- Refactor DowngradingConsistencyRetryPolicy to use Scassandra and add extended retry policy tests to be in line with the other RetryPolicyTests.

This will also need to be done on 2.1.   Will do so directly on #538 if/when this PR is approved.

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/policies/DefaultRetryPolicyIntegrationTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/policies/DefaultRetryPolicyIntegrationTest.java
@@ -19,8 +19,8 @@ import com.datastax.driver.core.*;
 import com.datastax.driver.core.exceptions.*;
 import org.assertj.core.api.Fail;
 import org.scassandra.Scassandra;
+import org.scassandra.http.client.ClosedConnectionConfig;
 import org.scassandra.http.client.PrimingRequest;
-import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
 
 import java.util.Collections;
@@ -174,7 +174,7 @@ public class DefaultRetryPolicyIntegrationTest extends AbstractRetryPolicyIntegr
                 query();
                 fail("expected a NoHostAvailableException");
             } catch (NoHostAvailableException e) {
-                assertThat(e.getErrors().keySet()).hasSize(3).containsOnlyOnce(
+                assertThat(e.getErrors().keySet()).hasSize(3).containsOnly(
                         host1.getSocketAddress(),
                         host2.getSocketAddress(),
                         host3.getSocketAddress());
@@ -199,13 +199,6 @@ public class DefaultRetryPolicyIntegrationTest extends AbstractRetryPolicyIntegr
         }
     }
 
-    @DataProvider
-    public static Object[][] serverSideErrors() {
-        return new Object[][]{
-                {server_error, ServerError.class},
-                {overloaded, OverloadedException.class}
-        };
-    }
 
     @Test(groups = "short", dataProvider = "serverSideErrors")
     public void should_try_next_host_on_server_side_error(PrimingRequest.Result error, Class<? extends DriverException> exception) {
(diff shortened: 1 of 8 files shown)


driver-core/src/test/java/com/datastax/driver/core/policies/DefaultRetryPolicyIntegrationTest.java
```diff
@@ -231,4 +224,29 @@ public void should_try_next_host_on_server_side_error(PrimingRequest.Result erro
...
 
+
+    @Test(groups = "short", dataProvider = "connectionErrors")
+    public void should_try_next_host_on_connection_error(ClosedConnectionConfig.CloseType closeType) {
+        simulateError(1, PrimingRequest.Result.closed_connection, new ClosedConnectionConfig(closeType));
+        simulateError(2, PrimingRequest.Result.closed_connection, new ClosedConnectionConfig(closeType));
+        simulateError(3, PrimingRequest.Result.closed_connection, new ClosedConnectionConfig(closeType));
+        try {
+            query();
+            Fail.fail("expected a TransportException");
```
adutra: Microscopic nit: this should read "expected a NoHostAvailableException".

adutra:
:100: finally we can test timeouts, very good job!

driver-core/src/test/java/com/datastax/driver/core/policies/DefaultRetryPolicyIntegrationTest.java
```diff
@@ -231,4 +224,29 @@ public void should_try_next_host_on_server_side_error(PrimingRequest.Result erro
...
+        } catch (NoHostAvailableException e) {
+            assertThat(e.getErrors().keySet()).hasSize(3).containsOnlyOnce(
```
olim7t: `containsOnly` would be more appropriate since this is a set.

tolbertam:
Thanks!  I incorporated the feedback, will squash and merge if this looks ok.

driver-core/src/test/java/com/datastax/driver/core/policies/DowngradingConsistencyRetryPolicyIntegrationTest.java
```diff
@@ -15,90 +15,338 @@
...
-            ccm.waitForDown(2);
-            // One replica remaining: should downgrade to 1 when CL > 1
-            checkAchievedConsistency(ALL, ONE, session);
-            checkAchievedConsistency(QUORUM, ONE, session);
-            checkAchievedConsistency(TWO, ONE, session);
-            checkAchievedConsistency(ONE, ONE, session);
+            query();
+            fail("expected an ReadTimeoutException");
+        } catch (ReadTimeoutException e) {
+            assertThat(e.getConsistencyLevel()).isEqualTo(expectedCL);
```
olim7t: How come this matches? Is it because you passed `received + 1` when priming, and SCassandra uses that to determine the CL it returns in the response?

olim7t:
Could you format the code?

tolbertam:
Ah, yes forgot completely about that, i'll do that now.

driver-core/src/test/java/com/datastax/driver/core/policies/DowngradingConsistencyRetryPolicyIntegrationTest.java
tolbertam: It's really confusing so i'll add some docs, but Scassandra ignores the CL that you use, it just always returns the error you tell it to so what actually happens is:

For 4 alive, but 5 required:

Query is tried -> Unavailable exception with 4 alive replicas, 5 required.
Query is tried again with CL 3 -> Unavailable exception with 4 alive replicas, 5 required.
UnavailableException is raised because we've already tried once.

So all the test is really trying to do is validate a retry is attempted with the correct consistency level based on the number of alive hosts.

tolbertam: Clarified in docs why we expect a cl.

tolbertam:
On second thought, going to wait on formatting until I get in sync with you.  For some reason with the default formatters it changes quite a bit in code I haven't changed which doesn't seem right.

driver-core/src/test/java/com/datastax/driver/core/policies/DowngradingConsistencyRetryPolicyIntegrationTest.java
olim7t: Your changes do make it clearer, thanks.

PR #561, opened 2015-12-15
JAVA-1014 New DseAuthProvider implementation for DSE 5.0

mike-tr-adamson:


driver-dse/src/main/java/com/datastax/driver/auth/DseAuthProvider.java
```diff
@@ -65,20 +91,153 @@
...
- *                          .build();
- * </pre>
  */
 public class DseAuthProvider implements AuthProvider {
-    @Override
+    private static final String DSE_AUTHENTICATOR = "com.datastax.bdp.cassandra.auth.DseAuthenticator";
+    private final String username;
+    private final String password;
+
+    public DseAuthProvider() {
```
adutra: It might be useful to users to provide javadocs for both constructors, specially because using one or the other induces the authentication mechanism used behind the scenes.
Or alternatively, one could make this class abstract and provide two small subclasses, `PlainTextDseAuthProvider` and `GSSAPIDseAuthProvider`, to make things even clearer for users.

driver-dse/src/main/java/com/datastax/driver/auth/DseAuthProvider.java
```diff
@@ -65,20 +91,153 @@
...
     public Authenticator newAuthenticator(InetSocketAddress host, String authenticator) throws AuthenticationException {
-        return new KerberosAuthenticator(host);
+        // Need both username and password for PLAIN auth
+        if ((username != null) && (password != null)) {
+            return new PlainTextAuthenticator(authenticator, username, password);
+        }
+        return new GSSAPIAuthenticator(authenticator, host);
+    }
+
+    public static abstract class BaseDseAuthenticator implements Authenticator {
```
adutra: This class could be private.

driver-dse/src/main/java/com/datastax/driver/auth/DseAuthProvider.java
```diff
@@ -65,20 +91,153 @@
...
+            throw new RuntimeException("Incorrect challenge from server");
+        }
+    }
+
+    private static class GSSAPIAuthenticator extends BaseDseAuthenticator {
+        private static final String JAAS_CONFIG_ENTRY = "DseClient";
+        private static final String[] SUPPORTED_MECHANISMS = new String[]{"GSSAPI"};
+        private static final String SASL_PROTOCOL_NAME = "dse";
+        private static final String SASL_PROTOCOL_NAME_PROPERTY = "dse.sasl.protocol";
+        private static final Map<String, String> DEFAULT_PROPERTIES = new HashMap<String, String>()
```
adutra: This does not look thread-safe, maybe replace with `ImmutableMap.<String, String>builder().put(Sasl.SERVER_AUTH, "true").put(Sasl.QOP, "auth").build()`?

driver-dse/src/main/java/com/datastax/driver/auth/DseAuthProvider.java
mike-tr-adamson: This was left public so that we can extend the `DseAuthProvider` and provide additional `Authenticator` implementations

driver-dse/src/main/java/com/datastax/driver/auth/DseAuthProvider.java
mike-tr-adamson: I like your second idea much bettter. That would make it very clear what was being selected. If it's ok I'd prefer `DsePlainTextAuthProvider` and `DseGSSAPIAuthProvider` but if you'd prefer your versions I'll stick with them.

mike-tr-adamson:
New commit with review changes

driver-dse/src/main/java/com/datastax/driver/auth/DseAuthProvider.java
```diff
@@ -18,67 +18,176 @@
...
+    private static final String DSE_AUTHENTICATOR = "com.datastax.bdp.cassandra.auth.DseAuthenticator";
+    private final String username;
+    private final String password;
+
+    public DseAuthProvider(String username, String password) {
+        this.username = username;
+        this.password = password;
+    }
+
     public Authenticator newAuthenticator(InetSocketAddress host, String authenticator) throws AuthenticationException {
```
adutra: I'd rather leave this method for subclasses to implement. Imho DseAuthProvider shouldn't know anything about username and passwords (they only make sense for plain text mechanism).
The way I would see it, this class would contain just the following:

``` java
public abstract class DseAuthProvider implements AuthProvider {

    private static final String DSE_AUTHENTICATOR = "com.datastax.bdp.cassandra.auth.DseAuthenticator";

    public static abstract class BaseDseAuthenticator implements Authenticator {

        private final String authenticator;

        public BaseDseAuthenticator(String authenticator) {
            this.authenticator = authenticator;
        }

        public abstract byte[] getMechanism();

        public abstract byte[] getInitialServerChallenge();

        @Override
        public byte[] initialResponse() {
            if (isDseAuthenticator())
                return getMechanism();
            else
                return evaluateChallenge(getInitialServerChallenge());
        }

        @Override
        public void onAuthenticationSuccess(byte[] token) {
        }

        private boolean isDseAuthenticator() {
            return authenticator.equals(DSE_AUTHENTICATOR);
        }
    }

}
```

And subclasses would look like this:

``` java
public class DseGSSAPIAuthProvider extends DseAuthProvider {

    @Override
    public Authenticator newAuthenticator(InetSocketAddress host, String authenticator) throws AuthenticationException {
        return new GSSAPIAuthenticator(authenticator, host);
    }

    private static class GSSAPIAuthenticator extends BaseDseAuthenticator {
        ...
    }
}

public class DsePlainTextAuthProvider extends DseAuthProvider {

    private final String username;

    private final String password;

    public DsePlainTextAuthProvider(String username, String password) {
        checkNotNull(username, "username cannot be null");
        checkNotNull(password, "password cannot be null");
        this.username = username;
        this.password = password;
    }

    @Override
    public Authenticator newAuthenticator(InetSocketAddress host, String authenticator) throws AuthenticationException {
        return new PlainTextAuthenticator(authenticator, username, password);
    }

    private static class PlainTextAuthenticator extends DseAuthProvider.BaseDseAuthenticator {
        ....
    }
}
```

driver-dse/src/main/java/com/datastax/driver/auth/DseAuthProvider.java
adutra: OK but in this case it will be part of the driver's API, and we should at least document it. Could you complete the javadocs for the class and the public methods (`getMechanism` and `getInitialServerChallenge`)?
And in that case, unless I'm missing something I think we could keep the constructor and the methods I mentioned protected instead of public (we're a bit reluctant about making things public because once you committed to the API, there is no turning back).

driver-dse/src/main/java/com/datastax/driver/auth/DseAuthProvider.java
adutra: Now that I think of it, `BaseDseAuthenticator` could be a top-level class; I don't think we actually need an abstract `DseAuthProvider` at all (the static field `DSE_AUTHENTICATOR` would be better declared in `BaseDseAuthenticator` where it is accessed).

adutra:
Thanks @mike-tr-adamson for addressing my comments! I've made a few more, and I also have one general question: Is this new implementation backwards-compatible with older versions of DSE? IOW. if the authenticator is not DSE_AUTHENTICATOR, is its behavior the same as before (i.e. trigger a Kerberos authentication)?

driver-dse/src/main/java/com/datastax/driver/auth/DseAuthProvider.java
adutra: Sorry for the spam :) but since we are modifying the API there should also be an entry in `upgrade_guide/README.md` to explain how to migrate existing code, can you add that too? I guess it boils down to replacing `DseAuthProvider` with `DseGSSAPIAuthProvider`. Thanks!

mike-tr-adamson:
@adutra I agree with you latest comments about break the AuthProviders apart completely but how do you feel about keeping `DseAuthProvider` as a simple extension of `DseGSSAPIAuthProvider` and mark it as deprecated. That would make migration to the new AuthProviders less painful as the `DseAuthProvider` would still provide kerberos auth without any changes.

As to you question about backward compatibility. The new AuthProviders are fully compatible with older versions of DSE. That is what the authenticator check is for. For any authenticators other than `DseAuthenticator` the behaviour reverts to the older behaviour (I hope that makes sense).

adutra:
> That would make migration to the new AuthProviders less painful as the DseAuthProvider would still provide kerberos auth without any changes.

+100 on this! It is always best when users can migrate without having to fix compilation errors.

> For any authenticators other than DseAuthenticator the behaviour reverts to the older behaviour (I hope that makes sense).

Thanks for the clarification, makes perfectly sense.

olim7t:
> Now that I think of it, BaseDseAuthenticator could be a top-level class; I don't think we actually need an abstract DseAuthProvider at all (the static field DSE_AUTHENTICATOR would be better declared in BaseDseAuthenticator where it is accessed).

+1 on that. `BaseDseAuthenticator` would be top-level and package-private, and we could move each implementation as an inner class of the corresponding AuthProvider (`PlainTextAuthenticator` in `DsePlainTextAuthProvider`, etc.). Then we need to override `newAuthenticator` in each AuthProvider but the implementation is trivial.

> keeping DseAuthProvider as a simple extension of DseGSSAPIAuthProvider and mark it as deprecated

Also agree with that.

mike-tr-adamson:
@adutra The latest commit should address all your review comments. `DseAuthProvider` is now a simple extension of `DseGSSAPIAuthProvider`.

olim7t:
Latest commit looks great!
Could you format the code and squash everything together before we merge? We use the default IntelliJ IDEA settings for formatting, if you don't use it I can send you a commit with the formatting.

mike-tr-adamson:
@olim7t I don't use intellij. If you could send me a commit, I'll update and squash my branch.

mike-tr-adamson:
@adutra @olim7t Formatted and squashed

PR #562, opened 2015-12-16
Fixed typos dealing with by-index get/set in Row objects etc.

jbisso:
Made corrections per email. Might need to make changes in 2.2 also, because that is the branch that has API docs currently generated and posted from.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Row.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Row.java
@@ -20,8 +20,8 @@ import com.datastax.driver.core.exceptions.InvalidTypeException;
 /**
  * A CQL Row returned in a {@link ResultSet}.
  * <p/>
- * The values of a CQL Row can be retrieve by either index (index starts at 0)
- * or name. When setting them by name, names follow the case insensitivity
+ * The values of a CQL Row can be retrieved by either index (index starts at zero)
+ * or name. When getting them by name, names follow the case insensitivity
  * rules explained in {@link ColumnDefinitions}.
  */
 public interface Row extends GettableData {
--- a/driver-core/src/main/java/com/datastax/driver/core/SettableByIndexData.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/SettableByIndexData.java
@@ -26,7 +26,7 @@ import java.nio.ByteBuffer;
 import java.util.*;
 
 /**
- * Collection of (typed) CQL values that can set by index (starting a 0).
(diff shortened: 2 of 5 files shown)


driver-core/src/main/java/com/datastax/driver/core/Row.java
```diff
@@ -20,8 +20,8 @@
 /**
  * A CQL Row returned in a {@link ResultSet}.
  * <p/>
- * The values of a CQL Row can be retrieve by either index (index starts at 0)
- * or name. When setting them by name, names follow the case insensitivity
+ * The values of a CQL Row can be retrieve by either index (index starts at zero)
+ * or name. When getting or setting them by name, names follow the case insensitivity
```
adutra: This really should be "When getting them" only, as it is not possible to set anything in a Row.

driver-core/src/main/java/com/datastax/driver/core/Row.java
```diff
@@ -20,8 +20,8 @@
 /**
  * A CQL Row returned in a {@link ResultSet}.
  * <p/>
- * The values of a CQL Row can be retrieve by either index (index starts at 0)
- * or name. When setting them by name, names follow the case insensitivity
+ * The values of a CQL Row can be retrieve by either index (index starts at zero)
```
adutra: Another typo: "can be retrieve**d**"

driver-core/src/main/java/com/datastax/driver/core/SettableByIndexData.java
```diff
@@ -26,7 +26,7 @@
 import java.util.*;
 
 /**
- * Collection of (typed) CQL values that can set by index (starting a 0).
+ * Collection of (typed) CQL values that can set by index (starting at zero).
```
adutra: "that can **be** set"

driver-core/src/main/java/com/datastax/driver/core/SettableData.java
```diff
@@ -16,7 +16,7 @@
 package com.datastax.driver.core;
 
 /**
- * Collection of (typed) CQL values that can set either by index (starting a 0) or by name.
+ * Collection of (typed) CQL values that can set either by index (starting at zero) or by name.
```
adutra: "that can **be** set" again – I suppose we need to amend SettableByNameData as well.

jbisso:
Second try.

jbisso:
One more time.

olim7t:
:+1: LGTM

olim7t:
Could you squash everything into one commit ?

PR #564, opened 2015-12-18
Add support for using DataStax Enterprise with CCMBridge.

tolbertam:
Adds support for running integration tests with DataStax Enterprise.

To run tests with DSE, set the system property 'dse.version' to the version of DSE you want to test with.  Note that you will still need to pass in a 'cassandra.version' property to indicate which major version of cassandra is associated with this DSE version (as this affects behavior of some tests).

Rather than adding system properties for DSE credentials, this relies on a [recent change](pcmanus/ccm#https://github.com/pcmanus/ccm/pull/426) in CCM to support providing credentials via $HOME/.ccm/.dse.ini.  The contents of this file needs to be formed in this way:

``` ini
[dse_credentials]
dse_username = myusername
dse_password = mypassword
```

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
@@ -51,9 +51,11 @@ public class CCMBridge {
 
     public static final String IP_PREFIX;
 
-    static final String CASSANDRA_VERSION;
+    private static final String CASSANDRA_VERSION;
 
-    static final String CASSANDRA_INSTALL_ARGS;
+    private static final String CASSANDRA_INSTALL_ARGS;
+
+    private static final boolean IS_DSE;
 
     public static final String DEFAULT_CLIENT_TRUSTSTORE_PASSWORD = "cassandra1sfun";
     public static final String DEFAULT_CLIENT_TRUSTSTORE_PATH = "/client.truststore";
@@ -86,6 +88,49 @@ public class CCMBridge {
     private static final Map<String, String> ENVIRONMENT_MAP;
 
     /**
+     * A mapping of full DSE versions to their C* counterpart.  This is not meant to be comprehensive.  Used by
+     * {@link #getCassandraVersion()}.  If C* version cannot be derived, the method makes a 'best guess'.
+     */
+    private static final Map<String, String> dseToCassandraVersions = ImmutableMap.<String, String>builder()
+            .put("5.0", "3.0")
+            .put("4.8.3", "2.1.11")
+            .put("4.8.2", "2.1.11")
+            .put("4.8.1", "2.1.11")
+            .put("4.8", "2.1.9")
+            .put("4.7.6", "2.1.11")
+            .put("4.7.5", "2.1.11")
+            .put("4.7.4", "2.1.11")
+            .put("4.7.3", "2.1.8")
+            .put("4.7.2", "2.1.8")
+            .put("4.7.1", "2.1.5")
+            .put("4.6.11", "2.0.16")
+            .put("4.6.10", "2.0.16")
+            .put("4.6.9", "2.0.16")
+            .put("4.6.8", "2.0.16")
+            .put("4.6.7", "2.0.14")
+            .put("4.6.6", "2.0.14")
+            .put("4.6.5", "2.0.14")
+            .put("4.6.4", "2.0.14")
+            .put("4.6.3", "2.0.12")
+            .put("4.6.2", "2.0.12")
+            .put("4.6.1", "2.0.12")
+            .put("4.6", "2.0.11")
+            .put("4.5.9", "2.0.16")
+            .put("4.5.8", "2.0.14")
+            .put("4.5.7", "2.0.12")
+            .put("4.5.6", "2.0.12")
+            .put("4.5.5", "2.0.12")
+            .put("4.5.4", "2.0.11")
+            .put("4.5.3", "2.0.11")
+            .put("4.5.2", "2.0.10")
+            .put("4.5.1", "2.0.8")
+            .put("4.5", "2.0.8")
+            .put("4.0", "2.0")
+            .put("3.2", "1.2")
+            .put("3.1", "1.2")
+            .build();
+
+    /**
      * The command to use to launch CCM
      */
     private static final String CCM_COMMAND;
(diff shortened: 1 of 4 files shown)


tolbertam:
Moving this to work in progress as I think it would also be useful to have a means decorate tests as DSE specific so they are only ran if a DSE version is provided.

adutra:
Is it possible to launch a DSE instance from a git branch or an install directory? If so, I would suggest:
- Re-use the `cassandra.*` properties for DSE version/branch/dir
- Add a boolean system property `dse` that triggers the inclusion of `--dse` in the CCM command line if its value is `true`

tolbertam:
> Is it possible to launch a DSE instance from a git branch or an install directory?

It does indeed work from install directory using 'cassandra.directory' system property granted it was previously built (same behavior as oss cassandra).   Git branch would not work as of yet as CCM does not support this, but if it did, it would work via cassandra.branch.

> If so, I would suggest:
> - Re-use the cassandra.\* properties for DSE version/branch/dir
> - Add a boolean system property dse that triggers the inclusion of --dse in the CCM command line if its value is true.

It should work this way for both cases, the code sets an isDSE boolean which is used for determining whether or not '--dse' needs to be appended to args when adding/creating a node (only 2 CCM commands that seem to be depend on this that i've seen so far).

tolbertam:
After speaking with @adutra about this, I understand and agree with his points.   I will overload `-Dcassandra.version` to support handling the DSE version and add a `-Ddse` parameter such that if true will interpret that as a DSE release.  In addition, will add a mapping of DSE <-> C\* versions to determine test compatibility.   I've also added a `@DseVersion` annotation that will only run tests if the version criteria is met and DSE is in use (otherwise the tests are skipped).

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
```diff
@@ -685,7 +728,7 @@ public Builder withCassandraConfiguration(String key, String value) {
         }
 
         public CCMBridge build() {
-            CCMBridge ccm = new CCMBridge();
+            CCMBridge ccm = new CCMBridge(isDSE);
             ccm.execute(buildCreateCommand());
             ccm.updateConfig(cassandraConfiguration);
```
tolbertam: Need to add a `withDSEConfiguration` method and call `updateDSEConfig` as appropriate.

tolbertam:
Updated to use '-Dcassandra.version' for both C\* and DSE versions.  If '-Ddse' is provided, it is assumed a DSE version is passed.  A mapping for determining C\* version from DSE version is described in CCMBridge.

Example usages:

**DSE 4.8.3**:

`-Ddse -Dcassandra.version=4.8.3`
`-Ddse=true -Dcassandra.version=4.8.3`

Emits:

`[CCMBridge] Using DSE version: 4.8.3 (C* 2.1.11, install arguments: -v4.8.3 --dse)`

**Cassandra 1.2.19**:

`-Dcassandra.version=1.2.19`

Emits:

`[CCMBridge] Using Cassandra version: 1.2.19 (install arguments: -v1.2.19)`

**Custom local install of DSE 5.0**

`-Dcassandra.version=5.0 -Ddse -Dcassandra.directory=/Users/atolbert/Documents/Projects/dse`

Emits:

`[CCMBridge] Using DSE version: 5.0 (C* 3.0, install arguments: --install-dir=/Users/atolbert/Documents/Projects/dse --dse)`

Also added `@DseVersion` annotation which validates against DSE version.  If DSE not in use, skips test.   Also updated version checks to allow partial versions (i.e. C\* 3.0) which is treated ad Major.MAX_VALUE for cases where we can't determine the exact C\* version from the DSE version.

tolbertam:
Since switching to the docker-based infra for travis we often have `should_stop_if_cancelled_during_failed_reconnect(com.datastax.driver.core.AbstractReconnectionHandlerTest` fail.  This test is pretty sensitive/timing based, I'll investigate in a separate PR.

tolbertam:
I think this is ready to go, removing WIP tag.  If it looks ok I'll rebase before merging.

driver-core/src/test/java/com/datastax/driver/core/TestListener.java
```diff
@@ -28,7 +29,13 @@
     private int test_index = 0;
 
     static {
-        System.out.println("[CCMBridge] Using Cassandra version: " + CCMBridge.CASSANDRA_VERSION + " (install arguments: " + CCMBridge.CASSANDRA_INSTALL_ARGS + ")");
+        if (CCMBridge.isDSE()) {
+            System.out.println("[CCMBridge] Using DSE version: " + CCMBridge.getDSEVersion() + " (C* " +
+                    CCMBridge.getCassandraVersion() + ", install arguments: " + CCMBridge.getInstallArguments() + ")");
+        } else {
+            System.out.println("[CCMBridge] Using Cassandra version: " + CCMBridge.getCassandraVersion() +
```
adutra: Minor nit: I always considered this message a bit misleading. It makes believe that a CCM cluster is being launched at that moment, which is not true; besides, this message is printed even when running only unit tests. Maybe we could rephrase it "[CCMBridge] CCM tests for this test session will use ...."

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
```diff
@@ -147,8 +208,68 @@ private static boolean isWindows() {
...
+    public static String getCassandraVersion() {
+        if (isDSE()) {
+            String cassandraVersion = dseToCassandraVersions.get(CASSANDRA_VERSION);
+            if (cassandraVersion != null) {
+                return cassandraVersion;
+            } else if (CASSANDRA_VERSION.startsWith("3.") || CASSANDRA_VERSION.compareTo("3") <= 0) {
+                return "1.2";
+            } else if (CASSANDRA_VERSION.startsWith("4.")) {
+                if (CASSANDRA_VERSION.compareTo("4.7") >= 0) {
+                    return "2.0";
```
adutra: Aren't the if and else blocks inverted?

tolbertam: Good catch, they are!  Fixing.

driver-core/src/test/java/com/datastax/driver/core/TestListener.java
tolbertam: :+1: on that idea, i'll update this.

tolbertam: Updated and changed this to use a logger instead of println.  I also moved this block to CCMBridge as that seems like a better home for it.  It will also defer parsing to when CCMBridge is first referenced instead of in TestListener (which may have some negative impacts, will observe test results).  output now looks like:

**DSE**

```
    0      [main] INFO  com.datastax.driver.core.CCMBridge  - Tests requiring CCM will use DSE version 4.8.3 (C* 2.1.11, install arguments: --install-dir=/Users/atolbert/Documents/Projects/dse-4.8.3 --dse)
```

**Cassandra**

```
    0      [main] INFO  com.datastax.driver.core.CCMBridge  - Tests requiring CCM will use Cassandra version 2.1.12 (install arguments: -v2.1.12)
```

tolbertam:
commits have been squashed into 1 (no other changes).

adutra:
Squashed again on top of current 2.1 (no changes).

adutra:
Looks very very good so far; I have a few more general remarks:
- We should document somewhere in the code (CCMBridge class?) that when using DSE one needs to create the file `$HOME/.ccm/.dse.ini` with proper credentials.
- I was wondering if `cassandra.version` shouldn't be renamed to `ccm.version` or stg similar to denote that it can be either a C\* version or a DSE version.

olim7t:
This is awesome.
I'm fine leaving the name as `cassandra.version`.
As for documentation, I think we should do that in the `CONTRIBUTING.md` of the module that will run DSE tests.

PR #566, opened 2015-12-30
Refactor CCM tests.

adutra:


driver-core/src/test/java/com/datastax/driver/core/CCMTestsSupport.java
```diff
@@ -0,0 +1,687 @@
...
+            }
+        }
+
+    }
+
+    /**
+     * A LoadingCache that stores running CCM clusters.
+     */
+    private static final LoadingCache<CCMTestContextKey, CCMTestContext> CACHE = CacheBuilder.newBuilder()
+            .initialCapacity(5)
```
tolbertam, line 379: I like this idea but I think we should reduce the maximum size to something smaller (somewhere between 2-5?).  How many instances we'd want to run probably depends on the system it is on, maybe we should set the maximum size to # of processors or something like that?

driver-core/src/test/java/com/datastax/driver/core/CCMTestsSupport.java
```diff
@@ -0,0 +1,687 @@
...
+     * Returns an alternate cluster builder to use for this test,
+     * with no event debouncing.
+     * <p/>
+     * It's not required to call {@link com.datastax.driver.core.Cluster.Builder#addContactPointsWithPorts},
+     * it will be done automatically.
+     *
+     * @return The cluster builder to use for the tests.
+     */
+    @SuppressWarnings("unused")
+    public Cluster.Builder createClusterBuilderNoDebouncing() {
```
tolbertam, line 534: Nice, this is definitely common enough that it's nice to provide at this level.

driver-core/src/test/java/com/datastax/driver/core/CCMTestsSupport.java
```diff
@@ -0,0 +1,687 @@
...
+     * Useful to create tables or other objects inside the test keyspace,
+     * or to insert test data.
+     * <p/>
+     * Statements do not need to be qualified with keyspace name.
+     * <p/>
+     * The default implementation returns an empty list (no DDL statements required).
+     *
+     * @return The DDL statements to execute before the tests.
+     */
+    public Collection<String> createTestFixtures() {
```
tolbertam, line 549: The previous version of this method (getTableDefinitions) was abstract.  I think I like defaulting this but wanted to point out the difference.

adutra, line 549: Indeed, but I thought that not all tests need fixtures, so providing a default implementation saves some code.

driver-core/src/test/java/com/datastax/driver/core/CaseSensitivityTest.java
```diff
@@ -20,69 +20,38 @@
...
-        KeyspaceMetadata km = c.cluster.getMetadata().getKeyspace(fetchName);
-        assertNotNull(km);
-        assertEquals(realName, km.getName());
-    }
-
-    private void assertNotExists(CCMBridge.CCMCluster c, String name) {
-        assertNull(c.cluster.getMetadata().getKeyspace(name));
-    }
+@CCMConfig(clusterProvider = "createClusterBuilderNoDebouncing")
+public class CaseSensitivityTest extends CCMTestsSupport {
```
tolbertam, line 24: Nice, really shows how concise this makes tests that require individual clusters.

driver-core/src/test/java/com/datastax/driver/core/CCMTestsSupport.java
adutra, line 534: Yes! I use it with CaseSensitivityTest, but other tests can definitely use this alternate builder.

driver-core/src/test/java/com/datastax/driver/core/CaseSensitivityTest.java
tolbertam, line 24: Ah i see, this one didn't even really require individual clusters anyways.  Still nice for the clusterProvider.

driver-core/src/test/java/com/datastax/driver/core/ClusterDelegateTest.java
```diff
@@ -22,9 +22,9 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
-public class ClusterDelegateTest extends CCMBridge.PerClassSingleNodeCluster {
+public class ClusterDelegateTest extends CCMTestsSupport {
     @Override
-    protected Collection<String> getTableDefinitions() {
+    public Collection<String> createTestFixtures() {
```
tolbertam: Since this is the default behavior of createTestFixes, couldn't this method just be removed?

driver-core/src/test/java/com/datastax/driver/core/ControlConnectionTest.java
```diff
@@ -69,7 +69,7 @@ public long nextDelayMs() {
         };
 
         try {
-            ccm = CCMBridge.builder("test").withNodes(2).build();
```
tolbertam, line 72: This test could be expressed using CCMTestsSupport couldn't it?  Doesn't need to be done here, but just thinking if it is possible, I think it should be able to with

``` java
@CCMTestMode(PER_METHOD)
....
// For each method
@CCMConfig(numberOfNodes=X, createCluster=false)
```

tolbertam, line 72: Some tests would need clusterProviders as well..will think about this some more.

tolbertam:
Looks great to me so far.  Really like the concept.  Eventually we can look to express other tests still using CCMBuilder on it's own, but I think that can be done later. :+1:

tolbertam:
Was just looking at some tests that could use this and one thought I had was that it would be nice if we could do the following with this:
1. Have access to the CCMBridge so we can call operations on it.  For example, if a test needs to stop, decommission, bootstrap nodes, etc.
2. If we implement 1, we have a slight risk that shared context CCMBridge's could affect later tests.  To alleviate this we could consider marking created CCM Cluster's as 'single-use' or 'disposable'.  One possibility is to create a special subclass of CCMTestsSupport that makes CCMBridge accessible that only creates disposable context's  (The context is thrown away after the scope (per method vs. per class).  CCMBridge could remain private for CCMTestsSupport.

tolbertam:
Pushed a small fix for EventDebouncerIntegrationTest.  Also removed the `stop(clusterName)` and `remove(clusterName)` methods in CCMBridge since those are no longer used.

All other tests ran great.  There was a case where a node timed out while trying to start, I think that may be because we had too many nodes running (I think we need to reduce the cache size for jenkins).

driver-core/src/test/java/com/datastax/driver/core/ClusterDelegateTest.java
adutra: I removed all calls to this method that return an empty collection.

driver-core/src/test/java/com/datastax/driver/core/ControlConnectionTest.java
adutra, line 72: Most tests that use CCM with more than one node couldn't use `PerClassSingleNodeCluster` before for obvious reasons; they can now be refactored with `CCMTestsSupport`.
Tests that use `CCMBridge.CCMCluster` could also be refactored, `CCMCluster` being merely an encapsulation of a `CCMBridge` instance + a `Cluster` instance.

adutra:
> Have access to the CCMBridge so we can call operations on it. For example, if a test needs to stop, decommission, bootstrap nodes, etc.

Done, one can now call `CCMTestsSupport.ccmBridge()`.

> If we implement 1, we have a slight risk that shared context CCMBridge's could affect later tests. To alleviate this we could consider marking created CCM Cluster's as 'single-use' or 'disposable'. 

I took some inspiration from Spring Tests and created a new config option: `CCMConfig.dirtiesContext`. If this option is true, the context is stopped after the tests and removed from the cache.

adutra:
Also added support for multi-node and multi-dc setups:

``` java
@CCMConfig(numberOfNodes = { 3, 3 })
```

However there is a problem when starting more than one node individually with `ccm nodeX start`: the command fails starting from 2nd node, because of `--wait-other-notice` : previous nodes do not acknowledge the arrival of new nodes.

It seems that multi-node setups need `ccm create` to specify the `-n` option that operates some extra magic. Need to investigate further.

Once this is sorted out tests such as `ConsistencyTest` could be refactored and `CCMBridge.CCMCluster` removed.

EDIT:
- The problem with multi node setups was the storage port: it needs to be the same for all nodes in the cluster (fixed)
- The problem with multi DC setups was the snitch: it need to be manually configured (`ccm create` and `ccm populate` apparently do it, but `ccm add` doesn't)

driver-core/src/test/java/com/datastax/driver/core/CCMTestsSupport.java
```diff
@@ -0,0 +1,898 @@
...
+            for (int nodesPerDc : numberOfNodes) {
+                total *= nodesPerDc;
+            }
+            PORT_LOCK.lock();
+            try {
+                // storage port must be the same for all nodes in CCM cluster;
+                // find first available for first ip, and hope it will work for all ips
+                int storagePort = nextPort(1);
+                List<TokenRange> ranges = new TokenRange(FACTORY.minToken(), FACTORY.minToken(), FACTORY).splitEvenly(total);
+                int n = 1;
```
tolbertam: One issue with this is that the token ranges assigned to nodes aren't really DC aware, in this case it looks like nodes in DC1 will only be in the first 1/dc_count of the ring, DC2 in 2/dc_count of the ring, which will have some nodes own more data than others.

I got around this in [ScassandraCluster](https://github.com/datastax/java-driver/blob/2.1/driver-core/src/test/java/com/datastax/driver/core/ScassandraCluster.java#L250-L259), by having each DC range be offset by 100 the previous range's node at the same position (like CCM does), but I didn't use `TokenRange` (I probably should have).

adutra: Ha :) in the meanwhile I looked at CCM code and did the same 100-offset-shift magic – in the end, we have roughly the same code. 

But I confess still don't get why we need this offset; the [docs](https://docs.datastax.com/en/cassandra/2.1/cassandra/configuration/configGenTokens_c.html) say: "Calculate the tokens for each data center so that the hash range is evenly divided for the nodes in each data center." Why can't each datacenter have exactly the same initial tokens?

tolbertam: Even though nodes are in different data centers, they are still on the same ring.  Because of that, no node can have the same token.  They are offset by 100 (instead of 1 for example) to allow for nodes being replaced (you usually will bootstrap a new node at the node you are replacing token + 1) and offsetting them consistently by 100 from the other datacenter keeps them evenly distributed.

adutra:
Squashed and rebased on top of current 2.1.

tolbertam:
I noticed that the log4j.properties changed to only log at ERROR and omit timestamps, thread name, etc.  Is this intentional?   It's useful to have INFO logs for debugging in CI.

tolbertam:
Updated a couple of tests to use CCMTestsSupport (just to get myself more familiar with the framework / test a few things).  This is working super great :+1:

olim7t:
Closing in favor of #567

PR #567, opened 2016-01-05
Refactor CCM tests.

tolbertam:
This is #566 opened up against 2.1 branch.   I did this so was easier for me to review.


tolbertam:
Updated a couple of tests to use CCMTestsSupport (just to get myself more familiar with the framework / test a few things). This is working super great :+1:

adutra:
@tolbertam In the end I'm now in favor of allowing dirtiesContext with PER_CLASS; you can order test methods so it could make sense to have one context per class and allow methods to modify it.

I also reverted log4j.properties, that was a mistake (I use a less verbose version specially to avoid DCAwareRoundRobinPolicy to pollute the console).

olim7t:
A few general remarks:
- maybe we could rename `CCMTestMode` to something more explicit; I think something like `@CreateCCM(PER_METHOD)` reads more nicely.
- it's not clear to me when you would extend `CCMTestsSupport` vs "mix it in" like it's done in `ControlConnectionTest`.
- on a related note, couldn't we use `TestListener` to process the annotations and avoid having to reference `CCMTestsSupport` altogether (except for the provider methods of course)?

adutra:
> maybe we could rename `CCMTestMode` to something more explicit; I think something like `@CreateCCM(PER_METHOD)` reads more nicely.

Agreed, done.

> it's not clear to me when you would extend `CCMTestsSupport` vs "mix it in" like it's done in `ControlConnectionTest`.

The mixin mode would be useful if we one day create the equivalent functionality for Scassandra tests, in which case some tests could mix both CCM and Scassandra clusters. I this case, inheriting from one or the other would not be ideal.

> on a related note, couldn't we use `TestListener` to process the annotations and avoid having to reference `CCMTestsSupport` altogether (except for the provider methods of course)?

I'll give it a try. I initially discarded the idea because a) the interface is cumbersome and b) we would still need an instance of CCMTestsSupport to benefit from the helper methods.

tolbertam:
Big :+1: on the register concept, reduces putting try/finallys all over the place to make sure the cluster gets close.

adutra:
Squashed commits.

tolbertam:
very nice!  I like the recent change to removing the ccmBridgeProvider field in CCMConfig in favor of pushing the methods of CCMBridge.builder into the annotation.   Assuming everything is passing in CI, I am :+1: on this.

tolbertam:
I have a couple commits to fix some intermittently failing tests that I thought would be good to cherry-pick in if you agree (otherwise can wait until later):
- 8da0b39 - Reintroduce sleep to password-based tests, since apparently they are needing for 1.2.19 (I had thought that it had been fixed by 1.2.19 since I had trouble reproducing, but I must be wrong).
- bbeea1c - Change `RefreshConnectedHostTest#should_refresh_single_connected_host` to start 2 nodes instead of 3 (and stopping node3), and bootstrap node 3 in when needed (instead of starting it again).  Saves some time and makes the test pass more consistently in it's mock validation.

adutra:
Thanks @tolbertam I cherry-picked the suggested commits.

adutra:
We're almost done with this pull request and aiming to have repeatable, successful builds.

A summary of recent changes:
- new [ConditionChecker](https://github.com/datastax/java-driver/blob/3cff540d5423a02486d8b4e4a5983ea33a0fd92f/driver-core/src/test/java/com/datastax/driver/core/ConditionChecker.java) to periodically/asynchronously evaluate a condition. Can replace `Thread.sleep()` and `Mockito.verify(... , timeout())` in which cases it provides a more deterministic outcome, and is usually faster.
- [Using ConditionChecker](https://github.com/datastax/java-driver/commit/95732c9c139a10a018159445a57804f34988d21f) to evaluate if nodes are up or down, see [here](https://github.com/datastax/java-driver/blob/95732c9c139a10a018159445a57804f34988d21f/driver-core/src/test/java/com/datastax/driver/core/TestUtils.java#L331-L348) and [here](https://github.com/datastax/java-driver/blob/95732c9c139a10a018159445a57804f34988d21f/driver-core/src/test/java/com/datastax/driver/core/TestUtils.java#L527-L542).
- Some tests that use ConditionChecker: HostConnectionPoolTest, QueryOptionsTest, AbstractPoliciesTest, AbstractReconnectionHandlerTest.
- There is one change to production code in [AbstractReconnectionHandler](https://github.com/datastax/java-driver/commit/b44e59830231cc1502dd7dfaf6904ef6801e28ba#diff-f86e9c024179dcbb908c7eaafe5ac221L134).
- I am also proposing a less verbose version of log4j.properties, with optional loggers that one can uncomment when debugging specific tests.

adutra:
Squashed commits and rebased.

PR #572, opened 2016-01-15
Fixes for tests using enable_user_defined_functions 

tolbertam:
Some quick fixes after 2.1 -> 3.0 merge around the 'enable_user_defined_functions' configuration.

Tests that use this config option are failing with < 2.2 cassandra versions since C\* doesn't allow unknown configs in cassandra.yaml.  Added a `configVersionRequirements` map to CCMTestsSupport that contains a listing of configurations and their version requirements.  If a config is found that requires a certain version and the version requirement can't be met, it is not used.  Not sure if this is the best approach, but does the job.

Also updated FunctionExecutionExceptionTest to use enable_user_defined_functions.

There may be more issues with the merge.  I noticed that the 3.0 C\* version builds didn't run well, but 2 of the 3 failed because of what appeared to be CI issues, so there might not actually be a problem there.

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/exceptions/FunctionExecutionExceptionTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/exceptions/FunctionExecutionExceptionTest.java
@@ -15,6 +15,7 @@
  */
 package com.datastax.driver.core.exceptions;
 
+import com.datastax.driver.core.CCMConfig;
 import com.datastax.driver.core.CCMTestsSupport;
 import com.datastax.driver.core.utils.CassandraVersion;
 import com.google.common.collect.Lists;
@@ -23,6 +24,7 @@ import org.testng.annotations.Test;
 import java.util.Collection;
 
 @CassandraVersion(major = 2.2)
+@CCMConfig(config = "enable_user_defined_functions:true")
 public class FunctionExecutionExceptionTest extends CCMTestsSupport {
 
     @Override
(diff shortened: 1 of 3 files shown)


driver-core/src/test/java/com/datastax/driver/core/exceptions/FunctionExecutionExceptionTest.java
```diff
@@ -23,6 +24,7 @@
 import java.util.Collection;
 
 @CassandraVersion(major = 2.2)
+@CCMConfig(config = "enable_user_defined_functions:true")
```
adutra, line 27: Isn't `@CCMConfig` enough to make the test pass under all versions of C\* ? I believe `@CassandraVersion` is already filtering out this test for C\* < 2.2, although I don't see any harm in adding an extra check for invalid configurations in CCMTestsSupport.

EDIT: hmm I think I see the use case: it's for tests like SchemaChangesTest where only some methods are annotated with `@CassandraVersion` but the creation mode is PER_CLASS right?

adutra:
Added a little commit with a dummy fix for a test failing because the query trace wasn't available, I hope you don't mind.
I opened #573 for fixes that apply to both 2.1 and 3.0; let's keep this pull request for fixes applying only to 3.0.

adutra:
:+1:

driver-core/src/test/java/com/datastax/driver/core/exceptions/FunctionExecutionExceptionTest.java
tolbertam, line 27: correct, the filtering of enable_user_defined_functions is needed for the test classes that don't require 2.2, but where some of the test methods do and require user defined functions.  We could have gotten around it by using PER_METHOD on those particular tests, but it would have needed to create a separate cluster, so this is pretty much just a small optimization.

tolbertam:
I think this is good to merge, but think it would be good to wait until #573 is on 3.0

adutra:
Squashed and rebased with #573 on 3.0.

PR #574, opened 2016-01-19
JAVA-1037: Allow named parameters in simple statements.

olim7t:


driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
```diff
@@ -95,6 +99,30 @@ public String getQueryString() {
...
     /**
+     * The named values to use for this statement.
+     * <p/>
+     * A statement can use either positional or named values, but not both. So if this method returns a non-null result,
+     * {@link #getValues(ProtocolVersion, CodecRegistry)} will return {@code null}.
+     * <p/>
+     * Values for a RegularStatement (i.e. if this method does not return
+     * {@code null}) are not supported with the native protocol version 1: you
+     * will get an {@link UnsupportedProtocolVersionException} when submitting
+     * one if version 1 of the protocol is in use (i.e. if you've force version
```
adutra: Small typo: "you've force**d**"

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
```diff
@@ -95,6 +99,30 @@ public String getQueryString() {
     public abstract ByteBuffer[] getValues(ProtocolVersion protocolVersion, CodecRegistry codecRegistry);
 
     /**
+     * The named values to use for this statement.
+     * <p/>
+     * A statement can use either positional or named values, but not both. So if this method returns a non-null result,
+     * {@link #getValues(ProtocolVersion, CodecRegistry)} will return {@code null}.
+     * <p/>
+     * Values for a RegularStatement (i.e. if this method does not return
+     * {@code null}) are not supported with the native protocol version 1: you
```
adutra, line 108: Technically, the correct sentence would be "... if this method _or the following one_ does not return null".

driver-core/src/main/java/com/datastax/driver/core/SimpleStatement.java
```diff
@@ -88,6 +91,29 @@ public SimpleStatement(String query, Object... values) {
             throw new IllegalArgumentException("Too many values, the maximum allowed is 65535");
         this.query = query;
         this.values = values;
+        this.namedValues = null;
+    }
+
+    /**
+     * Creates a new {@code SimpleStatement} with the provided query string and named values.
+     * <p/>
+     * This version requires that the query string use named placeholders, for example:
```
adutra: Suggestion: "This constructor" would be clearer than "This version"

driver-core/src/main/java/com/datastax/driver/core/SimpleStatement.java
```diff
@@ -88,6 +91,29 @@ public SimpleStatement(String query, Object... values) {
...
+     * <pre>{@code
+     * new SimpleStatement("SELECT * FROM users WHERE id = :i", ImmutableMap.<String, Object>of("i", 1));}
+     * </pre>
+     * The type of the values will be handled the same way as with anonymous placeholders (see
```
adutra: I would write "type**s**", and also amend some issues in the javadoc of the other constructor: 
- "Note that the type**s** of the values provided [...] hence the driver cannot know what those value**s** should be"
- The sentence about "The codec to serialize each value will be chosen in the codec registry associated with this session's cluster" seems odd. maybe a remnant from JAVA-1007.

driver-core/src/main/java/com/datastax/driver/core/SimpleStatement.java
```diff
@@ -224,27 +268,7 @@ public SimpleStatement setRoutingKey(ByteBuffer... routingKeyComponents) {
         return this;
     }
 
-    /**
-     * <p/>
```
adutra, line 228: I would have kept the javadocs for internal reference, even if the method is now private.

tolbertam:
Looking good.  I have a few tests I've added for testing negative cases (using with protocol v1 with named parameters, missing parameters, etc.).  Will push that shortly!

tolbertam:
Slight nit.  I just noticed (and not sure if expected), but if you use the wrong type for a named parameter value the exception it returns makes it appear as if the Map is used as the value (which doesn't seem right) and returns an InvalidTypeException (which seems appropriate)

```
org.testng.TestException: 
Expected exception com.datastax.driver.core.exceptions.InvalidQueryException but got com.datastax.driver.core.exceptions.InvalidTypeException: Value 0 of type class com.google.common.collect.RegularImmutableMap does not correspond to any CQL3 type

    at org.testng.internal.Invoker.handleInvocationResults(Invoker.java:1497)
    at org.testng.internal.Invoker.invokeMethod(Invoker.java:754)
    at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:901)
    at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1231)
    at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:127)
    at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:111)
    at org.testng.TestRunner.privateRun(TestRunner.java:767)
    at org.testng.TestRunner.run(TestRunner.java:617)
    at org.testng.SuiteRunner.runTest(SuiteRunner.java:348)
    at org.testng.SuiteRunner.runSequentially(SuiteRunner.java:343)
    at org.testng.SuiteRunner.privateRun(SuiteRunner.java:305)
    at org.testng.SuiteRunner.run(SuiteRunner.java:254)
    at org.testng.SuiteRunnerWorker.runSuite(SuiteRunnerWorker.java:52)
    at org.testng.SuiteRunnerWorker.run(SuiteRunnerWorker.java:86)
    at org.testng.TestNG.runSuitesSequentially(TestNG.java:1224)
    at org.testng.TestNG.runSuitesLocally(TestNG.java:1149)
    at org.testng.TestNG.run(TestNG.java:1057)
    at org.testng.IDEARemoteTestNG.run(IDEARemoteTestNG.java:72)
    at org.testng.RemoteTestNGStarter.main(RemoteTestNGStarter.java:122)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:25)
    at java.lang.reflect.Method.invoke(Method.java:597)
    at com.intellij.rt.execution.application.AppMain.main(AppMain.java:144)
Caused by: com.datastax.driver.core.exceptions.InvalidTypeException: Value 0 of type class com.google.common.collect.RegularImmutableMap does not correspond to any CQL3 type
    at com.datastax.driver.core.SimpleStatement.convert(SimpleStatement.java:294)
    at com.datastax.driver.core.SimpleStatement.getValues(SimpleStatement.java:128)
    at com.datastax.driver.core.SessionManager.makeRequestMessage(SessionManager.java:548)
    at com.datastax.driver.core.SessionManager.executeAsync(SessionManager.java:131)
    at com.datastax.driver.core.AbstractSession.execute(AbstractSession.java:55)
    at com.datastax.driver.core.SimpleStatementIntegrationTest.should_fail_if_query_with_named_values_but_using_invalid_type(SimpleStatementIntegrationTest.java:75)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:39)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:25)
    at java.lang.reflect.Method.invoke(Method.java:597)
    at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:84)
    at org.testng.internal.Invoker.invokeMethod(Invoker.java:714)
    ... 22 more
Caused by: com.datastax.driver.core.exceptions.InvalidTypeException: Invalid type for map value, expecting java.lang.String but got class java.lang.Integer
    at com.datastax.driver.core.TypeCodec$AbstractMapCodec.serialize(TypeCodec.java:2001)
    at com.datastax.driver.core.TypeCodec$AbstractMapCodec.serialize(TypeCodec.java:1879)
    at com.datastax.driver.core.SimpleStatement.convert(SimpleStatement.java:291)
    ... 33 more
Caused by: java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.String
    at com.datastax.driver.core.TypeCodec$StringCodec.serialize(TypeCodec.java:788)
    at com.datastax.driver.core.TypeCodec$AbstractMapCodec.serialize(TypeCodec.java:1999)
    ... 35 more
```

tolbertam:
ah my mistake, I have a map of `<String,Integer>` and i'm trying to set a string, makes sense.

tolbertam:
Added an additional test for testing with wrong type (Feel free to squash as needed).

tolbertam:
Thought just popped into my head, should we add a `Session.execute(String, Map<String,Object> values)` to correlate with `Session.execute(String, Object... values)`?

tolbertam:
Another thing I noticed is that the QueryLogger does not log parameter values for SimpleStatements.  This extends to namedValues,  should I open up a separate ticket for that or should we handle that here?

tolbertam:
Was reading through the native protocol spec and it looks like you can't use named values until Protocol Version 3, so I think we need to adjust some of the validation and documentation to go around that requirement instead of requiring protocol version 2.

tolbertam:
It looks like it'll actually work with Protocol V2 and Cassandra 2.1.X+ since it is somewhat compatible since its just a new flag and knows how to parse it, but it won't work with Protocol V2 and Cassandra 2.0.X since it's unaware of the flag and doesn't know how to parse it:

`java.lang.IndexOutOfBoundsException: Invalid combined index of 158099, maximum is 59`

olim7t:
> should we add a Session.execute(String, Map<String,Object> values) to correlate with Session.execute(String, Object... values)?

I was initially reluctant, but in hindsight I don't see any good reason not to. I've added the methods.

newkek:
> It looks like it'll actually work with Protocol V2 and Cassandra 2.1.X+ since it is somewhat compatible since its just a new flag and knows how to parse it, but it won't work with Protocol V2 and Cassandra 2.0.X since it's unaware of the flag and doesn't know how to parse it:

I agree we could maybe check for the Cassandra version to be 2.1+.

Imo logging the values in the QueryLogger could be done in another ticket, for the rest I'm :+1:.

olim7t:
I did the check on the protocol version. Formally the spec says it only works with protocol v3, I'd rather not rely on the fact that it "happens to work" with C\* 2.1 and protocol v2. Also, testing the C\* version is unreliable because of mixed clusters.

Regarding the QueryLogger, I agree to create another ticket.

tolbertam:
> Regarding the QueryLogger, I agree to create another ticket.

Sounds good, opened as [JAVA-1040](https://datastax-oss.atlassian.net/browse/JAVA-1040)

tolbertam:
looks good to me! :+1: :shipit:

PR #575, opened 2016-01-20
JAVA-1039: Minor bugs in Event Debouncer

adutra:


tolbertam:
Both changes look great too me. I also added a test for verifying that overflown events are dropped. :+1:

driver-core/src/main/java/com/datastax/driver/core/EventDebouncer.java
```diff
@@ -166,7 +169,7 @@ private void scheduleImmediateDelivery() {
         while (state == State.RUNNING) {
             DeliveryAttempt previous = immediateDelivery.get();
             if (previous != null && !previous.isDone())
-                return;
+                previous.cancel();
```
olim7t, line 173: deliverEvents does not handle interruptions, so this is essentially a no-op if `previous` is already executing.

adutra, line 173: Absolutely, even the poll() method doesn't block. I decided to keep this call because 1) the intention counts as well (we would like the execution to be canceled, even if we know it's not possible) and 2) the callback could theoretically block and/or respond to interrupt requests. But I'm fine if you prefer to remove this instruction completely.

driver-core/src/test/java/com/datastax/driver/core/EventDebouncerTest.java
```diff
@@ -65,6 +67,33 @@ public void should_deliver_single_event() throws InterruptedException {
     }
 
     @Test(groups = "unit")
+    public void should_log_and_drop_events_on_overflow() throws InterruptedException {
```
adutra, line 71: Hmm are you sure we are not risking an OOM error here? Also, how long does this test take to complete?

adutra, line 71: Pushed a small commit to lower maxQueuedEvents.

tolbertam, line 71: When executed alone it takes < 300ms (a lot of that could be overhead in the framework though).  Makes sense to lower the max though just to verify the behavior in tests though :+1:

adutra, line 71: Oh really? I thought naively it would take quite  a while to browse 10000 events, but it's probably better to avoid that if we can :+1:

adutra:
Bad news: after applying the fixes in this branch the build in Jenkins (and Travis) failed at least twice with the following stack trace:

```
Starting com.datastax.driver.core.EventDebouncerTest.should_deliver_n_events_in_order_even_if_queue_full [Test #41]...
19:18:55 deliver 17
19:18:55 events received 17
19:23:55 java.lang.AssertionError: Timed out waiting for events
19:23:55    at com.datastax.driver.core.EventDebouncerTest$MockDeliveryCallback.awaitEvents(EventDebouncerTest.java:193)
```

So it seems that not returning early from scheduleImmediateDelivery() is not enough. I will investigate today.

adutra:
OK I got it:

```
initial state: events = [], eventCount = 0, maxPendingEvents = 1

T1: eventReceived(event1)
T1: count = eventCount.incrementAndGet()
T1: events.add(event1)
state: events=[event1], eventCount = 1, count = 1
T1: count == maxPendingEvents => scheduleImmediateDelivery()
T1: current.executeNow() => spawns T2
for X in [2..17]
    T1: eventReceived(eventX)
    T1: count = eventCount.incrementAndGet()
    T1: events.add(eventX)
    T1: count > maxPendingEvents => no delivery
state: events = [event1..event17], eventCount = 17, count = 17
for [1..17]
    T2: events.poll()
state: events = [], eventCount = 17
T1: eventReceived(event18)
T1: count = eventCount.incrementAndGet()
state: events = [], eventCount = 18, count = 18
T2: eventCount.addAndGet(-17)
state: events = [], eventCount = 1, count = 18
T2: if (!events.isEmpty()) => false, no delayed delivery scheduled => exit
T1: events.add(event18)
state: events = [event18], eventCount = 1, count = 18
T1: count > maxPendingEvents => no delivery
for X in [19..50]
    T1: eventReceived(eventX)
    T1: count = eventCount.incrementAndGet()
    T1: events.add(eventX)
    T1: count > maxPendingEvents => no delivery

final state: events = [event18..event50], eventCount = 33
```

It probably only happens for `maxPendingEvents = 1`. 

I think the fundamental problems are that:
- `events` and `eventCount` are not updated atomically;
- the local variable `count` is stale when read.

Suppressing the local variable `count` should be enough.

olim7t:
We should use `if (eventCount.get() > 0` instead of `if (!events.isEmpty())` at the end of `deliverEvents`. I'm not sure if it makes a big difference, but we maintain a separate count because we can't rely on `events.size()` (which is not constant time), so let's always rely on the count (even if `events.isEmpty` _is_ constant time).

tolbertam:
:+1:

PR #578, opened 2016-01-25
More CCM test fixes

adutra:
A few more fixes for intermittently failing tests. Nothing extraordinary, most changes consist in raising timeouts or pausing before/after tests. Note that this includes fixes for JAVA-1039 from #575 .

This PR also makes a few changes to allow CCMTestsSupport to be easily subclassed.
- Direct access to ccm, cluster and session is replaced with methods (also provides better protection against tests that would reassign these fields);
- `createTestFixtures` has been refactored into `onTestContextInitialized`, to allow e.g. graph queries to be executed at this step as well as normal queries.

The PR looks big mostly because of 183795f and 2289c5b.

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/AbstractPoliciesTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/AbstractPoliciesTest.java
@@ -15,6 +15,10 @@
  */
 package com.datastax.driver.core;
 
+import com.google.common.util.concurrent.Uninterruptibles;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testng.annotations.AfterMethod;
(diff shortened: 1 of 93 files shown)


adutra:
@tolbertam pushed 2 new commits with support for DSE workloads and also a way to fully customize your CCMBridge.Builder : 

``` java
@CCMConfig(ccmProvider = "createCcmBuilder")
public class MyTest extends CCMTestsSupport {

    public CCMBridge.Builder createCcmBuilder() {
        return CCMBridge.builder().withDSE().withVersion("4.8.3").withWorkload(1, "spark");
    }

}
```

tolbertam:
@adutra that looks perfect!  Both commits solve different issues I was having :+1:

PR #579, opened 2016-01-25
Document extra codecs

olim7t:
\+ minor changes to object mapper doc

### DIFF
--- a/manual/custom_codecs/extras/README.md
+++ b/manual/custom_codecs/extras/README.md
@@ -1,0 +1,419 @@
+## Optional codecs
+
+The driver's "extras" module provides additional [codec](../) implementations: these codecs are not required by core
+driver features, but will probably prove useful in a lot of client applications. You can also study their source code as
+a reference to write your own.
+
+The module is published as a separate Maven artifact:
+
+```xml
+<dependency>
+  <groupId>com.datastax.cassandra</groupId>
+  <artifactId>cassandra-driver-extras</artifactId>
+  <version>3.0.0-rc1</version>
+</dependency>
+```
+
+It also declares a number of optional dependencies, that you'll need to explicitly re-declare in your application
+depending on which codecs you plan to use (this will be covered for each case below).
+
+### Temporal types
+
+JDK 8 and Joda Time support have often been requested; they're not in the core module because we want to preserve
+compatibility with older JDKs and avoid extra dependencies.
+
+#### JDK 8
+
+[InstantCodec] maps [Instant] to CQL's `timestamp`:
+
+```java
+import com.datastax.driver.extras.codecs.jdk8.InstantCodec;
+import java.time.Instant;
+
+cluster.getConfiguration().getCodecRegistry()
+    .register(InstantCodec.instance);
+
+// schema: CREATE TABLE example(id int PRIMARY KEY, t timestamp)
+session.execute("INSERT INTO example (id, t) VALUES (1, ?)",
+        Instant.parse("2010-06-30T01:20:30Z"));
+```
+
+Similarly:
+
+* [LocalDateCodec] maps [LocalDate] to `date`;
+* [LocalTimeCodec] maps [LocalTime] to `time`.
+
+One problem with `timestamp` is that it does not store time zones. [ZonedDateTimeCodec] addresses that, by mapping a
+[ZonedDateTime] to a `tuple<timestamp,varchar>`:
+
+```java
+// Due to internal implementation details, you have to retrieve the tuple type from a live cluster:
+TupleType tupleType = cluster.getMetadata()
+        .newTupleType(DataType.timestamp(), DataType.varchar());
+cluster.getConfiguration().getCodecRegistry()
+        .register(new ZonedDateTimeCodec(tupleType));
+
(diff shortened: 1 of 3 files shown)


manual/custom_codecs/extras/README.md
```diff
@@ -0,0 +1,417 @@
...
+  <groupId>com.datastax.cassandra</groupId>
+  <artifactId>cassandra-driver-extras</artifactId>
+  <version>3.0.0-rc1</version>
+</dependency>
+```
+
+It also declares a number of optional dependencies, that you'll need to explicitly re-declare in your application
+depending on which codecs you plan to use (this will be covered for each case below).
+
+### Time types
```
adutra: I would suggest the term "temporal" to avoid any confusion with the CQL `TIME` type.

manual/custom_codecs/extras/README.md
```diff
@@ -0,0 +1,417 @@
...
+There is no extra codec for `time`, because by default the driver already maps that type to a `long` representing the
+number of milliseconds since midnight.
+
+[SimpleTimestampCodec]: http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/extras/codecs/date/SimpleTimestampCodec.html
+[SimpleDateCodec]: http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/extras/codecs/date/SimpleDateCodec.html
+
+
+### Enums
+
+Enums can be mapped two different ways:
```
adutra: "can be mapped _in_ two different ways".

manual/custom_codecs/extras/README.md
```diff
@@ -0,0 +1,417 @@
...
+// v2 saved as 12
+```
+
+See the JDK8 Optional section above for explanations about [TypeToken].
+
+[OptionalCodec_guava]: http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/extras/codecs/guava/OptionalCodec.html
+[Optional_guava]: http://docs.guava-libraries.googlecode.com/git-history/v16.0.1/javadoc/com/google/common/base/Optional.html
+
+
+### Primitive arrays
```
adutra: Actually, this package also contains an `ObjectArrayCodec` class – not sure though if it's worth mentioning.

manual/custom_codecs/extras/README.md
```diff
@@ -0,0 +1,417 @@
...
+
+[IntArrayCodec]: http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/extras/codecs/arrays/IntArrayCodec.html
+[arrays]: http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/extras/codecs/arrays/package-summary.html
+
+### Abstract utilities
+
+The module provides two abstract classes that act as building blocks for your own codecs:
+
+[MappingCodec] is useful when you can easily convert your Java type to another type that already has a codec.
+For example, if you have a `Price` object that translates to a plain `Integer`, you can piggyback on the built-in
```
adutra, line 395: I would personally map a Price to a BigDecimal :)

manual/custom_codecs/extras/README.md
olim7t: done

manual/custom_codecs/extras/README.md
olim7t: done

manual/custom_codecs/extras/README.md
olim7t: done

manual/custom_codecs/extras/README.md
olim7t, line 395: You can use an integer if you have a fixed number of decimals (e.g. multiply by 100 if you only go down to cents). Anyway that's not really the point, I think the example is clear enough.

manual/custom_codecs/extras/README.md
```diff
@@ -0,0 +1,419 @@
...
+driver features, but will probably prove useful in a lot of client applications. You can also study their source code as
+a reference to write your own.
+
+The module is published as a separate Maven artifact:
+
+```xml
+<dependency>
+  <groupId>com.datastax.cassandra</groupId>
+  <artifactId>cassandra-driver-extras</artifactId>
+  <version>3.0.0-rc1</version>
```
tolbertam, line 13: Is there anyway we can parameterize the version here?  Either that or we can put some kind of fake templating here like `DRIVER_VERSION`.

manual/custom_codecs/extras/README.md
```diff
@@ -0,0 +1,419 @@
...
+</dependency>
+```
+
+It also declares a number of optional dependencies, that you'll need to explicitly re-declare in your application
+depending on which codecs you plan to use (this will be covered for each case below).
+
+### Temporal types
+
+JDK 8 and Joda Time support have often been requested; they're not in the core module because we want to preserve
+compatibility with older JDKs, and avoid extra dependencies.
```
tolbertam: Small nit:  The comma isn't necessary here (I think).

manual/custom_codecs/extras/README.md
olim7t, line 13: I've thought about that, but the hard part is that we want it to be the version in the sources, because people can browse them from github.com.
We do a global replace before each release, but that's not 100% foolproof.

PR #580, opened 2016-01-26
JAVA-1042: Include DSE version and workload in Host data.

adutra:
This PR applies the patch by @djatnieks with additional minor changes to the javadocs of `Host` class.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
@@ -503,6 +503,14 @@ class ControlConnection implements Connection.Owner {
                 : null;
         host.setListenAddress(listenAddress);
 
+        if (row.getColumnDefinitions().contains("workload")) {
+            String dseWorkload = row.getString("workload");
+            host.setDseWorkload(dseWorkload);
+        }
+        if (row.getColumnDefinitions().contains("dse_version")) {
+            String dseVersion = row.getString("dse_version");
+            host.setDseVersion(dseVersion);
+        }
     }
 
     private static void updateLocationInfo(Host host, String datacenter, String rack, boolean isInitialConnection, Cluster.Manager cluster) {
@@ -565,7 +573,10 @@ class ControlConnection implements Connection.Owner {
         List<String> racks = new ArrayList<String>();
         List<String> cassandraVersions = new ArrayList<String>();
         List<InetAddress> broadcastAddresses = new ArrayList<InetAddress>();
+        List<InetAddress> listenAddresses = new ArrayList<InetAddress>();
         List<Set<String>> allTokens = new ArrayList<Set<String>>();
+        List<String> dseVersions = new ArrayList<String>();
+        List<String> dseWorkloads = new ArrayList<String>();
 
         for (Row row : peersFuture.get()) {
             InetSocketAddress rpcAddress = rpcAddressForPeerHost(row, connection.address, cluster, logMissingRpcAddresses);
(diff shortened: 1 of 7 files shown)


driver-core/src/main/java/com/datastax/driver/core/Host.java
```diff
@@ -221,6 +241,32 @@ public VersionNumber getCassandraVersion() {
...
     /**
+     * The DSE version the host is running.
+     * <p/>
+     * It is also possible for this information to be unavailable. In that case
+     * this method returns {@code null}, and the caller should always be aware of this
+     * possibility.
+     *
+     * @return the DSE version the host is running.
+     */
+    public VersionNumber getDseVersion() {
```
tolbertam, line 252: I've noticed that this can happen even when using DSE as the information isn't populated immediately as the host becomes available on 9042, but eventually the data is populated.  This could mean the driver may also not reliably have this data even if using DSE, but doesn't seem likely (except maybe in bootstrap cases).  Do we have a way to force refresh of this data?  That might be nice to have (this would be a separate request).

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
```diff
@@ -579,6 +581,10 @@ private static void refreshNodeListAndTokenMap(Connection connection, Cluster.Ma
             broadcastAddresses.add(row.getInet("peer"));
             if (metadataEnabled)
                 allTokens.add(row.getSet("tokens", String.class));
+            String dseWorkload = row.getColumnDefinitions().contains("workload") ? row.getString("workload") : null;
```
tolbertam, line 595: It looks like this seems to only set the workload and dse version for peers, not for the contact (local) node.

driver-core/src/main/java/com/datastax/driver/core/Host.java
```diff
@@ -113,6 +116,22 @@ void setListenAddress(InetAddress listenAddress) {
         this.listenAddress = listenAddress;
     }
 
+    void setDSEVersion(String dseVersion) {
```
tolbertam: There is a slight inconsistency here in that we use mixed case for getters and upper case for setters for 'DSE' (`getDseVersion` vs. `setDSEVersion`),we should make consistent and use mixed case.

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
tolbertam, line 595: have a fix for this (and the case issue), pushing shortly.

driver-core/src/main/java/com/datastax/driver/core/Host.java
tolbertam, line 252: It looks like I was wrong about this, the problem was the control host not being set, after that was fixed everything seems to be fine.

tolbertam, line 252: Seems like i was somewhat right in that it isn't populated immediately on node start, so having a nice API for forcing a refresh on this might be nice (as a separate pr later).

tolbertam, line 252: So after doing a little more testing, it seems that newer DSE versions handles this much better.  On 4.8.3 I saw that a Solr workload never sets the dse_version column, where it works just fine.

tolbertam:
Added tests for this and JAVA-1035.  Also made a change to refreshNodeListAndTokenMap to handle listen_address in peers table (which C\* doesn't currently provide).

olim7t:
:+1:

PR #581, opened 2016-01-29
JAVA-1046: (Dynamic)CompositeTypes need to be parsed as string literal, not blob.

adutra:


driver-core/src/main/java/com/datastax/driver/core/AggregateMetadata.java
```diff
@@ -101,11 +106,29 @@ static AggregateMetadata build(KeyspaceMetadata ksm, Row row, VersionNumber vers
         if (version.getMajor() >= 3) {
             stateType = DataTypeCqlNameParser.parse(stateTypeName, cluster, ksm.getName(), ksm.userTypes, null, false, false);
             String rawInitCond = row.getString("initcond");
-            initCond = rawInitCond == null ? null : codecRegistry.codecFor(stateType).parse(rawInitCond);
+            if (rawInitCond == null) {
+                initCond = null;
+            } else {
+                try {
+                    initCond = codecRegistry.codecFor(stateType).parse(rawInitCond);
```
tolbertam, line 113: It looks like this actually succeeds in the case the test case was created for, so the exception isn't raised and `rawInitCond` isn't used..  `initCond` is set to a `ByteBuffer` (so `getInitCond()` also returns a `ByteBuffer`).  It looks like the test is expecting it to return a string literal (i.e.: `'s@foo:i@32'`).  The cql string for the aggregate looks like:

> CREATE AGGREGATE ks_1.ag() SFUNC id STYPE 'org.apache.cassandra.db.marshal.DynamicCompositeType(s=>org.apache.cassandra.db.marshal.UTF8Type,i=>org.apache.cassandra.db.marshal.Int32Type)' INITCOND 0x80730003666f6f00806900040000002000;

Is this the expected behavior?

tolbertam, line 113: It seems like this is exactly how it is represented in cqlsh as well:

> CREATE AGGREGATE ks_1.ag()
>     SFUNC id
>     STYPE 'org.apache.cassandra.db.marshal.DynamicCompositeType(s=>org.apache.cassandra.db.marshal.UTF8Type,i=>org.apache.cassandra.db.marshal.Int32Type)'
>     INITCOND 0x80730003666f6f00806900040000002000;

So maybe this is working as desired, but it isn't clear to me what the expected behavior should be, will double back on the C\* change to see if this is what we are to expect.

tolbertam, line 113: It looks like it's possible this change is no longer needed, especially given @pcmanus' [comment](https://datastax-oss.atlassian.net/browse/JAVA-1046?focusedCommentId=26546&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26546):

> Fyi, I think we should actually change INITCOND to use a blob literal for custom type values (see my comment on CASSANDRA-11064 so you might want to hold on this a bit, as you may not have to change anything.

With and without this change, the behavior is exactly the same, although the PR seems harmless.  I think at the very least we should update the test and include that.

olim7t, line 113: Which Cassandra version did you test against? IIRC, some versions between 3.0 and 3.4 use string literals, then at some point it was changed to a blob.

tolbertam, line 113: I've tested with 3.0.4 and 3.4 which work.  < 3.4 and 3.0.4 raise an exception because of [CASSANDRA-11064](https://issues.apache.org/jira/browse/CASSANDRA-11064), so the code isn't really exercised.  

One concern I have is that the aggregate cql string we generate (`CREATE AGGREGATE ks_1.ag() SFUNC id STYPE 'org.apache.cassandra.db.marshal.DynamicCompositeType(s=>org.apache.cassandra.db.marshal.UTF8Type,i=>org.apache.cassandra.db.marshal.Int32Type)' INITCOND 0x80730003666f6f00806900040000002000;`) does work if we execute it against 3.0.4 and 3.4, however it fails against 2.2.5 with the error:

>  Invalid component representation: 0x80730003666f6f00806900040000002000)

tolbertam, line 113: To clarify, executing the aggregate cql fails on the C\* side with:

> java.lang.RuntimeException: java.util.concurrent.ExecutionException: org.apache.cassandra.exceptions.SyntaxException: Failed parsing CQL term: [s@foo:i@32] reason: SyntaxException line 1:1 no viable alternative at character '@'

tolbertam, line 113: To sum up:

# Using Byte HexString for initcond:

> CREATE AGGREGATE ks_1.ag() SFUNC id STYPE 'org.apache.cassandra.db.marshal.DynamicCompositeType(s=>org.apache.cassandra.db.marshal.UTF8Type,i=>org.apache.cassandra.db.marshal.Int32Type)' INITCOND 0x80730003666f6f00806900040000002000;

On 2.2.X, 3.0.3, 3.3 yields:

> Invalid component representation: 0x80730003666f6f00806900040000002000)

On 3.0.4, 3.4 it works.

# Using Composite Type Literal for initcond:

> CREATE AGGREGATE ks_1.ag() SFUNC id STYPE 'org.apache.cassandra.db.marshal.DynamicCompositeType(s=>org.apache.cassandra.db.marshal.UTF8Type,i=>org.apache.cassandra.db.marshal.Int32Type)' INITCOND 's@foo:i@32';

On 3.0.3, 3.3 yields:

> java.lang.RuntimeException: java.util.concurrent.ExecutionException: org.apache.cassandra.exceptions.SyntaxException: Failed parsing CQL term: [s@foo:i@32] reason: SyntaxException line 1:1 no viable alternative at character '@'

On 2.2.5, 3.0.4, and 3.4 it works.

tolbertam:
Updated test to reflect the desired behavior (getInitCond should return a ByteBuffer and the CQL string should contain the cql hexstring).  Also added test to validate you can create an aggregate using the CQL generated by `AggregateMetadata.toString`.  It works for 3.0.4+, 3.4+ but not 2.2.X.

tolbertam:
Didn't notice this at the time, but this PR probably needs to be reopened against 3.0.x right?

olim7t:
Closing to reopen against 3.0.x.

PR #584, opened 2016-02-08
[WIP] JAVA-1069: Bootstrap driver-examples module

adutra:


driver-examples/src/main/java/com/datastax/driver/examples/basic/Example0001.java
```diff
@@ -0,0 +1,80 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This example shows how to connect to a Cassandra cluster and
+ * extract some basic information from it.
+ *
+ * @see <a href="http://datastax.github.io/java-driver/manual/">Java driver online manual</a>
+ */
+public class Example0001 {
```
olim7t: It would be nice if the class names gave some indication of what they do:
- Example0001 => ReadCassandraVersion
- Example0002 => ReadHostsAndTablesMetadata
- Example0003 => CreateAndPopulateSchema

Numbering the examples by increasing order of complexity is compelling, but it might not scale well: if the counter is at 500 and we realize a simple example is missing, there will be no way to insert it.

driver-examples/src/main/java/com/datastax/driver/examples/basic/Example0003.java
```diff
@@ -0,0 +1,143 @@
+package com.datastax.driver.examples.basic;
+
+import com.datastax.driver.core.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This example shows how to create schema objects,
+ * load data into Cassandra then retrieve and display them.
+ *
```
olim7t: We could have a section here that indicates what each test mutates in the database.

driver-examples/src/main/java/com/datastax/driver/examples/basic/Example0001.java
```diff
@@ -0,0 +1,80 @@
...
+         * This step is important because it frees underlying resources (TCP connections, thread pools…).
+         * In a real application, you would typically do this at shutdown (for example, when undeploying your webapp).
+         */
+
+        Cluster cluster = null;
+
+        try {
+
+            cluster = Cluster.builder()                                                    // (1)
+                    .addContactPoint("127.0.0.1")
```
olim7t: Let's use `addContactPoints(String...)` and extract the array into a `CONTACT_POINTS` constant that will be the same in all examples. And also add a `PORT` constant.

olim7t:
Closing as this will now target 3.0.x.

PR #585, opened 2016-02-08
More enhancements to CCM tests.

adutra:
Highlights:
- Apply annotation-level configuration to CCM builders obtained via ccmProvider.
- Generic way to randomize ports in a CCMBridge.Builder instance.
- Generate CCM cluster names only when building CCMBridge instances.
- Improve algorithm to find available ports.


tolbertam:
This looks good to me!  :+1:

driver-core/src/test/java/com/datastax/driver/core/policies/DowngradingConsistencyRetryPolicyIntegrationTest.java
```diff
@@ -40,10 +40,9 @@
         numberOfNodes = 3,
         config = {
                 // tests fail often with write or read timeouts
-                "hinted_handoff_enabled:true",
                 "phi_convict_threshold:5",
-                "read_request_timeout_in_ms:100000",
```
tolbertam: Seems weird to require timeouts this large, will be interesting to see if it improves things in CI.

adutra: Yes, these timeouts are insane, and yet this test and ConsistencyTest both fail very frequently with write or read timeouts where the query is expected to succeed. Raising the timeouts has definitely made them less error-prone. My best guess is that in CI we are maybe sharing CPU resources among several vms and consequently processes get stuck very frequently waiting for an available core, so I suppose the replica that is meant to process the query gets paused, and the coordinator eventually triggers a timeout.

tolbertam:
Targeting 2.1.10 for this as it would be good to include, there are still a few problems I need to debug (i.e. why skip is being called for beforeTest many times, skewing our numbers a bit).

tolbertam:
With regards to the massive number of skips, it looks like something to do with how TestNG marks things, so not particularly our problem.  For example if you have a test with a beforeMethod & afterMethod and 8 test cases that all get skipped, it ends up creating 24 skips (3 for each test) since beforeMethod and afterMethod get a skip for each test.

adutra:
> With regards to the massive number of skips, it looks like something to do with how TestNG marks things, so not particularly our problem.

I have noticed that too. It advocates for a more ambitious refactoring so that all these before and after methods get converted into a Listener. Maybe in JAVA-1091?

adutra:
Rebased on current 2.1.

tolbertam:
Added commits from #617 and rebased on 2.1

tolbertam:
I think this is good to go from my perspective.   There is still an outstanding item that needs to be solved in ccm, and that is updating of nested yaml properties.  We should save that for another PR since it is pretty complex and we can work around it for now.

PR #586, opened 2016-02-09
Percentile tracker improvements

olim7t:


adutra:
One general remark: `PercentileTracker`is being used as a raw type in `PercentileSpeculativeExecutionPolicy` and `QueryLogger`.

olim7t:
> PercentileTracker is being used as a raw type in PercentileSpeculativeExecutionPolicy and QueryLogger

Right, I didn't notice this. Looking back at it, there's no real value in expressing the key type in the API, so I'll just make the class non-generic and have `computeKey` return `Object`.

> use Cluster as key

Good idea. `equals` and `hashcode` are not really an issue, identity comparison is what we want in this case.

> What is the problem with always registering the tracker, and hence maybe more than once?

That's right, I hadn't realized trackers were stored in a set. We can skip `onRegister` if it's already registered.

olim7t:
Force-pushed since this completely changed the 3rd commit. All pending remarks addressed.

olim7t:
Except your remark about JAVA-1055, but I think we should address that in a separate PR.

adutra:
lgtm. Regarding filtering errors in `PercentileTracker`:

1) Even without JAVA-1055, I still am in favor of not filtering anything in the default implementation of `PercentileTracker`, so that users can override the `include` method if they need to.
2) Speaking of filtering, `LatencyAwarePolicy` also filters the same errors in its private tracker, and that's not configurable. I would suggest that we leave that for now of course, but in JAVA-1055, we should imagine a pluggable way for both `LatencyAwarePolicy` and `PercentileTracker` to filter out requests based on whatever criteria the user wants.

driver-core/src/main/java/com/datastax/driver/core/PercentileTracker.java
```diff
@@ -0,0 +1,339 @@
...
+    /**
+     * Computes a key used to categorize measurements. Measurements with the same key will be recorded in the same
+     * histogram.
+     *
+     * @param host      the host that was queried.
+     * @param statement the statement that was executed.
+     * @param exception if the query failed, the corresponding exception.
+     * @return the key, or {@code null} if we're not yet ready to record.
+     */
+    protected abstract Object computeKey(Host host, Statement statement, Exception exception);
```
tolbertam, line 95: Nice, really like this design, gives a lot of flexibility for those who want to implement their own Percentile Tracker's (i.e. track by certain queries in an application.)

driver-core/src/main/java/com/datastax/driver/core/PercentileTracker.java
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    /**
+     * Computes a key used to categorize measurements. Measurements with the same key will be recorded in the same
+     * histogram.
+     *
+     * @param host      the host that was queried.
+     * @param statement the statement that was executed.
+     * @param exception if the query failed, the corresponding exception.
+     * @return the key, or {@code null} if we're not yet ready to record.
```
tolbertam: maybe this should be, `the key, or {@code null} if we don't want to record this measurement.`

driver-core/src/main/java/com/datastax/driver/core/PercentileTracker.java
tolbertam, line 95: Since we maintain a Histogram & Recorder for each key, should we make note too be careful not to have too many possibly unique keys?

driver-core/src/main/java/com/datastax/driver/core/PercentileTracker.java
```diff
@@ -0,0 +1,339 @@
...
+        protected final long highestTrackableLatencyMillis;
+        protected int numberOfSignificantValueDigits = 3;
+        protected int minRecordedValues = 1000;
+        protected long intervalMs = MINUTES.toMillis(5);
+
+        Builder(long highestTrackableLatencyMillis) {
+            this.highestTrackableLatencyMillis = highestTrackableLatencyMillis;
+        }
+
+        protected abstract B self();
```
tolbertam, line 277: Nice!  I've never seen Builders generalized so they may be extended by a subclass, usually there is one for each implementation. The self trick is cool.

driver-core/src/main/java/com/datastax/driver/core/policies/PercentileSpeculativeExecutionPolicy.java
```diff
@@ -27,41 +23,26 @@
...
+     *                                 {@link Cluster#register(LatencyTracker) registered} with the cluster when this
+     *                                 policy initializes.
+     * @param percentile               the percentile that a request's latency must fall into to be considered slow (ex:
+     *                                 {@code 99.0}).
+     * @param maxSpeculativeExecutions the maximum number of speculative executions that will be triggered for a given
+     *                                 request (this does not include the initial, normal request). Must be strictly
+     *                                 positive.
      */
-    public PercentileSpeculativeExecutionPolicy(PerHostPercentileTracker percentileTracker, double percentile, int maxSpeculativeExecutions) {
+    public PercentileSpeculativeExecutionPolicy(PercentileTracker percentileTracker,
```
tolbertam, line 44: Should we add some guidance here to use `ClusterWidePercentileTracker` over `PerHostPercentileTracker` in the general case or is that better left to the docs?

driver-core/src/main/java/com/datastax/driver/core/PercentileTracker.java
```diff
@@ -0,0 +1,339 @@
...
+     *
+     * @param host       the host (if this is relevant in the way percentiles are categorized).
+     * @param statement  the statement (if this is relevant in the way percentiles are categorized).
+     * @param exception  the exception (if this is relevant in the way percentiles are categorized).
+     * @param percentile the percentile (for example, {@code 99.0} for the 99th percentile).
+     * @return the latency (in milliseconds) at the given percentile, or a negative value if it's not available yet.
+     * @see #computeKey(Host, Statement, Exception)
+     */
+    public long getLatencyAtPercentile(Host host, Statement statement, Exception exception, double percentile) {
+        checkArgument(percentile >= 0.0 && percentile < 100,
```
tolbertam, line 124: This doesn't fill in the percentile value, just logs %f

driver-core/src/test/java/com/datastax/driver/core/PercentileTrackerTest.java
```diff
@@ -0,0 +1,166 @@
...
+        MemoryAppender appender = new MemoryAppender();
+        percentileLogger.addAppender(appender);
+
+        try {
+            long startTime = System.currentTimeMillis();
+            for (int i = 0; i < 100; i++) {
+                tracker.update(defaultHost, defaultStatement, defaultException, TimeUnit.NANOSECONDS.convert(1, TimeUnit.MILLISECONDS));
+            }
+
+            // HdrHistogram adjusts its max based on byte count, with a max of 1000, that allows a max vaue of 2047 (2^8-1).
```
tolbertam: my math is a bit off there :)  (should be 2^11-1).  Curious as to why setting a max to 1000 ends up allowing to 2047 though, will look into it, probably some implementation detail in hdrhisto

driver-core/src/main/java/com/datastax/driver/core/PercentileTracker.java
olim7t, line 277: AssertJ does something similar.

driver-core/src/main/java/com/datastax/driver/core/policies/PercentileSpeculativeExecutionPolicy.java
olim7t, line 44: I tried to add more explanations in the javadoc. I'll put a some guidance in the manual.

tolbertam:
Looks good to me!  :+1:

adutra:
:+1:

PR #587, opened 2016-02-09
JAVA-1050: Backport JAVA-819 to 2.1

adutra:

PR #588, opened 2016-02-10
JAVA-764: LWT / CAS Consistency bug on Retry Policy.

adutra:
This is based on top of #587.

Amend all retry policies and handle serial consistency levels accordingly (rethrow should be the preferred decision).

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/policies/CASRetryPolicyIntegrationTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/policies/CASRetryPolicyIntegrationTest.java
@@ -1,0 +1,108 @@
+/*
+ *      Copyright (C) 2012-2015 DataStax Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+package com.datastax.driver.core.policies;
+
+import com.datastax.driver.core.*;
+import com.datastax.driver.core.exceptions.NoHostAvailableException;
+import com.datastax.driver.core.exceptions.UnavailableException;
+import com.datastax.driver.core.utils.CassandraVersion;
+import org.testng.annotations.Test;
+
+import static com.datastax.driver.core.Assertions.assertThat;
+import static com.datastax.driver.core.ConsistencyLevel.ANY;
+import static com.datastax.driver.core.ConsistencyLevel.SERIAL;
+import static com.datastax.driver.core.CreateCCM.TestMode.PER_METHOD;
+import static org.assertj.core.api.Assertions.fail;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.*;
+
+/**
+ * Tests for retry policies in combination with lightweight transactions.
+ * <p/>
+ * Note: we can't extend {@link AbstractRetryPolicyIntegrationTest} here, because we are
+ * testing LWT statements and Scassandra doesn't support priming consistency levels yet.
+ *
+ * @jira_ticket JAVA-764
+ */
+@CCMConfig(
+        numberOfNodes = 3,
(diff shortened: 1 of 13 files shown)


driver-core/src/test/java/com/datastax/driver/core/policies/CASRetryPolicyIntegrationTest.java
```diff
@@ -0,0 +1,83 @@
...
+        ccm().waitForDown(2);
+        TestUtils.waitForDown(ccm().addressOfNode(2).getHostName(), cluster);
+
+        String ks = TestUtils.generateIdentifier("ks_");
+        session.execute(String.format("CREATE KEYSPACE %s WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 3}", ks));
+        useKeyspace(session, ks);
+        session.execute("CREATE TABLE foo(k int primary key, c int)");
+        session.execute("INSERT INTO foo(k, c) VALUES (0, 0)");
+
+        Statement s = new SimpleStatement("UPDATE foo SET c = 1 WHERE k = 0 IF EXISTS")
```
tolbertam: This will fail with C\* 1.2 because it doesn't support CAS operations, can you add `@CassandraVersion(major=2.0)` to this?

adutra: Thanks, good catch! I also modified the statement because UPDATE IF EXISTS is only possible since CASSANDRA-8610.

olim7t:
We're losing some information though, namely the fact that it's the paxos phase, not the mutation phase, that timed out. If we do the switch in the query handler, the retry policy will have no way to differentiate the two cases.
And the default behavior if the paxos phase timed out should probably be to not retry. If a quorum of nodes was not reachable to establish a proposal, it's unlikely that switching to the next host would fix the issue.

adutra:
Rebased on top of current JAVA-1050. Addressed @olim7t concerns.

Summary of recent changes:
- New method `ConsistencyLevel.isSerial()`;
- Amended `onUnavailable` implementations of `DefaultRetryPolicy`and `DowngradingConsistencyRetryPolicy` to rethrow on serial consistency levels (see a12838c);
- It turns out that `onWriteTimeout` implementations need not be amended (or so I mistakenly assumed);
- Javadoc improvements to clarify the meaning of consistency levels.

tolbertam:
> New method ConsistencyLevel.isSerial();

Nice, that also takes care of [JAVA-982](https://datastax-oss.atlassian.net/browse/JAVA-982)

tolbertam:
Now that #587 is merged, can we rebase this on 2.1?

olim7t:
I've rebased on 2.1 and removed the commits related to our initial approach. This will require a bit of cleanup/squashing before merging.

olim7t:
After giving it more thought, I think there is no reason not to retry on the next host when the CL is serial: the "network-isolated coordinator" hypothesis applies as well.
I pushed a commit to try the next host with the statement's original CL.

olim7t:
One more change: retry with the previous CL instead of the original CL. We don't have the case in default policies, but the CL might already have been downgraded in a previous retry.

adutra:
lgtm :+1: 
Small note before we squash commits together: let's try to keep the commit for JAVA-982 separate.

olim7t:
Done.

driver-core/src/test/java/com/datastax/driver/core/policies/CASRetryPolicyIntegrationTest.java
```diff
@@ -0,0 +1,103 @@
...
+import static com.datastax.driver.core.CreateCCM.TestMode.PER_METHOD;
+import static org.assertj.core.api.Assertions.fail;
+import static org.mockito.Matchers.eq;
+import static org.mockito.Mockito.*;
+
+/**
+ * Tests for retry policies in combination with lightweight transactions.
+ * <p/>
+ * Note: we can't extend {@link AbstractRetryPolicyIntegrationTest} here, because we are
+ * testing LWT statements and Scassandra doesn't support priming consistency levels yet.
```
tolbertam, line 36: I'll add support for that to scassandra and create a branch with tests.

tolbertam, line 36: I think these tests need to be changed now that we retry once on unavailable, and then there are no remaining hosts to try, so we get an NHAE

tolbertam:
Opened scassandra/scassandra-server#143 to add the capability to simulate errors returning CL other than what the query specified to help test this.  Created [java764tests](https://github.com/datastax/java-driver/tree/java764tests) branch with added tests which pass against this branch that will open PR for when version of scassandra released with this feature.

PR #589, opened 2016-02-10
JAVA-1011: Expose PoolingOptions defaults values.

newkek:


adutra:
Missing an entry to the changelog, otherwise :+1:

tolbertam:
looks good to me as well :+1:, we should make note to do a follow on action when merged in 3.0 to add an entry into DEFAULTS for Protocol V4 that references the same map as V3.

newkek:
Pushed the changelog change. Good point @tolbertam, not sure how to do that, maybe opening another PR for 3.0 would be simpler no?

tolbertam:
:+1:  I think that is the correct route to go, another PR after this is merged into 3.0.

olim7t:
:+1: 
It would be nice to have javadocs, at least for the map.
Regarding protocol v4, I think it's better not to duplicate the entries if the values don't change. We can document that you need to use the highest version that's lower than or equal to your version. The immutable map iterates in insertion order, so that's easy to do, even from client code.

newkek:
Pushed the javadoc, thanks for the feedback.

olim7t:
Re-reading my last comment, I realized it would be more logical to use V1 as the first key, to enforce the "highest key <= actual version" rule. I added that in a separate commit.

PR #591, opened 2016-02-11
JAVA-727: Allow monotonic timestamp generators to drift in the future + use microsecond precision when possible.

adutra:


driver-core/src/test/java/com/datastax/driver/core/AtomicMonotonicTimestampGeneratorTest.java
```diff
@@ -29,10 +30,58 @@
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.testng.Assert.assertEquals;
 import static org.testng.Assert.fail;
 
-public class AtomicMonotonicTimestampGeneratorTest {
+public class AtomicMonotonicTimestampGeneratorTest extends CCMTestsSupport {
```
olim7t: Could we keep it unit test only? It pains me that we have to spin up a Cassandra cluster to test such a simple component.

olim7t:
Open question: would there be some value in leaving the old behavior as an alternative? One issue with a drifting generator is that you could loose the real-time ordering between multiple clients if only one of them has drifted in the future. Although this was possible as well with the previous implementations... At least we should document it.

Speaking of documentation, we should also update the javadoc of the two implementations, and probably briefly mention the drifting behavior in the manual (`manual/query_timestamps/`).

driver-core/src/main/java/com/datastax/driver/core/Clock.java
```diff
@@ -15,25 +15,169 @@
...
+
+/**
+ * Provides the current time with microseconds precision with some reasonable accuracy through
+ * the use of clock_gettime (http://man7.org/linux/man-pages/man2/clock_gettime.2.html via JNA).
+ * <p/>
+ * But because calling clock_gettime is slightly expensive, we only call it once per UPDATE_INTERVAL_MS
+ * and adds the number of nanoseconds since the last call to get the current time, which is good
+ * enough accuracy for our purpose (see CASSANDRA-6106).
+ */
+class NativeClock implements Clock {
```
adutra, line 87: This comes from Sylvain Lebresne's  suggestion in CASSANDRA-6106.

driver-core/src/main/java/com/datastax/driver/core/Clock.java
```diff
@@ -15,25 +15,169 @@
...
     /**
-     * Returns the current time in milliseconds
+     * Returns the current time in microseconds
      *
-     * @return the difference, measured in milliseconds, between the current time and midnight, January 1, 1970 UTC.
-     * @see System#currentTimeMillis()
+     * @return the difference, measured in microsencods, between the current time and midnight, January 1, 1970 UTC.
      */
-    long currentTime();
+    long currentTimeMicroseconds();
```
adutra: Changed the meaning of this method to return microseconds instead of milliseconds.

adutra:
> Open question: would there be some value in leaving the old behavior as an alternative? One issue with a drifting generator is that you could loose the real-time ordering between multiple clients if only one of them has drifted in the future. 

I very much doubt that two clients would be in sync enough to be able to order their statements so that the real last one wins. But if everyone agrees, I can leave the old behavior as an alternative.

> Speaking of documentation, we should also update the javadoc of the two implementations, and probably briefly mention the drifting behavior in the manual (manual/query_timestamps/).

Yes good catch, I will add a note on that.

adutra:
You will notice that commit 07ab2a0 introduces a second Clock implementation that uses JNA to obtain a real microsecond-precision clock. It can only be activated on linux by setting the appropriate system property to true.

driver-core/src/main/java/com/datastax/driver/core/AbstractMonotonicTimestampGenerator.java
```diff
@@ -0,0 +1,58 @@
...
+ * Generally speaking, this granularity is the millisecond, and
+ * the sub-millisecond part is simply a counter that gets incremented
+ * until the next clock tick, as provided by {@link System#currentTimeMillis()};
+ * on Linux systems however, it is possible to have a better granularity by using a JNA
+ * system call to {@code clock_gettime}. To benefit from this system call, set the system
+ * property {@code com.datastax.driver.USE_NATIVE_CLOCK} to {@code true}.
+ * <p/>
+ * This implementation is inspired by {@code org.apache.cassandra.service.ClientState#getTimestamp()}
+ */
+public abstract class AbstractMonotonicTimestampGenerator implements TimestampGenerator {
```
adutra: Note that this class was previously package private; I'm proposing to make it public so that the documentation here is visible and valid for the two subclasses shipped with the driver: AtomicMonotonicTimestampGenerator and ThreadLocalMonotonicTimestampGenerator.

driver-core/src/main/java/com/datastax/driver/core/AbstractTimestampGenerator.java
```diff
@@ -1,49 +0,0 @@
...
-package com.datastax.driver.core;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-/**
- * Base implementation for generators based on {@link System#currentTimeMillis()} and a counter to generate
- * the sub-millisecond part.
- */
-abstract class AbstractMonotonicTimestampGenerator implements TimestampGenerator {
```
adutra, line 25: I fixed the filename which wasn't in sync with the class name.

adutra:
Switched to JNR call to `gettimeofday` as it has at least 2 advantages:
- It is 10x faster than a JNA call to `clock_gettime`
- It is available under more systems (including Mac OS - not tested under Windows)

The call costs avg. 200-300 ns in my benchmarks where a call to `System.currentTimeMillis` is 30-50 ns, so it is still more expensive. To compensate, I kept the updating system initially designed by Sylvain in CASSANDRA-6106. That mechanism brings the cost of a call to `Native.currentTimeMicros` to 30-50 ns, that is, equivalent to `System.currentTimeMillis`.

Complete benchmark results under Linux:

```
Benchmark                     Mode  Cnt     Score     Error  Units
ClockBenchmark.testCurMillis  avgt   20    48.024 ±   1.805  ns/op
ClockBenchmark.testNanoTime   avgt   20    38.167 ±   2.352  ns/op
ClockBenchmark.testJNA        avgt   20  4028.256 ± 166.911  ns/op
ClockBenchmark.testJNAApprox  avgt   20    37.290 ±   0.979  ns/op
ClockBenchmark.testJNR        avgt   20   279.229 ±  16.476  ns/op
ClockBenchmark.testJNRApprox  avgt   20    38.222 ±   2.153  ns/op
```

Note: The "approx" benchmarks use Sylvain's method.

adutra:
Squashed and rebased

driver-core/src/main/java/com/datastax/driver/core/AbstractTimestampGenerator.java
olim7t, line 25: :+1:

driver-core/src/main/java/com/datastax/driver/core/AtomicMonotonicTimestampGenerator.java
```diff
@@ -18,17 +18,25 @@
...
  * <p/>
- * If that rate is exceeded, a warning is logged and the timestamps don't increment anymore until the next clock
- * tick. If you consistently exceed that rate, consider using {@link ThreadLocalMonotonicTimestampGenerator}.
+ * Generally speaking, this granularity is the millisecond, and
+ * the sub-millisecond part is simply a counter that gets incremented
+ * until the next clock tick, as provided by {@link System#currentTimeMillis()}.
+ * <p/>
+ * On Linux systems, however, it is possible to have a better granularity by using a JNA
+ * call to {@code clock_gettime}. To benefit from this system call, set the system
+ * property {@code com.datastax.driver.USE_NATIVE_CLOCK} to {@code true}.
```
olim7t: It's now set to true by default.

driver-core/src/main/java/com/datastax/driver/core/Native.java
```diff
@@ -0,0 +1,106 @@
...
+     * Returns the current timestamp with microsecond precision
+     * via a system call to {@code gettimeofday}.
+     *
+     * @return the current timestamp with microsecond precision.
+     */
+    static long currentTimeMicros() {
+        LibC.Timeval tv = new LibC.Timeval(LIB_C_RUNTIME);
+        if (LIB_C.gettimeofday(tv, null) != 0)
+            LOGGER.error("gettimeofday failed");
+        return tv.tv_sec.get() * 1000000 + tv.tv_usec.get();
```
olim7t, line 106: What happens under windows? Does `isLibCLoaded()` return false, or is it loaded but `currentTimeMicros()` fails and returns garbage ?

manual/query_timestamps/README.md
```diff
@@ -59,6 +59,48 @@ session.execute(statement);
...
+But in practice, the real accuracy of generated timestamps is largely dependent on the
+granularity of the underlying operating system's clock.
+
+For most systems, this minimum granularity is the millisecond, and 
+the sub-millisecond part of generated timestamps is simply a counter that gets incremented
+until the next clock tick, as provided by `System.currentTimeMillis()`.
+
+On some systems, however, it is possible to have a better granularity by using a [JNR]
+call to [gettimeofday]. This system call will be used when available, unless the system
+property `com.datastax.driver.USE_NATIVE_CLOCK` is explicitly set to `false`.
```
olim7t, line 87: The optional dependency also has to be pulled in.

manual/query_timestamps/README.md
```diff
@@ -59,6 +59,48 @@ session.execute(statement);
 
 [tsg]: http://docs.datastax.com/en/drivers/java/2.1/com/datastax/driver/core/TimestampGenerator.html
 
+### Overview of Timestamp Generators
+
+The driver ships with two different implementations of [TimestampGenerator][tsg]:
+
+1. [AtomicMonotonicTimestampGenerator][amtsg], which guarantess monotonicity of timestmaps for all threads;
+2. [ThreadLocalMonotonicTImestampGenerator][tlmtsg], which guarantees per-thread monotonicity of timestamps.
```
olim7t: timestmaps => timestamps

ThreadLocalMonotonicTImestampGenerator => ThreadLocalMonotonicT**_i_**mestampGenerator (also in the referenced URL)

driver-core/pom.xml
```diff
@@ -85,6 +85,13 @@
     </dependency>
 
     <dependency>
+      <groupId>com.github.jnr</groupId>
+      <artifactId>jnr-ffi</artifactId>
+      <version>2.0.7</version>
```
olim7t: We might want to extract the version as a property, for the reasons explained in 0853ed8.

driver-core/src/main/java/com/datastax/driver/core/AtomicMonotonicTimestampGenerator.java
adutra: Good catch, will amend, thanks.

driver-core/src/main/java/com/datastax/driver/core/Native.java
adutra, line 106: `isLibCLoaded` returns false. I added a check that will throw `UnsupportedOperationException` if one attempts to call `currentTimeMicros()` while LibC is not loaded (previously, it would throw an NPE).
I just checked Windows, LibC is not available and  `isLibCLoaded` returns false.
I also checked that when JNR library is not available on the classpath, `isLibCLoaded` returns false too.
In both cases the driver logs this:

```
    0      [main] INFO  com.datastax.driver.core.Native - Could not load JNR LibC Library, native calls will not be available
```

manual/query_timestamps/README.md
adutra, line 87: Amended, thanks.

adutra:
All remarks addressed, thanks for your feedback.

adutra:
Squashed again in preparation for JAVA-444, sorry for the inconvenience.

olim7t:
Rebased on current 2.1 (2708bd7)

adutra:
Squashed and rebased on top of current 2.1.

olim7t:
Rebased again, with a few minor changes (mainly doc).

driver-core/src/main/java/com/datastax/driver/core/AbstractMonotonicTimestampGenerator.java
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * Compute the next timestamp, given the last timestamp previously generated.
+     * The returned timestamp should be strictly greater than the previous one.
+     * <p/>
+     * This implementation is inspired by {@code org.apache.cassandra.service.ClientState#getTimestamp()}
+     *
+     * @return the next timestamp to use
+     */
+    protected long computeNext(long last) {
+        long current = clock.currentTimeMicros();
```
tolbertam: Could there be value in adding a warning statement if we detect current being 1000ms less than last?  That could indicate a clock drift, which might be worth logging (although we wouldn't want to spam it(.

driver-core/src/main/java/com/datastax/driver/core/Clock.java
```diff
@@ -15,25 +15,145 @@
...
\ No newline at end of file
+
+}
+
+/**
+ * Provides the current time with microseconds precision with some reasonable accuracy through
+ * the use of {@link Native#currentTimeMicros()}.
+ * <p/>
+ * But because calling JNR methods is slightly expensive (roughly 300 ns for {@link Native#currentTimeMicros()}),
+ * we only call it once per {@code UPDATE_INTERVAL_MS}
```
tolbertam: Shouldn't this be `UPDATE_INTERVAL_NS` since that is the time precision used?

tolbertam: Woops, scratch that, I misunderstood, it is milliseconds.

driver-core/src/main/java/com/datastax/driver/core/Clock.java
```diff
@@ -15,25 +15,145 @@
...
+ * is comparable to a call to {@link System#currentTimeMillis()}.
+ */
+class NativeClock implements Clock {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(NativeClock.class);
+    private static final long UPDATE_INTERVAL_MS = 1000;
+    private static final TimeFetcher TIME_FETCHER = new TimeFetcher();
+
+    static {
+        Thread updater = new Thread(TIME_FETCHER, "Native Clock Fetcher");
```
tolbertam: Question about the use of a thread.  Couldn't we change currentTimeMicros() to check if enough time has elapsed instead and then calculate?  Or is the overhead of that check too much?  I suppose a problem with that could be multiple calls to currentTimeMicros() concurrently, but we could synchronize that part of it somehow since it's pretty fast (30ns).

manual/query_timestamps/README.md
```diff
@@ -59,6 +59,58 @@ session.execute(statement);
...
+[amtsg]: http://docs.datastax.com/en/drivers/java/2.1/com/datastax/driver/core/AtomicMonotonicTimestampGenerator.html
+[tlmtsg]: http://docs.datastax.com/en/drivers/java/2.1/com/datastax/driver/core/ThreadLocalMonotonicTimestampGenerator.html
+
+#### Accuracy
+
+Both implementations strive to achieve microsecond resolution on a best-effort basis.
+But in practice, the real accuracy of generated timestamps is largely dependent on the
+granularity of the underlying operating system's clock.
+
+For most systems, this minimum granularity is the millisecond, and
```
tolbertam: 'the millisecond' can just be 'millisecond'

tolbertam:
Looks good to me, since #593 is based on this, I think we should work to merge that one in stead.

adutra:
@tolbertam added two new commits to address your excellent comments:
1. I benchmarked the current native clock implementation against your idea of fetching timestamps inline. They gave identical results, so I agree to remove the updater thread.
2. Added a warning in case of clock drift/skew. The warning is logged once every second. While I think the idea is interesting, I'm afraid we could spam the logs even with the one-second threshold: e.g. when a machine enters daylight savings time and goes back one hour earlier, the warning will be logged every second until the generator is capable of generating a timestamp that is ahead of the last one, which, depending on the rate timestamps are requested, can take quite a while. 

Logging a warning also comes at a cost, namely, the cost of a System.nanoTime() call for _every_ bad timestamp (50 ns on my machine in average), not counting the cost of the logger call itself (which is paid only once every second).

tolbertam:
Looks excellent!   I am working on a couple of tests to add to simulate the warning events that i'll push up shortly.  Granted that all works as I expect I think this is good to go :+1:

driver-core/src/main/java/com/datastax/driver/core/AbstractMonotonicTimestampGenerator.java
```diff
@@ -0,0 +1,84 @@
...
+            maybeLogWarning(current, last);
+            return last + 1;
+        }
+        return current;
+    }
+
+    private void maybeLogWarning(long current, long last) {
+        if (LOGGER.isWarnEnabled()) {
+            long now = System.nanoTime();
+            if (now > lastClockDriftWarning + CLOCK_DRIFT_WARNING_INTERVAL) {
```
tolbertam: It looks like it's possible for this to be logged multiple times with concurrent access, since we check lastClockDriftWarning and then set it.  I think if we change lastClockDriftWarning to an atomic long and do a CAS it will fix this.   Do you think that change is worth it?  Have it locally so I can push with my tests.

tolbertam: Went ahead and pushed a change for that.  Can revert if isn't the appropriate fix.

driver-core/src/main/java/com/datastax/driver/core/Clock.java
```diff
@@ -15,25 +15,121 @@
...
+    }
+
+    private final AtomicReference<FetchedTime> lastFetchedTime = new AtomicReference<FetchedTime>(fetchTimeMicros());
+
+    @Override
+    public long currentTimeMicros() {
+        FetchedTime spec = lastFetchedTime.get();
+        long curNano = System.nanoTime();
+        if (curNano > spec.nanoTimeAtCheck + ONE_SECOND_NS) {
+            spec = fetchTimeMicros();
```
tolbertam: We could in theory do this multiple times within a second with concurrent access, but it doesn't seem like that big of a concern.  To get around it we could create a new `FetchedTime` if the check passes and then do a CAS, and if the CAS succeeds, call fetchTimeMicros and fill in the fields of the new `FetchedTime`.

driver-core/src/main/java/com/datastax/driver/core/AbstractMonotonicTimestampGenerator.java
adutra: AtomicLong sounds good!

driver-core/src/main/java/com/datastax/driver/core/Clock.java
tolbertam: Here's what I had in mind for that: https://gist.github.com/tolbertam/99a4653c73280e2ce5a1 , not really sure if it is a worthy optimization though or how much time it'd actually save.

adutra: You mean, create an empty `FetchedTime` with zeroes ? I think that could lead to a situation where one thread CAS a new (empty) fetched time, then another thread reads it before the first thread gets a chance to fill in the fields.

adutra: What about instead do a regular CAS if the check passes; this will not prevent two threads from calling fetchTimeMicros() in parallel, but seems more correct indeed.

Smtg along the lines of:

``` java
    public long currentTimeMicros() {
        FetchedTime spec = lastFetchedTime.get();
        long curNano = System.nanoTime();
        if (curNano > spec.nanoTimeAtCheck + ONE_SECOND_NS) {
            lastFetchedTime.compareAndSet(spec, spec = fetchTimeMicros());
        }
        return spec.timeInMicros + ((curNano - spec.nanoTimeAtCheck) / 1000);
    }
```

Also I think we could turn NativeClock and SystemClock into singletons.

tolbertam: > You mean, create an empty FetchedTime with zeroes ? I think that could lead to a situation where one thread CAS a new (empty) fetched time, then another thread reads it before the first thread gets a chance to fill in the fields.

Good point, I hadn't even really considered that.  That will definitely be a problem.

> lastFetchedTime.compareAndSet(spec, spec = fetchTimeMicros());

Wouldn't that not really provide much value since it requires evaluation of fetchTimeMicros() whether or not the CAS succeeds?

> Also I think we could turn NativeClock and SystemClock into singletons.

:+1:

tolbertam: Oh, I misinterpreted. You do say it will not prevent two threads from calling in parallel, but seems more correct.  On that point I agree :+1:.   Since we are talking about nanoseconds here I agree that there isn't too much penalty of doing it multiple times concurrently.

adutra: Went ahead and introduced a CAS like in my example.

driver-core/src/main/java/com/datastax/driver/core/Native.java
```diff
@@ -0,0 +1,106 @@
...
+    private static final Runtime LIB_C_RUNTIME;
+
+    static {
+        LibC libc = null;
+        Runtime runtime = null;
+        try {
+            libc = LibraryLoader.create(LibC.class).load("c");
+            runtime = Runtime.getRuntime(libc);
+        } catch (Throwable t) {
+            LOGGER.info("Could not load JNR LibC Library, native calls will not be available");
```
tolbertam: it would be nice to include the throwable stack (maybe at DEBUG only though?) So it's easier to understand why it couldn't be loaded.  In one case I had a classloader issue which wasn't obvious so it would be good to have a way to expose the exception.

tolbertam:
The OSGI example suite was failing because jnr was missing, I created [java727-osgi](https://github.com/datastax/java-driver/tree/java727-osgi) (commit 9e6e7b6) which may resolve this.  However, the solution isn't that simple because some modules JNR depends on are not configured as OSGi modules.  These are:
- jnrx86-asm ([Pull request to enable OSGi](https://github.com/jnr/jnr-x86asm/pull/1))
- jffi ([Pull request to enable OSGi](https://github.com/jnr/jffi/pull/29))

When building & installing the maven modules w/ those PRs (and removing SNAPSHOT from the versions), the OSGi tests pass.

I think if we want OSGi work in the near term, we should make the JNR dependency optional and have the driver OSGi module not import their packages.  If those PRs get merged and versions with them released, we could add JNR back.

adutra:
Thank you very much @tolbertam for catching the OSGi problem. After reviewing your branch another possible solution came to me:

Actually JNR is de facto optional for the driver: if the dependency is missing, the driver falls back to SystemClock automatically. So my idea was to make it optional for OSGi bundles only. See [here](https://github.com/datastax/java-driver/blob/013e70123c637f3fea972ad5b42ccfb893cdc13a/driver-core/pom.xml#L181).

The OSGi tests pass with this configuration, and we do see the warning:

```
[FelixStartLevel] INFO com.datastax.driver.core.Native - Could not load JNR LibC Library, native calls will not be available (set this logger level to DEBUG to see the full stack trace)
```

I also cherry-picked your changes to the OSGi application.

tolbertam:
Nice @adutra!  That was not something I was aware was possible.  Given this and that JNR cannot work in OSGi without the aforementioned PRs against jffi and jnr-x86asm, do we even need `jnr.*` in `Import-Packages`?

adutra:
> we even need jnr.\* in Import-Packages?

Right now, no; but my idea was that if one day those PRs are merged and OSGi bundles are available for all JNR libraries, then people could start including JNR in their OSGi runtimes and the driver bundle would simply detect that the package is present and activate it.

tolbertam:
> Right now, no; but my idea was that if one day those PRs are merged and OSGi bundles are available for all JNR libraries, then people could start including JNR in their OSGi runtimes and the driver bundle would simply detect that the package is present and activate it.

Got it, that makes sense, thanks for clearing that up.  I need to look into whether or not it will actually work with just `jnr.*`.   I messed around quite a bit with my environment last night, so it's possible it does work.  Will try it with the OSGified modules to see if it works.

tolbertam:
Looks like it doesn't work, because it doesn't pull in jnr.ffi.provider into Import-Packages for some reason, but it does pull in others:

```
Manifest-Version: 1.0
Bnd-LastModified: 1457105298481
Build-Jdk: 1.8.0_60
Built-By: atolbert
Bundle-Description: A driver for Apache Cassandra 1.2+ that works exclus
 ively with the Cassandra Query Language version 3 (CQL3) and Cassandra'
 s binary protocol.
Bundle-License: http://www.apache.org/licenses/LICENSE-2.0.txt
Bundle-ManifestVersion: 2
Bundle-Name: DataStax Java Driver for Apache Cassandra - Core
Bundle-SymbolicName: com.datastax.driver.core
Bundle-Version: 2.1.10.SNAPSHOT
Created-By: Apache Maven Bundle Plugin
Export-Package: com.datastax.driver.core;uses:="com.codahale.metrics,com
 .datastax.driver.core.exceptions,com.datastax.driver.core.policies,com.
 google.common.reflect,com.google.common.util.concurrent,io.netty.bootst
 rap,io.netty.buffer,io.netty.channel,io.netty.channel.socket,io.netty.h
 andler.codec,io.netty.util,javax.net.ssl,jnr.ffi,jnr.ffi.annotations,or                                                                                                              
 g.slf4j";version="2.1.10.SNAPSHOT",com.datastax.driver.core.exceptions;
 uses:="com.datastax.driver.core";version="2.1.10.SNAPSHOT",com.datastax
 .driver.core.policies;uses:="com.datastax.driver.core,com.google.common
 .base";version="2.1.10.SNAPSHOT",com.datastax.driver.core.querybuilder;
 uses:="com.datastax.driver.core,com.datastax.driver.core.policies";vers
 ion="2.1.10.SNAPSHOT",com.datastax.driver.core.schemabuilder;uses:="com
 .datastax.driver.core,com.google.common.base";version="2.1.10.SNAPSHOT"
 ,com.datastax.driver.core.utils;uses:="com.google.common.util.concurren
 t";version="2.1.10.SNAPSHOT"
Import-Package: com.google.common.base;version="[14.0,19)",com.google.co
 mmon.collect;version="[14.0,19)",com.google.common.primitives;version="
 [14.0,19)",com.google.common.reflect;version="[14.0,19)",com.google.com
 mon.util.concurrent;version="[14.0,19)",jnr.ffi;resolution:=optional;ve
 rsion="[2.0,3)",jnr.ffi.annotations;resolution:=optional;version="[2.0,
 3)",com.codahale.metrics;version="[3.0,4)",io.netty.bootstrap;version="
 [4.0,5)",io.netty.buffer;version="[4.0,5)",io.netty.channel;version="[4
 .0,5)",io.netty.channel.group;version="[4.0,5)",io.netty.channel.nio;ve
 rsion="[4.0,5)",io.netty.channel.socket;version="[4.0,5)",io.netty.chan
 nel.socket.nio;version="[4.0,5)",io.netty.handler.codec;version="[4.0,5
 )",io.netty.handler.ssl;version="[4.0,5)",io.netty.handler.timeout;vers
 ion="[4.0,5)",io.netty.util;version="[4.0,5)";resolution:=optional,io.n
 etty.util.concurrent;version="[4.0,5)";resolution:=optional,javax.namin
 g,javax.naming.directory,javax.net.ssl,net.jpountz.lz4;resolution:=opti
 onal,org.HdrHistogram;resolution:=optional,org.slf4j;version="[1.7,2)",
 org.xerial.snappy;version="[1.0,2)";resolution:=optional
Tool: Bnd-2.1.0.20130426-122213
```

```
[FelixStartLevel] INFO com.datastax.driver.core.Native - Could not load JNR LibC Library, native calls will not be available
java.lang.IllegalArgumentException: interface jnr.ffi.provider.LoadedLibrary is not visible from class loader
    at java.lang.reflect.Proxy$ProxyClassFactory.apply(Proxy.java:581)
    at java.lang.reflect.Proxy$ProxyClassFactory.apply(Proxy.java:557)
    at java.lang.reflect.WeakCache$Factory.get(WeakCache.java:230)
    at java.lang.reflect.WeakCache.get(WeakCache.java:127)
    at java.lang.reflect.Proxy.getProxyClass0(Proxy.java:419)
    at java.lang.reflect.Proxy.newProxyInstance(Proxy.java:719)
    at jnr.ffi.LibraryLoader.createErrorProxy(LibraryLoader.java:306)
    at jnr.ffi.LibraryLoader.load(LibraryLoader.java:301)
    at jnr.ffi.LibraryLoader.load(LibraryLoader.java:269)
    at com.datastax.driver.core.Native.<clinit>(Native.java:75)
    at com.datastax.driver.core.ClockFactory.newInstance(Clock.java:51)
    at com.datastax.driver.core.AbstractMonotonicTimestampGenerator.<init>(AbstractMonotonicTimestampGenerator.java:37)
    at com.datastax.driver.core.AtomicMonotonicTimestampGenerator.<init>(AtomicMonotonicTimestampGenerator.java:42)
    at com.datastax.driver.osgi.impl.Activator.start(Activator.java:44)
    at org.apache.felix.framework.util.SecureAction.startActivator(SecureAction.java:697)
    at org.apache.felix.framework.Felix.activateBundle(Felix.java:2223)
    at org.apache.felix.framework.Felix.startBundle(Felix.java:2141)
    at org.apache.felix.framework.Felix.setActiveStartLevel(Felix.java:1368)
    at org.apache.felix.framework.FrameworkStartLevelImpl.run(FrameworkStartLevelImpl.java:308)
    at java.lang.Thread.run(Thread.java:745)
```

tolbertam:
This is what I needed to do to support JNR in OSGi while making it optional at the same time, seems to work on my branch:

```
<![CDATA[com.google.common*;version="[14.0,19)",jnr.*;resolution:=optional,jnr.ffi.provider;resolution:=optional,jnr.ffi.provider.jffi;resolution:=optional,jnr.ffi.provider.converters;resolution:=optional,com.kenai.jffi;resolution:=optional,*]]>
```

Shaded Imports

```
<![CDATA[com.google.common.*;version="[14.0,19)",jnr.*;resolution:=optional,jnr.ffi.provider;resolution:=optional,jnr.ffi.provider.jffi;resolution:=optional,jnr.ffi.provider.converters;resolution:=optional,com.kenai.jffi;resolution:=optional,!io.netty.*,javax.security.cert,*]]>
```

Unfortunately all of those packages must be explicitly listed as wildcarding doesn't seem to work (it must have something to do with how things are loaded).

adutra:
> Unfortunately all of those packages must be explicitly listed as wildcarding doesn't seem to work (it must have something to do with how things are loaded).

That's unfortunate :( shouldn't we simply exclude JNR from the OSGi bundle right now (`!jnr.*`) and if/when everything is available as bundles, change it later?

tolbertam:
> That's unfortunate :(

Yeah, definitely.  Have tried a number of permutations and that's the only one I can seem to make work both ways.  

>  shouldn't we simply exclude JNR from the OSGi bundle right now (!jnr.*) and if/when everything is available as bundles, change it later?

I'm :+1: on that idea if it works.  Making it explicit that we don't import it should hopefully make it clear to users that we don't support using JNR with OSGi.

tolbertam:
It looks like !jnr.\* works, so I'm :+1: on that.

driver-core/src/main/java/com/datastax/driver/core/AbstractMonotonicTimestampGenerator.java
```diff
@@ -0,0 +1,86 @@
...
+        return current;
+    }
+
+    private void maybeLogWarning(long current, long last) {
+        if (LOGGER.isWarnEnabled()) {
+            long now = System.nanoTime();
+            long lastWarning = lastClockDriftWarning.get();
+            if (now > lastWarning + CLOCK_DRIFT_WARNING_INTERVAL && lastClockDriftWarning.compareAndSet(lastWarning, now)) {
+                if (current == last) {
+                    LOGGER.warn(
```
tolbertam: Have you noticed when running the OSGi tests that it occasionally logs:

```
[main] WARN com.datastax.driver.core.TimestampGenerator - Clock drift detected: same timestamp was generated twice (1457112131130000), returned timestamps will be artificially incremented to guarantee monotonicity. This message will only be logged once every second.
```

With a system clock this seems very likely since you can have concurrent access in the same millisecond.   I'm wondering if we should consider removing this particular warning and depending on the 'Clock skew' one instead.

adutra: That, or make `lastClockDriftWarning` static maybe?

adutra:
> It looks like !jnr.\* works, so I'm :+1: on that.

Done!

driver-core/src/main/java/com/datastax/driver/core/AbstractMonotonicTimestampGenerator.java
tolbertam: making it static does seem like a good idea.  I think this will still happen though, because clock.currentTimeMicros() for `SystemClock` is only going to have millisecond resolution and the likelihood of multiple calls in an ms seems very high.  The previous implementation only printed a warning if 999 entries were created for the same millisecond, so it was much less sensitive to this.

olim7t:
The clock drift warning condition is too aggressive: with a [simple load test](https://gist.github.com/olim7t/d86275240b6486cb61bc) I can reproduce "clock drift" almost every second, and even "clock skew" regularly. This is with the microsecond-resolution clock; with `System.currentTimeMillis()`, the warning appears as soon as two timestamps are generated in the same millisecond.
Granted, my example is contrived, but if I can reproduce it on my machine it's not unlikely that it could happen in a high-throughput environment.
I think a better test would be `last > current + delta`, with `delta` provided as a constructor parameter, with a sensible default (1 second comes to mind).

olim7t:
Pushed PoC for a "drift threshold" (some javadocs missing, will complete later).

adutra:
The threshold improves things, but I'm more and more thinking that issuing warnings might not be the best approach. 

We could also provide a new Metrics meter, e.g. a Timer that would record the N last abnormal deltas (i.e. deltas for which last - current <= 0, with maybe a threshold other than zero to filter out minor skews). This way users could query this metrics to find out if there were skews (timer.getCount() > 0) and what was the average delta for them (timer.getSnapshot().getMean()).

olim7t:
Metrics complicate things though, because we'd have to link timestamp generators to the `Cluster` instance in some way (right now they are standalone components). We shouldn't overthink this too much in advance. I think a good compromise would be to make `maybeLogWarning` protected so that it can be overridden, and give it a more meaningful name like `onClockDrift`.
What I like about the configurable threshold is that you can completely disable warnings by setting it to a negative value.

adutra:
>  I think a good compromise would be to make maybeLogWarning protected so that it can be overridden, and give it a more meaningful name like onClockDrift.

Sounds like a good compromise.

olim7t:
New commit based on previous discussion.

adutra:
Pushed a small commit that reorganizes the documentation among the class hierarchy; hope that's ok otherwise I can revert it.

tolbertam:
Looks great to me now :+1:

barhouum7:
But how to make the JNR optional for OSGi bundles only?? PLEASE HELP ME!

barhouum7:
But how to make the JNR optional for OSGi bundles only??

PR #592, opened 2016-02-15
JAVA-852: Ignore peers with null entries during auto-discovery.

newkek:
Also: 
- Modified the ScassandraCluster mock for the system.peers query to return the host_id column.

This is a tentative solution for this ticket in the sense that we completely ignore a host with null values in columns `host_id`/`rack`/`tokens`/all columns tried for address. It may be a too tough decision because we are not sure whether these fields being null mean that the node is having issues, or if it is only the Gossiper that failed to broadcast the node's state to the `ControlConnection`, or if 'null' is tolerated in some corner cases of configuration of a cluster.

However, the "short" test suite passes with this change, which would maybe mean that it is reasonable to say that these fields need to be not null.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
@@ -38,6 +38,8 @@ class ControlConnection implements Host.StateListener, Connection.Owner {
 
     private static final Logger logger = LoggerFactory.getLogger(ControlConnection.class);
 
+    private static final boolean EXTENDED_PEER_CHECK = SystemProperties.getBoolean("com.datastax.driver.EXTENDED_PEER_CHECK", true);
+
     private static final InetAddress bindAllAddress;
 
     static {
@@ -280,7 +282,7 @@ class ControlConnection implements Host.StateListener, Connection.Owner {
             // We need to refresh the node list again;
             // We want that because the token map was not properly initialized by the first call above,
             // since it requires the list of keyspaces to be loaded.
-            refreshNodeListAndTokenMap(connection, cluster, false, true);
+            refreshNodeListAndTokenMap(connection, cluster, false, false);
 
             return connection;
         } catch (BusyConnectionException e) {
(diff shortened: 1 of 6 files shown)


adutra:
Missing the entry in the changelog. It does seem scary to ignore all these hosts, but I guess we'll have to unleash the beast and see how much damage it does :)

newkek:
Pushed the changelog. I agree we could give it a try and see if that causes troubles.

tolbertam:
I am pretty certain that tightening the validation on peers won't cause any issues, but I still think it might be worth adding a System Property that allows you to override whether or not `isValidPeer` is called just in some unanticipated case where a user sees null racks or host_ids in their peer table but still wants to connect.  Something like `-Dcom.datastax.driver.VALIDATE_PEERS` with the default being true.  What do you think? It's possible that i'm overworrying ;).

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
```diff
@@ -647,6 +651,19 @@ private static void refreshNodeListAndTokenMap(Connection connection, Cluster.Ma
             cluster.metadata.rebuildTokenMap(partitioner, tokenMap);
     }
 
+    private static boolean isValidPeer(Row peerRow, InetSocketAddress addr) {
+        return peerRow.getColumnDefinitions().contains("host_id")
+                && peerRow.getColumnDefinitions().contains("rack")
+                && peerRow.getColumnDefinitions().contains("tokens")
+                && addr != null
```
adutra: Small nit: the `addr` parameter doesn't seem to belong here; imo the code would be a bit more readable if this method only cared about `peerRow`; the if block above could then be `if (addr == null || !isValidPeer(row))`

olim7t:
I've tried starting Cassandra with a custom snitch that returns a null datacenter or rack, it fails in both cases. So it seems quite safe to assume that these should never be null in system tables.
Nevertheless I like Andy's idea of a property to disable the extended checks, I'll add that.

olim7t:
Also checked `host_id`, it's [generated](https://github.com/apache/cassandra/blob/cassandra-3.4/src/java/org/apache/cassandra/db/SystemKeyspace.java#L1028-L1045) the first time the server uses it, and that happens [before it joins the ring](https://github.com/apache/cassandra/blob/cassandra-3.4/src/java/org/apache/cassandra/service/StorageService.java#L817). So a peer should definitely not broadcast a null value.

adutra:
lgtm :+1:

tolbertam:
have one more test change pending to validate each column being null.

newkek:
lgtm, thanks for the improvements.

tolbertam:
looks good to me too! :+1:

PR #593, opened 2016-02-15
JAVA-444: Add Java process information to UUIDs.makeNode() hash.

adutra:
This is based on #591 and supersedes #521.


adutra:
Benchmark under Windows getpid() vs JMX:

```
Benchmark               Mode  Cnt      Score     Error  Units
Benchmark.testJMX  avgt   20  20070.951 ± 406.583  ns/op
Benchmark.testJNR  avgt   20     66.059 ±   1.798  ns/op
```

adutra:
Improved documentation for Native and UUIDs classes. Squashed commits again.

driver-core/src/main/java/com/datastax/driver/core/utils/UUIDs.java
```diff
@@ -27,17 +31,42 @@
...
- * Utility methods to work with UUID and most specifically with time-based ones
- * (version 1).
+ * Utility methods to help working with UUIDs, and more specifically, with time-based UUIDs
+ * (also known as Version 1 UUIDs).
+ * <h3>Notes on the algorithm used to generate time-based UUIDs</h3>
+ * The algorithm follows roughly the description in RFC-4122, but with the following adaptations:
+ * <ol>
+ * <li>Since Java does not provide direct access to the host's MAC address, that information
+ * is replaced with a digest of all IP addresses available on the host;</li>
+ * <li>The process ID (PID) isn't easily available to Java neither, so it is determined by one of the
```
olim7t: isn't available ... _either_

driver-core/src/main/java/com/datastax/driver/core/utils/UUIDs.java
```diff
@@ -108,6 +138,43 @@ private static long makeNode() {
         }
     }
 
+    private static String getProcessPiece() {
+        Integer pid = null;
+        String pidProperty = System.getProperty(PID_SYSTEM_PROPERTY);
+        if (pidProperty != null) {
+            try {
+                pid = Integer.parseInt(pidProperty);
+                LOGGER.debug("PID obtained from System property {}: {}", PID_SYSTEM_PROPERTY, pid);
```
olim7t: I would log that at level INFO, same at lines 155 and 164.

olim7t:
Looks good to me, apart from 2 minor remarks and the "optional" discussion.

olim7t:
Rebased on current 2.1 (2708bd7)

adutra:
Rebased again on top of current #591.

driver-core/src/main/java/com/datastax/driver/core/utils/UUIDs.java
```diff
@@ -108,6 +138,43 @@ private static long makeNode() {
...
+            try {
+                pid = Native.processId();
+                LOGGER.info("PID obtained through native call to getpid(): {}", pid);
+            } catch (Exception e) {
+                LOGGER.warn("Native call to getpid() failed", e);
+            }
+        }
+        if (pid == null) {
+            try {
+                String pidJmx = ManagementFactory.getRuntimeMXBean().getName().split("@")[0];
```
tolbertam, line 163: Really hard to test this as is as I need a non-POSIX system :D, I commented out the code above and it seems to work :+1:

driver-core/src/test/java/com/datastax/driver/core/MockClocks.java
```diff
@@ -21,7 +21,7 @@
         int calls;
 
         @Override
-        public long currentTime() {
+        public long currentTimeMicros() {
```
tolbertam, line 24: We could probably remove the 'public' modifiers from these since the parent class is not public right?  (we'd probably need to add from clirr exceptions though, doesn't seem important).

adutra, line 24: Unfortunately that would not work here, as this method is inherited from Clock interface, so it needs to be public.

adutra:
Rebased on top of current #591 to benefit from OSGi fixes.
Changed the loading process of JNR POSIX library in Native class to use reflection; otherwise it would fail in an unexpected manner when this library is not present in the classpath.

tolbertam:
This looks good to me :+1:, think it can be safely merged after #591 is settled/merged (or if you just want to merge this containing #591).

tolbertam:
:+1: on merging after rebased on 2.1 after #591 is merged.

PR #595, opened 2016-02-17
JAVA-630: Don't process DOWN events for nodes that have active connec…

olim7t:
…tions.

Tested manually:
- 3-node ccm cluster on localhost
- start a Java program with a host listener that logs events
- block all gossip to simulate control host isolation: `sudo ./tcpkill -i lo0 dst port 7000`

I tried various combinations:
- initialized cluster with no session: should always process down events (no pooled connections)
- with connected session: should ignore down events (hosts have working pooled connections)
- use whitelist policy to ignore one node: should always process down events for that node

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -2625,8 +2625,16 @@ public class Cluster implements Closeable {
                             // But it is unlikely, and don't have too much consequence since we'll try reconnecting
                             // right away, so we favor the detection to make the Host.isUp method more reliable.
                             Host downHost = metadata.getHost(address);
-                            if (downHost != null)
-                                futures.add(execute(hostDown(downHost)));
+                            if (downHost != null) {
+                                // Only process DOWN events if we have no active connections to the host . Otherwise, we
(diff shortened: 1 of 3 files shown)


driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -2625,8 +2625,16 @@ public String toString() {
...
                             // right away, so we favor the detection to make the Host.isUp method more reliable.
                             Host downHost = metadata.getHost(address);
-                            if (downHost != null)
-                                futures.add(execute(hostDown(downHost)));
+                            if (downHost != null) {
+                                // Only process DOWN events if we have no active connections to the host . Otherwise, we
+                                // wait for the connections to fail. This is to prevent against a bad control host
+                                // aggressively marking DOWN all of its peers.
+                                if (downHost.convictionPolicy.hasActiveConnections()) {
+                                    logger.debug("Ignoring down event on {} because it still has active connections");
```
tolbertam: downHost should be added as an argument here so it gets logged.

olim7t: Done in rebased commit, thanks

PR #599, opened 2016-02-19
JAVA-923: Position idempotent flag on object mapper queries.

adutra:


olim7t:
Maybe we could also position the `idempotent` flag on the queries generated for `Mapper`. I think they are all idempotent.

adutra:
> Maybe we could also position the idempotent flag on the queries generated for Mapper. I think they are all idempotent.

I like the idea, but can we guarantee idempotence if the entity contains a computed field, e.g.

``` java
@Computed(value = "now()")
UUID lastUpdated;
```

This kind of thing might not make sense, but since people could write it, the SAVE query for such an entity would not be strictly idempotent.

tolbertam:
Looks good to me!  :+1: Added a couple more tests and some test doc.

olim7t:
> I like the idea, but can we guarantee idempotence if the entity contains a computed field

Computed fields are not included in save queries.

adutra:
> Computed fields are not included in save queries.

Alright, went ahead and flagged all mapper-generated queries as idempotent.

PR #600, opened 2016-02-19
JAVA-831: Mapper can't load an entity where the PK is a UDT.

adutra:


driver-mapping/src/main/java/com/datastax/driver/mapping/ReflectionMapper.java
```diff
@@ -162,6 +162,12 @@ public void setValue(Object entity, Object value) {
 
             super.setValue(entity, udtMapper.toEntity((udtValue)));
         }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        Object toSerializableValue(Object value) {
+            return value == null ? null : udtMapper.toUDT((U) value);
+        }
```
olim7t, line 167: We also need to override that in `NestedUDTMapper` and `EnumMapper` (using an enum as partition key, though unlikely, is possible).

adutra, line 167: Good point, I pushed a commit that overrides the method in both `NestedUDTMapper` and `EnumMapper`. I think that `NestedUDTMapper.toSerializableValue()` will never be invoked really as collections cannot be part of the primary key, but the override doesn't do any harm.
And by the way, your comment made me realize that delete queries should be fixed too - which I did in the same commit.

olim7t:
I moved the call to `toSerializableValue` for the get method to the parent class.
LGTM :+1:

PR #601, opened 2016-02-19
JAVA-974: Validate accessor parameter types against bound statement.

adutra:


olim7t:
:+1: 
The `dataType` in `ParamMapper` wasn't used anymore, I pushed an extra commit to simplify the code.

PR #603, opened 2016-02-19
JAVA-1070: The Mapper should not prepare queries synchronously.

adutra:


driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
```diff
@@ -160,7 +167,11 @@ public MappingManager getManager() {
      * @return a query that saves {@code entity} (based on it's defined mapping).
      */
     public Statement saveQuery(T entity) {
```
tolbertam, line 174: It might be nice to have a public version of `saveQueryAsync` (for get and delete too) for those who want to write a completely async application (and using Statements returned here instead of calling `saveAsync`).   Will open on a follow on ticket for that.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperAsyncTest.java
```diff
@@ -0,0 +1,184 @@
...
+
+import static com.datastax.driver.core.TestUtils.CREATE_KEYSPACE_SIMPLE_FORMAT;
+import static java.util.concurrent.TimeUnit.MINUTES;
+import static org.testng.Assert.fail;
+
+/**
+ * Basic tests for the mapping module.
+ */
+@SuppressWarnings("unused")
+public class MapperAsyncTest extends CCMTestsSupport {
```
tolbertam, line 44: For whatever reason this isn't running in CI...looking into it.

tolbertam, line 44: It looks like it is running, just not being collected for whatever reason.

tolbertam, line 44: I get it now, the tests aren't being recorded because our second run with the 'isolated' profile is just running unit tests which blows away our reports.  I'll fix that, looks like it broke once I introduced that, my bad!

driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
tolbertam, line 174: [JAVA-1093](https://datastax-oss.atlassian.net/browse/JAVA-1093), although this could be done in this PR too if we decide its worth doing now.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperAsyncTest.java
tolbertam, line 44: #609

tolbertam:
I went ahead and rebased this on 2.1 since there was only a conflict in README.md (hope that's ok).  Added a small commit to add some test doc.  I was able to reproduce the issue on 2.1 with the provided test, the workaround seems to be to use a separate executor in the Future callbacks.   Looks good to me :+1:

driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
```diff
@@ -104,29 +103,37 @@ Session session() {
...
-                    String queryString = type.makePreparedQueryString(tableMetadata, mapper, manager, columns, options.values());
-                    logger.debug("Preparing query {}", queryString);
-                    stmt = session().prepare(queryString);
-                    Map<MapperQueryKey, PreparedStatement> newQueries = new HashMap<MapperQueryKey, PreparedStatement>(preparedQueries);
-                    newQueries.put(pqk, stmt);
-                    preparedQueries = newQueries;
+            String queryString = type.makePreparedQueryString(tableMetadata, mapper, manager, columns, options.values());
+            logger.debug("Preparing query {}", queryString);
+            final SettableFuture<PreparedStatement> future = SettableFuture.create();
+            Futures.addCallback(session().prepareAsync(queryString), new FutureCallback<PreparedStatement>() {
```
olim7t: It's still possible for two threads to prepare the same statement concurrently. It's not that bad, but it prints a warning log.
We can avoid it with a couple of changes:

``` java
// store futures instead of plain statements
private final ConcurrentMap<MapperQueryKey, ListenableFuture<PreparedStatement>> preparedQueries
...
SettableFuture<PreparedStatement> future = SettableFuture.create();
ListenableFuture<PreparedStatment> old = preparedQueries.putIfAbsent(pqk, future);
if (old != null) return old;
else {
   // call prepareAsync, complete `future` in the callback
}
```

tolbertam: Is this any different than someone calling session.prepare concurrently with the same query?   I think the difference is that there is no implication that the user is creating a prepared statement when they use saveQuery, so I agree that it makes sense that if we can mitigate this we should.

tolbertam: Was just thinking, is there any reason why we couldn't do the same thing in `Cluster.Manager#preparedQueries`?  This seems like a nice optimization.

olim7t: > Is this any different than someone calling session.prepare concurrently with the same query? I think the difference is that there is no implication that the user is creating a prepared statement when they use saveQuery, so I agree that it makes sense that if we can mitigate this we should.

Exactly. We don't want warnings randomly popping up when people make normal use of the mapper.

> Was just thinking, is there any reason why we couldn't do the same thing in Cluster.Manager#preparedQueries? This seems like a nice optimization.

That's a great idea, could you create a ticket?

tolbertam: :+1: , created [JAVA-1096](https://datastax-oss.atlassian.net/browse/JAVA-1096) for this and referenced this discussion thread.

adutra: Excellent ideas! I updated the code with @olim7t suggestion.

driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
adutra, line 174: Given the reticences expressed in that Jira ticket, I would suggest that we hold off on doing this in this PR.

olim7t:
Added a commit with a small refactoring that makes the code more readable IMO.

PR #604, opened 2016-02-23
JAVA-1083: Document server-side RPC address configuration.

olim7t:
No changelog since this is a simple doc update.


tolbertam:
very nice, :+1:

PR #606, opened 2016-02-24
JAVA-1072: Defunct connection when setKeyspaceAsync call fails.

olim7t:


adutra:
:+1:

tolbertam:
This looks good to me, I will have to do some targeted testing to reproduce this particular bug, but the fix makes complete sense.  I think we should go ahead and merge it and I'll incorporate this into stress regression for 2.1.10.

PR #607, opened 2016-02-24
JAVA-1021: Improve error message when connect() is called with an inv…

olim7t:
…alid keyspace name.

Contrary to what I said in the ticket description, I simply made the error message more explicit (I didn't want to parse the keyspace name client-side).


tolbertam:
Looks good to me :+1:.  Added additional test for connectAsync.  Just to clarify, the [jira ticket](https://datastax-oss.atlassian.net/browse/JAVA-1021) alludes to checking the keyspace before sending the request, but instead this decorates the error message to indicate to check the keyspace.  I think I remember you mentioned that this was the route to go (let the server error out, not preprocess on the client), but wanted to make sure.

olim7t:
Yes, I'll update the ticket's description.

PR #608, opened 2016-02-24
JAVA-879: Mapper.map() accepts mapper-generated and user queries.

newkek:
Also :
- _ExecutionInfo_ exposes a new method `getStatement()`. The statement is also now consistently put in the _ExecutionInfo_, and not only for queries with multiple pages.
- `Mapper.map()` accepts user queries and Mapper-generated queries.
- `mapAliased()` becomes `mapAliasedUnsafe()` and can allow users to use ResultSet produced from their own queries, containing their aliases. But the `mapAliasedUnsafe()` requires the order of the columns in the query, to be matching the orders of the fields in the mapped class. _This change causes a clirr test to fail because of backward compat_.
- Fields and methods in some classes, declared as package-local, were public. Some clirr tests fails because _ColumnMapper_ became package-local rather than public, but _ColumnMapper_ is never used in the public API, neither as a return object nor as a function parameter. So it looks like a false positive.


olim7t:
> mapAliased() becomes mapAliasedUnsafe() 

We could keep the same name but simply deprecate it, since `map` handles all cases now. This way we avoid a breaking change.

> Fields and methods in some classes, declared as package-local, were public

Yes, this is common in some part of the code. It's fine to mark it as false positives.

newkek:
Done for the `@Deprecated`. I don't know how to ignore the clirr errors.

olim7t:
You need to add entries to `clirr-ignores.xml` at the root of the project.

tolbertam:
Went ahead and added the clirr ignore (hope that's ok).

tolbertam:
This looks good to me, looks like it will definitely solve the problem as described in the issue. :+1:

olim7t:
Rebased on 2.1, extracted unrelated visibility changes to a separate commit, added a commit to avoid creating too many ExecutionInfo instances.
The fact that we rely on `DefaultPreparedStatement` being public bothers me, IMO this was an oversight and that class should have been private.

olim7t:
LGTM.

PR #609, opened 2016-02-24
Define isolated profile that disables tests at parent level.

tolbertam:
The 'isolated' profile was introduced in driver-core a while back to run tests in their own JVM fork when needed.  This causes problems however when another submodule like driver-mapping doesn't have an isolated profile in which it will run unit tests and overwrite any existing surefire report.  This change disables tests when running with isolated group unless you override it explicitly in a module that desires to use it.

### DIFF
--- a/driver-core/pom.xml
+++ b/driver-core/pom.xml
@@ -352,23 +352,10 @@
             <artifactId>maven-surefire-plugin</artifactId>
             <version>2.16</version>
             <configuration>
-              <forkCount>1</forkCount>
-              <reuseForks>false</reuseForks>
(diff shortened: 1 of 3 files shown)


tolbertam:
Looks to be working good, although I've found another issue while debugging this in that some tests are being skipped that shouldn't.  I think this is good to merge, and i will pursue the other issue.

adutra:
:+1:  this will have to be rebased on #603 before merge right?

tolbertam:
Thanks! It shouldn't need to be rebased as it is more of a global fix (we have existing tests that are being ran but not recorded because of this).

PR #610, opened 2016-02-25
JAVA-1040: SimpleStatement parameters support in QueryLogger

tommystendahl:


driver-core/src/main/java/com/datastax/driver/core/QueryLogger.java
```diff
@@ -746,6 +760,61 @@ protected String parameterValueAsString(ColumnDefinitions.Definition definition,
         return valueStr;
     }
 
+    protected int appendParameters(SimpleStatement statement, StringBuilder buffer, int remaining) {
```
tolbertam, line 763: There is quite a bit of duplication of code between the appendParameters/parameterValueAsString methods for the SimpleStatement and BoundStatement implementations.  However, they are probably different enough to justify it being the way it is.

tolbertam:
Looks good to me :+1:

driver-core/src/test/java/com/datastax/driver/core/QueryLoggerTest.java
```diff
@@ -610,11 +723,32 @@ public void should_truncate_blob_parameter_when_max_length_exceeded() throws Exc
                 .contains("Query completed normally")
                 .contains(ipOfNode(1))
                 .contains("c_blob:0x0102" + TRUNCATED_OUTPUT)
-                .doesNotContain("123456");
+                .doesNotContain("0x010203");
```
adutra, line 726: Thanks for spotting this :)

driver-core/src/main/java/com/datastax/driver/core/QueryLogger.java
olim7t, line 763: Agreed

olim7t:
Merged, thanks for your contribution!

PR #611, opened 2016-02-25
JAVA-1040: SimpleStatement parameters support in QueryLogger

tommystendahl:


driver-core/src/main/java/com/datastax/driver/core/SimpleStatement.java
```diff
@@ -193,6 +194,19 @@ public Object getObject(int i) {
     }
 
     /**
+     * Returns the named values as a {@code Map<String, Object>}.
+     *
+     * @return the named values of this statement.
+     * @throws IllegalStateException     if this statement does not have named values.
+     */
+    public Map<String, Object> getObjectMap() {
```
adutra: Sorry if I'm arriving a bit late for the party (since #610 is already merged).
1. Couldn't we rename this method to `getNamedValues()`?
2. Since we're already augmenting the API, and for the sake of symmetry, couldn't we also provide a `Object[] getValues()` method for positional parameters? I recall users asking for this frequently, precisely for debugging purposes.
3. While at it, since we have `Object getObject(int)` can't we have a symmetric  `Object getObject(String)`?

I know all of this should have been addressed in JAVA-1037, but it sounds like a good opportunity here to add those missing methods.

tommystendahl: No problem. What you say make good sense, I can take care of this tomorrow and update the PR.

adutra: Thanks @tommystendahl :) ! Though let's wait for more input: @tolbertam @olim7t @newkek your thoughts?

tommystendahl: I have done the changes in a local branch, couldn't help my self:-). I think it looks good, let me know if you want me to update the PR.

adutra: Go ahead then, let's see how it looks :) btw sorry for the delay but we've been focusing on 2.1.10 lately.

olim7t:
I'm not a big fan of the new `getValues()` and `getNamedValues()` because they collide with the existing ones. So we get overloads that do two different things:
- when called with parameters, return serialized values
- when called with no parameters, return the original objects :/

In the spirit of `GettableData`, we should have one "metadata" method that gives you an idea of which values are present, and then getter methods to retrieve the values. For positional values we already have `valuesCount()` and `getObject(int)` (and I don't think we need to add a method that returns an array). So for named values I suggest `Set<String> getValueNames()` and `getObject(String)`.

adutra:
> So for named values I suggest `Set<String> getValueNames()` and `getObject(String)`.

Good point, I agree to follow your suggestion instead.

tommystendahl:
Ok, I will update this according to your comment.

olim7t:
:+1: 
I merged the PR manually because it needs to go to the 3.0.x branch. I squashed your 3 commits into b8a961b (you're still the author).

PR #612, opened 2016-02-26
JAVA-1100: Exception when connecting with shaded java driver in OSGI

emolsson:


tolbertam:
Thank you for your contribution!  [We were explicitly working around this in our tests](https://github.com/datastax/java-driver/blob/2.1/driver-examples/osgi/src/test/java/com/datastax/driver/osgi/MailboxServiceIT.java#L84), it didn't register with us at the time that this was something we could fix in our build.   I think no PR for 3.0 is needed, we will take care of merging this upstream when we merge 2.1 into 3.0

tolbertam:
Confirmed, we can now remove `bootDelegationPackages("javax.security.*")` from our test and our shaded configuration passes without it, where previously I could not.  :+1:    Would you mind removing that line in that test?  Otherwise I can do that post merge.

emolsson:
Great that it solved another problem as well! I can remove it in the patch directly.

tolbertam:
Looks great!  :+1:

adutra:
Thanks for solving 2 issues with one PR! :+1:

olim7t:
Merged, thanks.

PR #613, opened 2016-02-26
JAVA-1089: Set LWT made from BuiltStatements to non-idempotent.

newkek:


newkek:
I think that covers all the LWT possible I could think of.

tolbertam:
Would this also apply to `Delete#ifExists`?

tolbertam:
Not completely related, but we should also override `isIdempotent()` on `SchemaStatement` to always return false?  I don't think a DDL change would ever be considered idempotent.

tolbertam:
Another unrelated,  shouldn't `Batch` consider it's inner statements in `isIdempotent`?  For example:

``` java
batch().add(update("foo").with(set("a", "b")).where(eq("a", "c")).ifExists());
```

the inner update statement is nonIdempotent but the batch is considered Idempotent.  It looks like `BatchStatement` doesn't do this either, [JAVA-1101](https://datastax-oss.atlassian.net/browse/JAVA-1101).

newkek:
> Would this also apply to Delete#ifExists?

Good catch, in fact there's `ifExists()` and `onlyIf()` in `Delete`, that I forgot, thanks.

newkek:
Summary of cases covered : 
```
UPDATE .... IF EXISTS
UPDATE .... IF ....
DELETE .... IF EXISTS
DELETE .... IF ....
INSERT INTO .... IF NOT EXISTS
```

newkek:
> Not completely related, but we should also override isIdempotent() on SchemaStatement to always return false? I don't think a DDL change would ever be considered idempotent.

I think that's a good idea to be conservative on that point, however the inferred default idempotence is already to `false`, so these statements would still be considered as non-idempotent. I'll make the change, in case the default idempotence one day changes.

tolbertam:
> I think that's a good idea to be conservative on that point

:+1: on that!   PR looks good to me now :)

newkek:
Summary of cases covered :

```
UPDATE .... IF EXISTS
UPDATE .... IF ....
DELETE .... IF EXISTS
DELETE .... IF ....
INSERT INTO .... IF NOT EXISTS
```

newkek:
Summary of additional cases covered by the ticket :

```
DELETE .... WHERE column OP FCALL()
UPDATE .... WHERE column OP FCALL()
DELETE list[1] FROM .... // non-idempotent because deletion of a list element

// LWT
UPDATE .... IF EXISTS
UPDATE .... IF ....
DELETE .... IF EXISTS
DELETE .... IF ....
INSERT INTO .... IF NOT EXISTS
```

newkek:
@tolbertam I reverted the change for `SchemaStatement`, instead I created JAVA-1105, where we will be able to study more the situation. 
And also because hard coding it like that by overriding the `isIdempotent()` method was making it impossible to change it afterward, since the method was always returning `false`. Where we would rather have the driver set it to `false` at the statement's creation, but still allowing users to override it explicitly if they want to, wdyt? 
I.e. before we implement JAVA-1105, they could set idempotent to true if the DDL contains a `IF NOT EXISTS`.

tolbertam:
That sounds sensible to me, i agree there really isn't any reason to rush on it for schema statements :+1:

driver-core/src/main/java/com/datastax/driver/core/querybuilder/Utils.java
```diff
@@ -307,6 +307,9 @@ static boolean isIdempotent(Object value) {
                 if (!isIdempotent(entry.getKey()) || !isIdempotent(entry.getValue()))
                     return false;
             }
+        } else if (value instanceof Clause){
+            Object clauseValue = ((Clause) value).firstValue();
+            return !(clauseValue instanceof FCall);
```
adutra: This should be: `return isIdempotent(clauseValue)`.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/Delete.java
```diff
@@ -39,6 +39,15 @@
         this.where = new Where(this);
         this.usings = new Options(this);
         this.conditions = new Conditions(this);
+
+        // This is for JAVA-1089, if the query deletes an element in a list, the statement should be non-idempotent.
+        for (Selector sel : columns) {
+            if (sel instanceof CollectionElementSelector) {
+                if (((CollectionElementSelector) sel).key instanceof Integer) {
```
adutra: I believe this is wrong: it would also return true for maps with integer keys. We will need to create separate subclasses or find another way to distinguish list element deletions from other cases.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/Delete.java
```diff
@@ -114,6 +132,7 @@ public Where where() {
      * @return the conditions of this query to which more conditions can be added.
      */
     public Conditions onlyIf(Clause condition) {
```
adutra, line 133: Could you add in this class the same javadoc warnings about idempotence as in `Insert` and `Update`?

driver-core/src/main/java/com/datastax/driver/core/querybuilder/Delete.java
newkek: Indeed it does not work well, will correct that.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/Delete.java
```diff
@@ -466,6 +495,36 @@ boolean containsBindMarker() {
...
     }
 
+    private static class ListElementSelector extends CollectionElementSelector {
+
+        ListElementSelector(String columnName, Object key) {
+            super(columnName, key);
+        }
+    }
+
+    private static class SetElementSelector extends CollectionElementSelector {
```
newkek, line 502: Set and Map `ElementSelector` are not useful now but it looked cleaner to add them imo and they could be useful for the future.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/Delete.java
```diff
@@ -33,21 +35,34 @@
     private boolean ifExists;
 
     Delete(String keyspace, String table, List<Selector> columns) {
-        super(keyspace);
-        this.table = table;
```
newkek, line 37: Tried to avoid all those code duplications in the BuiltStatement class and it's children, not sure if it was worth it but it's done.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/Delete.java
newkek, line 133: Done.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/Delete.java
adutra, line 502: :+1:

driver-core/src/test/java/com/datastax/driver/core/StatementIdempotenceTest.java
```diff
@@ -137,6 +137,23 @@ public void should_override_inferred_value_when_manually_set_on_built_statement(
...
 
+                delete().from("foo").where(lt("k", fcall("now"))),
+                delete().from("foo").where(lt("k", now())),
+                update("foo").where(eq("k", fcall("now"))),
+                delete().listElt("a", 1).from("test_coll"),
+
+                // LWT
+                update("foo").where(eq("is", "charlie?")).ifExists(),
+                update("foo").where(eq("good", "drivers")).onlyIf(contains("developers", "datastax")),
+                update("foo").onlyIf().and(contains("developers", "datastax")).where(eq("good", "drivers")),
```
olim7t, line 148: This is a weird corner case, but the DSL allows it so we have to support it.

PR #614, opened 2016-03-02
JAVA-1109: Document SSLOptions changes in upgrade guide.

tolbertam:
Also clarify pros/cons of using netty-tcnative.

/cc @al3xandru

### DIFF
--- a/manual/ssl/README.md
+++ b/manual/ssl/README.md
@@ -125,7 +125,7 @@ configuration on the `SSLEngine` (for example hostname verification).
 #### Netty
 
 [NettySSLOptions] allows you to use Netty's `SslContext` instead of
-the JDK directly. The advantage is that Netty can use OpenSSL directly,
+the JDK's. The advantage is that Netty can use OpenSSL directly,
 which provides better performance and generates less garbage.  A disadvantage of
 using the OpenSSL provider is that it requires platform-specific dependencies,
 unlike the JDK provider.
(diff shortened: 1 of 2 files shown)


manual/ssl/README.md
```diff
@@ -125,8 +125,11 @@ configuration on the `SSLEngine` (for example hostname verification).
 #### Netty
 
 [NettySSLOptions] allows you to use Netty's `SslContext` instead of
-the JDK directly. The advantage is that Netty can use OpenSSL if
-available, which provides better performance.
+the JDK directly. The advantage is that Netty can use OpenSSL directly,
```
adutra: I'm not sure but shouldn't this be "instead of the JDK**'s**"?

manual/ssl/README.md
```diff
@@ -149,7 +152,8 @@ Netty-tcnative provides the native integration with OpenSSL. Follow
 [these instructions](http://netty.io/wiki/forked-tomcat-native.html) to
 add it to your dependencies.
 
-Note that using netty-tcnative requires JDK 1.7 or above.
+Note that using netty-tcnative requires JDK 1.7 or above and requires the
+presence of OpenSSL on the system.  It will not fall back to the JDK implementation.
```
adutra: maybe clarify the last sentence: "It will _not_ fall back to the default JDK implementation, therefore you should only use it if you are sure that OpenSSL is installed on your system."

adutra:
:+1:

tolbertam:
Thanks for the feedback, addressed and ready to go after a squash :)

tolbertam:
Will reopen against 3.0.x

PR #615, opened 2016-03-03
JAVA-1108 add param to stress app, to set consistency level

alramlechner:
see https://datastax-oss.atlassian.net/browse/JAVA-1108; added functionality to set consistency level to stress app.


olim7t:
Thanks, we'll look at this for the 3.0.1 release.

tolbertam:
Looks like this works, thanks 👍 .  Would you mind squashing your 4 commits into 1?  

I do wonder if we should consider removing this example though, since cassandra-stress uses the java driver and is more complete/maintained.

Note that it looks like this module will be moved to `driver-tests` as part of #643

tolbertam:
Thanks for squashing 👍 , looks good to me.

olim7t:
Merged as eec8b31.
Sorry for not merging this PR directly, but this needed to go to the 3.0.x branch. The commit is still properly attributed to you.

PR #616, opened 2016-03-07
JAVA-905: Detect if many null fields are being written from Mapped objects.

adutra:
This solution uses `SlidingTimeWindowReservoir` which, according to the docs, can have a huge memory footprint. If we go forward, we need to figure out a way to prevent this reservoir from taking up the whole heap memory.


adutra:
Updated the pull request with a custom implementation of a sliding time window counter. Rebased on top of current 2.1.

PR #617, opened 2016-03-07
Fix issue where tests were being skipped in CI.

tolbertam:
Because we have `@Before*` methods on CCMTestsSupport, any test class
extending this class that had a @CassandraVersion annotation which
causes tests to be skipped for that class would also cause subsequent
tests to be skipped, even if they lacked the @CassandraVersion
annotation.

This happens because TestNG takes the declaring Class of the method
being skipped instead of the Class implementation being evaluated.  To
bypass this, setting configfailurepolicy to true causes TestNG to
evaluate the class implementation instead of the declaring class of the
@Before-annotated method.

Documentation on configurefailurepolicy [here](http://testng.org/doc/documentation-main.html#running-testng).

An alternative to this is to use [annotation transformers](http://testng.org/doc/documentation-main.html#annotationtransformers) and set `@Test`'s `enabled` field to `false`.  I tested this out by having TestListener implement `IAnnotationTransformer` and applied the `@CassandraVersion` logic in `transform`.  That worked well, but the downside there is that those tests no longer show up as skipped, instead they don't show up at all (which may be ok?).

A good example demonstrating the issue:

```
mvn test -Pshort -Dtest=MapperDefaultKeyspaceTest,MapperEnumTest -Dcassandra.version=1.2.19
```

`MapperDefaultKeyspaceTest` has a `@CassandraVersion` annotation which causes it to be skipped, but also `MapperEnumTest` is skipped where it shouldn't be.


driver-core/pom.xml
```diff
@@ -334,6 +334,13 @@
               <classpathDependencyExcludes>
                 <classpathDependencyExcludes>io.netty:netty-transport-native-epoll</classpathDependencyExcludes>
               </classpathDependencyExcludes>
+              <properties>
```
tolbertam: Not so great that this needs to be repeated, we could follow the same pattern I recently introduced for the `isolated` profile where the profile is declared in the root pom and the child modules only override the profile where needed.  Can fix that now or in a later PR.

tolbertam:
assigning to 2.1.10 as this would be nice to have.

tolbertam:
It looks like we have some mapper tests that are running now that were previously skipped that are failing on 2.0 and 2.1 (compat issues with C\* versions),  I'll fix those tomorrow.

driver-core/pom.xml
adutra: Also the profiles are almost identical, except for the groups property. In driver-extras I tried to factorize all the settings out of the profiles into the main declaration, and only set the `test.groups` property in each profile.

tolbertam: Good point, I'll follow that example in driver-extras as I think it will work very well.

tolbertam: I followed the pattern in driver-extras and I also did some refactoring to declare all the profiles in the parent pom and have the child profiles extend from them.  It seems to work very well and removes a lot of duplication in our poms, but need to give CI a spin to see if it behaves as we'd like.

tolbertam:
Looks like this is working good.   Squashed and rebased and this should be ready to go pending review.

adutra:
lgtm :+1:

adutra:
Hmm looks like we could improve things for the failsafe plugin as well. Pushed a branch [here](https://github.com/datastax/java-driver/tree/fix-skipped-tests-issue-adu), we'll see how things go.

adutra:
@tolbertam looks like CI is happy with my additional commits, would you be OK to integrate them into this PR?

tolbertam:
Cool, so it was the surefire version giving us problems previously?  Good to see that going back to 2.16 fixed it but weird that newer versions behave that way (can save that research for another day).  I'll cherry-pick your commits over and squash :+1:

tolbertam:
Squashed and rebased your commits on this PR.  After some discussion we've decided to take the commits from this branch into #585, so closing this one.

PR #621, opened 2016-03-10
Fix checkSchemaAgreement bug introduced in JAVA-852

tolbertam:
There was a small bug that I didn't catch until after #592 was merged in the `isValidPeer` check in `checkSchemaAgreement` requires the presence of columns that weren't returned in `SELECT_SCHEMA_PEERS`.  Updated to use `SELECT_PEERS` as we want to only valid peers are considered.

Also moved a test in ControlConnectionTest into a separate class so it can be run in isolated mode.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
@@ -427,10 +427,9 @@ class ControlConnection implements Host.StateListener, Connection.Owner {
         InetAddress peer = peersRow.getInet("peer");
         InetAddress addr = peersRow.getInet("rpc_address");
 
-        // We've already called isValid on the row, which checks this
-        assert addr != null;
-
-        if (peer.equals(connectedHost.getAddress()) || addr.equals(connectedHost.getAddress())) {
+        if (addr == null) {
+            return null;
+        } else if (peer.equals(connectedHost.getAddress()) || addr.equals(connectedHost.getAddress())) {
             // Some DSE versions were inserting a line for the local node in peers (with mostly null values). This has been fixed, but if we
             // detect that's the case, ignore it as it's not really a big deal.
             logger.debug("System.peers on node {} has a line for itself. This is not normal but is a known problem of some DSE version. Ignoring the entry.", connectedHost);
@@ -732,9 +731,6 @@ class ControlConnection implements Host.StateListener, Connection.Owner {
             versions.add(localRow.getUUID("schema_version"));
 
         for (Row row : peersFuture.get()) {
-
-            if (!isValidPeer(row, false))
-                continue;
 
             InetSocketAddress addr = addressToUseForPeerHost(row, connection.address, cluster);
             if (addr == null || row.isNull("schema_version"))
--- a/driver-core/src/test/java/com/datastax/driver/core/ExtendedPeerCheckDisabledTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/ExtendedPeerCheckDisabledTest.java
@@ -1,0 +1,36 @@
+/*
+ *      Copyright (C) 2012-2015 DataStax Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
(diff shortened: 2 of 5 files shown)


driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
```diff
@@ -720,7 +719,7 @@ boolean checkSchemaAgreement() throws ConnectionException, BusyConnectionExcepti
         if (connection == null || connection.isClosed())
             return false;
 
-        DefaultResultSetFuture peersFuture = new DefaultResultSetFuture(null, cluster.protocolVersion(), new Requests.Query(SELECT_SCHEMA_PEERS));
+        DefaultResultSetFuture peersFuture = new DefaultResultSetFuture(null, cluster.protocolVersion(), new Requests.Query(SELECT_PEERS));
```
tolbertam: This fixes things, but wasn't sure if it was the best solution, as we only want to consider peers that are valid, which requires retrieving all of the peer rows.  Alternatively we could skip peers that we don't have in host metadata, although that may be more involved / less reliable.

newkek:
+1 makes sense to me

driver-core/src/test/java/com/datastax/driver/core/ExtendedPeerCheckDisabledTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * Validates that if the com.datastax.driver.EXTENDED_PEER_CHECK system property is set to false that a peer
+     * with null values for host_id, data_center, rack, tokens is not ignored.
+     *
+     * @test_category host:metadata
+     * @jira_ticket JAVA-852
+     * @since 2.1.10
+     */
+    @Test(groups = "isolated", dataProvider = "disallowedNullColumnsInPeerData", dataProviderClass = ControlConnectionTest.class)
```
adutra, line 30: Do we really need isolated profile because of system properties?
I thought a try-finally block could suffice, e,g, [here](https://github.com/datastax/java-driver/blob/47419b8be04758d22d77534402c4d4c613a41694/driver-core/src/test/java/com/datastax/driver/core/SSLEncryptionTest.java#L118-L131).

tolbertam, line 30: I think we have to do it here since the property is read at static initialization of `ControlConnection` so I don't think there is any way to cause it to reread the property after first touch (unless i'm missing something)

adutra, line 30: Oh sorry you are absolutely right. What about making that field non-final and @VisibleForTesting? Or maybe we could inline it, it's used in two places only. My point is that imo we should avoid the isolated profile as much as we can... but I'm fine with the current state also so giving my :+1:

tolbertam, line 30: I really like that idea.   I'll make that change.  I agree about keeping the number of isolated tests to a minimum.

tolbertam, line 30: On second thought I think i'd like to leave it as it is for now.  Of all our system property driven fields (ones that use `SystemProperties`) all of them are currently final.  Although I do like the suggestion and the approach would be good to apply it for all tests that currently behave this way (we have a few).  I'll take that on in a separate PR if that's ok.

adutra, line 30: Sounds good, no worries! :+1:

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
olim7t: `SELECT_PEERS` retrieves more columns, notably the tokens. So there's a tradeoff between the payload size (tokens can be non-negligible with virtual nodes), and considering bogus entries for schema agreement (which will probably make it fail). In the attachment to JAVA-852 there are lines that have a non-null rpc_address but other null fields, so using one query or the other would make a difference...

tolbertam:
:+1:

PR #622, opened 2016-03-11
JAVA-1111: Example Eclipse RCP application built with Maven and Tycho.

adutra:
Note that I'm not declaring this new example as a child of driver-examples, because I'm not sure we should include it in our release lifecycle. Besides it has a different version number.


adutra:
Closing in favor of #692 and the new examples repository.

PR #624, opened 2016-03-17
2.1.10 doc

olim7t:


olim7t:
Work in progress, more to come.

manual/idempotence/README.md
```diff
@@ -0,0 +1,132 @@
+## Query idempotence
+
+A query is *idempotent* if it can be applied multiple times without changing the result of the initial application. For
```
tolbertam, line 3: Should we add emphasis here that you should make sure that you use client timestamp generators when using idempotent queries?  Although that is default now so maybe not as important to note.  EDIT:   Its not the default in 2.1 though I just recalled.

manual/idempotence/README.md
```diff
@@ -0,0 +1,132 @@
...
+from different clients, there must be a way to reorder them to yield a sequential history that is correct. From our
+clients' point of view, there were two operations:
+
+* client 1 executed a CAS from 1 to 4, that was not applied;
+* client 2 executed a CAS from 4 to 2, that was applied.
+
+But overall the column changed from 1 to 2. There is no ordering of the two operations that can explain that change. We
+broke linearizability by doing a transparent retry at step 6.
+
+To avoid this, the driver considers lightweight transactions as idempotent, and provides a [retry policy](../retries/)
```
tolbertam: Don't you mean non idempotent here?

manual/retries/README.md
```diff
@@ -1,6 +1,201 @@
...
+
+A read request reached the coordinator, which initially believed that there were enough live replicas to process it.
+But, for some reason, one or several replicas were too slow to answer within the predefined timeout
+(`read_request_timeout_in_ms` in `cassandra.yaml`), and the coordinator replied to the client with a `READ_TIMEOUT`
+error.
+
+This could be due to a temporary overloading of these replicas, or even
+that they just failed or were turned off. During reads, Cassandra doesn't request data from every replica to minimize
+internal network traffic; instead, some replicas are only asked for a checksum of the data. A read timeout may occur
+even if enough replica responded to fulfill the consistency level, but only checksum responses were received (the
```
tolbertam: should be 'if enough replicas'

manual/idempotence/README.md
tolbertam, line 3: Hrm, think I've changed my mind about this.  This might be more of a matter of perspective, the client doesn't get a completed Future until the operation succeeds, so even if it fails and retries and there end up being multiple mutations, the user knows that the operation completed.  Where it matters is for LWTs as you mentioned in the documentation.

driver-core/src/main/java/com/datastax/driver/core/utils/UUIDs.java
```diff
@@ -52,8 +52,10 @@
  * </ol>
  * </li>
  * </ol>
+ * Note: currently, native calls won't work from an OSGI container. This is because JNR artifacts don't include proper
```
adutra, line 55: This note would be useful in `AbstractMonotonicTimestampGenerator` as well.

adutra:
:+1: Very good documentation effort.

manual/retries/README.md
```diff
@@ -1,6 +1,201 @@
...
+avoid breaking legacy implementations. All built-in policies implement `ExtendedRetryPolicy`, and we recommend that you
+do too. If you pass a policy that only implements `RetryPolicy`, the driver will use a default `onRequestError`
+implementation that always retries on the next host with the same consistency level (this is consistent with what
+previous 2.1 versions did). In 3.0+, the two interfaces have been merged back together.
+
+The method receives the exception as a parameter, so that implementations can refine their decision based on what
+happened. The possible exceptions are:
+
+* [ServerError]: thrown by the coordinator when an unexpected error occurs. This is generally a Cassandra bug;
+* [OperationTimedOutException]: thrown by the client when it didn't hear back from the coordinator within the
```
adutra, line 101: So are we really planning to turn `OperationTimedOutException` and `ConnectionException` public?

manual/idempotence/README.md
olim7t, line 3: If queries are idempotent it shouldn't matter that they might get executed twice with different timestamps.
If they're not, and this matters to you, you should mark them as such and use the relevant policy that avoids retries.
So I think all in all it's fine.

manual/idempotence/README.md
olim7t: Yes, good catch.

driver-core/src/main/java/com/datastax/driver/core/utils/UUIDs.java
olim7t, line 55: Done.

PR #625, opened 2016-03-21
Support multiple workloads for DSE nodes, updateDSENodeConfig

tolbertam:
Adds support for multiple workloads on a node (i.e. spark and solr on a single node).  Also adds method for calling updatedseconf on a single node as this is now supported by ccm.


tolbertam:
One downside is that you can't set multiple workloads using the `@CCMConfig` annotation.  I couldn't really find a good way of expressing that.  We could have prebaked workloads for common combinations (i.e. spark + graph).  wdyt?  We probably don't need to worry about it now, but might be nice.

adutra:
As discussed, I added one more commit to address multi-workload setups using the `@CCMConfig` annotation.

tolbertam:
Can this get merged before we put 2.1 upstream into 3.0?

PR #626, opened 2016-03-21
JAVA-1174: CAS insert with Mapper class

mgurevin:


tolbertam:
Hi @mgurevin ,  thank you for your contribution!  I agree that it would be a great idea for the Mapper to support `IF NOT EXISTS` and other lwt operations.  After thinking it over and discussing with others I think this option offers a lot of flexibility and convenience, but I do have a concern.

My concern is that the current `Option` types (ttl, timestamp, cl, tracing, save null fields) are all for configuring the queries generated by the mapper, but not changing the query text itself (other than save null fields, which can change the query if disabled).   The mapper is currently not very expressive in the queries it generates.  There is only save/get/delete operations (can't make partial row column deletes/updates for example).  I think in general to enrich the mapper we need to figure out approach that enables more flexibility in mapper queries, sort of like `QueryBuilder` but for entities.

Another thought is that if you are using LWTs on a table that it may be best to always use LWT operations when updating rows in that table.   Perhaps there is something we could configure at the `@Table` level to enforce LWTs.   The only weakness here is that the mapper currently doesn't distinguish between inserts and updates so you wouldn't know when to use `IF EXISTS` and `IF NOT EXISTS`.

koscejev:
Would greatly appreciate if something like this is implemented. For both `if exists` and `if not exists`. Created a JIRA issue: https://datastax-oss.atlassian.net/browse/JAVA-1174

olim7t:

driver-mapping/src/main/java/com/datastax/driver/mapping/QueryType.java
```diff
@@ -69,6 +70,11 @@ String makePreparedQueryString(TableMetadata table, EntityMapper<?> mapper, Mapp
                 Insert.Options usings = insert.using();
                 for (Mapper.Option opt : options) {
                     opt.checkValidFor(QueryType.SAVE, manager);
+
+                    if (shouldIfNotExists(opt)) {
```
newkek, line 74: Maybe we should change the signature of `appendTo` to take `Insert/Delete` in parameter instead of `Insert.Options/Delete.Options`, so that you don't have to do a special case here and can apply `ifNotExists()` in `appendTo`.

olim7t:
Closing in favor of #826. Sorry for opening a separate PR, but I wanted to rebase on 3.x and clean up the option classes before, it was simpler that way. I've cherry-picked your commit as 621b22a, still properly attributed to you.

@koscejev unfortunately, it's not possible to implement IF EXIST. As Andy explained, the mapper uses INSERT for both new and existing objects, and INSERT does not support IF EXIST. That's a limitation of the current implementation, the only workaround is to use an accessor method.

driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
```diff
@@ -1017,11 +1083,7 @@ boolean isIncludedInQuery() {
             Preconditions.checkNotNull(columnMappers);
             this.queryType = queryType;
             this.columns = columnMappers;
-            this.optionTypes = EnumSet.noneOf(Option.Type.class);
-            for (Option opt : options.values()) {
-                if (opt.isIncludedInQuery())
-                    this.optionTypes.add(opt.type);
-            }
+            this.options = options.values();
```
olim7t, line 1086: For the record, we want to keep testing `isIncludedInQuery()`, to avoid storing extra prepared statement when unnecessary. For example, ttl(100) and ttl(2000) both generate the same query containing "TTL (?)", so it's not efficient to reprepare the query for each one.

PR #627, opened 2016-03-22
JAVA-1132 [2.1]: Executing bound statement with no variables results in exception with protocol v1.

adutra:


olim7t:
lgtm :+1:

tolbertam:
implementation and tests look great!  Running well in CI as well :+1:

PR #628, opened 2016-03-22
JAVA-1132 [3.0]: Executing bound statement with no variables results in exception with protocol v1. 

adutra:


adutra:
Closing in favor of #634.

PR #629, opened 2016-03-23
JAVA-1002: Avoid deadlock when re-preparing a statement on other hosts.

olim7t:


tolbertam:
Working great!  We decided not to incorporate an integration test for this (as would be difficult to make it programmatic as the log entry is out of band), but instead will run our duration tests with a single threaded event loop which will easily exhibit the problem.

PR #630, opened 2016-03-24
JAVA-1133: Add OSGi headers to cassandra-driver-extras.

adutra:
The problem was simply due to the fact that this module had a packaging type of jar instead of bundle.

This commit also improves OSGi tests in driver-examples/osgi:

1) Included driver-mapping and driver-extras as a dependency.
Tests now also validate that these modules are fully operational when accessed from within an OSGi container.

2) Split tests in different classes for a better reporting (different test names for each configuration)

3) Updated pax-exam to 4.8.0.

4) Added Guava 19 as a supported Guava version for all bundles (driver-core, driver-mapping, driver-extras and driver-osgi).
Tests in driver-examples/osgi now also test a configuration including Guava 19.0 explicitly.

5) Removed log4j dependency from parent pom.
This can lead to multiple bindings on the classpath if some submodules (e.g. osgi) use a different logging backend.

6) Updated logback configuration file in osgi module.

7) Added project.reporting.outputEncoding Maven property in osgi pom file (Failsafe plugin requires it).

8) Deleted GuavaSanityCheckNegativeIT.


adutra:
Closing in favor of #635.

PR #635, opened 2016-03-24
JAVA-1133: Add OSGi headers to cassandra-driver-extras.

adutra:
The problem was simply due to the fact that this module had a packaging type of jar instead of bundle.

This commit also improves OSGi tests in driver-examples/osgi:

1) Included driver-mapping and driver-extras as a dependency.
Tests now also validate that these modules are fully operational when accessed from within an OSGi container.

2) Split tests in different classes for a better reporting (different test names for each configuration)

3) Updated pax-exam to 4.8.0.

4) Added Guava 19 as a supported Guava version for all bundles (driver-core, driver-mapping, driver-extras and driver-osgi).
Tests in driver-examples/osgi now also test a configuration including Guava 19.0 explicitly.

5) Removed log4j dependency from parent pom.
This can lead to multiple bindings on the classpath if some submodules (e.g. osgi) use a different logging backend.

6) Updated logback configuration file in osgi module.

7) Added project.reporting.outputEncoding Maven property in osgi pom file (Failsafe plugin requires it).

8) Deleted GuavaSanityCheckNegativeIT.


driver-examples/osgi/src/main/java/com/datastax/driver/osgi/impl/Activator.java
```diff
@@ -53,6 +79,17 @@ public void start(BundleContext context) throws Exception {
...
+            Allow Netty ThreadDeathWatcher to terminate;
+            unfortunately we can't explicitly call ThreadDeathWatcher.awaitInactivity()
+            because Netty could be shaded.
+            If this thread isn't terminated when this bundle is closed,
+            we could get exceptions such as this one:
+            Exception in thread "threadDeathWatcher-2-1" java.lang.NoClassDefFoundError: xxx
+            Caused by: java.lang.ClassNotFoundException: Unable to load class 'xxx' because the bundle wiring for xxx is no longer valid.
+            Although ugly, they are harmless and can be safely ignored.
+            */
+            Thread.sleep(10000);
```
tolbertam: Does the sleep need to be 10 seconds?  Is it because that's how often the ThreadDeathWatcher does its check?

tolbertam:
This looks excellent, 👍

tolbertam:
We should wait to merge this until after #636 as it looks like jdk7 is failing to compile extras, which causes OSGi tests to fail.

tolbertam:
I tried rebasing this on #636 and everything worked great.   There were a few conflicts in pom.xml but they were pretty easy to overcome.  I think after merging that branch this one should be good to rebase and merge as well.

driver-examples/osgi/src/main/java/com/datastax/driver/osgi/impl/Activator.java
adutra: I chose 10 secs thinking that in our current Jenkins CI this could be necessary.

driver-examples/osgi/pom.xml
```diff
@@ -23,27 +23,49 @@
...
     <packaging>bundle</packaging>
     <name>DataStax Java Driver for Apache Cassandra Examples - OSGi</name>
     <description>An example of using DataStax Java Driver in an OSGi container.</description>
     <url>https://github.com/datastax/java-driver</url>
 
     <properties>
+        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
         <felix.version>4.6.0</felix.version>
-        <pax-exam.version>3.6.0</pax-exam.version>
+        <pax-exam.version>4.8.0</pax-exam.version>
```
tolbertam: Unfortunately it looks like 4.8.0 requires JDK 7, didn't catch this because our CI bypasses build failures (I should probably look into fixing that).  Switching back to 3.6.0 seems to work fine, do you think we should switch back?

driver-examples/osgi/src/main/java/com/datastax/driver/osgi/impl/Activator.java
tolbertam: Sounds good, makes sense.  The exception doesn't really cause any particular issues right other than being logged?  I've encountered it in the past, but seems everything continues to work with it (just curious on its impact).

driver-examples/osgi/pom.xml
adutra: Ah darn :( sorry for not verifying this. I will revert back and see if the tests compile/pass.

driver-examples/osgi/src/main/java/com/datastax/driver/osgi/impl/Activator.java
adutra: Now that I think of it, the exception isn't harmful, so we could lower the the sleep to 1 or 2 seconds, which is more than enough when running tests on a macbook.

PR #636, opened 2016-03-24
JAVA-1136: Enable JDK signature check in module driver-extras.

adutra:


tolbertam:
Rebased on 3.0.x (only conflict was README.md), to get other jdk6 fix changes to see if it fixes remaining problems.

driver-extras/pom.xml
```diff
@@ -142,18 +142,39 @@
...
-                    https://github.com/mojohaus/animal-sniffer/issues/1
-                    -->
-                    <skip>true</skip>
-                    <signature>
-                        <groupId>org.codehaus.mojo.signature</groupId>
-                        <artifactId>java18</artifactId>
-                        <version>1.0</version>
-                    </signature>
-                </configuration>
+                <executions>
```
tolbertam, line 145: Took me a little bit to wrap my head around this, but this is very cool.  The annotation solution is pretty elegant.  It would be even nicer if `animal-sniffer` supported package level annotations (i.e. if a package is annotated, it skips all of its classes), but it doesn't look like that currently works (but maybe a nice change request for animal sniffer).

driver-extras/pom.xml
```diff
@@ -142,18 +142,39 @@
...
+                                <artifactId>java16</artifactId>
+                                <version>1.0</version>
+                            </signature>
+                            <annotations>
+                                <annotation>com.datastax.driver.extras.codecs.jdk8.IgnoreJDK6Requirement</annotation>
+                            </annotations>
+                        </configuration>
+                    </execution>
+                    <execution>
+                        <id>check-jdk8</id>
```
tolbertam, line 164: One thing that tripped me up was why the `check-jdk8` execution?   My guess is this is for future-proofing if we ever do jdk9 specific stuff?

tolbertam:
This looks good to me and is running well in CI 👍

tolbertam:
I should note that the OSGi tests aren't running on this branch, but i think that is a carry over from 2.1 as it doesn't appear to be running it either :(.  It looks like #635 fixes this, so that will take care of that.

tolbertam:
I went ahead and rebased this since the only conflict was README.md.

PR #638, opened 2016-03-25
JAVA-1086: Support Cassandra 3.2 CAST function in QueryBuilder.

olim7t:
I found a way to add methods to `select()` without breaking binary compatibility, so I seized the opportunity to also add `select().raw()` (which will provide a workaround for missing query builder features in the future).

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -10,6 +10,7 @@ cassandra:
   - 2.1
   - 2.2
   - 3.0
+  - 3.4
 build:
   - type: maven
     version: 3.2.5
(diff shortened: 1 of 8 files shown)


tolbertam:
I noticed in the tests for [CASSANDRA-10310](https://issues.apache.org/jira/browse/CASSANDRA-10310) that is was valid to cast a parameter to a function call, i.e.:

```
SELECT avg(cast(count as double)) FROM myTable
```

To support this I thought it would be useful to add `QueryBuilder.cast` (similar to `QueryBuilder.fcall`), that way we can do something like this to generate that query:

``` java
select().fcall("avg", cast(column("count"), DataType.cdouble())).from("myTable");
```

Hopefully that makes sense (feel free to back out my commit if it doesn't :)).

PR #639, opened 2016-03-25
JAVA-1113: Support Cassandra 3.4 LIKE operator in QueryBuilder.

olim7t:
Note: the new QueryBuilder method permits a bind marker as the LIKE argument.

That's currently not supported by Cassandra, I'm trying to find out if it's a bug. If so, I'll add a warning in the QueryBuilder's javadoc; if not, I'll change the method to accept only a string.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -12,6 +12,7 @@
 - [improvement] JAVA-1136: Enable JDK signature check in module driver-extras.
 - [improvement] JAVA-866: Support tuple notation in QueryBuilder.eq/in.
 - [bug] JAVA-1140: Use same connection to check for schema agreement after a DDL query.
+- [improvement] JAVA-1113: Support Cassandra 3.4 LIKE operator in QueryBuilder.
 
 Merged from 2.1 branch:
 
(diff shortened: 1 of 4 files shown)


tolbertam:
Noticed that you mentioned that 'like' should be added to reserved keywords in [JAVA-1113](https://datastax-oss.atlassian.net/browse/JAVA-1113):

> Add it to the list of reserved keyword that we check when validating identifiers.

Does that still need to be done?  Did we opt not to add it because 'like' isn't a reserved word in older C\* versions?

tolbertam:
Encountered exactly what you were mentioning, in that it doesn't work with a bind marker, which `QueryBuilder` uses.  Looks like it is fixed in 3.6 ([CASSANDRA-1146](https://issues.apache.org/jira/browse/CASSANDRA-11456)).  Tried it out on trunk and it works!   Will push up a test shortly.

adutra:
> Does that still need to be done? Did we opt not to add it because 'like' isn't a reserved word in older C\* versions?

Looks like it's not necessary, see JAVA-1138.

tolbertam:
> Looks like it's not necessary, see JAVA-1138.

Oh hah, I missed the comment.  Sounds good then.  In that case 👍

PR #640, opened 2016-03-29
Retro-fitted FAQs from DITA to Documentor ...

jbisso:
I have added the FAQs from the DITA section (and removed that one) to the Documentor source fq/README.md file.

### DIFF
--- a/faq/README.md
+++ b/faq/README.md
@@ -10,3 +10,121 @@ page](../manual/paging/) for more information.
 Native protocol v1 does not support paging, but you can emulate it in
 CQL with `LIMIT` and the `token()` function. See
 [this conversation](https://groups.google.com/a/lists.datastax.com/d/msg/java-driver-user/U2KzAHruWO4/6vDmUVDDkOwJ) on the mailing list.
+
+### Can I check if a conditional statement (lightweight transaction) was successful?
+
+When executing a conditional statement, the `ResultSet` will contain a single `Row` with a
+column named "applied" of type boolean. This tells whether the conditional statement was
+successful or not.
+
+The driver provides a convenience method [wasApplied] to check this on the result set directly:
+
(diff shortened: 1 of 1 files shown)


faq/README.md
```diff
@@ -54,20 +73,22 @@ Yes. A batch can include both bound statements and simple statements:
...
-atch.add(new SimpleStatement( "INSERT INTO contacts (email, firstname, lastname) VALUES (?, ?, ?)", ...)); 
+batch.add(new SimpleStatement( "INSERT INTO contacts (email, firstname, lastname) VALUES (?, ?, ?)", ...));
 session.execute(batch);
 ```
+
 ### Can I get the raw bytes of a text column?
 
 If you need to access the raw bytes of a text column, call the
-`Row.getBytesUnsafe(“columnName”)` method.
+`Row.getBytesUnsafe("columnName")` method.
```
tolbertam, line 88: Good catch, didn't notice the difference in quotes.

faq/README.md
```diff
@@ -13,40 +13,59 @@ CQL with `LIMIT` and the `token()` function. See
...
 ```java
 ResultSet rset = session.execute(conditionalStatement);
-Row row = rset.one(); // if this is true then the statement was successful 
+Row row = rset.one();
 row.getBool(0);       // this is equivalent row.getBool("applied")
 ```
 
+Note that, unlike manual inspection, `wasApplied` does not consume the first row.
+
+[wasApplied]: http://docs.datastax.com/en/drivers/java/2.1/com/datastax/driver/core/ResultSet.html#wasApplied--
```
tolbertam, line 37: Good idea to link to the relevant pages in the manual for further reading :+1:

tolbertam:
:+1: on the changes as well.

PR #642, opened 2016-04-04
JAVA-1065: Add method to create token from partition key values.

olim7t:


tolbertam:
Looks nice and straightforward!   @RussellSpitzer can you confirm that this would meet your need for [JAVA-1065](https://datastax-oss.atlassian.net/browse/JAVA-1065)?

RussellSpitzer:
This would be fine, we are currently able to do the routing key generation on our own so we just use the tokenFactory hash at the moment. But I think we could change to use this interface instead.

https://github.com/datastax/spark-cassandra-connector/blob/master/spark-cassandra-connector/src/main/java/com/datastax/driver/core/MetadataHook.java

PR #643, opened 2016-04-06
Bootstrap driver-examples module

olim7t:


tolbertam:
This will probably need to be rebased/changed a little bit after #635 is merged.

tolbertam:
Also #615

adutra:
Squashed and rebased on top of current 3.0.x.

driver-examples/src/main/java/com/datastax/driver/examples/datatypes/Blobs.java
```diff
@@ -0,0 +1,251 @@
...
+
+        // One way to read from the buffer is to use absolute getters. Do NOT start reading at index 0, as the buffer
+        // might start at a different position (we'll see an example of that later).
+        for (int i = buffer.position(); i < buffer.limit(); i++) {
+            assert buffer.get(i) == (byte) 0xFF;
+        }
+
+        // Another way is to use relative getters.
+        while (buffer.hasRemaining()) {
+            assert buffer.get() == (byte) 0xFF;
```
tolbertam: If you don't have assertions enabled (`-ea` vm argument) this will tight loop since buffer.get() is not evaluated.   Maybe we should change these to not use assertions?  Either that or we could document that assertions are required.

tolbertam: Alternatively we could call .get() every time and then assert on whatever it returns separately, i.e.:

``` java
while (buffer.hasRemaining()) {
    byte b = buffer.get();
    assert b == (byte) 0xFF;
}
```

tolbertam:
Looks great!  The only thing I think that should be addressed before merge is my [comment on asserts](https://github.com/datastax/java-driver/pull/643#discussion_r61632539) other than that this is good to merge! 👍

driver-examples/src/main/java/com/datastax/driver/examples/datatypes/Blobs.java
adutra: Went ahead and adopted your suggestion in 2 places where there were risks of tight loops.

PR #645, opened 2016-04-08
Added changes to support columns list for GET queries.

vagifzeynalov:
Example:

```
public T get(K1 pk1, K2 pk2, Set<String> columns) {
    return mapper().get(pk1, pk2, columns);
}

final Entity entiry = dao.get(key1, key2, ImmutableSet.of("col1", "col2"));
```

### DIFF
--- a/driver-mapping/src/main/java/com/datastax/driver/mapping/ColumnMapper.java
+++ b/driver-mapping/src/main/java/com/datastax/driver/mapping/ColumnMapper.java
@@ -56,10 +56,14 @@ abstract class ColumnMapper<T> {
 
     public abstract void setValue(T entity, Object value);
 
-    public String getColumnName() {
-        return kind == Kind.COMPUTED
+    public String getColumnName(boolean quoted) {
+        return kind == Kind.COMPUTED || !quoted
                 ? columnName
                 : Metadata.quote(columnName);
+    }
+    
+    public String getColumnName() {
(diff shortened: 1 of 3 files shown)


olim7t:

adutra:
This can be easily achieved already with one of the following methods:

1. Using an accessor:

```
@Accessor
public interface EntityAccessor {

    @Query("SELECT col1, col2 FROM entity WHERE key1 = ? AND key1 = ?")
    Entity findEntity(int key1, int key2);

}
```

2. Using a custom resultset:

```
ResultSet rs = session.execute("SELECT col1, col2 FROM entity WHERE key1 = ? AND key2 = ?", key1, key2);
Entity entity = dao.map(rs).one();
```

Globally I'm not sure the use case is general enough to fit inside the mapper.

vagifzeynalov:
@adutra it is general enough, since the mapper itself uses it inside of it at line 107
`PreparedStatement getPreparedQuery(QueryType type, Set<ColumnMapper<?>> columns, EnumMap<Option.Type, Option> options) {
`
It was there before my change.

Why not to provide the ability for the top level classes to pass the columns list?

adutra:

@vagifzeynalov sorry for closing so abruptly but you didn't include any tests so it's hard to evaluate how stable this PR is, and I still think that we could achieve the same effect using accessors. Thanks anyway for your time and effort.

PR #646, opened 2016-04-08
Added changes to support columns list for GET queries.

vagifzeynalov:
Example:

```
public T get(K1 pk1, K2 pk2, Set<String> columns) {
    return mapper().get(pk1, pk2, columns);
}

final Entity entiry = dao.get(key1, key2, ImmutableSet.of("col1", "col2"));
```

### DIFF
--- a/driver-mapping/src/main/java/com/datastax/driver/mapping/ColumnMapper.java
+++ b/driver-mapping/src/main/java/com/datastax/driver/mapping/ColumnMapper.java
@@ -54,10 +54,14 @@ abstract class ColumnMapper<T> {
 
     abstract void setValue(T entity, Object value);
 
-    String getColumnName() {
-        return kind == Kind.COMPUTED
+    public String getColumnName(boolean quoted) {
+        return kind == Kind.COMPUTED || !quoted
                 ? columnName
                 : Metadata.quote(columnName);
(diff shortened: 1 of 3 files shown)


vagifzeynalov:
Hi guys!
Here is a quick example how it works
https://github.com/vagifzeynalov/dbl_dse_example/tree/mapper_add_columns_for_get
Especially this test file https://github.com/vagifzeynalov/dbl_dse_example/blob/mapper_add_columns_for_get/src/test/java/com/zeynalov/rw/core/dbl/dao/UserAddressDaoTest.java

olim7t:
Is the goal to restrict the columns in the SELECT statement?
One problem I see is that `Set<String>` is a valid type for a partition key component, for example with this table:

``` sql
CREATE TABLE foo(s frozen<set<text>> PRIMARY KEY );
```

vagifzeynalov:
Yes, you've got the goal right.
Actually it could be some special class for columns, like Columns extends HashSet< String >, or even List instead... I don't know what is the best in terms of the compatibility.
Do you want me to change it, or you can change it by yourself?

adutra:
Closing in favor of #645 as this should target a minor release, not a patch one. Thanks!

PR #650, opened 2016-04-12
JAVA-1154: Allow individual Statement to cancel the read timeout.

olim7t:


tolbertam:
looks good to me.  Added a small test for testing with '0' timeout at statement level (feel free to squash). 👍

PR #651, opened 2016-04-14
JAVA-866: Support tuple notation in QueryBuilder.eq/in.

olim7t:


driver-core/src/test/java/com/datastax/driver/core/querybuilder/QueryBuilderExecutionTest.java
```diff
@@ -182,4 +186,136 @@ public void should_delete_map_entry_with_bind_marker() throws Exception {
         assertThat(actual).containsExactly(entry(2, "bar"));
     }
 
+    /**
+     * Validates that {@link QueryBuilder} appropriately builds a select query using a tuple 'eq' comparison, i.e.:
```
adutra: I think the correct jargon is "multi-column equal relation", but I'm fine with "tuple equal comparison".

driver-core/src/main/java/com/datastax/driver/core/querybuilder/QueryBuilder.java
```diff
@@ -277,6 +277,26 @@ public static Clause eq(String name, Object value) {
...
+     * will generate the CQL WHERE clause {@code (a, b) = (2, 'test') }.
+     * <p/>
+     * Please note that this variant is only supported starting with Cassandra 2.0.6.
+     *
+     * @param names  the column names
+     * @param values the values
+     * @return the corresponding where clause.
+     * @throws IllegalArgumentException if {@code names.size() != values.size()}.
+     */
+    public static Clause eq(List<String> names, List<?> values) {
```
adutra, line 292: Should we provide overloaded versions of these two methods that take array arguments, e.g. `eq(String[] names, Object... values)`?

olim7t, line 292: The existing methods (`lt`, `gt`, etc.) didn't, I'd rather avoid overloading the API.

PR #652, opened 2016-04-18
Override copy() in CodecNotFoundException and remove JDK-specific behavior dependency in test.

tolbertam:
Just a small test adjustment.

This is needed because the Exception is propagated as the query is now done in a separate thread (JAVA-1070).


tolbertam:
passing CI, merging :)

PR #656, opened 2016-04-22
JAVA-1161 ZonedDateTime should be serialized and formatted with full information about TimeZone

chibenwa:
The java driver failed to serialize / deserialize dates like : 

```
ZonedDateTime.parse("2016-04-07T02:01+07:00[Asia/Vientiane]")
```

Basicaly the timeZone is lost, and this line fails : 

```
Assertions.assertThat(codec).withProtocolVersion(V4).canSerialize(input);
```

Note that timeZone is Important when modifying date. For instance French timeZone (Central Europe Time) changes of offset 2 times a year. If I want to know time in France 48 h later, answer might be in CET timeZone +47h, +48h or +49h, leading to different results. With only offset, this information is lost.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -2,6 +2,7 @@
 
 ### 3.0.1
 
+- [bug] JAVA-1161: ZonedDateTimeCodec does not preserve timeZone, but only the offset
 - [bug] JAVA-1132: Executing bound statement with no variables results in exception with protocol v1.
 - [improvement] JAVA-1040: SimpleStatement parameters support in QueryLogger.
 - [improvement] JAVA-1151: Fail fast if HdrHistogram is not in the classpath.
--- a/driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/ZonedDateTimeCodec.java
+++ b/driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/ZonedDateTimeCodec.java
@@ -19,29 +19,32 @@ import com.datastax.driver.core.*;
 import com.datastax.driver.core.exceptions.InvalidTypeException;
 
 import java.nio.ByteBuffer;
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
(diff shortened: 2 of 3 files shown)


adutra:
@chibenwa Now I understand your use case, and indeed adding a zone ID would help you when computing durations between timestamps.

But I still think that this is a rather uncommon case and I'm not comfortable with printing  a hardcoded zone ID such as `source.getZone().getId()`.

Instead I would suggest:
1. To improve `ZONE_FORMATTER` so that it optionally prints the zone ID, if available:

``` java
private static final DateTimeFormatter DEFAULT_ZONE_FORMATTER = new DateTimeFormatterBuilder()
        .optionalStart()
        .appendOffset("+HH:MM", "+00:00")
        .optionalEnd()
        .optionalStart()
        .appendLiteral('[')
        .parseCaseSensitive()
        .appendZoneRegionId()
        .appendLiteral(']')
        .toFormatter();
```

I don't think this would break user code as I suspect few of them care about zone IDs.
1. Allow users to specify their own formatters; this way users with special use cases have a workaround that wouldn't require changing the code again and again. One of our main concerns is to provide code that is useful to the most of our users, and customizable enough to be adapted to even the most rare cases.

I have a gist ready here:
https://gist.github.com/adutra/015fa2d135ce80ce22f3059a09a1db37

Feel free to cherry-pick and update your PR if you have time. Otherwise I can take care of it (and I will mention your contributions).

One last thing: the Joda equivalent, `DateTimeCodec`, would have to be amended the same way.

chibenwa:
Thank you for the time you took to read and answer this. Modifying the formatter seem to work in this case and do not change existing code much.

I will also add a constructor option for the formatter. (It is easy to do so...)

I will also have a look for yoda-time.

I think I can have a look before next thursday.

adutra:
@chibenwa great thanks! I updated my gist with a few minor changes.

chibenwa:
Hi @adutra . I had a look to your Gist. Looks great.

You forgot annotations : 

@IgnoreJDK6Requirement
@SuppressWarnings("Since15")

I will update the PR with your changes tonight, and will take care of yoda-time tomorrow.

Thanks

adutra:
@chibenwa I will take care of the rest from now on. Thanks for your contribution!

PR #658, opened 2016-04-29
JAVA-1171: Add Host method to determine if DSE Graph is enabled.

adutra:
This PR applies the patch supplied by @djatnieks with very minor formatting changes.


olim7t:
👍

tolbertam:
Tried this out with a Graph enabled cluster with some nodes enabled/not enabled and it works great.  Also tried with an older DSE version and it had no problems.  👍 .

PR #660, opened 2016-05-03
JAVA-1179: Request objects should be copied when executed.

adutra:
Will add a test later if not too complex to do.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Connection.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Connection.java
@@ -502,11 +502,10 @@ class Connection {
 
     public ResponseHandler write(ResponseCallback callback, boolean startTimeout) throws ConnectionException, BusyConnectionException {
 
-        Message.Request request = callback.request();
-
         ResponseHandler handler = new ResponseHandler(this, callback);
         dispatcher.add(handler);
-        request.setStreamId(handler.streamId);
+
+        Message.Request request = callback.request().setStreamId(handler.streamId);
 
         /*
          * We check for close/defunct *after* having set the handler because closing/defuncting
--- a/driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
@@ -64,7 +64,7 @@ class RequestHandler {
     private final long startTime;
 
     private final AtomicBoolean isDone = new AtomicBoolean();
-    private AtomicInteger executionCount = new AtomicInteger();
+    private final AtomicInteger executionCount = new AtomicInteger();
 
     public RequestHandler(SessionManager manager, Callback callback, Statement statement) {
         this.id = Long.toString(System.identityHashCode(this));
@@ -106,10 +106,6 @@ class RequestHandler {
 
         Message.Request request = callback.request();
         int position = executionCount.incrementAndGet();
-        // Clone the request after the first execution, since we set the streamId on it later and we
-        // don't want to share that across executions.
-        if (position > 1)
-            request = request.copy();
 
         SpeculativeExecution execution = new SpeculativeExecution(request, position);
         runningExecutions.add(execution);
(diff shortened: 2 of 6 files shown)


driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -502,10 +502,11 @@ public ResponseHandler write(ResponseCallback callback) throws ConnectionExcepti
 
     public ResponseHandler write(ResponseCallback callback, boolean startTimeout) throws ConnectionException, BusyConnectionException {
 
-        Message.Request request = callback.request();
-
         ResponseHandler handler = new ResponseHandler(this, callback);
         dispatcher.add(handler);
+
+        // JAVA-1179: defensively guard against reusing same request object twice
+        Message.Request request = callback.request().copy();
```
adutra: The copy() call is superfluous on the very first execution of the request.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -404,6 +397,7 @@ void cancel() {
 
         @Override
         public Message.Request request() {
+            Message.Request request = callback.request();
```
tolbertam: Is the idea to always copy because the request could be used again in a Retry (independent of a speculative execution being executed?)

adutra: Yes, but not here, rather in `Connection.write()`. Here we are in a higher level IMO, so we should only copy the original request if something in it should be modified, e.g. the CL.

tolbertam:
Added a test that reproduces this on 2.0 branch typically within 2 seconds, but does not manifest against this branch.  If it fails, you should see something like:

```
    2743   [cluster1-nio-worker-3] ERROR com.datastax.driver.core.ReusedStreamIdTest  - Got response without column schema_version, got columns Columns[thrift_version(varchar)] from Host /127.0.1.2:9042.
    2743   [cluster1-nio-worker-2] ERROR com.datastax.driver.core.ReusedStreamIdTest  - Got response without column rack, got columns Columns[release_version(varchar)] from Host /127.0.1.1:9042.
    2744   [cluster1-nio-worker-1] ERROR com.datastax.driver.core.ReusedStreamIdTest  - Got response without column thrift_version, got columns Columns[schema_version(uuid)] from Host /127.0.1.1:9042.
    3749   [cluster1-nio-worker-2] ERROR com.datastax.driver.core.ReusedStreamIdTest  - Got response without column broadcast_address, got columns Columns[rack(varchar)] from Host 

java.lang.AssertionError: Error triggered at or before 20 of 500 requests after 2040ms.
```

driver-core/src/main/java/com/datastax/driver/core/Connection.java
tolbertam: Is there any way around it or is an optimization to no copy not really worth it?

olim7t: It's also superfluous when `callback.request()` already did a copy because of a different retry CL.

olim7t:
Pushed an alternative proposal: copy the object only when needed in `setStreamId()`. This is maybe a bit more hackish but avoids the superfluous initial copy.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -237,6 +238,7 @@ private Metrics metrics() {
      */
     class SpeculativeExecution implements Connection.ResponseCallback {
         final String id;
+        private final Message.Request request;
```
adutra, line 241: Why store this as a final field if `callback.request()` is accessible from the enclosing `RequestHandler` instance?

tolbertam:
looks good, nice to avoid that copy on first use 👍

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
olim7t, line 241: Not much difference really, if anything it makes it clear that the call to `callback.request()` always returns the same object. I'll leave it this way for now since it's not part of the fix.

PR #661, opened 2016-05-04
JAVA-1180: Flusher should check if requests are canceled before flushing them

adutra:


tolbertam:
Nice!  Can exercise this with ReusedStreamIdTest from #660:

```
    2496   [cluster1-nio-worker-1] DEBUG com.datastax.driver.core.Connection$Flusher  - Not flushing request QUERY select schema_version from system.local([cl=ONE, vals=[], skip=false, psize=5000, state=null, serialCl=SERIAL]) because its handler has been canceled
    2496   [cluster1-nio-worker-2] DEBUG com.datastax.driver.core.Connection$Flusher  - Not flushing request QUERY select release_version from system.local([cl=ONE, vals=[], skip=false, psize=5000, state=null, serialCl=SERIAL]) because its handler has been canceled
    2496   [cluster1-nio-worker-3] DEBUG com.datastax.driver.core.Connection$Flusher  - Not flushing request QUERY select thrift_version from system.local([cl=ONE, vals=[], skip=false, psize=5000, state=null, serialCl=SERIAL]) because its handler has been canceled
```

In addition, this seems to address JAVA-1179, since I haven't seen a case of a callback getting the wrong response.  Although there is a small window that we discussed previously where the callback could be cancelled between when we check if it's cancelled and when its written, but this covers the case where a callback is blocking flushing pretty well.

olim7t:
In addition, it would be nice to release the Request's streamId if we decide not to write it (assuming JAVA-1179 is fixed and the streamId is still the right one). Otherwise it will never be released.

That being said, I wonder if it's worth applying this fix. JAVA-1179 already covers the issue, if Requests have the right streamIds, it's not a big deal to send a few already-timed-out ones, only slightly inefficient. Considering that the issue arises with improper usage of the driver (which we'll try to prevent with JAVA-1182), I think we can live with it.

adutra:
> In addition, it would be nice to release the Request's streamId if we decide not to write it (assuming JAVA-1179 is fixed and the streamId is still the right one). Otherwise it will never be released.

I think it will, when the handler gets canceled.

> That being said, I wonder if it's worth applying this fix. JAVA-1179 already covers the issue, if Requests have the right streamIds, it's not a big deal to send a few already-timed-out ones, only slightly inefficient. 

It might happen more often than we think, because in 99.9% of the cases, it goes unnoticed. So I think we could keep the fix: it's small, and if anything else, it should save some network bandwidth.

olim7t:
> It might happen more often than we think

If the flusher is getting so delayed that it sends timed-out requests, yes. My point is that this situation should not happen in the first place, unless something is really wrong.

adutra:
One other related aspect that I would like to bring your attention to is [these two lines](https://github.com/datastax/java-driver/blob/2.0/driver-core/src/main/java/com/datastax/driver/core/Connection.java#L1249-L1250):

``` java
if (callback.onTimeout(connection, System.nanoTime() - startTime, retryCount))
    cancelHandler();
```

In extremely rare situations (and the fix for JAVA-1180 wouldn't solve it), I believe a request could be executed twice:
1. Thread worker-1 enqueues Request1a
2. Thread timeouter invokes `onTimeout`
3. Thread worker-2 enqueues Request1b (retry)
4. Thread nio-worker-Z flushes Request1a
5. Thread nio-worker-X flushes Request1b
6. Thread nio-worker-W receives response for Request1a, corresponding handler is invoked and GCed
7. Thread nio-worker-Y receives response for Request1b, corresponding handler is invoked and GCed
8. Thread timeouter invokes `cancelHandler`
9. Thread timeouter finds no handler

It seems that the Flusher and Timeouter code are full of such possible, yet unlikely, race conditions. I don't know if there is something simple we could do to make them 100% reliable.

olim7t:
The `QueryState` should handle this case: at step 3, the state will change to (1, inProgress). Step 6 will do nothing because the response is for retry 0, but the current retryCount is 1. Step 7 will complete the `RequestHandler`. And I don't see step 9 causing any problem, the removeHandler call will return early.

olim7t:
Closing this as we've decided not to fix the ticket (see my explanation on [JAVA-1180](https://datastax-oss.atlassian.net/browse/JAVA-1180)).
@adutra we can continue the discussion about potential race conditions here or elsewhere.

adutra:
@olim7t Thanks for pointing me to `QueryState`, that component indeed makes my scenario impossible.

PR #662, opened 2016-05-04
JAVA-1184 Unwrap StatementWrappers when extracting column definitions.

Aulust:


driver-core/src/main/java/com/datastax/driver/core/DefaultResultSetFuture.java
```diff
@@ -156,7 +156,7 @@ public void run() {
                     setException(new DriverInternalError(String.format("Got unexpected %s response from %s", response.type, connection.address)));
                     break;
             }
-        } catch (RuntimeException e) {
+        } catch (Throwable e) {
```
adutra, line 159: Not a big fan of this, are you trying to catch the `AssertionError` thrown by `ArrayBackedResultSet`?

olim7t, line 159: I think the change makes sense: if we don't catch a `Throwable` here, it goes to [Dispatcher#exceptionCaught](https://github.com/datastax/java-driver/blob/2.1/driver-core/src/main/java/com/datastax/driver/core/Connection.java#L1042) which will only log it at DEBUG level, then defunct the connection. If the exception is thrown early enough, the future is not completed and hangs forever.
By setting the future with the exception it's rethrown to the user code, which makes it much easier to debug.

Aulust, line 159: Apart from catching `AssertionError` (in case of my test) this might be useful with identifying `OutOfMemoryError` or `StackOverflowError`. Otherwise you might end up with all of application threads been blocked waiting for futures to resolve, as @olim7t wrote. And thread dump won't help you find the cause of the problem.

Of course, there might be a better solution, so maybe it is worth to revert this change and create a new issue to address the problem.

olim7t:
Closing in favor of #665.
@Aulust, I opened a different PR because we decided to target this change for 2.0. Your commit is still there, properly attributed to you (I just renamed a couple of test methods). I added another commit to fix a similar issue in the query logger.
We're going to include this in a patch release for both 2.0, 2.1 and 3.0.

PR #663, opened 2016-05-05
JAVA-1182: Throw error when synchronous call made on I/O thread.

olim7t:
Note: for cases where the error cannot be propagated, like `Futures.addCallback`, Guava will log the error with java.util.logging. I'm not sure if we can/should do anything about that, or just expect that client apps will properly monitor those logs or catch the error before.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/AbstractSession.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/AbstractSession.java
@@ -51,6 +51,7 @@ public abstract class AbstractSession implements Session, AsyncInitSession {
      */
     @Override
     public ResultSet execute(Statement statement) {
+        checkNotInEventLoop();
         return executeAsync(statement).getUninterruptibly();
     }
 
@@ -75,6 +76,7 @@ public abstract class AbstractSession implements Session, AsyncInitSession {
      */
     @Override
     public PreparedStatement prepare(String query) {
+        checkNotInEventLoop();
         try {
             return Uninterruptibles.getUninterruptibly(prepareAsync(query));
         } catch (ExecutionException e) {
@@ -87,6 +89,7 @@ public abstract class AbstractSession implements Session, AsyncInitSession {
      */
     @Override
     public PreparedStatement prepare(RegularStatement statement) {
+        checkNotInEventLoop();
         try {
             return Uninterruptibles.getUninterruptibly(prepareAsync(statement));
         } catch (ExecutionException e) {
@@ -131,5 +134,14 @@ public abstract class AbstractSession implements Session, AsyncInitSession {
         } catch (InterruptedException e) {
             Thread.currentThread().interrupt();
         }
+    }
+
+    /**
+     * Checks that the current thread is not one of the Netty I/O threads used by the driver.
+     * <p/>
+     * This is called from the synchronous methods of this class to prevent deadlock issues.
+     */
+    protected void checkNotInEventLoop() {
+        // This method is concrete only to avoid a breaking change. See subclass for the actual implementation.
     }
 }
--- a/driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
@@ -27,6 +27,7 @@ import com.google.common.base.Functions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.*;
+import io.netty.util.concurrent.EventExecutor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -45,6 +46,9 @@ import java.util.concurrent.atomic.AtomicReference;
 class SessionManager extends AbstractSession {
 
     private static final Logger logger = LoggerFactory.getLogger(Session.class);
+
+    private static final boolean CHECK_IO_DEADLOCKS = SystemProperties.getBoolean(
+            "com.datastax.driver.CHECK_IO_DEADLOCKS", true);
 
     final Cluster cluster;
     final ConcurrentMap<Host, HostConnectionPool> pools;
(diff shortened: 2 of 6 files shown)


driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -611,6 +615,21 @@ void cleanupIdleConnections(long now) {
         }
     }
 
+    @Override
+    protected void checkNotInEventLoop() {
+        if (!CHECK_IO_DEADLOCKS)
+            return;
+        for (EventExecutor executor : cluster.manager.connectionFactory.eventLoopGroup) {
+            if (executor.inEventLoop()) {
+                throw new IllegalStateException(
```
adutra, line 624: Weren't we supposed to log a warning first, then turn it into an exception at the next major release? Maybe you can turn your system property into stg like `IO_DEADLOCK_CHECK_MODE` with values `NONE`, `WARN`, `FAIL`.

driver-core/src/test/java/com/datastax/driver/core/AsyncQueryTest.java
```diff
@@ -129,6 +130,25 @@ public void should_propagate_error_to_chained_query_if_session_init_fails() thro
...
+        ListenableFuture<Void> f2 = Futures.transform(f, new Function<ResultSet, Void>() {
+            @Override
+            public Void apply(ResultSet input) {
+                session.execute("select release_version from system.local");
+                return null;
+            }
+        });
+        try {
+            f2.get();
+        } catch (Exception e) {
```
adutra, line 147: You should add a `fail()` call here (after `f2.get()`).

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
olim7t, line 624: Oh yes, I explained that, but on the jira ticket:

> After looking further into this, I think just warning makes no sense. Synchronous calls on an I/O thread are always a bug, and I can't think of any situation where we would want to leave users' applications running with a risk of deadlock, with potentially very hard to debug consequences.
> So I'm going for the exception directly.

Funny that your suggestion is exactly how I had done it before changing my mind :)

adutra, line 624: Ha 👍

driver-core/src/main/java/com/datastax/driver/core/AbstractSession.java
```diff
@@ -51,6 +51,7 @@ public ResultSet execute(String query, Object... values) {
      */
     @Override
     public ResultSet execute(Statement statement) {
+        checkNotInEventLoop();
```
adutra, line 54: Not sure we should do the check in `AbstractSession`. I understand the motivation to make all possible subclasses resilient to IO thread misuse but 1) users will have to implement `checkNotInEventLoop` anyway (and I bet they won't) 2) some implementations may not even use Netty at all, in which case the method would sound inappropriate. Why not simply keep this check internal to `SessionManager`?

olim7t, line 54: Because the synchronous methods are defined in `AbstractSession`. But now I realize I could also redefine them in `SessionManager` to do the check and then just call super().

olim7t, line 54: Mmm actually I don't like that solution. `SessionManager` needs to override only some of the overloaded variants, to avoid performing the check twice (for example it's needed for `execute(Statement)` but not `execute(String)` because it calls the former). And that info is an implementation detail of `AbstractSession`, so I prefer keeping it there.
Implementations that don't need the check don't need to do anything, since by default the method is empty.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -611,6 +615,21 @@ void cleanupIdleConnections(long now) {
...
+                        "Detected a synchronous Session call (execute() or prepare()) on an I/O thread, " +
```
tolbertam, line 625: "This is a bug" could be interpreted as a bug in the driver code by users who don't understand the issue.  Can we remove that statement, i.e. instead: 

> Detected a synchronous Session call (execute() or prepare()) on an I/O thread, this can cause deadlocks or unpredictable behavior. Make sure your Future callbacks only use async calls, or schedule them on a different executor.

tolbertam:
What about the case where a setKeyspace is being done as part of borrowing a connection ([JAVA-893](https://datastax-oss.atlassian.net/browse/JAVA-893))?  I know that can be done in a netty thread still, even with the `JAVA-1002` fix (which isn't on 2.0.X, but can't remember whether or not that was needed there) and its not really the user's fault.

tolbertam:
> What about the case where a setKeyspace is being done as part of borrowing a connection (JAVA-893)? I know that can be done in a netty thread still, even with the JAVA-1002 fix (which isn't on 2.0.X, but can't remember whether or not that was needed there) and its not really the user's fault.

its not a problem since the check is only done in prepare & execute, were the set keyspace call does a connection.write directly, so the behavior there is the same as before.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
olim7t, line 625: Good point, changed it.

tolbertam, line 625: In the newest commit looks like there is an extra and that can be removed.

> this and can cause

olim7t:
Blocking setKeyspace calls are definitely a problem and documented as such, that's why we recommend using a keyspace-less session for async.

tolbertam:
> Blocking setKeyspace calls are definitely a problem and documented as such, that's why we recommend using a keyspace-less session for async.

Was able to confirm that was still a problem, was just curious as to whether or not this change touched that path at all (which it doesn't)

olim7t:
Yes, internal setKeyspace calls don't go through Session.execute. This change will only affect user queries.

PR #664, opened 2016-05-10
JAVA-743: Add JSON support to QueryBuilder

adutra:


driver-core/src/main/java/com/datastax/driver/core/querybuilder/Insert.java
```diff
@@ -126,6 +148,57 @@ public Insert values(List<String> names, List<Object> values) {
...
+     * </pre>
+     * <h3>Escaping quotes in column values</h3>
+     * Double quotes should be escaped with a backslash, but single quotes should be escaped in
+     * the CQL manner, i.e. by another single quote. For example, the column value
+     * {@code foo"'bar} should be inserted in the JSON string
+     * as {@code "foo\"''bar"}.
+     * <p>
+     * <h3>Null values and tombstones</h3>
+     * Any columns which are omitted from the JSON string will be defaulted to a {@code NULL} value
+     * (which will result in a tombstone being created).
```
olim7t, line 193: Great doc 👍

adutra:

driver-core/src/test/java/com/datastax/driver/core/querybuilder/QueryBuilderTest.java
```diff
@@ -922,6 +922,7 @@ public void should_return_object_at_ith_index() {
     }
 
     @Test(groups = "unit")
+    @CassandraVersion(major = 2.1)
```
tolbertam: Is the `@CassandraVersion` annotation necessary here?  These are just unit tests so the cassandra configuration shouldn't matter.

tolbertam:
This works super nice with the `Jsr353JsonCodec` will push up some tests using `QueryBuilder` with it that I added to the `driver-extras` module, but would be nice to have something for `driver-examples` too, what do you think?   (I can do it as a separate PR if you guys agree that it makes sense).

tolbertam:
Added some tests in driver-extras.  At first glance it may seem awkward to have it here, but it allows us the benefit of using `Jsr353JsonCodec` and `javax.json`.

adutra:
> This works super nice with the Jsr353JsonCodec

Your tests are simply brilliant, well done!

> would be nice to have something for driver-examples too, what do you think?

Yes, another great idea :) We definitely need some examples for our JSON fans. I think it's best to open a Jira first before submitting the PR.

driver-core/src/test/java/com/datastax/driver/core/querybuilder/QueryBuilderTest.java
adutra: Good point I removed them (I was very proud of myself for not forgetting them for once...).

adutra:
Squashed and rebased.

> At first glance it may seem awkward to have it here, but it allows us the benefit of using Jsr353JsonCodec and javax.json.

That brings the question: would it be crazy to add support for `javax.json` in the `QueryBuilder` directly? Obviously it's another dependency that will be bound to our API but `javax.json` should be quite safe, wdyt?

There would be one huge advantage: methods that currently take a JSON string parameter are frankly horrible, especially if the user needs to quote CQL identifiers; I don't expect enthusiasm from users here. Having an overloaded method that takes a `JsonObject` would be of great help. 

/cc @al3xandru

tolbertam:
> That brings the question: would it be crazy to add support for javax.json in the QueryBuilder directly? Obviously it's another dependency that will be bound to our API but javax.json should be quite safe, wdyt?

I think this is a great idea, especially since the jsr 353 dependency would be optional.

> There would be one huge advantage: methods that currently take a JSON string parameter are frankly horrible, especially if the user needs to quote CQL identifiers; I don't expect enthusiasm from users here. Having an overloaded method that takes a JsonObject would be of great help.

I agree, overloading seems like a great idea.

tolbertam:
> We definitely need some examples for our JSON fans. I think it's best to open a Jira first before submitting the PR.

Opened [JAVA-1237](https://datastax-oss.atlassian.net/browse/JAVA-1237) and will do that as a follow on.

tolbertam:
Another idea, if we decide to overload in the `QueryBuilder`, we could also consider making json more first class by adding `getJson` and `setJson` methods for `Gettable`/`Settable` implementations (`Row`,`Statement`,`UDTValue`, etc.)

adutra:
> I think this is a great idea, especially since the jsr 353 dependency would be optional.

Hmm if the method is in class `QueryBuilder` I don't think it can be optional. `QueryBuilder` would fail to load if `JsonObject` isn't in the classpath.

> we could also consider making json more first class by adding getJson and setJson methods for Gettable/Settable implementations (Row,Statement,UDTValue, etc.)

So that would boil down to migrating JsonCodec to driver-core, right? What I'm concerned about is that I'm not sure that java JSON API is that popular. So it would bring the question of why not including Jackson while we're at it.
It seems like we are opening the pandora box. I'm not sure how to proceed, but maybe opening a Jira and engage the discussion could be the best first step.

tolbertam:
> Hmm if the method is in class QueryBuilder I don't think it can be optional. QueryBuilder would fail to load if JsonObject isn't in the classpath.

Ack, yes that would make this difficult.

> I'm not sure how to proceed, but maybe opening a Jira and engage the discussion could be the best first step.

That sounds best, definitely think there is more we can do to make json more of a first class citizen, but don't think we should do that at the expense of adding scope creep to this ticket.    ~~Will open up a ticket for this, thanks!~~  Opened up [JAVA-1239](https://datastax-oss.atlassian.net/browse/JAVA-1239) to capture that.

PR #665, opened 2016-05-10
JAVA-1184: Unwrap StatementWrappers when extracting column definitions.

olim7t:


driver-core/src/main/java/com/datastax/driver/core/QueryLogger.java
```diff
@@ -614,6 +614,9 @@ public void setMaxLoggedParameters(int maxLoggedParameters) {
      */
     @Override
     public void update(Host host, Statement statement, Exception exception, long newLatencyNanos) {
+        if (statement instanceof StatementWrapper)
```
adutra, line 617: The unwrapping was already done line 751, but doing the unwrapping here makes the query logger log parameters for wrapped statements as well, so I agree that it is better to unwrap here than at line 751.
One thing to consider for a future refactoring of this class is to allow for a customizable way to log statement wrappers (and maybe all kinds of statements), see JAVA-1191 for a use case for that. Currently it is not possible at all.

PR #666, opened 2016-05-11
JAVA-1193 + JAVA-1120: Refresh token and replica metadata synchronously when schema is altered.

adutra:
I'm also including a fix for JAVA-1120 in the same PR as the affected code is pretty much the same.

Note that I don't have specific tests for JAVA-1120 but I did benchmark quickly the creation of 50 tables with default debouncer options. Without the fix it takes in avg. 50 secs on my laptop; with the fix, it takes 3 secs in avg.

The 3rd commit brings minor enhancements to `SchemaChangesTest`: missing assertions and verifications basically.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -2266,7 +2266,15 @@ public class Cluster implements Closeable {
                         if (!schemaInAgreement)
                             logger.warn("No schema agreement from live replicas after {} s. The schema may not be up to date on some nodes.", configuration.getProtocolOptions().getMaxSchemaAgreementWaitSeconds());
 
-                        ListenableFuture<Void> schemaReady = refreshSchema ? submitSchemaRefresh(targetType, targetKeyspace, targetName, targetSignature) : MoreFutures.VOID_SUCCESS;
+                        ListenableFuture<Void> schemaReady;
+                        if (refreshSchema) {
+                            schemaReady = submitSchemaRefresh(targetType, targetKeyspace, targetName, targetSignature);
+                            // JAVA-1120: skip debouncing delay and force immediate delivery
+                            if (!schemaReady.isDone())
+                                schemaRefreshRequestDebouncer.scheduleImmediateDelivery();
+                        } else {
+                            schemaReady = MoreFutures.VOID_SUCCESS;
+                        }
                         final boolean finalSchemaInAgreement = schemaInAgreement;
                         schemaReady.addListener(new Runnable() {
                             @Override
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -1851,9 +1851,10 @@ public class Cluster implements Closeable {
                     if (controlConnection.refreshNodeInfo(host)) {
                         logger.debug("Successful reconnection to {}, setting host UP", host);
                         try {
-                            if (isHostAddition)
+                            if (isHostAddition) {
                                 onAdd(host, connection);
-                            else
+                                submitNodeListRefresh();
+                            } else
                                 onUp(host, connection);
                         } catch (InterruptedException e) {
                             Thread.currentThread().interrupt();
@@ -2226,15 +2227,15 @@ public class Cluster implements Closeable {
             return schemaRefreshRequestDebouncer.eventReceived(request);
         }
 
-        void submitNodeListRefresh() {
+        ListenableFuture<Void> submitNodeListRefresh() {
             logger.trace("Submitting node list and token map refresh");
-            nodeListRefreshRequestDebouncer.eventReceived(new NodeListRefreshRequest());
+            return nodeListRefreshRequestDebouncer.eventReceived(new NodeListRefreshRequest());
         }
 
-        void submitNodeRefresh(InetSocketAddress address, HostEvent eventType) {
+        ListenableFuture<Void> submitNodeRefresh(InetSocketAddress address, HostEvent eventType) {
             NodeRefreshRequest request = new NodeRefreshRequest(address, eventType);
             logger.trace("Submitting node refresh: {}", request);
-            nodeRefreshRequestDebouncer.eventReceived(request);
+            return nodeRefreshRequestDebouncer.eventReceived(request);
         }
 
         // refresh the schema using the provided connection, and notice the future with the provided resultset once done
@@ -2709,6 +2710,7 @@ public class Cluster implements Closeable {
                     public void runMayThrow() throws Exception {
                         if (controlConnection.refreshNodeInfo(host)) {
                             onAdd(host, null);
+                            submitNodeListRefresh();
                         } else {
                             logger.debug("Not enough info for {}, ignoring host", host);
                         }
@@ -2745,6 +2747,7 @@ public class Cluster implements Closeable {
                         if (metadata.remove(host)) {
                             logger.info("Cassandra host {} removed", host);
                             onRemove(host);
+                            submitNodeListRefresh();
                         }
                     }
                 };
--- a/driver-core/src/main/java/com/datastax/driver/core/Metadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Metadata.java
@@ -41,7 +41,7 @@ public class Metadata {
     volatile String partitioner;
     private final ConcurrentMap<InetSocketAddress, Host> hosts = new ConcurrentHashMap<InetSocketAddress, Host>();
     final ConcurrentMap<String, KeyspaceMetadata> keyspaces = new ConcurrentHashMap<String, KeyspaceMetadata>();
-    volatile TokenMap tokenMap;
+    private volatile TokenMap tokenMap;
 
     final ReentrantLock lock = new ReentrantLock();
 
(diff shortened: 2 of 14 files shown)


driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -2265,7 +2267,20 @@ public void run() {
                         if (!schemaInAgreement)
                             logger.warn("No schema agreement from live replicas after {} s. The schema may not be up to date on some nodes.", configuration.getProtocolOptions().getMaxSchemaAgreementWaitSeconds());
 
-                        ListenableFuture<Void> schemaReady = refreshSchema ? submitSchemaRefresh(targetType, targetKeyspace, targetName, targetSignature) : MoreFutures.VOID_SUCCESS;
+                        ListenableFuture<Void> schemaReady;
+                        if (refreshSchema) {
+                            schemaReady = submitSchemaRefresh(targetType, targetKeyspace, targetName, targetSignature);
+                            if (refreshNodeList) {
+                                schemaReady = Futures.transform(schemaReady, new AsyncFunction<Object, Void>() {
```
adutra: Note that a call to `submitNodeListRefresh()` is already done in `ControlConnection.refreshSchema()` for other cases (notably when a schema change event is received), but here we need to wait on that future. I think the cost of enqueueing 2 node list refresh requests is negligible compared to the advantage of having the token metadata guaranteed to be up-to-date.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -2265,7 +2267,25 @@ public void run() {
...
+                        ListenableFuture<Void> schemaReady;
+                        if (refreshSchema) {
+                            schemaReady = submitSchemaRefresh(targetType, targetKeyspace, targetName, targetSignature);
+                            // JAVA-1120: skip debouncing delay and force immediate delivery
+                            //schemaRefreshRequestDebouncer.scheduleImmediateDelivery();
+                            if (refreshNodeList) {
+                                schemaReady = Futures.transform(schemaReady, new AsyncFunction<Object, Void>() {
+                                    @Override
+                                    public ListenableFuture<Void> apply(Object input) throws Exception {
+                                        ListenableFuture<Void> nodeListRefreshReady = submitNodeListRefresh();
```
adutra: Note that a call to `submitNodeListRefresh()` is already done in `ControlConnection.refreshSchema()` for other cases (notably when a schema change event is received), but here we need to wait on that future. I think the cost of enqueueing 2 node list refresh requests is negligible compared to the advantage of having the token metadata guaranteed to be up-to-date.

adutra:
Added two new commits:

1. 1de87db suppresses recursive call to `submitNodeListRefresh()`
2. b7e9635 optimizes token metadata computation during schema resfreshes (and makes it synchronous)

Please review those carefully as they affect sensitive areas, and I might have let slip something important.

adutra:
Rebased and squashed on top of current 3.0.x.

tolbertam:
We can probably limit the usage of `TestUtils.nonDebouncingQueryOptions` in some cases with JAVA-1120.  I'll work on that, would be a good way of ensuring this fix, as these options should no longer be needed in the schema change case.

tolbertam:
Not sure why appveyor triggered a check here..please ignore that.

adutra:
Looks like CI is happy with the removal of `TestUtils.nonDebouncingQueryOptions`, and more importantly, the overall build time remained constant.

driver-core/src/main/java/com/datastax/driver/core/QueryOptions.java
```diff
@@ -299,7 +301,19 @@ public QueryOptions setMetadataEnabled(boolean enabled) {
...
+            // which will be incomplete due to the lack of keyspace metadata
+            Futures.addCallback(manager.submitNodeListRefresh(), new MoreFutures.SuccessCallback<Void>() {
+                @Override
+                public void onSuccess(Void result) {
+                    // 2. then call submitSchemaRefresh() to
+                    // refresh schema metadata and re-compute the token map
+                    // this time with information about keyspaces
+                    manager.submitSchemaRefresh(null, null, null, null);
+                }
+            });
```
olim7t, line 316: Do we need to refresh the node list here? `metadataEnabled == false` doesn't disable the handling of topology events, so the node list should be up to date.

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -633,20 +660,20 @@ private TokenMap(Token.Factory factory,
...
             }
+            this.primaryToTokens = primaryToTokens;
         }
 
-        public static TokenMap build(Token.Factory factory, Map<Host, Collection<String>> allTokens, Collection<KeyspaceMetadata> keyspaces) {
+        public static TokenMap build(Token.Factory factory, Map<Host, ? extends Collection<Token>> allTokens, Collection<KeyspaceMetadata> keyspaces) {
 
             Set<Host> hosts = allTokens.keySet();
             Map<Token, Host> tokenToPrimary = new HashMap<Token, Host>();
             Map<Host, Set<Token>> primaryToTokens = new HashMap<Host, Set<Token>>();
```
olim7t: Now that the string -> token conversion is handled upstream, this is just a deep copy of `allTokens`.

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -633,20 +660,20 @@ private TokenMap(Token.Factory factory,
                 Host host = entry.getKey();
                 host.setTokens(ImmutableSet.copyOf(entry.getValue()));
             }
+            this.primaryToTokens = primaryToTokens;
         }
 
-        public static TokenMap build(Token.Factory factory, Map<Host, Collection<String>> allTokens, Collection<KeyspaceMetadata> keyspaces) {
+        public static TokenMap build(Token.Factory factory, Map<Host, ? extends Collection<Token>> allTokens, Collection<KeyspaceMetadata> keyspaces) {
```
olim7t: I like the idea of a simplified computation when only the keyspaces have changed (not the node list). But there's still a lot of keyspace-independent stuff that's recomputed here, basically everything except `hostsToRanges` (even `replStrategyToHosts` would be nice to cache between invocations). I wonder if we should not cache more aggressively to avoid this.

driver-core/src/main/java/com/datastax/driver/core/QueryOptions.java
adutra, line 316: The hosts list is kept up-to-date, but not the token map, which is likely to be `null` at this point. The field `Metadata.tokenMap` is only ever updated if `metadataEnabled == true` (see [here](https://github.com/datastax/java-driver/blob/3.0.x/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java#L656-L657)), and the only way to build the token map from scratch for the first time is to trigger a node list refresh.

adutra:
Rebased and addressed feedback: the `TokenMap` class now caches all the information required to rebuild itself when only the schema has changed, but not the topology.

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -66,34 +64,43 @@
 
     // rebuilds the token map with the current hosts, typically when refreshing schema metadata
     void rebuildTokenMap() {
-        if (tokenMap == null) return;
-        Token.Factory factory = getTokenFactory(partitioner);
-        if (factory == null) return;
-        rebuildTokenMap(factory, tokenMap.primaryToTokens);
+        lock.lock();
```
tolbertam, line 67: Very nice!

adutra, line 67: @tolbertam FYI I added another commit with further optimizations around TokenMap re-creation.

olim7t:
Looks good to me. We cache more info than before, but it's O(numTokens) so it should still be reasonable, even for large clusters.
@adutra could you squash the commits as appropriate, and reword the messages? (no "should be" or "consider...", just imperative sentences)

driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
```diff
@@ -63,6 +63,9 @@ void refresh(Cluster cluster,
                 assert rows.keyspaces != null;
                 Map<String, KeyspaceMetadata> keyspaces = buildKeyspaces(rows, cassandraVersion, cluster);
                 updateKeyspaces(metadata, metadata.keyspaces, keyspaces, targetKeyspace);
+                // If we rebuild all from scratch or have an updated keyspace, rebuild the token map
+                // since some replication on some keyspace may have changed
+                metadata.rebuildTokenMap();
```
adutra, line 68: I am still a bit worried bout this call being synchronous, but not atomic. Here, right before the call to `rebuildTokenMap`, the field `Metadata.keyspaces` would contain the changes, but not the field `Metadata.tokenMap`. There is still a slight, albeit improbable, chance that an observer could retrieve inconsistent data. What makes it difficult to fix is that updates to `Metadata.keyspaces` are made on-the-go.

olim7t, line 68: > I am still a bit worried bout this call being synchronous, but not atomic

I agree, but fixing that is much more involved: I think the best way would be to make `Metadata` an immutable object that contains a snapshot of _all_ metadata at a given point in time. After a single call to `cluster.getMetadata()` the user would have the guarantee that all fields are consistent with each other (which is not possible currently since you call methods on a mutable object).
Internally we would handle updates by storing this object in a volatile field, and switching it with a new copy when something gets updated.
This is likely to break the API though.

PR #668, opened 2016-05-11
JAVA-1192: Make EventDebouncer settings updatable at runtime.

adutra:


driver-core/src/main/java/com/datastax/driver/core/EventDebouncer.java
```diff
@@ -151,9 +155,9 @@ private void completeAllPendingFutures() {
 
         if (state == State.RUNNING) {
             int count = eventCount.get();
-            if (count < maxPendingEvents) {
+            if (count < maxPendingEvents()) {
                 scheduleDelayedDelivery();
-            } else if (count == maxPendingEvents) {
+            } else if (count == maxPendingEvents()) {
                 scheduleImmediateDelivery();
```
olim7t, line 162: If `maxPendingEvents()` is decreased dynamically, we might miss that condition and not schedule an immediate delivery (since now `count > maxPendingEvents()`).
Not sure if that's a huge issue since a delayed delivery should still be scheduled.

adutra, line 162: Good catch, I updated the code to only call `maxPendingEvents()` once.

adutra:
Squashed and rebased on top of current 3.0.

driver-core/src/main/java/com/datastax/driver/core/EventDebouncer.java
```diff
@@ -151,6 +155,7 @@ private void completeAllPendingFutures() {
 
         if (state == State.RUNNING) {
             int count = eventCount.get();
+            int maxPendingEvents = maxPendingEvents();
```
olim7t, line 158: To follow up on my previous comment about maxPendingEvents, the issue is not only when it changes while the method is executing, but also when it changes between calls, for example:
- maxPendingEvents = 100, 49 events are enqueued
- eventReceived is called, count increments to 50 => we schedule a delayed delivery
- maxPendingEvents is changed to 50
- eventReceived is called, count increments to 51 => we do nothing

In this scenario we go over the max without ever triggering the condition `(count == maxPendingEvents)`, so the immediate delivery is never scheduled. As I noted before, a scheduled delivery is still planned, so maybe we can overlook this as an edge case.

adutra, line 158: Hmm right. It could be a concern because the scheduled delivery could happen before your 51th event is added to the queue, in which case it would remain in the queue until the next event is received. I'm thinking about changing the if condition to `count >= maxPendingEvents`. That would certainly trigger too many immediate deliveries, but on the other hand, no event would ever remain in the queue without triggering a delivery (delayed or immediate).

EDIT: nevermind, I double checked and it's not possible for an event to remain indefinitely in the queue in this case, it would be delivered by a delayed delivery (either the one triggered by the 50th element, or the next one that would be triggered [here](https://github.com/datastax/java-driver/blob/java1192/driver-core/src/main/java/com/datastax/driver/core/EventDebouncer.java#L247)).

So I think we can safely dismiss this case.

PR #670, opened 2016-05-12
Appveyor CI enablement

tolbertam:
- Enables Appveyor CI for running integration tests against windows.  Initial goal is to run 'short' suite against jdk8 w/ C\* 3.0.5, 'unit' suites against jdk6 and jdk7.  Takes about ~50 minutes and should run rather reliably.
- Eventually to enable daily or weekly full suite runs against C\* 2.2+ once more confidence is established.
- Provides validation in ensuring that code that requires native os calls (netty-tcnative, jnr, etc.) works on windows.  Prebuilds libapr (if not present) and uses provided openssl for validating netty-tcnative.

Changes:
- Use powershell and use `--quiet-windows` flag on start (new C\* versions require this)
- Some platform specific quoting adjustments since windows is fickle about spaces.
- Some refactoring in `HostConnectionPoolTest` using `ConditionChecker` to ensure pools grow/shrink.
- Mark some tests as long that should be as they require multi-node clusters or take a long time.
- Add `exactRequired` field to `CassandraVersion` and `DseVersion` annotations to indicate that a test requires a specific version of Cassandra to run.
- Upgrade netty-tcnative test dependency from 1.1.33.Fork17 to 1.1.33.Fork18 as Fork17 does not work on windows.
- Upgrade scassandra to 1.0.9 (reduced thread count improves test stability on windows).

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
@@ -193,7 +193,7 @@ public class CCMBridge implements CCMAccess {
      *
      * @return <code>true</code> if the operating system is a Windows one, <code>false</code> otherwise.
      */
-    private static boolean isWindows() {
+    public static boolean isWindows() {
 
         String osName = System.getProperty("os.name");
         return osName != null && osName.startsWith("Windows");
--- a/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
@@ -167,7 +167,7 @@ public class CCMBridge implements CCMAccess {
         }
 
         if (isWindows()) {
-            CCM_COMMAND = "cmd /c ccm.py";
+            CCM_COMMAND = "powershell.exe -ExecutionPolicy Unrestricted ccm.py";
         } else {
             CCM_COMMAND = "ccm";
         }
@@ -386,7 +386,11 @@ public class CCMBridge implements CCMAccess {
         if (logger.isDebugEnabled())
             logger.debug("Starting: {} - free memory: {} MB", this, TestUtils.getFreeMemoryMB());
         try {
-            execute(CCM_COMMAND + " start --wait-other-notice --wait-for-binary-proto" + jvmArgs);
+            String cmd = CCM_COMMAND + " start --wait-other-notice --wait-for-binary-proto" + jvmArgs;
+            if (isWindows() && this.version.compareTo(VersionNumber.parse("2.2.4")) >= 0) {
+                cmd += " --quiet-windows";
+            }
+            execute(cmd);
         } catch (CCMException e) {
             logger.error("Could not start " + this, e);
             logger.error("CCM output:\n{}", e.getOut());
@@ -444,7 +448,11 @@ public class CCMBridge implements CCMAccess {
     public void start(int n) {
         logger.debug(String.format("Starting: node %s (%s%s:%s) in %s", n, TestUtils.IP_PREFIX, n, binaryPort, this));
         try {
-            execute(CCM_COMMAND + " node%d start --wait-other-notice --wait-for-binary-proto" + jvmArgs, n);
+            String cmd = CCM_COMMAND + " node%d start --wait-other-notice --wait-for-binary-proto" + jvmArgs;
+            if (isWindows() && this.version.compareTo(VersionNumber.parse("2.2.4")) >= 0) {
+                cmd += " --quiet-windows";
+            }
+            execute(cmd, n);
         } catch (CCMException e) {
             logger.error(String.format("Could not start node %s in %s", n, this), e);
             logger.error("CCM output:\n{}", e.getOut());
@@ -875,9 +883,14 @@ public class CCMBridge implements CCMAccess {
 
         private String joinJvmArgs() {
             StringBuilder allJvmArgs = new StringBuilder("");
+            String quote = isWindows() ? "\"" : "";
             for (String jvmArg : jvmArgs) {
-                allJvmArgs.append(" --jvm_arg=");
+                // Windows requires jvm arguments to be quoted, while *nix requires unquoted.
+                allJvmArgs.append(" ");
+                allJvmArgs.append(quote);
+                allJvmArgs.append("--jvm_arg=");
                 allJvmArgs.append(randomizePorts(jvmArg));
+                allJvmArgs.append(quote);
             }
             return allJvmArgs.toString();
         }
--- a/driver-core/src/test/java/com/datastax/driver/core/ProtocolV1Test.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/ProtocolV1Test.java
@@ -16,6 +16,7 @@
 package com.datastax.driver.core;
 
 import com.datastax.driver.core.exceptions.UnsupportedFeatureException;
+import org.testng.SkipException;
 import org.testng.annotations.Test;
 
 import static com.datastax.driver.core.ProtocolVersion.V1;
@@ -31,6 +32,13 @@ public class ProtocolV1Test extends CCMTestsSupport {
     public Cluster.Builder createClusterBuilder() {
         return super.createClusterBuilder()
                 .withProtocolVersion(V1);
+    }
+
+    @Override
+    public void beforeTestClass(Object testInstance) throws Exception {
+        if (CCMBridge.isWindows())
+            throw new SkipException("C* 1.2 is not supported on Windows.");
+        super.beforeTestClass(testInstance);
     }
 
     /**
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Datastax Java Driver for Apache Cassandra
 
-[![Build Status](https://travis-ci.org/datastax/java-driver.svg?branch=3.0.x)](https://travis-ci.org/datastax/java-driver)
+[![Build Status](https://travis-ci.org/datastax/java-driver.svg?branch=3.0.x)](https://travis-ci.org/datastax/java-driver) [![Build status](https://ci.appveyor.com/api/projects/status/717ysg5l1si1h3pi/branch/3.0.x?svg=true)](https://ci.appveyor.com/project/DataStax/java-driver/branch/3.0.x)
 
 *If you're reading this on github.com, please note that this is the readme
 for the development version and that some features described here might
--- a/ci/appveyor.ps1
+++ b/ci/appveyor.ps1
@@ -1,0 +1,129 @@
+Add-Type -AssemblyName System.IO.Compression.FileSystem
+
+$dep_dir="C:\Users\appveyor\deps"
+If (!(Test-Path $dep_dir)) {
+  Write-Host "Creating $($dep_dir)"
+  New-Item -Path $dep_dir -ItemType Directory -Force
+}
+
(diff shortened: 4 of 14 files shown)


tolbertam:
Targeting for 3.0.3 as would be good to include getting this stable as part of the 3.0.3 test effort.

tolbertam:
tests using netty-tcnative for openssl broke on windows with 4.0.37 / Fork17.  Not sure why yet, will look into it.

tolbertam:
There are a few finicky tests that need to be solidified before i'd feel comfortable pulling this in and made part of PR checks.  I'll fix these and remove WIP once I feel confident.

tolbertam:
Removing WIP tag, I think this is now stable enough to be deemed ready for review.   I will set it up to rebuild constantly to tease out any other tests we can solidify, but there are some that may be unavoidable (like CCM failing to start for whatever reason).

driver-core/src/test/java/com/datastax/driver/core/ProtocolV1Test.java
```diff
@@ -24,7 +25,7 @@
 /**
  * Tests targeting protocol v1 specifically.
  */
-@CCMConfig(version = "1.2.19")
+@CassandraVersion(major = 1.2, exactRequired = true)
```
olim7t: So `exactRequired` skips the test if major.minor > 1.2?

driver-core/src/test/java/com/datastax/driver/core/SchemaChangesTest.java
```diff
@@ -482,8 +484,7 @@ public void should_notify_of_keyspace_drop(String keyspace) throws InterruptedEx
         }
         assert added != null;
         refreshSchema();
-        for (Metadata m : metadatas())
-            assertThat(m.getReplicas(keyspace, Bytes.fromHexString("0xCAFEBABE"))).isNotEmpty();
+        replicasPresentForKeyspace(keyspace).becomesTrue();
```
olim7t: Sorry but I'm not a big fan of this, kind of hard to read.

driver-core/src/test/java/com/datastax/driver/core/SchemaChangesTest.java
```diff
@@ -500,8 +501,8 @@ public void should_notify_of_keyspace_drop(String keyspace) throws InterruptedEx
         refreshSchema();
         for (Metadata m : metadatas()) {
             assertThat(m.getKeyspace(keyspace)).isNull();
-            assertThat(m.getReplicas(keyspace, Bytes.fromHexString("0xCAFEBABE"))).isEmpty();
         }
+        replicasPresentForKeyspace(keyspace).becomesFalse();
```
olim7t: The condition checks that the keyspace is present in the token map of all Clusters. So the negative would be "the keyspace is absent for at least one Cluster". Is it really what we want, or rather "the keyspace is absent for _all_ Clusters"?

driver-core/src/test/java/com/datastax/driver/core/ProtocolV1Test.java
tolbertam: Effectively yes.   It skips any cassandra versions that doesn't match the criteria exactly.  i.e. if `major` is 1.2, only C\* versions matching 1.2.X will qualify.   If `major` is `1.2`, minor is `19`, only 1.2.19 will qualify.   I agree it's a bit crude.  @adutra has a good solution for this that allows a version range specification that I think will be much better/flexible that we can use in the future.

driver-core/src/test/java/com/datastax/driver/core/SchemaChangesTest.java
tolbertam: You are right, that doesn't seem like a complete validation.  I'll fix it.

driver-core/src/test/java/com/datastax/driver/core/SchemaChangesTest.java
tolbertam: Based on your comment below, I think I will need to separate out the validation between the true/false cases anyways.  Were you more against the naming strategy?  Or the gist of how the validation is done? (or both :P?)

tolbertam: Come to think of it, @adutra has addressed all of this in #666, so i'll wait for that to merge and rebase and take in his changes.

tolbertam:
Something like ~7/85 of the builds I ran with this over the last week failed some of which I've addressed, so I think we may rarely see some intermittent failures, but I think this is tolerable for merge (after we got the other 3.0.3 tickets merged and this has been reviewed).

driver-core/src/test/java/com/datastax/driver/core/SchemaChangesTest.java
tolbertam: I deleted this change as it is no longer needed after #666 was merged.

driver-core/src/test/java/com/datastax/driver/core/SchemaChangesTest.java
tolbertam: No longer applicable after #666.

tolbertam:
Rebased against 3.0.x and deleted commit that made changes to `SchemaChangesTest` since it was no longer needed after #666.  Should be ready pending review and successful run.

driver-core/src/test/java/com/datastax/driver/core/utils/CassandraVersion.java
```diff
@@ -34,7 +34,12 @@
     /**
      * @return The minor version required to execute this test, i.e. "0"
      */
-    int minor() default 0;
+    int minor() default -1;
+
+    /**
+     * @return Whether or not the exact major version is exactly required to run this test.
+     */
+    boolean[] exactRequired() default {false};
```
adutra: Why do you need this attribute to be a boolean array? It seems a simple boolean is enough.

driver-core/src/test/java/com/datastax/driver/core/ProtocolV1Test.java
adutra: So if I get it right, this test used to always be run against a 1.2.19 cluster, regardless of the default CCM version (`cassandra.version`), whereas now it would only run if the default CCM version _is_ 1.2.X, and would be skipped for all other configurations.

I'm fine with that, but indeed my feeling is that we should avoid skipping tests. IMHO there should be a default CCM version for the test suite that would be used for all tests that do not declare any specific CCM requirement, but if a test – like this one – has very specific requirements, such as a specific version or a specific version range, then a non-default CCM cluster should be spun for it, rather than simply skipping it. I basically came to this conclusion seeing how many times we realize very late that a test fails for a specific C\* version, just because the test was being skipped on our local machines.

In the new toolbox repository, things are being done that way (there is no test skipping). I understand that in a Jenkins matrix build that would make some tests be executed several times always against the same C\* version, but I'd rather have that, than risking to have some tests maybe being skipped everywhere.

driver-core/src/test/java/com/datastax/driver/core/utils/CassandraVersion.java
tolbertam: It's needed for defaulting the value if unspecified I believe.   The same thing is done in `@CCMConfig`

driver-core/src/test/java/com/datastax/driver/core/ProtocolV1Test.java
tolbertam: Right,  I just need a way of skipping this particular test since 1.2.x does not work well in CCM on windows.  I sort of agree with what you are saying though, should I find an alternative way for skipping this particular test on windows?

driver-core/src/test/java/com/datastax/driver/core/utils/CassandraVersion.java
adutra: `@CCMConfig` is complex in that it needs to distinguish `true`, `false` and `unset`, and so it uses boolean arrays = `{true}`, `{false}` and `{}`. But here I think you should only need:

``` java
boolean exactRequired() default false;
```

driver-core/src/test/java/com/datastax/driver/core/ProtocolV1Test.java
adutra: Oh. So we need OS requirements too, I will look into that for the new test framework. 

Ideally, we would have to amend `@CCMConfig` to add a new attribute, e.g. `os = {NIX, MAC_OS}` and have the test skipped/ignored if the os doesn't match.

But that's complex, so in the meanwhile I think that what you currently have done so far is enough.

tolbertam: I think to keep things really simple, I can just put a platform check in this test itself since I only need it for the 1 test.   This way we don't need to pollute `@CassandraVersion/@DseVersion/@CCMConfig` for the time being.  I'm sure we'll want it for more at some point, but by that time we'll be using the new test framework.

tolbertam: Hrmm, unfortunately it doesn't seem like this is an option because there isn't a way for me prevent it creating the CCM cluster other than at annotation level, so may need to keep this in here for now.

driver-core/src/test/java/com/datastax/driver/core/utils/CassandraVersion.java
tolbertam: Fixed and rebased commit 5bc921e (figured was simple enough to fix inline).

driver-core/src/test/java/com/datastax/driver/core/ProtocolV1Test.java
adutra: You need to hook in early enough to prevent the CCM cluster from being created. Try this:

``` java
    @Override
    public void beforeTestClass(Object testInstance) throws Exception {
        if (CCMBridge.isWindows())
            throw new SkipException("Only works on proper OS's");
        super.beforeTestClass(testInstance);
    }
```

(assuming `CCMBridge.isWindows()` is made accessible)

tolbertam: Ah I see,  I added a `@BeforeClass` annotation and it got invoked after `CCMBridge` kicked in, but I see how overriding this could do the trick.  I'll do that.

tolbertam: That worked great, thanks!  Pushed as 2d6c76c and deleted this commit.

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
```diff
@@ -876,8 +884,13 @@ public int weight() {
         private String joinJvmArgs() {
             StringBuilder allJvmArgs = new StringBuilder("");
             for (String jvmArg : jvmArgs) {
-                allJvmArgs.append(" --jvm_arg=");
+                // Windows requires jvm arguments to be quoted, while *nix requires unquoted.
+                String quote = isWindows() ? "\"" : "";
```
adutra: This could be moved outside the for loop.

tolbertam: thanks, will fix.

tolbertam: Done, rebased it into 317b4d3 since it was a small change.

PR #671, opened 2016-05-18

adutra:


driver-core/src/main/java/com/datastax/driver/core/BoundStatement.java
```diff
@@ -270,7 +270,7 @@ public BoundStatement setRoutingKey(ByteBuffer routingKey) {
      */
     @Override
     public String getKeyspace() {
-        return statement.getPreparedId().metadata.size() == 0 ? null : statement.getPreparedId().metadata.getKeyspace(0);
+        return statement.getPreparedId().getVariables().size() == 0 ? null : statement.getPreparedId().getVariables().getKeyspace(0);
```
adutra, line 273: With the refactoring I came across this line and thought: this would return `null` for statements without bound variables. But we could also inspect the resultset metadata (i.e. the returned columns) to find out the keyspace.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -2133,6 +2133,13 @@ public PreparedStatement addPrepared(PreparedStatement stmt) {
             return stmt;
         }
 
+        void removePrepared(PreparedStatement stmt) {
+            if (!preparedQueries.remove(stmt.getPreparedId().getId(), stmt)) {
+                // we probably raced
+                logger.trace("Failed to remove statement {}", stmt.getPreparedId().getId());
```
adutra, line 2139: This would happen if 2 threads prepare the same statement simultaneously.

driver-core/src/main/java/com/datastax/driver/core/DefaultPreparedStatement.java
```diff
@@ -52,72 +49,13 @@ private DefaultPreparedStatement(PreparedId id, String query, String queryKeyspa
 
     static DefaultPreparedStatement fromMessage(Responses.Result.Prepared msg, Cluster cluster, String query, String queryKeyspace) {
         assert msg.metadata.columns != null;
-
```
adutra, line 55: Moved the `PreparedId` computation code to that class.

driver-core/src/main/java/com/datastax/driver/core/PreparedId.java
```diff
@@ -15,25 +15,125 @@
...
+
+    int[] getRoutingKeyIndexes() {
+        return routingKeyIndexes;
+    }
+
+    ProtocolVersion getProtocolVersion() {
+        return protocolVersion;
+    }
+
+    void swap(PreparedId updated) {
```
adutra, line 73: This is a bit hackish but has the advantage of saving us from creating a subinterface for `PreparedStatement`. Note that reads and writes are not atomic (it is possible to read an inconsistent state if state is being updated); we _could_ make them atomic if this is of a concern.

olim7t, line 73: Couldn't we swap the `preparedId` field in `DefaultPreparedStatement` instead? This way we could keep `PreparedId` immutable, which might makes things easier.

olim7t:
On a side note, could we keep the convention of naming branches after ticket numbers, as in `java1196` (lowercase, no dash). This avoids looking up the branch name each time.

driver-core/src/main/java/com/datastax/driver/core/PreparedId.java
adutra, line 73: But what if the `PreparedStatement` isn't an instance of `DefaultPreparedStatement`?

adutra:
> On a side note, could we keep the convention of naming branches after ticket numbers, as in java1196 (lowercase, no dash).

Sure, Alex wasn't aware of the naming convention, but I will take care of renaming the branch.

adutra:
Closing in favor of #675 (branch renaming).

PR #675, opened 2016-05-19

adutra:


driver-core/src/main/java/com/datastax/driver/core/BoundStatement.java
```diff
@@ -270,7 +270,7 @@ public BoundStatement setRoutingKey(ByteBuffer routingKey) {
      */
     @Override
     public String getKeyspace() {
-        return statement.getPreparedId().metadata.size() == 0 ? null : statement.getPreparedId().metadata.getKeyspace(0);
+        return statement.getPreparedId().getVariables().size() == 0 ? null : statement.getPreparedId().getVariables().getKeyspace(0);
```
adutra: With the refactoring I came across this line and thought: this would return `null` for statements without bound variables. But we could also inspect the resultset metadata (i.e. the returned columns) to find out the keyspace.

driver-core/src/main/java/com/datastax/driver/core/PreparedId.java
```diff
@@ -15,25 +15,125 @@
...
+
+    int[] getRoutingKeyIndexes() {
+        return routingKeyIndexes;
+    }
+
+    ProtocolVersion getProtocolVersion() {
+        return protocolVersion;
+    }
+
+    void swap(PreparedId updated) {
```
adutra, line 89: Resuming the discussion:

> > > @adutra This is a bit hackish but has the advantage of saving us from creating a subinterface for `PreparedStatement`. Note that reads and writes are not atomic (it is possible to read an inconsistent state if state is being updated); we could make them atomic if this is of a concern.
> > 
> > @olim7t Couldn't we swap the preparedId field in `DefaultPreparedStatement` instead? This way we could keep `PreparedId` immutable, which might makes things easier.
> 
> @adutra But what if the `PreparedStatement` isn't an instance of `DefaultPreparedStatement`?

adutra:
@ifesdjeen I presume you finished the remaining work on a local branch of yours, is that correct? Can we then close this pull request?

ifesdjeen:
This is correct. I think I've borrowed a couple of lines from here, but since we ended up taking a completely different path, I mostly did it from scratch...

PR #676, opened 2016-05-20
JAVA-1161: Preserve full time zone info in ZonedDateTimeCodec and DateTimeCodec

adutra:
The present PR proposes to store zone IDs (such as `UTC`, `GMT`, or `Europe/Paris`) instead of raw offsets (such as `+00:00` or `-07:00`), when this information is available.

Zone IDs carry daylight savings time information, which offsets do not, and so can be more precise. One small drawback is that zone IDs are not ISO-8601 normalized, but I guess we can live with that.

I also strived to make both affected codecs, `ZonedDateTimeCodec` and `DateTimeCodec`, compatible with each other, i.e., a tuple serialized by the former should be deserializable by the latter.

And finally, I opened `ZonedDateTimeCodec` to custom formatters, in case users would want to format or parse in a specific format. I did the same for `DateTimeCodec`  but only for the timestamp part, mainly because there is no proper parser/formatter for time zones: they _have_ to be parsed by the built-in parser in `DateTimeZone.forID()`, which does not leave much room for customization.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -6,6 +6,7 @@
 - [improvement] JAVA-743: Add JSON support to QueryBuilder.
 - [improvement] JAVA-1233: Update HdrHistogram to 2.1.9.
 - [improvement] JAVA-1233: Update Snappy to 1.1.2.6.
+- [bug] JAVA-1161: Preserve full time zone info in ZonedDateTimeCodec and DateTimeCodec.
 
 Merged from 3.0.x branch:
 
--- a/driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/ZonedDateTimeCodec.java
+++ b/driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/ZonedDateTimeCodec.java
@@ -34,14 +34,19 @@ import static com.google.common.base.Preconditions.checkArgument;
  * Since Cassandra's <code>timestamp</code> type preserves only
  * milliseconds since epoch, any timezone information
  * would normally be lost. By using a
- * <code>tuple&lt;timestamp,varchar&gt;</code> a timezone ID can be
+ * <code>tuple&lt;timestamp,varchar&gt;</code> a timezone can be
  * persisted in the <code>varchar</code> field such that when the
- * value is deserialized the timezone is
- * preserved.
- * <p/>
- * <strong>IMPORTANT</strong>: this codec's {@link #format(Object) format} method formats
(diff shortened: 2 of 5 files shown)


adutra:

adutra:
Rebased again. Il would like to have the team's opinion about how to handle warnings in IntelliJ when a class uses APIs from Java 7+. The default behavior is to raise a compile warning, which is very useful. These errors can be suppressed with `@SuppressWarnings("Since15")` – which is why all classes in `com.datastax.driver.extras.codecs.jdk8` have this annotation. Unfortunately, this does not apply to imports, so we have to avoid imports and specify the class FQDN each time (example [here](https://github.com/datastax/java-driver/blob/3feeffadf53688a5433330345824175ad149fae1/driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/ZonedDateTimeCodec.java#L61)).
But in hindsight, it makes the whole code quite verbose, so maybe a better approach would be to raise the alert from error to warning or weak warning. Thoughts?

tolbertam:
> The default behavior is to raise a compile warning, which is very useful. These errors can be suppressed with @SuppressWarnings("Since15") – which is why all classes in com.datastax.driver.extras.codecs.jdk8 have this annotation.

If I remove the FQDN from `DateTimeFormatter` I don't get the warning, maybe this is an IntellIJ bug that has since been fixed?  I'm on version 15.0.6.

adutra:
> maybe this is an IntellIJ bug that has since been fixed? I'm on version 15.0.6.

I'm on – err – 2016.1.3 (they completely changed their versioning scheme, so I'm a bit lost).

By default, under "Preferences > Editor > Inspections > Java > Java language level migration aids", I had an inspection called "Usages of API which isn't available at the configured language level", set to Error severity.

tolbertam:
> By default, under "Preferences > Editor > Inspections > Java > Java language level migration aids", I had an inspection called "Usages of API which isn't available at the configured language level", set to Error severity.

I see, I don't have that option, but I do have `Usages of API documented as @since 1.5 (1.6|1.7)` which I bet is the same thing.  It's currently unchecked but when I turn it on I see the errors you are referencing on the import statement.

olim7t:
I've tried `noinspection Since15` comments, but I can't find a way to apply them to the whole file. Another way would be to exclude the files using the Settings dialog, but that means storing IntelliJ settings in SCM which I'm not a fan of.
This is a minor problem, I vote to keep the fully-qualified names for now.

PR #677, opened 2016-05-22
IF_NOT_EXISTS mapper option for SAVE queries.

raggzy:
Added IF_NOT_EXISTS mapper option for save operations.

Useful in cases when you want to insert something atomically, but you are not sure it exists or not. And your app logic actually depends on  whether it was inserted or not. 

You still can build Insert by hands, but this way looks much much prettier.

For example, user registration process, your User table has PrimaryKey `username`. You try to register  user, but if 'it exists', e.g. wasApplied returned false - you tell about that on UI with error/warning "User already exists". Until try to insert, you'd never know that.

### DIFF
--- a/driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
+++ b/driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
@@ -736,7 +736,7 @@ public class Mapper<T> {
      */
     public static abstract class Option {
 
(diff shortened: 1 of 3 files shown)


adutra:
This looks very similar to #626. Also already logged as [JAVA-1174](https://datastax-oss.atlassian.net/browse/JAVA-1174).

raggzy:
Yeah, it is the same. Sorry, missed it out. I'd love to see any of those in next release :)

olim7t:
Closing in favor of #626 then.

PR #678, opened 2016-05-23
JAVA-1126: Mapper should be resilient when table or UDT is modified.

adutra:
This PR proposes the following enhancements:
- Each mapping manager subscribes to schema change events and reacts to table and UDT removed / modified notifications (could extend to keyspace notifications too); the basic reaction consists of a log waning;
- For modified UDTs, the manager also tries to update the previously registered `MappedUDTCodec` with a new one;
- Column names and field names are validated upfront (upon creation of a `Mapper`) in a fail-fast attempt to detect wrong configurations;
- Existence of tables and UDTs are also validated upfront.

This however cannot prevent existing `Mapper` instances from misbehaving, i.e. if a user creates a `Mapper` instance, then issues an `ALTER` statement, that `Mapper` instance would be stale and most likely, unusable.


driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java
```diff
@@ -74,6 +77,74 @@ public MappingManager(Session session, ProtocolVersion protocolVersion) {
         // which nodes might join the cluster later.
         // At least if protocol >=2 we know there won't be any 1.2 nodes ever.
         this.isCassandraV1 = (protocolVersion == ProtocolVersion.V1);
+        session.getCluster().register(new SchemaChangeListenerBase() {
+
+            @Override
+            public void onTableRemoved(TableMetadata table) {
+                synchronized (mappers) {
+                    Iterator<Mapper<?>> it = mappers.values().iterator();
+                    while (it.hasNext()) {
```
olim7t, line 83: Why not a regular for-each loop?

olim7t, line 83: Nevermind, I hadn't noticed the `remove()` call.

driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java
```diff
@@ -74,6 +77,74 @@ public MappingManager(Session session, ProtocolVersion protocolVersion) {
...
+            @Override
+            public void onTableChanged(TableMetadata current, TableMetadata previous) {
+                synchronized (mappers) {
+                    Iterator<Mapper<?>> it = mappers.values().iterator();
+                    while (it.hasNext()) {
+                        Mapper<?> mapper = it.next();
+                        if (previous.equals(mapper.getTableMetadata())) {
+                            LOGGER.warn("Table {} has been altered; existing mappers for @Entity annotated {} might not work properly anymore",
+                                    previous.getName(), mapper.getMappedClass());
+                            it.remove();
```
olim7t, line 102: Is the rationale that the next call to `mapper()` will regenerate it?

driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java
```diff
@@ -74,6 +77,74 @@ public MappingManager(Session session, ProtocolVersion protocolVersion) {
...
+                    Iterator<MappedUDTCodec<?>> it = udtCodecs.values().iterator();
+                    while (it.hasNext()) {
+                        MappedUDTCodec<?> codec = it.next();
+                        if (previous.equals(codec.getCqlType())) {
+                            LOGGER.warn("User type {} has been altered; existing mappers for @UDT annotated {} might not work properly anymore",
+                                    previous, codec.getUdtClass());
+                            it.remove();
+                            // try to register an updated version of the previous codec
+                            try {
+                                getUDTCodec(codec.getUdtClass());
```
olim7t: This mutates the map. Aren't we risking a `ConcurrentModificationException` here? We could memorize the codecs to update and do it outside of the iteration.

driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java
adutra, line 102: Yes, correct. No need to regenerate it upfront imo – contrary to what happens with udts.

adutra:
Squashed and rebased on top of current 3.0.x.

driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java
```diff
@@ -74,6 +74,78 @@ public MappingManager(Session session, ProtocolVersion protocolVersion) {
...
+                        if (mapper.getTableMetadata().equals(table)) {
+                            LOGGER.error("Table {} has been removed; existing mappers for @Entity annotated {} will not work anymore", table.getName(), mapper.getMappedClass());
+                            it.remove();
+                        }
+                    }
+                }
+            }
+
+            @Override
+            public void onTableChanged(TableMetadata current, TableMetadata previous) {
```
tolbertam, line 94: On the case of a user type being changed, it looks like it's removed from mappers, but then readded via `getUDTCodec`, but this doesn't do the same for table via `getMapper`, should it?

tolbertam, line 94: Ah maybe it is needed in the UDT case because it updates the Codec, where no analog is needed for table.  I suppose only benefit is maybe we'd be able to log the error here instead of when the user tries to use the mapper.

tolbertam, line 94: I suppose the other issue is replacing the Mapper doesn't do a lot of good since the user would still hold a reference to the old one so it makes sense to not attempt to replace it.

driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java
```diff
@@ -74,6 +74,78 @@ public MappingManager(Session session, ProtocolVersion protocolVersion) {
...
+                            LOGGER.error("User type {} has been removed; existing mappers for @UDT annotated {} will not work anymore",
+                                    type, codec.getUdtClass());
+                            it.remove();
+                        }
+                    }
+                }
+            }
+
+            @Override
+            public void onUserTypeChanged(UserType current, UserType previous) {
```
tolbertam, line 124: Wondering, I don't think what I'm about to suggest is necessary, but just curious.

If a UDT is changed, shouldn't we also in theory clear out any `Mapper` that has a field that maps to that type?  Alternatively we could update the underlying `ColumnMapper` for that field.  I don't think it's necessary as if the change is compatible it'll keep working, if not it'll stop working and there's no way to fix it (other than updating your code).

tolbertam, line 124: Turns out my point is moot, apparently when the UDT changes, you also get a table change event too, very nice!

```
    27802  [cluster1-worker-0] WARN  com.datastax.driver.mapping.MappingManager  - Table users has been altered; existing mappers for @Entity annotated class com.datastax.driver.mapping.MapperUDTTest$User might not work properly anymore
    27803  [cluster1-worker-0] WARN  com.datastax.driver.mapping.MappingManager  - User type frozen<ks_1.address> has been altered; existing mappers for @UDT annotated class com.datastax.driver.mapping.MapperUDTTest$Address might not work properly anymore
```

tolbertam:
rebased (3.0.x had some test fixes) and added some tests and documentation.   I also made a slight change so `IllegalArgumentException` is thrown instead of `IllegalStateException` in the case where you attempt to create a UDT codec where the UDT doesn't exist in metadata to be consistent with other error cases.  Feel free to squash.

lgtm 👍 assuming CI looks good.

driver-mapping/src/main/java/com/datastax/driver/mapping/AnnotationParser.java
```diff
@@ -106,18 +112,24 @@ private AnnotationParser() {
         validateOrder(pks, "@PartitionKey");
         validateOrder(ccs, "@ClusteringColumn");
 
-        mapper.addColumns(createColumnMappers(pks, factory, mapper.entityClass, mappingManager, columnCounter),
-                createColumnMappers(ccs, factory, mapper.entityClass, mappingManager, columnCounter),
-                createColumnMappers(rgs, factory, mapper.entityClass, mappingManager, columnCounter));
+        mapper.addColumns(
+                createColumnMappers(pks, factory, mapper.entityClass, mappingManager, columnCounter, tableMetadata, ksName, tableName),
+                createColumnMappers(ccs, factory, mapper.entityClass, mappingManager, columnCounter, tableMetadata, ksName, tableName),
+                createColumnMappers(rgs, factory, mapper.entityClass, mappingManager, columnCounter, tableMetadata, ksName, tableName));
```
olim7t, line 118: `ksName` and `tableName` could be retrieved from `tableMetadata` instead of passing them as parameters (also applies to `createFieldMappers` below).

adutra:
> added some tests and documentation.

Great tests, thanks for that!

driver-mapping/src/main/java/com/datastax/driver/mapping/AnnotationParser.java
adutra, line 118: That's true but frankly I prefer to leave it as is for practical reasons:
1. Quoting of case-sensitive identifiers is not very consistent, and is done in different places, so should we quote or not the result of `TableMetadata.getName()`?  The code coverage for case-sensitive use cases is not very good, so let's not introduce regressions for that.
2. In #686 I added specific tests for case-sensitiveness and also applied a methodic quoting of identifiers). See btw JAVA-564.
3. The less we modify this class the easier it will be to merge it with JAVA-541 :)

PR #679, opened 2016-05-24
Allow to set default value when getting primitive type value is null

jiafu1115:
Motivation:

When we got primitive type such as calling getInt(int i),
we would get ‘0’ when the value is null. We can’t set the default value
if the value is null when expected the value isn't 0/false .
If we want to distinguish 0 and null , we should use isNull() or get(i,
Integer.class). It isn’t convenient for coding.
So add new methods to allow to setting the default value when got value
is Null.

Modifications:

add extra new methods with default value such as: 
public int getInt(int i, int defaultValue);

Result:

More flexible and convenient.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/AbstractGettableByIndexData.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/AbstractGettableByIndexData.java
@@ -96,10 +96,18 @@ abstract class AbstractGettableByIndexData implements GettableByIndexData {
      */
     @Override
     public boolean getBool(int i) {
+         return getBool(i, false);
+    }
+    
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean getBool(int i, boolean defaultValue) {
         ByteBuffer value = getValue(i);
         TypeCodec<Boolean> codec = codecFor(i, Boolean.class);
         if (codec instanceof TypeCodec.PrimitiveBooleanCodec)
-            return ((TypeCodec.PrimitiveBooleanCodec) codec).deserializeNoBoxing(value, protocolVersion);
+            return ((TypeCodec.PrimitiveBooleanCodec) codec).deserializeNoBoxing(value, protocolVersion, defaultValue);
         else
             return codec.deserialize(value, protocolVersion);
     }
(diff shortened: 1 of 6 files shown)


jiafu1115:
@olim7t  why need to check the new added method for clirr-maven-plugin.
What's more, can you help to review this PR? 3ks in advanced.

[ERROR] 7012: com.datastax.driver.core.GettableByIndexData: Method 'public boolean getBool(int, boolean)' has been added to an interface
[ERROR] 7012: com.datastax.driver.core.GettableByIndexData: Method 'public byte getByte(int, byte)' has been added to an interface
[ERROR] 7012: com.datastax.driver.core.GettableByIndexData: Method 'public double getDouble(int, double)' has been added to an interface
[ERROR] 7012: com.datastax.driver.core.GettableByIndexData: Method 'public float getFloat(int, float)' has been added to an interface
[ERROR] 7012: com.datastax.driver.core.GettableByIndexData: Method 'public int getInt(int, int)' has been added to an interface
[ERROR] 7012: com.datastax.driver.core.GettableByIndexData: Method 'public long getLong(int, long)' has been added to an interface
[ERROR] 7012: com.datastax.driver.core.GettableByIndexData: Method 'public short getShort(int,

olim7t:
Clirr complains because adding a method to an interface is a breaking change: if a 3rd party wrote their own implementation (for example to wrap the driver's type), then their code doesn't compile anymore when they upgrade the driver.

Regarding the change itself, I think there is a fine line between convenience and spreading the API too thin. Here we're talking about replacing:

``` java
row.isNull("c") ? 0 : row.getInt("c")
```

With:

``` java
row.getInt("c", 0)
```

I don't understand how it's more flexible. I agree it's slightly more convenient, but IMHO it's not worth the added API clutter (and breaking change we have to deal with).

jiafu1115:
@olim7t 

Thanks for your comment. Everybody wants to make code clean/simple/thin. So when I use the driver's api in my project, Without new added methods I proposed my project code will looks not very beautiful.

For example:
too many "isNull" and every getting call two apis: row.isNull and getX();
`return  new DO(
                        row.isNull("a") ? 0 : row.getInt("a"),
                        row.isNull("b") ? 0 : row.getLong("b"),
                        row.isNull("c") ? 0 : row.getShort("c"),
                        row.isNull("d") ? 0.0 : row.getDouble("d"),
                        row.isNull("e") ? false : row.getBoolean("e"),
                       );`
instead of it:
`return  new DO(
row.getInt("a",0),
 row.getLong("b",0),
row.getShort("c",0),
row.getDouble("d",0.0),
row.getBoolean("e",false),
 );`

With the new API, it looks more simpler. 

And with only one getInt("name") existed, It will may involved bug  easier due to not pay attention to distinguish the null an 0 when the coder doesn't read the API doc: when null return 0. But if the api added, it will make the coder pay more attention about it. He/She may know there must existed Null due to there is another method with default value.
So it may prevent mistake.It is one good point for API design.

What's more, most APIs/Open source consider the default value setting instead of call 2 apis together such as:

http://grepcode.com/file/repo1.maven.org/maven2/commons-configuration/commons-configuration/1.10/org/apache/commons/configuration/Configuration.java#Configuration

```
byte getByte(String key);
byte getByte(String key, byte defaultValue);
```

Whatever, as you talked, it is balance between thin and convenience. So it is up to your decision.
I just give my suggestions.

olim7t:
Closing based on our discussion, in this case I feel API surface and stability take precedence over convenience. Sorry @jiafu1115

jiafu1115:
@olim7t  NP.Thanks

PR #683, opened 2016-05-25
JAVA-1157: Allow asynchronous paging of Mapper Result

adutra:
This commit also extracts a common super-interface for both `ResultSet` and `Result`, as suggested by the ticket reporter. Clirr plugin was happy with that so I went ahead and implemented the idea.


driver-core/src/main/java/com/datastax/driver/core/Fetchable.java
```diff
@@ -0,0 +1,166 @@
...
+     * when needed (i.e. when {@code getAvailableWithoutFetching() == 0} and
+     * {@code isFullyFetched() == false}).
+     * <p/>
+     * You can however call this method manually to force the fetching of the
+     * next page of results. This can allow to prefetch results before they are
+     * strictly needed. For instance, if you want to prefetch the next page of
+     * results as soon as there is less than 100 rows readily available in this
+     * result set, you can do:
+     * <pre>
+     *   ResultSet rs = session.execute(...);
```
adutra: The javadocs still mention `ResultSet` in some places but I guess it's ok since `ResultSet` is the only child interface of this one in this module.

driver-core/src/main/java/com/datastax/driver/core/ResultSet.java
```diff
@@ -39,7 +34,11 @@
  * <p/>
  * Note that this class is not thread-safe.
  */
-public interface ResultSet extends Iterable<Row> {
+public interface ResultSet extends Fetchable<ResultSet, Row> {
+
+    // redeclared only to make clirr happy
+    @Override
+    Row one();
```
adutra, line 41: That was the only complaint by clirr plugin. All other methods could be pulled up without problems.

driver-core/src/main/java/com/datastax/driver/core/Fetchable.java
```diff
@@ -0,0 +1,166 @@
...
+     * result sets can be paged and thus be retrieved by multiple queries. In that
+     * case this method return the {@link ExecutionInfo} for the last query
+     * performed. To retrieve the information for all queries, use {@link #getAllExecutionInfo}.
+     * <p/>
+     * The returned object includes basic information such as the queried hosts,
+     * but also the Cassandra query trace if tracing was enabled for the query.
+     *
+     * @return the execution info for the last query made for this result set.
+     */
+    ExecutionInfo getExecutionInfo();
```
olim7t: These two methods feel a bit out-of-place here, they're more specific to CQL statements.

driver-core/src/main/java/com/datastax/driver/core/Fetchable.java
```diff
@@ -0,0 +1,166 @@
...
+import com.google.common.util.concurrent.ListenableFuture;
+
+import java.util.Iterator;
+import java.util.List;
+
+/**
+ * Defines a result set whose rows can be fetched and paged,
+ * possibly asynchronously.
+ */
+public interface Fetchable<S extends Fetchable<S, T>, T> extends Iterable<T> {
```
olim7t: How about `PagingIterable`?

driver-core/src/main/java/com/datastax/driver/core/Fetchable.java
adutra: I thought the same, but went for placing them here 1) for practical reasons (both `ResultSet` and `Result` expose them) and 2) thinking that if you are fetching results, you are implicitly sending requests, so it's not so awkward after all to expose these methods here. But I'm fine with removing if everyone agrees.

driver-core/src/main/java/com/datastax/driver/core/Fetchable.java
adutra: I like it, it stresses the fact that it extends `Iterable`. `Pageable` sounds nice too; the ticket reporter proposed initially `AsyncPageable`. The votes are open :)

adutra:
Squashed and rebased on top of current 3.0. Note that I also renamed `Fetchable`to `PagingIterable`.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperAsyncResultTest.java
```diff
@@ -0,0 +1,140 @@
...
+        public String getName() {
+            return name;
+        }
+
+        public void setName(String name) {
+            this.name = name;
+        }
+
+        @Override
+        public boolean equals(Object other) {
```
olim7t: Minor remark: strictly speaking you don't need to implement `equals` and `hashCode` here, since you're accumulating the ids below.

changelog/README.md
```diff
@@ -2,6 +2,9 @@
 
 ### 3.1.0 (in progress)
 
+- [new feature] JAVA-1157: Mapper should allow asynchronous Result paging.
```
olim7t: Minor: could you switch to an imperative style for this and the commit message? For example "Allow asynchronous Mapper Result paging".

adutra:
Rebased and addressed feedback. Does everyone agree that `getExecutionInfo()` should stay in `PagingIterable`?

driver-core/src/main/java/com/datastax/driver/core/ResultSet.java
```diff
@@ -39,7 +34,11 @@
  * <p/>
  * Note that this class is not thread-safe.
  */
-public interface ResultSet extends Iterable<Row> {
+public interface ResultSet extends PagingIterable<ResultSet, Row> {
+
+    // redeclared only to make clirr happy
+    @Override
+    Row one();
```
tolbertam, line 41: Since this is going into 3.1.0, does it matter that clirr isn't compliant for this particular case?

tolbertam:
> Rebased and addressed feedback. Does everyone agree that `getExecutionInfo()` should stay in `PagingIterable`?

Sounds good to me, I like that you don't really lose anything mapping from `ResultSet` -> `Result` now, everything is still accessible other than `getColumnDefinitions` (doesn't seem relevant when using mapped entities) and `wasApplied` (only applies to updates/deletes so doesn't seem relevant for mapping).

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperAsyncResultTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+        public void setName(String name) {
+            this.name = name;
+        }
+
+    }
+
+    @BeforeMethod(groups = "short")
+    public void cleanup() {
+        session().execute("TRUNCATE users");
```
tolbertam, line 77: I worry that truncate can be somewhat/slow unreliable, but it should not remotely be a problem with a smaller set of data, so not worried in this case.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperAsyncResultTest.java
```diff
@@ -0,0 +1,126 @@
...
+                results,
+                accumulator);
+        Futures.getUnchecked(future);
+        assertThat(accumulator.all.size()).isEqualTo(totalCount);
+    }
+
+    /**
+     * Dummy transformation that accumulates all traversed results
+     */
+    static class ResultsAccumulator implements AsyncFunction<Result<User>, Result<User>> {
```
tolbertam, line 107: Nice

driver-core/src/main/java/com/datastax/driver/core/ResultSet.java
adutra, line 41: You mean, you would prefer that I add the exception to `clirr-ignores.xml`? the exact error is 

```
7002: com.datastax.driver.core.ResultSet: Method 'public com.datastax.driver.core.Row one()' has been removed
```

PR #686, opened 2016-05-30
JAVA-541 + JAVA-636 + JAVA-984: Polymorphism support to object mapper

adutra:
This addresses both JAVA-541, JAVA-636, and JAVA-984.

The second commit and following introduce a refactoring of the mapper internals:
1. `ColumnMapper` renamed to `PropertyMapper`;
2. No more factory patterns to create mappers;
3. `ReflectionX` classes removed and merged into their parent classes;
4. `QueryType` class cleanup.
5. Replaced `IllegalArgumentException` with `IllegalStateException` where more appropriate.

If you want to read the docs first, jump to the new "[polymorphism support](https://github.com/datastax/java-driver/blob/java541/manual/object_mapper/creating/README.md#polymorphism-support)" section.


adutra:

manual/object_mapper/creating/README.md
```diff
@@ -222,3 +362,146 @@ private Map<String, List<Address>> frozenValueMap;
...
+    
+    @Override
+    public double getVolume() {
+        return 4d / 3d * Math.PI * Math.pow(getRadius(), 3);
+    }
+
+}
+```
+
+One powerful advantage of annotating getter methods is that
```
tolbertam, line 503: Awesome example, like how it shows polymorphism for both tables and UDTs.

adutra:
Guidelines for reviewers:
1. The `QueryType` class is now an enum. There were some other instances that were planned for slice queries, but these were never implemented, so it's actually dead code.
2. There were factories to obtain instances of `EntityMapper` and `ColumnMapper`. The factories though would always return the same instances. My guess, reading the commit history, is that the instances that we currently have use reflection to read and write values, so Sylvain or Michael probably left the door open for other solutions (bytecode generation, etc.). But as far as I know nobody ever asked for this. So the current state of code is very over-engineered. I took the opportunity to remove the factory classes.
3. The `ColumnMapper` class has been renamed to `PropertyMapper`. That wasn't required, but I guess the name `ColumnMapper` was chosen before UDTs arrived, and now this class applies to fields as well. So I thought that it would better stress the Java side of the mapping, because both columns and fields map to java properties, hence `PropertyMapper`.

If anyone disagrees with these decisions, please chime in.

olim7t:
Sorry but I feel that the JAVA-541 commit is missing the point: it merely adds the ability to inherit field definitions from a parent class. That's useful in itself, but what the ticket asks for is a way to encode inheritance relationships in the model. Borrowing from the Caffinitas terminology, and considering an inheritance hierarchy `Shape > Circle, Shape > Rectangle > Square`, that would mean:
- table-per-class inheritance: there are 4 tables `shapes`, `circles`, `rectangles` and `squares` sharing the same primary key. A `Square` instance is persisted as a row in each of `shapes`, `rectangles` and `squares`.
- single-table inheritance: there is a single table `shapes` with a discriminator column, and a column for any possible field in the hierarchy. A `Square` instance is persisted as a single row in this table, any non-applicable column (e.g. radius) is null.

adutra:
> Sorry but I feel that the JAVA-541 commit is missing the point.  it merely adds the ability to inherit field definitions from a parent class.

This is exactly what one of the commenters in the ticket asked for: fields in superclasses. And I think this guy has it right: keep it simple, but extend to class hierarchies.

So yes, I went for only that strategy, and as stated in the docs: "Each concrete class must correspond to a table or a UDT, and should be annotated accordingly with @Table or @UDT. This is analogous to the so-called "table per concrete class" mapping strategy."

Frankly, proposing other strategies is the red line I did not want to cross. Especially the single-table strategy; i wonder if this model is efficient at all for cassandra, given the high number of nulls. There is too much of Hibernate in it IMHO.

And while at it, why not propose strategies to map the same entity to many tables, each one "denormalized" differently (JAVA-549 and JAVA-524)?

I think we need to draw a clear line here and define the scope of "polymorphism support" and how far we want to go.

@al3xandru Could you please give us your insight here?

olim7t:
I agree with your arguments, and polymorphic mapping would certainly cross a line in term of implementation complexity. But at least we should not introduce property inheritance as a fix for JAVA-541, as I feel users watching that ticket are expecting much more.

olim7t:
I find 169e7e0 extremely hard to review. There are at least 3 different changes in there:
- remove factories
- rename MappedProperty to PropertyMapper
- allow annotations on getters and setters (JAVA-636)

What's heartbreaking is that the first two would be trivial to review separately, and they probably represent a large portion of the diff. With everything mixed together, you have to pay attention to each line to determine which change (or combination of changes) impacted it.

In the future it would really simply reviews if 1 commit = 1 change.

adutra:
> I find 169e7e0 extremely hard to review. 

Yes, my bad, please accept my apologies. I found myself in that situation and couldn't separate the changeset in two or three distinct commits, and so had to commit everything in one.

driver-mapping/src/main/java/com/datastax/driver/mapping/AnnotationParser.java
```diff
@@ -21,263 +21,170 @@
...
+ * - {@link #parseEntity(Class, MappingManager)}: entity classes into {@link EntityMapper} instances
+ * - {@link #parseUDT(Class, MappingManager)}: UDT classes into {@link MappedUDTCodec} instances.
+ * - {@link #parseAccessor(Class, MappingManager)}: Accessor interfaces into {@link AccessorMapper} instances.
  */
+@SuppressWarnings({"unchecked", "WeakerAccess"})
 class AnnotationParser {
 
-    private static final Comparator<Field> fieldComparator = new Comparator<Field>() {
+    @VisibleForTesting
+    static final Set<Class<? extends Annotation>> VALID_PROPERTY_ANNOTATIONS = ImmutableSet.of(
```
tolbertam: Wasn't clear to me initially that VALID_FIELD_ANNOTATIONS apply to UDTs while PROPERTY apply to tables, would be good to clarify that here (although it may be somewhere else but haven't gotten to it yet).

driver-mapping/src/main/java/com/datastax/driver/mapping/AnnotationParser.java
```diff
@@ -21,263 +21,170 @@
...
     private AnnotationParser() {
     }
 
-    public static <T> EntityMapper<T> parseEntity(Class<T> entityClass, EntityMapper.Factory factory, MappingManager mappingManager) {
+    static <T> EntityMapper<T> parseEntity(final Class<T> entityClass, MappingManager mappingManager) {
         Table table = AnnotationChecks.getTypeAnnotation(Table.class, entityClass);
 
-        String ksName = table.caseSensitiveKeyspace() ? table.keyspace() : table.keyspace().toLowerCase();
-        String tableName = table.caseSensitiveTable() ? table.name() : table.name().toLowerCase();
+        String ksName = table.caseSensitiveKeyspace() ? Metadata.quote(table.keyspace()) : table.keyspace().toLowerCase();
```
tolbertam, line 82: Was this an existing bug (That keyspaces / tables were not quoted)?   Had thought previously that you had to escape it in your annotation definition.

driver-mapping/src/main/java/com/datastax/driver/mapping/PropertyMapper.java
```diff
@@ -0,0 +1,196 @@
...
+    final TypeToken<Object> javaType;
+    final TypeCodec<Object> customCodec;
+    final int position;
+
+    private final Field field;
+    private final Method getter;
+    private final Method setter;
+    private final Map<Class<? extends Annotation>, Annotation> annotations;
+
+    PropertyMapper(Class<?> baseClass, String propertyName, String alias, Field field, PropertyDescriptor property) {
```
tolbertam, line 54: From what I've seen in the code it's never possible for both `Field` and `PropertyDescriptor` to each be defined or both be null, why not have two separate constructors, or even 2 separate subtypes?  I see in the getters/setters a lot of null checking, where if you had two different types you can have separate implementations.

driver-mapping/src/main/java/com/datastax/driver/mapping/AnnotationParser.java
tolbertam, line 82: Ah I see that `EntityMapper` used to do quoting on demand, where now it's done early, this change makes sense.

driver-mapping/src/main/java/com/datastax/driver/mapping/PropertyMapper.java
```diff
@@ -0,0 +1,196 @@
...
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * Maps a Java bean property to a table column or a UDT field.
+ * <p>
+ * Properties can be either accessed through getter and setter pairs,
+ * or by direct field access, depending on what is available in the
+ * entity/UDT class.
+ */
+class PropertyMapper {
```
tolbertam, line 40: Not sure this matters, but I noticed that `ColumnMapper` was generic, where its replacement, `PropertyMapper,` isn't.  Is there any particular reason why it was removed?  I see that it really wasn't needed in the first place, since the callers don't particularly deal with the type that is being mapped directly.

driver-mapping/src/main/java/com/datastax/driver/mapping/AnnotationChecks.java
```diff
@@ -40,7 +40,7 @@
     static <T extends Annotation> T getTypeAnnotation(Class<T> annotation, Class<?> annotatedClass) {
         T instance = annotatedClass.getAnnotation(annotation);
         if (instance == null)
-            throw new IllegalArgumentException(String.format("@%s annotation was not found on %s",
+            throw new IllegalStateException(String.format("@%s annotation was not found on %s",
```
olim7t: Admittedly `IllegalArgumentException` isn't the most appropriate type for mapper annotation misuse, but `IllegalStateException` isn't either: 
- `IllegalArgumentException`: _Thrown to indicate that a method has been passed an illegal or inappropriate argument_
- `IllegalStateException`: _Signals that a method has been invoked at an illegal or inappropriate time_

Between the two, I'd rather keep IAE for the sake of backward compatibility.

driver-mapping/src/main/java/com/datastax/driver/mapping/ReflectionUtils.java
```diff
@@ -15,26 +15,179 @@
...
+    // they cannot be both null at the same time
+    static <T> Map<String, Object[]> scanFieldsAndProperties(Class<T> baseClass) {
+        Map<String, Object[]> fieldsAndProperties = new HashMap<String, Object[]>();
+        Map<String, Field> fields = scanFields(baseClass);
+        for (Map.Entry<String, Field> entry : fields.entrySet()) {
+            fieldsAndProperties.put(entry.getKey(), new Object[]{entry.getValue(), null});
+        }
+        Map<String, PropertyDescriptor> properties = scanProperties(baseClass);
+        for (Map.Entry<String, PropertyDescriptor> entry : properties.entrySet()) {
+            Object[] value = fieldsAndProperties.get(entry.getKey());
```
tolbertam, line 66: What are the cases where you could have the multiple `PropertyDescriptor` with the same name?

driver-mapping/src/main/java/com/datastax/driver/mapping/ReflectionUtils.java
```diff
@@ -15,26 +15,179 @@
...
+    }
+
+    private static Map<Class<? extends Annotation>, Annotation> scanFieldAnnotations(Field field, Map<Class<? extends Annotation>, Annotation> annotations) {
+        for (Annotation annotation : field.getAnnotations()) {
+            annotations.put(annotation.annotationType(), annotation);
+        }
+        return annotations;
     }
+
+    private static Map<Class<? extends Annotation>, Annotation> scanMethodAnnotations(Method method, Map<Class<? extends Annotation>, Annotation> annotations) {
```
tolbertam, line 122: Ah nice, so this is where most of the magic happens to pull Annotations from parent classes/interfaces?

driver-mapping/src/main/java/com/datastax/driver/mapping/AnnotationParser.java
adutra, line 82: > Was this an existing bug (That keyspaces / tables were not quoted)?

I believe so. In the current code, only UDT names were quoted here, the rest was quoted on demand later on.

> Had thought previously that you had to escape it in your annotation definition.

Well, there wasn't any test so far testing case sensitiveness, and nothing in the docs. From the code I infer that we do not expect users to quote identifiers themselves, i.e., a column `"Address"` would be declared as:

``` java
@Column(name = "Address", caseSensitive = true)
```

And not:

``` java
@Column(name = "\"Address\"", caseSensitive = true)
```

> Ah I see that EntityMapper used to do quoting on demand, where now it's done early, this change makes sense.

Same thing for `ColumnMapper`, which is the purpose of JAVA-564.

driver-mapping/src/main/java/com/datastax/driver/mapping/PropertyMapper.java
adutra, line 54: Unfortunately it _is_ possible for both `Field` and `PropertyDescriptor` to be defined. It's even the most common case, i.e. you have a field (e.g. `foo`) whose name matches the property name (e.g. `getFoo()/setFoo()`).

driver-mapping/src/main/java/com/datastax/driver/mapping/ReflectionUtils.java
adutra, line 66: So the possible cases are:

<table>
<tr>
<th>Field</th>
<th>Property</th>
<th>Example</th>
</tr>
<tr>
<td>Present</td>
<td>Present</td>
<td>
<pre>
private String foo;
public String getFoo() { return foo; }
</pre>
</td>
</tr>

<tr>
<td>Present</td>
<td>Absent</td>
<td>
<pre>
private String foo;
// no getter nor setter
</pre>
</td>
</tr>
<tr>
<td>Absent</td>
<td>Present</td>
<td>
<pre>
private String _foo; // field name mismatch
public String getFoo() { return _foo; }
</pre>
</td>
</tr>
</table>

driver-mapping/src/main/java/com/datastax/driver/mapping/ReflectionUtils.java
adutra, line 122: Yes exactly!

driver-mapping/src/main/java/com/datastax/driver/mapping/PropertyMapper.java
adutra, line 40: The main reason I removed it is because no caller code actually cares about this type parameter. 

The second reason is that I find it misleading, because `T` was the _entity_ type, not the property type. E.g. the following column mapping:

``` java
class Foo {
@Column String bar;
}
```

...would generate a `ColumnMapper<Foo>`; I would naturally expect it to be `ColumnMapper<String>` instead.

driver-mapping/src/main/java/com/datastax/driver/mapping/AnnotationChecks.java
adutra: You have a fair point, I will revert exceptions back to IAE. 

Introducing a new `MappingException` would also be an option imo. Granted, it wouldn't be backwards-compatible, but I believe that in this special case, nobody would attempt to catch an exception raised by the mapper, simply because if this happens, it means that the mapping is wrong and needs to be fixed until no more exceptions are raised. So changing one unchecked exception for another shouldn't affect client code.

driver-mapping/src/main/java/com/datastax/driver/mapping/PropertyMapper.java
tolbertam, line 54: Ah, that makes sense to me now, thanks for clearing that up!

driver-mapping/src/main/java/com/datastax/driver/mapping/PropertyMapper.java
tolbertam, line 40: Good point, I didn't even make that connection initially, makes sense to remove it then.

driver-mapping/src/main/java/com/datastax/driver/mapping/AnnotationParser.java
```diff
@@ -43,8 +42,10 @@
 @SuppressWarnings({"unchecked", "WeakerAccess"})
 class AnnotationParser {
 
-    @VisibleForTesting
-    static final Set<Class<? extends Annotation>> VALID_PROPERTY_ANNOTATIONS = ImmutableSet.of(
+    /**
+     * Annotations allowed on a property that maps to a table column.
+     */
+    private static final Set<Class<? extends Annotation>> VALID_COLUMN_ANNOTATIONS = ImmutableSet.of(
```
tolbertam, line 49: Nice, especially like renaming property -> column

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperInvalidAnnotationsTest.java
```diff
@@ -0,0 +1,264 @@
...
+ */
+package com.datastax.driver.mapping;
+
+import com.datastax.driver.mapping.annotations.*;
+import org.testng.annotations.Test;
+
+import static org.mockito.Mockito.mock;
+
+@SuppressWarnings({"unused", "WeakerAccess"})
+public class MapperInvalidAnnotationsTest {
```
tolbertam, line 26: This is covers a lot of invalid cases! 👍

kford:
I'm using the 3.1.0 driver here (https://mvnrepository.com/artifact/com.datastax.cassandra/cassandra-driver-mapping/3.1.0) and not sure how to take advantage of the polymorphic support from this PR.  Is there an updated example that can demonstrate the capabilities?

Thanks

newkek:
Hi @kford, examples can be found in our docs: http://datastax.github.io/java-driver/manual/object_mapper/creating/#polymorphism-support

kford:
@newkek and @adutra, It says superclasses are scanned, but it seems not subclasses (by omission in the docs).  

*edit: should I move this elsewhere since this PR is already merged?

I have a use case like the following

class Properties{
  // no members
}

class UserTaskProperties extends Properties {
  // some members
}
/\* other types of Properties, like UserApprovalProperties etc., all of which extend Properties */

I created an example application that you can easily run with maven and attached the zip.. It just needs host/user/pass edited in pom.xml.. It demonstrates that saving these mapped subtypes is lossy.  Am I doing something wrong?  If I'm not, the only alternative I can think of is introducing custom codecs for the sub-types and manage the mapping manually there.. That would create significant maintenance overhead.

Thanks in advance

[cassandra-mapping.zip](https://github.com/datastax/java-driver/files/425866/cassandra-mapping.zip)

update: forgot to attach cql/ddl
[cassandra-poc-cql-ddl.txt](https://github.com/datastax/java-driver/files/425954/cassandra-poc-cql-ddl.txt)

tolbertam:
Hi @kford, 

This kind of polymorphism is not currently supported by the Mapper even with this change.   From the jira ticket ([JAVA-541](https://datastax-oss.atlassian.net/browse/JAVA-541)) we added a note that this doesn't cover all cases:

> We're solving this in 3.1.0 with the ability to inherit properties from parent classes. This allows you to use a "table-per-class" polymorphic mapping, where the table of each concrete class contains all the columns (its own + the ones inherited from the parents). Note that polymorphic queries are not supported (e.g. if Dog extends Animal and both are concrete, queries on the animals table will not return rows from the dogs table).
> 
> We're aware that polymorphic mapping can have many different meanings; table-per-class is a good compromise in terms of complexity and compatibility with the existing API. If this is not what you were waiting for, please open a separate ticket (and link back to this one) so that we can track the other alternatives.

Unfortunately this case is not covered, but Cassandra doesn't support this either as far as I know.  Using the provided schema:

```
CREATE TYPE cass_poc.vehicle (id uuid);
create type cass_poc.vehicle_type_truck (towing_capacity int);
create type cass_poc.vehicle_type_car (miles_per_gallon int, passenger_seats int);
CREATE TABLE cass_poc.employees (employee_id uuid PRIMARY KEY, employee_type text, name text, vehicles list<frozen<vehicle>>);
```

I see in your code that you want to be able to have `Truck` and `Car` types in a `List<Vehicle>` which maps to the `vehicles` column (`list<frozen<vehicle>>`) on the `employees` table.  While that is perfectly valid java, C\* won't allow it since the `vehicles` column requires `vehicle` UDTs, and there is no inheritance for UDTs in Cassandra currently.

As far as solving this particular issue, a couple ways it could be handled.
1. Make the vehicle UDT a union of car and truck, i.e.:  `create type cass_poc.vehicle (id uuid, vehicle_type text, towing_capacity int, miles_per_gallon int, passenger_seats int);`.  The `vehicle_type` column can hold what type of vehicle it is.  This doesn't really map well to the java side, but you could check the vehicle_type to know what fields are relevant...that's not very clean.
2. As I think you were alluding to, Store the vehicle in C\* as some other type (blob, map, etc.), and then register a Codec that maps Vehicle <-> `ByteBuffer` (or some other type).  This should allow you to represent your java model the way you like as long as the codec is registered to convert between `Vehicle` and the type cassandra needs.  I agree it is a bit of work, but it would work and I think is the best option.   Internally in C\* UDTs are effectively serialized blobs anyways (1 frozen cell that you cannot manipulate a field at a time), but C\* nicely represents them in a unified format that the drivers understand, where with your own custom blobs that would not be the case.

kford:
@tolbertam Thank you for the thorough response/explanation.  I guess this also means that if I wanted to query by one of the properties that is part of my logical Vehicle class, I'd have to extract that out so it's either not part of the blob, or redundantly stored as (text for example), outside the blob...

like...  type vehicle (vehicleType text, otherQueryableField text, vehicleBlob blob)  where the vehicleBlob may or may not redundantly contain vehicleType and otherQueryableField, depending on how I want to write my custom codec.

tolbertam:
> @tolbertam Thank you for the thorough response/explanation. I guess this also means that if I wanted to query by one of the properties that is part of my logical Vehicle class, I'd have to extract that out so it's either not part of the blob, or redundantly stored as (text for example), outside the blob...

That is correct, you can't query by individual UDT fields (yet: see [CASSANDRA-6832](https://issues.apache.org/jira/browse/CASSANDRA-6382)).  You would have to have an extra column that is part of the primary key or separately indexed to do that.

newkek:
> *edit: should I move this elsewhere since this PR is already merged?

Just a quick note since you asked @kford , our mailing is the most appropriate place to ask those kinds of questions: https://groups.google.com/a/lists.datastax.com/forum/#!forum/java-driver-user
Thanks.

PR #687, opened 2016-06-01
JAVA-1204: Update documentation to indicate tcnative version requirement

tolbertam:
For [JAVA-1204](https://datastax-oss.atlassian.net/browse/JAVA-1204)

Should also apply forward to 3.0 branch.

### DIFF
--- a/manual/ssl/README.md
+++ b/manual/ssl/README.md
@@ -152,8 +152,12 @@ Netty-tcnative provides the native integration with OpenSSL. Follow
 [these instructions](http://netty.io/wiki/forked-tomcat-native.html) to
 add it to your dependencies.
 
-Note that using netty-tcnative requires JDK 1.7 or above and requires the
-presence of OpenSSL on the system.  It will not fall back to the JDK implementation.
+There are known runtime incompatibilities between newer versions of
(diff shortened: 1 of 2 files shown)


manual/ssl/README.md
```diff
@@ -152,8 +152,12 @@ Netty-tcnative provides the native integration with OpenSSL. Follow
 [these instructions](http://netty.io/wiki/forked-tomcat-native.html) to
 add it to your dependencies.
 
-Note that using netty-tcnative requires JDK 1.7 or above and requires the
-presence of OpenSSL on the system.  It will not fall back to the JDK implementation.
+There are known runtime incompatibilities between newer versions of
+netty-tcnative and the version of netty that the driver uses.  For best
+results, use version 1.1.33.Fork10.
```
adutra: Is this the recommended version for Netty 4.0.36 as well (cf. JAVA-1147)?

tolbertam: I'd imagine so, i think forks 10-13 may work as the incompatibility was introduced in Fork14, which only works with 4.1+, but I will definitely test it!

tolbertam:
Updated to indicate Fork17 is now required, assuming #694 is merged.

PR #688, opened 2016-06-01
Fix typo in javadoc.

justinpitts:


olim7t:
Changed the base branch since we plan to rename 3.0 to 3.x.
@justinpitts you also need to sign the CLA in order for us to merge this.

justinpitts:
Do I need to resolve these build errors for JDK 6 and 7? It is not immediately obvious to me what the issue here is.

tolbertam:
It could be that the base branch you started from originally had a problem building on jdk 7 and 8 and has since been fixed.  Since you changed javadoc only it is unlikely this is introducing anything that breaks it and it looks like the 3.x branch is currently ok.

justinpitts:
I agree that it is unlikely my fault, but it's my PR so I assume it is
ultimately my responsibility to ensure it builds.

olim7t:
I'm going to cherry-pick your commit on the 3.0.x branch to include it in 3.0.4. Don't worry about the build.

olim7t:
Done. Unfortunately I can't mark this as merged since I changed the target branch, but your commit is in the repo: 06913ca

PR #689, opened 2016-06-06
JAVA-1206: log custom type columns in query logger.

olim7t:


driver-core/src/main/java/com/datastax/driver/core/DataType.java
```diff
@@ -900,16 +900,16 @@ public String getCustomTypeClassName() {
...
+        //        @Override
+//        public Object parse(String value) {
+//            throw new InvalidTypeException("Cannot parse values of custom types");
+//        }
+//
+//        @Override
+//        public String format(Object value) {
+//            throw new InvalidTypeException("Cannot format values of custom types");
+//        }
+//
```
adutra: What is the reason why you commented out these 2 methods?

adutra: Oh nevermind, it's because `QueryLogger` now calls `format()` on custom types.

olim7t: Yes, actually I meant to remove them.

PR #690, opened 2016-06-08
JAVA-1209: ProtocolOptions.getProtocolVersion() should return null instead of throwing NPE if Cluster has not been init'd

dmitrymurashenkov:


olim7t:
LGTM.
One minor thing: I'm not sure if having a unit test in a class extending CCMTestsSupport works well; @adutra?

tolbertam:
> One minor thing: I'm not sure if having a unit test in a class extending CCMTestsSupport works well; @adutra?

It looks like the test runs and is included as part of the unit tests, so it appears to work.  I think it is because all of the CCM logic is done as part of `@Before/@BeforeClass` annotations that are bound to short/long profiles.

adutra:
> One minor thing: I'm not sure if having a unit test in a class extending CCMTestsSupport works well; @adutra?

Yes, it works, we have some other occurrences of this in our test codebase.

olim7t:
Merged to 3.0.x.
Very minor remark: we tend to use AssertJ assertions (`assertThat` syntax) instead of TestNG's in new tests. I changed that in the merge commit.

PR #691, opened 2016-06-08
JAVA-1139: ConnectionException.getMessage() throws NPE if address is null

dmitrymurashenkov:


driver-core/src/test/java/com/datastax/driver/core/exceptions/ConnectionExceptionTest.java
```diff
@@ -0,0 +1,40 @@
...
+    public void getHost_should_return_null_if_address_is_null() {
+        assertNull(new ConnectionException(null, "Test message").getHost());
+    }
+
+    /**
+     * @jira_ticket JAVA-1139
+     */
+    @Test(groups = "unit")
+    public void getMessage_should_return_message_if_address_is_null() {
+        assertEquals("[null] Test message", new ConnectionException(null, "Test message").getMessage());
```
olim7t: We might as well remove `[null]` altogether in that case, that is have the message `"Test message"`.

PR #692, opened 2016-06-08
JAVA-1201: Document OSGi.

adutra:


faq/osgi/README.md
```diff
@@ -0,0 +1,179 @@
...
+This is correct, but could be a problem for some OSGi provisioning tools,
+and notably for [Tycho], because it does not consider optional
+dependencies when computing the target platform.
+If you are facing provisioning issues related to compression libraries,
+you might need to reference them in your project's manifest explicitly; 
+with the [Maven bundle plugin], this can be achieved with the following 
+[BND] `Import-Package` instruction (the example below is for
+LZ4, but the same applies to Snappy as well):
+
+    <Import-Package>net.jpountz.lz4,*</Import-Package>
```
tolbertam, line 69: It looks like the \* here throws off the markdown somehow, does it need to be escaped?

adutra, line 69: Hmm you're right, the syntax highlighting on Github seems confused by this asterisk, but I checked the grammar and inside a code block like this, there should be no parsing, so I guess the problem is with Github. Anyway the rendered HTML looks fine.

adutra:
Closing in favor of #695.

PR #693, opened 2016-06-09
JAVA-1101: Batch and BatchStatement should consider inner statements to determine query idempotence

dmitrymurashenkov:
BatchStatement and Batch now share common algorithm for isIdempotent():
1. If any statement in batch is non-idempotent - whole batch is non idempotent
2. If no non-idempotent statements, but some have isIdempotent=null then
   batch has null idempotency
3. Otherwise batch is idempotent

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -2,6 +2,7 @@
 
 ### 3.0.3 (in progress)
 
+- [bug] JAVA-1101: Batch and BatchStatement should consider inner statements to determine query idempotence
 
 ### 3.0.2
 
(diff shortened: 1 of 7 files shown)


tolbertam:
Looks great, thank you for your contribution @dmitrymurashenkov !

olim7t:
👍 merged to 3.0.x.
I had a few minor remarks but I addressed them directly in the merge commit:
- simplify `isBatchIdempotent` with an early return when we detect a non-idempotent statement
- use AssertJ assertions in the tests and rename the parent class

PR #694, opened 2016-06-09
JAVA-1147: Upgrade Netty to 4.0.37.

adutra:
This commit also solves JAVA-1213: Updates and inserts to BLOB column using ReadOnly ByteBuffer fail.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -2,6 +2,9 @@
 
 ### 3.0.3 (in progress)
 
+- [improvement] JAVA-1147: Upgrade Netty to 4.0.37.
+- [bug] JAVA-1213: Allow updates and inserts to BLOB column using read-only ByteBuffer.
+
 
 ### 3.0.2
 
--- a/driver-core/pom.xml
+++ b/driver-core/pom.xml
@@ -130,7 +130,7 @@
         <dependency>
             <groupId>io.netty</groupId>
             <artifactId>netty-tcnative</artifactId>
-            <version>1.1.33.Fork10</version>
+            <version>1.1.33.Fork17</version>
             <classifier>${os.detected.classifier}</classifier>
             <scope>test</scope>
         </dependency>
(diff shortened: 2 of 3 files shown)


tolbertam:
Looks like we'll need to upgrade netty-tcnative as well:

```
org.testng.TestException: 
Expected exception com.datastax.driver.core.exceptions.NoHostAvailableException but got java.lang.NoSuchMethodError: org.apache.tomcat.jni.SSLContext.getMode(J)I
    at io.netty.handler.ssl.OpenSslContext.<init>(OpenSslContext.java:207)
    at io.netty.handler.ssl.OpenSslContext.<init>(OpenSslContext.java:139)
    at io.netty.handler.ssl.OpenSslClientContext.<init>(OpenSslClientContext.java:187)
    at io.netty.handler.ssl.SslContext.newClientContextInternal(SslContext.java:746)
    at io.netty.handler.ssl.SslContextBuilder.build(SslContextBuilder.java:395)
    at com.datastax.driver.core.SSLTestBase.getSSLOptions(SSLTestBase.java:134)
    at com.datastax.driver.core.SSLEncryptionTest.should_not_connect_with_ssl_without_trusting_server_cert(SSLEncryptionTest.java:57)
```

I'll figure out what version is compatible.

tolbertam:
Just remembered, the driver doesn't have an explicit dependency on netty-tcnative, just our tests.  We'll need to update our test dependency and docs (since we added the version recommendation in #687)

tolbertam:
The good news is Fork17 (the latest) appears to work!  Updated the dependency, feel free to squash it in (should wait to see how CI goes first though).  There are a few more tests I'd like to add for tcnative (testing with cert files for trust in addition to the ones that already test with trust store) that I will do in a separate PR.

tolbertam:
Looks like CI ran clean.  👍  on merging that and will make sure during perf regression that this doesn't impact performance.

PR #695, opened 2016-06-20
JAVA-1201: Document OSGi.

adutra:


faq/osgi/README.md
```diff
@@ -0,0 +1,205 @@
+## Frequently Asked Questions - OSGi
```
olim7t, line 1: Awesome docs 👍

driver-tests/osgi/README.md
```diff
@@ -1,12 +1,21 @@
...
+[OSGi examples repository]:https://github.com/datastax/java-driver-examples-osgi
 
 ## Usage
 
+`MailboxService` is an OSGi service that uses Cassandra to
+store messages that can be retrieved by email address.
+
 To build the bundle and run tests, execute the following Maven goal:
 
     mvn verify -P short
```
olim7t, line 34: Note on line 26 (not part of this diff): it says to run `mvn package` on the 3 dependent modules, this actually leads to the test-jar issue if you start from a clean environment. I think it's simpler to recommend `mvn install` on the full project like it says in the POM.

faq/osgi/README.md
```diff
@@ -0,0 +1,205 @@
...
+_not_ its Maven artifact id.
+
+    
+### How to use the driver shaded jar?
+
+The driver [shaded jar](../../manual/shaded_jar/) can be used 
+in any OSGi application, although the same limitations explained in
+the manual apply.
+
+Be sure _not_ to include any non-shaded Netty bundle in your project.
```
tolbertam: Shouldn't this still work? The driver will just not use it, but other libraries that depend on netty should be able to use that version right?

faq/osgi/README.md
tolbertam, line 1: agreed, these are great!

driver-tests/osgi/README.md
adutra, line 34: Good point, thanks. Actually the pax runner plugin _always_ resolves everything from the local Maven repository, and never from the reactor build, that's why the POM says `mvn install`. For tests however we can access the reactor build artifacts (in a hacky fashion, but that works) – but obviously, we still need the test jar in the local repo. Thanks for thinking about this corner case.

faq/osgi/README.md
adutra: You are perfectly right. I tested a shaded configuration with Netty bundles provisioned as well, and it worked. I removed this section entirely.

adutra:
Rebased and addressed feedback.

faq/osgi/README.md
```diff
@@ -0,0 +1,201 @@
...
+dependencies when computing the target platform.
+
+If you are facing provisioning issues related to compression libraries,
+you might need to either add them explicitly to your OSGi runtime,
+or explicitly reference them in your project's manifest.
+With the [Maven bundle plugin], this second option can be achieved with the following 
+[BND] `Import-Package` instruction (the example below is for
+LZ4, but the same applies to Snappy as well):
+
+    <Import-Package>net.jpountz.lz4,*</Import-Package>
```
adutra, line 69: This asterisk is indeed annoying :)

PR #696, opened 2016-06-21
JAVA-1212: Don't retry non-idempotent statements by default.

olim7t:


driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -789,6 +793,15 @@ private void setFinalResult(Connection connection, Message.Response response) {
         }
     }
 
+    private void logIdempotenceWarning() {
+        if (WARNED_IDEMPOTENT.compareAndSet(false, true))
```
tolbertam, line 797: What do you think about establishing a common logger implementation for messages like this?  This way users can easily suppress warnings like this globally for all changes.   Some users pay very close attention to warnings and try their best to suppress all the ones that doesn't affect their application health.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -490,12 +523,17 @@ public void onSet(Connection connection, Message.Response response, long latency
                                 connection.release();
                                 assert err.infos instanceof WriteTimeoutException;
                                 WriteTimeoutException wte = (WriteTimeoutException) err.infos;
-                                retry = retryPolicy.onWriteTimeout(statement,
-                                        wte.getConsistencyLevel(),
-                                        wte.getWriteType(),
-                                        wte.getRequiredAcknowledgements(),
-                                        wte.getReceivedAcknowledgements(),
-                                        retriesByPolicy);
+                                if (statement.isIdempotentWithDefault(manager.cluster.getConfiguration().getQueryOptions()))
```
tolbertam, line 526: Shouldn't we get rid of `IdempotenceAwareRetryPolicy` since we now consider idempotence outside of Retry policy?

tolbertam:
A bunch of tests look to have been broken with this, but I think it should be pretty easy to fix.  Tests need to either be updated to mark certain statements as idempotent, or update their expected exceptions to account for the fact that they won't be retried, i'll take care of that.   EDIT: Fixed

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
tolbertam, line 526: Ah I see, it was marked deprecated.

driver-core/src/test/java/com/datastax/driver/core/policies/AbstractRetryPolicyIntegrationTest.java
```diff
@@ -44,10 +44,10 @@
 
 /**
  * Base class for retry policy integration tests.
- * <p/>
+ * <p>
```
tolbertam: ack, my bad, my environment is targeting JDK 8 which uses `<p>` instead of `<p/>` will fix.

PR #697, opened 2016-06-23
JAVA-1229: Allow specifying the keyspace for individual queries.

bannmann:
See https://datastax-oss.atlassian.net/browse/JAVA-1229.


olim7t:

bannmann:
@olim7t, I rebased my commit as well.

Now that 3.1.0 was released, I also moved my `changelog` entry to a new `3.2.0` section. Is that correct?

driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
```diff
@@ -929,6 +934,20 @@ public static Option saveNullFields(boolean enabled) {
...
+         * Creates a new Option object to specify the keyspace to use. This is valid for save, delete and get
+         * operations.
+         * <p/>
+         * If this option is not specified, it defaults to {@code null} (use the keyspace defined in the session or the
+         * {@link com.datastax.driver.mapping.annotations.Table} annotation).
+         *
+         * @param keyspace the name of the keyspace.
+         * @return the option.
+         */
+        public static Option keyspace(String keyspace) {
```
newkek, line 947: Seems conflicting that we would have an `Option` **and** an annotation for setting the keyspace. Sounds like here the `Option` takes over the annotation, right?

bannmann: If you specify the `keyspace` option, the keyspace annotation will be ignored, yes. Are you suggesting that the Javadoc doesn't make this clear? Please indicate what changes you want me to make.

Please make sure to read the use case in [JAVA-1229](https://datastax-oss.atlassian.net/browse/JAVA-1229).

bannmann:
I will resolve the conflicts if I get any indication that this PR has a chance to be considered.

adutra:
Thanks again for your work so far @bannmann. 

Before you rebase or refactor anything, please note that JAVA-1310 and JAVA-1316 have modified quite deeply the mapper internal classes. And they introduced the notion of strategies to configure certain aspects at MappingManager level.

So let's discuss the current design first under this new context.

First of all, I think that if we go for dynamic keyspace names, we should go for dynamic table names too, and maybe even UDT names. Who knows how people are dealing with multitenancy out there :)

Secondly, I wonder if this new feature wouldn't be a good fit for another strategy. I.e. instead of switching keyspaces per query, one would switch keyspaces per `MappingManager` instances. I understand that your primary concern here is multitenancy. My idea would mean that you would have as many `MappingManager` instances as you have tenants in your app, but they could share the same session under the hoods. Do you think this approach would work for you? If so, I would be in favor of making object names configurable through a new strategy instead of through query options.

Does that make sense?

bannmann:
Yeah, I noticed the refactoring and decided to manually re-apply my changes to current `3.x` instead of rebasing/merging. This would still work.

Anyway, you raise a good point. I have the vague impression that the keyspace name is not a very good fit for query options as the query text itself is not modified.

Having one mapper per tenant would require changes to our architecture that we have to think through first. I'll get back when we have an answer; this should only be a matter of a few days.

How would you envision the table name and UDT stuff to work?

adutra:
> How would you envision the table name and UDT stuff to work?

We would have another strategy, let's call it `ResolvingStrategy` for the moment. It's API would be stg like:

```
KeyspaceMetadata resolveKeyspace(Class<?> entityClass, Session session);
TableMetadata resolveTable(Class<?> entityClass, Session session);
UserType resolveUserType(Class<?> entityClass, Session session);
```

The default implementation would basically do everything that is currently done in `AnnotationParser`, i.e. inspect the entity annotation first, then fallback to the session's keyspace.

I would also take the opportunity of this refactoring to apply the `NamingStrategy` introduced in JAVA-1316 to keyspace, table and UDT names as well (i.e. if the keyspace is `myKeyspace` and the naming strategy is `LOWER_CAMEL_CASE`, then this strategy would look for a keyspace named `my_keyspace`).

bannmann:
> I understand that your primary concern here is multitenancy. My idea would mean that you would have as many MappingManager instances as you have tenants in your app, but they could share the same session under the hoods. Do you think this approach would work for you?

We have evaluated this approach and think it should work fine for our app.

> We would have another strategy, let's call it `ResolvingStrategy` for the moment.

I suppose the `ResolvingStrategy` would be set within the `MappingConfiguration`, right?

> I would also take the opportunity of this refactoring to apply the `NamingStrategy` introduced in JAVA-1316 to keyspace, table and UDT names as well

I can see how the `NamingStrategy` is used for inferring column names from field names in `DefaultPropertyMapper.inferMappedName()`, and understand that the new code would infer table/UDT names similarly based on the class name. However, I don't understand how it would apply to keyspace names, as those are not based on any Java identifiers, but strings given in the annotation. So there should be no need for conversion between different naming conventions. The multi tenancy requirement would already be fulfilled by the `ResolvingStrategy` itself.

These questions aside, I will start implementing your suggested approach. Is there a rough date or even deadline for 3.3.0 apart from "when it's done"?

adutra:
> I suppose the ResolvingStrategy would be set within the MappingConfiguration, right?

Correct.

> I can see how the NamingStrategy is used for inferring column names from field names in DefaultPropertyMapper.inferMappedName(), and understand that the new code would infer table/UDT names similarly based on the class name. However, I don't understand how it would apply to keyspace names, as those are not based on any Java identifiers, but strings given in the annotation. So there should be no need for conversion between different naming conventions. The multi tenancy requirement would already be fulfilled by the ResolvingStrategy itself.

Again you got it right, congrats!

As for the identifiers, I still think it could be valid to apply a `NamingStrategy` to keyspace, table and UDT names, even if for your specific case it isn't required.

Let's assume the following CQL model:

```sql
CREATE TABLE user_profile (user_id uuid PRIMARY KEY, user_name varchar, ...);
```

Currently it is possible to map this to the Java class below:

```java
@Table(name = "user_profile")
class UserProfile {
    @PartitionKey UUID userId;
    String userName;
}
```

Using the `LOWER_SNAKE_CASE` naming strategy, fields `userId` and `userName` are transparently mapped into `user_id` and `user_name`. However we still need to explicitly set the table name in the annotation, otherwise the mapper would infer `userprofile`. With my suggestion, the mapper would correctly infer `user_profile`, and the annotation could be simplified to just `@Table`. Same for UDTs.

But as I type this, I realize that this is probably not so useful for keyspace names because keyspaces do not correspond to any Java structure. I think it wouldn't hurt to apply the naming strategy for consistency sake, but I'l leave the decision to you.

> Is there a rough date or even deadline for 3.3.0 apart from "when it's done"?

I'm afraid no :( but we learned lessons from the last two minors (3.1 and 3.2) and we definitely would like to have shorter release cycles than before.

adutra:
Hmm we might need to change the signatures a bit. I think it would be unavoidable to pass a `NamingStrategy` instance to all methods in `ResolvingStrategy`, e.g.

```java
TableMetadata resolveTable(Class<?> entityClass, Session session, NamingStrategy namingStrategy);
```

This is unfortunate as I'd rather keep strategies orthogonal to each other, but these two are arguably a bit related.

adutra:
Hi @bannmann we're currently wrapping up 3.3.0 release. Do you think you could update your PR in the next days?

bannmann:
Thanks for the heads up! I think it's best if we just target the 3.4.0 release instead.

bannmann:
@adutra, do you have an idea where the call sites of `ResolvingStrategy` would get the naming strategy from? I could add a field in the `MappingConfiguration` for this, but for some reason this doesn't feel right.

My gut feeling is that the fact that the resolving strategy has a reference to a naming strategy is not relevant to these call sites or the mapping configuration. After all, the naming strategy used by the resolver is not different for each call, but very stable (set once when instantiating the mapping system). Thus, I would consider the reference to the naming strategy to be an implementation detail of `DefaultResolvingStrategy`, which should have a setter and/or constructor parameter for it.

Also note that we would need two fields because one might have different naming strategies for tables and UDTs.

adutra:
> After all, the naming strategy used by the resolver is not different for each call, but very stable (set once when instantiating the mapping system). Thus, I would consider the reference to the naming strategy to be an implementation detail of DefaultResolvingStrategy, which should have a setter and/or constructor parameter for it.

Sounds reasonable, 👍 

>  we would need two fields because one might have different naming strategies for tables and UDTs

I haven't considered that option. Do you think it would be a common thing to have different resolvers? If you think so, than I don't have any strong objections, but let's try to keep this new API as concise as possible.

bannmann:
> Do you think it would be a common thing to have different resolvers?

Actually I don't. That's the good thing about keeping the naming strategy as an implementation detail: we only have to decide what the `DefaultResolvingStrategy` does. Now if people want different naming strategies for tables and UDTs, we did not create any obstacles for them (unlike with the original idea of having naming strategy in interface method signatures).

Another topic I need input on is what exactly from `AnnotationParser` moves to ` ResolvingStrategy`:

1. easiest solution: have `ResolvingStrategy.resolveTable(Class<T> entityClass, MappingManager mappingManager)` return an `EntityMapper<T>`, so the method is effectively just a moved version of `AnnotationParser.parseEntity()`
    * advantage: annotation parsing is only triggered once, exactly in the place where the `EntityMapper` instance is created.
    * disadvantage: `ResolvingStrategy` does more than just table and keyspace names as due to the need to construct an `EntityMapper` it also knows about consistencies, keys and mapped properties.
2. alternative solution: limit the `ResolvingStrategy` to table name and keyspace; return an instance of a new class called `TableNameInfo` (or similar) that is used by the caller when constructing the `EntityMapper`
    * advantage: narrows the scope of `ResolvingStrategy` to a minimum
    * disadvantage: yet another layer of indirection (data flow: annotation -> `TableNameInfo` -> `EntityMapper`)
    * disadvantage: `AnnotationChecks.getTypeAnnotation()` would be called twice, once inside the `DefaultResolvingStrategy` and once in the class that constructs the rest of the `EntityMapper`

I tend towards 2. @adutra, do you agree?

bannmann:
Come to think of it, the new class mentioned in solution 2 can be used for tables and UDTs, making for a nice, uniform interface. Being able to reuse the same class for both makes it less controversial to add this layer of indirection, and one cannot really say this is "class creep".

olim7t:
I'm revisiting this PR in an effort to include it in 3.4.0. It looks like I don't have push access to your branch, so I've created #927.
I went with the initial proposal of using `Option`; with Java driver 4 on the horizon, we don't want to spend too much effort refactoring the 3.x branch, so this has the advantage of being simpler.

PR #698, opened 2016-06-23
JAVA-1200: Upgrade LZ4 to 1.3.0.

adutra:
This commit also upgrades Snappy version to 1.1.2.6 and HdrHistogram
version to 2.1.9.

Besides, it also introduces new OSGi tests to validate that the
following libraries can be used in an OSGi environment: LZ4, Snappy, and
HdrHistogram.

As these new tests rely on system properties that might be present or
not, this commit switches the Pax Exam container from "native" to
"forked", because native containers are not confined enough to isolate
system properties correctly.

### DIFF
--- a/driver-tests/osgi/pom.xml
+++ b/driver-tests/osgi/pom.xml
@@ -34,6 +34,7 @@
         <felix.version>4.6.0</felix.version>
         <!-- more recent version require JDK7+ -->
         <pax-exam.version>3.6.0</pax-exam.version>
+        <url.version>2.4.0</url.version>
         <logback.version>1.1.3</logback.version>
         <test.groups>none</test.groups>
         <!--
@@ -72,6 +73,12 @@
         </dependency>
 
         <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-api</artifactId>
+            <version>1.7.5</version>
+        </dependency>
+
+        <dependency>
             <groupId>com.datastax.cassandra</groupId>
             <artifactId>cassandra-driver-core</artifactId>
             <version>${project.parent.version}</version>
@@ -95,7 +102,7 @@
 
         <dependency>
             <groupId>org.ops4j.pax.exam</groupId>
-            <artifactId>pax-exam-container-native</artifactId>
+            <artifactId>pax-exam-container-forked</artifactId>
             <version>${pax-exam.version}</version>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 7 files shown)


driver-tests/osgi/pom.xml
```diff
@@ -95,7 +96,7 @@
 
         <dependency>
             <groupId>org.ops4j.pax.exam</groupId>
-            <artifactId>pax-exam-container-native</artifactId>
+            <artifactId>pax-exam-container-forked</artifactId>
```
adutra, line 105: Had to switch to forked to better isolate system properties.

driver-tests/osgi/pom.xml
```diff
@@ -108,6 +109,13 @@
         </dependency>
 
         <dependency>
+            <groupId>org.ops4j.pax.url</groupId>
+            <artifactId>pax-url-reference</artifactId>
```
adutra, line 119: Had to add this dependency to resolve `reference:` urls (it was probably a dependency of the native container before).

driver-tests/osgi/src/test/java/com/datastax/driver/osgi/BundleOptions.java
```diff
@@ -48,6 +49,19 @@ public static MavenArtifactProvisionOption guavaBundle() {
         return mavenBundle("com.google.guava", "guava", "16.0.1");
     }
 
+    public static CompositeOption lz4Bundle() {
+        return new CompositeOption() {
+
+            @Override
+            public Option[] getOptions() {
+                return options(
+                        systemProperty("cassandra.compression").value(ProtocolOptions.Compression.LZ4.name()),
```
adutra, line 59: Because this system property is not always present, subsequent tests would fail with the native container (runs all tests in the same JVM), while they succeed with the forked one.

driver-tests/osgi/src/test/java/com/datastax/driver/osgi/BundleOptions.java
```diff
@@ -84,6 +98,7 @@ public static CompositeOption defaultOptions() {
                         mavenBundle("ch.qos.logback", "logback-classic", "1.1.3"),
                         mavenBundle("ch.qos.logback", "logback-core", "1.1.3"),
                         mavenBundle("io.dropwizard.metrics", "metrics-core", "3.1.2"),
+                        mavenBundle("org.testng", "testng", "6.8.8"),
```
adutra, line 103: Had to add that too, probably another side effect of switching containers.

olim7t:
I moved the upgrades of Snappy and HdrHistogram out of this PR, will reintroduce them in JAVA-1233.

PR #699, opened 2016-06-25
JAVA-1232: Fix NPE in IdempotenceAwareRetryPolicy.isIdempotent.

AntonIvanov87:
https://datastax-oss.atlassian.net/browse/JAVA-1232


driver-core/src/test/java/com/datastax/driver/core/policies/LoggingRetryPolicyIntegrationTest.java
```diff
@@ -54,7 +55,7 @@ public void setUpRetryPolicy() {
         setRetryPolicy(new LoggingRetryPolicy(new CustomRetryPolicy()));
     }
 
-    @BeforeMethod(groups = {"short", "unit"})
```
AntonIvanov87, line 57: Cluster is not initialized in group "unit", that's why cluster.getConfiguration() below results in NPE.

olim7t:
👍 
I've checked other types of policies (load balancing...) for similar errors.

PR #702, opened 2016-06-30
JAVA-1160: Fix NPE in VersionNumber.getPreReleaseLabels()

tommystendahl:


olim7t:
👍 
However I'd rather target this to the 3.0 branch since we don't actively maintain 2.1 anymore. (don't re-create the PR, I'll handle this when merging)

olim7t:
Merged to 3.0.x.
Your commit is at f840fca, I also used the opportunity to refactor the unit test (082c047).

PR #703, opened 2016-06-30
JAVA-1227: Document "SELECT *" issue with prepared statement

tolbertam:
For [JAVA-1227](https://datastax-oss.atlassian.net/browse/JAVA-1227).

Added section to 'Prepared Statements' doc and FAQ.  The first paragraph is the same in each section, with the FAQ entry linking to the doc.

Assigned to 3.0.3 milestone as the jira ticket targets this.

### DIFF
--- a/faq/README.md
+++ b/faq/README.md
@@ -88,6 +88,17 @@ session.execute(batch);
 ```
 
 
+### Why do my 'SELECT *' `PreparedStatement`-based queries stop working after a schema change?
+
+Both the driver and Cassandra maintain a mapping of `PreparedStatement` queries to their
+metadata.  When a change is made to a table, such as a column being added or dropped, there
+is currently no mechanism for Cassandra to invalidate the existing metadata.  Because of this,
+the driver is not able to properly react to these changes and will improperly read rows after
+a schema change is made.
+
+See [Prepared statements](../manual/statements/prepared) for more information.
+
+
 ### Can I get the raw bytes of a text column?
 
 If you need to access the raw bytes of a text column, call the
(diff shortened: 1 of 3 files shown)


manual/statements/prepared/README.md
```diff
@@ -241,9 +241,27 @@ Changing the driver's defaults should be done with care and only in
...
+
+Both the driver and Cassandra maintain a mapping of `PreparedStatement` queries to their
+metadata.  When a change is made to a table, such as a column being added or dropped, there
+is currently no mechanism for Cassandra to invalidate the existing metadata.  Because of this,
+the driver is not able to properly react to these changes and will improperly read rows after
+a schema change is made.
+
+Therefore it is currently recommended to not create prepared statements
+for 'select *' queries unless you don't plan on making schema changes involving
+adding or dropping columns. Alternatively you should list all columns of interest
```
olim7t, line 254: I would replace "unless you don't plan" with "if you plan" (the double negative is harder to read).

olim7t:
👍

manual/statements/prepared/README.md
tolbertam, line 254: I don't think I should not fix this :P    (fixing shortly)

tolbertam, line 254: Fixed in 609f3a2, can be squashed before merge.

faq/README.md
```diff
@@ -88,6 +88,17 @@ session.execute(batch);
 ```
 
 
+### Why do my 'select *' `PreparedStatement`-based queries stop working after a schema change?
```
adutra: I see that we tend to use upper case letters for CQL keywords throughout the manual, so for consistency I suggest changing 'select *' to `SELECT *`.

tolbertam: Always good to be consistent,  will fix.

PR #704, opened 2016-07-06
JAVA-1235: Move the query to the end of "re-preparing .." log message…

vkasar:
… as a key value.


olim7t:
Merged to 3.0.x, thanks.
One minor detail: we generally add the latest tickets at the end of the changelog. I fixed it in the merge commit.

PR #707, opened 2016-07-13
Ensure decode takes readerIndex() into account.

normanmaurer:
Motivation:

Frame.Decoder and Frame.Decoder.DecoderForStreamIdSize assume that the passed in ByteBuf always should be read starting at index 0. This is not a safe assumption.

Modifications:

Not start explicit on index 0 but use readerIndex().

Result:

Correctly handle the case where the readerIndex() is not 0.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Frame.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Frame.java
@@ -197,7 +197,7 @@ class Frame {
             if (buffer.readableBytes() < 1)
                 return;
 
-            int version = buffer.getByte(0);
+            int version = buffer.getByte(buffer.readerIndex());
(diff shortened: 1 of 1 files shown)


olim7t:
Changing the milestone back to 3.0.4 since this is a bugfix. Also changing the base branch since we plan to rename 3.0 (no action needed, I'll handle the details when I merge).

tolbertam (approved):
lgtm, 👍 .  There is another issue when we hit `TooLargeFrameException`.  If the entire frame hasn't been received yet, the `LengthFieldBasedFrameDecoder` skips the bytes we have thus far, so next time we go to decode we are in a bad position.  I hope to address that in #748.

olim7t:
Cherry-picked on 3.0.x as c658e73.

PR #709, opened 2016-07-14
JAVA-1246: Driver swallows the real exception in a few cases

vkasar:


vkasar:
This PR was created against datastax:3.0 instead of datastax:3.0.x. Closing.

PR #667, opened 2016-05-11
Adding root cause to IllegalArgumentException on Cluster.addContactPoint

brezende:

PR #710, opened 2016-07-14
JAVA-1246: Driver swallows the real exception in a few cases

vkasar:


driver-core/src/main/java/com/datastax/driver/core/ReplicationStategy.java
```diff
@@ -55,6 +56,7 @@ static ReplicationStrategy create(Map<String, String> replicationOptions) {
             }
         } catch (NumberFormatException e) {
             // Cassandra wouldn't let that pass in the first place so this really should never happen
+            logger.error("Failed to parse replication options: " + replicationOptions, e);
```
adutra, line 59: Did you manage to actually reproduce a `NumberFormatException` here? It is obviously possible to reproduce this error artificially, i.e. by invoking this method directly with invalid parameters, but is it ever triggered when dealing with real keyspaces? I tried with different C\* versions and replication strategies, but I invariably get back an `InvalidConfigurationInQueryException` if the replication factor is not numeric, so the exception is never raised here. 
If you _did_ manage to trigger this error somehow, I would suggest rephrasing the comment  above as it seems inaccurate (and if possible, describe a situation where such an error can happen).
Otherwise, given that `NumberFormatException` is an unchecked exception and does not require to be caught, I would suggest that we simply get rid of this catch block altogether.

adutra:
lgtm 👍 , thanks for contributing!

driver-core/src/main/java/com/datastax/driver/core/ReplicationStategy.java
vkasar, line 59: We did not reproduce a NumberFormatException. All that we did was add a log where an exception was silently being swallowed. Sylvain added that catch block as part of rev 1497aedfb4d29277489bc285b3642a4387b28f3b. I am not sure about the implications of removing that block. Let me know if you are OK leaving it there as part of this PR. Feel free to remove the catch block in another commit.

adutra, line 59: The advantage of removing this catch block is that it removes some dead code and increases our test coverage. But I don't mind leaving it neither, so I'd suggest that we wait for some more input from others in the team: @olim7t, @newkek, @tolbertam, what are your thoughts?

adutra:
FYI other cases where the driver swallows the original exception were reported a while ago in #545 (probably using a static analysis tool). We chose to handle them case by case and declined the PR. Here, on the other hand, I agree that the reported cases should be fixed.

driver-core/src/main/java/com/datastax/driver/core/ReplicationStategy.java
tolbertam, line 59: I think it makes sense to leave the catch block.  While in practice it shouldn't happen as the values should be numeric, it doesn't hurt to catch the exception, since its otherwise an unchecked exception and in the event it does happen (for whatever reason) it'd be nice to handle and log it.

olim7t, line 59: 👍 for leaving the catch block. Maybe that's a bit far-fetched, but if replication factors can ever be something other than integers, we'd want older driver versions to be resilient. If we let the exception propagate it would go all the way to `SchemaParser#refresh`, skipping everything after the keyspace, vs. just having one `KeyspaceMetadata` with missing replication settings.

olim7t:
The `addContactPoint` case was also suggested in #667.

olim7t:
Looks like there was an accidental merge on your branch, I cherry-picked your commit directly on 3.0.x (091fb2a13daa7e606f5b9963136c4002582e7156) to have a cleaner history. In the future you can rebase your branches and do a `push --force` (it's fine for pull requests).
I have to close this, but your commit is on 3.0.x with proper attribution.

PR #713, opened 2016-07-15
Various test fixes after 3.0 merge

tolbertam:
Work in progress.  Various fixes to tests after 3.0 merge to fix some intermittently failing tests.  Will close and reopen against appropriate source branches (2.1/3.0/3.0.x,etc.) once all issues are hammered out.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -10,7 +10,7 @@ cassandra:
   - 2.1
   - 2.2
   - 3.0
-  - 3.4
+  - 3.7
 build:
   - type: maven
     version: 3.2.5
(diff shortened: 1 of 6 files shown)


tolbertam:
Removing the wip tag.  Still have a few occasional (i.e. `com.datastax.driver.core.QueryLoggerErrorsTest.should_log_queries_beyond_dynamic_threshold`) timing issues to knock out (will open up a separate PR), but I think it would be good to merge this now primarily for 4649ce6 since it fails pretty often and 4a09cf4 since it always fails.  Merging this should cause builds to pass more often than not (where now they always fail).

I will also go back and cherry-pick 4649ce6 onto 2.1.

PR #714, opened 2016-07-19
Update tests to account for retry behavior change

tolbertam:
As of JAVA-1212 (#696), non-idempotent queries are no longer retried for writes
and request errors.  Tests that expected this have been updated to
account for this.


tolbertam:
Looks like this has fixed all the issues and should be ready to merge.

PR #716, opened 2016-07-27
JAVA-1237: Add JSON examples

olim7t:
This is a low-hanging fruit and it would be nice to have in 3.1.0 since JAVA-743 was also introduced in that version.
I've initialized the example with the raw string variants, I think it would also be nice to include similar examples with custom types and the Jackson codec (see #715). driver-examples will have to depend on driver-extras, but I don't see any issue with that.
@adutra could you take a look at it tomorrow if you have time?

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
@@ -21,7 +21,6 @@ import com.datastax.driver.core.policies.RetryPolicy;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
@@ -297,12 +296,16 @@ public abstract class BuiltStatement extends RegularStatement {
     @Override
     public String toString() {
         try {
-            if (forceNoValues) {
-                return getQueryString();
+            try {
+                if (forceNoValues)
+                    return getQueryString();
+                // 1) try first with all values inlined (will not work if some values require custom codecs)
+                return maybeAddSemicolon(buildQueryString(null, CodecRegistry.DEFAULT_INSTANCE)).toString();
+            } catch (CodecNotFoundException e) {
+                // 2) try next with bind markers for all values to avoid usage of custom codecs
+                return maybeAddSemicolon(buildQueryString(new ArrayList<Object>(), CodecRegistry.DEFAULT_INSTANCE)).toString();
             }
-            StringBuilder queryString = buildQueryString(null, CodecRegistry.DEFAULT_INSTANCE);
-            return maybeAddSemicolon(queryString).toString();
-        } catch (CodecNotFoundException e) {
+        } catch (RuntimeException e) {
             // Ugly but we have absolutely no context to get the registry from
             return String.format("built query (could not generate with default codec registry: %s)", e.getMessage());
         }
--- a/driver-examples/src/main/java/com/datastax/driver/examples/json/JacksonJsonColumn.java
+++ b/driver-examples/src/main/java/com/datastax/driver/examples/json/JacksonJsonColumn.java
@@ -1,0 +1,170 @@
+/*
+ *      Copyright (C) 2012-2015 DataStax Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+package com.datastax.driver.examples.json;
+
+import com.datastax.driver.core.*;
+import com.datastax.driver.extras.codecs.json.JacksonJsonCodec;
+import com.fasterxml.jackson.annotation.JsonCreator;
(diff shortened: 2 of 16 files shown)


driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
```diff
@@ -297,14 +296,16 @@ public Boolean isIdempotent() {
     @Override
     public String toString() {
         try {
-            if (forceNoValues) {
-                return getQueryString();
-            }
-            StringBuilder queryString = buildQueryString(null, CodecRegistry.DEFAULT_INSTANCE);
-            return maybeAddSemicolon(queryString).toString();
+            // try first with all values inlined (will not work if some values require custom codecs)
```
tolbertam: What are the implications of no longer check forceNoValues?  Isn't that here for protocol version 1?  Although toString (maybe) isn't used by the driver itself, so it is not considered relevant here?

driver-examples/src/main/java/com/datastax/driver/examples/json/JacksonJsonColumn.java
```diff
@@ -0,0 +1,170 @@
...
+import static com.datastax.driver.core.querybuilder.QueryBuilder.*;
+
+/**
+ * Illustrates how to map a single table column of type {@code VARCHAR},
+ * containing JSON payloads, into a Java object using
+ * the <a href="http://wiki.fasterxml.com/JacksonHome">Jackson</a> library.
+ * <p/>
+ * This example makes usage of a custom {@link TypeCodec codec},
+ * {@link JacksonJsonCodec}, which is declared in the driver-extras module.
+ * If you plan to reuse this example, make sure to include the following
```
tolbertam: I think it would be better to rephrase as 'If you plan to follow this example in your own project' instead of 'If you plan to reuse this example' or something to that extent.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
adutra: I might be wrong, but since if `forceNoValues = true` then `getQueryString()` calls `buildQueryString(null, codecRegistry)`, it boils down to the same, except that we do not take advantage of the cache anymore (but is it important here?).

driver-examples/src/main/java/com/datastax/driver/examples/json/JacksonJsonFunction.java
```diff
@@ -0,0 +1,214 @@
...
+/**
+ * Illustrates how to map a single table column of an arbitrary type
+ * to a Java object using
+ * the <a href="http://wiki.fasterxml.com/JacksonHome">Jackson</a> library,
+ * and leveraging the {@code toJson()} and {@code fromJson()} functions
+ * introduced in Cassandra 2.2.
+ * <p/>
+ * This example makes usage of a custom {@link TypeCodec codec},
+ * {@link JacksonJsonCodec}, which is declared in the driver-extras module.
+ * If you plan to reuse this example, make sure to include the following
```
tolbertam: same comment as above

driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
tolbertam: I see so it should behave the same but without using the cache.  I'm not completely sure i understand the motivation behind the change though.  What is the difference of handling CodecNotFoundException by trying with the empty list, that the values are included in the toString?

driver-examples/src/main/java/com/datastax/driver/examples/json/JacksonJsonFunction.java
```diff
@@ -0,0 +1,214 @@
...
+    // Mapping JSON payloads to table columns of arbitrary types,
+    // using fromJson() function
+    private static void insertFromJson(Session session) {
+
+        User alice = new User("alice", 30);
+        User bob = new User("bob", 35);
+
+        ObjectNode aliceScores = JsonNodeFactory.instance.objectNode()
+                .put("call_of_duty", 4.8)
+                .put("pockemon_go", 9.7);
```
tolbertam: pockemon -> pokemon.

tolbertam: Also I think you are giving pokemon go way too much credit 😄

driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
adutra: Sorry, I should have explained better my rationale:
1. First, try to generate a query string avoiding bind markers as much as possible (because `UPDATE t SET x = 'foo' WHERE y = 42` is more precise than `UPDATE t SET x = ? WHERE y = ?`);
2. If some variable causes a `CodecNotFoundException` to be raised, try again without the values (because `UPDATE t SET x = ? WHERE y = ?` is still better than an error message);
3. If that raises an error again, then return the dummy `built query ...` error message.

But you are right that if `forceNoValues = true`, there is no reason why we should not use the cache. I will amend that.

driver-examples/src/main/java/com/datastax/driver/examples/json/JacksonJsonFunction.java
tolbertam: oh I think I misread it as a review score, in that case it is fine :)

driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
```diff
@@ -297,12 +296,16 @@ public Boolean isIdempotent() {
     @Override
     public String toString() {
         try {
-            if (forceNoValues) {
-                return getQueryString();
+            try {
+                if (forceNoValues)
+                    return getQueryString();
```
tolbertam, line 301: Nice, i'm more comfortable now that it doesn't replace the forceNoValues case 👍

PR #717, opened 2016-07-29
Add web UI paging examples.

olim7t:


adutra:
The build is failing because `animal-sniffer` plugin complains about `com.sun.net.httpserver.HttpServer` and alike not being part of the API. IMHO it would be better to replace it with a more popular HTTP server (Jetty, Tomcat, Grizzly...).

Also I think pagination would be better demonstrated if the table had a clustering column, and the query was actually returning ordered CQL rows for a given partition key.

And finally, maybe a REST example would be more appropriate as we wouldn't care so much about the HTML formatting boilerplate code.

adutra:
Went ahead and implemented my own suggestions by adding a new `ForwardPagingRestUi` example.

Also fixed minor issues with other examples.

olim7t:
> pagination would be better demonstrated if the table had a clustering column, and the query was actually returning ordered CQL rows for a given partition key.

Good point.

> maybe a REST example would be more appropriate as we wouldn't care so much about the HTML formatting boilerplate code

For a quick test in the browser HTML is much more user-friendly, you get a formatted output and clickable links.

> The build is failing because animal-sniffer plugin complains about com.sun.net.httpserver.HttpServer and alike not being part of the API. IMHO it would be better to replace it with a more popular HTTP server (Jetty, Tomcat, Grizzly...).

I'm 👎 on that. IMO the goal of the examples is to be as standalone as possible, ideally you should be able to browse the examples on github, pick the one you're interested in, copy-paste it in your project and it works directly. Therefore: avoid extra dependencies, don't factor common code (no parent or utility classes -- yes there will be some code duplication). Your REST example introduces 9 extra dependencies.
Granted, the `com.sun.*` classes are not part of the official API, but they're in Oracle JDK and OpenJDK. Let's be honest, what percentage of users are using something else? I'd rather have something that works out of the box in 99% of cases.

Since we don't deploy the examples module in central, is it that important to run animal-sniffer on it?

adutra:
> For a quick test in the browser HTML is much more user-friendly, you get a formatted output and clickable links.

That's debatable, some browser extensions provide very easy navigation and display of REST services, but fair enough, I would be in favor of providing both.

>  Your REST example introduces 9 extra dependencies. [...]  I'd rather have something that works out of the box in 99% of cases.

Well the REST example works in probably 100% of cases, precisely because it declares the dependencies it requires. Some of them (Jersey) can hardly be avoided I'm afraid. But I admit that Grizzly could be easily replaced with the `com.sun` HTTP server, I recall that there is a Jersey container for that server. So I will amend the REST server to at least get rid of the Grizzly dependencies.
But let's be realistic: the more complex the examples will get the more dependencies we will have. I don't think we can provide high-level examples with no extra dependencies at all

> Since we don't deploy the examples module in central, is it that important to run animal-sniffer on it?

That's a very good point, we should disable it altogether.

olim7t:
> the more complex the examples will get the more dependencies we will have

So maybe the examples should not get complex. We can cover quite a lot of features with the "single Java file with no external dependencies" model.

olim7t:
Sorry to insist, but IMO the added dependencies (even reduced) outweigh the benefits. What does the REST example demonstrate (with regard to Java driver paging) that the others don't already cover?

adutra:
> Sorry to insist, but IMO the added dependencies (even reduced) outweigh the benefits. What does the REST example demonstrate (with regard to Java driver paging) that the others don't already cover?

I think we have different visions of what an example is meant to be. You seem to have a restrictive acceptation: an example should simply demonstrate a driver feature. I tend to consider that an example should also demonstrate how to _integrate_ a driver feature with existing software.

A very basic example such as `ReadCassandraVersion` indeed does nothing more than show how to make reads with the driver. However, if you look into the JSON examples, you will notice that they actually showcase the driver integrated with third-party libraries such as Jackson (and btw, they too require additional dependencies, and they too cover exactly the same features, only with different libraries).

In the paging examples, I simply see this same principle going one step further: the driver is here presented in the context of web applications. So to answer your question: "What does the REST example cover?" Well, it covers the usage of the driver in a REST application :) In particular, it shows how to manage and inject the session, how to use JSON codecs to leverage the conversion from database rows -> Java objects -> JSON, etc. I might be wrong but I think users developing applications with this kind of software stack would be happy to see how to handle these common needs. Granted, that comes with a cost: 6 new dependencies (note that these could very well be useful to other examples).

And sorry to say, but applying your restrictive meaning, I don't see what an HTML example buys us. You could as well print the pages on the console along with the hashed paging state.  No need for [this boilerplate code](https://github.com/datastax/java-driver/blob/ba3bbcddc82fe8142a7cccf6f34569f8f8316d38/driver-examples/src/main/java/com/datastax/driver/examples/paging/ForwardPagingWebUi.java#L168-L223) and we could get rid of the `com.sun` dependency.

I do agree with you though that "examples should not get complex". But where to draw the line? if the amount of required dependencies is really big, we should create separate modules, that seems obvious. That's what happened for OSGi, and that's what would happen if, say, we decide to write a full-blown Spring example.

Anyway, I would suggest to ping others in the team and have their thoughts about it.

We could consider that it is useless and delete it. We could also consider that it "crossed the red line" and is too complex to be shown here, but would deserve to be presented in a separate project, just like the OSGi ones. I'm fine with either outcome.

^^ cc @newkek @tolbertam @al3xandru

olim7t:
> I don't see what an HTML example buys us. You could as well print the pages on the console along with the hashed paging state

The only reason  chose HTML is interactivity. Paging is a bit complex so it's nice that you can set up breakpoints, then click the next/previous links and debug what happens (you can also do it with the rest example but there are extra steps, like curl and copy-pasting, or installing a custom browser extension).

Now that you mention it, an interactive console app would also achieve this goal. I'll look into it.

To restate my position on where to draw the line for examples: it would be nice to keep the "copy-paste in my project and it just works" aspect. Otherwise you force users who want to execute and debug examples to:
- either checkout and compile the whole driver project
- or fish out which dependencies are required for that particular example. My worry is that the module grows to a gazillion dependencies which would make this a pain in the ass.

olim7t:
> Now that you mention it, an interactive console app would also achieve this goal. I'll look into it.

After more thought on this, a webapp is a better example because of the stateless aspect: the paging state is stored in the URL, you could bookmark it, restart the server and it would still work. A console app wouldn't illustrate that as well because it would store the paging state in memory.

Also, 99.9% of practical use cases will revolve around HTTP.

newkek:
I've been pinged here so I'll give my opinion... 

The main goal for these examples is to show how paging works, we expect users to go check the code. That said, I'm looking at the WebUI and the RestUI paging code and I must admit that the boiler plate code of HTML formatting makes things much harder to understand. 
In the RestUI example since it's all made with Object mappings, the core code can focus on how the paging works and I think that this is most important. We also know that having an example that uses Rest services and Object mapping would be much more popular Less people would be interested in having to understand all that paging code on the HTML example, and re-write everything because most users will write services like the Rest service.
I do understand the "having the least dependencies possible" but we need to define "Do we want something that people can actually reuse in their use case, or just an abstract example?" if it is only an abstract example (what I consider to be the WebUI) then it is as valuable as Docs, and we already have docs for paging.
Also since driver-examples is not a dependency on any other module and is not deployed in mvn central, it has no impact to add those dependencies for people that do not want to use it, so it seems acceptable to me.
I lean towards keeping the RestUI examples.

olim7t:
OK, let's go with REST then.

olim7t:
Changed base branch to 3.x, make sure you use that for future rebases.

olim7t:
Migrated all examples to REST and rebased on latest 3.x. This should be ready now so we can include it in 3.1.1.

tolbertam (approved):
This looks great!  👍

PR #718, opened 2016-08-01
JAVA-1248: Implement "beta" flag for native protocol v5.

ifesdjeen:
@adutra hey I've addressed the comments you've had [here](https://github.com/datastax/java-driver/commit/ee3e855a989575f33191040e0fe991e3339b7977) (did a force-push, so there's no direct link anymore). 

Unfortunately, since Cassandra version is not yet released, I had to hardcode the `install-dir` for CCM.

Thank you for the help.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -686,6 +686,7 @@ public class Cluster implements Closeable {
         private SSLOptions sslOptions = null;
         private boolean metricsEnabled = true;
         private boolean jmxEnabled = true;
+        private boolean allowBetaProtocolVersion = false;
 
         private Collection<Host.StateListener> listeners;
 
@@ -740,6 +741,27 @@ public class Cluster implements Closeable {
         }
 
         /**
+         * Create cluster connection using latest development protocol version,
+         * which is currently in beta. Calling this method will result into setting
+         * USE_BETA flag in all outgoing messages, which allows server to negotiate
+         * the supported protocol version even if it is currently in beta.
+         * <p/>
+         * This feature is only available starting with version {@link ProtocolVersion#V5 V5}.
+         * <p/>
+         * Use with caution, refer to the server and protocol documentation for the details
+         * on latest protocol version.
+         * @return this Builder.
+         */
+        public Builder allowBetaProtocolVersion() {
+            if (protocolVersion != null)
+                throw new IllegalArgumentException("Can't use beta flag with initial protocol version of " + protocolVersion);
+
+            this.allowBetaProtocolVersion = true;
+            this.protocolVersion = ProtocolVersion.NEWEST_BETA;
+            return this;
+        }
+
+        /**
          * Sets the maximum time to wait for schema agreement before returning from a DDL query.
          * <p/>
          * If not set through this method, the default value (10 seconds) will be used.
@@ -759,7 +781,7 @@ public class Cluster implements Closeable {
         /**
          * The native protocol version to use.
          * <p/>
-         * The driver supports versions 1 to 3 of the native protocol. Higher versions
+         * The driver supports versions 1 to 5 of the native protocol. Higher versions
          * of the protocol have more features and should be preferred, but this also depends
          * on the Cassandra version:
          * <p/>
@@ -769,6 +791,8 @@ public class Cluster implements Closeable {
          * <tr><td>1</td><td>1.2</td></tr>
          * <tr><td>2</td><td>2.0</td></tr>
          * <tr><td>3</td><td>2.1</td></tr>
+         * <tr><td>4</td><td>2.2</td></tr>
+         * <tr><td>5</td><td>3.10</td></tr>
          * </table>
          * <p/>
          * By default, the driver will "auto-detect" which protocol version it can use
@@ -779,6 +803,11 @@ public class Cluster implements Closeable {
          * the driver connects to is a Cassandra 1.2 node and auto-detection is used
          * (the default), then the native protocol version 1 will be use for the lifetime
          * of the Cluster instance.
+         * <p/>
+         * By using {@link Builder#allowBetaProtocolVersion()}, it is
+         * possible to force driver to connect to Cassandra node that supports the latest
+         * protocol beta version. Leaving this flag out will let client to connect with
+         * latest released version.
          * <p/>
          * This method allows to force the use of a particular protocol version. Forcing
          * version 1 is always fine since all Cassandra version (at least all those
@@ -804,6 +833,12 @@ public class Cluster implements Closeable {
          * @return this Builder.
          */
         public Builder withProtocolVersion(ProtocolVersion version) {
+            if (allowBetaProtocolVersion)
+                throw new IllegalStateException("Can not set the version explicitly if `allowBetaProtocolVersion` was used.");
+            if (version.compareTo(ProtocolVersion.NEWEST_SUPPORTED) > 0)
+                throw new IllegalArgumentException("Can not use " + version + " protocol version. " +
+                        "Newest supported protocol version is: " + ProtocolVersion.NEWEST_SUPPORTED + ". " +
+                        "For beta versions, use `allowBetaProtocolVersion` instead");
             this.protocolVersion = version;
             return this;
         }
--- a/driver-core/src/main/java/com/datastax/driver/core/Connection.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Connection.java
@@ -273,6 +273,7 @@ class Connection {
                             case V2:
                             case V3:
                             case V4:
+                            case V5:
                                 return authenticateV2(authenticator, protocolVersion, initExecutor);
                             default:
                                 throw defunct(protocolVersion.unsupported());
@@ -1389,6 +1390,7 @@ class Connection {
         private static final Message.ProtocolEncoder messageEncoderV2 = new Message.ProtocolEncoder(ProtocolVersion.V2);
         private static final Message.ProtocolEncoder messageEncoderV3 = new Message.ProtocolEncoder(ProtocolVersion.V3);
         private static final Message.ProtocolEncoder messageEncoderV4 = new Message.ProtocolEncoder(ProtocolVersion.V4);
+        private static final Message.ProtocolEncoder messageEncoderV5 = new Message.ProtocolEncoder(ProtocolVersion.V5);
         private static final Frame.Encoder frameEncoder = new Frame.Encoder();
 
         private final ProtocolVersion protocolVersion;
@@ -1451,6 +1453,8 @@ class Connection {
                     return messageEncoderV3;
                 case V4:
                     return messageEncoderV4;
+                case V5:
+                    return messageEncoderV5;
                 default:
                     throw new DriverInternalError("Unsupported protocol version " + protocolVersion);
             }
--- a/driver-core/src/main/java/com/datastax/driver/core/Frame.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Frame.java
@@ -111,6 +111,7 @@ class Frame {
                 return fullFrame.readByte();
             case V3:
             case V4:
+            case V5:
                 return fullFrame.readShort();
             default:
                 throw version.unsupported();
@@ -153,6 +154,7 @@ class Frame {
                     return 8;
                 case V3:
                 case V4:
+                case V5:
                     return 9;
                 default:
                     throw version.unsupported();
@@ -164,7 +166,8 @@ class Frame {
             COMPRESSED,
             TRACING,
             CUSTOM_PAYLOAD,
-            WARNING;
+            WARNING,
+            USE_BETA;
 
             static EnumSet<Flag> deserialize(int flags) {
                 EnumSet<Flag> set = EnumSet.noneOf(Flag.class);
@@ -277,6 +280,7 @@ class Frame {
                     break;
                 case V3:
                 case V4:
+                case V5:
                     header.writeShort(streamId);
                     break;
                 default:
--- a/driver-core/src/test/java/com/datastax/driver/core/ProtocolBetaVersionTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/ProtocolBetaVersionTest.java
@@ -1,0 +1,137 @@
+/*
+ *      Copyright (C) 2012-2015 DataStax Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+package com.datastax.driver.core;
+
+import org.testng.annotations.Test;
+
+import static com.datastax.driver.core.ProtocolVersion.V4;
+import static com.datastax.driver.core.ProtocolVersion.V5;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+/**
+ * Tests for the new USE_BETA flag introduced in protocol v5
+ * and Cassandra 3.10.
+ */
+@CCMConfig(version = "3.10")
+public class ProtocolBetaVersionTest extends CCMTestsSupport {
+
+    /**
+     * Verifies that the cluster builder fails when version is explicitly set and user attempts to set beta flag.
+     *
+     * @jira_ticket JAVA-1248
+     */
+    @Test(groups = "short")
+    public void should_not_initialize_when_version_explicitly_required_and_beta_flag_is_set() throws Exception {
+        try {
+            Cluster.builder()
+                    .addContactPoints(getContactPoints())
+                    .withPort(ccm().getBinaryPort())
+                    .withProtocolVersion(V4)
+                    .allowBetaProtocolVersion()
+                    .build();
+            fail("Expected IllegalArgumentException");
+        } catch (IllegalArgumentException e) {
+            assertThat(e.getMessage()).isEqualTo("Can't use beta flag with initial protocol version of V4");
+        }
+    }
+
+    /**
+     * Verifies that the cluster builder fails when beta flag is set and user attempts to pass a version explicitly.
+     *
+     * @jira_ticket JAVA-1248
+     */
+    @Test(groups = "short")
+    public void should_not_initialize_when_beta_flag_is_set_and_version_explicitly_required() throws Exception {
+        try {
+            Cluster.builder()
+                    .addContactPoints(getContactPoints())
+                    .withPort(ccm().getBinaryPort())
+                    .allowBetaProtocolVersion()
+                    .withProtocolVersion(V4)
+                    .build();
+            fail("Expected IllegalStateException");
+        } catch (IllegalStateException e) {
+            assertThat(e.getMessage()).isEqualTo("Can not set the version explicitly if `allowBetaProtocolVersion` was used.");
+        }
+    }
+
+    /**
+     * Verifies that the driver CANNOT connect to 3.10 with the following combination of options:
+     * Version V5
+     * Flag UNSET
+     *
+     * @jira_ticket JAVA-1248
+     */
+    @Test(groups = "short")
+    public void should_not_connect_when_beta_version_explicitly_required_and_flag_not_set() throws Exception {
+        try {
+            Cluster.builder()
+                    .addContactPoints(getContactPoints())
+                    .withPort(ccm().getBinaryPort())
+                    .withProtocolVersion(V5)
+                    .build();
+            fail("Expected IllegalArgumentException");
+        } catch (IllegalArgumentException e) {
+            assertThat(e.getMessage()).startsWith("Can not use V5 protocol version. Newest supported protocol version is: V4");
+        }
+    }
+
+    /**
+     * Verifies that the driver can connect to 3.10 with the following combination of options:
+     * Version UNSET
+     * Flag SET
+     * Expected version: V5
+     *
+     * @jira_ticket JAVA-1248
+     */
+    @Test(groups = "short")
+    public void should_connect_with_beta_when_no_version_explicitly_required_and_flag_set() throws Exception {
+        // Note: when the driver's ProtocolVersion.NEWEST_SUPPORTED will be incremented to V6 or higher
+        // a renegotiation will start taking place here and will downgrade the version from V6 to V5,
+        // but the test should remain valid
+        Cluster cluster = Cluster.builder()
+                .addContactPoints(getContactPoints())
+                .withPort(ccm().getBinaryPort())
+                .allowBetaProtocolVersion()
+                // no version explicitly required -> renegotiation allowed
+                .build();
+        cluster.connect();
+        assertThat(cluster.getConfiguration().getProtocolOptions().getProtocolVersion()).isEqualTo(V5);
+    }
+
+    /**
+     * Verifies that the driver can connect to 3.10 with the following combination of options:
+     * Version UNSET
+     * Flag UNSET
+     * Expected version: V4
+     *
+     * @jira_ticket JAVA-1248
+     */
+    @Test(groups = "short")
+    public void should_connect_after_renegotiation_when_no_version_explicitly_required_and_flag_not_set() throws Exception {
+        // Note: when the driver's ProtocolVersion.NEWEST_SUPPORTED will be incremented to V6 or higher
+        // the renegotiation will start downgrading the version from V6 to V4 instead of V5 to V4,
+        // but the test should remain valid
+        Cluster cluster = Cluster.builder()
+                .addContactPoints(getContactPoints())
+                .withPort(ccm().getBinaryPort())
+                // no version explicitly required -> renegotiation allowed
+                .build();
+        cluster.connect();
+        assertThat(cluster.getConfiguration().getProtocolOptions().getProtocolVersion()).isEqualTo(V4);
+    }
+}
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -3,6 +3,7 @@
 ### 3.2.0 (in progress)
 
 - [new feature] JAVA-1347: Add support for duration type.
+- [new feature] JAVA-1248: Implement "beta" flag for native protocol v5.
 
 
 ### 3.1.3
--- a/driver-core/src/main/java/com/datastax/driver/core/CodecUtils.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/CodecUtils.java
@@ -67,6 +67,7 @@ public final class CodecUtils {
                 return getUnsignedShort(input);
             case V3:
             case V4:
+            case V5:
                 return input.getInt();
             default:
                 throw version.unsupported();
@@ -91,6 +92,7 @@ public final class CodecUtils {
                 break;
             case V3:
             case V4:
+            case V5:
                 output.putInt(size);
                 break;
             default:
@@ -129,6 +131,7 @@ public final class CodecUtils {
                 break;
             case V3:
             case V4:
+            case V5:
                 if (value == null) {
                     output.putInt(-1);
                 } else {
@@ -214,6 +217,7 @@ public final class CodecUtils {
                 return 2;
             case V3:
             case V4:
+            case V5:
                 return 4;
             default:
                 throw version.unsupported();
@@ -230,6 +234,7 @@ public final class CodecUtils {
                 return 2 + elemSize;
             case V3:
             case V4:
+            case V5:
                 return value == null ? 4 : 4 + value.remaining();
             default:
                 throw version.unsupported();
--- a/driver-core/src/main/java/com/datastax/driver/core/Message.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Message.java
@@ -314,6 +314,8 @@ abstract class Message {
             EnumSet<Frame.Header.Flag> flags = EnumSet.noneOf(Frame.Header.Flag.class);
             if (request.isTracingRequested())
                 flags.add(Frame.Header.Flag.TRACING);
+            if (protocolVersion == ProtocolVersion.NEWEST_BETA)
+                flags.add(Frame.Header.Flag.USE_BETA);
             Map<String, ByteBuffer> customPayload = request.getCustomPayload();
             if (customPayload != null) {
                 if (protocolVersion.compareTo(ProtocolVersion.V4) < 0)
--- a/driver-core/src/main/java/com/datastax/driver/core/ProtocolEvent.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ProtocolEvent.java
@@ -132,6 +132,7 @@ class ProtocolEvent {
                     return new SchemaChange(change, targetType, targetKeyspace, targetName, targetSignature);
                 case V3:
                 case V4:
+                case V5:
                     change = CBUtil.readEnumValue(Change.class, bb);
                     targetType = CBUtil.readEnumValue(SchemaElement.class, bb);
                     targetKeyspace = CBUtil.readString(bb);
--- a/driver-core/src/main/java/com/datastax/driver/core/ProtocolVersion.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ProtocolVersion.java
@@ -29,12 +29,15 @@ public enum ProtocolVersion {
     V1("1.2.0", 1),
     V2("2.0.0", 2),
     V3("2.1.0", 3),
-    V4("2.2.0", 4);
+    V4("2.2.0", 4),
+    V5("3.10.0", 5);
 
     /**
      * The most recent protocol version supported by the driver.
      */
     public static final ProtocolVersion NEWEST_SUPPORTED = V4;
+
+    public static final ProtocolVersion NEWEST_BETA = V5;
 
     private final VersionNumber minCassandraVersion;
     private final int asInt;
--- a/driver-core/src/main/java/com/datastax/driver/core/Requests.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Requests.java
@@ -326,6 +326,7 @@ class Requests {
                 case V2:
                 case V3:
                 case V4:
+                case V5:
                     CBUtil.writeConsistencyLevel(consistency, dest);
                     dest.writeByte((byte) QueryFlag.serialize(flags));
                     if (flags.contains(QueryFlag.VALUES)) {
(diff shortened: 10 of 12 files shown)


driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -243,7 +246,8 @@ private static String extractMessage(Throwable t) {
                         // Testing for a specific string is a tad fragile but well, we don't have much choice
                         // C* 2.1 reports a server error instead of protocol error, see CASSANDRA-9451
                         if ((error.code == ExceptionCode.PROTOCOL_ERROR || error.code == ExceptionCode.SERVER_ERROR) &&
-                                error.message.contains("Invalid or unsupported protocol version"))
+                                error.message.contains("Invalid or unsupported protocol version") ||
```
adutra: I think the precedence of boolean operators would require parentheses here:

```
(
    error.code == ExceptionCode.PROTOCOL_ERROR || 
    error.code == ExceptionCode.SERVER_ERROR
) && (
    error.message.contains("Invalid or unsupported protocol version") || 
    error.message.contains("Beta version of the protocol used")
)
```

Or is it enough to detect the `"Beta version of the protocol used"` message, regardless of the error code?

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -1319,6 +1327,10 @@ protected void initChannel(SocketChannel channel) throws Exception {
             pipeline.addLast("frameDecoder", new Frame.Decoder());
             pipeline.addLast("frameEncoder", frameEncoder);
 
+            if (useBeta) {
+                pipeline.addLast(new Frame.BetaVersionFlagAdder());
```
adutra: Very minor suggestions:
1. I don't know the practical implications of _not_ giving a name to a handler, but since all the other handlers are given names in this method, I'd suggest we name this new one too.
2. Since this handler is stateless, couldn't we make it a singleton, just like `Frame.Encoder`? Also it would be good to annotate it with `@ChannelHandler.Sharable`.

driver-core/src/main/java/com/datastax/driver/core/Frame.java
```diff
@@ -175,7 +178,10 @@ static int lengthFor(ProtocolVersion version) {
                 return set;
             }
 
-            static int serialize(EnumSet<Flag> flags) {
+            static int serialize(EnumSet<Flag> flags, ProtocolVersion protocolVersion) {
+                if (protocolVersion.toInt() < ProtocolVersion.V5.toInt() && flags.contains(USE_BETA))
```
adutra: I think we have a problem here. `Frame.BetaVersionFlagAdder` is executed _after_  `Frame.Encoder` so the flag will never be set at this point. Maybe a better approach is to move this check to `Frame.BetaVersionFlagAdder`.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
ifesdjeen: You're right, it's much better to check error code.

driver-core/src/main/java/com/datastax/driver/core/ProtocolOptions.java
```diff
@@ -231,4 +232,23 @@ public AuthProvider getAuthProvider() {
...
+    public ProtocolOptions setAllowBetaProtocolVersions(boolean useBeta) {
+        this.useBeta = useBeta;
+        return this;
+    }
+
+    /**
+     * Whether or not connecting to server with protocol version which is currently in beta is allowed.
+     * @return {@code true} if connecting with beta protocol version is allowed, {@code false} otherwise.
+     */
+    public boolean getAllowBetaProtocolVersions() {
```
adutra: The correct name would be `isAllowBetaProtocolVersions()`.

driver-core/src/main/java/com/datastax/driver/core/ProtocolOptions.java
```diff
@@ -231,4 +232,23 @@ public AuthProvider getAuthProvider() {
...
     }
 
+    /**
+     * Allows client to connect to server using latest development protocol version,
+     * which is currently in beta.
+     *
+     * @param useBeta
+     * @return this {@code ProtocolOptions} object.
+     */
+    public ProtocolOptions setAllowBetaProtocolVersions(boolean useBeta) {
```
adutra: I don't think we can change this value once the `Cluster` is created, because we would have to remove the `Frame.BetaVersionFlagAdder` handler from all channels. So I'd suggest we make this field `final` and remove this setter.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -744,6 +745,22 @@ public Builder withPort(int port) {
...
+         * which is currently in beta.
+         * <p/>
+         * Use with caution, refer to the server and protocol documentation for the details
+         * on latest protocol version.
+         *
+         * @param allowBetaProtocolVersions {@code true} if connecting with beta protocol version should be allowed,
+         *                                  {@code false} otherwise
+         * @return this Builder.
+         */
+        public Builder setAllowBetaProtocolVersions(boolean allowBetaProtocolVersions) {
```
adutra: Following the example of other boolean properties in this builder, I would suggest turning this method into a more simple `allowBetaProtocolVersions()` that would internally set the corresponding flag to `true` (with no possibility of turning back to `false`).

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -744,6 +745,22 @@ public Builder withPort(int port) {
         }
 
         /**
+         * Allows client to connect to server using latest development protocol version,
+         * which is currently in beta.
```
adutra: Maybe add here some explanation about what are the implications of calling this method (i.e. the flag `USE_BETA` would be added to all outgoing messages) and also warn that it only makes sense for protocol versions >= 5.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -784,6 +803,11 @@ public Builder withMaxSchemaAgreementWaitSeconds(int maxSchemaAgreementWaitSecon
          * (the default), then the native protocol version 1 will be use for the lifetime
          * of the Cluster instance.
          * <p/>
+         * With {@link ProtocolOptions#setAllowBetaProtocolVersions(boolean)}, it is
```
adutra: If the referenced method is deleted as I suggested, please delete this paragraph as well (or rephrase accordingly).

driver-core/src/main/java/com/datastax/driver/core/Frame.java
adutra: Actually it is not a good idea to throw exceptions from channel handlers as these would defunct the connection and throw a rather cryptic error message ("Error writing").
So instead I would suggest moving the check to `Frame.BetaVersionFlagAdder` but in a more lenient version:

``` java
static class BetaVersionFlagAdder extends MessageToMessageEncoder<Frame> {
    @Override
    protected void encode(ChannelHandlerContext ctx, Frame frame, List<Object> out) throws Exception {
        ProtocolVersion protocolVersion = frame.header.version;
        if (protocolVersion.toInt() >= ProtocolVersion.V5.toInt())
            frame.header.flags.add(USE_BETA);
        out.add(frame);
    }
}
```

One might wonder why the check is required at all, since this handler is only added to the pipeline if the user requests the `USE_BETA` flag explicitly. It is still required in the case of renegotiation: e.g. the driver tries `V5` + `USE_BETA`, gets a `PROTOCOL_ERROR` because the server only handles `V4`, and then retries with `V4`: in this case, because `USE_BETA` is not supported in `V4`, the handler should check the version and refrain itself from adding the flag.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
adutra: Also, as a convenience for our users, we might want to cross-check invalid situations such as:

``` java
Cluster cluster = Cluster.builder()
        .addContactPoint(...)
        .withProtocolVersion(V4)
        .setAllowBetaProtocolVersions(true)
        .build();
```

...and throw an `IllegalArgumentException` right away instead of letting the user proceed.

driver-core/src/test/java/com/datastax/driver/core/ProtocolBetaVersionTest.java
```diff
@@ -0,0 +1,123 @@
...
+import static com.datastax.driver.core.ProtocolVersion.V4;
+import static com.datastax.driver.core.ProtocolVersion.V5;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+/**
+ * Tests for the new USE_BETA flag introduced in protocol v5
+ * and Cassandra 3.10.
+ */
+@CCMConfig(createCluster = false, version = "3.10")
```
adutra: This test requires  version 3.10 _precisely_, just as `ProtocolV1Test` requires 1.2.19.

Until 3.10 is released, here is what you should do to make the tests pass:

```
cd /path/to/cassandra/local/git/repo
git fetch
git checkout 12142-trunk
ant build
cd ~/.ccm/repository
ln -s /path/to/cassandra/local/git/repo 3.10
```

Note that you will have to delete the symlink after 3.10 is released.

driver-core/src/test/java/com/datastax/driver/core/ProtocolBetaVersionTest.java
```diff
@@ -15,48 +15,109 @@
...
+    /**
+     * Verifies that the driver can connect to 3.10 with the following combination of options:
+     * Version V5
+     * Flag SET
+     * Expected version: V5
+     *
+     * @jira_ticket JAVA-1248
+     */
+    @Test(groups = "short")
+    public void should_connect_with_beta_when_beta_version_explicitly_required_and_flag_set() throws Exception {
```
ifesdjeen: very good test, along with unset flag one.

driver-core/src/main/java/com/datastax/driver/core/ProtocolOptions.java
ifesdjeen: true, I've only added it here because I wanted to reduce the amount of constructors. But you're right, it's better to do it here.

driver-core/src/main/java/com/datastax/driver/core/ProtocolOptions.java
ifesdjeen: Sorry, didn't notice / know about this convention.

driver-core/src/test/java/com/datastax/driver/core/ProtocolBetaVersionTest.java
ifesdjeen: Since the branch is merged, one can already use the snapshot version or trunk, so you can do something like:

```
cd /tmp
wget http://cassci.datastax.com/job/trunk/lastSuccessfulBuild/artifact/build/apache-cassandra-3.10-SNAPSHOT-src.tar.gz
tar -xvf apache-cassandra-3.10-SNAPSHOT-bin.tar.gz
cd ~/.ccm/repository
ln -s /tmp/apache-cassandra-3.10-SNAPSHOT-bin/ 3.10
```

(with auto-delete™)

ifesdjeen:
I've addressed the comments you mentioned. Hope it's better now.

driver-core/src/main/java/com/datastax/driver/core/ProtocolOptions.java
```diff
@@ -135,11 +136,30 @@ public ProtocolOptions(int port) {
...
+     * @param protocolVersion the protocol version to use. This can be {@code null}, in which case the
+     *                        version used will be the biggest version supported by the <em>first</em> node the driver connects to.
+     *                        See {@link Cluster.Builder#withProtocolVersion} for more details.
+     * @param sslOptions      the SSL options to use. Use {@code null} if SSL is not
+     *                        to be used.
+     * @param authProvider    the {@code AuthProvider} to use for authentication against
+     *                        the Cassandra nodes.
+     * @param useBeta         whether or not this client is allowed to connect to servers with protocol versions that are in beta.
+     */
+    public ProtocolOptions(int port, ProtocolVersion protocolVersion, int maxSchemaAgreementWaitSeconds, SSLOptions sslOptions, AuthProvider authProvider, boolean useBeta) {
```
adutra: Those constructors are really annoying because each new setting requires a new constructor. I wonder if we shouldn't introduce a builder pattern, turn this constructor private and mark the other as `@Deprecated`. This way the private constructor is not part of the API and can be modified any time without risking to break backwards compatibility.

olim7t: I'm not sure, it sounds a bit overkill. By hiding the constructors we also prevent extension of the class (that's an edge case, but we've had users extend these classes to add their own options).
On the other hand I get your point about extensibility. But we rarely add new protocol options.

driver-core/src/main/java/com/datastax/driver/core/ProtocolOptions.java
```diff
@@ -231,4 +251,11 @@ public AuthProvider getAuthProvider() {
         return authProvider;
     }
 
+    /**
+     * Whether or not connecting to server with protocol version which is currently in beta is allowed.
+     * @return {@code true} if connecting with beta protocol version is allowed, {@code false} otherwise.
+     */
+    public boolean isAllowBetaProtocolVersions() {
```
olim7t: I'm not a big proponent of `isXxx` in all cases, here I think `allowsBetaProtocolVersions` would be easier to read.

driver-core/src/main/java/com/datastax/driver/core/Frame.java
```diff
@@ -328,4 +332,16 @@ protected void encode(ChannelHandlerContext ctx, Frame frame, List<Object> out)
...
     }
+
+    @ChannelHandler.Sharable
+    static class BetaVersionFlagAdder extends MessageToMessageEncoder<Frame> {
+        @Override
+        protected void encode(ChannelHandlerContext ctx, Frame frame, List<Object> out) throws Exception {
+            ProtocolVersion protocolVersion = frame.header.version;
+            // Check for the protocol version, if the downgrade has occurred during renegotiation
+            if (protocolVersion.toInt() >= ProtocolVersion.V5.toInt())
+                frame.header.flags.add(Header.Flag.USE_BETA);
```
olim7t: I don't understand what's happening here: it seems like there's a case where outgoing frames did not contain the beta flag and we add it on the fly. How can that be? Either the user set the beta flag in their config and it should already be set on the frame, or they didn't and forcing it transparently sounds wrong (user wouldn't realize they're in beta).

olim7t: Possibly related to this: how does auto-negotiation behave if
- driver is an older version that supports a beta v5
- Cassandra is a newer version that has the complete v5
- you set allowBetas = false

Could negotiation somehow succeed and result in v5/no beta? That would be bad because the driver could fail on some of the not-yet-implemented-on-the-client v5 features.

I feel like the driver should have the notion of whether its latest supported version is fully implemented or not (and not even allow non-beta attempts if it is not).

ifesdjeen: > Possibly related to this: how does auto-negotiation behave if

This is what we currently have. The initial patch was adding `beta` to the protocol version, although we discussed and decided to go the protocol-negotiation path. 

> outgoing frames did not contain the beta flag and we add it on the fly

That was a bit easier than having ProtocolOptions (or that flag) passed through to `ProtocolEncoder` (since it's currently static). 

What we can do is adding a "soft" coding to fix both issues. If we know that driver supports v5 only in beta, we'd throw a configuration exception in cluster builder. Same simplification can be done in `ProtocolEncoder`, where we can just hardwire beta flag for all v5 messages on creation instead of having it added dynamically.

cc @adutra

driver-core/src/main/java/com/datastax/driver/core/ProtocolOptions.java
ifesdjeen: Good point. I'm pretty sure @adutra mentioned it in a different context`allowsBetaProtocolVersions`, sorry I missed it.

driver-core/src/main/java/com/datastax/driver/core/Frame.java
adutra: > Could negotiation somehow succeed and result in v5/no beta? That would be bad because the driver could fail on some of the not-yet-implemented-on-the-client v5 features.

Indeed you have a point here.

Some background first: so far we've been working on the following assumptions:
1. The driver should have `Protocol.NEWEST_SUPPORTED` pointing to the highest version available, even if it is a beta one.
2. Suppose that current driver has versions: X beta, X-1 stable
3. `Protocol.NEWEST_SUPPORTED` points to X, which is beta
4. Driver requests connection with `Protocol.NEWEST_SUPPORTED` = X (beta)

Expected Results:

<table>
<tr>
<th></th>
<th>1. Without USE_BETA flag</th>
<th>2. With USE_BETA flag</th>
</tr>
<tr>
<th>A. X is beta server-side</th>
<td>Connects with X-1 (after renegotiation) <br/> :white_check_mark:</td>
<td>Connects with X <br/> :white_check_mark:</td>
</tr>
<tr>
<th>B. X is not beta server-side</th>
<td>Connects with X <br/> :x:</td>
<td>Connects with X <br/> :warning:</td>
</tr>
</table>


As we can see, our initial assumptions had problems in some cases: if the server is newer and has X stable, our driver would most likely fail to connect if `USE_BETA` was not requested (case B1). It would probably also fail to connect even if the flag was requested (case B2), but that's more acceptable because `USE_BETA` implies "Bad Things ™️  can happen".

So I would suggest the following improvements:
1. It would indeed be better if the driver completely prevent B1 from happening at all. That would mean the driver would have to store information about beta versions:
   1. `Protocol.NEWEST_SUPPORTED` would be the last stable (V4)
   2. `Protocol.NEWEST_BETA` would be the last beta, or null
   3. If the user does not specify a version and do not call `allowBetaVersions()`, use `Protocol.NEWEST_SUPPORTED`
   4. If the user does not specify a version and _does_ call `allowBetaVersions()`, use `Protocol.NEWEST_BETA`, if not null, or `Protocol.NEWEST_SUPPORTED` otherwise
2. `BetaVersionFlagAdder` might become obsolete in this new design; the flag could be added in `ProtocolEncoder` and `com.datastax.driver.core.Message.Request` could convey the request for the flag, pretty much like `isTracingRequested()`.

An alternative solution would be to have the _server_ reject B1 and B2 on the grounds that from its point of view, X is not beta. That would also simplify things a lot, but I don't know if it is still time to amend the code server-side.

driver-core/src/main/java/com/datastax/driver/core/ProtocolOptions.java
adutra: Fair enough, let's keep it that way.

driver-core/src/main/java/com/datastax/driver/core/Frame.java
olim7t: > our driver would most likely fail to connect if USE_BETA was not requested (case B1)

I'm not sure how the current version of the PR behaves, but what I would expect is "connect with X-1" (the highest stable supported by both sides).

> It would probably also fail to connect even if the flag was requested (case B2), but that's more acceptable

Agreed. That also applies to the case where the server is on a _different_ X-beta -- i.e. more beta features were added to X since the driver started supporting it (see my [comment on 10786](https://issues.apache.org/jira/browse/CASSANDRA-10786?focusedCommentId=15344506&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15344506)).

> Request could convey the request for the flag, pretty much like isTracingRequested()

Yes, that would avoid the additional handler.

> If the user does not specify a version and do not call allowBetaVersions(), use Protocol.NEWEST_SUPPORTED

Wouldn't it be slightly simpler if we didn't allow auto-negotiation with beta versions?
If you want to use a beta, you would have to specify the version as well (we could even have an alternate method `Cluster.Builder#withBetaProtocolVersion(ProtocolVersion)`).
With auto-negotiation, you would always end up with a stable.

olim7t:
Just to confirm that I have reproduced the problematic case "B1" locally. I recompiled Cassandra with `CURRENT_VERSION = VERSION_5` in `Server.java` to simulate completed V5 support. With that change, `should_not_connect_when_beta_version_explicitly_required_and_flag_not_set` (force V5 without beta flag) fails because the driver successfully connects with V5.

The driver should have the notion of whether the latest version is beta on its side, regardless of what the server says. See Alexandre's suggestions above.

driver-core/src/main/java/com/datastax/driver/core/Frame.java
adutra: > Wouldn't it be slightly simpler if we didn't allow auto-negotiation with beta versions?
> If you want to use a beta, you would have to specify the version as well (we could even have an alternate method `Cluster.Builder#withBetaProtocolVersion(ProtocolVersion)`).

Yes we could, and that was close to what @ifesdjeen had in his early versions. I found it a bit unfriendly as that would throw an `UnsupportedProtocolVersionException` back to the user in case A1 in my table above. So I suggested instead the opposite as I thought it would be interesting for users to simply declare that they are willing to use beta versions (and cope with the consequences), without actually specifying the version. This way renegotiation would take care of downgrading the version gracefully if need be.

But long story short, I'm fine if we decide to go down the path you suggested, specially if it simplifies the code.

olim7t: Considering this from the Cassandra ticket:

> This is primarily useful for driver authors to start work against a new protocol version when the work on that spans multiple releases. Users would not generally be expected to utilize this flag, although it could potentially be used to offer early feedback on new protocol features.

It's fair to assume that clients requesting USE_BETA will always have a specific version in mind, so negotiation is not needed. I vote for taking the simple path and not allowing negotiation with betas.

> that would throw [...] in case A1 in my table above

Even with the above suggestion it shouldn't. If X is beta server-side and the driver requests negotiation (hence without USE_BETA), I would expect X-1 as a result. In other words, negotiation would result in the highest common stable.

ifesdjeen:
I've done the changes you suggested, with the exception of `withBetaProtocolVersion` doesn't accept a specific version, since I thought we'll have supported 1 in-development beta version at a time.

Trying to set a version explicitly in combination with beta flag will now fail in both cases: when beta flag is set before version or vice versa.

I wanted to point out that the driver in beta will _still_ be able to connect to newer non-beta server, since server would "ignore" beta flag when v5 is released and allow connecting to it despite the beta version of driver. If we want to avoid it, we can open a follow-up issue in Cassandra and discuss it.

olim7t:
> I wanted to point out that the driver in beta will still be able to connect to newer non-beta server

I think that's acceptable under the "bad things might happen when the driver is in beta" rule.

ifesdjeen:
Agreed.

ifesdjeen:
@olim7t @adutra could you take a look at the final version?

olim7t:
Changed the base branch since we plan to rename 3.0 to 3.x.
@ifesdjeen we've not forgotten you, I will look at your latest changes as soon as 3.0.4 is released.

ifesdjeen:
> we've not forgotten you

Thank you! 🎆  Looking forward to 3.0.4!

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -808,6 +837,12 @@ public Builder withMaxSchemaAgreementWaitSeconds(int maxSchemaAgreementWaitSecon
          * @return this Builder.
          */
         public Builder withProtocolVersion(ProtocolVersion version) {
+            if (allowBetaProtocolVersion)
+                throw new IllegalStateException("Can not set the version explicitly if `allowBetaProtocolVersion` was used.");
+            if (version.toInt() > ProtocolVersion.NEWEST_SUPPORTED.toInt())
```
olim7t: Small nit: I would prefer `version.compareTo(ProtocolVersion.NEWEST_SUPPORTED) > 0`

olim7t (commented):
LGTM 👍 
I like this version much better, the addition of `NEWEST_BETA` makes things very clear.

stamhankar999:
I notice that the v5 protocol spec returns a slightly different response format for read-failure and write-failure. Shouldn't you account for that? See section 9 of the v5 spec for details.

olim7t:
@stamhankar999 this pull request focuses on the beta flag. But indeed other aspects of the new protocol version will need to be supported as they become available.

ifesdjeen:
@olim7t sorry it took so long. I've fixed the last comment and rebased on top of the latest `3.x`.

ifesdjeen:
Is there any change we could get this one merged? We have an issue (or multiple) that may rely on this. Also, Cassandra patch is already in trunk for some time...

driver-core/src/test/java/com/datastax/driver/core/ProtocolBetaVersionTest.java
```diff
@@ -0,0 +1,152 @@
...
+import static com.datastax.driver.core.ProtocolVersion.V4;
+import static com.datastax.driver.core.ProtocolVersion.V5;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+/**
+ * Tests for the new USE_BETA flag introduced in protocol v5
+ * and Cassandra 3.10.
+ */
+@CCMConfig(version = "3.10")
```
tolbertam, line 29: Unfortunately this doesn't work at the moment since 3.10 isn't released (yet) and the `cassandra-3.10` branch is gone and the tag isn't there yet.  However, 👍 on leaving it as is as it will eventually work and we can verify manually until then.

driver-core/src/test/java/com/datastax/driver/core/ProtocolBetaVersionTest.java
tolbertam: Isn't there now no difference between this test and `should_connect_with_beta_when_no_version_explicitly_required_and_flag_set()`?  I think technically this test can be deleted now that the protocol version should not be provided with `allowBetaProtocolVersion`

driver-core/src/test/java/com/datastax/driver/core/ProtocolBetaVersionTest.java
```diff
@@ -0,0 +1,152 @@
...
+     * @jira_ticket JAVA-1248
+     */
+    @Test(groups = "short")
+    public void should_connect_with_beta_when_beta_version_explicitly_required_and_flag_set() throws Exception {
+        Cluster cluster = Cluster.builder()
+                .addContactPoints(getContactPoints())
+                .withPort(ccm().getBinaryPort())
+                .allowBetaProtocolVersion()
+                .build();
+        cluster.connect();
```
tolbertam: Also oddly against `cassandra-3.11` this test currently fails, here's the error on the C* side:

```
ERROR [Native-Transport-Requests-2] 2017-01-05 21:54:33,272 Message.java:623 - Unexpected exception during request; channel = [id: 0xc8ec7c60, L:/127.0.1.1:60045 - R:/127.0.1.1:60141]
java.lang.IndexOutOfBoundsException: readerIndex(50) + length(4) exceeds writerIndex(51): SlicedAbstractByteBuf(ridx: 50, widx: 51, cap: 51/51, unwrapped: PooledUnsafeDirectByteBuf(ridx: 118, widx: 160, cap: 1024))
        at io.netty.buffer.AbstractByteBuf.checkReadableBytes0(AbstractByteBuf.java:1179) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.buffer.AbstractByteBuf.readInt(AbstractByteBuf.java:638) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.buffer.AbstractByteBuf.readUnsignedInt(AbstractByteBuf.java:646) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at org.apache.cassandra.cql3.QueryOptions$Codec.decode(QueryOptions.java:405) ~[main/:na]
        at org.apache.cassandra.cql3.QueryOptions$Codec.decode(QueryOptions.java:366) ~[main/:na]
        at org.apache.cassandra.transport.messages.QueryMessage$1.decode(QueryMessage.java:48) ~[main/:na]
        at org.apache.cassandra.transport.messages.QueryMessage$1.decode(QueryMessage.java:44) ~[main/:na]
        at org.apache.cassandra.transport.Message$ProtocolDecoder.decode(Message.java:281) ~[main/:na]
        at org.apache.cassandra.transport.Message$ProtocolDecoder.decode(Message.java:262) ~[main/:na]
        at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:88) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:366) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:352) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:345) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.handler.codec.MessageToMessageDecoder.channelRead(MessageToMessageDecoder.java:102) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:366) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:352) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:345) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:293) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.handler.codec.ByteToMessageDecoder.fireChannelRead(ByteToMessageDecoder.java:280) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:396) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:248) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:366) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:352) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:345) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1294) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:366) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:352) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:911) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:131) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:572) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:513) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:427) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:399) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.util.concurrent.SingleThreadEventExecutor$2.run(SingleThreadEventExecutor.java:140) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:144) ~[netty-all-4.0.39.Final.jar:4.0.39.Final]
        at java.lang.Thread.run(Thread.java:745) [na:1.8.0_112]
```

tolbertam: I see, it's because [Protocol V5 expects](https://github.com/apache/cassandra/blob/cassandra-3.11/doc/native_protocol_v5.spec#L1183-L1184) unsigned int for query options flags [QueryOptions code](https://github.com/apache/cassandra/blob/cassandra-3.11/src/java/org/apache/cassandra/cql3/QueryOptions.java#L405)  (see also: [PYTHON-684](https://datastax-oss.atlassian.net/browse/PYTHON-684)).  I've fixed it locally.  I guess this shows the value of using beta to begin with :smile: (to allow us to evaluate and make protocol changes as we go)!  I'm not sure we need to fix in this PR, but here's [a gist of the patch](https://gist.github.com/tolbertam/eb157f2991f907a66626e9365592436d) to get things working (opened up [JAVA-1362](https://datastax-oss.atlassian.net/browse/JAVA-1362) for that).

tolbertam:
Other than my feed back [here](https://github.com/datastax/java-driver/pull/718/files#r94896858) about the now-extraneous test case, this looks good to me, great work!

driver-core/src/test/java/com/datastax/driver/core/ProtocolBetaVersionTest.java
tolbertam: Isn't there now no difference between this test and `should_connect_with_beta_when_no_version_explicitly_required_and_flag_set()`?  I think technically this test can be deleted now that the protocol version should not be provided with `allowBetaProtocolVersion`.

tolbertam (changes requested):
Other than my comment about ` should_connect_with_beta_when_beta_version_explicitly_required_and_flag_set()` no longer being needed (I think)?  This looks good to me!

driver-core/src/test/java/com/datastax/driver/core/ProtocolBetaVersionTest.java
olim7t: Agreed, I'll clean up the tests.

tolbertam (approved):

ifesdjeen:
Thank you!

PR #719, opened 2016-08-03
JAVA-1255: Allow nested UDTs to be used in Mapper

tolbertam:
For [JAVA-1255](https://datastax-oss.atlassian.net/browse/JAVA-1255)

### DIFF
--- a/driver-mapping/src/main/java/com/datastax/driver/mapping/AnnotationParser.java
+++ b/driver-mapping/src/main/java/com/datastax/driver/mapping/AnnotationParser.java
@@ -201,6 +201,9 @@ class AnnotationParser {
                 throw new IllegalArgumentException(String.format("Field %s does not exist in type %s.%s",
                         propertyMapper.columnName, ksName, userType.getTypeName()));
 
+            for (Class<?> fieldUdt : TypeMappings.findUDTs(propertyMapper.javaType.getType()))
+                mappingManager.getUDTCodec(fieldUdt);
+
             propertyMappers.put(propertyMapper.columnName, propertyMapper);
         }
 
--- a/driver-mapping/src/test/java/com/datastax/driver/mapping/MapperNestedUDTTest.java
+++ b/driver-mapping/src/test/java/com/datastax/driver/mapping/MapperNestedUDTTest.java
@@ -1,0 +1,271 @@
+/*
+ *      Copyright (C) 2012-2015 DataStax Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


driver-mapping/src/main/java/com/datastax/driver/mapping/AnnotationParser.java
```diff
@@ -194,6 +194,9 @@ else if (propertyMapper.isClusteringColumn())
                 throw new IllegalArgumentException(String.format("Field %s does not exist in type %s.%s",
                         propertyMapper.columnName, ksName, userType.getTypeName()));
 
+            for (Class<?> fieldUdt : TypeMappings.findUDTs(field.getGenericType()))
```
tolbertam: Not sure if this is the right fix, but it does work.  Alternatively, `TypeMappings.findUDTs` could recurse nested UDTs like it does for Collections, wdyt?

adutra:
That looks like an oversight when I refactored the code. The old `ReflectionFactory` used to [scan for UDTs](https://github.com/datastax/java-driver/blob/66cfc42660509b8ddf27606fafd42246d820555d/driver-mapping/src/main/java/com/datastax/driver/mapping/ReflectionMapper.java#L99-L100) for both columns and fields, but I only ported these two lines of code to [`parseEntity`](https://github.com/datastax/java-driver/blob/b52e7dca65ce8ea6ec8f1e95dd49bde0e134ab8e/driver-mapping/src/main/java/com/datastax/driver/mapping/AnnotationParser.java#L142-L142), and forgot to call them in `parseUDT`. 
What is sad is that we did not have any nested UDT test so far, so I'm glad that we now have one!

driver-mapping/src/main/java/com/datastax/driver/mapping/AnnotationParser.java
adutra: Your fix is in the right place, however, you need to adjust the code to:

```
for (Class<?> fieldUdt : TypeMappings.findUDTs(propertyMapper.javaType.getType()))
                mappingManager.getUDTCodec(fieldUdt);
```

Because some properties might only be present through getters and setters and not through fields.

adutra:
Went ahead and applied my comment above. lgtm 👍 

Although, we might want to create a separate branch, 3.1.x, to host fixes for 3.1.0, wdyt? We might also consider renaming 3.0 to 3.x as well.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperNestedUDTTest.java
```diff
@@ -135,85 +137,74 @@ public int hashCode() {
...
-            this.a = a;
-            this.b = b;
-            this.c = c;
-            this.d = d;
-        }
+        // a somewhat artificial way to store coordinates,
+        // but the goal is to test that nested UDTs can be discovered through
+        // getters instead of fields
+        @Transient
+        private final Point[] points;
```
tolbertam, line 145: Nice 👍

tolbertam:
Thanks for reviewing and adjusting for getter/setters support :).  I agree with reopening this against a 3.1 specific branch when we have that 👍

olim7t (approved):

PR #722, opened 2016-08-08
JAVA-1260: Backport JAVA-1002 to 2.0

tolbertam:
For [JAVA-1260](https://datastax-oss.atlassian.net/browse/JAVA-1260)


tolbertam:
the jdk6 build is broken in travis for 2.0 (have fixed for newer versions).  We can depend on our internal jenkins ci for that.

PR #723, opened 2016-08-11
JAVA-1261: Throw error when attempting to page in I/O thread

olim7t:
Iterating a result set past the current page triggers a blocking call
for the next page. This should never be done in a Netty I/O thread (for
example by iterating in a future callback), because it can deadlock.

Similar to JAVA-1182, we throw an error so that client code can detect
this mistake early.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -3,6 +3,8 @@
 ### 3.0.4 (in progress)
 
 - [improvement] JAVA-1246: Driver swallows the real exception in a few cases
+- [improvement] JAVA-1261: Throw error when attempting to page in I/O thread.
+
 
 ### 3.0.3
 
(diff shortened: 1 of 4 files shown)


tolbertam:
looks good to me 👍 .  Made a few changes in tests that can be squashed.

engharb:
Dears,
 I am a biggener in using Apache-Cassandra. My question How can I integrate your contribution (this java driver implementation) to my Application/Project?

best regards,

olim7t:
@engharb if you're referring to this specific fix, the easiest way would be to wait for the 3.0.4 release (follow the [mailing list](https://groups.google.com/a/lists.datastax.com/forum/#!forum/java-driver-user) or [twitter](https://twitter.com/dsJavaDriver) for release notifications). We don't commit to release dates in advance, but it should be a few weeks from now. Otherwise you can checkout the `java1261` branch and build a snapshot.

PR #725, opened 2016-08-11
Set appveyor os to 'Previous Visual Studio 2015' temporarily.

tolbertam:
[CASSANDRA-12278](https://issues.apache.org/jira/browse/CASSANDRA-12278) prevents C\* from working with j8u100+.  Previous image
has an older version of the JDK.

Also log hardwareinfo and update to cassandra 3.7.

Will cherry pick to 3.0 afterwords if this gets merged.


tolbertam:
Might just be better to wait for 3.8 which should be coming soon.  Lets leave this PR open until then.

PR #726, opened 2016-08-12
JAVA-1263: Eliminate memory copies in LZ4Compressor

normanmaurer:
Motivation:

Use ByteBuffer for compress / decompress to reduce memory copies / object allocations.

Modifications:
- Use ByteBuffer and so not need to do any extra memory copies for compress / decompress.

Result:

Faster compression / decompression when using lz4.

PR #727, opened 2016-08-12
JAVA-1263: Eliminate memory copies in SnappyFrameCompressor and use netty's allo…

normanmaurer:
…cator for allocations

Motiviation:

Snappy allows to use ByteBuffer for compress/decompress when direct buffers are used. At the moment a memory copy to heap buffers is done, which is wasteful.

Modification:

Use ByteBuffer for compress/decompress when possible and allocate the buffer from the ByteBufAllocator of netty which may be a pooled implementation.

Result:

Less memory copies == Better performance.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/FrameCompressor.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/FrameCompressor.java
@@ -16,6 +16,7 @@
 package com.datastax.driver.core;
 
 import com.datastax.driver.core.exceptions.DriverInternalError;
+import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
 import net.jpountz.lz4.LZ4Factory;
 import org.slf4j.Logger;
@@ -23,6 +24,7 @@ import org.slf4j.LoggerFactory;
 import org.xerial.snappy.Snappy;
 
 import java.io.IOException;
+import java.nio.ByteBuffer;
 
 abstract class FrameCompressor {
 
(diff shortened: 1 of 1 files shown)


tolbertam:
I noticed a small difference between this PR and #726 is that this checks if the input buffer is a direct buffer and then behaves accordingly when creating the output buffer (whether to use heap buffer or direct buffer).   For some reason (still debugging) when using `-Dio.netty.noPreferDirect=true` performance takes a really big hit.  Will get a profile and share why shortly.

normanmaurer:
@tolbertam which one take a big hit.. This one or #726 ?

tolbertam:
Ah yes, sorry I should have clarified, it is this one.  Maybe i have something misconfigured, but JFC is showing heapBuffer() allocating a lot of memory outside TLAB:

image

For context previously the combined memory allocated without `-Dio.netty.noPreferDirect=true` is around 330MB in new TLAB after 1 million writes.   With the system property ~1TB was allocated outside TLAB after ~60k writes.

normanmaurer:
@tolbertam ha! I know why ;) Because I suck... Let me fix the bug :)

normanmaurer:
@tolbertam pushed an update... Please check again, I think this is fixed now

normanmaurer:
@tolbertam PTAL again buddy :)

tolbertam:
Ah, I see what you did there!  Trying it out now and will let you know how it goes :)

normanmaurer:
@tolbertam yep basically I not calculated the max length correctly

tolbertam:
Looking great!   Some numbers recorded from 1 million op stress with the following configurations:

`pre`: Existing code
`post`: This PR
`pre_npd`: Pre with `-Dio.netty.noPreferDirect=true`
`post_npd`: Post with `-Dio.netty.noPreferDirect=true`

There are a few things I need to look into:
- why decompress() allocates more in the write cases for the new code (maybe some small overhead when the response payload is small?).  
- why compress() is so small in the `post_npd` cases, guessing it has something to do with how i'm parsing the flight recording (heapBuffer allocates outside TLAB, and I think I'm capturing that right, but could be wrong).  I think it might be because the `HeapArenas` are already created (i see 9 objects outside TLAB that were created by other parts of code.

```
     pre_write: TLAB Size: compress(): 642.79MB decompress(): 137.69MB combined: 780.48MB, which is 07.74% of overall:   9.85GB
    post_write: TLAB Size: compress(): 139.28MB decompress(): 190.87MB combined: 330.15MB, which is 03.76% of overall:   8.58GB

 pre_npd_write: TLAB Size: compress(): 615.91MB decompress(): 134.00MB combined: 749.90MB, which is 06.48% of overall:  11.30GB
post_npd_write: TLAB Size: compress():  21.50MB decompress(): 193.15MB combined: 214.65MB, which is 02.23% of overall:   9.40GB

---

     pre_read : TLAB Size: compress(): 185.45MB decompress(): 511.39MB combined: 696.84MB, which is 08.15% of overall:   8.35GB
    post_read : TLAB Size: compress(): 125.78MB decompress(): 160.83MB combined: 286.61MB, which is 03.34% of overall:   8.38GB

 pre_npd_read : TLAB Size: compress(): 227.82MB decompress(): 498.09MB combined: 725.92MB, which is 07.98% of overall:   8.88GB
post_npd_read : TLAB Size: compress():  13.75MB decompress(): 213.96MB combined: 227.71MB, which is 02.14% of overall:  10.40GB

---

     pre_mixed: TLAB Size: compress(): 459.93MB decompress(): 320.22MB combined: 780.15MB, which is 08.89% of overall:   8.57GB
    post_mixed: TLAB Size: compress(): 151.64MB decompress(): 194.06MB combined: 345.70MB, which is 04.14% of overall:   8.16GB

 pre_npd_mixed: TLAB Size: compress(): 461.39MB decompress(): 299.57MB combined: 760.96MB, which is 08.61% of overall:   8.63GB
post_npd_mixed: TLAB Size: compress():  22.59MB decompress(): 210.88MB combined: 233.48MB, which is 02.69% of overall:   8.47GB
```

tolbertam:
Another thing I noticed is that when using `-Dio.netty.noPreferDirect=true` that on the decompress path we are still working with Direct ByteBufs.  Might be somewhere in the driver code path where we are calling `directBuffer` explicitly, so the non-direct path is not exercised.  I'll look into that but not something to do with this PR.

normanmaurer:
@tolbertam thats most likely because what you see on the decompress part is still a direct buffer (as a direct buffer is needed to read from the socket).

tolbertam:
@normanmaurer, ahh i see, that makes sense.  I did some debugging to see where things were allocated on the read path and found it here in `AbstractByteBufAllocator#ioBuffer`

``` java
    @Override
    public ByteBuf ioBuffer(int initialCapacity) {
        if (PlatformDependent.hasUnsafe()) {
            return directBuffer(initialCapacity);
        }
        return heapBuffer(initialCapacity);
    }
```

If I use `-Dio.netty.noUnsafe=true`, I might be able to exercise that path, trying it out.

tolbertam:
Looks like that worked!  I'll give it a spin.

tolbertam:
Looking good, in all cases the overall memory allocation is cut at least in half.  In the noPreferDirect/noUnsafe cases the existing pool arenas are reused.

It does seem odd how in the 'write' case `decompress` uses more memory with this patch, but it is probably a consequence in there being a little more overhead when there is little/no payload in the response?   This also happens on the LZ4 PR too, I think it is ok.
- `pre`: Existing code
- `post`: This PR
- `pre_npd_u`: Pre with `-Dio.netty.noPreferDirect=true -Dio.netty.noUnsafe=true`
- `post_npd_u`: Post with `-Dio.netty.noPreferDirect=true -Dio.netty.noUnsafe=true`

```
       pre_write: TLAB Size: compress(): 642.79MB decompress(): 137.69MB combined: 780.48MB, which is 07.74% of overall:   9.85GB
      post_write: TLAB Size: compress(): 139.28MB decompress(): 190.87MB combined: 330.15MB, which is 03.76% of overall:   8.58GB

 pre_npd_u_write: TLAB Size: compress(): 558.51MB decompress(): 129.94MB combined: 688.45MB, which is 06.39% of overall:  10.52GB
post_npd_u_write: TLAB Size: compress():  21.52MB decompress():  31.42MB combined:  52.94MB, which is 00.58% of overall:   8.92GB

---

       pre_read : TLAB Size: compress(): 185.45MB decompress(): 511.39MB combined: 696.84MB, which is 08.15% of overall:   8.35GB
      post_read : TLAB Size: compress(): 125.78MB decompress(): 160.83MB combined: 286.61MB, which is 03.34% of overall:   8.38GB

 pre_npd_u_read : TLAB Size: compress(): 286.68MB decompress(): 486.23MB combined: 772.91MB, which is 08.30% of overall:   9.10GB
post_npd_u_read : TLAB Size: compress():  18.34MB decompress():  28.46MB combined:  46.81MB, which is 00.53% of overall:   8.63GB

---

       pre_mixed: TLAB Size: compress(): 459.93MB decompress(): 320.22MB combined: 780.15MB, which is 08.89% of overall:   8.57GB
      post_mixed: TLAB Size: compress(): 151.64MB decompress(): 194.06MB combined: 345.70MB, which is 04.14% of overall:   8.16GB

 pre_npd_u_mixed: TLAB Size: compress(): 442.32MB decompress(): 286.99MB combined: 729.31MB, which is 08.10% of overall:   8.79GB
post_npd_u_mixed: TLAB Size: compress():  24.79MB decompress():  34.16MB combined:  58.94MB, which is 00.73% of overall:   7.91GB
```

normanmaurer:
@tolbertam can you give some details about the "write" case ? what exactly this is all about ?

tolbertam:
I should have included that here.  The scenario is as follows:

I'm using cassandra-stress with the built driver code against a single node on cassandra trunk with the following parameters:

```
tools/bin/cassandra-stress <mode> n=1000000 -rate threads=500 -mode native cql3 compression=snappy
```

Recorded allocation profiling using JFC with the following parameters to cassandra-stress ([profile](https://gist.github.com/jmiettinen/1976f9bbe156ecb5e232))

```
-XX:FlightRecorderOptions=dumponexit=true,dumponexitpath=$JFR,maxsize=10G,maxage=1d,disk=true, \
-XX:StartFlightRecording=defaultrecording=true,settings=allocation_profiling.jfc \
```

The modes used are 'write', 'read', and 'mixed'.   Each write inserts 5 34-byte blob columns for a key, and the read and mixed workloads read the data back.  So in the 'write' case, we expect at least 170 bytes to compress, very little to decompress (since we get an empty result back), where in the read case it is the opposite, and the 'mixed' case is a mixture of both.

normanmaurer:
@tolbertam so what you say is that if you test the write case decompress(...) will be called with a very small buffer and with this pr the allocations are bigger compared to before ? can you tell me if we go in the `isDirect()` block or not ? And also if you use the `PooledByteBufAllocator` for this test or not ?

tolbertam:
> so what you say is that if you test the write case decompress(...) will be called with a very small buffer and with this pr the allocations are bigger compared to before ?

I think that is what is happening.  There is very little coming back in the response payload in the `write` case since there are no rows (since it was an insert query).  So using ByteBuffer instead of a byte array maybe introduces a very small overhead that wouldn't otherwise be noticeable if you actually had some payload to decompress.  I don't think we need to optimize around that case though.

> can you tell me if we go in the isDirect() block or not?

In the `post` case it does, its only when i used `-Dio.netty.noUnsafe=true` that it doesn't (i.e. `post_npd_u_write`)

> And also if you use the PooledByteBufAllocator for this test or not ?

I am not, but I think if I used that the memory allocations would be reduced and look a lot like when I was using `-Dio.netty.noPreferDirect=true` since heapBuffer uses a pooled buffer allocator.   Using `-Dio.netty.allocator.type=pooled` should kick that into gear right?  I'll give it a shot.

tolbertam:
Guess I was wrong about that, [we are using PooledByteBufAllocator already](https://github.com/datastax/java-driver/blob/3.0.x/driver-core/src/main/java/com/datastax/driver/core/NettyOptions.java#L139-L142).

screen shot 2016-08-15 at 3 58 55 pm

tolbertam:
For context, the previous implementation:

screen shot 2016-08-15 at 4 12 30 pm

tolbertam:
I see, I might have misunderstood the benefit initially of using ByteBuf making use of a pooled  implementation (this is a bit new to me).  In the case of using a pooled implementation there isn't actually new bytes being allocated for the compressed/uncompressed payloads as it can re-use the memory allocated from the pool, there is just some overhead in creating new DirectByteBuffer instances in internalNioBuffer (duplicate() / slice() - not actually copying the bytes but creating DirectByteBuffer instances).

In the write scenario,  `decompress()` with the previous implementation didn't allocate much to begin with (137.69MB overall), where in the read scenario it did (511.39MB).   With the changes, the overhead in creating new `DirectByteBuffer` objects surpasses that of the previous implementation where it was creating tiny byte arrays and then wrapping them.  Where in the case where there is actually data to decompress (read) that overhead does not surpass it (511.39MB vs. 160.83MB).

The allocations are effectively the same in all three modes (read, write, mixed) for both compress and decompress within some margin.  This makes sense given that a pooled byte buf allocator is used

```
compress() post:
write: 139.28MB
read: 125.78MB
mixed: 151.64MB

decompress() post:
write: 190.87MB
read: 160.83MB
mixed: 194.06MB
```

compared to the previous implementation where it was more of a function of the size of the request/response payloads:

```
compress() pre:
write: 642.79MB
read: 184.45MB
mixed: 459.93MB

decompress() pre:
write: 137.69MB
read: 511.39MB
mixed: 320.22MB
```

Hopefully I'm understanding that right.

driver-core/src/main/java/com/datastax/driver/core/FrameCompressor.java
```diff
@@ -58,23 +60,119 @@ private SnappyCompressor() {
...
-            byte[] output = new byte[Snappy.uncompressedLength(input)];
-            int size = Snappy.uncompress(input, 0, input.length, output, 0);
-            return frame.with(Unpooled.wrappedBuffer(output, 0, size));
+            ByteBuf input = frame.body;
+            final ByteBuf frameBody;
+
+            if (input.isDirect()) {
+                // Using internalNioBuffer(...) as we only hold the reference in this method and so can
+                // reduce Object allocations.
+                ByteBuffer in = input.internalNioBuffer(input.readerIndex(), input.readableBytes());
```
tolbertam, line 124: Just noticed on #726 that for the input buffer you made a convenience method that was used checked the nioBufferCount on the input buffer, guessing we need this here too?

``` java
private static ByteBuffer inputNioBuffer(ByteBuf buf, int index, int len) {
    // Using internalNioBuffer(...) as we only hold the reference in this method and so can
    // reduce Object allocations.
    return buf.nioBufferCount() == 1 ? buf.internalNioBuffer(index, len) : buf.nioBuffer(index, len);
}
```

tolbertam:
Created a combined PR (#753) with commit from this and #726 with some additional changes, thanks!

PR #728, opened 2016-08-12
JAVA-1262: Use ParseUtils for quoting & unquoting

tolbertam:
For [JAVA-1262](https://datastax-oss.atlassian.net/browse/JAVA-1262)

Metadata.handleId and quote use String.replace or replaceAll, which each
create a lot of garbage as they compile a Pattern each time.  Instead
use ParseUtils.quote and unquote for quoting and unquoting.

Adds doubleQuote, unDoubleQuote, and isDoubleQuoted as analogs to the
single quoted counterparts in ParseUtils.

Memory allocation difference running a 5,000,000 write cassandra-stress:

| configuration | Metadata.quote() | Metadata.handleId() | total | % of total |
| --- | --- | --- | --- | --- |
| 3.0.x | 3.18GB | 3.77GB | 43.54GB | 15.96% |
| java1262 | 496.47MB | 1.30GB | 36.69GB | 4.87% |

A good portion of Metadata.handleId() (880.63 MB) is from pattern checking for determining whether or not to lowercase a non-quoted string.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
@@ -232,7 +232,7 @@ public abstract class ParseUtils {
      * by single quotes, and {@code false} otherwise.
      */
     public static boolean isQuoted(String value) {
-        return value != null && value.length() > 1 && value.charAt(0) == '\'' && value.charAt(value.length() - 1) == '\'';
+        return isQuoted(value, '\'');
     }
 
     /**
@@ -243,7 +243,7 @@ public abstract class ParseUtils {
      * @return The quoted string.
      */
     public static String quote(String value) {
-        return '\'' + replaceChar(value, '\'', "''") + '\'';
+        return quote(value, '\'');
     }
 
     /**
@@ -254,9 +254,41 @@ public abstract class ParseUtils {
      * @return The unquoted string.
      */
     public static String unquote(String value) {
-        if (!isQuoted(value))
-            return value;
-        return value.substring(1, value.length() - 1).replace("''", "'");
+        return unquote(value, '\'');
+    }
+
+    /**
+     * Return {@code true} if the given string is surrounded
+     * by double quotes, and {@code false} otherwise.
+     *
+     * @param value The string to inspect.
+     * @return {@code true} if the given string is surrounded
+     * by double quotes, and {@code false} otherwise.
+     */
+    public static boolean isDoubleQuoted(String value) {
+        return isQuoted(value, '\"');
+    }
+
+    /**
+     * Double quote the given string; double quotes are escaped.
+     * If the given string is null, this method returns a quoted empty string ({@code ""}).
+     *
+     * @param value The value to double quote.
+     * @return The double quoted string.
+     */
+    public static String doubleQuote(String value) {
+        return quote(value, '"');
+    }
+
+    /**
+     * Unquote the given string if it is double quoted; double quotes are unescaped.
+     * If the given string is not double quoted, it is returned without any modification.
+     *
+     * @param value The string to un-double quote.
+     * @return The un-double quoted string.
+     */
+    public static String unDoubleQuote(String value) {
+        return unquote(value, '"');
     }
 
     /**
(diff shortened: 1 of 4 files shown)


driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
```diff
@@ -428,45 +460,114 @@ public static String formatTime(long value) {
...
-     * @param text        The text.
-     * @param search      The character to search for.
-     * @param replacement The replacement.
-     * @return The text with all occurrences of {@code search} replaced with {@code replacement}.
+     * @param text      The text.
+     * @param quoteChar The character to use as a quote.
+     * @return The text with surrounded in quotes with all existing quotes escaped with (i.e. ' becomes '')
      */
-    private static String replaceChar(String text, char search, String replacement) {
+    private static String quote(String text, char quoteChar) {
```
tolbertam, line 498: `quote` and `unquote` are versions of replaceChar with enough differences to justify different methods.  I removed `replaceChar` altogether as it isn't used otherwise, can add it back if you think it is appropriate to keep.

driver-core/src/test/java/com/datastax/driver/core/ParseUtilsTest.java
```diff
@@ -60,4 +60,43 @@ public void testIsQuoted() {
         assertThat(ParseUtils.isQuoted("'''foo'''")).isTrue();
     }
 
+    @Test(groups = "unit")
```
tolbertam, line 66: These tests are pretty much copy-pasted versions of the quoted tests, but figured it would be better to have a little duplication then to risk making a mistake when refactoring around common code.  Can refactor though if you like.

driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
```diff
@@ -428,45 +460,114 @@ public static String formatTime(long value) {
...
+            return emptyQuoted(quoteChar);
 
         int nbMatch = 0;
         int start = -1;
         do {
-            start = text.indexOf(search, start + 1);
+            start = text.indexOf(quoteChar, start + 1);
             if (start != -1)
                 ++nbMatch;
         } while (start != -1);
```
newkek, line 509: Would be cool to have a fast exit if `nbMatch == 0` as there was before

driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
```diff
@@ -428,45 +460,114 @@ public static String formatTime(long value) {
...
+        String search = emptyQuoted(quoteChar);
+        int nbMatch = 0;
+        int start = -2;
+        do {
+            start = text.indexOf(search, start + 2);
+            // ignore the second to last character occurrence, as the last character is a quote.
+            if (start != -1 && start != text.length() - 2)
+                ++nbMatch;
+        } while (start != -1);
+
```
newkek, line 559: We could also fast exit here if no quote is found and just remove the surrounding quotes

driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
```diff
@@ -428,45 +460,114 @@ public static String formatTime(long value) {
...
+        // length of the new string will be its current length - the number of occurrences.
+        int newLength = text.length() - nbMatch - 2;
         char[] result = new char[newLength];
         int newIdx = 0;
-        for (int i = 0; i < text.length(); i++) {
+        boolean firstFound = false;
```
newkek, line 569: Why isn't it initialized to `true`?

driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
```diff
@@ -428,45 +460,114 @@ public static String formatTime(long value) {
...
+                if (firstFound) {
+                    // don't add to result, this action in effect removes consecutive characters.
+                    firstFound = false;
+                } else {
+                    firstFound = true;
+                    result[newIdx++] = c;
+                }
             } else {
                 result[newIdx++] = c;
+                firstFound = false;
```
newkek, line 585: Shouldn't it be set to `true`? Unless I'm misunderstanding something

driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
tolbertam, line 509: The reason it was removed was for adding quotes to the new String, where previously the quotes were added after calling `replaceChar`.  Alternatively if `nbMatch == 0` we could simply return `quoteChar + text + quoteChar`, but that doesn't buy much.

driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
tolbertam, line 569: `firstFound` is set to true when a `quoteChar` is found in the text, so we know that if we find that the next character is a `quoteChar` as well, we know to remove it.    It gets set to false when we find the consecutive `quoteChar` or when we encounter something that isn't a `quoteChar`.   I'll document that.

newkek, line 569: Ah ok I got, I understood it the other way around, as `"the first is found" -> true ? don't do anything : put it in the result array` whereas it's actually the other way around

tolbertam, line 569: exactly, if you find a quote character and the next character is also a quote, we skip adding the second quote to the result array and set firstFound to false.

driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
```diff
@@ -428,45 +460,118 @@ public static String formatTime(long value) {
...
+        return value != null && value.length() > 1
+                && value.charAt(0) == quoteChar && value.charAt(value.length() - 1) == quoteChar;
+    }
+
+    /**
+     * @param quoteChar " or '
+     * @return A quoted empty string.
+     */
+    private static String emptyQuoted(char quoteChar) {
+        // don't handle non quote characters, this is done so the these are interned and doesn't create
```
adutra: Small nit: "so that these are interned and don't create".

driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
newkek, line 509: > Alternatively if nbMatch == 0 we could simply return quoteChar + text + quoteChar, but that doesn't buy much.

Yeah that was the idea, we avoid looping on the whole char sequence if there's no match (`nbMatch == 0`) as this is apparently a hot part of the code. As maybe we can assume it's likely that there's not going to be many quoted Strings containing other quotes, people use it more for Caps letters and other symbols. But if you think it's not worth it ok

driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
adutra, line 559: Expanding on Kevin's suggestion, I think it would indeed be good to optimize around 2 cases:
- "simple" quoted identifiers (e.g. `"foobar"`) - i.e. when `nbMatch == 0`;
- empty strings.

These cases are probably statistically way more frequent than weird identifiers such as `"foo""bar"`.

I ran a quick JMH benchmark and optimizations do perform better by about 10 ns/op for the special cases (weird identifiers get slower though by about 5 ns/op).

tolbertam, line 559: Hrmm, I think i'm coming around on that.  The empty strings case is already covered for both quote/unquote.  I'll take care of simple identifiers.

The only problem is the quoting case.   I had done some testing and:

```
"'" + text + "'";
```

Seems to be optimized by the compiler (in terms of memory allocation) because the quotes are constant, but `quoteChar + text + quoteChar` isn't because quoteChar isn't a constant.   I'll do some more testing around that.

driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
adutra, line 509: >  if nbMatch == 0 we could simply return quoteChar + text + quoteChar, but that doesn't buy much.

Indeed, my JMH benchmark did not show any significant performance gain for `quote` (only 2 ns/op better) – much unlike the `unquote` method.

driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
adutra, line 559: > The empty strings case is already covered for both quote/unquote.

My bad, what I meant was to optimize empty _quoted_ strings, i.e. `""` by adding something like this:

``` java
if (search.equals(text))
            return "";
```

> `quoteChar + text + quoteChar` isn't because quoteChar isn't a constant

That's a valid point, maybe we should not optimize `quote` any further.

tolbertam, line 559: I see, that would also be covered if we check nbMatch == 0.  Rerunning the memory benchmark to see if I was remembering that correctly or not, should know shortly.

tolbertam, line 559: Perhaps I was wrong about that.  Short circuiting and returning seems to be the right course of action in the case of memory allocation as well (the following output is from a 1 million write cassandra-stress run):

```
1262_exist_write : TLAB Size: quote(): 104.10MB unquote(): 104.72MB combined: 208.83MB, which is 02.79% of overall:   7.31GB
1262_return_write: TLAB Size: quote():  64.35MB unquote():  59.53MB combined: 123.88MB, which is 01.44% of overall:   8.40GB
```

Thanks for the feedback @newkek and @adutra!  Will push up the change shortly.

tolbertam:
For apples to apples comparison with the original benchmark with 5 million writes (note that the total seems to fluctuate independent of the change, no worries there):

| configuration | Metadata.quote() | Metadata.handleId() | total | % of total |
| --- | --- | --- | --- | --- |
| 3.0.x | 3.18GB | 3.77GB | 43.54GB | 15.96% |
| java1262_o | 496.47MB | 1.30GB | 36.69GB | 4.87% |
| java1262 | 315.53MB | 1.14GB | 48.45GB | 2.99% |

java1262_o is the original PR, java1262 is with the recent change based on feedback (9019ada) to return early if there is no escaping/unescaping needed.    This reduced allocations some more which is nice.  The dominator is still the alphanumeric matcher, but the memory use from these two methods has reduced quite a bit (from 6.95GB to 1.45GB).

driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
```diff
@@ -428,45 +460,126 @@ public static String formatTime(long value) {
...
+     * @param quoteChar The character to use as a quote.
+     * @return The text with surrounding quotes removed and non surrounding quotes unescaped (i.e. '' becomes ')
+     */
+    private static String unquote(String text, char quoteChar) {
+        if (!isQuoted(text, quoteChar))
             return text;
 
-        int newLength = text.length() + nbMatch * (replacement.length() - 1);
+        String search = emptyQuoted(quoteChar);
+        int nbMatch = 0;
```
adutra, line 551: Looks good so far! We could be even more aggressive and return early here in case of empty quoted strings. How about:

```
if (search.equals(text))
            return "";
```

(not sure though if we gain a lot compared to what we currently have, will re-run my benchmark).

olim7t: Given that we already know that `isQuoted` holds true, we could even test with `if (text.length()  == 2)`.

driver-core/src/main/java/com/datastax/driver/core/ParseUtils.java
```diff
@@ -428,45 +460,126 @@ public static String formatTime(long value) {
...
+        int start = -2;
+        do {
+            start = text.indexOf(search, start + 2);
+            // ignore the second to last character occurrence, as the last character is a quote.
+            if (start != -1 && start != text.length() - 2)
+                ++nbMatch;
```
olim7t, line 557: > `start != text.length() - 2`

Is the goal here to handle malformed strings that end with a double quote, as in `ParseUtils.unquote("'foo''")`? It looks like it also covers the quoted empty string `"''"`, however it fails on `"'''"` (three consecutive quotes). This can be solved by initializing `start` at -1 instead of -2.

tolbertam: Exactly, that's why that is there.  Good catch, I do see it throws an out of bounds exception and that changing start to `-1` fixes it.  I'll fix and add a test, thanks!

PR #729, opened 2016-08-12
Update testng to 6.9.10

tolbertam:
For some reason, prior to testng 6.9.7 the driver-core test suite would
take upwards of 20 seconds to create reports for.  This is fixed by
testng 6.9.7.

Found via profiling that during this time the reporter is opening and closing upwards of 12k files per second:

screen shot 2016-08-12 at 3 33 07 pm

Looking at file probes, it appeared to be opening and closing the same files over and over again in the target/surefire-reports/old directory.


tolbertam:
testng 6.9.x doesn't support JDK 6.  Closing and will find an alternative solution.

PR #731, opened 2016-08-22
Enforce source formatting

tolbertam:
Based on discussion, here is a pull request that enforces source formatting.   This uses [formatter-maven-plugin](https://github.com/revelc/formatter-maven-plugin) which uses eclipse to validate and update source formatting.

I attempted to find a happy medium between eclipse defaults and the formatting standards we were currently using.  The end result is that the changes weren't too dramatic compared to our current standards.

This format is the same as the Eclipse standards with the following exceptions:
- Spaces instead of Tabs with 4 space indentation
- No joining of wrapped lines (for leniency)
- Disable auto-line wrapping (for leniency)
- Disable formatting of comments (for leniency)
- No spaces inserted before and after braces in array initializer
- Indent statements within switch body
- Add new lines after each enum constant

The last two could be removed, it would result in changes of a lot of our code, but would get us closer to the Eclipse conventions (not sure how important that is).

The remaining naunces between the existing format and this new one:
- For some reason parentheses are aligned with arguments if they are on a line on their own.  i.e. see `CodecRegistry`.   I can change this so parenthesis are joined with the last argument, which is how a lot of our code already is.
- Spacing alignment on constructor/method definition after 1st new line is now indented 8 characters instead of aligned with the 1st parentheses. Not sure (yet) if there is a way to align that with eclipse formatting.  i.e. `AbstractTableMetadata` constructor.
- Enum constants are now declared one line a piece.  Some of our code did this already (i.e. DataType) while others didn't.

Some other notes:
- Using version 1.4.0 of the plugin instead of 1.8.0 (the latest).  This is for JDK 6 support.
- When using `mvn formatter:format` it has to be run in each module individually since it doesn't recurse sub modules like validate does for some reason.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,8 +24,35 @@ in-depth knowledge of the codebase.
 
 ## Editor configuration
 
-We use IntelliJ IDEA with the default formatting options, with one exception: check
-"Enable formatter markers in comments" in Preferences > Editor > Code Style.
+We use Eclipse formatting conventions with a few exceptions:
+
+* Spaces instead of Tabs with 4 space indentation
+* No joining of wrapped lines (for leniency)
+* Disable auto-line wrapping (for leniency)
+* Disable formatting of comments (for leniency)
+* No spaces inserted before and after braces in array initializer
+* Indent statements within switch body
+* Add new lines after each enum constant
+
+A formatting configuration is provided in [src/formatter/java.xml](src/formatter/java.xml)
+that can be importted into both Eclipse and IntelliJ.
+
+Additionally, you may opt to use the formatting-maven-plugin to do the formatting for you
+by executing:
+
+```
+mvn formatter:format
+```
+
+In the module that you wish to do formatting for.  This is required because the formatting
+plugin doesn''t currently recurse child modules.
+
+Source format validation is done as part of the maven build and may be validated
+individually by executing:
+
+```
+mvn formatter:validate
+```
 
 Please format your code and organize imports before submitting your changes.
 
--- a/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
@@ -15,7 +15,6 @@
  */
 package com.datastax.driver.core;
 
-
 import com.google.common.base.Predicate;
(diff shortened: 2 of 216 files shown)


adutra:
I'm afraid to say that I'm not a big fan of automatic formatting tools. The main reason is that it is always possible to commit something ill-formatted if one forgets to run the plugin before committing. Another reason is that this plugin is Eclipse-based and as such, does not honor IntelliJ's formatting directives such as `@formatter:off` and `@formatter:on` (although it is certainly possible that the plugin provides equivalent functionality with different directives).

olim7t:
> I'm afraid to say that I'm not a big fan of automatic formatting tools. The main reason is that it is always possible to commit something ill-formatted if one forgets to run the plugin before committing.

One of the main goals of the PR is to address this, by adding a validation step to the build.

> Another reason is that this plugin is Eclipse-based and as such, does not honor IntelliJ's formatting directives such as `@formatter:off` and `@formatter:on` (although it is certainly possible that the plugin provides equivalent functionality with different directives).

It does, with the same (configurable) directives: https://github.com/datastax/java-driver/blob/2238d92b6eb3f07ab896a8bfdf12a4ccea1f011e/src/formatter/java.xml#L47

I'm not a big fan of using the Eclipse formatter either, but it would be really nice to have something that's not IDE-dependent and can be enforced on external PRs.

tolbertam:
I think I would be more comfortable with this if if I could get it working more reliably, i'll keep looking into it.   My current issues with it is that even when using the Eclipse Formatting Plugin for IntelliJ, the maven formatter goal behaves slightly differently than the IDE in some cases.

I think the version of eclipse used as the basis for formatting matters unfortunately.  It is really frustrating that formatting rules are sometimes not consistent between Eclipse versions!  You can have the rules set up correctly in the IDE and it is still wrong :(.   Since we maintain support for building with java6, that leaves us with an older version of the plugin that uses an older version of Eclipse (although we have discussed requiring JDK 8 for building the project possibly).  I think we can get around it by using very flexible rules (which i've done), but then this loses some of its utility.

Would be nice if someone made a `gofmt` for the java community 🙏 .

I'll keep looking around to see if there are projects out there that we can draw inspiration from, as it would be really nice to enforce a specific style.

tolbertam:
Closing this for now since it's out of sync, can always reopen if we decide we want to do this.

PR #732, opened 2016-08-26
Astyanax docs

newkek:


upgrade_guide/migrating_from_astyanax/configuration/README.md
```diff
@@ -0,0 +1,243 @@
+# Configuration
+
+## How Configuring the Java driver works
+
+The two basic components in the Java driver are the `Cluster` and the `Session`.
+The `Cluster` is the object to create first and on which to will apply all
```
adutra: This needs to be rephrased: "on which to will apply"

upgrade_guide/migrating_from_astyanax/configuration/README.md
```diff
@@ -0,0 +1,243 @@
...
+```java
+Cluster cluster = Cluster.builder()
+       .addContactPoint("127.0.0.1")
+       .withPort(9042)
+```
+
+The *Java driver* by default connects to port *9042*, hence you can supply only
+host names with the `addContactPoints(String...)` method. Note that the contact
+points are only the entry points to the `Cluster` for the *Automatic discovery
+phase*, the driver will connect to all hosts of the cluster, in the DC of the
```
adutra: That depends on the LBP, it's not necessarily true for all setups.

upgrade_guide/migrating_from_astyanax/language_level_changes/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+
+                      -----------------------------------------------------------------------
+
+
+                                                     CQL
+                                                     
+             /-                              
+             |                               
+             |  /--------------------+---------------------------------+-----------------------------\
```
adutra, line 43: This is not the way I would intuitively represent a CQL row. It would be more like this:

<table>
<tr><th>key</th><th>column: 1</th><th>column: 2</th><th>column: 3</th></tr>
<tr><td>cRED   1</td><td>'a'</td><td>'b'</td><td>'c'</td></tr>
<tr><td>cRED   2</td><td>'a'</td><td>'b'</td><td>null</td></tr>
</table>

upgrade_guide/migrating_from_astyanax/language_level_changes/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+Some of the columns of a *CQL Table* have a special role that is specifically
+related to the *Cassandra* architecture. Indeed, the *Row key* of the *Thrift Row*,
+becomes the *Partition Key* in the *CQL Table*, and can be composed of 1 or multiple
+*CQL columns* (the key column in Figure 1). The *“Column”* part of the Column-value
+component in a *Thrift Row*, becomes the *Clustering ColumnKey* in *CQL*, and can
+also be composed of multiple columns (in the figure, column1 is the only column 
+composing the *Clustering ColumnKey*).
+
+Here is the basic architectural concept of *CQL*, this post will not dive more
```
adutra: This should be rephrased as it's not a blog post anymore.

upgrade_guide/migrating_from_astyanax/language_level_changes/README.md
newkek: Agreed, pushed a revised version thanks. However, it is strange that cqlsh does not represent things this way, but the way I had done it before...

newkek:
Rebased on top of current 3.x and squashed.

upgrade_guide/migrating_from_astyanax/language_level_changes/README.md
```diff
@@ -0,0 +1,66 @@
...
+             \-
+```
+
+Some of the columns of a *CQL Table* have a special role that is specifically
+related to the *Cassandra* architecture. Indeed, the *Row key* of the *Thrift Row*,
+becomes the *Partition Key* in the *CQL Table*, and can be composed of 1 or multiple
+*CQL columns* (the key column in Figure 1). The *“Column”* part of the Column-value
+component in a *Thrift Row*, becomes the *Clustering ColumnKey* in *CQL*, and can
+also be composed of multiple columns (in the figure, column1 is the only column 
+composing the *Clustering ColumnKey*).
```
olim7t: In the thrift example, are "col1, col2, col3" the cell names, and "a, b, c" the cell values? If so, I don't agree with the second part of the schema, cell names should translate to different values of the clustering column.
See the "Dynamic Column family" example in [this blog post](http://www.datastax.com/dev/blog/thrift-to-cql3): the thrift cell names (timestamps) become the _values_ of the "time" CQL column. The text says

> map [...] the second component [of the CQL primary key, i.e. the clustering column] (time) to the internal cell name. And the last CQL3 column (url) will be mapped to the cell value"

It gets more complicated with multiple clustering columns and multiple non-primary key rows (see the "Composite" and "Non-compact tables" examples respectively), but the _value_ of the CQL clustering column always goes into the _name_ of the internal cell.

I think the example should also illustrate a case where a single Thrift wide row translates to multiple CQL rows, it seems like that's a common hurdle for people coming from Thrift.

upgrade_guide/migrating_from_astyanax/configuration/README.md
```diff
@@ -0,0 +1,242 @@
+# Configuration
+
+## How Configuring the Java driver works
+
+The two basic components in the Java driver are the `Cluster` and the `Session`.
+The `Cluster` is the object to create first, and on to which apply all
+global configuration options. Connecting to the `Cluster` creates a
```
olim7t: suggestion: "on to which all global configuration options apply"

upgrade_guide/migrating_from_astyanax/README.md
```diff
@@ -0,0 +1,10 @@
+# Migrating from Astyanax
+
+This section is a guide for users previously using *Astyanax* and looking for
+migrating to the *DataStax Java driver*.
+
+See the child pages for more information:
+
+* [Changes at the language level](language_level_changes/)
+* [Migrating Astyanax configurations to DataStax Java driver configurations](configuration/)
+* [Querying and retrieving results comparisons.](queries_and_result/)
```
tolbertam: This should be queries_and_results I think

upgrade_guide/migrating_from_astyanax/configuration/README.md
```diff
@@ -0,0 +1,242 @@
...
+`Session`. Queries are executed through the `Session`.
+
+The `Cluster` object then is to be viewed as the equivalent of the `AstyanaxContext`
+object. ’Starting’ an `AstyanaxContext` object typically returns a `Keyspace`
```
olim7t: [typo maniac warning] Surrounding "Starting" with two closing quotes looks ugly. If you use plain quotes in the source (`'Starting'`), documentor will do the right thing and generate an opening and closing quote. This works with double quotes as well.

upgrade_guide/migrating_from_astyanax/language_level_changes/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+Here is the basic architectural concept of *CQL*, a detailed explanation and *CQL*
+examples can be found in this article : [http://www.planetcassandra.org/making-the-change-from-thrift-to-cql/].
```
tolbertam: This appears somewhat weird in the generated docs:

image

upgrade_guide/migrating_from_astyanax/configuration/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+## Connection pools
+
+Configuration of connection pools in *Astyanax* are made through the
+`ConnectionPoolConfigurationImpl`. This object gathers important configurations
+that the *Java driver* has categorized in multiple *Option* and *Policy* kinds.
+
+### Connections pools internals
+Everything concerning the internal pools of connections to the *Cassandra nodes*
+will be gathered in the Java driver in the `PoolingOptions` :
```
olim7t: Link to ../../../manual/pooling

upgrade_guide/migrating_from_astyanax/configuration/README.md
```diff
@@ -0,0 +1,242 @@
...
+PoolingOptions poolingOptions =
+       new PoolingOptions()
+           .setMaxRequestsPerConnection(HostDistance.LOCAL, 1024)
+           .setCoreConnectionsPerHost(HostDistance.LOCAL, 2)
+           .setMaxConnectionsPerHost(HostDistance.LOCAL, 3);
+```
+
+Note that the *Java driver* allows multiple simultaneous requests on one single
+connection, as it is based upon the *Native protocol*, an asynchronous binary
+protocol that can handle up to 32768 simultaneous requests.
```
olim7t: Link "native protocol" to ../../../manual/native_protocol

upgrade_guide/migrating_from_astyanax/configuration/README.md
```diff
@@ -0,0 +1,242 @@
...
+to. So to get the same behaviour than the *TokenAware* pool type of *Astyanax*, 
+users shouldn’t need to specify a load balancing policy since the default one 
+should cover it.
+
+Important : Note that since *CQL* is an abstraction of the Cassandra’s architecture, a simple 
+query needs to have the *Row key* specified explicitly on a `Statement` in order 
+to benefit from the *TokenAware* routing (the *Row key* in the *Java driver* is 
+referenced as *Routing Key*), unlike the *Astyanax* driver. 
+Some differences occur related to the different kinds of `Statements` the *Java
+driver* provides. Please see [this link](../../../manual/statements) for specific information.
```
olim7t: It would be more appropriate to link to `../../../manual/load_balancing/#token-aware-policy`, it details how to set the routing key for each statement type.

upgrade_guide/migrating_from_astyanax/configuration/README.md
```diff
@@ -0,0 +1,242 @@
...
+this case, setting the CL on the `PreparedStatement`, causes the `BoundStatements` to 
+inherit the CL from the prepared statements they were prepared from. More
+informations about how `Statement`s work in the *Java driver* are detailed
+in the [“Queries and Result” section](../queries_and_results/).
+
+
+## Authentication
+
+Authentication settings are managed by the `AuthProvider` class in the *Java driver*.
+It can be highly customizable, but also comes with default simple implementations :
```
olim7t: [typo maniac warning] remove space before column

upgrade_guide/migrating_from_astyanax/configuration/README.md
```diff
@@ -0,0 +1,242 @@
...
+Concerning connection pools, the Java driver’s default settings should allow 
+most of the users to get the best out of the driver in terms of throughput, 
+they have been thoroughly tested and tweaked to accommodate the users’ needs. 
+If one still wishes to change those, first [Monitoring the pools](../../../manual/pooling/#monitoring-and-tuning-the-pool) is
+advised, then a [deep dive in the Pools management mechanism](../../../manual/pooling/) should
+provide enough insight.
+
+A lot more options are available in the different `XxxxOption`s classes, policies are
+also highly customizable since the base drivers implementations can easily be 
+extended and implement users specific actions.
```
olim7t: suggestion: user-specific actions

upgrade_guide/migrating_from_astyanax/configuration/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+Concerning connection pools, the Java driver’s default settings should allow 
+most of the users to get the best out of the driver in terms of throughput, 
+they have been thoroughly tested and tweaked to accommodate the users’ needs. 
+If one still wishes to change those, first [Monitoring the pools](../../../manual/pooling/#monitoring-and-tuning-the-pool) is
+advised, then a [deep dive in the Pools management mechanism](../../../manual/pooling/) should
+provide enough insight.
+
+A lot more options are available in the different `XxxxOption`s classes, policies are
+also highly customizable since the base drivers implementations can easily be
```
olim7t: only one driver?

upgrade_guide/migrating_from_astyanax/queries_and_results/README.md
```diff
@@ -0,0 +1,106 @@
...
+*Java driver*:
+
+```java
+ResultSet rs = session.execute("SELECT value FROM table1 WHERE key = '1'" +
+       " AND column1 > '3'" +
+       " AND column1 < '5'");
+for (Row row : rs) {
+   String value = row.getString("value");
+}
+```
```
olim7t, line 96: 👍 love those examples

olim7t (changes requested):
My main comment is about the Thrift to CQL schema.

The others are trivial, I can address them if you want.

upgrade_guide/migrating_from_astyanax/language_level_changes/README.md
tolbertam: That was tripping me up as well, initially I was conflating how things are stored on disk with the chart here.    I think it would be more clear if the chart represented the mental models for Thrift and CQL and Thrift has 1 row per Partition with the cell names like:

`'a:col2', 'a:col3', 'b:col2', 'b:col3'`

Where the CQL chart would represent multiple rows within a partition with each Row being 1 line each within a partition, where thrift is 1 line of rows overall within the partition.

upgrade_guide/migrating_from_astyanax/configuration/README.md
```diff
@@ -0,0 +1,242 @@
...
+               .setMaxConnsPerHost(3)
+```
+
+*Java driver*:
+
+```java
+PoolingOptions poolingOptions =
+       new PoolingOptions()
+           .setMaxRequestsPerConnection(HostDistance.LOCAL, 1024)
+           .setCoreConnectionsPerHost(HostDistance.LOCAL, 2)
```
tolbertam: I think it would be better to use `setConnectionsPerHost(HostDistance.LOCAL, 2, 3)`, since I think we should encourage using that method.

upgrade_guide/migrating_from_astyanax/configuration/README.md
tolbertam: To my comment about not changing options, I think it would be good emphasize that configuring pooling with the java driver is less important because it allows multiple requests on a connection.   There shouldn't be a compelling reason to increase the number of connections in the general case except for very high throughputs (can link to ../../../manual/pooling/#tuning-protocol-v3-for-very-high-throughputs).

upgrade_guide/migrating_from_astyanax/language_level_changes/README.md
newkek: Initially the schema was like this: https://gist.github.com/newkek/4a0cbe91577886383aaa9ef89701cf03
I changed it to this version after a discussion with @adutra, I think it kind of represents more "the concept" even though it's not the exact memory representation. I don't mind changing it back as I believe it is rather subjective. Also noticing that I updated the schema without updating the explanation so actually currently both don't match.

upgrade_guide/migrating_from_astyanax/configuration/README.md
```diff
@@ -0,0 +1,242 @@
...
+ConnectionPoolConfigurationImpl cpool =
+       new ConnectionPoolConfigurationImpl("myConnectionPool")
+               .setSocketTimeout(3000)
+               .setConnectTimeout(3000)
+```
+
+*Java Driver :*
+
+```java
+SocketOptions so =
```
tolbertam, line 76: would be good to add note that timeouts should not be changed unless you are changing the timeouts in cassandra.yaml.

upgrade_guide/migrating_from_astyanax/configuration/README.md
```diff
@@ -0,0 +1,242 @@
...
+`Session`. Queries are executed through the `Session`.
+
+The `Cluster` object then is to be viewed as the equivalent of the `AstyanaxContext`
+object. ’Starting’ an `AstyanaxContext` object typically returns a `Keyspace`
+object, the `Keyspace` object is the equivalent of the *Java driver*’s `Session`.
+
+Configuring a `Cluster` works with the *Builder* pattern. The `Builder` takes all
+the configurations into account before building the `Cluster`.
+
+Following are some examples of the most important configurations that were
```
tolbertam, line 17: It would be good to add a general comment that you should depend on the default configuration unless you have a good reason.  It'd be nice to have a sentence after each section saying "You shouldn't need to configure this unless....".  I think users often change a lot of configuration options because they see it in examples, but in reality they shouldn't need to.

upgrade_guide/migrating_from_astyanax/configuration/README.md
```diff
@@ -0,0 +1,242 @@
...
+
+Since the *Java driver* only executes *CQL* statements, which can be either reads
+or writes to *Cassandra*, it is not possible to globally configure the
+Consistency Level for only reads or only writes. To do so, since the Consistency 
+Level can be set per-statement, you can either set it on every statement, or use 
+`PreparedStatements` (if queries are to be repeated with different values): in
+this case, setting the CL on the `PreparedStatement`, causes the `BoundStatements` to 
+inherit the CL from the prepared statements they were prepared from. More
+informations about how `Statement`s work in the *Java driver* are detailed
+in the [“Queries and Result” section](../queries_and_results/).
```
tolbertam: Should be "Queries and Results"

upgrade_guide/migrating_from_astyanax/language_level_changes/README.md
```diff
@@ -0,0 +1,66 @@
+# Language change : from Thrift to CQL
+The data model changes when using *CQL* (Cassandra Query Language).
+*CQL* is providing an abstraction of the low-level data stored in *Cassandra*, in
+opposition to *Thrift* that aims to expose the low-level data structure directly.
+[But note that this changes with Cassandra 3’s new storage engine.](http://www.datastax.com/2015/12/storage-engine-30)
+
+*Thrift* exposes *Keyspaces*, and these *Keyspaces* contain *Column Families*. A
+*ColumnFamily* contains *Rows* in which each *Row* has a list of an arbitrary number
+of column-values. With *CQL*, the data is **tabular**, *ColumnFamily* gets viewed
+as a *Table*, the **Table Rows** get a **fixed and finite number of named columns**.
```
tolbertam, line 10: In one case Table Rows is bolded for emphasis, and in another case it's underlined, is that intentional?

upgrade_guide/migrating_from_astyanax/queries_and_results/README.md
```diff
@@ -0,0 +1,106 @@
+# Queries and Results
+There are many ressources such as [this post][planetCCqlLink] or [this post][dsBlogCqlLink] to learn
```
tolbertam: ressources -> resources

upgrade_guide/migrating_from_astyanax/queries_and_results/README.md
```diff
@@ -0,0 +1,106 @@
...
+how to transform previous Thrift operations to CQL queries.
+ 
+The *Java driver* executes CQL queries through the `Session`. 
+The queries can either be simple *CQL* Strings or represented in the form of 
+`Statement`s. The driver offers 4 kinds of statements, `SimpleStatement`, 
+`Prepared/BoundStatement`, `BuiltStatement`, `BatchStatement`. All necessary
```
tolbertam: , and BatchStatement

upgrade_guide/migrating_from_astyanax/queries_and_results/README.md
```diff
@@ -0,0 +1,106 @@
+# Queries and Results
+There are many ressources such as [this post][planetCCqlLink] or [this post][dsBlogCqlLink] to learn
+how to transform previous Thrift operations to CQL queries.
+ 
+The *Java driver* executes CQL queries through the `Session`. 
+The queries can either be simple *CQL* Strings or represented in the form of 
+`Statement`s. The driver offers 4 kinds of statements, `SimpleStatement`, 
+`Prepared/BoundStatement`, `BuiltStatement`, `BatchStatement`. All necessary 
+information can be [found here](../../../manual/statements/) about the natures of the different
```
tolbertam: natures -> nature

upgrade_guide/migrating_from_astyanax/queries_and_results/README.md
```diff
@@ -0,0 +1,106 @@
...
+`Statement`s.
+
+As explained in [this documentation section](../../../manual/#running-queries),
```
tolbertam: suggest renaming 'this documentation section' 'the 'running queries' section'

tolbertam (changes requested):
Looks great!  Had a few minor suggestions.

upgrade_guide/migrating_from_astyanax/language_level_changes/README.md
tolbertam: I like the current version compared to the original diagram.  The two suggestions I would have:
- The column/cell names in the thrift diagram should be clusteringkeyvalue:columnname (i.e. a:col2, a:col3 and so on).  The clustering key shouldn't be it's own column.
- I think it would be good to show what it would look like though if you had multiple rows per partition.  The current diagram only has 1 cql row/unique cluster per partition, it would be good to demonstrate what it would look like if there were at least 2.  (i.e. have two rows for key=1 where key=1,col1=a and key=1,col1=b).  Hopefully that makes sense.

tolbertam: What is really confusing to me is that with the pre-3.0 storage engine, whether or not you use CQL or Thrift,   the way data is stored is a 1:1 with how I visualize Thrift.   Where with the 3.0 storage engine, it is nearly 1:1 with how I visualize CQL.  That's why I think it is very easy for the reader to conflate the data model with how things are stored.

tolbertam: ^ That isn't so much a comment on your documentation, just an inner monologue of how things get mixed up in my mind :)

upgrade_guide/migrating_from_astyanax/configuration/README.md
newkek: Hmm not 100% agreeing with that, as per previous discussions we have mentioned that some clients may want to "give up" on a request earlier than others, and so in this case it does not impact the C\* yaml... I think it's better to keep it vague here, and I've clearly mentioned at the beginning that "if you change this options it means you know what you're doing", I might repeat it here

upgrade_guide/migrating_from_astyanax/language_level_changes/README.md
newkek: Yep, wanted to emphasize the **Table** _Rows_ difference...

upgrade_guide/migrating_from_astyanax/language_level_changes/README.md
newkek: > The column/cell names in the thrift diagram should be clusteringkeyvalue:columnname (i.e. a:col2, a:col3 and so on).

I think this is too low level in the storage engine and too far from actual CQL queries and would confuse people more than anything... When creating a Columnfamily with Astyanax or any Thrift client, what you end up with in the CQL representation is what is described in this schema https://gist.github.com/newkek/4a0cbe91577886383aaa9ef89701cf03 at least that's how CQLSH would represent it, in a CQL query you would query a Column Family created with a Thrift client as `SELECT key, column1, value FROM CF1` at least for a simple example, and I think for the sake of understandability we should stick to that and let the other links describe specificities better.

newkek:
Looks like I've corrected all comments, except the schema. Seems like there are 3 different suggestions there... Please note that I did not intend to go into much details in the CQL level changes and stay "basic" and give a "general idea" as I have linked to other posts that can explain those CQL changes much better and more detailed than I can do and it is not the intention of this doc to focus on CQL. I'd rather go back to the schema in https://gist.github.com/newkek/4a0cbe91577886383aaa9ef89701cf03 because it's closer to what CQLSH shows and what CQL queries look like.

upgrade_guide/migrating_from_astyanax/configuration/README.md
tolbertam: In the past we have advised users to not decrease `SocketOptions#readTimeoutMillis` and to depend on the cassandra side timeouts instead and in particular not to set readTimeoutMillis to a value less than the cassandra timeouts.   If the client timeout is less than the cassandra timeout, you don't give RetryPolicy an opportunity to kick in.

upgrade_guide/migrating_from_astyanax/language_level_changes/README.md
tolbertam: Putting that aside, what do you think about extending the diagram to have multiple rows within a partition?   I think that would better demonstrate that a Thrift Row has all columns for a given rowkey/partition key, where a CQL Row contains the columns for a unique partition&clustering key.

olim7t:
Yes I prefer your other schema too, because it illustrates the case where a single Thrift row translates to multiple CQL rows, which was the main concern I think.

upgrade_guide/migrating_from_astyanax/language_level_changes/README.md
newkek: > I think that would better demonstrate that a Thrift Row has all columns for a given rowkey/partition key, where a CQL Row contains the columns for a unique partition&clustering key.

Sounds like that is what is described by https://gist.github.com/newkek/4a0cbe91577886383aaa9ef89701cf03 ?

tolbertam: Maybe i'm interpreting that diagram wrong, but isn't there only 1 clustering value for the thrift diagram for each row (if Col : 1 is the clustering value)?

olim7t (approved):

olim7t:
LGTM pending the schema. I saw a couple minor issues but I addressed them directly.

upgrade_guide/migrating_from_astyanax/language_level_changes/README.md
newkek: Hm I see what you mean, was it possible to have two "Col:1" for the same Row Key in Thrift? I don't see how I can do that, with Astyanax at least, I tried inserting twice for the Same Row Key and The Same "Col:x" 2 different values, and one erases the other, since the "Col:" is part of the Clustering key

tolbertam: Sorry, it's easy to conflate terms so i'm probably confusing things a bit.

Taking a step back to describe what I'm trying to say with an example data model.   I was thinking of the table that you were sharing having a composite comparator that was `CompositeType(Int32, UTF8Type)`  where `Int32` is the value of `Col1`, and `UTF8Type` is the name of the column.

As an example, say you want to represent readings from a thermostat.  The thermostat is identified by an integer id which will be our row key. A thermostat has 'readings' identified by a reading id, and each reading has a `temp` (for the current temperature), a `mode` (cooling, heating, off) and a setpoint (for target temperature).

Using cassandra-cli, i could create a schema like so:

```
create keyspace ks;
use ks;
create column family readings
with key_validation_class = Int32Type
and comparator = 'CompositeType(Int32Type, UTF8Type)'
and default_validation_class = UTF8Type;
```

I could get even fancier and provide column metadata, but leaving this as is to keep it simple.

I could then create multiple readings for an individual thermostat, i.e.:

```
set readings[0]['0:temp'] = '72';
set readings[0]['0:mode'] = 'heating';
set readings[0]['0:setpoint'] = '73';
set readings[0]['1:temp'] = '75';
set readings[0]['1:mode'] = 'cooling';
set readings[0]['1:setpoint'] = '70';
set readings[1]['5:temp'] = '65';
set readings[1]['5:mode'] = 'off';
set readings[1]['5:setpoint'] = '70';
```

When reading back the data, I get:

```
[default@ks] get readings[0];
=> (name=0:mode, value=heating, timestamp=1474483110889000)
=> (name=0:setpoint, value=73, timestamp=1474483119088000)
=> (name=0:temp, value=72, timestamp=1474483101433000)
=> (name=1:mode, value=cooling, timestamp=1474483137460000)
=> (name=1:setpoint, value=70, timestamp=1474483147380000)
=> (name=1:temp, value=75, timestamp=1474483127437000)
[default@ks] get readings[1];
=> (name=5:mode, value=off, timestamp=1474483198322000)
=> (name=5:setpoint, value=70, timestamp=1474483205968000)
=> (name=5:temp, value=65, timestamp=1474483192274000)
```

In that example, for a Thrift row for a given key,  I have multiple readings.  In CQL, The concept of a row layout would depend on the clustering, but I would preferably model this like:

```
create table readings_cql (k int, reading_id int, mode text, setpoint float, temp float, primary key (k, reading_id));

-- insert data
insert into readings_cql (k, reading_id, mode, setpoint, temp) values (0, 0, 'heating', 73, 72);
insert into readings_cql (k, reading_id, mode, setpoint, temp) values (0, 1, 'cooling', 70, 75);
insert into readings_cql (k, reading_id, mode, setpoint, temp) values (1, 5, 'off', 70, 65);
```

This would give me back:

```
select * from readings_cql;

 k | reading_id | mode    | setpoint | temp
---+------------+---------+----------+------
 1 |          5 |     off |       70 |   65
 0 |          0 | heating |       73 |   72
 0 |          1 | cooling |       70 |   75
```

This looks more like your second (current) diagram.

Where as the thrift table represented as a CQL table looks more like your first (https://gist.github.com/newkek/4a0cbe91577886383aaa9ef89701cf03)

```
cqlsh> select * from ks.readings;

 key | column1 | column2  | value
-----+---------+----------+---------
   1 |       5 |     mode |     off
   1 |       5 | setpoint |      70
   1 |       5 |     temp |      65
   0 |       0 |     mode | heating
   0 |       0 | setpoint |      73
   0 |       0 |     temp |      72
   0 |       1 |     mode | cooling
   0 |       1 | setpoint |      70
   0 |       1 |     temp |      75
```

So what the `One CQL Table` will look like will ultimately depend on the CQL schema.  Ultimately the stored data is similar, it's just how CQL determines what is a 'Row' depends on the clustering.

Comparing the data for readings (thrift) and readings_cql, the layout on disk is virtually the same (other than the values because my value type is different, and the ghost cell for CQL):

```
Thrift table:
[default@ks] get readings[0];
=> (name=0:mode, value=heating, timestamp=1474483110889000)
=> (name=0:setpoint, value=73, timestamp=1474483119088000)
=> (name=0:temp, value=72, timestamp=1474483101433000)
=> (name=1:mode, value=cooling, timestamp=1474483137460000)
=> (name=1:setpoint, value=70, timestamp=1474483147380000)
=> (name=1:temp, value=75, timestamp=1474483127437000)

CQL table:
[default@ks] get readings_cql[0];
=> (name=0:, value=, timestamp=1474483980386000)
=> (name=0:mode, value=68656174696e67, timestamp=1474483980386000)
=> (name=0:setpoint, value=42920000, timestamp=1474483980386000)
=> (name=0:temp, value=42900000, timestamp=1474483980386000)
=> (name=1:, value=, timestamp=1474483996207000)
=> (name=1:mode, value=636f6f6c696e67, timestamp=1474483996207000)
=> (name=1:setpoint, value=428c0000, timestamp=1474483996207000)
=> (name=1:temp, value=42960000, timestamp=1474483996207000)
```

With all that defined, what was confusing to me is the Thrift Diagram (`One Thrift COLUMNFAMILY`) which is the same for both diagrams.

In your diagram you have three colors:
1. red: Thrift: Row Key, CQL: Partition Key
2. blue: Thrift: Column Comparator, CQL: Clustering Column Value
3. orange: Thrift: Column Value, CQL: Non-Clustering Column Value

If I'm interpreting that correctly and 'Col1' is a clustering column, I wouldn't expect it to be represented as its own column for Thrift.  Instead if would be part of the column name in Thrift diagram where it would not for the CQL diagram.  

I know that is maybe i'm getting too semantical and thinking too much about the storage model, but I think it is hard to divorce thrift from the storage model, as to me they are the same.  [One of the older thrift -> CQL guides](http://www.datastax.com/dev/blog/thrift-to-cql3) uses the storage model to describe CQL and how different column family schemas would be modeled with CQL:

> CQL3 (the Cassandra Query Language) provides a new API to work with Cassandra. Where the legacy thrift API exposes the internal storage structure of Cassandra pretty much directly, CQL3 provides a thin abstraction layer over this internal structure.

Hopefully that makes sense, I had to revisit thrift to level set my thoughts.

tolbertam: Here's an example for inserting / retrieving data with that schema: https://gist.github.com/tolbertam/b12135fb4223896f14aa7c80426ae9a9

I'll admit that Composites are complicated, maybe a better example is a dynamic column family like the `clicks` example in [this guide](http://www.datastax.com/dev/blog/thrift-to-cql3) which maps to a cql table with 1 partition key, 1 clustering column, and 1 non-clustering column.

newkek: Thanks for the investigation. Yeah you're right Composite Column types are probably going to get more complicated however my first schema is not wrong it's just basic since it exposes the storage model in thrift but for a more basic use-case... As I said I would rather let more experienced people take care of explaining the more complicated aspects of the data model change for CQL by linking to other blog posts that were already written, rather than trying to explain all myself again in this docs...
Btw thanks for the code sample I had tried the composite columns in Astyanax a litle while ago but I couldn't get the CompositeColumn Serializer to work so that's cool

tolbertam (approved):

newkek:
Ok so after private discussions with @tolbertam I pushed a revised version of the first schema in https://gist.github.com/newkek/4a0cbe91577886383aaa9ef89701cf03 that should be slightly clearer. Seems to me we're all done for that PR.

tolbertam:
The changes look great! 👍

olim7t:
Squashed and rebased on top of 3.0.x.

PR #733, opened 2016-08-31
fix README of assign SimpleStatement routing key

kaicao:
When following document to assign routing key to SimpleStatement found issue DateType.cint() doesn't has serialize method, instead should use TypeCodec.cint


olim7t:
Thanks!
I've cherry-picked your commit on the 3.0.x branch, so that it's in the upcoming 3.0.4 release. I'm closing this but your commit is here with proper attribution: 9eff44d

kaicao:
Thank you

PR #734, opened 2016-08-31
JAVA-1275: Use Netty's default thread factory

olim7t:
This makes all of our threads instances of Netty's
FastThreadLocalThread, which allows an optimization of ThreadLocal
variables.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -25,6 +25,7 @@ import com.google.common.base.Strings;
 import com.google.common.base.Throwables;
 import com.google.common.collect.*;
 import com.google.common.util.concurrent.*;
+import io.netty.util.concurrent.DefaultThreadFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1515,7 +1516,12 @@ ProtocolVersion protocolVersion() {
         }
 
         ThreadFactory threadFactory(String name) {
-            return new ThreadFactoryBuilder().setNameFormat(clusterName + "-" + name + "-%d").build();
+            return new ThreadFactoryBuilder()
```
tolbertam, line 1519: Couldn't we use:

``` java
return new DefaultThreadFactory(clusterName + "-" + name);
```

Although I noticed that this gives a slightly different name, i.e.: `cluster1-nio-worker-7-1` instead of `cluster1-nio-worker-1` since it seems to assign a unique pool id to each thread factory, so maybe the way it is done here is best for compatibility with the existing naming (also that unique id seems extraneous since we're sort of doing that ourselves with the cluster name).

driver-core/src/main/java/com/datastax/driver/core/NettyOptions.java
```diff
@@ -89,7 +89,9 @@
      *
      * @param threadFactory The {@link ThreadFactory} to use when creating a new {@code EventLoopGroup} instance;
      *                      The driver will provide its own internal thread factory here.
-     *                      It is safe to ignore it and use another thread factory.
+     *                      It is safe to ignore it and use another thread factory. Note however that for optimal
+     *                      performance it is recommended to use a factory that returns
+     *                      {@link io.netty.util.concurrent.FastThreadLocalThread} instances.
```
tolbertam: It would be helpful to indicate `io.netty.util.concurrent.DefaultThreadFactory` as an example, so users know how to create a thread factory this way.  Although `NettyOptions` is for advanced users, so they might already know this.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
olim7t: Yes, keeping the exact same naming was my motivation here.

PR #735, opened 2016-08-31
JAVA-1276: FAQ Entry for tombstone messages

tolbertam:
Adds an FAQ entry for how tombstones could possibly be created using the
driver and how to avoid doing so.

Figured this would be good to include for 3.0.4 so opened up against 3.0.x, but we can defer to later.

### DIFF
--- a/faq/README.md
+++ b/faq/README.md
@@ -157,4 +157,59 @@ and we've had many reports where the problem turned out to be in user code.
 
 See [Blobs.java] in the `driver-examples` module for some examples and explanations.
 
+
+### Why am I seeing messages about `tombstone_warn_threshold` or `tombstone_fail_threshold` being exceeded in my Cassandra logs?
+
+Applications which use the object mapper or set `null` values in their 
+statements may observe that many tombstones are being stored in their tables
+which subsequently may lead to poor query performance, failed queries, or
+columns being mysteriously deleted.
+
+This is caused by INSERT/UPDATE statements containing `null` values for columns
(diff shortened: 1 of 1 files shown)


faq/README.md
```diff
@@ -157,4 +157,59 @@ and we've had many reports where the problem turned out to be in user code.
...
+columns being mysteriously deleted.
+
+This is caused by INSERT/UPDATE statements containing `null` values for columns
+that a user does not intend to change.  Common circumstances around this come
+from using the object mapper or writing your own persistence layer and
+attempting to reuse the same `PreparedStatement` for inserting data, even with
+partial updates.
+
+Prior to cassandra 2.2, there was no means of reusing the same
+`PreparedStatement` for making partial updates to difference columns.
```
adutra: small nit : `different columns`

faq/README.md
```diff
@@ -157,4 +157,59 @@ and we've had many reports where the problem turned out to be in user code.
 
 See [Blobs.java] in the `driver-examples` module for some examples and explanations.
 
-[Blobs.java]: https://github.com/datastax/java-driver/tree/3.0.x/driver-examples/src/main/java/com/datastax/driver/examples/datatypes/Blobs.java
\ No newline at end of file
+
+### Why am I seeing messages about tombstone_warn_threshold or tombstone_fail_threshold being exceeded?
```
adutra: Maybe you should specify that such error messages do not appear in the driver logs, but in Cassandra logs instead.

PR #737, opened 2016-09-06
JAVA-1279: Mapper should exclude Groovy's "metaClass" property when looking for mapped properties

adutra:


driver-mapping/src/main/java/com/datastax/driver/mapping/ReflectionUtils.java
```diff
@@ -25,12 +26,19 @@
...
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 
 /**
  * Utility methods related to reflection.
  */
 class ReflectionUtils {
 
+    private static final Set<String> EXCLUDED_PROPERTIES = ImmutableSet.of(
```
beckje01, line 36: Would it be worth making this something users could configure?

adutra: I thought about that too, but I'm hesitating – that could be a bit over-engineered because we would have to introduce some configuration class, say `MappingConfiguration`, and pass it to `MappingManager`, e.g.:

``` java
public MappingManager(Session session, ProtocolVersion protocolVersion, MappingConfiguration mappingConfiguration) {
....
}
```

Also I'm hesitating on the easiest way to expose that configuration in particular. Of course we could simply do:

``` java
public class MappingConfiguration {
    public Set<String> transientProperties() {
        return ...
    }
}
```

But if users need to specify transient properties this way, that's probably because their models inherit from parent classes that they do not have control on; so that list could grow bigger and bigger.

Maybe a better solution would be to turn the opt-out approach for transient properties into an opt-in one, i.e., _only_ properties annotated with `@Column` (or `@Field`, etc.) would be mapped, and all the others would be ignored. That could be achieved with a configuration like below:

``` java
public enum TransiencyStrategy {

    /**
     * Default. Maps all properties except those annotated with {@link com.datastax.driver.mapping.annotations.Transient}.
     */
    ALL,

    /**
     * Maps only properties explicitly annotated with {@link com.datastax.driver.mapping.annotations.Column},
     * {@link com.datastax.driver.mapping.annotations.Field}, {@link com.datastax.driver.mapping.annotations.PartitionKey}
     * or {@link com.datastax.driver.mapping.annotations.ClusteringColumn}.
     */
    ANNOTATED
}

public class MappingConfiguration {
    public TransiencyStrategy transiencyStrategy() {
        return ANNOTATED;
    }
}
```

@al3xandru @olim7t @newkek @tolbertam can we have your input on this topic?

newkek: I would actually lean towards the pre-3.1.0 behaviour and a opt-in approach for having the mapper scanning all fields in superclasses like the Groovy's metaclass and so on, via indeed something like `MappingConfiguration` if that's possible. Because if we start making a special case for the Groovy metaclass we probably will end up having to do other special cases over and over later.

driver-mapping/pom.xml
```diff
@@ -108,11 +109,37 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.codehaus.groovy</groupId>
+            <artifactId>groovy-all</artifactId>
+            <version>${groovy.version}</version>
+            <scope>test</scope>
+        </dependency>
```
olim7t, line 117: Is there any way we can test this without depending on the groovy runtime?

driver-mapping/src/main/java/com/datastax/driver/mapping/ReflectionUtils.java
olim7t: What about a system property (to keep things simple)?

driver-mapping/pom.xml
tolbertam: Keeping [JAVA-1296](https://datastax-oss.atlassian.net/browse/JAVA-1296) in mind as well, i'm also wondering about that.   Testing all sorts of different language runtimes could get kind of overbearing after a while.  wdyt about a separate module for these kinds of one off tests?

adutra: This is a test dependency. I wouldn't care so much about it, but I could live with a separate module if this is a concern. Do you think it could pollute our test classpath is some dangerous way?

olim7t: No, it was more to avoid downloading the internet when you first run tests. We just want to ignore a `metaClass` property, we could test it without Groovy. But I see the value in a full-fledged Groovy integration test as well. We can keep that as-is for now, and we'll reconsider the tests structure for 1296.

driver-mapping/src/main/java/com/datastax/driver/mapping/ReflectionUtils.java
olim7t: One limitation with a system property is that you only control the list globally. Someone might need a per-mapper configuration.
Yet another way would be a property on the `Table` and `UDT` annotations.

olim7t:
I've pushed a commit with the exclusions as a class-level annotation property, it looks like a good solution.

adutra (commented):
Class-level transients are convenient, but I am concerned that if users want to exclude some inherited property, they will probably have to amend every concrete class in their data model, which is error-prone (i.e. if someone adds a new table or UDT to the model, he must not forget to include the appropriate class-level transients). Therefore I think if we could provide users with a way to exclude such properties only once in their code, that would be better.

olim7t:
We don't want to delay 3.1.1 any more, so I'm merging this with the hard-coded list.
I've opened [JAVA-1310](https://datastax-oss.atlassian.net/browse/JAVA-1310) to explore the configuration aspect, and I'll create a branch for it and cherry-pick my commit on it.

olim7t (approved):

PR #738, opened 2016-09-07
JAVA-1272: BuiltStatement not able to print its query string if it contains mapped UDTs.

adutra:


driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
```diff
@@ -296,18 +296,19 @@ public Boolean isIdempotent() {
     @Override
     public String toString() {
         try {
+            if (forceNoValues)
+                return getQueryString();
+            // 1) try first with all values inlined (will not work if some values require custom codecs,
+            // or if the required codecs are registered in a different CodecRegistry instance than the default one)
+            return maybeAddSemicolon(buildQueryString(null, CodecRegistry.DEFAULT_INSTANCE)).toString();
+        } catch (RuntimeException e1) {
```
adutra, line 300: Catching `RuntimeException` prevents other unchecked exceptions to propagate, and in particular, exceptions thrown from codecs (NPE, `InvalidTypeException`, `UnsupportedOperationException`...)

adutra:
Note to the unfortunate developer that will merge this to 3.x and 3.1.x: the implementation of `parseAndSetField()`and `formatField()` in `MappedUDTCodec` will have to be completely rewritten (thanks to JAVA-541) as below:

``` java
@Override
protected String formatField(T source, String fieldName) {
    PropertyMapper propertyMapper = columnMappers.get(fieldName);
    if (propertyMapper == null)
        return null;
    Object value = propertyMapper.getValue(source);
    TypeCodec<Object> codec = propertyMapper.customCodec;
    if (codec == null)
        codec = codecRegistry.codecFor(cqlUserType.getFieldType(propertyMapper.columnName), propertyMapper.javaType);
    return codec.format(value);
}

@Override
protected T parseAndSetField(String input, T target, String fieldName) {
    PropertyMapper propertyMapper = columnMappers.get(fieldName);
    if (propertyMapper != null) {
        TypeCodec<Object> codec = propertyMapper.customCodec;
        if (codec == null)
            codec = codecRegistry.codecFor(cqlUserType.getFieldType(propertyMapper.columnName), propertyMapper.javaType);
        propertyMapper.setValue(target, codec.parse(input));
    }
    return target;
}
```

driver-core/src/main/java/com/datastax/driver/core/TypeCodec.java
```diff
@@ -506,6 +506,8 @@ public DataType getCqlType() {
      * Implementing this method is not strictly mandatory: internally, the driver only uses it to
      * parse the INITCOND when building the metadata of an aggregate function (and in most cases it
      * will use a built-in codec, unless the INITCOND has a custom type).
+     * If you choose not to implement this method, you should not throw an exception but
+     * instead return a constant string (for example "XxxCodec.parse not implemented").
```
olim7t: Is this misplaced? The method returns a `T`, not a string.

driver-core/src/main/java/com/datastax/driver/core/TypeCodec.java
```diff
@@ -533,8 +535,8 @@ public DataType getCqlType() {
      * {@link TupleValue}, and the internal representation of a {@code ROWS} response),
      * which may appear in driver logs.</li>
      * </ol>
-     * If choose not to implement this method, you can return a constant string (for example
-     * "XxxCodec.format not implemented").
+     * If you choose not to implement this method, you should not throw an exception but
+     * instead return a constant string (for example "XxxCodec.format not implemented").
```
olim7t, line 537: Why? Don't we catch the exception in `BuiltStatement.toString` now?

adutra: Because we still do not catch it in the `QueryLogger`, for instance. Any exception raised in a `LatencyTracker` bubbles up to [here](https://github.com/datastax/java-driver/blob/3.0.x/driver-core/src/main/java/com/datastax/driver/core/Cluster.java#L1575) before being swallowed.
On a second thought, I think it would be even better to remove this paragraph altogether and protect our code against unchecked exceptions thrown by a codec, like we did in `BuiltStatement.toString()`.

driver-core/src/main/java/com/datastax/driver/core/TypeCodec.java
adutra: Sorry, stubborn copy/paste, I will amend that.

driver-core/src/test/java/com/datastax/driver/core/querybuilder/QueryBuilderTest.java
```diff
@@ -999,4 +994,63 @@ public void should_include_original_cause_when_arguments_invalid() {
...
+     * the query string should contain all variables inlined and formatted properly.
+     *
+     * @jira_ticket JAVA-1272
+     */
+    @Test(groups = "unit")
+    public void should_inline_custom_codec() throws Exception {
+        assertThat(
+                insertInto("users").value("id", new Foo(42)).toString())
+                .isEqualTo("INSERT INTO users (id) VALUES (?);");
+        CodecRegistry.DEFAULT_INSTANCE.register(new FooCodec());
```
tolbertam, line 1050: Since `DEFAULT_INSTANCE` is static, this will impact all other tests, but since Foo is only used here I don't see it being a problem.

driver-mapping/src/main/java/com/datastax/driver/mapping/MappedUDTCodec.java
```diff
@@ -87,13 +87,35 @@ protected T deserializeAndSetField(ByteBuffer input, T target, String fieldName,
 
     @Override
     protected String formatField(T source, String fieldName) {
-        // This codec implementation is internal-use only, and its format method is never used
-        throw new UnsupportedOperationException();
+        if (!fieldName.startsWith("\""))
```
tolbertam, line 90: I was initially very confused that we were quoting fields, but then I noticed that column names are quoted because `ColumnMapper.getColumnName` quotes non computed Columns, in case of UDT, these are never computed.

driver-mapping/src/main/java/com/datastax/driver/mapping/MappedUDTCodec.java
```diff
@@ -87,13 +87,35 @@ protected T deserializeAndSetField(ByteBuffer input, T target, String fieldName,
...
+            codec = codecRegistry.codecFor(cqlUserType.getFieldType(columnMapper.getColumnName()), columnMapper.getJavaType());
+
+        return codec.format(value);
     }
 
     @Override
     protected T parseAndSetField(String input, T target, String fieldName) {
-        // This codec implementation is internal-use only, and its parse method is never used
-        throw new UnsupportedOperationException();
+        if (!fieldName.startsWith("\""))
```
tolbertam: In this case it isn't escaped before passing it in, although maybe that should be done in `AbstractUDTCodec#parse`

driver-mapping/src/main/java/com/datastax/driver/mapping/MappedUDTCodec.java
tolbertam: Hrmm, actually this would always quote the fieldName, do we want that?

driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
```diff
@@ -293,14 +292,19 @@ public Boolean isIdempotent() {
...
+            // 2) try next with bind markers for all values to avoid usage of custom codecs
+            try {
+                return maybeAddSemicolon(buildQueryString(new ArrayList<Object>(), CodecRegistry.DEFAULT_INSTANCE)).toString();
+            } catch (RuntimeException e2) {
+                // Ugly but we have absolutely no context to get the registry from
+                return String.format("built query (could not generate with default codec registry: %s)", e2.getMessage());
```
tolbertam, line 306: Is this final case even possible now if we don't try parsing the values?  I guess it could depend on the implementation of `BuiltStatement`, but none of the ones we define do from what I can tell.  Still seems like a good idea to have this here just in case.

tolbertam (approved):
lgtm 👍

driver-core/src/main/java/com/datastax/driver/core/TypeCodec.java
olim7t: Mmm after re-reading this javadoc I remember the rationale for returning a string and never throwing an exception. The real issue is that `MappedUdtCodec` was throwing an `UnsupportedOperationException`, that was an oversight.
Provided that all our internal codecs properly implement `format`, We don't even need the catch block in `BuiltStatement.toString()`, although it doesn't hurt to keep it there.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
olim7t: `buildQueryString` might still try to inline _some_ of the values even if you require bind markers. The [BuiltStatement javadoc](http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/core/querybuilder/BuiltStatement.html) lists those edge cases.

PR #740, opened 2016-09-08
JAVA-1283: Cache in Mapper is causing problem after reconnection to nodes

adutra:


olim7t:
👍

driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
```diff
@@ -128,6 +128,9 @@ public void onSuccess(PreparedStatement stmt) {
                     @Override
                     public void onFailure(Throwable t) {
                         future.setException(t);
+                        // do not keep a failed future in the query cache
+                        preparedQueries.remove(pqk, future);
+                        logger.error("Query preparation failed: " + queryString, t);
```
newkek, line 133: minor: couldn't we log before setting the exception on the future or does it not matter?

olim7t (approved):
Approved

PR #742, opened 2016-09-08
JAVA-1053: Surface all errors in RetryPolicy.

adutra:


driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -598,8 +560,19 @@ public void onSet(Connection connection, Message.Response response, long latency
                                 return;
                             default:
                                 connection.release();
-                                if (metricsEnabled())
+                                if (err.code == SERVER_ERROR)
```
adutra, line 563: You will notice that I removed the logger call, because it would be too verbose to log all kinds of errors, and because ultimately I think it's up to the retry policy to log the error if it needs to.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -653,29 +626,22 @@ public void onSet(Connection connection, Message.Response response, long latency
...
-                            logError(connection.address, new DriverException("Error preparing query, got " + response));
-                            if (metricsEnabled())
-                                metrics().getErrorMetrics().getOthers().inc();
-                            retry(false, null);
-                            break;
-                        default:
-                            // Something's wrong, so we return but we let setFinalResult propagate the exception
-                            SpeculativeExecution.this.setFinalResult(connection, response);
-                            break;
+                    if (response.type == RESULT && ((Responses.Result) response).kind == PREPARED) {
```
adutra, line 630: Here only the success case is bypassing the retry policy, all other (error) cases go through it.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -686,17 +652,7 @@ public void onException(Connection connection, Exception exception, long latency
 
                 @Override
                 public boolean onTimeout(Connection connection, long latency, int retryCount) {
-                    QueryState queryState = queryStateRef.get();
```
adutra, line 689: This was more or less identical to the main `onTimeout` implementation, so now that it goes through the retry policy I thought best to just call the main method here.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -714,21 +670,16 @@ public void onException(Connection connection, Exception exception, long latency
...
-                        metrics().getErrorMetrics().getConnectionErrors().inc();
-                        if (decision.getType() == Type.RETRY)
-                            metrics().getErrorMetrics().getRetriesOnConnectionError().inc();
-                        if (decision.getType() == Type.IGNORE)
-                            metrics().getErrorMetrics().getIgnoresOnConnectionError().inc();
-                    }
-                    processRetryDecision(decision, connection, exception);
-                    return;
+                RetryPolicy retryPolicy = retryPolicy();
+                RetryPolicy.RetryDecision decision = retryPolicy.onRequestError(statement, request().consistency(), (DriverException) exception, retriesByPolicy);
```
adutra, line 674: You will notice the cast to `DriverException`, which is harmless because in practice all exceptions are instances of `DriverException`.

driver-core/src/test/java/com/datastax/driver/core/AuthenticationTest.java
```diff
@@ -48,11 +48,11 @@ public void sleepIf12() {
     @Test(groups = "short")
     public void should_connect_with_credentials() throws InterruptedException {
         PlainTextAuthProvider authProvider = spy(new PlainTextAuthProvider("cassandra", "cassandra"));
-        Cluster cluster = Cluster.builder()
+        Cluster cluster = register(Cluster.builder()
```
adutra, line 51: Unrelated: missing call to `register()`.

driver-core/src/test/java/com/datastax/driver/core/AuthorizationTest.java
```diff
@@ -0,0 +1,77 @@
...
+    public void sleepIf12() {
+        // For C* 1.2, sleep before attempting to connect as there is a small delay between
+        // user being created.
+        if (ccm().getVersion().getMajor() < 2) {
+            Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
+        }
+    }
+
+    @Test(groups = "short", expectedExceptions = UnauthorizedException.class)
+    public void should_fail_to_access_unauthorized_resource() throws InterruptedException {
```
adutra, line 50: We did not have any authorization test so far, and since the Jira ticket mentions `UnauthorizedException` explicitly, I thought it would be a good idea to create one.

driver-core/src/test/resources/log4j.properties
```diff
@@ -20,6 +20,7 @@ log4j.rootLogger=INFO, A1
 # Adjust Scassandra's log level
 # (it seems some messages are conditioned by log4j.properties and others by reference.conf, so we need both)
 log4j.logger.org.scassandra=ERROR
+log4j.logger.spray.can.server=ERROR
```
adutra, line 23: Unrelated: reducing verbosity of this logger.

newkek:
+1, and I think we should put a mention of this change in the upgrade guide. Because people that were implementing their own RetryPolicies will have to deal with the new error types so I think we should guide that. Either by saying to extend `DefaultRetryPolicy` in order to keep things as they were before, or informing what are the new types of errors.

adutra:
Closing as the scope of this ticket has shifted.

PR #743, opened 2016-09-10
JAVA-1285 - Routing key auto-discovery is case sensitive on column names

briandilley:


driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
```diff
@@ -231,7 +231,7 @@ void maybeAddRoutingKey(String name, Object value) {
             return;
 
         for (int i = 0; i < partitionKey.size(); i++) {
-            if (name.equals(partitionKey.get(i).getName())) {
+            if (name.equalsIgnoreCase(partitionKey.get(i).getName())) {
```
tolbertam: Great catch!  I have one worry here which has to do with quoted identifiers.  For example you can have a table with the following schema:

```
CREATE TABLE simple.blah (
    "hI" int PRIMARY KEY,
    "Hi" int,
    hi int
)
```

By using equalsIgnoreCase, `"Hi"` would match against `"hI"` and be considered part of the routing key values, which we don't want.  I think to work around this you could use:

``` java
Metadata.handleId(name).equals(partitionKey.get(i).getName())
```

briandilley: `Metadata.handleId` has package protected access - should i just make it public?

briandilley: Addressed tolbertam's issue, please review.

tolbertam: Great point, I didn't even consider that `Metadata.handleId` wasn't public.  I'm not sure what the best direction is, but I think it seems reasonable to me to make it public although there could be reasons to not make it so.  @olim7t / @adutra / @newkek, thoughts?

adutra:
I would be more in favor of not exposing this method (because users really don't need it), but rather (and my fingers burn as I type) to create a copy of it in `com.datastax.driver.core.querybuilder.Utils`. Some other methods already suffered that same fate, e.g. `convert()` and `compose()`.

adutra:
Moreover, I think we should include this issue in 3.0.4 instead.

briandilley:
Ok, copied the method.  I'll let you deal with cherry-picking this down into other branches (ie: 3.0.4) if that's what you want to do. @adutra

briandilley:
So was this merged in? @adutra

adutra:
No it wasn't. I don't understand why it was closed. Anyway I will reopen and take care of cherry-picking or rebasing, don't worry (and authorship will be preserved of course). Thanks again for your contribution!

adutra:
Well, it looks like because we need to merge this into 3.0.x I will have to close this PR and open another one: #749 . Github fun :(

PR #744, opened 2016-09-11
JAVA-1286: Support setting and retrieving udt fields in QueryBuilder

tolbertam:
For [JAVA-1286](https://datastax-oss.atlassian.net/browse/JAVA-1286).

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/querybuilder/QueryBuilder.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/querybuilder/QueryBuilder.java
@@ -260,7 +260,7 @@ public final class QueryBuilder {
     public static String token(String... columnNames) {
         StringBuilder sb = new StringBuilder();
         sb.append("token(");
-        Utils.joinAndAppendNames(sb, null, ",", Arrays.asList((Object[]) columnNames));
+        Utils.joinAndAppendNames(sb, null, Arrays.asList((Object[]) columnNames));
         sb.append(')');
         return sb.toString();
     }
@@ -349,7 +349,7 @@ public final class QueryBuilder {
      * @param values the values
      * @return the corresponding where clause.
      * @throws IllegalArgumentException if the size of any tuple in {@code values} is not equal to {@code names.size()},
-     * or if {@code values} contains elements that are neither {@link List lists} nor {@link #bindMarker() bind markers}.
+     *                                  or if {@code values} contains elements that are neither {@link List lists} nor {@link #bindMarker() bind markers}.
      */
     public static Clause in(List<String> names, List<?> values) {
         return new Clause.CompoundInClause(names, values);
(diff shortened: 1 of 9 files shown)


driver-core/src/main/java/com/datastax/driver/core/querybuilder/Utils.java
```diff
@@ -32,7 +32,7 @@
 // Static utilities private to the query builder
 abstract class Utils {
 
-    private static final Pattern cnamePattern = Pattern.compile("\\w+(?:\\[.+\\])?");
+    private static final Pattern cnamePattern = Pattern.compile("\\w+(?:\\[.+\\]|\\.\\w+)?");
```
tolbertam: I admit that this may not be the best route.  Also supporting this may break existing user code (i.e. if they have columns named `"myc.ol"`).

driver-core/src/test/java/com/datastax/driver/core/querybuilder/QueryBuilderUDTExecutionTest.java
```diff
@@ -87,4 +88,39 @@ public void should_handle_collections_of_UDT() throws Exception {
...
+    public void should_support_setting_and_retrieving_udt_fields() {
+        //given
+        String table = "unfrozen_udt_table";
+        String udt = "person";
+        session().execute(createType(udt).addColumn("first", DataType.text()).addColumn("last", DataType.text()));
+        UserType userType = cluster().getMetadata().getKeyspace(keyspace).getUserType(udt);
+        assertThat(userType).isNotNull();
+
+        session().execute(createTable(table).addPartitionKey("k", DataType.text())
+                .addUDTColumn("u", udtLiteral(udt))
```
tolbertam, line 111: Maybe we should consider renaming `udtLiteral` to something else since unfrozen udts are now allowed (although maybe in future version?)

adutra:
To be honest I would be in favor of a different approach: allow `QueryBuilder.raw()` (or maybe `QueryBuilder.column()`) to be used as a column name in a set assignment. 

This would lead to a change (hopefully backwards-compatible) in the signature of `set()`: `public static Assignment set(Object name, Object value)`, and would require some changes in the `Assignment` class to detect when a variable name should or should not be enclosed in single quotes.

But I think this would be ultimately more in line with the way we currently protect strings from being mistankenly quoted by the query builder.

tolbertam:
> To be honest I would be in favor of a different approach: allow QueryBuilder.raw() (or maybe QueryBuilder.column()) to be used as a column name in a set assignment.

I am on the fence with that, but I think I've come around on your suggestion.  On the one hand, the current proposed code can break some users if they for whatever reason had periods in their column names and weren't quoting column names theirselves.   On the other hand, users should be in the right to expect the driver to not mistakenly quote the name in this particular case.

I think quoting the name for the user as it currently exists is inconsistent with other parts of our API and we should consider changing this in the future.  For example, we won't quote the keyspace when passed into `Cluster.connect(keyspace)`.  However, we can't simply change that here.  And even changing that in a future major release could be a problem too for users (even if we are really careful about communicating it).

In that case I agree that the right approach then is to come up with an alternative way to say "Here is my name, don't touch it" (as you've proposed) by allowing use of `column` or `raw` with `set`.   There will still be users who encounter their name being escaped when they wouldn't expect, but at least now they'll have a workaround.

newkek:
+1 for the `QueryBuilder.raw()` approach + we should study where else we could allow it to give people escape plans when the driver doesn't behave as expected.

olim7t:
Changed base to 3.x, make sure you use that branch for future rebases.

adutra:
Rebased on 3.x. 

On a side note, do we have an agreement on modifying `cnamePattern` vs modifying `set` signature?

adutra:
Turns out changing the parameter of a public method _is_ a breaking change :-/ it's a pity because this pattern could be applied in lots of places. But anyway, semver says no :)

adutra (dismissed):

driver-core/src/main/java/com/datastax/driver/core/querybuilder/Utils.java
```diff
@@ -33,7 +33,7 @@
 abstract class Utils {
 
     private static final Pattern alphanumeric = Pattern.compile("\\w+"); // this includes _
-    private static final Pattern cnamePattern = Pattern.compile("\\w+(?:\\[.+\\])?");
+    private static final Pattern cnamePattern = Pattern.compile("\\w+(?:\\[.+\\]|\\.\\w+)?");
```
olim7t: This does not handle "a.b.c" if there are multiple levels of nesting (is it possible?)

driver-core/src/test/java/com/datastax/driver/core/querybuilder/QueryBuilderUDTExecutionTest.java
olim7t: The name refers to expressing the type with a string literal, I think this still holds even if the type is unfrozen.

olim7t:
We could get around the breaking change by leaving `set(String, Object)` and adding an overload `set(Object, Object)`.
For selection, there is already a workaround with `select().raw("a.b")`.

Maybe we should also have a programmatic way to express it (`field("a", "b")`?), for cases when you generate a query from a UDT definition.

adutra:
> We could get around the breaking change by leaving set(String, Object) and adding an overload set(Object, Object).

Indeed, done.

> Maybe we should also have a programmatic way to express it (field("a", "b")?), for cases when you generate a query from a UDT definition.

Agreed and done. Although I named the method `path` instead of `field`, as I sense that in the future that dotted notation could be useful in more places than are currently supported.

tolbertam:
The changes look good to me, can't approve since I created the PR, but this has my thumbs up.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/QueryBuilder.java
```diff
@@ -589,6 +589,35 @@ public static Assignment set(String name, Object value) {
...
+     * Accepted types for the {@code name} parameter are:
+     * <ol>
+     * <li>String: {@code set("name", value)} generates {@code name = value}
+     * (this is equivalent to {@link #set(String, Object)});</li>
+     * <li>{@link #raw(String)}: {@code set(raw("name"), value)}
+     * generates {@code name = value};</li>
+     * <li>{@link #column(String)}: {@code set(column("name"), value)}
+     * generates {@code "name" = value}.</li>
+     * <li>{@link #path(String...)}: {@code set(path("udt", "field"), value)}
+     * generates {@code name.field = value}.</li>
```
olim7t: I wonder if it's worth mentioning `column`. I don't see any reason why you would want to use it here, and it's more confusing than helpful.

olim7t (approved):
LGTM, I did a few changes to the javadoc to address my remark.

PR #745, opened 2016-09-12
JAVA-1287: Add CDC to TableOptionsMetadata and Schema Builder

tolbertam:
In support of CASSANDRA-12041, add support for parsing cdc from table
options and add support for specification using schema builder.

For [JAVA-1287](https://datastax-oss.atlassian.net/browse/JAVA-1287).

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/schemabuilder/TableOptions.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/schemabuilder/TableOptions.java
@@ -67,6 +67,8 @@ public abstract class TableOptions<T extends TableOptions> extends SchemaStateme
 
     private Optional<SpeculativeRetryValue> speculativeRetry = Optional.absent();
 
+    private Optional<Boolean> cdc = Optional.absent();
+
     private List<String> customOptions = new ArrayList<String>();
 
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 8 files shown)


driver-core/src/main/java/com/datastax/driver/core/schemabuilder/TableOptions.java
```diff
@@ -358,6 +360,19 @@ public T speculativeRetry(SpeculativeRetryValue speculativeRetry) {
     }
 
     /**
+     * Define whether or not change data capture is enabled on this table.
+     * <p/>
+     * If no call is made to this method, the default value set by Cassandra is {@code false}.
```
adutra: Should we warn users that if they call this method and later on execute this statement against C\* < 3.8 they will get an error?

adutra:
👍 
It's odd because the C\* ticket also mentions a syntax change for keyspaces, e.g.:

``` sql
CREATE KEYSPACE ks WITH replication... AND cdc_datacenters= {'dc1','dc2'...};
ALTER KEYSPACE ks DROP CDCLOG;
```

But I couldn't find any reference to these changes in the current C\* trunk, so I guess it's been abandoned.

driver-core/src/main/java/com/datastax/driver/core/schemabuilder/TableOptions.java
tolbertam: Good point, we definitely should, will push update.

tolbertam:
> But I couldn't find any reference to these changes in the current C\* trunk, so I guess it's been abandoned.

I'll look into that, I would think there should still be a way to drop the existing logs, maybe it's not done through CQL now.

tolbertam:
> I'll look into that, I would think there should still be a way to drop the existing logs, maybe it's not done through CQL now.

Didn't find anything in the C\* commit other than the `with CDC` option for CQL, also didn't see any other way to drop the cdclogs from what I can tell.

olim7t:
Changed base to 3.x, make sure you use that branch for future rebases.

adutra (approved):

olim7t (approved):
Rebased on 3.x and fixed the changelog.
This looks good to me 👍

PR #747, opened 2016-09-12
JAVA-893: Refactor HostConnectionPool to a non-blocking implementation

olim7t:


driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -382,23 +404,11 @@ private void retry(final boolean retryCurrent, ConsistencyLevel newConsistencyLe
...
-                        sendRequest();
-                    } catch (Exception e) {
-                        setFinalException(null, new DriverInternalError("Unexpected exception while retrying query", e));
-                    }
-                }
-            });
+            if (queryStateRef.get().isCancelled())
+                return;
+
+            if (!retryCurrent || !query(h))
```
adutra, line 404: So retries are now executed on Netty I/O threads?

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -261,12 +264,18 @@ private RetryPolicy retryPolicy() {
                 logger.trace("[{}] Starting", id);
         }
 
-        void sendRequest() {
+        void findNextHostAndQuery() {
             try {
                 Host host;
                 while (!isDone.get() && (host = queryPlan.next()) != null && !queryStateRef.get().isCancelled()) {
+
+                    HostConnectionPool pool = manager.pools.get(host);
```
adutra: What is the motivation for checking the pool here _and_ inside the `query` method?

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -94,8 +96,10 @@
 
     private final AtomicReference<Owner> ownerRef = new AtomicReference<Owner>();
 
+    private final ListenableFuture<Connection> thisFuture;
+
```
adutra, line 102: minor nit: one extra blank line here.

driver-core/src/main/java/com/datastax/driver/core/PoolingOptions.java
```diff
@@ -400,30 +406,48 @@ public PoolingOptions setIdleTimeoutSeconds(int idleTimeoutSeconds) {
...
+        return this;
     }
 
     /**
-     * Sets the timeout when trying to acquire a connection from a host's pool.
+     * Returns the maximum number of requests that get enqueued if no connection is available.
+     *
+     * @return the maximum queue size.
+     */
+    public int getMaxQueueSize() {
```
adutra, line 430: Would it be worth giving this parameter a more meaningful name, e.g. `maxPendingConnectionRequests`, and making it clear in the javadocs that this number is meant _per connection pool_ (and not global)?

adutra (approved):

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
olim7t: Yes that's overkill, I'll get rid of this one.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
olim7t: Yes. There's nothing blocking or compute-intensive on that path anymore: if the pool is busy or the keyspace needs to be changed, that will be handled asynchronously via the future returned from borrow. So I figured the extra indirection wasn't worth it anymore.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
olim7t: removed

driver-core/src/main/java/com/datastax/driver/core/PoolingOptions.java
olim7t: I wouldn't use "request" because it's too similar to "network request", but if you can think of something else I'm open to suggestions, it's true that maxQueueSize is not great.

I added a sentence to the javadoc of the setter as you suggested.

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
```diff
@@ -384,7 +309,32 @@ public void returnConnection(Connection connection) {
...
+        while (!pendingBorrows.isEmpty()) {
+
+            // Try to acquire the right to use the connection
+            while (true) {
+                int inFlight = connection.inFlight.get();
+                if (inFlight >= Math.min(connection.maxAvailableStreams(), options().getMaxRequestsPerConnection(hostDistance))) {
+                    // Connection is full again, stop dequeuing
+                    return;
+                }
+                if (connection.inFlight.compareAndSet(inFlight, inFlight + 1))
```
tolbertam: Nice, I like this approach compared to the previous `signalAvailableConnection()` approach.    

It would be good to add a comment here why `inFlight` is incremented as it was initially confusing to me, but now I see that the previous implementation would increment inFlight as soon as it got the connection in `waitForConnection`.  It seems more natural in the previous implementation, but it also makes sense here (just took me a moment to get it).   It is done here instead in the different places where the behavior is handling when the Future completes and provides the connection (which at that point you could be outside scope of HostConnectionPool).  

My current interpretation is:

`if (connection.inFlight.compareAndSet(inFlight, inFlight + 1))`: Increment inFlight on the connection to account for a new borrow request.

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
```diff
@@ -62,9 +59,8 @@
     @VisibleForTesting
     final Set<Connection> trash = new CopyOnWriteArraySet<Connection>();
 
-    private volatile int waiter = 0;
-    private final Lock waitLock = new ReentrantLock(true);
-    private final Condition hasAvailableConnection = waitLock.newCondition();
+    private final Queue<SettableFuture<Connection>> pendingBorrows = new ConcurrentLinkedQueue<SettableFuture<Connection>>();
+    private final AtomicInteger pendingBorrowCount = new AtomicInteger();
```
tolbertam, line 61: Why not use `pendingBorrows.size()` instead of a separate counter?  I think it is because `ConcurrentLinkedQueue.size()` is not constant time, but was just wondering.   As we check this to determine whether or not we've reached `maxQueueSize` i worry about the count getting off somehow, but i see it is only incremented and decremented in 2 occurrences, so I am probably overworrying :)

driver-core/src/main/java/com/datastax/driver/core/PoolingOptions.java
```diff
@@ -400,30 +406,48 @@ public PoolingOptions setIdleTimeoutSeconds(int idleTimeoutSeconds) {
...
      */
+    @Deprecated
     public int getPoolTimeoutMillis() {
-        return poolTimeoutMillis;
+        return 0;
+    }
+
+    /**
+     * @deprecated the connection pool does not use a timeout anymore, incoming requests are now throttled with a
+     * threshold on the {@link #setMaxQueueSize(int)}  queue size}. This method has no effect.
```
tolbertam: dangling `}`, `{@link #setMaxQueueSize(int)}  queue size-->}<--`

driver-core/src/main/java/com/datastax/driver/core/PoolingOptions.java
tolbertam: I think the javadoc on the setter is really good.  I can't think of a better name other than `setMaxQueueSize` (but I am bad at coming up with names :) ).

driver-core/src/main/java/com/datastax/driver/core/PoolingOptions.java
```diff
@@ -151,7 +157,7 @@
     private volatile int maxRequestsPerConnectionRemote = UNSET;
 
     private volatile int idleTimeoutSeconds = DEFAULT_IDLE_TIMEOUT_SECONDS;
-    private volatile int poolTimeoutMillis = DEFAULT_POOL_TIMEOUT_MILLIS;
+    private volatile int maxQueueSize = DEFAULT_MAX_QUEUE_SIZE;
```
tolbertam, line 160: Thinking out loud, one small concern I have about using a queue instead of a timeout is that since there is no timeout on awaiting an available connection now it could be possible that a Future returned from `executeAsync` may never be completed if for some reason there is a bug in `HostConnectionPool`.  

Well, I suppose that possibility has always existed, there would need to be a bug in the driver for it to be a problem, which we would then have to fix :).    If a user wanted to be paranoid about it, they could timeout their Futures themselves (which is a good practice anyways).

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
```diff
@@ -612,8 +562,9 @@ public final CloseFuture closeAsync() {
 
         phase.set(Phase.CLOSING);
 
-        // Wake up all threads that wait
-        signalAllAvailableConnection();
+        for (SettableFuture<Connection> pendingBorrow : pendingBorrows) {
```
tolbertam, line 596: It looks like we use the pool phase to protect us from enqueuing borrows for a Pool that is already closed, but isn't there a possibility that a request may be added to `pendingBorrows` afterwards?

The possibility seems very slim, but couldn't the following happen:
1. thread-X: Enters `borrowConnection`, pool phase is `READY` so code proceeds.
2. thread-Y: Enters `closeAsync`, sets phase to `CLOSING`, fails futures in pendingBorrows.
3. thread-X: Notices `connections.isEmpty`evaluates to true, spawns new connection and calls `enqueue`.  Or any other condition that causes `enqueue` to be called.

There is now a future in `pendingBorrows` which never completes right?  That window seems really tiny but it does seem possible.  Might be able to reproduce it in a debugger and adding some breakpoints.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
tolbertam: Nice,  I remember seeing retries fill up the Cluster executor if your pool timeout > 0 on retries and pool was full.

tolbertam (commented):
Looks good to me so far!  Need to spend some time with the tests and run some perf tests but so far so good.

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
adutra: 👍 for commenting a bit more this line's intention, it made me pause for a couple seconds too.

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
adutra: It also reminded me some bad memories from `EventDebouncer` :) but indeed I think the counter is required because `size()` is O(N). I think the counter and the queue could be out of sync momentarily but I did not find any situation where this would lead to a bad behavior of the pool.

driver-core/src/main/java/com/datastax/driver/core/PoolingOptions.java
adutra: `maxConnectionWaitQueueSize`? Otherwise I throw in the towel too.

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
```diff
@@ -276,93 +264,30 @@ public Connection borrowConnection(long timeout, TimeUnit unit) throws Connectio
...
-                if (inFlight < minInFlight) {
-                    minInFlight = inFlight;
-                    leastBusy = connection;
-                }
+        while (true) {
+            int count = pendingBorrowCount.get();
+            if (count >= maxQueueSize) {
+                return Futures.immediateFailedFuture(new ConnectionException(host.getSocketAddress(),
+                        "Pool is busy (no available connection and the queue has reached its max size " + maxQueueSize
+                                + " )"));
```
tolbertam: There's an extra space here, i.e. : `reached its max size 256 )`

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
```diff
@@ -384,7 +309,32 @@ public void returnConnection(Connection connection) {
...
+                if (connection.inFlight.compareAndSet(inFlight, inFlight + 1))
+                    break;
+            }
+
+            SettableFuture<Connection> pendingBorrow = pendingBorrows.poll();
+            if (pendingBorrow == null) {
+                // Another thread has emptied the queue since our last check, restore the count
+                connection.inFlight.decrementAndGet();
+            } else {
+                pendingBorrowCount.decrementAndGet();
```
tolbertam, line 343: I observed in testing that `totalInFlight` was becoming offset and over time become a negative number.  I think it is because `totalInFlight` is not incremented here.

tolbertam (changes requested):

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -278,48 +279,64 @@ void sendRequest() {
...
-                    connection.release();
-                logger.error("Unexpected error while querying " + host.getAddress(), e);
-                logError(host.getSocketAddress(), e);
-                return false;
-            }
+
+                @Override
+                public void onFailure(Throwable t) {
+                    if (t instanceof TimeoutException) {
+                        logError(host.getSocketAddress(),
```
tolbertam: Is `TimeoutException` possible?  I noticed the `Pool is Busy` exception is a `ConnectionException`.

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
olim7t: Yes, it's because `size()` is not constant.

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
olim7t: I tried to explain it in the comment above the `while` loop: a successful increment is the precondition that gives you the "right" to use the connection. I'll see how I can explain it better.

driver-core/src/main/java/com/datastax/driver/core/PoolingOptions.java
olim7t: fixed

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
olim7t: Yes, I agree there's a race there.

driver-core/src/main/java/com/datastax/driver/core/exceptions/BusyPoolException.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    private BusyPoolException(InetSocketAddress address, int queueSize, Throwable cause) {
+        super(buildMessage(address, queueSize), cause);
+        this.address = address;
+        this.queueSize = queueSize;
+    }
+
+    private static String buildMessage(InetSocketAddress address, int queueSize) {
+        return String.format("[%s] Pool is busy (no available connection and the queue has reached its max size %d",
```
tolbertam: Missing closing `)`

driver-core/src/main/java/com/datastax/driver/core/PoolingOptions.java
```diff
@@ -415,7 +415,7 @@ public int getPoolTimeoutMillis() {
 
     /**
      * @deprecated the connection pool does not use a timeout anymore, incoming requests are now throttled with a
-     * threshold on the {@link #setMaxQueueSize(int)}  queue size}. This method has no effect.
+     * threshold on the {@link #setMaxQueueSize(int) queue size}. This method has no effect.
```
tolbertam, line 418: Should the closing `}` be after `(int)`?

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
```diff
@@ -285,6 +285,13 @@ private PoolingOptions options() {
 
         SettableFuture<Connection> future = SettableFuture.create();
         pendingBorrows.add(future);
+
+        // If we raced with shutdown, make sure the future will be completed. This has no effect if it was properly
+        // handled in closeAsync.
+        if (phase.get() == Phase.CLOSING) {
```
tolbertam, line 288: Nice, for some reason I thought handling this would be much more complex, but checking the state of the pool after adding the future makes perfect sense.

driver-core/src/main/java/com/datastax/driver/core/PoolingOptions.java
olim7t: it's a named link

tolbertam: oh, I didn't even know that was a thing you could do, nice!

tolbertam (approved):
Looks great so far.  Going to keep testing, but approving review since everything looks good.

tolbertam:
Just remembered, we should also update the Connection pooling doc, primary the `Acquisition Timeout` section (http://datastax.github.io/java-driver/manual/pooling/#acquisition-timeout) could be replaced.

tolbertam:
We can also remove the note about blocking on connections from `Known limitations` in the asynchronous programming doc (http://datastax.github.io/java-driver/manual/async/#known-limitations) 🎉

driver-core/src/main/java/com/datastax/driver/core/PoolingOptions.java
newkek: suggestion: `maxQueuedQueriesPerPool()` or `maxQueuedQueriesPerBusyPool()` if I got this right

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
```diff
@@ -196,17 +193,13 @@ private PoolingOptions options() {
...
+            return Futures.immediateFailedFuture(new ConnectionException(host.getSocketAddress(), "Pool is " + phase));
 
         if (connections.isEmpty()) {
-            if (!host.convictionPolicy.canReconnectNow())
-                throw new TimeoutException("Connection pool is empty, currently trying to reestablish connections");
-            else {
+            if (host.convictionPolicy.canReconnectNow()) {
                 int coreSize = options().getCoreConnectionsPerHost(hostDistance);
                 if (coreSize == 0) {
                     maybeSpawnNewConnection();
```
newkek, line 202: Not related to that ticket but I don't get that line, if users set the "core connections pool size" to 0, is the driver creating a new connection for all queries until the max number is reached?

olim7t:
Rebased on top of current 3.0.x.

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
olim7t: `maybeSpawnNewConnection()` protects against creating multiple connection in parallel, so it should create only one connection.

newkek: Well it is part of `borrowConnection()` which if I'm not mistaken is called for every request so even if it does one at a time, the call here only checks on options().getCoreConnectionsPerHost() which does not check the number of currently open connections, so it looks like we would keep on opening connections (one at a time) until the max is reached, is that correct? Not saying it's a bad behaviour, was just seeing that we don't mention that anywhere in docs

tolbertam:
Have completed testing on this and it is working great!    I made a few adjustments to the tests to accommodate for `ConnectionException` -> `BusyPoolException` and added another test.

tolbertam (changes requested):
Endurance tests revealed two bugs.  I made an attempt at fixing one of them on a [separate branch](https://github.com/datastax/java-driver/tree/java893_andytest):
1. When a new connection is added to the pool and is selected as `leastBusy` in `borrowConnection`, it calls `setKeyspaceAsync`.  If a bunch of threads enter here at same time, there is a run on `setKeyspaceAsync` which overwhelms stream ids < ProtocolV3.  It attempted to fix this in 01e8ee2 .
2. In `dequeue` we don't ensure that the keyspace is set correctly on the connection like in `borrowConnection`.  Because of this if you get a new connection in `addConnectionIfUnderConnection` and `dequeue` is called, no keyspace is set on the connection, so all borrow requests here will fail because no keyspace is set.   I added a work around for this to setKeyspace in `addConnectionIfUnderConnection` ( e07a220 ).  However, we should probably validate the keyspace in dequeue or do something proactive like update all pools' connections when the keyspace changes on a connection.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -492,23 +492,51 @@ void setKeyspace(String keyspace) throws ConnectionException {
...
+        SetKeyspaceAttempt attempt = new SetKeyspaceAttempt(keyspace);
+
+        // Check for an existing keyspace attempt.
+        while (true) {
+            SetKeyspaceAttempt existingAttempt = targetKeyspace.get();
+            // if existing attempts' keyspace matches what we are trying to set, use it.
+            if (attempt.equals(existingAttempt)) {
+                attempt = existingAttempt;
+                break;
+            } else if (targetKeyspace.compareAndSet(existingAttempt, attempt)) {
```
olim7t, line 511: If the CAS fails, we want to wait on the new attempt stored in the atomic reference (i.e. do `attempt = targetKeyspace.get()` again).
This will likely be a setKeyspace to the same keyspace that we raced with. Note that it could also target a different keyspace if the client program is constantly switching keyspaces, but in that case there's not much we can do to prevent requests from failing.

tolbertam: I agree, if the CAS fails, the new attempt is likely for the same keyspace, in that case the next iteration of the while loop should see the new `existingAttempt` has the same keyspace as what we are trying to set (line 501-502).   It isn't clear but the equals() on `SetKeyspaceAttempt` is on the keyspace (maybe I should just compare the keyspaces of the attempts so it is more clear?)

If the CAS succeeds and the existingAttempt is for a different keyspace, I think that's ok.  I added a debug statement (at 508) if the existingAttempt is still in flight, as this could cause some requests to fail because the keyspace is changing, as you said there is not much we can do about that.  Although maybe we should change that logging statement to warning so a user knows that this is happening, wdyt?

olim7t: Sorry I hadn't paid attention to the outer `while` loop :(

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -492,23 +492,51 @@ void setKeyspace(String keyspace) throws ConnectionException {
...
+                // Otherwise, start a new keyspace attempt replacing the existing one.
+                // This is ok, as any observers of the existing attempt will still complete.
+                if (existingAttempt != null && !existingAttempt.future.isDone()) {
+                    logger.debug("{} There is already an in progress attempt to set keyspace on this connection for keyspace '{}', beginning set keyspace for '{}'",
+                            this, existingAttempt.keyspace, keyspace);
```
olim7t: I agree we should change this to WARN, and add an explanation like "This can happen if you constantly change the keyspace at runtime". Otherwise the queries fail with errors that are not necessarily easy to debug, like "unconfigured table xxx".
The downside is that we will flood the logs if the client is really degenerate, but that's better than letting all queries fail without explanation.

tolbertam: Sounds good, will make that adjustment and push shortly (adding some tests too)

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
```diff
@@ -337,14 +337,33 @@ private void dequeue(Connection connection) {
...
                 // Another thread has emptied the queue since our last check, restore the count
                 connection.inFlight.decrementAndGet();
             } else {
                 totalInFlight.incrementAndGet();
                 pendingBorrowCount.decrementAndGet();
-                pendingBorrow.set(connection);
+                // Ensure that the keyspace set on the connection is the one set on the pool state, in the general case it will be.
+                ListenableFuture<Connection> setKeyspaceFuture = connection.setKeyspaceAsync(manager.poolsState.keyspace);
+                // Slight optimization, if the keyspace was already correct the future will be complete, so simply complete it here.
+                if (setKeyspaceFuture.isDone()) {
```
tolbertam, line 347: There's a bug here.  If the future is done and failed, we don't mark this future as failed and we probably should.  Also, if the future fails, we should probably try again instead of remaining in a broken state.  We should also be reactive to this too in `setKeyspaceAsync`

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -483,24 +488,62 @@ void setKeyspace(String keyspace) throws ConnectionException {
...
+            if (attempt.equals(existingAttempt)) {
+                return existingAttempt.future;
+            } else if (targetKeyspace.compareAndSet(existingAttempt, attempt)) {
+                // Otherwise, start a new keyspace attempt replacing the existing one.
+                // This is ok, as any observers of the existing attempt will still complete.
+                if (existingAttempt != null && !existingAttempt.future.isDone()) {
+                    logger.warn("{} There is already an in progress attempt to set keyspace on this connection for keyspace '{}', beginning set keyspace for '{}'.  "
+                                    + "This can happen if you constantly change the keyspace at runtime.",
+                            this, existingAttempt.keyspace, keyspace);
+                    ksFuture.setException(new DriverException("Aborting attempt to set keyspace to " + keyspace + " since there is already an in flight attempt to set keyspace to " + existingAttempt.keyspace + "."));
```
tolbertam: Should we create a specialized exception for this?   The other thing is perhaps we don't need the warning above since the code that handles the future failing will log an expected exception.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -483,24 +488,62 @@ void setKeyspace(String keyspace) throws ConnectionException {
...
                 }
+                logger.debug("{} Setting keyspace {}", this, keyspace);
+                // Note: we quote the keyspace below, because the name is the one coming from Cassandra, so it's in the right case already
+                Future future = write(new Requests.Query("USE \"" + keyspace + '"'));
+                Futures.addCallback(future, new FutureCallback<Message.Response>() {
+
+                    @Override
+                    public void onSuccess(Message.Response response) {
+                        if (response instanceof SetKeyspace) {
+                            logger.debug("{} Keyspace set to {}", Connection.this, keyspace);
```
tolbertam, line 521: Previously we'd set `Connection.keyspace` to whatever is in the response here.  Just thought I'd point this out since we are now just using the keyspace passed to this method, which I think will be equivalent (Cassandra doesn't return quoted version right?)

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -452,14 +456,15 @@ private void notifyOwnerWhenDefunct() {
     }
 
     String keyspace() {
-        return keyspace;
+        SetKeyspaceAttempt currentAttempt = targetKeyspace.get();
+        return currentAttempt == null ? null : currentAttempt.keyspace;
```
olim7t: `currentAttempt` cannot be null

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -483,24 +488,62 @@ void setKeyspace(String keyspace) throws ConnectionException {
...
-                    return MoreFutures.VOID_SUCCESS;
-                } else if (response.type == ERROR) {
-                    Responses.Error error = (Responses.Error) response;
-                    throw defunct(error.asException(address));
-                } else {
-                    throw defunct(new DriverInternalError("Unexpected response while setting keyspace: " + response));
+    ListenableFuture<Connection> setKeyspaceAsync(final String keyspace) throws ConnectionException, BusyConnectionException {
+        if (Objects.equal(keyspace(), keyspace)) {
+            return thisFuture;
+        }
```
olim7t: The switch to the current keyspace could still be in progress. I'd rather do:

```
SetKeyspaceAttempt existingAttempt = targetKeyspace.get();
if (Objects.equal(existingAttempt.keyspace, keyspace) {
    return existingAttempt.future;
}
```

And then remove that case from inside the `while` loop below.

tolbertam: Oh good point, I missed that.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
tolbertam: targetKeyspace gets set to null if the future fails, should it not do that?  The reason it does it is if set keyspace fails, we want to provide another opportunity to set the same keyspace.

tolbertam: can get around this by setting it to the same thing set in the Connection constructor, will do that.

tolbertam: Fixed.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
tolbertam: I made this change, but didn't remove the case from the while loop.  Justification for that is otherwise we'd have to create a new `SetKeyspaceAttempt` every time, whether or not we need to.  Instead a `SetKeyspaceAttempt` is only ever created if the current `targetKeyspace` doesn't match.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
olim7t: sgtm. There's no way to set the keyspace back to null, so resetting to `defaultKeyspaceAttempt` guarantees that the call will be retried.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
olim7t: The keyspace name always comes from a `SET_KEYSPACE` response, so it's the "internal" format: not quoted and in the correct case.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
olim7t: Since the error should be relatively rare, I think we can keep a generic DriverException.
+1 for removing the warning.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -505,10 +505,8 @@ void setKeyspace(String keyspace) throws ConnectionException {
                 return existingAttempt.future;
             } else if (!existingAttempt.future.isDone()) {
                 // If the existing attempt is still in flight, fail this attempt immediately.
-                logger.warn("{} There is already an in progress attempt to set keyspace on this connection for keyspace '{}', aborting set keyspace for '{}'.  "
-                                + "This can happen if you constantly change the keyspace at runtime.",
-                        this, existingAttempt.keyspace, keyspace);
-                ksFuture.setException(new DriverException("Aborting attempt to set keyspace to '" + keyspace + "' since there is already an in flight attempt to set keyspace to '" + existingAttempt.keyspace + "'."));
+                ksFuture.setException(new DriverException("Aborting attempt to set keyspace to '" + keyspace + "' since there is already an in flight attempt to set keyspace to '" + existingAttempt.keyspace + "'.  "
+                        + "This can happen if you constantly change the keyspace at runtime."));
```
olim7t: Suggestion: "This can happen if you try to USE different keyspaces from the same session simultaneously."

driver-core/src/main/java/com/datastax/driver/core/PoolingOptions.java
olim7t: Good catch ;)

PR #748, opened 2016-09-13
JAVA-1292: Adjusted frame length error breaks ability to read data

tolbertam:
For [JAVA-1292](https://datastax-oss.atlassian.net/browse/JAVA-1292)

Also addresses [JAVA-1293](https://datastax-oss.atlassian.net/browse/JAVA-1293): Make `DecoderForStreamId.MAX_FRAME_LENGTH` configurable by adding a SystemProperty `com.datastax.driver.NATIVE_TRANSPORT_MAX_FRAME_SIZE_IN_MB` which defaults to 256.

Should also consider backporting to 2.1.x

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Connection.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Connection.java
@@ -27,6 +27,7 @@ import io.netty.channel.*;
 import io.netty.channel.group.ChannelGroup;
 import io.netty.channel.group.DefaultChannelGroup;
 import io.netty.channel.socket.SocketChannel;
+import io.netty.handler.codec.DecoderException;
 import io.netty.handler.timeout.IdleStateEvent;
 import io.netty.handler.timeout.IdleStateHandler;
 import io.netty.util.Timeout;
@@ -993,7 +994,7 @@ class Connection {
             ResponseHandler handler = pending.remove(streamId);
             streamIdHandler.release(streamId);
             if (handler == null) {
-                /**
+                /*
                  * During normal operation, we should not receive responses for which we don't have a handler. There is
                  * two cases however where this can happen:
                  *   1) The connection has been defuncted due to some internal error and we've raced between removing the
@@ -1046,6 +1047,27 @@ class Connection {
             if (writer.get() > 0)
                 return;
 
+            if (cause instanceof DecoderException) {
+                Throwable error = cause.getCause();
+                // Special case, if we encountered a FrameTooLongException, raise exception on handler and don't defunct it since
+                // the connection is in an ok state.
+                if (error != null && error instanceof FrameTooLongException) {
+                    FrameTooLongException ftle = (FrameTooLongException) error;
+                    int streamId = ftle.getStreamId();
+                    ResponseHandler handler = pending.remove(streamId);
+                    streamIdHandler.release(streamId);
+                    if (handler == null) {
+                        streamIdHandler.unmark(streamId);
+                        if (logger.isDebugEnabled())
+                            logger.debug("{} FrameTooLongException received on stream {} but no handler set anymore (either the request has "
+                                    + "timed out or it was closed due to another error).", Connection.this, streamId);
+                        return;
+                    }
+                    handler.cancelTimeout();
+                    handler.callback.onException(Connection.this, ftle, System.nanoTime() - handler.startTime, handler.retryCount);
+                    return;
+                }
+            }
             defunct(new TransportException(address, String.format("Unexpected exception triggered (%s)", cause), cause));
         }
 
--- a/driver-core/src/main/java/com/datastax/driver/core/Frame.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Frame.java
@@ -16,6 +16,7 @@
 package com.datastax.driver.core;
 
 import com.datastax.driver.core.exceptions.DriverInternalError;
+import com.datastax.driver.core.exceptions.FrameTooLongException;
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelHandlerContext;
@@ -189,42 +190,44 @@ class Frame {
     }
 
     static final class Decoder extends ByteToMessageDecoder {
-        static final DecoderForStreamIdSize decoderV1 = new DecoderForStreamIdSize(1);
-        static final DecoderForStreamIdSize decoderV3 = new DecoderForStreamIdSize(2);
+        private DecoderForStreamIdSize decoder;
 
         @Override
         protected void decode(ChannelHandlerContext ctx, ByteBuf buffer, List<Object> out) throws Exception {
             if (buffer.readableBytes() < 1)
                 return;
 
-            int version = buffer.getByte(buffer.readerIndex());
-            // version first bit is the "direction" of the frame (request or response)
-            version = version & 0x7F;
+            // Initialize sub decoder on first message.  No synchronization needed as
+            // decode is always called from same thread.
+            if (decoder == null) {
+                int version = buffer.getByte(buffer.readerIndex());
+                // version first bit is the "direction" of the frame (request or response)
+                version = version & 0x7F;
+                decoder = new DecoderForStreamIdSize(version, version >= 3 ? 2 : 1);
+            }
 
-            DecoderForStreamIdSize decoder = (version >= 3) ? decoderV3 : decoderV1;
             Object frame = decoder.decode(ctx, buffer);
             if (frame != null)
                 out.add(frame);
         }
 
         static class DecoderForStreamIdSize extends LengthFieldBasedFrameDecoder {
-            private static final int MAX_FRAME_LENGTH = 256 * 1024 * 1024; // 256 MB
-            private final int opcodeOffset;
+            // The maximum response frame length allowed.  Note that C* does not currently restrict the length of its responses (CASSANDRA-12630).
+            private static final int MAX_FRAME_LENGTH = SystemProperties.getInt("com.datastax.driver.NATIVE_TRANSPORT_MAX_FRAME_SIZE_IN_MB", 256) * 1024 * 1024; // 256 MB
+            private final int protocolVersion;
 
-            DecoderForStreamIdSize(int streamIdSize) {
+            DecoderForStreamIdSize(int protocolVersion, int streamIdSize) {
                 super(MAX_FRAME_LENGTH, /*lengthOffset=*/ 3 + streamIdSize, 4, 0, 0, true);
-                this.opcodeOffset = 2 + streamIdSize;
+                this.protocolVersion = protocolVersion;
             }
 
             @Override
             protected Object decode(ChannelHandlerContext ctx, ByteBuf buffer) throws Exception {
+                // Capture current index in case we need to get the stream id.
+                // If a TooLongFrameException is thrown the readerIndex will advance to the end of
+                // the buffer (or past the frame) so we need the position as we entered this method.
+                int curIndex = buffer.readerIndex();
                 try {
-                    if (buffer.readableBytes() < opcodeOffset + 1)
-                        return null;
-
-                    // Validate the opcode (this will throw if it's not a response)
-                    Message.Response.Type.fromOpcode(buffer.getByte(buffer.readerIndex() + opcodeOffset));
-
                     ByteBuf frame = (ByteBuf) super.decode(ctx, buffer);
                     if (frame == null) {
                         return null;
@@ -232,11 +235,17 @@ class Frame {
                     // Do not deallocate `frame` just yet, because it is stored as Frame.body and will be used
                     // in Message.ProtocolDecoder or Frame.Decompressor if compression is enabled (we deallocate
                     // it there).
-                    return Frame.create(frame);
+                    Frame theFrame = Frame.create(frame);
+                    // Validate the opcode (this will throw if it's not a response)
+                    Message.Response.Type.fromOpcode(theFrame.header.opcode);
+                    return theFrame;
                 } catch (CorruptedFrameException e) {
                     throw new DriverInternalError(e);
                 } catch (TooLongFrameException e) {
-                    throw new DriverInternalError(e);
+                    int streamId = protocolVersion > 2 ?
+                            buffer.getShort(curIndex + 2) :
+                            buffer.getByte(curIndex + 2);
+                    throw new FrameTooLongException(streamId);
                 }
             }
         }
--- a/driver-core/src/test/java/com/datastax/driver/core/FrameLengthTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/FrameLengthTest.java
@@ -1,0 +1,121 @@
+/*
+ *      Copyright (C) 2012-2015 DataStax Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+package com.datastax.driver.core;
+
+import com.datastax.driver.core.exceptions.FrameTooLongException;
+import com.datastax.driver.core.querybuilder.Insert;
+import com.datastax.driver.core.schemabuilder.Create;
+import com.datastax.driver.core.schemabuilder.SchemaBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import java.nio.ByteBuffer;
+import java.util.Collection;
+import java.util.Random;
+
+import static com.datastax.driver.core.querybuilder.QueryBuilder.*;
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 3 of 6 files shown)


driver-core/src/test/java/com/datastax/driver/core/FrameLengthTest.java
```diff
@@ -0,0 +1,128 @@
...
+            // Both hosts should have been tried.
+            assertThat(e.getErrors()).hasSize(2);
+        } finally {
+            // Check that TooLongFrameException is encountered as it isn't raised to the NoHostAvailableException.
+            assertThat(logs.get()).contains("io.netty.handler.codec.TooLongFrameException");
+            connectionLogger.setLevel(originalLevel);
+            connectionLogger.removeAppender(logs);
+        }
+        // Connection will be lost for both hosts, but they should be able to reconnect.
+        waitForUp(ipOfNode(1), cluster(), 10);
```
tolbertam: When this exception is encountered a `DriverInternalError` is raised, which causes the connection to be defuncted and the request retried on the next host, which will also be defuncted.  Is this the right behavior?  It might be better to fail the request, closed the connection (nor not?) and not try on the next host.

driver-core/src/test/java/com/datastax/driver/core/FrameLengthTest.java
```diff
@@ -0,0 +1,128 @@
...
+import static com.datastax.driver.core.querybuilder.QueryBuilder.*;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Fail.fail;
+
+@CCMConfig(numberOfNodes = 2)
+public class FrameLengthTest extends CCMTestsSupport {
+
+    static {
+        // Set max frame size to 1MB to make it easier to manifest frame length error.
+        System.setProperty("com.datastax.driver.NATIVE_TRANSPORT_MAX_FRAME_SIZE_IN_MB", "1");
```
tolbertam: this gets evaluated earlier then I would have expected and breaks other tests.  Will fix.

driver-core/src/main/java/com/datastax/driver/core/Frame.java
```diff
@@ -205,10 +200,17 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf buffer, List<Object> ou
             int version = buffer.getByte(0);
             // version first bit is the "direction" of the frame (request or response)
             version = version & 0x7F;
+
+            DecoderForStreamIdSize decoder = decoderForStreamIdSize.get();
```
tolbertam: Needed because we can't be sure the protocol version the server is using until we actually get something from it.  Although I'm not sure an `AtomicReference` is necessary here, i'd expect `decode` will only called one at a time and from the same i/o thread, wdyt?  In general could there be a better approach than this?

driver-core/src/main/java/com/datastax/driver/core/Frame.java
```diff
@@ -201,19 +201,32 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf buffer, List<Object> ou
             // version first bit is the "direction" of the frame (request or response)
             version = version & 0x7F;
 
-            DecoderForStreamIdSize decoder = (version >= 3) ? decoderV3 : decoderV1;
+            DecoderForStreamIdSize decoder = decoderForStreamIdSize.get();
+            if (decoder == null) {
```
tolbertam, line 202: Needed because we can't be sure the protocol version the server is using until we actually get something from it.  Although I'm not sure an `AtomicReference` is necessary here, i'd expect `decode` will only called one at a time and from the same i/o thread, wdyt?  In general could there be a better approach than this?

adutra: I think the only change required here is to remove the `static` modifier so that the decoders are not shared among connections.

The following makes your test pass as well:

``` java
    static final class Decoder extends ByteToMessageDecoder {
        private final DecoderForStreamIdSize decoderV1 = new DecoderForStreamIdSize(1);
        private final DecoderForStreamIdSize decoderV3 = new DecoderForStreamIdSize(2);

        @Override
        protected void decode(ChannelHandlerContext ctx, ByteBuf buffer, List<Object> out) throws Exception {
            if (buffer.readableBytes() < 1)
                return;

            int version = buffer.getByte(0);
            // version first bit is the "direction" of the frame (request or response)
            version = version & 0x7F;

            DecoderForStreamIdSize decoder = (version >= 3) ? decoderV3 : decoderV1;
            Object frame = decoder.decode(ctx, buffer);
            if (frame != null)
                out.add(frame);
        }
...
```

tolbertam: That will work in the general case, the main reason for restricting to one `DecoderForStreamIdSize` is that if you reach the frame length limit, the `DecoderForStreamIdSize` instance tracks the remaining bytes to consume and discards them until remnant of the frame is discarded.  Although in general you should only expect frames sharing the same protocol version, and if that didn't happen the code I have here would fail in unexpected ways anyways, so maybe the best solution is to do as you suggest.

adutra: Hmm I see. Conceptually speaking, you are right that `Decoder` should instantiate only one `DecoderForStreamIdSize` and use it consistently, so I'm ok with your approach as well.

tolbertam: > you are right that Decoder should instantiate only one DecoderForStreamIdSize and use it consistently, so I'm ok with your approach as well.

Sounds good, I think if we can count on C\* always sending the same protocol version on a connection this approach works, but added an assertion just to see if that would ever be a problem (it helped me capture the bug i had previously where I passed in the protocolVersion originally).   Wasn't sure if using an `AtomicReference` was required or not.  I'd expect decode to only ever be called one at a time, but wasn't sure if that was a guarantee netty makes.

driver-core/src/test/java/com/datastax/driver/core/FrameLengthTest.java
olim7t: Yes, it would be better to fail the request (i.e. rethrow the error to the client) to stop retries. With your fix the connection is still working, so I don't see any reason to mark it defunct.

driver-core/src/main/java/com/datastax/driver/core/Frame.java
olim7t: The decoder is a `ChannelHandler`, so as you noted previously `decode` will always be invoked from the same thread (the channel's). You're safe using a plain private field without any synchronization.

driver-core/src/main/java/com/datastax/driver/core/Frame.java
```diff
@@ -201,19 +201,32 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf buffer, List<Object> ou
...
+
             Object frame = decoder.decode(ctx, buffer);
             if (frame != null)
                 out.add(frame);
         }
 
         static class DecoderForStreamIdSize extends LengthFieldBasedFrameDecoder {
-            private static final int MAX_FRAME_LENGTH = 256 * 1024 * 1024; // 256 MB
+            // The maximum response frame length allowed.  Note that C* does not currently restrict the length of its responses (CASSANDRA-12630).
+            private static final int MAX_FRAME_LENGTH = SystemProperties.getInt("com.datastax.driver.NATIVE_TRANSPORT_MAX_FRAME_SIZE_IN_MB", 256) * 1024 * 1024; // 256 MB
```
olim7t, line 216: This could also go into `ProtocolOptions`, however the annoying thing with that class is that every new field requires a new `Cluster.Builder` method (there is no `withProtocolOptions()`). Changing the max frame size is an edge case, so I guess this can stay as a system property.

driver-core/src/main/java/com/datastax/driver/core/Frame.java
tolbertam: > You're safe using a plain private field without any synchronization.

awesome, I thought that was the case but wasn't completely sure,  I'll nix the `AtomicReference`.

driver-core/src/main/java/com/datastax/driver/core/Frame.java
tolbertam: Sounds good 👍 , I agree that it would be unlikely that this will be a configuration users will very rarely touch.

driver-core/src/test/java/com/datastax/driver/core/FrameLengthTest.java
tolbertam: 👍 , I'll see if I can find a good way to make it fail and not retry.

tolbertam: There are a few things to overcome around that:
1. If we hit a frame too large, the decoder discards all readable bytes up until the frame length.  However, if readable bytes < frame length, all bytes are discarded and the Decoder enters a state where it will throw away bytes up to the frame size.  This works well, however in the parent `Frame.Decoder` it checks the protocol version and opcode every time regardless of whether or not we are in this state and we can't detect the current state in the subclass `DecoderForStreamIdSize` (from what I can tell).  Will keep looking into that.  EDIT:  We could disable failFast, which will cause the decoder to not fail until we've consumed the entire frame.  EDIT 2: That won't work either since the frame bytes get skipped until the frame is done.
2. From what I can tell, there's no where else in code where an exception raised while Decoding does anything but close the connection.   We can handle this by checking the exception cause in `Connection.Dispatcher.handlerCaught` and then notifying the appropriate handler (we'd have to parse the stream id to do that, but I can overcome that) and not defuncting the connection.

tolbertam: 1. Need to decide whether or not this exception should be passed to `RetryPolicy#onRequestError`.  I would think not, since the default policy retries all request errors, and we likely would not want this ever to be retried.

tolbertam: I think I can work around the first problem by doing the following:
1. After the first message, we assume the version of all frames will be the same on the same connection.  This way we don't have to parse version every time.
2. Check the opcode after creating the frame.  The Frame body is not parsed until later anyways, so this should be ok, the disadvantage of this is that we may fail later on in the process. I.E. if the payload is malformed and the full frame length is never received, the request would just time out instead.  That might not be too bad if the data is bad anyways.

tolbertam: That seems to work, but may not be the best route.   I'll push something up for review.

tolbertam:
Pushed a proof of concept to address the issue in [this discussion](https://github.com/datastax/java-driver/pull/748#discussion_r78486414).  Now when `TooLongFrameException` is encountered, the exception is raised to the user (wrapped in new `DriverException`) and the connection remains open.   Previously it would close the connection and try the next host, which will likely cause the error to be propagated to all hosts.

tolbertam:
Rebased on 3.0.x with the change from c658e73 and updated other code i added to take readerIndex into account.

driver-core/src/main/java/com/datastax/driver/core/Frame.java
```diff
@@ -189,54 +190,62 @@ Frame with(ByteBuf newBody) {
...
                     if (frame == null) {
                         return null;
                     }
                     // Do not deallocate `frame` just yet, because it is stored as Frame.body and will be used
                     // in Message.ProtocolDecoder or Frame.Decompressor if compression is enabled (we deallocate
                     // it there).
-                    return Frame.create(frame);
+                    Frame theFrame = Frame.create(frame);
+                    // Validate the opcode (this will throw if it's not a response)
+                    Message.Response.Type.fromOpcode(theFrame.header.opcode);
```
tolbertam, line 240: One difference w/ previous implementation was it previously parsed the opcode before calling `super.decode`, this wouldn't work if you were in the middle of discarding a frame because the previous frame bytes would have already been discarded.   This would explain why you previously would get an exception like:

```
io.netty.handler.codec.DecoderException: com.datastax.driver.core.exceptions.DriverInternalError: Unknown response opcode XX
```

after a `TooLongFrameException` because the opcode bytes would have been previously discarded.

This means that if you get an invalid opcode, it now won't defunct the connection until we've received the whole frame.  I don't think that is a problem, but thought I'd point out the difference.

driver-core/src/main/java/com/datastax/driver/core/exceptions/TooLongFrameException.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+package com.datastax.driver.core.exceptions;
+
+/**
+ * Indicates that the response frame for a request exceeded
+ * {@link com.datastax.driver.core.Frame.Decoder.DecoderForStreamIdSize#MAX_FRAME_LENGTH}
+ * (default: 256MB, configurable via com.datastax.driver.NATIVE_TRANSPORT_MAX_FRAME_SIZE_IN_MB
+ * system property) and thus was not parsed.
+ */
+public class TooLongFrameException extends DriverException {
```
adutra: Shouldn't this be considered a subtype of `DriverInternalError`?

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -993,14 +994,15 @@ protected void channelRead0(ChannelHandlerContext ctx, Message.Response response
...
-                /**
+                /*
                  * During normal operation, we should not receive responses for which we don't have a handler. There is
                  * two cases however where this can happen:
                  *   1) The connection has been defuncted due to some internal error and we've raced between removing the
                  *      handler and actually closing the connection; since the original error has been logged, we're fine
                  *      ignoring this completely.
                  *   2) This request has timed out. In that case, we've already switched to another host (or errored out
                  *      to the user). So log it for debugging purpose, but it's fine ignoring otherwise.
+                 *   3) The connection encountered a frame too large to process.
```
adutra: In case of frame too large, wouldn't `exceptionCaught()` be called instead of `channelRead0()`?

adutra (approved):

driver-core/src/main/java/com/datastax/driver/core/exceptions/TooLongFrameException.java
tolbertam: I considered that, but this comment in the `DriverInternalError` documentation scared me away:

```
 This should never be raised and indicates a bug (either in the driver or in Cassandra).
```

I didn't consider this to be a bug so I put it at the `DriverException` level instead, but I can be convinced to put it under `DriverInternalError` too.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
tolbertam: You are right, I was more commenting on the fact that this is a scenario where there is no handler, but it does seem wrong for this to be here since when this happens it'd be handled in `exceptionCaught()` like you said.   I'll fix this 👍

driver-core/src/main/java/com/datastax/driver/core/exceptions/TooLongFrameException.java
adutra: Ok makes sense to not inherit from DIE, thanks for clarifying.

olim7t: This will be exposed to the client (as the `DefaultResultSetFuture`'s cause, or rethrown from the sync API), so definite +1 for `DriverException`.
For the same reason, I think it should have a message, and it's probably not necessary to nest the cause (which is a Netty type so not part of our API, and doesn't bring any additional info).

olim7t: `FrameTooLongException` sounds more natural, WDYT? (think of `ClassNotFoundException`).

I know it will look weird in the code where we interact with the Netty exception, but I'd take that for a better name in our public API.

driver-core/src/main/java/com/datastax/driver/core/Frame.java
```diff
@@ -189,54 +190,62 @@ Frame with(ByteBuf newBody) {
...
+                    return theFrame;
                 } catch (CorruptedFrameException e) {
                     throw new DriverInternalError(e);
-                } catch (TooLongFrameException e) {
-                    throw new DriverInternalError(e);
+                } catch (io.netty.handler.codec.TooLongFrameException e) {
+                    int streamId = protocolVersion > 2 ?
+                            buffer.getShort(curIndex + 2) :
+                            buffer.getByte(curIndex + 2);
+                    throw new TooLongFrameException(streamId, e);
```
olim7t: We catch a subclass of `DecoderException`, wrap it in our own type which will itself be wrapped into another `DecoderException` by the calling Netty code. In `exceptionCaught` we unwrap the outer one to rethrow our exception to the client.
That's a bit convoluted but I think it's acceptable for this edge case. I've thought of rethrowing Netty's `TooLongFrameException` directly, but that means we'd have to store the streamId elsewhere (maybe as a context attribute) which is a bit less clean.
So no suggestion really but I just wanted to mention it :)

olim7t (changes requested):
Looks great overall, two minor suggestions wrt `TooLongFrameException`.

driver-core/src/main/java/com/datastax/driver/core/exceptions/TooLongFrameException.java
tolbertam: `FrameTooLongException` sounds more natural to me as well, i'll also not pass the netty exception in the cause.

driver-core/src/main/java/com/datastax/driver/core/Frame.java
tolbertam: I agree 100% with you there, it is very convoluted and this was the part of the change I was least comfortable with.   I didn't even catch the detail that `TooLongFrameException` was a subclass of `DecoderException` so it is even stranger than I had thought :).

tolbertam:
Renamed `TooLongFrameException` -> `FrameTooLongException` and rebased on 3.0.x

olim7t (approved):

PR #749, opened 2016-09-13
JAVA-1285: QueryBuilder routing key auto-discovery should handle case-sensitive column names.

adutra:
This is the same as #743 but targeting 3.0.x.


olim7t:
lgtm 👍

tolbertam:
lgtm too 👍

olim7t (approved):
Approved

PR #754, opened 2016-09-21
Relax auth error metric count validation in reconnection test

tolbertam:
Test previously asserted that the getAuthenticationErrors() metric
incremented at the same rate as the AuthProvider#newAuthenticator count,
which may not be true as we could be checking between a newAuthenticator
request and an authentication request happening.  Instead ensure that
the count goes up.


tolbertam:
Looks like this works, should be good to merge.

adutra (approved):

PR #755, opened 2016-09-21
JAVA-1257: Implement a customizable StatementFormatter.

adutra:
Note to reviewers:

* The core functionality is in `StatementFormatter`. It is only new code and should not introduce regressions in any way.
* This PR also introduces `EnhancedQueryLogger`, an equivalent to `QueryLogger` where the formatting tasks are delegated to `StatementFormatter`.
* This PR also refactors `Statement.toString()`. I see a huge advantage in having a common system to print statements. The problem is that it was implicitly assumed so far that `toString()` would print the query string, but that assumption is now broken. Users willing to retrieve the query string should always call `getQueryString()`. I updated the upgrade guide accordingly.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -1655,7 +1655,11 @@ public class Cluster implements Closeable {
 
         void reportQuery(Host host, Statement statement, Exception exception, long latencyNanos) {
             for (LatencyTracker tracker : latencyTrackers) {
-                tracker.update(host, statement, exception, latencyNanos);
+                try {
+                    tracker.update(host, statement, exception, latencyNanos);
+                } catch (Exception e) {
+                    logger.error(String.format("Tracker %s threw exception: %s", tracker, e.getMessage()), e);
+                }
             }
         }
 
--- a/driver-core/src/main/java/com/datastax/driver/core/EnhancedQueryLogger.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/EnhancedQueryLogger.java
@@ -1,0 +1,471 @@
+/*
+ *      Copyright (C) 2012-2015 DataStax Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+package com.datastax.driver.core;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
+
+/**
+ * A configurable {@link LatencyTracker} that logs all executed statements.
+ * <p/>
+ * Typically, client applications would instantiate one single query logger (using its {@link Builder builder}),
+ * configure it and register it on the relevant {@link Cluster} instance, e.g.:
+ * <p/>
+ * <pre>
+ * Cluster cluster = ...
+ * EnhancedQueryLogger queryLogger = EnhancedQueryLogger.builder()
+ *     .withConstantThreshold(...)
+ *     .build();
+ * cluster.register(queryLogger);
+ * </pre>
+ * <p/>
+ * Refer to the {@link Builder} documentation for more information on
+ * configuration settings for the query logger.
+ * <p/>
+ * Once registered, the query logger will log every statement executed by the driver;
+ * note that it will never log null statements nor any special statement used internally by the driver.
+ * <p/>
+ * There is one log for each request to a Cassandra node; because the driver sometimes retries the same statement on multiple nodes,
+ * a single statement execution (for example, a single call to {@link Session#execute(Statement)}) can produce multiple logs on
+ * different nodes.
+ * <p/>
+ * For more flexibility, the query logger uses 3 different {@link Logger} instances:
+ * <p/>
+ * <ol>
+ * <li>{@link #NORMAL_LOGGER}: used to log normal queries, i.e., queries that completed successfully
+ * within a configurable threshold in milliseconds.</li>
+ * <li>{@link #SLOW_LOGGER}: used to log slow queries, i.e., queries that completed successfully
+ * but that took longer than a configurable threshold in milliseconds to complete.</li>
+ * <li>{@link #ERROR_LOGGER}: used to log unsuccessful queries, i.e.,
+ * queries that did not completed normally and threw an exception.
+ * Note this this logger will also print the full stack trace of the reported exception.</li>
+ * </ol>
+ * <p/>
+ * The appropriate logger is chosen according to the following algorithm:
+ * <ol>
+ * <li>if an exception has been thrown: use {@link #ERROR_LOGGER};</li>
+ * <li>otherwise, if the reported latency is greater than the configured threshold in milliseconds: use {@link #SLOW_LOGGER};</li>
+ * <li>otherwise, use {@link #NORMAL_LOGGER}.</li>
+ * </ol>
+ * <p/>
+ * All loggers are activated by setting their levels to {@code DEBUG} or {@code TRACE} (including {@link #ERROR_LOGGER}).
+ * If the level is set to {@code TRACE}, then the query parameters (if any) will be logged as well (names and actual values).
+ * <p/>
+ * <strong>Constant thresholds vs. Dynamic thresholds</strong>
+ * <p/>
+ * Currently the EnhancedQueryLogger can track slow queries in two different ways:
+ * using a {@link Builder#withConstantThreshold(long)} constant threshold} in milliseconds (which is the default
+ * behavior), or using a {@link Builder#withDynamicThreshold(PercentileTracker, double) dynamic threshold}
+ * based on latency percentiles.
+ * <p/>
+ * This class is thread-safe.
+ *
+ * @since 3.2.0
+ */
+public abstract class EnhancedQueryLogger implements LatencyTracker {
+
+    /**
+     * The default latency threshold in milliseconds beyond which queries are considered 'slow'
+     * and logged as such by the driver.
+     */
+    public static final long DEFAULT_SLOW_QUERY_THRESHOLD_MS = 5000;
+
+    /**
+     * The default latency percentile beyond which queries are considered 'slow'
+     * and logged as such by the driver.
+     */
+    public static final double DEFAULT_SLOW_QUERY_THRESHOLD_PERCENTILE = 99.0;
+
+    // Loggers
+
+    /**
+     * The logger used to log normal queries, i.e., queries that completed successfully
+     * within a configurable threshold in milliseconds.
+     * <p/>
+     * This logger is activated by setting its level to {@code DEBUG} or {@code TRACE}.
+     * Additionally, if the level is set to {@code TRACE}, then the query parameters (if any)
+     * will be logged.
+     * <p/>
+     * The name of this logger is {@code com.datastax.driver.core.QueryLogger.NORMAL}.
+     */
+    public static final Logger NORMAL_LOGGER = LoggerFactory.getLogger("com.datastax.driver.core.QueryLogger.NORMAL");
+
+    /**
+     * The logger used to log slow queries, i.e., queries that completed successfully
+     * but whose execution time exceeded a configurable threshold in milliseconds.
+     * <p/>
+     * This logger is activated by setting its level to {@code DEBUG} or {@code TRACE}.
+     * Additionally, if the level is set to {@code TRACE}, then the query parameters (if any)
+     * will be logged.
+     * <p/>
+     * The name of this logger is {@code com.datastax.driver.core.QueryLogger.SLOW}.
+     */
+    public static final Logger SLOW_LOGGER = LoggerFactory.getLogger("com.datastax.driver.core.QueryLogger.SLOW");
+
+    /**
+     * The logger used to log unsuccessful queries, i.e., queries that did not complete normally and threw an exception.
+     * <p/>
+     * This logger is activated by setting its level to {@code DEBUG} or {@code TRACE}.
+     * Additionally, if the level is set to {@code TRACE}, then the query parameters (if any)
+     * will be logged.
+     * <p/>
+     * Note this this logger will also print the full stack trace of the reported exception.
+     * <p/>
+     * The name of this logger is {@code com.datastax.driver.core.QueryLogger.ERROR}.
+     */
+    public static final Logger ERROR_LOGGER = LoggerFactory.getLogger("com.datastax.driver.core.QueryLogger.ERROR");
+
+    // Message templates
+
+    private static final String NORMAL_TEMPLATE = "[%s] [%s] Query completed normally, took %s ms: %s";
+
+    private static final String SLOW_TEMPLATE_MILLIS = "[%s] [%s] Query too slow, took %s ms: %s";
+
+    private static final String SLOW_TEMPLATE_PERCENTILE = "[%s] [%s] Query too slow, took %s ms (%s percentile = %s ms): %s";
+
+    private static final String ERROR_TEMPLATE = "[%s] [%s] Query error after %s ms: %s";
+
+    protected volatile Cluster cluster;
+
+    private final StatementFormatter formatter;
+
+    private EnhancedQueryLogger(StatementFormatter formatter) {
+        this.formatter = formatter;
+    }
+
+    /**
+     * Creates a new {@link EnhancedQueryLogger.Builder} instance.
+     * <p/>
+     * This is a convenience method for {@code new EnhancedQueryLogger.Builder()}.
+     *
+     * @return the new EnhancedQueryLogger builder.
+     * @throws NullPointerException if {@code cluster} is {@code null}.
+     */
+    public static EnhancedQueryLogger.Builder builder() {
+        return new EnhancedQueryLogger.Builder();
+    }
+
+    @Override
+    public void onRegister(Cluster cluster) {
+        this.cluster = cluster;
+    }
+
+    @Override
+    public void onUnregister(Cluster cluster) {
+        // nothing to do
+    }
+
+    /**
+     * A QueryLogger that uses a constant threshold in milliseconds
+     * to track slow queries.
+     * This implementation is the default.
+     */
+    public static class ConstantThresholdQueryLogger extends EnhancedQueryLogger {
+
+        private volatile long slowQueryLatencyThresholdMillis;
+
+        private ConstantThresholdQueryLogger(long slowQueryLatencyThresholdMillis, StatementFormatter formatter) {
+            super(formatter);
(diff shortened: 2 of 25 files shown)


driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1593,7 +1593,11 @@ boolean removeSession(Session session) {
 
         void reportQuery(Host host, Statement statement, Exception exception, long latencyNanos) {
             for (LatencyTracker tracker : latencyTrackers) {
-                tracker.update(host, statement, exception, latencyNanos);
+                try {
```
adutra, line 1658: Not really related, but I thought it would be wise to catch exceptions here instead of letting them propagate.

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
```diff
@@ -189,21 +189,4 @@ public boolean hasValues() {
         return hasValues(CodecRegistry.DEFAULT_INSTANCE);
     }
 
-    /**
-     * Returns this statement as a CQL query string.
```
adutra: This is not true anymore as this method now prints much more than just the query string. Ironically, these very same javadocs state a few lines below that the query string only "does not convey all the information stored in `Statement` objects".

olim7t:
Sorry but this looks really complicated... Printing statements in the logs is a simple problem, I don't think it should require 6 new types.
Also I'm not comfortable with the breaking changes, even though they're on protected methods they're still breaking.

I think we should step back and reconsider what we're trying to solve here. What are the things that a user could reasonably expect from the logger and are currently not possible?
We should provide a generic format that's reasonable for most users, not necessarily make everything configurable; e.g. if someone wants to use a different separator than `,`, then it's probably time to write their own logger (we need to make sure that it's possible, iirc the logger uses some package-private APIs so we need a way to expose them).

adutra:
> Printing statements in the logs is a simple problem, I don't think it should require 6 new types.

Not so simple as that imo, if you also want to protected your code against huge statements (long query strings, tons of bound values etc.) and a the same time make the printing customizable to some extent.

> Also I'm not comfortable with the breaking changes, even though they're on protected methods they're still breaking.

Yeah it sucks; but without them, `QueryLogger` cannot benefit from `StatementFormatter`. If that's that bad, then let's simply let `QueryLogger` out of this PR. The workaround in this case is to create another query logger, let's call it `EnhancedQueryLogger` and deprecate the old one.

> What are the things a user could reasonably expect from the logger and are currently not possible?

Here are a few ones:

1. Bound statements do not print their values (see JAVA-1256);
2. We currently do not print consistency level (see JAVA-1340);
3. Users cannot log custom statements (see JAVA-1273).

> We should provide a generic format that's reasonable for most users, not necessarily make everything configurable

Here I agree, you have a point. We might not need so many configurable settings, I will give a try today and simplify these parts to the bare minimum.

adutra:
My two last commits address feedback received so far:

- Revert changes to `QueryLogger` to avoid binary braking changes;
- Simplify `StatementFormatter` (get rid of some configurable settings to make the API less verbose)

driver-core/src/main/java/com/datastax/driver/core/EnhancedQueryLogger.java
```diff
@@ -0,0 +1,472 @@
...
+            this.slowQueryLatencyThresholdPercentile = slowQueryLatencyThresholdPercentile;
+            constantThreshold = false;
+            return this;
+        }
+
+        /**
+         * Sets the {@link StatementFormatter formatter} to use
+         * to format logged statements.
+         * <p/>
+         * If this methods is not called, then the
```
tolbertam: minor nit:

> If this methods is not called

should be

> if this method is not called

upgrade_guide/README.md
```diff
@@ -3,6 +3,19 @@
...
 versions of the Java driver.
 
+### 3.2.0
+
+[JAVA-1257](https://datastax-oss.atlassian.net/browse/JAVA-1257) has deprecated the `QueryLogger`
+class and users are now encouraged to use `EnhancedQueryLogger` instead, which makes it easier for
+users to customize the way statements are logged. 
+Please note that `QueryLogger` class might be removed in a future major release.
+
+[JAVA-1257](https://datastax-oss.atlassian.net/browse/JAVA-1257) also modifies the behavior
```
tolbertam, line 25: I don't think we ever documented the fact that it returned the Query string, but I'm glad you documented this here because it does sound like this could break somebody (even our tests assumed it worked this way).

The other possible concern is that `toString` will now be more intensive since it includes the parameter values and such.  If you have a ton of parameters that need to be serialized, repeated calls to `toString` could create some work - but I think that would be unusual.

manual/logging/README.md
```diff
@@ -113,78 +113,73 @@ and provide hints about what's going wrong.
...
+`EnhancedQueryLogger` [API docs][query_logger] for more information.
 
-Secondly, you need to adjust your logging framework to accept log messages from the `QueryLogger`. The `QueryLogger`
+Secondly, you need to adjust your logging framework to accept log messages from the `EnhancedQueryLogger`. The `EnhancedQueryLogger`
 uses 3 different loggers:
 
-* `com.datastax.driver.core.QueryLogger.NORMAL` : Used to log normal queries, i.e., queries that completed successfully within a configurable threshold in milliseconds.
-* `com.datastax.driver.core.QueryLogger.SLOW` : Used to log slow queries, i.e., queries that completed successfully but that took longer than a configurable threshold in milliseconds to complete.
-* `com.datastax.driver.core.QueryLogger.ERROR`: Used to log unsuccessful queries, i.e., queries that did not complete normally and threw an exception. Note this this logger will also print the full stack trace of the reported exception.
+* `com.datastax.driver.core.EnhancedQueryLogger.NORMAL` : Used to log normal queries, i.e., queries that completed successfully within a configurable threshold in milliseconds.
```
tolbertam: What do you think about reusing `QueryLogger` for the logger class?  I imagine that in next major release, we will simply get rid of the existing `QueryLogger` and rename `EnhancedQueryLogger` to `QueryLogger`.   If that is the case, why not reuse `QueryLogger` as the class that the logger uses?   I don't expect users to simultaneously want to configure both.   I could be otherthinking it :D

manual/logging/README.md
```diff
@@ -113,78 +113,73 @@ and provide hints about what's going wrong.
 
 ### Logging query latencies
 
-The `QueryLogger` provides clients with the ability to log queries
-executed by the driver, and especially, it allows client to track slow
+The `EnhancedQueryLogger` class provides clients with the ability to log queries
```
tolbertam, line 116: Should we add documentation for customizing what the query logger emits via `StatementFormatter` and using a custom`StatementPrinter`?

driver-core/src/main/java/com/datastax/driver/core/StatementFormatter.java
```diff
@@ -0,0 +1,1108 @@
...
+        private int maxInnerStatements = DEFAULT_MAX_INNER_STATEMENTS;
+        private int maxOutgoingPayloadEntries = DEFAULT_MAX_OUTGOING_PAYLOAD_ENTRIES;
+        private int maxOutgoingPayloadValueLength = DEFAULT_MAX_OUTGOING_PAYLOAD_VALUE_LENGTH;
+
+        private final List<StatementPrinter<?>> printers = new ArrayList<StatementPrinter<?>>();
+
+        private Builder() {
+        }
+
+        public Builder addStatementPrinter(StatementPrinter<?> printer) {
```
tolbertam: Would be good to add javadoc here explaining why a user would register a custom statement printer.

tolbertam (commented):
Looks great so far, just a few small comments.  Also I have a few tests I'd like to add (although the coverage is already very good!) which I will push in a bit.

upgrade_guide/README.md
adutra: >I don't think we ever documented the fact that it returned the Query string 

Unfortunately for us, [we did it](https://github.com/datastax/java-driver/blob/3.1.x/driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java#L192-L209) for `RegularStatement` :(

manual/logging/README.md
adutra: I was thinking about a blog post actually. I think we will need one for the new mapper features anyway.

adutra: Hmm you are right, maybe a small example in the docs would also be helpful. Will add that.

adutra:
Addressed feedback + rebased.

manual/logging/README.md
adutra: Sorry I won't have time today. Will address that on Monday.

driver-core/src/main/java/com/datastax/driver/core/StatementFormatter.java
```diff
@@ -0,0 +1,1138 @@
...
+            for (StatementPrinter<?> printer : printers) {
+                registry.register(printer);
+            }
+            StatementFormatterLimits limits = new StatementFormatterLimits(
+                    maxQueryStringLength, maxBoundValueLength, maxBoundValues, maxInnerStatements, maxOutgoingPayloadEntries, maxOutgoingPayloadValueLength);
+            return new StatementFormatter(registry, limits);
+        }
+
+        private void registerDefaultPrinters(StatementPrinterRegistry registry) {
+            registry.register(new DefaultStatementPrinter());
```
tolbertam: One small observation here which may not be that important:  If someone registers a custom `StatementPrinter` with a `getSupportedStatementClass()` that is a parent of any of the default printers' `Statement` classes, it will not be used because they are overridden by the more specific formatters.   For example, if they have a StatementPrinter for `RegularStatement` it wouldn't be used because it'd be overridden by the more specific default printers.  Do you think that is expected behavior?

In the general case, I would not expected users to specialize an already existing Statement that is handled, but if they wanted to they'd probably just have to register the printer for each special type for it to work.

adutra: > Do you think that is expected behavior?

It was, but I think your arguments are sound. Maybe we should make it so that custom printers have higher priority over bulilt-in ones.

tolbertam: Sounds good, I pushed 7e3d373 with a (currently disabled) test that exercises that behavior.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
```diff
@@ -293,25 +311,6 @@ public Boolean isIdempotent() {
...
     }
 
-    @Override
-    public String toString() {
-        try {
-            if (forceNoValues)
-                return getQueryString();
-            // 1) try first with all values inlined (will not work if some values require custom codecs,
-            // or if the required codecs are registered in a different CodecRegistry instance than the default one)
-            return maybeAddSemicolon(buildQueryString(null, CodecRegistry.DEFAULT_INSTANCE)).toString();
```
tolbertam, line 303: Just noticed there is a bizarre nuance here.  If we pass in `null` for the first parameter to `buildQueryString` it will inline the value in the query, so if you called `toString()` it would include the parameter in the query string, otherwise if you called `getQueryString()` it wouldn't and rather it would be a value.  This affects something like this (from `Jsr353JsonCodecTest`):

```
insertInto(TABLE1).value("k", "key0").value(quote("miXeD"), "a").value("i", 1).value("f", 1.1).getQueryString()
```

The generated query is: 

```
INSERT INTO test1 (k,"miXeD",i,f) VALUES (?,?,1,1.1);
```

instead of 

```
INSERT INTO test1 (k,"miXeD",i,f) VALUES ('key0','a',1,1.1);
```

However, the BuiltStatement does include the variables when you execute it, so perhaps that doesn't matter so much.

adutra: The old behavior was an attempt to include as much information as possible in the query string itself, as a means to make the resulting string reflect the statement's contents.

This behavior is not needed anymore, because even if the query string itself contains bind markers, the associated bound values are now printed along with the query string.

driver-core/src/main/java/com/datastax/driver/core/StatementFormatter.java
adutra: The test is now enabled and passes – although I had to remove inner statements from the batch statement because they were producing false positives.

adutra:
Squashed commits and rebased on top of current 3.x.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
```diff
@@ -112,6 +112,24 @@ public String getQueryString(CodecRegistry codecRegistry) {
...
+     * The number of values for this statement, or zero, if this
+     * statement does not have values.
+     * @param codecRegistry the codec registry that will be used if the statement must be
+     *                      rebuilt in order to determine if it has values, and Java objects
+     *                      must be inlined in the process (see {@link BuiltStatement} for
+     *                      more explanations on why this is so).
+     *
+     * @return the number of values.
+     */
+    public int valuesCount(CodecRegistry codecRegistry) {
```
adutra: Required to be able to print values of a `BuiltStatement`.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
adutra: We could even see this as a new advantage: now the query string included in the `toString()` result is the very same query string that would be executed if the statement was passed to `session.execute(Statement)`.

tolbertam: I think one thing that isn't clear to me is why are String values considered part of parameters where numbers are inlined?  Is there an intrinsic reason it was done this way?   I can see users encountering this and asking the same question.

tolbertam: I know that this behavior already previously existed (for the statement that was actually executed).  The other consideration is that parameters in simple statements are only supported in protocol V2+, although at this point I don't think we need to optimize for protocol V1.  I wonder why we don't just parameterize everything?

adutra: It took me a while to figure that out too. The reason is explained [here](https://github.com/datastax/java-driver/blob/3.1.x/driver-core/src/main/java/com/datastax/driver/core/querybuilder/Utils.java#L222-L224). Basically it is done so to not force users to give an exact type for the parameters, i.e. "bigint" would require longs, "varint" `BigInteger` etc. But once inlined, you could give any `Number` instance and it would work just fine.

tolbertam: Ah I see, that does make sense.   In the case of `String`, I guess it wouldn't matter if it was a parameter or not right?  I think it is fine to leave that the way it is, as there as some values you can't inline (collections for example), and changing it for the sake of changing it doesn't seem to provide much value.

tolbertam:
> Squashed commits and rebased on top of current 3.x.

In the future can you hold off on squashing until the review is complete?   It is hard to understand what the recent changes are since last reviewed otherwise (unless there is an easy way to do it, my git fu can be lacking :D).

adutra:
> In the future can you hold off on squashing until the review is complete?

Oh I'm so sorry :( I should have waited indeed.

tolbertam (approved):
Overall I think this looks very good.  The `StatementFormatter` / `StatementPrinter` API seems pretty easy to use and gives a large amount of flexibility.  I like that it also doesn't get in the way if a user doesn't need to use it.  👍

driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
olim7t: I also added a more detailed explanation [here](https://github.com/datastax/java-driver/blob/3.1.x/driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java#L36). Maybe not the best place, we can move it or link to it from elsewhere.

adutra:
Note to reviewers: I push one commit that adds some cosmetic changes to formatting patterns and reduces the verbosity of printed statements by abbreviating common terms such as "CL" (consistency level) or "IDP" (for idempotence). I can revert if you think it is not desirable.

newkek:
I was trying to sum up what are the needs for the different tickets that would be resolved by this PR. So, aside form the tickets that we can resolve simply without the `StatementFormatter` there's the ticket about bringing a customizable way to print statements.

So on the topic of providing a customizable way to print statements there seems to be 2 cases:
1. Printing a custom user-defined statement configurably.
2. Be able to configure the way the driver prints the built-in statement classes.

For 1), I believe users are in control of the class they implement, hence there is no more straightforward way, to print their own way their statement, than to just override `Statement#toString()`. I am trying to find what would be the advantage in usability to force the user to configure the QueryLogger + create a custom StatementFormatter + potentially a custom StatementPrinter + registering their Statement class in the StatementFormatter/registry, VS just overriding `toString()`. It seems like the StatementFormatter solution doesn't bring more usability from what I can see, but involves a much more complex knowledge required to do just a specialized formatting, plus overriding `toString()` when you want some custom formatting would seem rather natural I think for users. Also, if they override `toString()` they can benefit of the nice formatting, not just only in the QueryLogger, but in their application and in other places. All considering the QueryLogger calls `toString()` on statements when logging. One argument would be the re-usability of a StatementPrinter for different Statement classes (you register the same printer for several statement classes), but I'll get to that a little further down.

For 2), instead of having to understand what the QueryLogger, StatementFormatter, etc.. classes are for, they could just extend the driver Statement class, and override `toString()` as well.

Now I get that printing a Statement can be tricky as you mentioned before, the size of the query, all the different options the driver has (CL, tracing, etc..), the number of bound values, and so on. Hence I think the `StatementPrinter`(and `StatementWriter`) classes are useful, but do not necessarily need to be used via the StatementFormatter/PrinterRegistry/etc.. How about we would provide the StatementPrinter classes as standalone, so that users would be free to use them/expand them in `Statement#toString()` if they decide to override them. And the users would be free to reuse the same StatementPrinters in different Statement classes, and so on. And the driver uses the built-in printers in the default `toString` methods.

In terms of internal details/performance, that's more minor but I think using a Registry, and doing a look up of a Printer by taking the class of a given Statement class, is convoluted instead of just calling `#toString()`.

Wdyt?

adutra:
Let's not forget that the only practical way to create custom statements is to extend `StatementWrapper` (extending `SimpleStatement` should work too but I never tried). But unfortunately, `StatementWrapper.toString()` will never be called because of [this line](https://github.com/avivcarmis/java-driver/blob/08e6fbb6155723a4d5d889518d11f58f2db2f152/driver-core/src/main/java/com/datastax/driver/core/QueryLogger.java#L618).

So the `toString()` solution is not always applicable.

But to be honest, the more it goes the less I like this PR for the same reasons as you, so I'm starting to agree that this shouldn't be merged.

newkek:
I took a little look at the line you mentioned, sounds indeed like we need to unwrap the statement, because later on in [logQuery](https://github.com/avivcarmis/java-driver/blob/08e6fbb6155723a4d5d889518d11f58f2db2f152/driver-core/src/main/java/com/datastax/driver/core/QueryLogger.java#L644) there's some logic regarding the kind of Statement that needs to be logged, right? But if we keep that logic in specific `StatementPrinters` inside `toString()` of the Statement implementations, I think we don't need to do a case by case in `logQuery`. So the unwrapping can be removed in the `QueryLogger`, and `StatementWrapper#toString` just calls by default `wrapped.toString()`. A little like what's happening in `EnhancedQueryLogger` which currently delegates the formatting to `StatementFormatter` but here instead we'd just call `statement.toString()`.

PR #760, opened 2016-10-18
JAVA-262: Make internal executors customizable

olim7t:


driver-core/src/main/java/com/datastax/driver/core/ThreadingOptions.java
```diff
@@ -0,0 +1,143 @@
...
+     * Builds the generic administration executor, used for tasks such as triggering registered
+     * {@link SchemaChangeListener}s, reacting to node state changes, and metadata updates.
+     * <p/>
+     * The default implementation sets the pool size to the number of available cores.
+     *
+     * @param clusterName the name of the cluster, as specified by
+     *                    {@link com.datastax.driver.core.Cluster.Builder#withClusterName(String)}.
+     * @return the executor.
+     */
+    public ThreadPoolExecutor executor(String clusterName) {
```
adutra: Couldn't we give it a more meaningful name? Suggestions: `mainExecutor`, `mainWorkerPool` etc...

driver-core/src/main/java/com/datastax/driver/core/ThreadingOptions.java
```diff
@@ -0,0 +1,143 @@
...
+                // an optimization around ThreadLocals (we could use DefaultThreadFactory directly but it creates
+                // slightly different thread names, so keep we keep a ThreadFactoryBuilder wrapper for backward
+                // compatibility).
+                .setThreadFactory(new DefaultThreadFactory("ignored name"))
+                .build();
+    }
+
+    /**
+     * Builds the generic administration executor, used for tasks such as triggering registered
+     * {@link SchemaChangeListener}s, reacting to node state changes, and metadata updates.
```
adutra: It is also used for speculative executions and retries (before JAVA-893), not sure it's worth mentioning.

driver-core/src/main/java/com/datastax/driver/core/ThreadingOptions.java
```diff
@@ -0,0 +1,143 @@
...
+                // Back with Netty's thread factory in order to create FastThreadLocalThread instances. This allows
+                // an optimization around ThreadLocals (we could use DefaultThreadFactory directly but it creates
+                // slightly different thread names, so keep we keep a ThreadFactoryBuilder wrapper for backward
+                // compatibility).
+                .setThreadFactory(new DefaultThreadFactory("ignored name"))
+                .build();
+    }
+
+    /**
+     * Builds the generic administration executor, used for tasks such as triggering registered
```
adutra: Nit: "generic administration" sounds a bit like a bag for any kind of tasks, I'd suggest "main internal executor" or something in the style.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1362,16 +1374,17 @@ synchronized void init() {
...
-            this.reconnectionExecutor = new ScheduledThreadPoolExecutor(2, threadFactory("reconnection"));
-            // scheduledTasksExecutor is used to process C* notifications. So having it mono-threaded ensures notifications are
-            // applied in the order received.
-            this.scheduledTasksExecutor = new ScheduledThreadPoolExecutor(1, threadFactory("scheduled-task-worker"));
-
-            this.reaper = new ConnectionReaper(this);
+            ThreadingOptions threadingOptions = this.configuration.getThreadingOptions();
+            ThreadPoolExecutor tmpExecutor = threadingOptions.executor(clusterName);
+            this.executorQueue = tmpExecutor.getQueue();
+            this.executor = MoreExecutors.listeningDecorator(tmpExecutor);
```
adutra, line 1374: I think that ideally, `ThreadingOptions` should return `Listening(Scheduled)ExecutorService` instances. Returning `(Scheduled)ThreadPoolExecutor` sounds like a half-way solution. To give users total flexibility, we need to be able to operate on any `ExecutorService` regardless of the implementation. 

But if I follow you correctly, you need `ThreadPoolExecutor` instances so that you can access their underlying queue.

However such queues seem only used for metrics anyway, so I'd suggest that we simply stop monitoring these altogether, on the grounds that now that they are customizable, it's up to the user to monitor them. Alternatively, we could expose new methods in `ThreadingOptions` that would return the queues to monitor, but to me it seems overkill.

driver-core/src/main/java/com/datastax/driver/core/ThreadingOptions.java
```diff
@@ -0,0 +1,143 @@
...
+
+import java.util.concurrent.*;
+
+/**
+ * A set of hooks that allow clients to customize the driver's internal executors.
+ * <p/>
+ * The methods in this class are invoked when the cluster initializes. To customize the behavior, extend the class and
+ * override the appropriate methods.
+ * <p/>
+ * This is mainly intended to allow customization and instrumentation of driver threads. Keep the executors separate, as
```
adutra: Nit: I would be even more explicit here and state stg like "each executorX method must return a newly-allocated thread pool; shared thread pools should be avoided as these could introduce etc.."

adutra: BTW I wonder if all methods in this class shouldn't be prefixed with `create` or `new` to make it crystal clear that each invocation should return a newly-allocated object.

driver-core/src/main/java/com/datastax/driver/core/ThreadingOptions.java
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * Builds the executor used to block on new connections before they are added to a pool.
+     * <p/>
+     * The default implementation uses 2 threads.
+     *
+     * @param clusterName the name of the cluster, as specified by
+     *                    {@link com.datastax.driver.core.Cluster.Builder#withClusterName(String)}.
+     * @return the executor.
+     */
+    public ThreadPoolExecutor blockingExecutor(String clusterName) {
```
adutra: Same here, the name is a bit vague, why not `connectionCreationExecutor`?

driver-core/src/main/java/com/datastax/driver/core/ThreadingOptions.java
```diff
@@ -0,0 +1,143 @@
...
+     * @param clusterName the name of the cluster, as specified by
+     *                    {@link com.datastax.driver.core.Cluster.Builder#withClusterName(String)}.
+     * @return the executor.
+     */
+    public ScheduledThreadPoolExecutor reconnectionExecutor(String clusterName) {
+        return new ScheduledThreadPoolExecutor(2, threadFactory(clusterName, "reconnection"));
+    }
+
+    /**
+     * Builds the executor to handle host state notifications from Cassandra.
```
adutra, line 119: It's also used for debouncing of such notification events, and for idle connections cleanup (again, not sure it's worth mentioning)

adutra (changes requested):
Very good work, the only important topic to address imho is method signatures in `ThreadingOptions`.

driver-core/src/main/java/com/datastax/driver/core/ThreadingOptions.java
tolbertam: I like that the names are 1:1 with the names in `Cluster.manager`.  I expect the people who will use this will will need to look at the source code anyways so it makes sense to sync them up.  I think in a future major release where we refactor our thread pools I agree it would be good to make the names more explicit.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
tolbertam: I think the need for monitoring thread pool queues is even greater now that the users have control over configuring them, so I think it would be good to keep it (the cost is low anyways).   Also think we should not remove the metrics in a patch release as it could break some users existing monitoring (i.e. if they were alerting on these metrics, although that likelihood seems low).

driver-core/src/main/java/com/datastax/driver/core/ThreadingOptions.java
```diff
@@ -0,0 +1,143 @@
...
+ *   limitations under the License.
+ */
+package com.datastax.driver.core;
+
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+import io.netty.util.concurrent.DefaultThreadFactory;
+
+import java.util.concurrent.*;
+
+/**
```
tolbertam, line 23: It would be good to use strong language discouraging the use of this, maybe indicate it is for expert use only as the default options will be sensible for most cases.  Sometimes users add configuration for everything just because the option to is there, but this is one of those where they should not be overriding behavior unless they have a really good reason.

driver-core/src/main/java/com/datastax/driver/core/ThreadingOptions.java
olim7t: Yes, I thought about changing the names but I agree it's better to stay in line with the code and especially the metrics (which we can't change).

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
olim7t: As discussed offline, I agree that we could use `*Service` interfaces, and `instanceof` checks and casts internally to access the queues.
I don't see any value in exposing the Guava wrappers. Implementations may choose to return them, when we call `MoreExecutors.listeningDecorator` internally it will be a no-op.

driver-core/src/main/java/com/datastax/driver/core/ThreadingOptions.java
olim7t: I'm a bit wary of "expert use" comments in general. It implies that we divide users into two groups, but how does someone know if they have the "expert" badge?
I'd rather say something along the lines of "things will break badly if this is misconfigured".

tolbertam: Maybe 'expert use only' isn't the right verbiage, but I think it would be good to have something strong and delineating this configuration from others as users shouldn't be using this without good reason.  `things will break badly if this is misconfigured` sounds good to me.

driver-core/src/main/java/com/datastax/driver/core/Metrics.java
```diff
@@ -262,6 +251,24 @@ void shutdown() {
             jmxReporter.stop();
     }
 
+    private static Gauge<Integer> buildQueueSizeGauge(final ExecutorService executor) {
+        if (executor instanceof ThreadPoolExecutor) {
```
tolbertam: I think this doesn't work for executor and blockingExecutor because `listeningDecorator` returns a `ListeningDecorator` which isn't an instance of `ThreadPoolExecutor`.  Noticed that it was returning `-1` for `blocking-executor-queue-depth` and `executor-queue-depth`

tolbertam (approved):
looks good to me 👍

adutra (approved):

adutra:
I allowed myself to add 3 commits on top of the branch, hopefully consensual enough not to require a full review:
1. c4dd6c0 simply removes unused fields;
2. b40a28d aligns `Metrics`javadocs with those of `ThreadingOptions`;
3. 34ffb3e Homogenizes the way gauges are created for executor queues.

Feel free to squash or even revert at your discretion.

tolbertam:
The test I introduced is not working on Windows for whatever reason.  Will fix this shortly.

tolbertam (approved):
test should be fixed now and recent changes look good 👍

PR #761, opened 2016-10-18
JAVA-1303: Add missing BoundStatement.setRoutingKey(ByteBuffer...)

olim7t:


adutra (approved):

tolbertam (approved):
This looks good to me.  I also added a small test for validating that an explicitly provided routing key is used with `BoundStatement` in `TokenAwarePolicy`.

PR #762, opened 2016-10-19
Use unsafesystem by default to speed up schema related tests

tolbertam:
From [this thread](http://www.mail-archive.com/user@cassandra.apache.org/msg49419.html) we learned about the `-Dcassandra.unsafesystem` system property introduced in 1.2.7 by [CASSANDRA-5704](https://issues.apache.org/jira/browse/CASSANDRA-5704) which forgoes a non-fast blocking flush of system tables to speed up schema changes.   For practical use this is not an option one shold use, but for the intents of our testing it should speed up tests that make and wait on schema changes.

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
@@ -228,7 +228,12 @@ public class CCMBridge implements CCMAccess {
         this.thriftPort = thriftPort;
         this.binaryPort = binaryPort;
         this.isDSE = isDSE;
-        this.jvmArgs = jvmArgs;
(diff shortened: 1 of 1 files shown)


adutra (commented):
I think you also need to modify schema creation statements to use `durable_writes=false` in order to fully benefit from this improvement, especially those created with the template in `TestUtils`.

tolbertam:
> I think you also need to modify schema creation statements to use durable_writes=false in order to fully benefit from this improvement, especially those created with the template in TestUtils.

👍 , I tried that and that seemed to speed things up as well (although not as pronounced).  From offline discussion we've decided to hold off on this for now since the order of magnitude of difference (100-200ms in best case) is not big enough to justify the change.  I think it may help in our windows CI since the schema tests always seems to be less reliable, but i'll look into that later.

PR #763, opened 2016-10-20
JAVA-1241: Updated netty to 4.1.6.Final

alexbool:


olim7t:
Hi,
Is there a specific feature in 4.1.6 that you're interested in?

alexbool:
No, just updating to prevent dll hell with other piece of software that is using `netty` (`async-http-client` in particular).

yvladimirov:
@olim7t 
Hi, new ElasticSearch 5 version used  netty 4.1.6 and haved conflicted with cassandra java driver

tolbertam:
@yvladimirov the java-driver has a 'shaded' configuration ([doc](https://datastax.github.io/java-driver/manual/shaded_jar/)) that embeds and shades netty under a different package name in the driver jar.

This would allow you to continue using the driver in the case where the version of netty that your application or another dependency has conflicts with the driver's netty dependency version.  Would that work in the meantime?

alexbool:
@tolbertam This could work, but in this case we cannot benefit from using native transport, which can lead to performance issues that we have not yet evaluated.
By the way, what is the state of this PR? What do I need to do to get it reviewed?

tolbertam:
> but in this case we cannot benefit from using native transport, which can lead to performance issues that we have not yet evaluated.

I haven't myself seen a measured benefit of the epoll module yet for the driver in particular ([some discussion here](https://groups.google.com/a/lists.datastax.com/d/msg/java-driver-user/QgwtAB7Mhn8/l1zEpYAZNb8J)), although it's been quite a while since i've taken some time to look at it and run some comparison tests (and look at the garbage profile in particular).  I imagine that you won't notice a difference unless running with very large throughputs (> 100k req/s).

> By the way, what is the state of this PR? What do I need to do to get it reviewed?

I think this would be a great consideration for the next minor release (i.e. 3.2.x), although i'm not sure when that will be.    Really it is just a matter of being accepted and going through our test cycle.  I'll give it a run through of our CI this afternoon and get back to you just so we have some initial confidence that this change won't cause any glaring issues.  I see the [appveyor build](https://ci.appveyor.com/project/DataStax/java-driver/build/610.3.x) (windows) failed an SSL test, likely because it uses the netty-tcnative-openssl module and it's possible that netty 4.1.x needs a newer version of libapr1.

I would prefer not to put this in a patch release (i.e. 3.1.3) because we should be limiting the scope of those changes to fixes and minor changes to keep patch releases predictable for users as there may be some backwards compatibility issues with netty 4.1.x and 4.0.x that we aren't aware of.

tolbertam:
It looks like CI ran well, the only exception being that we'll need to update our netty-tcnative dependency in our tests and also update our OSGI test to depend on netty 4.1.  Would you mind giving me access to update this branch so I can make the appropriate changes?  Otherwise I could create another PR.

tolbertam:
I think 19fbb5563 will fix the test issues if you wouldn't mind cherry-picking it.  I'm running things through CI now and if it looks good (if [this passes](https://ci.appveyor.com/project/DataStax/java-driver/build/637.netty416test/job/udgqalit9qv7uhfx) we're good) I think this is good to merge to the 3.x branch (if everyone else is comfortable with it).

alexbool:
Thanks for your help, I cherry-picked you commit.

alexbool:
AppVeyor failed with some strange errors

tolbertam:
Indeed it looks like there's still some work that needs to be done on the OSGi tests, i'll fix that up, thanks!

olim7t:
> I would prefer not to put this in a patch release (i.e. 3.1.3)

Agreed, let's target this for 3.2.0.
@yvladimirov if you have an urgent need for this, you can also try to override the version in your application to force the driver to use Netty 4.1. Given that there are no code changes it should work, but it's at your own risk until it's fully tested.

tolbertam:
It looks like the OSGi tests issues were around `netty-resolver`, that `netty-transport` now depends on in 4.1.  Adding this bundle to our tests fixes the issue (commit ad24dc4).

alexbool:
@tolbertam Thanks, cherry-picked!

adutra:
Netty 4.1.8 has been released on Jan 30, we should probably use that instead.

adutra:
Closing in favor of #800 . Thanks @alexbool !

PR #764, opened 2016-10-24
JAVA-1320: prevent unnecessary task creation on empty pool

olim7t:


tolbertam (approved):
looking great!  this appears to have fixed the issue 👍

PR #765, opened 2016-10-26
JAVA-1321: Wrong OSGi dependency version for Guava

emolsson:
Changed required Guava version back to 16.0

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -1,5 +1,10 @@
 ## Changelog
 
+### 3.1.1 (in progress)
+
+- [bug] JAVA-1321: Wrong OSGi dependency version for Guava.
+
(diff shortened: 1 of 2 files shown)


tolbertam:
Good catch!  I think the minimum requirement should be guava 16.0.1 though since the java-driver requires it to work around a `TypeToken` bug.

emolsson:
Sure, I'll change it to ${guava.version} to make sure that it's always set to the Guava version in the top pom.

olim7t:
Good catch, that was an oversight when I merged some 2.1 tickets for 3.1.
Cherry-picked on 3.1.x as 108ddca, it will be in 3.1.2.

PR #768, opened 2016-11-17
JAVA-1334: Provide the counterpart of the method addContactPoints that silently fail when a host is unknown

essobedo:
Fix for https://datastax-oss.atlassian.net/browse/JAVA-1334

This PR simply provides the ability to add only the contact points that can be resolved:

- If none can be resolved then an `IllegalArgumentException` will be thrown 

- For each contact point that cannot be resolved a warning will be logged

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### 3.2.0
+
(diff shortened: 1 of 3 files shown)


essobedo:
Close it as rejected by the Project Lead

PR #769, opened 2016-11-21
JAVA-1328: Upgrade guava 16.0.1 -> 20.0

mspangdal:


mspangdal:
SSL failures, not really sure how to interpret the failures though

adutra:
Thanks @mspangdal for contributing this, it is very helpful. Unfortunately, we cannot upgrade right now to 20.0 as a lot of people out there are stuck with older versions (especially the devs for the Spark-Cassandra connector). Read [JAVA-1328](https://datastax-oss.atlassian.net/browse/JAVA-1328) for more details and please vote for this ticket if you need the upgrade ASAP.

dn-mib:
Any progress on this? How would this be affected by the next version of Guava that will be Java 8-only if I have understood it correctly?

tolbertam:
> Any progress on this? How would this be affected by the next version of Guava that will be Java 8-only if I have understood it correctly?

Based on @olim7t's comments in [JAVA-1328](https://datastax-oss.atlassian.net/browse/JAVA-1328), there's an approach to support both >= 20 and < 20 versions of guava, but it is not trivial.  My opinion is that we don't try to support both simultaneously and make it a requirement for the next major version of the driver to shade guava and reduce the dependency on it (made easier if we require java 8).  Of course this depends on the timeline of our next major version and need from the community to have a support path for 20 (please vote on that ticket and add comments if you need it).

olim7t:
I've opened #784 to explore the idea I suggested on the Jira ticket. It supports both pre- and post-19 versions.
Sorry for creating a separate PR, but I wanted to base it on 3.0.x.

adutra:
@mspangdal @olim7t can we close this in favor of #784 ?

mspangdal:
Sure, no problem!

PR #772, opened 2016-12-03
Improve mapping manager synchronization

avivcarmis:
OK so following [this discussion](https://groups.google.com/a/lists.datastax.com/forum/#!topic/java-driver-user/i87MpKNw8l0), I've used `ConcurrentHashMap::putIfAbsent` to prevent an illegal state where two different instances of same mapper are created and returned.
Since this issue is non-deterministic (depends on execution timing to be reproduced) i don't really have a way to test it.

There are two issues with my implementation.
1) Since we can't use JDK8 interface, i can't use `ConcurrentHashMap`'s `computeIfAbsent`, which means that prior to the first put call there might be some redundant creations of instances, the good news is, they will never be returned to the user so at least the original issue is solved.
2) The logic of `SchemaChangeListenerBase` is now not synchronized with creation of instances, this means that there can be a scenario where a table is removed and the error message is not logged.

To solve both issues, we can revert back to regular `HashMap`, previous implementation, but create a dedicated final lock for each map, this way we get all we need. Down side is we implement synchronization ourselves, which means we probably get lower performance and might have concurrency issues. (Same as earlier actually)

Let me know what you think...

### DIFF
--- a/driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java
+++ b/driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java
@@ -23,6 +23,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Mapping manager from which to obtain entity mappers.
@@ -34,9 +35,9 @@ public class MappingManager {
     private final Session session;
     final boolean isCassandraV1;
 
-    private volatile Map<Class<?>, Mapper<?>> mappers = Collections.emptyMap();
-    private volatile Map<Class<?>, MappedUDTCodec<?>> udtCodecs = Collections.emptyMap();
-    private volatile Map<Class<?>, Object> accessors = Collections.emptyMap();
+    private final ConcurrentHashMap<Class<?>, Mapper<?>> mappers = new ConcurrentHashMap<Class<?>, Mapper<?>>();
+    private final ConcurrentHashMap<Class<?>, MappedUDTCodec<?>> udtCodecs = new ConcurrentHashMap<Class<?>, MappedUDTCodec<?>>();
+    private final ConcurrentHashMap<Class<?>, Object> accessors = new ConcurrentHashMap<Class<?>, Object>();
 
     /**
      * Creates a new {@code MappingManager} using the provided {@code Session}.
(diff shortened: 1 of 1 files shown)


adutra:
@avivcarmis would you mind rebasing this on top of 3.x? Otherwise I can take care of it.

driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java
```diff
@@ -218,57 +219,33 @@ public Session getSession() {
...
-                    mappers = newMappers;
-                }
-            }
+        if (!mappers.containsKey(klass)) {
+            EntityMapper<T> entityMapper = AnnotationParser.parseEntity(klass, ReflectionMapper.factory(), this);
+            Mapper<T> mapper = new Mapper<T>(this, klass, entityMapper);
+            mappers.putIfAbsent(klass, mapper);
         }
-        return mapper;
+        return (Mapper<T>) mappers.get(klass);
```
adutra, line 227: I might be nitpicking but I think the following would be still more efficient:

```java
        Mapper<T> mapper = (Mapper<T>) mappers.get(klass);
        if (mapper == null) {
            EntityMapper<T> entityMapper = ...;
            mapper = new Mapper<T>(this, klass, entityMapper);
            Mapper<T> old = (Mapper<T>) mappers.putIfAbsent(klass, mapper);
            if (old != null)
                mapper = old;
        }
        return mapper;
```
... because in the nominal case (i.e. there is already an entry in the map) there would be just one access to the map instead of two.

adutra (changes requested):

driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java
avivcarmis: Yeah, i'll change it.
I'm actually re-thinking that whole PR, these entire parts ([1](https://github.com/datastax/java-driver/blob/41d3d72c23f149b13303074046f72877461a3cd7/driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java#L82), [2](https://github.com/datastax/java-driver/blob/41d3d72c23f149b13303074046f72877461a3cd7/driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java#L96), etc...) may not work.

I think the safest tweak will be to go back to the [original ](https://github.com/datastax/java-driver/blob/3.x/driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java#L37) implementation and add
`final Object lock = new Object()`  for each of those maps and synchronize them when constructing. Less pretty, but less error prone i guess.
What do you think?

avivcarmis: @adutra

adutra: @avivcarmis `onTableRemoved` and `onTableChanged` should work on a best-effort basis. Honestly, if people out there are crazy enough to alter their schema while clients are issuing requests, it's their problem. I would actually suggest that we get rid of the `synchronized` blocks in the places you spotted and clearly specify in the documentation that concurrent modifications of the schema could end up in unexpected results.

avivcarmis:
Done.
it was a minor commit so it was easier to implement on top of 3.x.
Opened a [new PR](https://github.com/datastax/java-driver/pull/803)

PR #773, opened 2016-12-09
JAVA-1351: Include Custom Payload in Request.copy

tolbertam:
For [JAVA-1351](https://datastax-oss.atlassian.net/browse/JAVA-1351)

Motivation:

When a request is retried or speculatively executed, a new Request
instance is created from the existing one using Request.copy.  This has
a side effect of not carrying over custom payload from the previous
request.  This change ensures Request.copy carries over custom payload
from the previous Request if present.

Modifications:

Renamed existing copy implementations to copyInternal and introduced new
copy methods that call copyInternal and then setCustomPayload to carry
over custom payload from previous Request.

Result:

Custom payload is now carried over on retries and speculative
executions.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Message.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Message.java
@@ -180,9 +180,21 @@ abstract class Message {
             }
         }
 
-        abstract Request copy();
+        Request copy() {
+            Request request = copyInternal();
+            request.setCustomPayload(this.getCustomPayload());
+            return request;
+        }
+
+        protected abstract Request copyInternal();
 
         Request copy(ConsistencyLevel newConsistencyLevel) {
+            Request request = copyInternal(newConsistencyLevel);
+            request.setCustomPayload(this.getCustomPayload());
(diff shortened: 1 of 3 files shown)


driver-core/src/main/java/com/datastax/driver/core/Message.java
```diff
@@ -180,9 +180,21 @@ ByteBuffer pagingState() {
...
         }
 
-        abstract Request copy();
+        Request copy() {
+            Request request = copyInternal();
+            request.setCustomPayload(this.getCustomPayload());
+            return request;
+        }
+
+        abstract Request copyInternal();
```
tolbertam: Not sure if this is the best approach, but seemed like a good way to carry over custom payload globally for each request.  Also not sure if `copyInternal` is the best name, would be good to change if can think of a better one.

olim7t: I can't think of a better name either. If we make the methods protected it's acceptable.

olim7t (approved):
👍

PR #774, opened 2016-12-12
JAVA-1346: Reset heartbeat only on client reads (not writes)

adutra:


tolbertam:
looks good, have a test for this, will push up shortly.

tolbertam:
Added a test, fails on 3.0.x but passes on this branch, feel free to squash :) 👍

tolbertam (approved):

olim7t (approved):
LGTM
Nit: could we update the title to describe the actual change, rather than the initial problem? Suggestion: "Reset heartbeat only on client reads (not writes)"

adutra:
Squashed and rebased, should be ready to go.

PR #776, opened 2016-12-22
JAVA-1347: Add support for duration type

adutra:


driver-core/src/main/java/com/datastax/driver/core/DataType.java
```diff
@@ -759,4 +772,16 @@ public String toString() {
             return String.format("'%s'", customClassName);
         }
     }
+
+    /**
+     */
+    public static class DurationType extends CustomType {
```
olim7t, line 784: Just a reminder to either fill in or remove the javadoc comment

driver-core/src/main/java/com/datastax/driver/core/Duration.java
```diff
@@ -0,0 +1,587 @@
...
+     * In this case all the non zero values must be negatives.
+     *
+     * @param months      the number of months
+     * @param days        the number of days
+     * @param nanoseconds the number of nanoseconds
+     */
+    private Duration(int months, int days, long nanoseconds) {
+        // Makes sure that all the values are negatives if one of them is
+        assert (months >= 0 && days >= 0 && nanoseconds >= 0)
+                || ((months <= 0 && days <= 0 && nanoseconds <= 0));
```
olim7t: We should throw an `IllegalArgumentException` rather than using an assertion, this is called directly from `newInstance` which is public.

driver-core/src/main/java/com/datastax/driver/core/Duration.java
```diff
@@ -0,0 +1,587 @@
...
+
+        this.months = months;
+        this.days = days;
+        this.nanoseconds = nanoseconds;
+    }
+
+    public static Duration newInstance(int months, int days, long nanoseconds) {
```
olim7t, line 99: Needs a javadoc with an explanation on positive/negative arguments.

driver-core/src/main/java/com/datastax/driver/core/Duration.java
```diff
@@ -0,0 +1,587 @@
...
+     * <li>{@code us} or {@code µs}: microseconds</li>
+     * <li>{@code ns}: nanoseconds</li>
+     * </ul>
+     * </li>
+     * <li>ISO 8601 format:  P[n]Y[n]M[n]DT[n]H[n]M[n]S or P[n]W</li>
+     * <li>ISO 8601 alternative format: P[YYYY]-[MM]-[DD]T[hh]:[mm]:[ss]</li>
+     * </ul>
+     *
+     * @param input the <code>String</code> to convert
+     * @return a number of nanoseconds
```
olim7t: Returns a Duration rather.

driver-core/src/main/java/com/datastax/driver/core/Duration.java
```diff
@@ -0,0 +1,587 @@
...
+        } else if (s.equals("s")) {
+            return builder.addSeconds(number);
+        } else if (s.equals("ms")) {
+            return builder.addMillis(number);
+        } else if (s.equals("us") || s.equals("µs")) {
+            return builder.addMicros(number);
+        } else if (s.equals("ns")) {
+            return builder.addNanos(number);
+        }
+        throw new InvalidTypeException(String.format("Unknown duration symbol '%s'", symbol));
```
olim7t: I wonder if `InvalidTypeException` is the best error to throw here, it's closely related to the codec system but this class is not. Maybe `IllegalArgumentException` would be more appropriate.

driver-core/src/main/java/com/datastax/driver/core/DataType.java
```diff
@@ -759,4 +772,16 @@ public String toString() {
...
+
+        private static final DurationType instance = new DurationType();
+
+        private DurationType() {
+            super(Name.CUSTOM, "org.apache.cassandra.db.marshal.DurationType");
+        }
```
olim7t, line 790: Nit: we could override `toString()` here, as this produces a slightly better output, for example in `TableMetadata.toString()` (although the full custom type name is valid and entirely compatible).

driver-core/src/main/java/com/datastax/driver/core/DataType.java
```diff
@@ -503,6 +503,19 @@ public static CollectionType frozenMap(DataType keyType, DataType valueType) {
...
+     * Returns a Duration type, introduced in Cassandra 3.10.
+     * <p/>
+     * Note that a Duration type does not have a native representation in CQL, and
+     * technically, is merely a special {@link DataType#custom(String) custom type}
+     * from the driver's point of view.
+     *
+     * @return the Duration type.
+     */
+    public static DurationType duration() {
+        return DurationType.instance;
```
olim7t, line 517: We might want to add a special case in `DataType.decode()` to return this singleton as well.

Currently it returns a `CustomType` instance that is not a `DurationType`. It appears for example in the column definitions of a prepared statement.

olim7t (changes requested):

driver-core/src/main/java/com/datastax/driver/core/DataType.java
adutra: Good catch, thanks.

tolbertam (approved):
Looks good to me!  I updated DurationIntegrationTest to test with some sample values defined in the unit tests just to get a wider range of values tried at the integration level.  I also tried testing using Durations as a bind parameter for comparison operations w/ [CASSANDRA-11936](https://issues.apache.org/jira/browse/CASSANDRA-11936) (i.e. `WHERE reading_time < now() - ?`) but that doesn't seem to work yet on the C* side.

driver-core/src/main/java/com/datastax/driver/core/DataTypeCqlNameParser.java
```diff
@@ -65,6 +65,8 @@
                     .put("timeuuid", timeuuid())
                     .put("tinyint", tinyint())
                     .put("smallint", smallint())
+                    // duration is not really a native CQL type, but appears as so in system tables
+                    .put("duration", duration())
```
stamhankar999, line 69: It turns out it will be in the system tables in C* 4.x (v5 protocol), but is a custom type in v4 protocol.

driver-core/src/test/java/com/datastax/driver/core/DurationCodecTest.java
```diff
@@ -0,0 +1,92 @@
...
+import static com.datastax.driver.core.Duration.*;
+import static com.datastax.driver.core.ProtocolVersion.V4;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class DurationCodecTest {
+
+    @Test(groups = "unit")
+    public void testFromStringWithStandardPattern() {
+        assertCodec("1y2mo", Duration.newInstance(14, 0, 0));
+        assertCodec("-1y2mo", Duration.newInstance(-14, 0, 0));
```
stamhankar999: It is actually illegal to mix positive and negative attributes for a C* Duration. C* doesn't complain about it when you use bound parameters, but you can't feed literals like this in cql. Ben Lerer suggested that drivers validate the data because it is expensive for C* to validate on its end (though he said he'll mull over an efficient way to reject server-side).

driver-core/src/main/java/com/datastax/driver/core/DataTypeCqlNameParser.java
tolbertam: Nice, I remember that it was a similar story with UDTs being a custom type between a later version of 2.0 and 2.1 (with protocol v3).  I'm glad that is the case.

olim7t (approved):

PR #777, opened 2016-12-22
JAVA-1355: Update getReplicas to handle range covering multiple hosts

tolbertam:
For [JAVA-1355](https://datastax-oss.atlassian.net/browse/JAVA-1355):

Motivation:

Metadata.getReplicas(String, TokenRange) currently assumes that the
TokenRange provided represents a TokenRange for a single host.  Because
of this, it finds the replicas for the last Token on that range instead
of considering the entire range.

Arbitrary TokenRanges may be provided that span multiple hosts.  In
this case, this method should include each host spanning this range
and their replicas.

Modifications:

Updated method to evaluate which Host's TokenRanges intersect with the
input TokenRange and include those range's replicas.

Result:

Metadata.getReplicas(String TokenRange) now includes replicas for all
hosts covering a range, not just the replicas for the last token in the
range.

Targeted this against 3.0.x & 3.0.6, but wasn't sure if that was appropriate.  It could also be debated as to whether this is a bug fix or a functionality change, which could effect what target release could be.  If we decide to leave as is, I think it would be good to update the javadoc to make it clear that the method assumes the range only covers 1 host.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Metadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Metadata.java
@@ -18,6 +18,7 @@ package com.datastax.driver.core;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -321,8 +323,25 @@ public static boolean isReservedCqlKeyword(String id) {
         if (current == null) {
             return Collections.emptySet();
         } else {
-            Set<Host> hosts = current.getReplicas(keyspace, range.getEnd());
-            return hosts == null ? Collections.<Host>emptySet() : hosts;
+            Set<Host> replicas = Sets.newHashSet();
+            Set<Host> allHosts = getAllHosts();
+            // for each host's range, if it intersects with the input range, include that range's replicas.
+            // this is a bit repetitive as we may evaluate the same ranges multiple times.
```
tolbertam, line 329: As documented, this is not ideal since we could evaluate the same range multiple times, but couldn't find a direct way to get at a host's primary ranges.

tolbertam:
Closing this, will rebase on 3.2.x and can consider it for that release, will also open up a separate PR that just changes the docs on 3.0.x

PR #781, opened 2017-01-06
Change SSLOptions API to take a remote address

wimtie:
Enable creation of SSLHandler with remote address information. This is
information set, which is needed to enable hostname verification. We
cannot take this information from SocketChannel at this time since it is
not connected yet at the time Connection.Initializer creates the
SSLHandler.


wimtie:
Forgot to change clirr settings, will re-open new PR.

PR #782, opened 2017-01-06
JAVA-1364: Enable creation of SSLHandler with remote address information

wimtie:
Enable creation of SSLHandler with remote address information. This is
information set, which is enables doing hostname verification (for example
through a custom TrustManager). We cannot take this information from SocketChannel
since it is not connected yet at the time Connection.Initializer creates the
SSLHandler.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Connection.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Connection.java
@@ -1311,7 +1311,11 @@ class Connection {
             ChannelPipeline pipeline = channel.pipeline();
 
             if (sslOptions != null) {
-                pipeline.addLast("ssl", sslOptions.newSSLHandler(channel));
+                if (sslOptions instanceof AddressAwareSSLOptions) {
+                    pipeline.addLast("ssl", ((AddressAwareSSLOptions)sslOptions).newSSLHandler(channel, connection.address));
+                } else {
+                    pipeline.addLast("ssl", sslOptions.newSSLHandler(channel));
+                }
             }
 
 //            pipeline.addLast("debug", new LoggingHandler(LogLevel.INFO));
--- a/driver-core/src/main/java/com/datastax/driver/core/AddressAwareSSLOptions.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/AddressAwareSSLOptions.java
@@ -22,26 +22,17 @@ import io.netty.handler.ssl.SslHandler;
 
 /**
  * Defines how the driver configures SSL connections.
- *
- * @see JdkSSLOptions
- * @see NettySSLOptions
+ * 
+ * Child interface to SSLOptions with the possibility to pass remote endpoint data
+ * when instantiating SSLHandlers (and, by extension, SSLContext). This is needed
+ * when hostname verification is required, for example. The reason this is a
+ * child interface is to keep backward compatibility alive for SSLContext. This 
+ * interface maybe be merged into SSLOptions in a later major release.
+ * 
+ * @see <a href="https://datastax-oss.atlassian.net/browse/JAVA-1364">JAVA-1364</a>
  */
-public interface SSLOptions {
+public interface AddressAwareSSLOptions extends SSLOptions {
 
-    /**
-     * Creates a new SSL handler for the given Netty channel.
-     * <p/>
-     * This gets called each time the driver opens a new connection to a Cassandra host. The newly created handler will be added
-     * to the channel's pipeline to provide SSL support for the connection.
-     * <p/>
-     * You don't necessarily need to implement this method directly; see the provided implementations: {@link JdkSSLOptions} and
-     * {@link NettySSLOptions}.
-     *
-     * @deprecated please use AddressAwareSSLOptions.newSSLHandler(SocketChannel channel, InetSocketAddress address)
-     * @param channel the channel.
-     * @return the handler.
-     * 
-     */
-	@Deprecated
-    SslHandler newSSLHandler(SocketChannel channel);
+    SslHandler newSSLHandler(SocketChannel channel, InetSocketAddress address);
+
 }


olim7t:
Hi,

You're right, we overlooked the fact that the address might be useful to initialize the context, and indeed `channel.remoteAddress()` is still unset at that point. I imagine you're running into something like [this](http://stackoverflow.com/questions/13390964/java-ssl-fatal-error-80-unwrapping-net-record-after-adding-the-https-en).

Your changes are technically sound, but the problem is with the API break. We reserve those for major versions (i.e. 4.0), so if we want to provide the fix earlier we need to preserve binary compatibility. I see two ways:

1) a child interface:
    ```java
    public interface AddressAwareSSLOptions extends SSLOptions {
       SslHandler newSSLHandler(SocketChannel channel, InetSocketAddress address);
    }
    ```
    The channel initializer uses an `instanceof` check; if the options implement the child interface, it calls the method with the address, otherwise it calls the method without the address.
    We deprecate the method in the parent interface, and change `JdkSSLOptions` and `NettySSLOptions` to implement the child interface and have the original method throw `UnsupportedOperationException`.

2) Keep the method as-is, and pass the information as a channel attribute.
    In the channel initializer:
    ```java
    channel.attr(SSLOptions.ADDRESS_ATTRIBUTE_KEY).set(connection.address);
    ```
    To retrieve it from the SSL options:
    ```java
    InetSocketAddress address = channel.attr(SSLOptions.ADDRESS_ATTRIBUTE_KEY).get();
    ```

Neither solution is pretty (binary compatibility is not glamorous). Solution 1 is more explicit, solution 2 is simpler. I would lean towards 2 in this case.

Either way, this needs to be explained in the javadocs, and we'll create a JIRA ticket targeted to 4.0.0 for the breaking fix.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -1311,7 +1311,8 @@ protected void initChannel(SocketChannel channel) throws Exception {
             ChannelPipeline pipeline = channel.pipeline();
 
             if (sslOptions != null) {
-                pipeline.addLast("ssl", sslOptions.newSSLHandler(channel));
+                InetSocketAddress addr = new InetSocketAddress(connection.address.getAddress(), connection.address.getPort());
```
olim7t: Why not `InetSocketAddress addr = connection.address`?

wimtie:
Hi, thanks for your review!

I was afraid the API break would block this PR, but figured it was worth a shot. Do you mind if I update the PR with an implementation of your first suggestion? The second one feels a bit more like a workaround to me. Also, option 1 enables a clear deprecation process.

Cheerio,

Tim

driver-core/src/main/java/com/datastax/driver/core/Connection.java
wimtie: Ah, I didn't realize InetSocketAddress' fields are 'immutable' so I took this approach to avoid possible side effects of state change. You are right though, it's unnecessary.

olim7t:
> Do you mind if I update the PR with an implementation of your first suggestion?

Not at all, go ahead.

wimtie:
@olim7t I've rebased and force pushed a new version. Let's see what AppVeyor thinks of it. Please note that this patch does break compatibility for people who've overriden newSSLHandler on either JdkSSLptions or NettySSLOptions. I can also exclude those classes from the patch, if desired. We are not using them, but I thought it'd be nice to add the new functionality for the users of these.

adutra:
IMO we should merge this into 3.x instead of 3.2.x, which is the bugfix branch for 3.2.0 (which is not even released yet).

adutra:
Created [JAVA-1364](https://datastax-oss.atlassian.net/browse/JAVA-1364) to track this.

adutra:
@wimtie Would you mind updating the changelog has per our [guidelines](https://github.com/datastax/java-driver/blob/3.x/CONTRIBUTING.md)?

driver-core/src/main/java/com/datastax/driver/core/AddrInfoSSLOptions.java
```diff
@@ -0,0 +1,27 @@
...
+ *   limitations under the License.
+ */
+package com.datastax.driver.core;
+
+import java.net.InetSocketAddress;
+
+import io.netty.channel.socket.SocketChannel;
+import io.netty.handler.ssl.SslHandler;
+
+public interface AddrInfoSSLOptions extends SSLOptions {
```
adutra: `AddressAwareSSLOptions` is a clearer name imo.

adutra: Please add a javadoc explaining the reason of this child interface, a link to the Jira ticket (JAVA-1364) and a warning that this interface might be deleted in the next major release of the driver, and merged up into `SSLOptions`.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -1311,7 +1311,13 @@ protected void initChannel(SocketChannel channel) throws Exception {
             ChannelPipeline pipeline = channel.pipeline();
 
             if (sslOptions != null) {
-                pipeline.addLast("ssl", sslOptions.newSSLHandler(channel));
+                if (sslOptions instanceof AddrInfoSSLOptions) {
+                    InetSocketAddress addr = new InetSocketAddress(connection.address.getAddress(), connection.address.getPort());
+                    pipeline.addLast("ssl", ((AddrInfoSSLOptions)sslOptions).newSSLHandler(channel, addr));
+                } else {
+                    // Legacy compatibility
+                    pipeline.addLast("ssl",sslOptions.newSSLHandler(channel));
```
adutra: Currently, this would throw `UnsupportedOperationException` and break working code (see my other comments).

driver-core/src/main/java/com/datastax/driver/core/JdkSSLOptions.java
```diff
@@ -50,17 +53,36 @@ protected JdkSSLOptions(SSLContext context, String[] cipherSuites) {
...
+     * You might want to override this (and newSSLEngine) if you need to fine-tune the engine's configuration
+     * (for example enabling hostname verification).
+     * 
+     * @param channel the Netty channel for that connection.
+     * @param address the remote address of the connection.
+     */
     @Override
-    public SslHandler newSSLHandler(SocketChannel channel) {
+    public SslHandler newSSLHandler(SocketChannel channel, InetSocketAddress address) {
         SSLEngine engine = newSSLEngine(channel);
```
adutra, line 55: Here we should call `newSSLEngine(channel, address)`.

driver-core/src/main/java/com/datastax/driver/core/JdkSSLOptions.java
```diff
@@ -50,17 +53,36 @@ protected JdkSSLOptions(SSLContext context, String[] cipherSuites) {
...
         return new SslHandler(engine);
     }
 
     /**
+     * Creates an SSLHandler for a connection.
+     * Deprecated: code uses newSSLHandler(SocketChannel channel, InetSocketAddress address) instead.
+     * @param channel the Netty channel for that connection.
+     */
+    @Deprecated
```
adutra: The deprecation should be declared in `SSLOptions`. No javadoc needed here.

driver-core/src/main/java/com/datastax/driver/core/JdkSSLOptions.java
```diff
@@ -50,17 +53,36 @@ protected JdkSSLOptions(SSLContext context, String[] cipherSuites) {
...
+    @Override
+    public SslHandler newSSLHandler(SocketChannel channel) {
+        throw new UnsupportedOperationException("This is a placeholder, please use newSSLHandler(SocketChannel channel, InetSocketAddress address)");
```
adutra: This will break existing code. Instead we should keep the old implementation.

driver-core/src/main/java/com/datastax/driver/core/JdkSSLOptions.java
```diff
@@ -74,6 +96,25 @@ protected SSLEngine newSSLEngine(SocketChannel channel) {
...
+     * <p/>
+     * You might want to override this if you need to fine-tune the engine's configuration
+     * (for example enabling hostname verification).
+     *
+     * @param channel the Netty channel for that connection.
+     * @param address remote address for the connection
+     * @return the engine.
+     */
+    protected SSLEngine newSSLEngine(SocketChannel channel, InetSocketAddress address) {
+        SSLEngine engine = context.createSSLEngine(address.getHostName(), address.getPort());
```
adutra: It might be a better idea to make `newSSLEngine(SocketChannel channel)` delegate to `newSSLEngine(SocketChannel channel, InetSocketAddress address)` and test the nullity of parameter `address`. This would reduce the amount of duplicated code.

driver-core/src/main/java/com/datastax/driver/core/NettySSLOptions.java
```diff
@@ -37,7 +39,13 @@ public NettySSLOptions(SslContext context) {
     }
 
     @Override
+    public SslHandler newSSLHandler(SocketChannel channel, InetSocketAddress address) {
+        return context.newHandler(channel.alloc(), address.getHostName(), address.getPort());
+    }
+
+    @Override
     public SslHandler newSSLHandler(SocketChannel channel) {
-        return context.newHandler(channel.alloc());
```
adutra, line 41: Same here, the current implementation should be kept as is to avoid breaking user code.

driver-core/src/main/java/com/datastax/driver/core/JdkSSLOptions.java
```diff
@@ -50,17 +53,36 @@ protected JdkSSLOptions(SSLContext context, String[] cipherSuites) {
...
+    public SslHandler newSSLHandler(SocketChannel channel) {
+        throw new UnsupportedOperationException("This is a placeholder, please use newSSLHandler(SocketChannel channel, InetSocketAddress address)");
+    }
+
+    /**
      * Creates an SSL engine each time a connection is established.
      * <p/>
      * <p/>
-     * You might want to override this if you need to fine-tune the engine's configuration
+     * You might want to override this (and newSSLHandler) if you need to fine-tune the engine's configuration
```
adutra: nit: `(and its {@link #newSSLEngine(SocketChannel, InetSocketAddress) overloaded method})`

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -1311,7 +1311,13 @@ protected void initChannel(SocketChannel channel) throws Exception {
             ChannelPipeline pipeline = channel.pipeline();
 
             if (sslOptions != null) {
-                pipeline.addLast("ssl", sslOptions.newSSLHandler(channel));
+                if (sslOptions instanceof AddrInfoSSLOptions) {
+                    InetSocketAddress addr = new InetSocketAddress(connection.address.getAddress(), connection.address.getPort());
```
adutra: This seems unnecessary, `connection.address` is already an `InetSocketAddress`, so the following should suffice:

```java
pipeline.addLast("ssl", sslOptions.newSSLHandler(channel, connection.address));
```

driver-core/src/main/java/com/datastax/driver/core/JdkSSLOptions.java
```diff
@@ -50,17 +53,36 @@ protected JdkSSLOptions(SSLContext context, String[] cipherSuites) {
         this.cipherSuites = cipherSuites;
     }
 
+    /**
```
adutra: Javadoc here is unnecessary, this is an implementation method.

adutra (changes requested):

wimtie:
I've changed the PR to be a minimal change that allows creation of SSLContext with remote address. Whatever I do to JdkSSLOptions or NettyOptions will always break something, so I've left them alone for this change. @adutra: under what version should I add this to the changelog?

olim7t:
> Whatever I do to JdkSSLOptions or NettyOptions will always break something

What if you implement `newSSLHandler(SocketChannel, InetSocketAddress)` to call `newSSLHandler(SocketChannel)` by default? It won't break existing implementations, and for new ones we'd recommended redefining the method instead.

wimtie:
@olim7t that'd work, but it would give the false sense that the `SSLContext` would be initialized with the remote address, since `JdkSSLOptions` would now implement `AddressAwareSSLOptions`. I thought it'd be better to avoid that. What's the preferred option here?

olim7t:
I think it's acceptable if we deprecate the original methods and explain the situation in the comments.

adutra:
Closing in favor of #796. @wimtie thank you again for your contributions.

PR #784, opened 2017-01-11
JAVA-1328: Provide compatibility with Guava 20

olim7t:
Based on original work by @mspangdal.


adutra:
Is there a compelling reason to target 3.0.7 instead of 3.2.0? I would personally be more comfortable with 3.2.0.

driver-core/src/main/java/com/datastax/driver/core/GuavaCompatibility.java
```diff
@@ -0,0 +1,215 @@
...
+    private static String stripQualifiers(String fullVersion) {
+        Matcher matcher = GA_VERSION_EXTRACTOR.matcher(fullVersion);
+        if (matcher.matches()) {
+            return matcher.group(1);
+        } else {
+            throw new DriverInternalError(String.format("Could not strip qualifiers from full Guava version %s", fullVersion));
+        }
+    }
+
+    private static class Version18OrLower extends GuavaCompatibility {
```
adutra, line 196: Nit: adding `@SuppressWarnings("deprecation")` here would spare us a couple warnings.

driver-core/src/main/java/com/datastax/driver/core/GuavaCompatibility.java
```diff
@@ -0,0 +1,215 @@
...
+        } else {
+            logger.info("Detected Guava {} in the classpath, using 19+ compatibility layer", fullVersion);
+            return new Version19OrHigher();
+        }
+    }
+
+    private static Manifest loadGuavaManifest() {
+        InputStream in = null;
+        try {
+            Enumeration<URL> resources = Resources.class.getClassLoader()
```
adutra: `Resources` is itself part of Guava, and is marked `@Beta`. Why not use `GuavaCompatibility.class.getClassLoader()`?

driver-core/src/main/java/com/datastax/driver/core/GuavaCompatibility.java
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        @Override
+        public <I, O> ListenableFuture<O> transformAsync(ListenableFuture<I> input, AsyncFunction<? super I, ? extends O> function) {
+            return Futures.transform(input, function);
+        }
+
+        @Override
+        public <I, O> ListenableFuture<O> transformAsync(ListenableFuture<I> input, AsyncFunction<? super I, ? extends O> function, Executor executor) {
+            return Futures.transform(input, function, executor);
```
adutra, line 228: Found one occurrence of this method in `com.datastax.driver.core.AsyncQueryTest#connectAndQuery`

driver-core/src/main/java/com/datastax/driver/core/utils/MoreObjects.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.Arrays;
+
+/**
+ * Driver-specific implementation of utility object methods.
+ * <p>
+ * They are available in some versions of Java/Guava, but not across all versions ranges supported by the driver, hence
+ * the custom implementation.
+ */
+public class MoreObjects {
```
adutra, line 26: Since it's public, maybe we could add javadocs for methods in this class, even if I concede that their purposes are pretty obvious.

driver-core/src/test/java/com/datastax/driver/core/policies/LimitingLoadBalancingPolicy.java
```diff
@@ -40,8 +40,8 @@
 public class LimitingLoadBalancingPolicy extends DelegatingLoadBalancingPolicy {
```
adutra, line 39: The import `import com.google.common.collect.Sets` is unused.

adutra (changes requested):

adutra:
In addition to my review, we should also

1) in `/pom.xml`, replace the javadoc link with this one:

```
 <link>https://google.github.io/guava/releases/20.0/api/docs/</link>
```

2) Replace links to `code.google.com/p/guava-libraries` with `github.com/google/guava`; this happens in two files at least: `CloseFuture.java` and `ResultSetFuture.java`.

driver-core/src/main/java/com/datastax/driver/core/GuavaCompatibility.java
olim7t: I hadn't noticed it was beta. I think it's better to use a Guava class to make sure it's the classloader that loaded Guava, I've changed it to `Preconditions`.

olim7t:
Addressed all feedback.

tolbertam:
> Is there a compelling reason to target 3.0.7 instead of 3.2.0? I would personally be more comfortable with 3.2.0.

That's a good point, it doesn't break API compatibility, but is more of a new capability than a hotfix so 3.2.0 does seem more appropriate.

tolbertam:
CI looks happy with guava 16.0.1, 17.0, 18.0, 19.0, 20.0 and 21.0, excellent work!

adutra (approved):

Xaerxess:
Glad to see this approved! Is there any chance for a release in the near future?

adutra:
@Xaerxess no timeframe yet, but it shouldn't be long.

adutra:
Squashed and rebased on top of 3.x instead of 3.0.x.

adutra:
Note to reviewers: I added an extra commit to fix OSGi tests and also to revert the Import-Package instruction that we use to create our bundles. We need to explicitly declare that the driver bundle accepts any Guava version in the range [16.0.1,21) so I modified the instructions to read that.

palaiya (approved):

PR #785, opened 2017-01-12
Update tests to use scassandra 1.1.1

tolbertam:
Remove Protocol V2 specific callouts in Scassandra tests.  This is no longer needed since scassandra supports v3 and v4 now.

Rewrite speculative execution integration test using Scassandra since it now supports client timestamps.

Allow setting cassandra version in ScassandraCluster.  The cassandra_version returned in peers table is significant in determining if the driver should connect to that node depending on its
protocol version.  Defaulting to 3.0.10 is a fair route since Scassandra supports protocol version 4.

Also prime keyspace metadata based on cassandra version as system schema tables changed in 3.0.

Add adhoc job for running all configurations on a non-primary branch.  This job has no schedule and can be used for manually kicking off all configs with long suite.


adutra (approved):
Works like a charm, very good job! 👍

olim7t (approved):

PR #788, opened 2017-01-20
JAVA-682: added method to perform query based on partition key

stsourdos:
added method signature: getByPartitionKey(Object... objects), which returns a list of entities T

Jira issue described here: [](https://datastax-oss.atlassian.net/browse/JAVA-682)


adutra:

@stsourdos see JAVA-682 and JAVA-667. We decided to not offer range queries out of the box because it's usually not a good idea. The alternative is to use accessors. Thanks anyway for your time and effort.

PR #789, opened 2017-01-24
JAVA-1373: Encode flags in Prepare request for Protocol V5+

olim7t:


olim7t:
I'm 👍 on this (can't review because I created the PR). I tested against Tyler's branch for CASSANDRA-10145.
I agree we should merge ASAP to guarantee protocol compatibility once 10145 is merged, JAVA-1350 can wait a bit more.

adutra (approved):

tolbertam (approved):
👍

tolbertam:
Just remembered one possible consideration.  If someone is on 3.10 and tries using the beta flag, if this PR is merged, prepared statements won't work.  On the other hand, i'm not really concerned about that.  I'd rather we stay ahead of changes instead of intentionally lagging behind to try to work with released cassandra versions w/ beta.   The beta flag is more for development than anything else at this point right?

olim7t:
> I'd rather we stay ahead of changes instead of intentionally lagging behind

Agreed, this falls under the "things can break when in beta" rule.

PR #790, opened 2017-01-25
JAVA-1367: Make protocol negotiation more resilient

adutra:
Motivation:

CASSANDRA-11464 introduced in 3.0.8+, 3.8+ changed logic of protocol
error messages to be sent with the protocol version requested
by the client, instead of a version that the server supports,
to ensure that the client will be able to decode the response properly.

The driver currently uses the returned protocol version to determine
what protocol version to downgrade to and reattempt connection.
Because of this, the driver would retry the same protocol version and
fail again.

Modifications:

This commit changes the renegotiation algorithm to allow
more than one renegotiation attempt. It also makes no more distinction
between user-defined versions and inferred versions.
And finally, it now ignores the server response version
and always retries to connect using attempted version - 1.

Result:

Driver is compatible with Cassandra 3.0.8+, 3.8+.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -750,6 +750,7 @@ public class Cluster implements Closeable {
          * <p/>
          * Use with caution, refer to the server and protocol documentation for the details
          * on latest protocol version.
+         *
          * @return this Builder.
          */
         public Builder allowBetaProtocolVersion() {
@@ -1504,18 +1505,7 @@ public class Cluster implements Closeable {
             Set<Host> contactPointHosts = Sets.newHashSet(allHosts);
 
             try {
-                try {
-                    controlConnection.connect();
-                } catch (UnsupportedProtocolVersionException e) {
-                    logger.debug("Cannot connect with protocol {}, trying {}", e.getUnsupportedVersion(), e.getServerVersion());
-
-                    connectionFactory.protocolVersion = e.getServerVersion();
-                    try {
-                        controlConnection.connect();
-                    } catch (UnsupportedProtocolVersionException e1) {
-                        throw new DriverInternalError("Cannot connect to node with its own version, this makes no sense", e);
-                    }
-                }
+                negotiateProtocolVersionAndConnect();
 
                 // The control connection can mark hosts down if it failed to connect to them, or remove them if they weren't found
                 // in the control host's system.peers. Separate them:
@@ -1567,7 +1557,7 @@ public class Cluster implements Closeable {
                     // creations if a session is created right after this method returns).
                     logger.info("New Cassandra host {} added", host);
 
-                    if (!connectionFactory.protocolVersion.isSupportedBy(host)) {
+                    if (!host.supports(connectionFactory.protocolVersion)) {
                         logUnsupportedVersionProtocol(host, connectionFactory.protocolVersion);
                         continue;
                     }
@@ -1590,6 +1580,30 @@ public class Cluster implements Closeable {
             } catch (NoHostAvailableException e) {
                 close();
                 throw e;
+            }
+        }
+
+        private void negotiateProtocolVersionAndConnect() {
+            boolean shouldNegotiate = (configuration.getProtocolOptions().initialProtocolVersion == null);
+            while (true) {
+                try {
+                    controlConnection.connect();
+                    return;
+                } catch (UnsupportedProtocolVersionException e) {
+                    if (!shouldNegotiate) {
+                        throw e;
+                    }
+                    // Do not trust version of server's response, as C* behavior in case of protocol negotiation is not
+                    // properly documented, and varies over time (specially after CASSANDRA-11464). Instead, always
+                    // retry at attempted version - 1, if such a version exists; and otherwise, stop and fail.
+                    ProtocolVersion attemptedVersion = e.getUnsupportedVersion();
+                    ProtocolVersion retryVersion = attemptedVersion.getLowerSupported();
+                    if (retryVersion == null) {
+                        throw e;
+                    }
+                    logger.info("Cannot connect with protocol version {}, trying with {}", attemptedVersion, retryVersion);
+                    connectionFactory.protocolVersion = retryVersion;
+                }
             }
         }
 
@@ -1734,7 +1748,7 @@ public class Cluster implements Closeable {
             if (isClosed())
                 return;
 
-            if (!connectionFactory.protocolVersion.isSupportedBy(host)) {
+            if (!host.supports(connectionFactory.protocolVersion)) {
                 logUnsupportedVersionProtocol(host, connectionFactory.protocolVersion);
                 return;
             }
@@ -2028,7 +2042,7 @@ public class Cluster implements Closeable {
             if (isClosed())
                 return;
 
-            if (!connectionFactory.protocolVersion.isSupportedBy(host)) {
+            if (!host.supports(connectionFactory.protocolVersion)) {
                 logUnsupportedVersionProtocol(host, connectionFactory.protocolVersion);
                 return;
             }
--- a/driver-core/src/main/java/com/datastax/driver/core/Connection.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Connection.java
@@ -248,10 +248,7 @@ class Connection {
                         return checkClusterName(protocolVersion, initExecutor);
                     case ERROR:
                         Responses.Error error = (Responses.Error) response;
-                        // Testing for a specific string is a tad fragile but well, we don't have much choice
-                        // C* 2.1 reports a server error instead of protocol error, see CASSANDRA-9451
-                        if ((error.code == ExceptionCode.PROTOCOL_ERROR || error.code == ExceptionCode.SERVER_ERROR) &&
-                                error.message.contains("Invalid or unsupported protocol version"))
+                        if (isUnsupportedProtocolVersion(error))
                             throw unsupportedProtocolVersionException(protocolVersion, error.serverProtocolVersion);
                         throw new TransportException(address, String.format("Error initializing connection: %s", error.message));
                     case AUTHENTICATE:
@@ -405,9 +402,17 @@ class Connection {
         }
     }
 
+    private boolean isUnsupportedProtocolVersion(Responses.Error error) {
+        // Testing for a specific string is a tad fragile but well, we don't have much choice
+        // C* 2.1 reports a server error instead of protocol error, see CASSANDRA-9451
+        return (error.code == ExceptionCode.PROTOCOL_ERROR || error.code == ExceptionCode.SERVER_ERROR) &&
+                error.message.contains("Invalid or unsupported protocol version");
+    }
+
     private UnsupportedProtocolVersionException unsupportedProtocolVersionException(ProtocolVersion triedVersion, ProtocolVersion serverProtocolVersion) {
-        logger.debug("Got unsupported protocol version error from {} for version {} server supports version {}", address, triedVersion, serverProtocolVersion);
-        return new UnsupportedProtocolVersionException(address, triedVersion, serverProtocolVersion);
+        UnsupportedProtocolVersionException e = new UnsupportedProtocolVersionException(address, triedVersion, serverProtocolVersion);
+        logger.debug(e.getMessage());
+        return e;
     }
 
     boolean isDefunct() {
@@ -1000,7 +1005,7 @@ class Connection {
             ProtocolVersion protocolVersion = factory.protocolVersion;
             if (protocolVersion == null) {
                 // This happens for the first control connection because the protocol version has not been
-                // negociated yet.
+                // negotiated yet.
                 protocolVersion = ProtocolVersion.V2;
             }
             streamIdHandler = StreamIdGenerator.newInstance(protocolVersion);
--- a/driver-core/src/main/java/com/datastax/driver/core/ProtocolVersion.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ProtocolVersion.java
@@ -26,38 +26,36 @@ import java.util.Map;
  */
 public enum ProtocolVersion {
 
-    V1("1.2.0", 1),
-    V2("2.0.0", 2),
-    V3("2.1.0", 3),
-    V4("2.2.0", 4),
-    V5("3.10.0", 5);
+    V1("1.2.0", 1, null),
+    V2("2.0.0", 2, V1),
+    V3("2.1.0", 3, V2),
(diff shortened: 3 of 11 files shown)


driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -405,9 +405,52 @@ private void incrementAuthErrorMetric() {
         }
     }
 
-    private UnsupportedProtocolVersionException unsupportedProtocolVersionException(ProtocolVersion triedVersion, ProtocolVersion serverProtocolVersion) {
-        logger.debug("Got unsupported protocol version error from {} for version {} server supports version {}", address, triedVersion, serverProtocolVersion);
-        return new UnsupportedProtocolVersionException(address, triedVersion, serverProtocolVersion);
+    private static final Pattern SUPPORTED_VERSIONS_PATTERN = Pattern.compile("supported versions are \\(([^)]+?)\\)");
```
tolbertam: Since C* has been pretty inconsistent about the error messages it returns by release, what about just not parsing the error message at all and then tying the next version to try based on the protocol version that we previously attempted?   The problem we are experiencing at the moment is because we were using information provided by the server, so I think we should just not trust it unless we can get something in the protocol specification about how to handle this scenario.

In the general case, we could simply just retry at the attempted protocol version - 1.  The only downside of this is that it may take multiple attempts to find the right protocol version, but this is only done at initialization anyways, so it doesn't seem to have that much cost.  We could also add a method to `ProtocolVersion` for requesting the next version to try in the case where it's more complex than that, this is actually what the nodejs-driver [is proposing](https://github.com/datastax/nodejs-driver/blob/1ed4a37dc010b0e57e0ae55bfb4704d97e2d622d/lib/types/protocol-version.js#L113-L127), although it still keeps the optimization to use the protocol version in the error response if it is different (like we also do here).

adutra: > Since C* has been pretty inconsistent about the error messages it returns

Not so much actually, I crawled through the revisions of `Frame.java` and found 4 different messages only.

> In the general case, we could simply just retry at the attempted protocol version - 1

I guess what you mean is _if_ the server version differs from the client one, then use that version, otherwise downgrade to N-1, is that right? The only problem I see with the "-1" approach is that downgrading when the server replies with the same version is not useful at all (any lower version will fail as well – what we need in that case is actually upgrade).

olim7t: I don't understand, the client starts negotiation with its highest version. If it started with vN, it doesn't support N+1 nor N+2.

adutra: hmm now that I think of it, maybe we could simplify the algorithm to:

1. if the server version differs from the client one, then use that version;
2. otherwise upgrade to server version + 1, if possible;
3. otherwise fail.

Step 2 sounds counterintuitive, but that could actually work for all cases. Will give some more thought.

tolbertam: > I guess what you mean is if the server version differs from the client one, then use that version, otherwise downgrade to N-1, is that right? The only problem I see with the "-1" approach is that downgrading when the server replies with the same version is not useful at all (any lower version will fail as well – what we need in that case is actually upgrade).

Unless i'm misunderstanding, we would never upgrade the protocol version, only downgrade.

I think the most straightforward approach is what @jorgebay came up with in datastax/nodejs-driver#204.  That implementation is similar to what you said with a few caveats:

    1. if the server version differs from the client one, then use that version;
    2. otherwise downgrade from the previous protocol version you tried

Steps 1 and 2 would then be repeated until a suitable protocol version is found.

I think all the other drivers (other than C# and nodejs) only do step 2.

'downgrade' could mean different things depending on the context of the protocol version, so would be nice to have a method on `ProtocolVersion` for asking for the lower protocol version.

adutra:
Addressed all feedback so far:

1. Removed all parsing of error messages;
2. Simplified renegotiation algorithm.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1504,18 +1504,7 @@ long delayMs() {
...
-                    logger.debug("Cannot connect with protocol {}, trying {}", e.getUnsupportedVersion(), e.getServerVersion());
-
-                    connectionFactory.protocolVersion = e.getServerVersion();
-                    try {
-                        controlConnection.connect();
-                    } catch (UnsupportedProtocolVersionException e1) {
-                        throw new DriverInternalError("Cannot connect to node with its own version, this makes no sense", e);
-                    }
-                }
+                connect();
```
tolbertam: I like that a lot of logic was moved into separate methods, makes code a lot easier to follow 👍

driver-core/src/test/java/com/datastax/driver/core/ProtocolVersionRenegotiationTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+import com.datastax.driver.core.exceptions.NoHostAvailableException;
+import com.datastax.driver.core.exceptions.UnsupportedProtocolVersionException;
+import org.testng.annotations.Test;
+
+import static com.datastax.driver.core.ProtocolVersion.*;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@CreateCCM(CreateCCM.TestMode.PER_METHOD)
+public class ProtocolVersionRenegotiationTest extends CCMTestsSupport {
```
tolbertam, line 25: Excellent tests!

tolbertam:
So far this looks really great.  Going to spend some time experimenting with it, but the test coverage looks like it covers the scenarios well.

driver-core/src/test/java/com/datastax/driver/core/ProtocolVersionRenegotiationTest.java
```diff
@@ -0,0 +1,136 @@
...
+        Cluster cluster = connectWithoutVersion();
+        assertThat(actualProtocolVersion(cluster)).isEqualTo(V1);
+    }
+
+    /**
+     * @jira_ticket JAVA-1367
+     */
+    @Test(groups = "short")
+    @CCMConfig(version = "2.0.17", createCluster = false)
+    public void should_renegotiate_when_renegotiation_enabled_and_initial_version_too_high_2_0_17() throws Exception {
```
tolbertam: This is more of a note for me.  But I realized that in our CI environment we choose the JDK to run C* against based on the global cassandra version for the test.  If C* is less than 2.1, we choose JDK 7 because C* 2.0 doesn't work with JDK 8.  However, I think this assertion is untrue.  I think the issue is that C* 2.0 doesn't compile on Java 8 ([CASSANDRA-7936](https://issues.apache.org/jira/browse/CASSANDRA-7936)), but it will still run on it (works for me locally at least).  I'll fix that.

tolbertam: The other interesting thing about these tests is they run no matter what `-Dcassandra.version` is and I think that is ok.  it is good to get as much validation as possible done in one run.

tolbertam (approved):
Have experimented with this locally a bit and it's working great 👍

driver-core/src/test/java/com/datastax/driver/core/ProtocolVersionRenegotiationTest.java
adutra: `@CCMConfig(version = "XXX")` forces a specific version instead of skipping the test if the version doesn't match the default one. I think it is required sometimes too, as tests like these could fail if run against any other version.

driver-core/src/main/java/com/datastax/driver/core/ProtocolVersion.java
```diff
@@ -73,6 +68,26 @@ public int toInt() {
...
+                return V4;
+            case V4:
+                return V3;
+            case V3:
+                return V2;
+            case V2:
+                return V1;
+        }
+        return null;
+    }
```
olim7t: How about
```
return INT_TO_VERSION.get(asInt - 1);
```

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1593,10 +1582,35 @@ else if (host.state == Host.State.DOWN)
...
 
+        private void connect() {
+            try {
+                controlConnection.connect();
+            } catch (UnsupportedProtocolVersionException e) {
+                ProtocolVersion newVersion = e.getServerVersion();
+                if (newVersion == e.getUnsupportedVersion()) {
+                    newVersion = newVersion.getLowerSupported();
+                    if (newVersion == null) {
+                        throw e;
```
olim7t: This was a bit hard to follow for me. I'm pushing a proposal with intermediate variables and comments.

olim7t (approved):
lgtm with 2 minor comments.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1593,15 +1593,15 @@ private void connect() {
                 if (serverResponseVersion == attemptedVersion) {
                     // We're post CASSANDRA-11464: the server replies with the same version that the client tried (to
                     // ensure that the client can decode the response).
-                    // Retry with a lower version (if it's too low we'll eventually reach our min and fail).
+                    // Retry with a lower version (if it's too low we'll fail at the next attempt, as we only retry once).
```
olim7t: Right, I missed that.
So what happens if the server supports up to v10 and the client starts with v12?

adutra: > what happens if the server supports up to v10 and the client starts with v12?

The server would reply with v10. If the driver also supports v10, then v10 will be used; otherwise, it will fail when decoding the response frame ([here](https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/Frame.java#L93-L93) more precisely).

olim7t: OK, so the last piece I was missing is that C* only replies with the same version if the driver started too low (e.g. C* supports v12+ and the driver started with v10).
In that case, what's the point of even retrying? If the driver was already too low, going lower won't help.

adutra: > In that case, what's the point of even retrying?

That was my point since the beginning. But @jorgebay convinced me: the fact that retrying at N-1 is necessarily going to fail is not a documented feature, but an implementation detail at this point. We don't know if this is going to always be the case, so retrying at N-1 is more future-proof.

olim7t: > the fact that retrying at N-1 is necessarily going to fail is not a documented feature

If we're going that route, the same could be said for N-2.

This is getting complicated. How about going back to the simple algorithm used [before JAVA-361](https://github.com/datastax/java-driver/blob/2.1.0/driver-core/src/main/java/com/datastax/driver/core/Cluster.java#L1143-L1168):
* try with version N
* if we get a PROTOCOL_ERROR, simply retry with the driver's previous version. Ignore the version it's encoded in (that's not documented in the spec either after all).
* repeat until we run out of driver's supported versions.

Yes, we might do a couple of extra retries in cases where there is no chance of success. It's not a big deal, negotiation is not supposed to be efficient anyway, and we discourage its use in production.

adutra: > This is getting complicated. How about going back to the simple algorithm used before JAVA-361.

I don't have anything against the idea. I will push the modifications shortly if nobody disagrees.

adutra: Sorry, but your suggestion makes some tests fail. The main reason is that we do not retry more than once, so if we start too high, the second attempt will still be too high and fail, whereas currently the second attempt could potentially be much lower that the initial one. 

On the light of this, I'm actually reconsidering my position: I think we should rather keep the current algorithm.

adutra:
As discussed, I'm proposing a revised algorithm that basically gives us pre-ProtocolVersion enum behavior.

tolbertam:
🤦‍♂️ I spelled 'accommodate' wrong in my commit message for `f228596` when you squash can you either fix that or not include that in the final commit?

tolbertam:
Changes look good to me 👍

adutra:
Squashed and rebased. Ready to be merged.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1582,6 +1583,30 @@ else if (host.state == Host.State.DOWN)
...
+                    }
+                    // Do not trust version of server's response, as C* behavior in case of protocol negotiation is not
+                    // properly documented, and varies over time (specially after CASSANDRA-11464). Instead, always
+                    // retry at attempted version - 1, if such a version exists; and otherwise, stop and fail.
+                    ProtocolVersion attemptedVersion = e.getUnsupportedVersion();
+                    ProtocolVersion retryVersion = attemptedVersion.getLowerSupported();
+                    if (retryVersion == null) {
+                        throw e;
+                    }
+                    logger.info("Cannot connect with protocol version {}, trying with {}", attemptedVersion, retryVersion);
```
olim7t, line 1604: I raised the level of this message from DEBUG.

driver-core/src/test/java/com/datastax/driver/core/ProtocolVersionRenegotiationTest.java
```diff
@@ -53,45 +62,31 @@ public void should_fail_when_renegotiation_disabled_3_10() throws Exception {
...
     }
 
     /**
      * @jira_ticket JAVA-1367
      */
     @Test(groups = "short")
-    @CCMConfig(version = "2.0.17", createCluster = false)
-    public void should_renegotiate_and_initial_version_too_high_2_0_17() throws Exception {
-        Cluster cluster = connectWithVersion(V4);
-        assertThat(actualProtocolVersion(cluster)).isEqualTo(V2);
```
olim7t: There should be no negotiation when an explicit version is provided. `withProtocolVersion`'s javadocs say:
> This method allows to force the use of a particular protocol version.

tolbertam: I think that might have got lost in the shuffle or there might have been a misinterpretation as the commit message includes:

>  It also makes no more distinction between user-defined versions and inferred versions.

Agreed that renegotiation should only occur if the protocol verison is not provided.

adutra: Sorry, that wasn't at all what I understood from our conversation.

PR #791, opened 2017-01-25
Various test fixes for 3.2.0

adutra:
Motivation:

Many tests are failing when using recent versions of C* or DSE.

Modifications:

1) Bump the default C* version to 3.8
2) Redesign @CassandraVersion and @DseVersion to be able to fully accept
tick-tock releases and correctly compare them.
3) Reduce log verbosity for Scassandra tests
4) Prevent thrift from being enabled for C* >= 4.0
5) Fix decommission tests with C* >= 3.12

This commit contains contributions by Andrew Tolbert (@tolbertam).

Result:

All tests should pass.

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
@@ -282,6 +282,21 @@ public class CCMBridge implements CCMAccess {
     }
 
     /**
+     * @return {@link VersionNumber} representation of {@link CCMBridge#getCassandraVersion()}
+     */
+    public static VersionNumber getCassandraVersionNumber() {
+        return VersionNumber.parse(getCassandraVersion());
+    }
+
+    /**
+     * @param compareVersion The given version to compare with.
+     * @return Whether or not the configured cassandra version is greater than or equal to the given version.
+     */
+    public static boolean isCassandraVersionAtLeast(String compareVersion) {
+        return getCassandraVersionNumber().compareTo(VersionNumber.parse(compareVersion)) >= 0;
+    }
+
+    /**
      * @return The configured DSE version if '-Ddse=true' specified, otherwise null.
      */
     public static String getDSEVersion() {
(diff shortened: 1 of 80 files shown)


driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
```diff
@@ -941,6 +980,12 @@ public void run() {
             return ccm;
         }
 
+        private static boolean isThriftSupported(boolean dse, VersionNumber version) {
+            return
+                    (!dse && version.compareTo(VersionNumber.parse("3.10")) < 0)
```
tolbertam: I think C* 4.0 will be the first version without thrift.  It will still be present in 3.10, but not in trunk.

tolbertam (changes requested):
Other than my comment about C* and Thrift, this looks good to me 👍

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
```diff
@@ -941,6 +978,10 @@ public void run() {
             return ccm;
         }
 
+        private static boolean isThriftSupported(boolean dse, VersionNumber version) {
+            return dse || version.compareTo(VersionNumber.parse("4.0")) < 0;
```
tolbertam, line 983: Just curious, but what is the significance of DSE here?  Although all versions of DSE support Thrift at the moment, i imagine that will change too.  That being said I think it is fine this way for now.

tolbertam (approved):
👍

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
adutra: I'm just assuming that all versions of DSE support thrift for now, because I don't know when C* 4.0 will get pulled in DSE.

olim7t:
> Redesign CassandraVersion and DseVersion to be able to fully accept tick-tock releases and correctly compare them.

💯  for using a string for the version.

PR #792, opened 2017-01-25
JAVA-1371: Reintroduce connection pool timeout

olim7t:


driver-core/src/main/java/com/datastax/driver/core/PoolingOptions.java
```diff
@@ -435,7 +447,9 @@ public int getMaxQueueSize() {
      * Sets the maximum number of requests that get enqueued if no connection is available.
      * <p/>
      * If the queue grows past this value, new requests will be rejected immediately (and the driver will move to the
-     * next host in the query plan). This limit is per connection pool, not global to the driver.
+     * next host in the query plan). This limit is per connection pool, not global to the driver. This setting takes
+     * priority over {@link #setPoolTimeoutMillis(int)}; that is, if the queue is full, new attempts to borrow will be
+     * rejected, even if the timeout is not null.
```
adutra: Suggestion: "even if the timeout is not null" -> "even if no timeout has been set"

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
```diff
@@ -264,8 +269,8 @@ private PoolingOptions options() {
         return leastBusy.setKeyspaceAsync(manager.poolsState.keyspace);
     }
 
-    private ListenableFuture<Connection> enqueue(int maxQueueSize) {
-        if (maxQueueSize == 0) {
+    private ListenableFuture<Connection> enqueue(long timeout, TimeUnit unit, int maxQueueSize) {
+        if (timeout == 0 || maxQueueSize == 0) {
```
tolbertam, line 273: Nice side effect of this change!  Unless i'm misunderstanding, if we have to enqueue at all and timeout is 0 we move on to the next host immediately?

driver-core/src/main/java/com/datastax/driver/core/PoolingOptions.java
```diff
@@ -406,19 +406,31 @@ public PoolingOptions setIdleTimeoutSeconds(int idleTimeoutSeconds) {
...
     }
 
     /**
-     * @deprecated the connection pool does not use a timeout anymore, incoming requests are now throttled with a
-     * threshold on the {@link #setMaxQueueSize(int) queue size}. This method has no effect.
+     * Sets the timeout when trying to acquire a connection from a host's pool.
+     * <p/>
+     * If no connection is available within that time, the driver will try the next host from the query plan.
+     * {@link #setMaxQueueSize(int)} takes priority over this setting; that is, if the queue is full, new attempts to
+     * borrow will be rejected, even if the timeout is not null.
```
tolbertam: should be 'even if the timeout is non zero'.

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
tolbertam: Oh, that's what the docs say for the timeout so that is the case 👍

driver-core/src/test/java/com/datastax/driver/core/HostConnectionPoolTest.java
```diff
@@ -205,9 +231,9 @@ public void should_adjust_connection_keyspace_on_dequeue_if_pool_state_is_differ
             // will fail, and thus the enqueued connection futures will fail in the following ways:
             // 1. DriverInternalError for set keyspace returning 'ROWS' response
             // 2. TransportException for connection being closed, when this happens we stop checking these futures.
-            // 3. Never completing because no connections are available.
+            // 3. Timing out because no connections are available.
```
tolbertam: D'oh, it should have been apparent to me that this could be a problem.  Even documented that the future would never complete here!

adutra (approved):

olim7t:
I reworded the javadocs because "takes priority" was ambiguous (a zero timeout can actually take priority over a non-zero queue size).

tolbertam:
Added an additional test.  This is looking good to me.  I'm just running a few more tests since there were some changes around in flight, but I don't expect any problems there, should give my thumbs up shortly.

tolbertam (approved):
This is working great.  Tested a number of scenarios where we pending borrows were the norm to ensure inflight didn't get off (i reduce load to 0 intermittently to see if the pool recovers to 0 inflight) and that the connection states stayed ok.

PR #794, opened 2017-01-30

tolbertam:
Created from [ifesdjeen/java-driver#10786-release](https://github.com/ifesdjeen/java-driver/tree/10786-release) and rebased on 3.x.    Used [tolbertam/cassandra#10786-queryflags](https://github.com/tolbertam/cassandra/tree/10786-queryflags) to test it.

Still needs review, but the tests are running good against that C* branch.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/ArrayBackedResultSet.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ArrayBackedResultSet.java
@@ -39,7 +39,7 @@ abstract class ArrayBackedResultSet implements ResultSet {
 
     private static final Queue<List<ByteBuffer>> EMPTY_QUEUE = new ArrayDeque<List<ByteBuffer>>(0);
 
-    protected final ColumnDefinitions metadata;
+    protected volatile ColumnDefinitions metadata;
     protected final Token.Factory tokenFactory;
     private final boolean wasApplied;
 
@@ -60,18 +60,42 @@ abstract class ArrayBackedResultSet implements ResultSet {
             case ROWS:
                 Responses.Result.Rows r = (Responses.Result.Rows) msg;
 
-                ColumnDefinitions columnDefs;
-                if (r.metadata.columns == null) {
-                    Statement actualStatement = statement;
-                    if (statement instanceof StatementWrapper) {
-                        actualStatement = ((StatementWrapper) statement).getWrappedStatement();
-                    }
-                    assert actualStatement instanceof BoundStatement;
-                    columnDefs = ((BoundStatement) actualStatement).statement.getPreparedId().resultSetMetadata;
-                    assert columnDefs != null;
-                } else {
-                    columnDefs = r.metadata.columns;
+                Statement actualStatement = statement;
+                if (statement instanceof StatementWrapper) {
+                    actualStatement = ((StatementWrapper) statement).getWrappedStatement();
                 }
+
+                ColumnDefinitions columnDefs = r.metadata.columns;
+                if (columnDefs == null) {
+                    // If result set metadata is not present, it means the request had SKIP_METADATA set, the driver
+                    // only ever does that for bound statements.
+                    BoundStatement bs = (BoundStatement) actualStatement;
+                    columnDefs = bs.preparedStatement().getPreparedId().resultSetMetadata.variables;
+                } else {
+                    // Otherwise, always use the response's metadata.
+                    // In addition, if a new id is present it means we're executing a bound statement with protocol v5,
+                    // the schema changed server-side, and we need to update the prepared statement (see
+                    // CASSANDRA-10786).
+                    MD5Digest newMetadataId = r.metadata.metadataId;
+                    assert !(actualStatement instanceof BoundStatement) ||
+                            ProtocolFeature.PREPARED_METADATA_CHANGES.isSupportedBy(protocolVersion) ||
+                            newMetadataId == null;
+                    if (newMetadataId != null) {
+                        BoundStatement bs = ((BoundStatement) actualStatement);
+                        PreparedId preparedId = bs.preparedStatement().getPreparedId();
+                        // Extra test for CASSANDRA-13992: conditional updates yield a different result set depending on
+                        // whether the update was applied or not, so the prepared statement must never have result
+                        // metadata, and we should always execute with skip_metadata = false.
+                        // However the server sends a new_metadata_id in the response, so make sure we ignore it if the
+                        // prepared statement did not have metadata in the first place.
+                        // TODO remove the "if" (i.e. always assign resultSetMetadata) if CASSANDRA-13992 gets fixed before 4.0.0 GA
+                        if (preparedId.resultSetMetadata.variables != null) {
+                            preparedId.resultSetMetadata =
+                                    new PreparedId.PreparedMetadata(newMetadataId, columnDefs);
+                        }
+                    }
+                }
+                assert columnDefs != null;
 
                 Token.Factory tokenFactory = (session == null) ? null
                         : session.getCluster().manager.metadata.tokenFactory();
@@ -224,7 +248,7 @@ abstract class ArrayBackedResultSet implements ResultSet {
     private static class MultiPage extends ArrayBackedResultSet {
 
         private Queue<List<ByteBuffer>> currentPage;
-        private final Queue<Queue<List<ByteBuffer>>> nextPages = new ConcurrentLinkedQueue<Queue<List<ByteBuffer>>>();
+        private final Queue<NextPage> nextPages = new ConcurrentLinkedQueue<NextPage>();
 
         private final Deque<ExecutionInfo> infos = new LinkedBlockingDeque<ExecutionInfo>();
 
@@ -280,8 +304,8 @@ abstract class ArrayBackedResultSet implements ResultSet {
         @Override
         public int getAvailableWithoutFetching() {
             int available = currentPage.size();
-            for (Queue<List<ByteBuffer>> page : nextPages)
-                available += page.size();
+            for (NextPage page : nextPages)
+                available += page.data.size();
             return available;
         }
 
@@ -297,9 +321,12 @@ abstract class ArrayBackedResultSet implements ResultSet {
                 // Grab the current state now to get a consistent view in this iteration.
                 FetchingState fetchingState = this.fetchState;
 
-                Queue<List<ByteBuffer>> nextPage = nextPages.poll();
+                NextPage nextPage = nextPages.poll();
                 if (nextPage != null) {
-                    currentPage = nextPage;
+                    if (nextPage.metadata != null) {
+                        this.metadata = nextPage.metadata;
+                    }
+                    currentPage = nextPage.data;
                     continue;
                 }
                 if (fetchingState == null)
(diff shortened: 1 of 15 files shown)


driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -2201,7 +2201,14 @@ public PreparedStatement addPrepared(PreparedStatement stmt) {
 
                 // The one object in the cache will get GCed once it's not referenced by the client anymore since we use a weak reference.
                 // So we need to make sure that the instance we do return to the user is the one that is in the cache.
-                return previous;
+                if (previous.getPreparedId().getResultMetadataId().equals(stmt.getPreparedId().metadata)) {
```
adutra: This should be ` if (previous.getPreparedId().getResultMetadataId().equals(stmt.getPreparedId().getResultMetadataId()))`.

driver-core/src/main/java/com/datastax/driver/core/PreparedId.java
```diff
@@ -24,16 +24,33 @@
...
+    PreparedId(MD5Digest id, MD5Digest resultMetadataId, ColumnDefinitions metadata, ColumnDefinitions resultSetMetadata, int[] routingKeyIndexes, ProtocolVersion protocolVersion) {
         this.id = id;
+        this.resultMetadataId = resultMetadataId;
         this.metadata = metadata;
         this.resultSetMetadata = resultSetMetadata;
         this.routingKeyIndexes = routingKeyIndexes;
         this.protocolVersion = protocolVersion;
     }
+
+    public void swap(MD5Digest metadataId, ColumnDefinitions resultSetMetadata)
```
adutra: This swap is not atomic, we should either synchronize on `this` or create an `AtomicReference` to hold a value object containing both data.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
adutra: The warning line 2199 should imo only be printed if we are in this side of the if/else block.

driver-core/src/main/java/com/datastax/driver/core/Responses.java
```diff
@@ -380,6 +383,14 @@ static Metadata decode(ByteBuf body, boolean withPkIndices, ProtocolVersion prot
                     EnumSet<Flag> flags = Flag.deserialize(body.readInt());
                     int columnCount = body.readInt();
 
+                    MD5Digest resultMetadataId = null;
+                    if (flags.contains(Flag.METADATA_CHANGED))
+                    {
+                        assert protocolVersion == ProtocolVersion.V5 : "MetadataChanged flag is supported starting from v5";
```
adutra: Nit: `MetadataChanged` -> `METADATA_CHANGED`

adutra (commented):
A couple of minor concerns + an overall problem of code formatting.

adutra:
Went ahead and implemented my suggestions, and also refactored `PreparedId` a bit. Ready to review.

adutra:
Squashed and rebased.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -2208,16 +2208,35 @@ public void ensurePoolsSizing() {
         }
 
         public PreparedStatement addPrepared(PreparedStatement stmt) {
```
olim7t, line 2220: Do we still need to change this method?
It's not very important to handle race conditions here; even if we store an instance that has the wrong `resultSetMetadata`, the server will inform us in the first response and we'll mutate the instance then.

olim7t (changes requested):
Looks great overall, I think the cache update can be simplified.
One thing I wonder is if a paging query can be affected if the metadata changes between two pages, I'm going to test that.

olim7t:
It is indeed possible to observe a metadata change between two pages of the same `ArrayBackedResultSet.MultiPage`.
I think we should handle that, it will require a bit of refactoring to make the metadata mutable, and ensure that we make the switch at the right time.

olim7t:
> It is indeed possible to observe a metadata change between two pages of the same ArrayBackedResultSet.MultiPage

I'm working on a fix for this.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
adutra: Indeed, but at least the new version would not print a log warning about re-preparing the same statement, if it's being re-prepared because the resultset metadata has changed.

adutra:
👍 lgtm

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
olim7t: > if it's being re-prepared because the resultset metadata has changed.

This is no longer the case. When the resultset metadata has changed, we don't re-prepare and re-add the statement, only mutate it in place.
So the only way `addPrepared` is called twice is if the client explicitly called `session.prepare` twice with the same statement. Whether the resultset metadata changed or not between the two calls is irrelevant I think, this is still an anti-pattern.

olim7t (approved):

adutra (approved):

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -567,16 +566,23 @@ else if (fetchSize != Integer.MAX_VALUE)
...
             if (protocolVersion.compareTo(ProtocolVersion.V4) < 0)
                 bs.ensureAllSet();
-            boolean skipMetadata = protocolVersion != ProtocolVersion.V1 && bs.statement.getPreparedId().resultSetMetadata != null;
-            Requests.QueryProtocolOptions options = new Requests.QueryProtocolOptions(Message.Request.Type.EXECUTE, consistency, Arrays.asList(bs.wrapper.values), Collections.<String, ByteBuffer>emptyMap(),
-                    skipMetadata, fetchSize, usedPagingState, serialConsistency, defaultTimestamp);
-            request = new Requests.Execute(bs.statement.getPreparedId().id, options, statement.isTracing());
+
+            // skip resultset metadata if version > 1 (otherwise this feature is not supported)
+            // and if we already have metadata for the prepared statement being executed.
+            boolean skipMetadata = protocolVersion != ProtocolVersion.V1 && bs.statement.getPreparedId().resultSetMetadata.variables != null;
```
newkek, line 586: Isn't the driver supposed to always receive results metadata after a `PREPARE` message? I was wondering if the check for `bs.statement.getPreparedId().resultSetMetadata.variables != null` wasn't redundant.

driver-core/src/main/java/com/datastax/driver/core/Responses.java
```diff
@@ -380,6 +384,17 @@ static Metadata decode(ByteBuf body, boolean withPkIndices, ProtocolVersion prot
                     EnumSet<Flag> flags = Flag.deserialize(body.readInt());
                     int columnCount = body.readInt();
 
+                    ByteBuffer state = null;
+                    if (flags.contains(Flag.HAS_MORE_PAGES))
+                        state = CBUtil.readValue(body);
+
+                    MD5Digest resultMetadataId = null;
+                    if (flags.contains(Flag.METADATA_CHANGED)) {
+                        assert protocolVersion == ProtocolVersion.V5 : "METADATA_CHANGED flag is supported starting from v5";
```
newkek: Is it possible that we would ever receive from Cassandra the `METADATA_CHANGED` with protocol < V5? Maybe that check isn't necessary?

ifesdjeen: This is the point of the assertion: we can not receive it. If we somehow did - throw assertion error.

ifesdjeen:
It looks like the `cluster` is holding the old metadata, even after the session reconnect and table destruction, so the tests weren't idempotent (but which also means that technically re-creating the table with the same exact name might cause trouble). I've added `cluster` instance reset for now, but I think we should take a closer look at why that might be happening.

ifesdjeen:
Added a fix for the problem described above. This is mostly relevant/realistic for tests, but theoretically may happen in real life as well.

When the following sequence of events happens:
  * create a cluster; create a table with schema `schema`;
  * prepare statement (receive an id of `hash1` / `resultset1`)
  * add a column 
  * try running a `SELECT` query
  * ensure that the metadata part of the id is now `resultset2`

So far, this is pretty much the issue description (all good). Now, the interesting part starts: 

  * drop the table; re-create the table (with the same `schema` as the initial one); 
  * try preparing the statement again, you'll notice that the metadata part of the id is again `resultset2`

Of course the resultset metadata should have been `resultset1`. 

The problem is that (I assume) for performance reasons, [here](https://github.com/datastax/java-driver/pull/794/commits/a4620a4c3eb66d0e3e4ad06e80564b57360744d7#diff-3c61240d382b855ddcdc43fa82420d28R2227) the "old" statement is returned (comment mentions since it's in cache, so I assume the CPU cache is meant. Problem is that while it was correct in the previous setup: same statement id == same statements, it's not true anymore, since the metadata might have changed. I've added a check to compare metadata of the returning statement to make sure we return the correct one and skip the optimisation when it is not possible to use it.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1388,7 +1388,7 @@ private static String generateClusterName() {
         // new one join the cluster).
         // Note: we could move this down to the session level, but since prepared statement are global to a node,
         // this would yield a slightly less clear behavior.
-        ConcurrentMap<MD5Digest, PreparedStatement> preparedQueries;
+        public ConcurrentMap<MD5Digest, PreparedStatement> preparedQueries;
```
olim7t: Looks unrelated to the latest change, is it a leftover?

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -2218,14 +2218,15 @@ public void ensurePoolsSizing() {
...
+            PreparedStatement previous = preparedQueries.putIfAbsent(stmt.getPreparedId().boundValuesMetadata.id, stmt);
             if (previous != null) {
                 logger.warn("Re-preparing already prepared query is generally an anti-pattern and will likely affect performance. "
                         + "Consider preparing the statement only once. Query='{}'", stmt.getQueryString());
 
                 // The one object in the cache will get GCed once it's not referenced by the client anymore since we use a weak reference.
                 // So we need to make sure that the instance we do return to the user is the one that is in the cache.
-                return previous;
+                if (com.google.common.base.Objects.equal(previous.getPreparedId().resultSetMetadata.id, stmt.getPreparedId().resultSetMetadata.id))
+                    return previous;
```
olim7t: I don't think it's the right way to fix the issue you described. We're returning the correct PreparedStatement to the caller of this method, but leaving the cached instance with the old result metadata. Rather it should be:
```java
previous.getPreparedId().resultSetMetadata = stmt.getPreparedId().resultSetMetadata;
return previous;
```

olim7t (changes requested):

driver-core/src/main/java/com/datastax/driver/core/ArrayBackedResultSet.java
```diff
@@ -62,15 +62,34 @@ static ArrayBackedResultSet fromMessage(Responses.Result msg, SessionManager ses
...
+                    // only ever sets the flag SKIP_METADATA to true for bound statements,
+                    // never for regular ones.
                     Statement actualStatement = statement;
                     if (statement instanceof StatementWrapper) {
                         actualStatement = ((StatementWrapper) statement).getWrappedStatement();
                     }
-                    assert actualStatement instanceof BoundStatement;
-                    columnDefs = ((BoundStatement) actualStatement).statement.getPreparedId().resultSetMetadata;
+                    assert statement instanceof BoundStatement;
+                    columnDefs = ((BoundStatement) statement).statement.getPreparedId().resultSetMetadata.variables;
```
olim7t: Should be `actualStatement` on those two lines.

driver-core/src/main/java/com/datastax/driver/core/ArrayBackedResultSet.java
```diff
@@ -62,15 +62,34 @@ static ArrayBackedResultSet fromMessage(Responses.Result msg, SessionManager ses
...
                     assert columnDefs != null;
                 } else {
+                    // metadata is always present for regular statements (QUERY requests);
+                    // it might also be present for bound statements (EXECUTE requests)
+                    // if the metadata has changed server-side.
                     columnDefs = r.metadata.columns;
+                    if (statement instanceof BoundStatement) {
```
olim7t: We should unwrap like in the other branch.

olim7t (changes requested):
Sorry, I submitted my previous review too soon :(
Two more comments.

olim7t (approved):

olim7t:
I went ahead and implemented my suggestions.

ifesdjeen:
Thank you a lot @olim7t!

olim7t:
Added `ProtocolFeature` enum and rebased on current 3.x.

driver-core/src/main/java/com/datastax/driver/core/ProtocolFeature.java
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Determines whether or not the input version supports ths feature.
+     *
+     * @param version the version to test against.
+     * @return true if supported, false otherwise.
+     */
+    boolean isSupportedBy(ProtocolVersion version) {
+        switch (this) {
+            case PREPARED_METADATA_CHANGES:
+                return version == ProtocolVersion.V5;
```
ifesdjeen, line 40: Might be worth checking for `>`, will be hard to catch when  we bump the version again...

ifesdjeen:
Are we waiting for a follow-up Cassandra ticket to be resolved?.. It'd be good to get it committed at some point, as we might have more v5 changes.

olim7t:
Rebased on 3.3.x + changed base + fixed issues with conditional updates.

PR #796, opened 2017-02-08
JAVA-1364: Enable creation of SslHandler with remote address information

adutra:
Motivation:

SSLOptions do not currently allow to create SslHandler instances with remote address information.
This is needed to instantiate SSLEngine/SslHandler with remote address information set,
which enables doing hostname verification (for example through a custom TrustManager).
We cannot take this information from SocketChannel since it is not connected yet
at the time Connection.Initializer creates the SslHandler.

Modifications:

Create a child interface to SSLOptions, EndpointAwareSSLOptions, that exposes a new
method to create SslHandler instances with remote address information.
Modify Connection class so that it detects instances of EndpointAwareSSLOptions
and calls the new method instead of the old one.
Modify current implementations of SSLOptions, JdkSSLOptions and NettySSLOptions
to implement the new interface; delegate old methods to new ones.

Result:

Remote endpoint information is now available to implementors of EndpointAwareSSLOptions,
and to the driver built-in implementations JdkSSLOptions and NettySSLOptions.

This commit has substantial contributions from Tim Lamballais Tessensohn (@wimtie).

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/JdkSSLOptions.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/JdkSSLOptions.java
@@ -24,7 +24,11 @@ import java.security.NoSuchAlgorithmException;
 
 /**
  * {@link SSLOptions} implementation based on built-in JDK classes.
+ *
+ * @deprecated Use {@link RemoteEndpointAwareJdkSSLOptions} instead.
  */
+@SuppressWarnings("DeprecatedIsStillUsed")
+@Deprecated
 public class JdkSSLOptions implements SSLOptions {
 
     /**
@@ -36,8 +40,8 @@ public class JdkSSLOptions implements SSLOptions {
         return new Builder();
     }
 
-    private final SSLContext context;
-    private final String[] cipherSuites;
+    protected final SSLContext context;
+    protected final String[] cipherSuites;
 
     /**
      * Creates a new instance.
@@ -66,7 +70,7 @@ public class JdkSSLOptions implements SSLOptions {
      * @param channel the Netty channel for that connection.
      * @return the engine.
      */
-    protected SSLEngine newSSLEngine(SocketChannel channel) {
+    protected SSLEngine newSSLEngine(@SuppressWarnings("unused") SocketChannel channel) {
         SSLEngine engine = context.createSSLEngine();
         engine.setUseClientMode(true);
         if (cipherSuites != null)
@@ -86,8 +90,8 @@ public class JdkSSLOptions implements SSLOptions {
      * Helper class to build JDK-based SSL options.
      */
     public static class Builder {
-        private SSLContext context;
-        private String[] cipherSuites;
+        protected SSLContext context;
+        protected String[] cipherSuites;
 
         /**
          * Set the SSL context to use.
@@ -124,6 +128,7 @@ public class JdkSSLOptions implements SSLOptions {
          *
          * @return the new instance.
          */
+        @SuppressWarnings("deprecation")
         public JdkSSLOptions build() {
             return new JdkSSLOptions(context, cipherSuites);
         }
--- a/driver-core/src/main/java/com/datastax/driver/core/RemoteEndpointAwareSSLOptions.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/RemoteEndpointAwareSSLOptions.java
@@ -1,0 +1,53 @@
+/*
+ *      Copyright (C) 2012-2015 DataStax Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+package com.datastax.driver.core;
+
+import io.netty.channel.socket.SocketChannel;
+import io.netty.handler.ssl.SslHandler;
+
+import java.net.InetSocketAddress;
+
+/**
+ * Child interface to {@link SSLOptions} with the possibility to pass remote endpoint data
+ * when instantiating {@link SslHandler}s.
+ * <p/>
+ * This is needed when e.g. hostname verification is required.
+ * See <a href="https://datastax-oss.atlassian.net/browse/JAVA-1364">JAVA-1364</a> for details.
+ * <p/>
+ * The reason this is a child interface is to keep {@link SSLOptions} backwards-compatible.
+ * This interface may be be merged into {@link SSLOptions} in a later major release.
+ *
+ * @see <a href="https://datastax-oss.atlassian.net/browse/JAVA-1364">JAVA-1364</a>
+ * @since 3.2.0
+ */
+public interface RemoteEndpointAwareSSLOptions extends SSLOptions {
+
+    /**
+     * Creates a new SSL handler for the given Netty channel and the given remote endpoint.
+     * <p/>
+     * This gets called each time the driver opens a new connection to a Cassandra host. The newly created handler will be added
+     * to the channel's pipeline to provide SSL support for the connection.
+     * <p/>
+     * You don't necessarily need to implement this method directly; see the provided implementations:
+     * {@link RemoteEndpointAwareJdkSSLOptions} and {@link RemoteEndpointAwareNettySSLOptions}.
+     *
+     * @param channel        the channel.
+     * @param remoteEndpoint the remote endpoint address.
+     * @return a newly-created {@link SslHandler}.
+     */
+    SslHandler newSSLHandler(SocketChannel channel, InetSocketAddress remoteEndpoint);
+
+}
--- a/manual/ssl/README.md
+++ b/manual/ssl/README.md
@@ -71,7 +71,7 @@ if you've followed the steps for inter-node encryption).
 
 ### Driver configuration
 
-The base class to configure SSL is [SSLOptions]. It's very generic, but
+The base class to configure SSL is [RemoteEndpointAwareSSLOptions]. It's very generic, but
 you don't necessarily need to deal with it directly: the default
 instance, or the provided subclasses, might be enough for your needs.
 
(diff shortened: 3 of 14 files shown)


driver-core/src/main/java/com/datastax/driver/core/RemoteEndpointAwareSSLOptions.java
```diff
@@ -0,0 +1,62 @@
...
+     * {@inheritDoc}
+     *
+     * @deprecated use {@link #newSSLHandler(SocketChannel, InetSocketAddress)} instead.
+     */
+    @Deprecated
+    @Override
+    SslHandler newSSLHandler(SocketChannel channel);
+
+    /**
+     * Creates a new SSL handler for the given Netty channel and the given remote remoteEndpoint.
```
olim7t: > remote remoteEndpoint

typo?

driver-core/src/main/java/com/datastax/driver/core/JdkSSLOptions.java
```diff
@@ -52,7 +53,12 @@ protected JdkSSLOptions(SSLContext context, String[] cipherSuites) {
 
     @Override
     public SslHandler newSSLHandler(SocketChannel channel) {
-        SSLEngine engine = newSSLEngine(channel);
+        return newSSLHandler(channel, null);
```
olim7t: Now that the class implements `RemoteEndpointAwareSSLOptions`, this method should never get called by our code. We could throw an `UnsupportedOperationException` or `AssertionError` here to ensure that this is the case.

driver-core/src/main/java/com/datastax/driver/core/RemoteEndpointAwareSSLOptions.java
```diff
@@ -0,0 +1,62 @@
...
+public interface RemoteEndpointAwareSSLOptions extends SSLOptions {
+
+    /**
+     * {@inheritDoc}
+     *
+     * @deprecated use {@link #newSSLHandler(SocketChannel, InetSocketAddress)} instead.
+     */
+    @Deprecated
+    @Override
+    SslHandler newSSLHandler(SocketChannel channel);
```
olim7t: I would go as far as to deprecate it in the parent interface as well, with a comment that points to this interface. Any new implementation of the SSL options should use this interface.

olim7t (changes requested):

driver-core/src/main/java/com/datastax/driver/core/JdkSSLOptions.java
adutra: Agreed.

However your remark made me think of the following situation: If someone is extending `JdkSSLOptions` and overriding this method, or the method `newSSLEngine`, he will experience runtime problems because his overridden method will not be called anymore. Maybe something worth mentioning in the upgrade guide?

adutra: I'm even wondering if this wouldn't be a showstopper actually. I think it would be safer to create subclasses, i.e. `RemoteEndpointAwareJdkSSLOptions extends JdkSSLOptions`.

adutra:
Went ahead and implemented my own suggestion.

olim7t (approved):

tolbertam:
@wimtie, I was just experimenting with this PR and I was wondering what your planned usecase was for this?   Your original description states:

> Enable creation of SSLHandler with remote address information. This is
information set, which is enables doing hostname verification (for example
through a custom TrustManager). We cannot take this information from SocketChannel
since it is not connected yet at the time Connection.Initializer creates the
SSLHandler.

I was curious how you were accessing the additional information this provided so I could experiment with that myself.

I was able to make use of this by updating `newSSLEngine` in `RemoteEndpointAwareJdkSSLOptions` to the following:

```java

protected SSLEngine newSSLEngine(@SuppressWarnings("unused") SocketChannel channel, InetSocketAddress remoteEndpoint) {
    SSLEngine engine = remoteEndpoint == null
            ? context.createSSLEngine()
            : context.createSSLEngine(remoteEndpoint.getHostName(), remoteEndpoint.getPort());

    // begin added code
    SSLParameters parameters = engine.getSSLParameters();
    parameters.setEndpointIdentificationAlgorithm("HTTPS");
    engine.setSSLParameters(parameters);
    // end added code
        
    engine.setUseClientMode(true);
    if (cipherSuites != null)
        engine.setEnabledCipherSuites(cipherSuites);
    return engine;
}
```

Are you effectively doing the same in a custom SslHandler or SSLOptions?   I noticed that with this I am able to get certificate exceptions if the hostname of the address doesn't resolve to the CN of the cert provided by the cassandra node:

```
Caused by: java.security.cert.CertificateException: No name matching localhost found
    at sun.security.util.HostnameChecker.matchDNS(HostnameChecker.java:221)
    at sun.security.util.HostnameChecker.match(HostnameChecker.java:95)
    at sun.security.ssl.X509TrustManagerImpl.checkIdentity(X509TrustManagerImpl.java:455)
    at sun.security.ssl.X509TrustManagerImpl.checkIdentity(X509TrustManagerImpl.java:436)
    at sun.security.ssl.X509TrustManagerImpl.checkTrusted(X509TrustManagerImpl.java:252)
    at sun.security.ssl.X509TrustManagerImpl.checkServerTrusted(X509TrustManagerImpl.java:136)
    at sun.security.ssl.ClientHandshaker.serverCertificate(ClientHandshaker.java:1496)
    ... 26 more
```

In addition, if `createSSLEngine` is not initialized with the remote address info, I get an NPE:

```
...
Caused by: java.lang.NullPointerException
    at sun.net.util.IPAddressUtil.textToNumericFormatV4(IPAddressUtil.java:49)
    at sun.net.util.IPAddressUtil.isIPv4LiteralAddress(IPAddressUtil.java:241)
    at sun.security.util.HostnameChecker.isIpAddress(HostnameChecker.java:124)
    at sun.security.util.HostnameChecker.match(HostnameChecker.java:92)
    at sun.security.ssl.X509TrustManagerImpl.checkIdentity(X509TrustManagerImpl.java:455)
    at sun.security.ssl.X509TrustManagerImpl.checkIdentity(X509TrustManagerImpl.java:436)
    at sun.security.ssl.X509TrustManagerImpl.checkTrusted(X509TrustManagerImpl.java:252)
    at sun.security.ssl.X509TrustManagerImpl.checkServerTrusted(X509TrustManagerImpl.java:136)
    at sun.security.ssl.ClientHandshaker.serverCertificate(ClientHandshaker.java:1496)
    at sun.security.ssl.ClientHandshaker.processMessage(ClientHandshaker.java:216)
    at sun.security.ssl.Handshaker.processLoop(Handshaker.java:979)
    at sun.security.ssl.Handshaker$1.run(Handshaker.java:919)
    at sun.security.ssl.Handshaker$1.run(Handshaker.java:916)
    at java.security.AccessController.doPrivileged(Native Method)
    at sun.security.ssl.Handshaker$DelegatedTask.run(Handshaker.java:1369)
    at io.netty.handler.ssl.SslHandler.runDelegatedTasks(SslHandler.java:1124)
    at io.netty.handler.ssl.SslHandler.unwrap(SslHandler.java:1009)
    ... 18 more 
```

Was that the original impetus for this request?  If I'm on the mark there, I think this might be generally useful enough that we put the logic to enable hostname verification into the SSLOptions itself, what do you think?   We could make the case that this be the default behavior here (although maybe not until the next major release).  I have something working locally with tests that I can push up if you agree.  The additional option added would be `withHostnameVerification()` or something to that extent.

tolbertam:
Ah, I just realized we only have access to `SSLEngine` in JDK-based SSL options.  In the Netty-based options, I suppose something would have to be done in a custom `SslHandler` or through Netty's `SslContext` (i.e. `SslContext.newHandler(SSLEngine)`) somehow, is that right?  Guessing we could do the same there too.

manual/ssl/README.md
```diff
@@ -178,15 +177,16 @@ SslContextBuilder builder = SslContextBuilder
...
   .addContactPoint("127.0.0.1")
   .withSSL(sslOptions)
   .build();
 ```
 
-[SSLOptions]: http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/core/SSLOptions.html
-[JdkSSLOptions]: http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/core/JdkSSLOptions.html
-[NettySSLOptions]: http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/core/NettySSLOptions.html
-[NettyOptions]: http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/core/NettyOptions.html
+[RemoteEndpointAwareSSLOptions]:      http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/core/RemoteEndpointAwareSSLOptions.html
```
adutra, line 188: All these urls should be updated to 3.2, but I guess we can do that later when releasing 3.2.0 (I bet many other manual pages will have to be updated as well).

wimtie:
@tolbertam, we are actually using a custom TrustManager since we have IP-based certificates on our Cassandra boxes. But yeah, the general idea is the same as what you posted above.

tolbertam:
@wimtie, thanks for confirming!  How are you able to get at the remote address information from the `TrustManager` api?  I see that `checkServerTrusted` is given array of `X509Certificate`s and an `authType`, but no address information.

adutra:
I think that if your trust manager is a `X509ExtendedTrustManager` then you will have access to the `SSLSocket`, see [here](https://docs.oracle.com/javase/7/docs/api/javax/net/ssl/X509ExtendedTrustManager.html#checkServerTrusted(java.security.cert.X509Certificate[],%20java.lang.String,%20java.net.Socket)).

Then you can access peer info:

```
public void checkServerTrusted(X509Certificate[] chain,
                               String authType,
                               Socket socket) throws CertificateException {
    SSLSocket sslSocket = ((SSLSocket) socket);
    SSLSession session = sslSocket.getHandshakeSession();
    String peerHost = session.getPeerHost();
    int peerPort = session.getPeerPort();
}
```

tolbertam:
@adutra , I see, I overlooked that, that makes sense then, thanks!

wimtie:
@tolbertam FWIW, we override

`public void checkServerTrusted(X509Certificate[] chain, String authType, SSLEngine engine)`

from X509ExtendedTrustManager, where engine has `engine.getPeerHost()` `engine.getPeerPort()`

tolbertam:
Thanks!  I see that without passing the address into `createSSLEngine` the host info is indeed not there.  I'll push up a test for that.  I think we should also consider adding `withHostnameVerification(boolean enabled)` to the builders as well as I think it is generally useful, but will discuss that with the team 👍

tolbertam:
Just realized my test doesn't work with Java 6.  I'll set it up so it only runs against JDK 8 like some of our other tests this afternoon.

tolbertam:
Rebased on 3.x to fix conflict in changelog

tolbertam (approved):
Looks good to me 👍 .  I created a separate jira ticket ([JAVA-1403](https://datastax-oss.atlassian.net/browse/JAVA-1403)) as a follow on for adding an option to enforce hostname validation for future consideration.

PR #797, opened 2017-02-09
JAVA-1395-allow to disable request's timer metric for some statements

jiafu1115:
refer to :

https://datastax-oss.atlassian.net/browse/JAVA-1395?filter=-2

I hadn't clean up the code such as adding test code and squashing the commit due to I am n't sure if it is one improvement or requirement you also think it is need. So just write code demo for the issue-1395 to show the code change scope.

Thanks in advanced. if agree with this requirement, I will clean up this PR.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
@@ -85,7 +85,7 @@ class RequestHandler {
                 && statement.isIdempotentWithDefault(manager.configuration().getQueryOptions());
         this.statement = statement;
 
-        this.timerContext = metricsEnabled() && statement.isMetricing()
+        this.timerContext = metricsEnabled()
                 ? metrics().getRequestsTimer().time()
                 : null;
         this.startTime = System.nanoTime();
@@ -212,7 +212,7 @@ class RequestHandler {
     }
 
     private boolean metricsEnabled() {
-        return manager.configuration().getMetricsOptions().isEnabled();
+        return manager.configuration().getMetricsOptions().isEnabled() && statement.isMetricing();
     }
 
     private Metrics metrics() {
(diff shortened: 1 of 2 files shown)


tolbertam:
Hi @jiafu1115 , thank you for your contribution!   This does seem generally useful, but I wonder if it would be better to create a timer outside of the driver to measure time for specific queries instead of excluding queries from the timer in the driver itself. 

The metrics made available in the driver are more intended more for measuring the driver itself, rather than business logic in a user's application.   The metrics framework is really nice and I find it easy enough to add my own metrics outside of the driver, and I think in general that is probably a better choice for the user, since a Timer that is measured outside of the driver will be a more accurate representation since it would include the entire roundtrip time from when you submit a request and get a completed future from the driver, where the request timer in the driver just measures the time between when `RequestHandler` is created and it gets and parses the response from cassandra (before the future is actually completed).

jiafu1115:
@tolbertam  

Thanks for your detailed feedback. 

I agree with that metric outsite of driver may be more useful to trouble shooting or meause the acural operation.

Thus, consider the actural requirements, I only want to measure the overral performance for cassandra without involoved another metric outside of driver. But the keepalive or health check(select time or other simple command) influence the measure result a lot by high executation rate. So I only want to disable the "health check"'s measure.

Consider the inner metirc timer's usage, I think if no consider the driver's performance tunning requirement. It's goal must be to show the overral performace for all cassandra's operation.Thus, we want to disable some operation expecially about health check. or it is usefulness for the request timer metric for customer. If it is usefulness for this driver's request timer, we need to disable it but we want other metirs such as connnection count and error count,etc. now they are be providered together.

So WDYT now?  Thanks for your feedback. I still think most applications want to relay on the inner request timers metric to measue the overall performance for the cassandra operation. But most application don't want to meature theire healch check operation and they does't want involved other metric outside of the driver with more code changes. What's more, they believe the driver's metric.

BTW: It is only about  <10 lines code change to meet this requirements without any regression risk. but if our application need to support it we should add much more codes and disable the inner metrics, but we need other metics such as guawa.counter existed in driver.

tolbertam:
4.0.0-alpha3 was released earlier this week (see [manual](https://docs.datastax.com/en/developer/java-driver/4.0-alpha/manual/core/metrics/)).  The API is flexible enough in that you can completely control how metrics are updated, so you could do something such as this.  I don't think we want to change the existing Metrics behavior for 3.x, so closing this.

jiafu1115:
thanks

PR #798, opened 2017-02-09
JAVA-1313: Copy SerialConsistencyLevel to PreparedStatement

adutra:
Motivation:

A serial consistency level set on a statement that is going
to be prepared is not being propagated to the prepared statement.

Modifications:

Modify AbstractSession.prepareAsync(RegularStatement) and propagate
the serial consistency level to prepared statements, just as for
other inherited properties.

Result:

Serial consistency level is now being propagated to prepared
statements.

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/PreparedStatementTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/PreparedStatementTest.java
@@ -17,7 +17,10 @@ package com.datastax.driver.core;
 
 import com.datastax.driver.core.exceptions.InvalidQueryException;
 import com.datastax.driver.core.exceptions.UnsupportedFeatureException;
+import com.datastax.driver.core.policies.FallthroughRetryPolicy;
+import com.datastax.driver.core.utils.Bytes;
 import com.datastax.driver.core.utils.CassandraVersion;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.util.concurrent.Uninterruptibles;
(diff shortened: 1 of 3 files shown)


olim7t:
While we're at it, it would also make sense to propagate the keyspace (since we do for the routing key already), and the fetch size.
We'll need to add setters to PreparedStatement.

adutra:
> While we're at it, it would also make sense to propagate the keyspace (since we do for the routing key already), and the fetch size. We'll need to add setters to PreparedStatement.

Unfortunately we cannot add methods to `PreparedStatement` as it is an interface, and that would be a breaking change.

Furthermore, the keyspace is already stored in the prepared statement metadata and propagated to bound statements.

driver-core/src/test/java/com/datastax/driver/core/PreparedStatementTest.java
```diff
@@ -334,13 +338,30 @@ public void prepareStatementInheritPropertiesTest() {
 
         RegularStatement toPrepare = new SimpleStatement("SELECT * FROM test WHERE k=?");
         toPrepare.setConsistencyLevel(ConsistencyLevel.QUORUM);
+        toPrepare.setSerialConsistencyLevel(ConsistencyLevel.LOCAL_SERIAL);
```
tolbertam, line 341: Nice, I like that you tested that all the other properties get propagated as well 👍

tolbertam (approved):
Looks good 👍

PR #800, opened 2017-02-09
JAVA-1241: Upgrade Netty to 4.1.x

adutra:
This commit also updates netty-tcnative to Fork26, and
accounts for netty-transport depending on new netty-resolver module in 4.1.


tolbertam (approved):
👍 ing since I was 👍 on the previous PR.  I'd like to run some perf tests against this to make sure there are no glaring performance degradations, but I expect that to be very unlikely.   Since the scope of changes is very small, I think it's ok to merge and if in the oft-chance we hit any problems while perf testing we can always back out then.

PR #801, opened 2017-02-10
JAVA-1397: Handle duration as native datatype in protocol v5+

olim7t:
Protocol v5 is still beta, I temporarily added this (not committed) in DurationIntegrationTest to test:
```java
    @Override
    public Cluster.Builder createClusterBuilder() {
        return super.createClusterBuilder().allowBetaProtocolVersion();
    }
```

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/DataType.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/DataType.java
@@ -17,10 +17,8 @@ package com.datastax.driver.core;
 
 import com.datastax.driver.core.exceptions.DriverInternalError;
 import com.google.common.base.Objects;
-import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Sets;
 import io.netty.buffer.ByteBuf;
 
 import java.util.*;
@@ -35,6 +33,7 @@ public abstract class DataType {
      */
     public enum Name {
 
+        CUSTOM(0),
         ASCII(1),
         BIGINT(2),
         BLOB(3),
@@ -43,7 +42,6 @@ public abstract class DataType {
         DECIMAL(6),
         DOUBLE(7),
         FLOAT(8),
-        INET(16),
         INT(9),
         TEXT(10) {
             @Override
@@ -61,16 +59,17 @@ public abstract class DataType {
         },
         VARINT(14),
         TIMEUUID(15),
-        LIST(32),
-        SET(34),
-        MAP(33),
-        CUSTOM(0),
-        UDT(48, ProtocolVersion.V3),
-        TUPLE(49, ProtocolVersion.V3),
+        INET(16),
+        DATE(17, ProtocolVersion.V4),
+        TIME(18, ProtocolVersion.V4),
         SMALLINT(19, ProtocolVersion.V4),
         TINYINT(20, ProtocolVersion.V4),
-        DATE(17, ProtocolVersion.V4),
-        TIME(18, ProtocolVersion.V4);
+        DURATION(21, ProtocolVersion.V5),
+        LIST(32),
+        MAP(33),
+        SET(34),
+        UDT(48, ProtocolVersion.V3),
+        TUPLE(49, ProtocolVersion.V3);
 
         final int protocolId;
 
@@ -147,6 +146,7 @@ public abstract class DataType {
         primitiveTypeMap.put(Name.TINYINT, new DataType.NativeType(Name.TINYINT));
         primitiveTypeMap.put(Name.DATE, new DataType.NativeType(Name.DATE));
         primitiveTypeMap.put(Name.TIME, new DataType.NativeType(Name.TIME));
+        primitiveTypeMap.put(Name.DURATION, new DataType.NativeType(Name.DURATION));
     }
 
     private static final Set<DataType> primitiveTypeSet = ImmutableSet.copyOf(primitiveTypeMap.values());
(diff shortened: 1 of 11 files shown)


olim7t:
I think we should also update DataTypeIntegrationTest and similar, looking into that.

olim7t:
In hindsight, it's better to keep duration tests separate because the type has specific restrictions (e.g. can't be used in collections) so it can't be handled as a "regular" primitive type by DataTypeIntegrationTest.
I've added the logic to exclude it so that things don't break when v5 becomes non-beta.

driver-core/src/test/java/com/datastax/driver/core/DataTypeTest.java
```diff
@@ -331,7 +331,7 @@ public void should_not_return_v4_types_in_all_primitive_types_with_v3() {
         }
     }
 
-    @Test(groups = "unit")
+    @Test(groups = "unit", enabled = false)
     public void should_return_same_elements_with_all_primitive_types_using_latest_protocol_version() {
         assertThat(DataType.allPrimitiveTypes(ProtocolVersion.NEWEST_SUPPORTED)).isEqualTo(DataType.allPrimitiveTypes());
```
olim7t: This fails, I'm not sure if we should adapt it or remove it altogether.

driver-core/src/main/java/com/datastax/driver/core/DataType.java
```diff
@@ -61,16 +60,18 @@ public boolean isCompatibleWith(Name that) {
...
         SET(34),
         MAP(33),
         CUSTOM(0),
         UDT(48, ProtocolVersion.V3),
-        TUPLE(49, ProtocolVersion.V3),
-        SMALLINT(19, ProtocolVersion.V4),
-        TINYINT(20, ProtocolVersion.V4),
-        DATE(17, ProtocolVersion.V4),
-        TIME(18, ProtocolVersion.V4);
+        TUPLE(49, ProtocolVersion.V3);
```
adutra, line 72: If you re-order by protocolId, why not fully re-order?

driver-core/src/main/java/com/datastax/driver/core/DataType.java
```diff
@@ -593,7 +598,10 @@ public boolean isCollection() {
 
             @Override
             public boolean apply(DataType dataType) {
-                return protocolVersion.compareTo(dataType.getName().minProtocolVersion) >= 0;
```
adutra: This method is only used in tests, therefore I would suggest that we move it to `TestUtils`, wdyt? To buy us some time, I will push a commit to do that myself, but feel free to revert if you disagree.

driver-core/src/main/java/com/datastax/driver/core/DataType.java
```diff
@@ -593,7 +598,10 @@ public boolean isCollection() {
...
+                // Duration is handled separately in DurationIntegrationTest, because it has specific restrictions (e.g.
+                // not allowed in collections).
+                return dataType.getName() != Name.DURATION &&
```
adutra: Also, I think the "base" method should return all types for a given version, and let individual tests filter out undesired ones, like in `DataTypeTest.exclude(DataType)`.

driver-core/src/test/java/com/datastax/driver/core/DataTypeTest.java
adutra: I vote for removing it. The method under test – `DataType.allPrimitiveTypes(ProtocolVersion)` – is itself a test method and the assumption – `NEWEST_SUPPORTED` should contain all known primitive types – is simply not true anymore.

driver-core/src/main/java/com/datastax/driver/core/DataType.java
```diff
@@ -775,24 +783,4 @@ public String toString() {
         }
     }
 
-    /**
-     * The Duration type, introduced in Cassandra 3.10.
-     * <p/>
-     * This class is a singleton; to obtain its unique instance,
-     * call {@link #duration()}.
-     */
-    public static class DurationType extends CustomType {
```
adutra, line 784: Pushed a small commit to delete this.

adutra (approved):
Approving my own changes \o/

driver-core/src/main/java/com/datastax/driver/core/DataType.java
```diff
@@ -579,26 +582,6 @@ public boolean isCollection() {
...
-     * Returns a set of all primitive types supported by the given protocolVersion.
-     * <p>
-     * Primitive types are defined as the types that don't have type arguments
-     * (that is excluding lists, sets, and maps, tuples and udts).
-     * </p>
-     *
-     * @param protocolVersion protocol version to get types for.
-     * @return returns a set of all the primitive types for the given protocolVersion.
-     */
-    static Set<DataType> allPrimitiveTypes(final ProtocolVersion protocolVersion) {
```
adutra, line 591: Moved to TestUtils as per my own suggestion.

driver-core/src/main/java/com/datastax/driver/core/DataType.java
olim7t: Right, I missed a few of them.

driver-core/src/test/java/com/datastax/driver/core/DataTypeTest.java
olim7t: `should_not_return_v4_types_in_all_primitive_types_with_v3` (right above) could be deleted too.

driver-core/src/test/java/com/datastax/driver/core/DataTypeTest.java
```diff
@@ -317,22 +317,4 @@ public void serializeDeserializeCollectionsTest(ProtocolVersion version) {
...
-        // Ensure all values are <= V3.
-        for (DataType dataType : dataTypes) {
-            assertThat(dataType.getName().minProtocolVersion).isLessThanOrEqualTo(ProtocolVersion.V3);
-        }
-    }
-
-    @Test(groups = "unit")
-    public void should_return_same_elements_with_all_primitive_types_using_latest_protocol_version() {
-        assertThat(DataType.allPrimitiveTypes(ProtocolVersion.NEWEST_SUPPORTED)).isEqualTo(DataType.allPrimitiveTypes());
-    }
```
olim7t, line 337: Removed these two tests, as they cover a method that has been moved to TestUtils.

tolbertam (approved):
looks good to me! 👍

PR #803, opened 2017-02-12
Improve mapping manager synchronization

avivcarmis:
following discussion on [previous PR](https://github.com/datastax/java-driver/pull/772).


adutra (approved):
👍

tolbertam (approved):
looks great to me, just a note to make sure when merging to get rid of the 4ca1737 commit which just looks like a trivial no-change merge commit.

driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java
```diff
@@ -220,15 +222,11 @@ public Session getSession() {
...
-                    mapper = new Mapper<T>(this, klass, entityMapper);
-                    Map<Class<?>, Mapper<?>> newMappers = new HashMap<Class<?>, Mapper<?>>(mappers);
-                    newMappers.put(klass, mapper);
-                    mappers = newMappers;
-                }
+            EntityMapper<T> entityMapper = AnnotationParser.parseEntity(klass, this);
+            mapper = new Mapper<T>(this, klass, entityMapper);
+            Mapper<T> old = (Mapper<T>) mappers.putIfAbsent(klass, mapper);
+            if (old != null) {
+                mapper = old;
```
olim7t, line 229: This is not exactly the same as the previous version, now we create a new instance each time, even if the entry existed in the map.
For `getMapper` and `getAccessor` it doesn't matter that much, because they're only called directly by the client (and in general they will probably only do it once), but `getUDTCodec` is also called internally, each time a new class with an UDT field is parsed.
Checking the map first is simple enough, so I think we should do it in all cases:
```
Mapper<T> old = (Mapper<T>) mappers.get(klass);
if (old != null) {
    return old;
}
EntityMapper<T> entityMapper = AnnotationParser.parseEntity(klass, this);
mapper = new Mapper<T>(this, klass, entityMapper);
old = (Mapper<T>) mappers.putIfAbsent(klass, mapper);
if (old != null) {
    mapper = old;
}
return mapper;
```
It can still create an unnecessary instance, but the window is much shorter.

olim7t (changes requested):

driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java
avivcortica: @olim7t i think you've missed [this line](https://github.com/datastax/java-driver/blob/a542078d4de699ae88f410d3b2d3e715c133d898/driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java#L223) which does exactly what you've suggested - take a look and let me know...

olim7t: 🐼  my bad, I was looking at the diff view on github and I only focused on the changed lines.

olim7t (approved):

PR #804, opened 2017-02-13
PrimitiveTypeSamples Duration fix

tolbertam:
Duration is excluded because it can't be used in collections and udts.  It is tested separately in DurationIntegrationTest.

This doesn't actually impact testing yet so Protocol V5 is currently marked as beta, but eventually it'll tear down the test framework if running against a Cassandra version that supports it.


tolbertam:
Few more issues discovered, please hold off until then.

olim7t (approved):

PR #805, opened 2017-02-14
Backport #791 and remove ambiguity in Cassandra and DSE versions in tests

tolbertam:
Backports test changes from #791 and also removes ambiguity in Cassandra and DSE versions in tests (see b2e51ab for details).

If accepted, will need to make changes on each upstream branch as we go upwards.  I've tested this change on all upstream branches and made the individual fixes, so I should be able to do that quickly when the time comes.

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/CCMAccess.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/CCMAccess.java
@@ -32,13 +32,21 @@ public interface CCMAccess extends Closeable {
     String getClusterName();
 
     /**
-     * Returns the Cassandra version of this CCM cluster.
+     * Returns the Cassandra version of this CCM cluster.  If {@link #getDSEVersion()} is non-null it is assumed
+     * that this value is only used for representing the compatible Cassandra version for that DSE version.
      * <p/>
-     * By default the version is equal to {@link CCMBridge#getCassandraVersion()}.
      *
      * @return The version of this CCM cluster.
      */
-    VersionNumber getVersion();
+    VersionNumber getCassandraVersion();
+
+    /**
+     * Returns the DSE version of this CCM cluster if this is a DSE cluster, otherwise null.
+     * <p/>
+     *
+     * @return The version of this CCM cluster.
+     */
+    VersionNumber getDSEVersion();
 
     /**
      * @return The config directory for this CCM cluster.
@@ -236,4 +244,19 @@ public interface CCMAccess extends Closeable {
      */
     void waitForDown(int node);
 
+    /**
+     * @return The target protocolVersion to use when connecting to this CCM cluster.
+     * <p/>
+     * This should be based on the highest protocol version that both the cluster and driver support.
+     * <p/>
+     * For example, C* 2.0.17 should return {@link ProtocolVersion#V2} since C* supports up to V2 and the driver
+     * supports that version.
+     */
+    ProtocolVersion getProtocolVersion();
+
+    /**
+     * @param maximumAllowed The maximum protocol version to use.
+     * @return The target protocolVersion or maximumAllowed if {@link #getProtocolVersion} is greater.
+     */
+    ProtocolVersion getProtocolVersion(ProtocolVersion maximumAllowed);
 }
--- a/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
@@ -282,6 +282,21 @@ public class CCMBridge implements CCMAccess {
     }
 
     /**
+     * @return {@link VersionNumber} representation of {@link CCMBridge#getCassandraVersion()}
+     */
+    public static VersionNumber getCassandraVersionNumber() {
+        return VersionNumber.parse(getCassandraVersion());
+    }
+
+    /**
+     * @param compareVersion The given version to compare with.
+     * @return Whether or not the configured cassandra version is greater than or equal to the given version.
+     */
+    public static boolean isCassandraVersionAtLeast(String compareVersion) {
+        return getCassandraVersionNumber().compareTo(VersionNumber.parse(compareVersion)) >= 0;
+    }
+
+    /**
      * @return The configured DSE version if '-Ddse=true' specified, otherwise null.
      */
     public static String getDSEVersion() {
@@ -290,6 +305,22 @@ public class CCMBridge implements CCMAccess {
         } else {
             return null;
         }
+    }
+
+    /**
+     * @return {@link VersionNumber} representation of {@link CCMBridge#getDSEVersion()}
+     */
+    public static VersionNumber getDSEVersionNumber() {
+        return VersionNumber.parse(getDSEVersion());
+    }
+
+    /**
+     * @param compareVersion The given version to compare with.
+     * @return Whether or not the configured dse version is greater than or equal to the given version.
+     */
+    public static boolean isDSEVersionAtLeast(String compareVersion) {
+        VersionNumber dseVersion = getDSEVersionNumber();
+        return dseVersion != null && dseVersion.compareTo(VersionNumber.parse(compareVersion)) >= 0;
     }
 
     /**
@@ -549,7 +580,12 @@ public class CCMBridge implements CCMAccess {
     @Override
     public void decommission(int n) {
         logger.debug(String.format("Decommissioning: node %s (%s%s:%s) from %s", n, TestUtils.IP_PREFIX, n, binaryPort, this));
-        execute(CCM_COMMAND + " node%d decommission", n);
+        // Special case for C* 3.12+, DSE 5.1+, force decommission (see CASSANDRA-12510)
+        String cmd = CCM_COMMAND + " node%d decommission";
+        if ((!isDSE() && isCassandraVersionAtLeast("3.12")) || (isDSE() && isDSEVersionAtLeast("5.1"))) {
+            cmd += " --force";
+        }
+        execute(cmd, n);
     }
 
     @Override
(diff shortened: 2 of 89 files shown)


driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
```diff
@@ -40,7 +40,11 @@
 
     private static final Logger logger = LoggerFactory.getLogger(CCMBridge.class);
 
-    private static final String CASSANDRA_VERSION;
+    private static final String INPUT_CASSANDRA_VERSION;
```
adutra: This is actually only used in the static block below and could be turned into a local variable.

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
```diff
@@ -144,7 +148,7 @@
     private static final String CCM_COMMAND;
 
     static {
-        CASSANDRA_VERSION = System.getProperty("cassandra.version");
+        INPUT_CASSANDRA_VERSION = System.getProperty("cassandra.version");
```
adutra: What about distinguishing two different system properties for default versions to use: one for OSS C*, one for DSE? 

One could e.g. specify that tests requiring DSE should by default use 5.0.3, but tests requiring OSS C* should rather use 3.10.

In this case, the `dse` system property is useless and the mapping from a DSE version to an OSS C* one should only be used when determining the corresponding C* version of a DSE version (which might be different than the "global" C* version).

I can give it a try myself if you don't mind.

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
```diff
@@ -40,7 +40,11 @@
 
     private static final Logger logger = LoggerFactory.getLogger(CCMBridge.class);
 
-    private static final String CASSANDRA_VERSION;
+    private static final String INPUT_CASSANDRA_VERSION;
+
+    private static final VersionNumber CASSANDRA_VERSION_NUMBER;
```
adutra: How about calling these constants `GLOBAL_XXX` to match their getter method names?

driver-core/src/test/java/com/datastax/driver/core/CCMAccess.java
```diff
@@ -34,11 +34,18 @@
...
-    VersionNumber getVersion();
+    VersionNumber getCassandraVersion();
+
+    /**
+     * Returns the DSE version of this CCM cluster.
+     * <p/>
+     *
+     * @return The version of this CCM cluster.
+     */
+    VersionNumber getDSEVersion();
```
adutra, line 49: Not sure how much it makes sense to have a CCM cluster instance reporting two different versions. A CCMAccess is a bridge to either a DSE cluster or a C* cluster, not both.

driver-core/src/test/java/com/datastax/driver/core/CCMAccess.java
```diff
@@ -236,4 +243,14 @@
      */
     void waitForDown(int node);
 
+    /**
+     * @return The desired target protocol version based on the 'cassandra.version' System property.
+     */
+    ProtocolVersion getDesiredProtocolVersion();
```
adutra: Suggestion: replace "desired" – its meaning is quite unclear here – with "target", e.g. `getTargetProtocolVersion` os even simply `getProtocolVersion`.
The javadocs need to be updated, here the returned version is the version that matches this specific instance, not the global system property.

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
```diff
@@ -724,6 +694,26 @@ public void waitForDown(int node) {
         TestUtils.waitUntilPortIsDown(addressOfNode(node));
     }
 
+    @Override
+    public ProtocolVersion getDesiredProtocolVersion() {
+        VersionNumber version = getCassandraVersion();
```
adutra: this is only checking cassandra versions, shouldn't it be checking dse versions as well? Anyway as per my previous comment I think a single CCMBridge instance should carry only one version, and a boolean flag to indicate if it is a DSE cluster or not.

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
```diff
@@ -840,41 +830,26 @@ public Builder notStarted() {
...
-            return this;
-        }
-
-        /**
-         * Sets this cluster to be a non-DSE cluster (defaults to {@link #isDSE()} if this is never called).
+         * The Cassandra version to use.
          */
-        public Builder withoutDSE() {
-            this.isDSE = false;
+        public Builder withCassandraVersion(VersionNumber versionNumber) {
```
adutra: Again, not sure this makes sense, since `withCassandraVersion` and `withDSEVersion` are mutually exclusive (see assertions) why not keep just one method `withVersion`?

adutra (changes requested):
I love your ideas! I just wonder if we go down this path, why not go even further and distinguish DSE from C* from the very top of the stack, i.e. starting with system properties themselves (see my comment in the review).

driver-core/src/test/java/com/datastax/driver/core/CCMAccess.java
adutra: Oh my bad, I see it now, getCassandraVersion() is filled automatically with the corresponding C* version if this instance is a DSE cluster, correct? If so, please disregard my previous comment.

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
tolbertam: I didn't want to change the existing system properties, I'll think on this for a bit

driver-core/src/test/java/com/datastax/driver/core/CCMAccess.java
tolbertam: 👍 in the case the dse version is provided, the cassandra version is just the derived cassandra verson from dse for deriving test behavior based on that version.  I'll update the comments to reflect that.

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
tolbertam: For now the Cassandra version (which is the compatible cassandra version when DSE version is provided) is adequate for determining the protocol version.

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
tolbertam: Agreed, reverting this back as it isn't necessary in this particular case.

tolbertam:
woops, didn't mean to delete this 🤦‍♂️

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
tolbertam: Discussion for system properties is on the `consistentTestVersioning-adu` branch didn't want to address that here.

driver-core/src/test/java/com/datastax/driver/core/CCMAccess.java
olim7t: I would describe it as the highest version common to the driver and the server, is that accurate?

tolbertam: Yep, that is the intent of this

tolbertam: Ah I see that the documentation doesn't completely reflect that, i'll fix that up:

```
 The target protocol version based on the 'cassandra.version' System property.
```

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
```diff
@@ -1075,16 +1105,16 @@ private String randomizePorts(CharSequence str) {
         }
 
         @Override
-        @SuppressWarnings("SimplifiableIfStatement")
         public boolean equals(Object o) {
-            // do not include cluster name and start, only
```
adutra: I think we should keep the comment about cluster name because indeed a cluster name is unique, but two CCM instances should still be considered equal if their characteristics are identical. As for the `start` property, I am not sure anymore, but I think we can include it the computation of equals() and hashcode().

driver-core/src/test/java/com/datastax/driver/core/TestUtils.java
```diff
@@ -838,7 +811,7 @@ public static long getFreeMemoryMB() {
 
     /**
      * Helper for generating a DynamicCompositeType {@link ByteBuffer} from the given parameters.
-     *
+     * <p>
```
adutra: This is minor but I think the convention is to use `<p/>`

adutra (changes requested):
Looks great, only minor remarks 👍

driver-core/src/test/java/com/datastax/driver/core/TestUtils.java
tolbertam: Ah, good catch, IntelliJ's autoformatter behaves differently between when you have JDK 8 configured and otherwise.  For some reason with JDK 8 it omits `/`.  I'll fix it.

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
tolbertam: Good catch will fix

tolbertam: Hrmm it doesn't look like cluster name is even part of the builder anymore since this is randomly generated, so maybe that comment isn't relevant anymore?

adutra: Ah right, silly me :)

adutra (approved):

tolbertam:
Squashed the `Additional Feedback` commits into 202770e, will merge this later today (will let it run through CI for an additional sanity check).

PR #808, opened 2017-02-16
Use Cassandra 3.7 instead of 3.6 in ProtocolVersionRenegotiationTest

tolbertam:
Cassandra 3.6 does not work with > JDK 1.8u100 because of [CASSANDRA-11661](https://issues.apache.org/jira/browse/CASSANDRA-11661), use 3.7 instead.

This was causing this test to repeatedly fail in appveyor.  It does not fail in our jenkins environment because we are using an older JDK update, but I will fix that.

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/ProtocolVersionRenegotiationTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/ProtocolVersionRenegotiationTest.java
@@ -38,8 +38,8 @@ public class ProtocolVersionRenegotiationTest extends CCMTestsSupport {
      * @jira_ticket JAVA-1367
      */
     @Test(groups = "short")
-    @CCMConfig(version = "3.6", createCluster = false)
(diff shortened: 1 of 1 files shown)


adutra (approved):

tolbertam:
Waiting for the appveyor build to see if this fixes the test (pretty confident but not 100%) and if it does I'll merge.

tolbertam:
D'oh!  It didn't work 😢 , it might have something to do with the C* download not being previously downloaded for CCM.  I'll try something else.  I think I may just rewrite this test so it only runs when the configured cassandra version is >= 3.0 since we run the full matrix anyways.

tolbertam:
On second thought I'd like to revisit this test class.  Instead of creating a cluster for each test at a specific version, we can just check the version configured for this test run and then behave appropriately.  The CCM download process is unreliable as mirrors sometimes omit newer C* version bizarrely and we predownload on the images to prevent downloading everytime.  Will do this as we merge the consistentTestVersioning branch up.

PR #809, opened 2017-02-16
JAVA-1355: Update getReplicas to handle range covering multiple hosts

adutra:
Motivation:

Metadata.getReplicas(String, TokenRange) currently assumes that the
TokenRange provided represents a TokenRange for a single host.  Because
of this, it finds the replicas for the last Token on that range instead
of considering the entire range.

Arbitrary TokenRanges may be provided that span multiple hosts.  In
this case, this method should include each host spanning this range
and their replicas.

Modifications:

Updated method to evaluate which Host's TokenRanges intersect with the
input TokenRange and include those range's replicas.
Created method Metadata.getReplicas(String, Token) to query for
individual tokens.

Result:

Metadata.getReplicas(String, TokenRange) now includes replicas for all
hosts covering a range, not just the replicas for the last token in the
range.


adutra:
Note to reviewers: I just added a1f972f as an exercice and also in the case where we would need a `TokenRange.contains(TokenRange)` method.
It turns out that to correctly implement that method, a change has had to be made to the private method `contains(Token, boolean)`.

driver-core/src/main/java/com/datastax/driver/core/TokenRange.java
```diff
@@ -166,6 +166,27 @@ public boolean intersects(TokenRange that) {
...
+     * @param that the other range.
+     * @return whether this range contains the given range.
+     */
+    public boolean contains(TokenRange that) {
+        // Empty ranges never contain any other range
+        if (this.isEmpty() || that.isEmpty())
+            return false;
+
+        return this.contains(that.start, true)
+                && this.contains(that.end, false);
```
olim7t: This doesn't work if the other range has both ends inside but wraps around, for example `]4,8].contains(]7,5])` should be false.
A straightforward way to implement `contains` is
```java
this.intersects(that) && 
    this.intersectWith(that).equals(that)
```
If we add it, we can inline and maybe simplify the code.

driver-core/src/main/java/com/datastax/driver/core/TokenRange.java
```diff
@@ -226,9 +247,8 @@ public boolean contains(Token token) {
     // * ]1,2] does not contain 1, but it contains the start of ]1,3]
     private boolean contains(Token token, boolean isStart) {
         boolean isAfterStart = isStart ? token.compareTo(start) >= 0 : token.compareTo(start) > 0;
-        boolean isBeforeEnd = end.equals(factory.minToken()) ||
-                (isStart ? token.compareTo(end) < 0 : token.compareTo(end) <= 0);
-        return isWrappedAround()
+        boolean isBeforeEnd = isStart ? token.compareTo(end) < 0 : token.compareTo(end) <= 0;
+        return isWrappedAround() || end.equals(factory.minToken())
```
olim7t, line 251: There is indeed a bug when `token.equals(minToken) && !isStart && this.end.equals(minToken)`. I'm opening a separate ticket to keep track of it.

olim7t:
Closing as we've decided to keep the current behavior.

PR #810, opened 2017-02-17
JAVA-1247: Disable idempotence warnings

adutra:
In 3.1.0, we added a warning when the first non-idempotent statement
is not retried (see JAVA-1225), as a way to ease the transition to
the new behavior.

By 3.2.0 users should be more familiar with this, so this commit
removes the warning.


tolbertam (approved):
👍

olim7t (approved):

PR #812, opened 2017-02-17
Update ProtocolVersionRenegotiationTest to use global config

tolbertam:
Since we run a test matrix in CI we can get full coverage of how the
driver behaves when connecting to different C* versions collectively.
This is quicker / more reliable than creating a different version
cluster each time.

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/ProtocolVersionRenegotiationTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/ProtocolVersionRenegotiationTest.java
@@ -16,42 +16,53 @@
 package com.datastax.driver.core;
 
 import com.datastax.driver.core.exceptions.UnsupportedProtocolVersionException;
+import com.datastax.driver.core.utils.CassandraVersion;
+import org.testng.SkipException;
+import org.testng.annotations.BeforeMethod;
 import org.testng.annotations.Test;
 
 import static com.datastax.driver.core.ProtocolVersion.*;
 import static org.assertj.core.api.Assertions.assertThat;
 
-@CreateCCM(CreateCCM.TestMode.PER_METHOD)
 public class ProtocolVersionRenegotiationTest extends CCMTestsSupport {
 
-    /**
-     * @jira_ticket JAVA-1367
-     */
-    @Test(groups = "short")
-    @CCMConfig(version = "2.1.16", createCluster = false)
-    public void should_succeed_when_version_provided_and_matches() throws Exception {
-        Cluster cluster = connectWithVersion(V3);
-        assertThat(actualProtocolVersion(cluster)).isEqualTo(V3);
+    private ProtocolVersion protocolVersion;
+
+    @BeforeMethod(groups = "short")
+    public void setUp() {
+        protocolVersion = ccm().getProtocolVersion();
     }
 
     /**
      * @jira_ticket JAVA-1367
      */
     @Test(groups = "short")
-    @CCMConfig(version = "3.6", createCluster = false)
-    public void should_fail_when_version_provided_and_too_low_3_6() throws Exception {
+    public void should_succeed_when_version_provided_and_matches() throws Exception {
+        Cluster cluster = connectWithVersion(protocolVersion);
+        assertThat(actualProtocolVersion(cluster)).isEqualTo(protocolVersion);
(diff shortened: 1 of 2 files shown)


tolbertam:
Would like to hold off on merge until know for sure that this fixes appveyor.

tolbertam:
Looks like this works now

adutra:
Sorry to re-open an old philosophical debate, but I couldn't resist :)

We are again hesitating between 2 different ways to write our tests:

**Option 1**
```java
@RequireServer("ANY")
public void test_something() {
    if (thisServer)
        testThis();
    else if (thatServer)
        testThat();
}
```
**Option 2**
```java
@RequireServer("this")
public void test_something_with_this_server() {
        testThis();
}
@RequireServer("that")
public void test_something_with_that_server() {
        testThat();
}
```

I know the first approach is intended to facilitate our CI builds, and I'm not questioning that. What annoys me with the first approach is that the test itself is not deterministic, it is environment-dependent. And even worse, it could actually never run or never test anything, if its skip logic does not match any of the environments defined in our CI.

Again, I personally would be more in favor for running all tests in a single job; a test can either require some specific versions (or ranges thereof), or be valid for all C* versions. In any case, it should never be skipped – skipped tests always made me a bit uncomfortable.

This could be the strategy used for commit builds (i.e. on each commit). Under this perspective, commit builds could be quicker than they actually are, because they only require one Jenkins slave. It would also benefit to developers, because one single `mvn test` would be enough to be sure that they did not break anything before committing.

Nightly builds, on the contrary, could use the current matrix job to add an extra validation: ensure that tests designed to work for any version of C* indeed do not break when played against this or that specific version.

Anyways I'm 👍 on this PR but I think we should definitely revisit our test framework for the next major release.

adutra (approved):
Approving because we need to move forward, but I would like us to reserve some time in the future to revisit our test framework in depth. 👍

tolbertam:
Agreed, I think we need to spend time revisiting the test framework.  I agree that we should be able to run in such a way that we get full test coverage.   On the other hand, we also need to be mindful of test run time and reliability, and as our test suite grows our turn around time grows (there's also the fact that every release cassandra takes a little bit longer to start up than the last, so creating new CCM clusters adds up).  There's a lot of wins we can gain in terms of execution time.

olim7t:
> Again, I personally would be more in favor for running all tests in a single job; a test can either require some specific versions (or ranges thereof), or be valid for all C* versions. In any case, it should never be skipped – skipped tests always made me a bit uncomfortable.

The problem with specific versions is that it's harder to maintain in the long run. In theory we should curate all `@CCMConfig(version = "x")` all the time, make sure that they don't force old versions that we're not interested in anymore, and check if there's anything we can regroup to avoid having too many different versions. It also leads to duplicate runs in the matrix builds (unless I've missed something in the test framework).

With the skipping strategy, we can pick 2/3 branches, and run the test suite with the latest version of each branch.

driver-core/src/test/java/com/datastax/driver/core/ProtocolVersionRenegotiationTest.java
```diff
@@ -16,42 +16,53 @@
...
      * @jira_ticket JAVA-1367
      */
     @Test(groups = "short")
-    @CCMConfig(version = "3.6", createCluster = false)
-    public void should_fail_when_version_provided_and_too_low_3_6() throws Exception {
+    @CassandraVersion("3.0")
+    public void should_fail_when_version_provided_and_too_low_less_than_3_8() throws Exception {
+        if (ccm().getCassandraVersion().compareTo(VersionNumber.parse("3.8")) >= 0) {
+            throw new SkipException("CASSANDRA-11464 (3.8+) uses the attempted protocol version");
+        }
```
olim7t, line 53: `CassandraVersion` could have a min and a max to handle that kind of situation.

olim7t (approved):

driver-core/src/test/java/com/datastax/driver/core/ProtocolVersionRenegotiationTest.java
tolbertam: Agreed that might be nice to have eventually as there will inevitably be some other places we'll need to do this.

PR #813, opened 2017-02-21
JAVA-1404: Fix min token handling in TokenRange.contains

olim7t:
Quick fix for a bug discovered by @adutra in #809.
Targeting to 3.0.x, we'll need to merge downstream if we include it in 3.2.0.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/TokenRange.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/TokenRange.java
@@ -215,7 +215,7 @@ public final class TokenRange implements Comparable<TokenRange> {
      * Checks whether this range contains a given token.
      *
      * @param token the token to check for.
-     * @return whether this range contains the token, i.e. {@code range.start &lt; token &lt;= range.end}.
+     * @return whether this range contains the token, i.e. {@code range.start < token <= range.end}.
      */
     public boolean contains(Token token) {
         return contains(token, false);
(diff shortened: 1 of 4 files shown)


driver-core/src/main/java/com/datastax/driver/core/TokenRange.java
```diff
@@ -224,13 +224,27 @@ public boolean contains(Token token) {
...
+            } else {
+                return isStart ? token.compareTo(start) >= 0 : token.compareTo(start) > 0;
+            }
+        } else {
+            boolean isAfterStart = isStart ? token.compareTo(start) >= 0 : token.compareTo(start) > 0;
+            boolean isBeforeEnd = isStart ? token.compareTo(end) < 0 : token.compareTo(end) <= 0;
+            return isWrappedAround()
+                    ? isAfterStart || isBeforeEnd
+                    : isAfterStart && isBeforeEnd;
+        }
```
olim7t: Not the most concise but I think clarity is more important.

adutra: May I propose another way to do it:

```java
@VisibleForTesting
boolean contains(Token token, boolean isStart) {
    if (isEmpty()) {
        return false;
    }
    if (isFull()) {
        return true;
    }
    Token minToken = factory.minToken();
    if (token.equals(minToken)) {
        return isStart
                ? start.equals(minToken)
                : end.equals(minToken);
    }
    return isWrappedAround()
            ? isAfterStart(token, isStart) || isBeforeEnd(token, isStart)
            : isAfterStart(token, isStart) && isBeforeEnd(token, isStart);
}

private boolean isFull() {
    Token minToken = factory.minToken();
    return start.equals(minToken) && end.equals(minToken);
}

private boolean isAfterStart(Token token, boolean isStart) {
    Token minToken = factory.minToken();
    int c = start.equals(minToken) ? Integer.MIN_VALUE : start.compareTo(token);
    return isStart ? c <= 0 : c < 0;
}

private boolean isBeforeEnd(Token token, boolean isStart) {
    Token minToken = factory.minToken();
    int c = end.equals(minToken) ? Integer.MAX_VALUE : end.compareTo(token);
    return isStart ? c > 0 : c >= 0;
}
```

I find it slightly easier to follow.

adutra (approved):

tolbertam (approved):

driver-core/src/main/java/com/datastax/driver/core/TokenRange.java
olim7t:
```java
if (token.equals(minToken)) {
        return isStart
                ? start.equals(minToken)
```
Not necessarily, for example this fails with the test "]2,1] contains the start of ]min,5]".

I need to merge this and merge to the downstream branches, so I'm going to proceed with the current version.

PR #814, opened 2017-02-23
JAVA-1310: Make mapper's ignored properties configurable

olim7t:
Reboot of @avivcarmis original pull request:
- target 3.x branch
- rebase on top of current 3.x
- get rid of my initial implementation

Reminder of the remaining tasks:

1. `PropertyScanScope` class will be changed to an enum containing `{GETTERS_AND_SETTERS, FIELDS, BOTH}`.
2. `HierarchyScanStrategy` class will be migrated to a single field of type Class<?> indicating the deepest ancestor to scan.
    => I think it would also be useful to indicate if the bound is inclusive or not. In some cases you may have a common parent and you'll want to say "up to this parent included", but in others you may have multiple independent trees, and it will be "up to `Object` excluded".
3. ~~Add [this change](https://github.com/datastax/java-driver/pull/771#discussion-diff-101653251R175).~~
4. ~~Document the fact that `MappingManager` [getters](https://github.com/datastax/java-driver/blob/fbcd7d50e57845adfaf8bca7fb938943fb86c955/driver-mapping/src/main/java/com/datastax/driver/mapping/MappingManager.java#L236) will account for configuration object only for the first time, then they'll be ignored [as discussed here](https://github.com/datastax/java-driver/pull/771#discussion-diff-101658604R312).~~
5. Make configuration immutable (maybe switch to builders) and remove duplication on mapper creation [as discussed here](https://github.com/datastax/java-driver/pull/771#discussion-diff-101656586R46).
6. [Move class docs to getter methods](https://github.com/datastax/java-driver/pull/771#discussion-diff-101656317R36).
7. [Move nested classes to top level ones](https://github.com/datastax/java-driver/pull/771#discussion-diff-101656811R76).
8. ~~Rename transient config properties and document transient hierarchy (config transient props vs. class annotation transient properties vs. transient annotation) [as discussed here](https://github.com/datastax/java-driver/pull/771#discussion-diff-101765523R101).~~ => I've removed the annotation-level config property
9. ~~Rename `PropertyMappingStrategy` values to `{OPT_IN, OPT_OUT}` [as discussed here](https://github.com/datastax/java-driver/pull/771#discussion-diff-101766284R334).~~
10. ~~Reuse `AnnotationParser.VALID_COLUMN_ANNOTATIONS` [as discussed here](https://github.com/datastax/java-driver/pull/771#discussion-diff-101769396R44).~~
11. ~~Account for java `transient` modifier [as discussed here](https://github.com/datastax/java-driver/pull/771#discussion-diff-101766808R138) (i'm actually +1 on that, fell for it myself when i started using the mapper).~~
12. ~~Rename `MapperConfiguration` to `MappingConfiguration`.~~
13. ~~Last thing left open is whether we want to remove [`PropertyScanConfiguration`](https://github.com/datastax/java-driver/blob/fbcd7d50e57845adfaf8bca7fb938943fb86c955/driver-mapping/src/main/java/com/datastax/driver/mapping/MapperConfiguration.java#L46) field and move it's props to `MappingConfiguration` top level.~~ => no, let's keep it separate

### DIFF
--- a/driver-mapping/src/main/java/com/datastax/driver/mapping/DefaultPropertyMapper.java
+++ b/driver-mapping/src/main/java/com/datastax/driver/mapping/DefaultPropertyMapper.java
@@ -1,0 +1,542 @@
+/*
+ *      Copyright (C) 2012-2015 DataStax Inc.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+package com.datastax.driver.mapping;
+
+import com.datastax.driver.core.Metadata;
+import com.datastax.driver.mapping.annotations.*;
+import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Sets;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.beans.BeanInfo;
+import java.beans.IntrospectionException;
+import java.beans.Introspector;
+import java.beans.PropertyDescriptor;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.AccessibleObject;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.*;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * The default {@link PropertyMapper} used by the mapper.
+ * <p/>
+ * This mapper can be configured to scan for fields, getters and setters, or both.
+ * The default is to scan for both.
+ * <p/>
+ * This mapper can also be configured to skip transient properties.
+ * By default, all properties will be mapped (non-transient),
+ * unless explicitly marked with {@link Transient @Transient}.
+ * <p/>
+ * This mapper recognizes standard getter and setter methods
+ * (as defined by the Java Beans specification),
+ * and also "relaxed" setter methods, i.e., setter methods
+ * whose return type are not {@code void}.
+ *
+ * @see DefaultMappedProperty
+ */
+public class DefaultPropertyMapper implements PropertyMapper {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(DefaultPropertyMapper.class);
+
+    private static final HashSet<String> DEFAULT_TRANSIENT_PROPERTY_NAMES = Sets.newHashSet(
+            "class",
+            // JAVA-1279: exclude Groovy's metaClass property
+            "metaClass"
+    );
+
+    private static final Set<Class<?>> NON_TRANSIENT_ANNOTATIONS = ImmutableSet.<Class<?>>of(
+            Column.class,
+            PartitionKey.class,
+            ClusteringColumn.class,
+            com.datastax.driver.mapping.annotations.Field.class,
+            Computed.class,
+            Frozen.class,
+            FrozenKey.class,
+            FrozenValue.class
+    );
+
+    /**
+     * Annotations allowed on a property that maps to a table column.
+     */
+    private static final Set<Class<? extends Annotation>> VALID_COLUMN_ANNOTATIONS = ImmutableSet.<Class<? extends Annotation>>builder()
+            .add(Column.class)
+            .add(Computed.class)
+            .add(ClusteringColumn.class)
+            .add(Frozen.class)
+            .add(FrozenKey.class)
+            .add(FrozenValue.class)
+            .add(PartitionKey.class)
+            .add(Transient.class)
+            .build();
+
+    /**
+     * Annotations allowed on a property that maps to a UDT field.
+     */
+    private static final Set<Class<? extends Annotation>> VALID_FIELD_ANNOTATIONS = ImmutableSet.of(
+            com.datastax.driver.mapping.annotations.Field.class,
+            Frozen.class,
+            FrozenKey.class,
+            FrozenValue.class,
+            Transient.class);
+
+    private PropertyAccessStrategy propertyAccessStrategy = PropertyAccessStrategy.BOTH;
+
+    private PropertyTransienceStrategy propertyTransienceStrategy = PropertyTransienceStrategy.OPT_OUT;
+
+    private HierarchyScanStrategy hierarchyScanStrategy = new DefaultHierarchyScanStrategy();
+
+    private NamingStrategy namingStrategy = new DefaultNamingStrategy();
+
+    private Set<String> transientPropertyNames = new HashSet<String>(DEFAULT_TRANSIENT_PROPERTY_NAMES);
+
+    /**
+     * Sets the {@link PropertyAccessStrategy property access strategy} to use.
+     * The default is {@link PropertyAccessStrategy#BOTH}.
+     *
+     * @param propertyAccessStrategy the {@link PropertyAccessStrategy property access strategy} to use; may not be {@code null}.
+     * @return this {@link DefaultPropertyMapper} instance (to allow for fluent builder pattern).
+     */
+    public DefaultPropertyMapper setPropertyAccessStrategy(PropertyAccessStrategy propertyAccessStrategy) {
+        this.propertyAccessStrategy = checkNotNull(propertyAccessStrategy);
+        return this;
+    }
+
+    /**
+     * Sets the {@link PropertyTransienceStrategy property transience strategy} to use.
+     * The default is {@link PropertyTransienceStrategy#OPT_OUT}.
+     *
+     * @param propertyTransienceStrategy the {@link PropertyTransienceStrategy property transience strategy} to use; may not be {@code null}.
+     * @return this {@link DefaultPropertyMapper} instance (to allow for fluent builder pattern).
+     */
+    public DefaultPropertyMapper setPropertyTransienceStrategy(PropertyTransienceStrategy propertyTransienceStrategy) {
+        this.propertyTransienceStrategy = checkNotNull(propertyTransienceStrategy);
+        return this;
+    }
+
+    /**
+     * Sets the {@link HierarchyScanStrategy hierarchy scan strategy} to use.
+     * The default is {@link DefaultHierarchyScanStrategy}.
+     *
+     * @param hierarchyScanStrategy the {@link HierarchyScanStrategy hierarchy scan strategy} to use; may not be {@code null}.
+     * @return this {@link DefaultPropertyMapper} instance (to allow for fluent builder pattern).
+     */
+    public DefaultPropertyMapper setHierarchyScanStrategy(HierarchyScanStrategy hierarchyScanStrategy) {
+        this.hierarchyScanStrategy = checkNotNull(hierarchyScanStrategy);
+        return this;
+    }
+
+    /**
+     * Sets the {@link NamingStrategy naming strategy} to use.
+     * The default is {@link DefaultNamingStrategy}.
+     *
+     * @param namingStrategy the {@link NamingStrategy naming strategy} to use; may not be {@code null}.
+     * @return this {@link DefaultPropertyMapper} instance (to allow for fluent builder pattern).
+     */
+    public DefaultPropertyMapper setNamingStrategy(NamingStrategy namingStrategy) {
+        this.namingStrategy = checkNotNull(namingStrategy);
+        return this;
+    }
+
+    /**
+     * Sets transient property names. This
+     * will completely replace any names already configured for this object.
+     * <p/>
+     * The default set comprises the following property names:
+     * {@code class} and {@code metaClass}.
+     * These properties pertain to the {@link Object} class –
+     * {@code metaClass} being specific to the Groovy language.
+     * <p/>
+     * Property names provided here will always be considered transient;
+     * if a more fine-grained tuning is required, it is also possible
+     * to use the {@link Transient @Transient} annotation
+     * on a specific property.
+     * <p/>
+     * Subclasses can also override {@link #isTransient(String, Field, Method, Method, Map)} to gain
+     * complete control over which properties should be considered transient.
+     *
+     * @param transientPropertyNames a set of property names to exclude from mapping; may not be {@code null}. This
+     *                               will completely replace any names already configured for this object.
+     */
+    public DefaultPropertyMapper setTransientPropertyNames(Set<String> transientPropertyNames) {
+        this.transientPropertyNames = checkNotNull(transientPropertyNames);
+        return this;
+    }
+
+    /**
+     * Adds new values to the existing set of transient property names.
+     * <p/>
+     * The default set comprises the following property names:
+     * {@code class} and {@code metaClass}.
+     * These properties pertain to the {@link Object} class –
+     * {@code metaClass} being specific to the Groovy language.
+     * <p/>
+     * Property names provided here will always be considered transient;
+     * if a more fine-grained tuning is required, it is also possible
+     * to use the {@link Transient @Transient} annotation
+     * on a specific property.
+     * <p/>
+     * Subclasses can also override {@link #isTransient(String, Field, Method, Method, Map)} to gain
+     * complete control over which properties should be considered transient.
+     *
+     * @param transientPropertyNames the values to add; may not be {@code null}.
+     */
+    public DefaultPropertyMapper addTransientPropertyNames(String... transientPropertyNames) {
+        return addTransientPropertyNames(Arrays.asList(checkNotNull(transientPropertyNames)));
+    }
+
+    /**
+     * Adds new values to the existing set of transient property names.
+     * <p/>
+     * The default set comprises the following property names:
+     * {@code class} and {@code metaClass}.
+     * These properties pertain to the {@link Object} class –
+     * {@code metaClass} being specific to the Groovy language.
+     * <p/>
+     * Property names provided here will always be considered transient;
+     * if a more fine-grained tuning is required, it is also possible
+     * to use the {@link Transient @Transient} annotation
+     * on a specific property.
+     * <p/>
+     * Subclasses can also override {@link #isTransient(String, Field, Method, Method, Map)} to gain
+     * complete control over which properties should be considered transient.
+     *
+     * @param transientPropertyNames the values to add; may not be {@code null}.
+     */
+    public DefaultPropertyMapper addTransientPropertyNames(Collection<String> transientPropertyNames) {
+        this.transientPropertyNames.addAll(checkNotNull(transientPropertyNames));
+        return this;
+    }
+
+    @Override
+    public Set<? extends MappedProperty<?>> mapTable(Class<?> tableClass) {
+        return mapTableOrUdt(tableClass, VALID_COLUMN_ANNOTATIONS);
+    }
+
+    @Override
+    public Set<? extends MappedProperty<?>> mapUdt(Class<?> udtClass) {
+        return mapTableOrUdt(udtClass, VALID_FIELD_ANNOTATIONS);
+    }
+
+    private Set<? extends MappedProperty<?>> mapTableOrUdt(Class<?> entityClass, Collection<? extends Class<? extends Annotation>> allowed) {
+        Map<String, Object[]> fieldsGettersAndSetters = new HashMap<String, Object[]>();
+        List<Class<?>> classHierarchy = hierarchyScanStrategy.filterClassHierarchy(entityClass);
+        if (propertyAccessStrategy.isFieldScanAllowed()) {
+            Map<String, Field> fields = scanFields(classHierarchy);
+            for (Map.Entry<String, Field> entry : fields.entrySet()) {
+                String propertyName = entry.getKey();
+                Field field = tryMakeAccessible(entry.getValue());
+                fieldsGettersAndSetters.put(propertyName, new Object[]{field, null, null});
+            }
+        }
+        if (propertyAccessStrategy.isGetterSetterScanAllowed()) {
+            Map<String, PropertyDescriptor> properties = scanProperties(classHierarchy);
+            for (Map.Entry<String, PropertyDescriptor> entry : properties.entrySet()) {
+                PropertyDescriptor property = entry.getValue();
+                Method getter = tryMakeAccessible(locateGetter(entityClass, property));
+                Method setter = tryMakeAccessible(locateSetter(entityClass, property));
+                Object[] value = fieldsGettersAndSetters.get(entry.getKey());
+                if (value != null) {
+                    value[1] = getter;
+                    value[2] = setter;
+                } else if (getter != null || setter != null) {
+                    fieldsGettersAndSetters.put(entry.getKey(), new Object[]{null, getter, setter});
+                }
+            }
+        }
+        Set<MappedProperty<?>> mappedProperties = new HashSet<MappedProperty<?>>(fieldsGettersAndSetters.size());
+        for (Map.Entry<String, Object[]> entry : fieldsGettersAndSetters.entrySet()) {
+            String propertyName = entry.getKey();
+            Field field = (Field) entry.getValue()[0];
+            Method getter = (Method) entry.getValue()[1];
+            Method setter = (Method) entry.getValue()[2];
+            Map<Class<? extends Annotation>, Annotation> annotations = scanPropertyAnnotations(field, getter);
+            AnnotationChecks.validateAnnotations(propertyName, annotations, allowed);
+            if (isTransient(propertyName, field, getter, setter, annotations)) {
+                LOGGER.debug(String.format("Property '%s' is transient and will not be mapped", propertyName));
+                continue;
+            }
+            if (!annotations.containsKey(Computed.class) && field == null && getter == null) {
+                throw new IllegalArgumentException(String.format("Property '%s' is not readable", propertyName));
+            }
+            if (field == null && setter == null) {
+                throw new IllegalArgumentException(String.format("Property '%s' is not writable", propertyName));
+            }
+            String mappedName = inferMappedName(entityClass, propertyName, annotations);
+            MappedProperty<?> property = createMappedProperty(entityClass, propertyName, mappedName, field, getter, setter, annotations);
+            mappedProperties.add(property);
+        }
+        return mappedProperties;
+    }
+
+    /**
+     * Returns {@code true} if the given property is transient,
+     * {@code false} otherwise.
+     * <p/>
+     * If this method returns {@code true} the given property will not be mapped.
+     * The implementation provided here relies on the
+     * {@link #setPropertyTransienceStrategy(PropertyTransienceStrategy) transience strategy}
+     * and the {@link #setTransientPropertyNames(Set) transient property names}
+     * configured on this mapper.
+     * <p/>
+     * Subclasses may override this method to take full control of which properties
+     * should be mapped and which should be considered transient.
+     *
+     * @param propertyName the property name; may not be {@code null}.
+     * @param field        the property field; may be {@code null}.
+     * @param getter       the getter method for this property; may be {@code null}.
+     * @param setter       the setter method for this property; may be {@code null}.
+     * @param annotations  the annotations found on this property; may be empty but never {@code null}.
+     * @return {@code true} if the given property is transient (i.e., non-mapped), {@code false} otherwise.
+     */
+    @SuppressWarnings("unused")
+    protected boolean isTransient(String propertyName, Field field, Method getter, Method setter, Map<Class<? extends Annotation>, Annotation> annotations) {
+        if (propertyTransienceStrategy == PropertyTransienceStrategy.OPT_OUT)
+            return annotations.containsKey(Transient.class)
+                    || (transientPropertyNames.contains(propertyName)
+                    && Collections.disjoint(annotations.keySet(), NON_TRANSIENT_ANNOTATIONS));
+        else
+            return Collections.disjoint(annotations.keySet(), NON_TRANSIENT_ANNOTATIONS);
+    }
+
+    /**
+     * Locates a getter method for the given mapped class and given property.
+     * <p/>
+     * Most users should rely on the implementation provided here.
+     * It is however possible to return any non-standard method, as long as it does
+     * not take parameters, and its return type is assignable to (and covariant with) the property's type.
+     * This might be particularly useful for boolean properties whose names are verbs, e.g. "{@code hasAccount}":
+     * one could then return the non-standard method {@code boolean hasAccount()} as its getter.
+     * <p/>
+     * This method is never called if {@link PropertyAccessStrategy#isGetterSetterScanAllowed()} returns {@code false}.
+     * Besides, implementors are free to return {@code null} if access to the property through reflection is not required
+     * (in which case, they will likely have to provide a custom implementation of {@link MappedProperty}).
+     *
+     * @param mappedClass The mapped class; this is necessarily a class annotated with
+     *                    either {@link Table @Table} or
+     *                    {@link UDT @UDT}.
+     * @param property    The property to locate a getter for; never {@code null}.
+     * @return The getter method for the given base class and given property, or {@code null} if no getter was found, or reflection is not required.
+     */
+    protected Method locateGetter(@SuppressWarnings("unused") Class<?> mappedClass, PropertyDescriptor property) {
+        return property.getReadMethod();
+    }
+
+    /**
+     * Locates a setter method for the given mapped class and given property.
+     * <p/>
+     * Most users should rely on the implementation provided here.
+     * It is however possible to return any non-standard method, as long as it accepts one single parameter type
+     * that is contravariant with the property's type.
+     * <p/>
+     * This method is never called if {@link PropertyAccessStrategy#isGetterSetterScanAllowed()} returns {@code false}.
+     * Besides, implementors are free to return {@code null} if access to the property through reflection is not required
+     * (in which case, they will likely have to provide a custom implementation of {@link MappedProperty}).
+     *
+     * @param mappedClass The mapped class; this is necessarily a class annotated with
+     *                    either {@link Table @Table} or
+     *                    {@link UDT @UDT}.
+     * @param property    The property to locate a setter for; never {@code null}.
+     * @return The setter method for the given base class and given property, or {@code null} if no setter was found, or reflection is not required.
+     */
+    protected Method locateSetter(Class<?> mappedClass, PropertyDescriptor property) {
+        Method setter = property.getWriteMethod();
+        if (setter == null) {
+            // JAVA-984: look for a "relaxed" setter, ie. a setter whose return type may be anything
+            String propertyName = property.getName();
+            String setterName = "set" + propertyName.substring(0, 1).toUpperCase() + propertyName.substring(1);
+            try {
+                Method m = mappedClass.getMethod(setterName, property.getPropertyType());
+                if (!Modifier.isStatic(m.getModifiers()))
+                    setter = m;
+            } catch (NoSuchMethodException ignored) {
+            }
+        }
+        return setter;
+    }
+
+    /**
+     * Infers the Cassandra object name corresponding to given the property name.
+     * <p/>
+     * Most users should rely on the implementation provided here.
+     * It relies on annotation values and ultimately resorts to the
+     * {@link NamingStrategy} configured on this mapper.
+     * <p/>
+     * Subclasses may override this method if they need full control
+     * over generating Cassandra object names.
+     *
+     * @param mappedClass The mapped class; this is necessarily a class annotated with
+     *                    either {@link Table @Table} or
+     *                    {@link UDT @UDT}.
+     * @param propertyName The property name; may not be {@code null} nor empty.
+     * @param annotations  The property annotations (found on its field and getter method); may not be {@code null} but can be empty.
+     * @return The inferred Cassandra object name.
+     */
+    protected String inferMappedName(@SuppressWarnings("unused") Class<?> mappedClass, String propertyName, Map<Class<? extends Annotation>, Annotation> annotations) {
+        if (annotations.containsKey(Computed.class)) {
+            String expression = ((Computed) annotations.get(Computed.class)).value();
+            if (expression.isEmpty())
+                throw new IllegalArgumentException(String.format("Property '%s': attribute 'value' of annotation @Computed is mandatory for computed properties", propertyName));
+            return expression;
+        }
+
+        // If a name is explicitly provided with @Column or @Field, use it
+        boolean caseSensitive = false;
+        String mappedName = null;
+        if (annotations.containsKey(Column.class)) {
+            Column column = (Column) annotations.get(Column.class);
+            caseSensitive = column.caseSensitive();
+            if (!column.name().isEmpty())
+                mappedName = column.name();
+        } else if (annotations.containsKey(com.datastax.driver.mapping.annotations.Field.class)) {
+            com.datastax.driver.mapping.annotations.Field udtMappedField =
+                    (com.datastax.driver.mapping.annotations.Field) annotations.get(com.datastax.driver.mapping.annotations.Field.class);
+            caseSensitive = udtMappedField.caseSensitive();
+            if (!udtMappedField.name().isEmpty())
+                mappedName = udtMappedField.name();
+        }
+        if (mappedName != null) {
+            return caseSensitive ? Metadata.quote(mappedName) : mappedName.toLowerCase();
+        }
+
+        // Otherwise delegate to the naming strategy
+        mappedName = namingStrategy.toCassandraName(propertyName);
+        if (mappedName == null || mappedName.isEmpty())
+            throw new IllegalArgumentException(String.format("Property '%s': could not infer mapped name", propertyName));
+        return Metadata.quoteIfNecessary(mappedName);
+    }
+
+    /**
+     * Creates a {@link MappedProperty} instance.
+     * <p>
+     * Instances returned by the implementation below will use the Java reflection API to read and write values.
+     * Subclasses may override this method if they are capable of accessing
+     * properties without incurring the cost of reflection.
+     *
+     * @param mappedClass The mapped class; this is necessarily a class annotated with
+     *                    either {@link Table @Table} or
+     *                    {@link UDT @UDT}.
+     * @param propertyName The property name; may not be {@code null} nor empty.
+     * @param mappedName   The mapped name; may not be {@code null} nor empty.
+     * @param field        The property field; may be {@code null}.
+     * @param getter       The property getter method; may be {@code null}.
+     * @param setter       The property setter method; may  be {@code null}.
+     * @param annotations  The property annotations (found on its field and getter method); may not be {@code null} but can be empty.
+     * @return a newly-allocated {@link MappedProperty} instance.
+     */
+    protected MappedProperty<?> createMappedProperty(Class<?> mappedClass, String propertyName, String mappedName, Field field, Method getter, Method setter, Map<Class<? extends Annotation>, Annotation> annotations) {
+        return DefaultMappedProperty.create(mappedClass, propertyName, mappedName, field, getter, setter, annotations);
+    }
+
+    private static Map<String, Field> scanFields(List<Class<?>> classHierarchy) {
+        HashMap<String, Field> fields = new HashMap<String, Field>();
+        for (Class<?> clazz : classHierarchy) {
+            for (Field field : clazz.getDeclaredFields()) {
+                if (field.isSynthetic() || Modifier.isStatic(field.getModifiers()) || Modifier.isTransient(field.getModifiers()))
+                    continue;
+                // never override a more specific field masking another one declared in a superclass
+                if (!fields.containsKey(field.getName()))
+                    fields.put(field.getName(), field);
+            }
+        }
+        return fields;
+    }
+
+    private static Map<String, PropertyDescriptor> scanProperties(List<Class<?>> classHierarchy) {
+        Map<String, PropertyDescriptor> properties = new HashMap<String, PropertyDescriptor>();
+        for (Class<?> clazz : classHierarchy) {
+            // each time extract only current class properties
+            BeanInfo beanInfo;
+            try {
+                beanInfo = Introspector.getBeanInfo(clazz, clazz.getSuperclass());
+            } catch (IntrospectionException e) {
+                throw Throwables.propagate(e);
+            }
+            for (PropertyDescriptor property : beanInfo.getPropertyDescriptors()) {
+                if (!properties.containsKey(property.getName())) {
+                    properties.put(property.getName(), property);
+                }
+            }
+        }
+        return properties;
+    }
+
+    private static Map<Class<? extends Annotation>, Annotation> scanPropertyAnnotations(Field field, Method getter) {
+        Map<Class<? extends Annotation>, Annotation> annotations = new HashMap<Class<? extends Annotation>, Annotation>();
+        // annotations on getters should have precedence over annotations on fields
+        if (field != null)
+            scanFieldAnnotations(field, annotations);
+        if (getter != null)
+            scanMethodAnnotations(getter, annotations);
+        return annotations;
+    }
+
+    private static Map<Class<? extends Annotation>, Annotation> scanFieldAnnotations(Field field, Map<Class<? extends Annotation>, Annotation> annotations) {
+        for (Annotation annotation : field.getAnnotations()) {
+            annotations.put(annotation.annotationType(), annotation);
+        }
+        return annotations;
+    }
+
+    private static Map<Class<? extends Annotation>, Annotation> scanMethodAnnotations(Method method, Map<Class<? extends Annotation>, Annotation> annotations) {
+        // 1. direct method annotations
+        for (Annotation annotation : method.getAnnotations()) {
+            annotations.put(annotation.annotationType(), annotation);
+        }
+        // 2. Class hierarchy: check for annotations in overridden methods in superclasses
+        Class<?> getterClass = method.getDeclaringClass();
+        for (Class<?> clazz = getterClass.getSuperclass(); clazz != null && !clazz.equals(Object.class); clazz = clazz.getSuperclass()) {
+            maybeAddOverriddenMethodAnnotations(annotations, method, clazz);
+        }
+        // 3. Interfaces: check for annotations in implemented interfaces
+        for (Class<?> clazz = getterClass; !clazz.equals(Object.class); clazz = clazz.getSuperclass()) {
+            for (Class<?> itf : clazz.getInterfaces()) {
+                maybeAddOverriddenMethodAnnotations(annotations, method, itf);
+            }
+        }
+        return annotations;
+    }
+
+    private static void maybeAddOverriddenMethodAnnotations(Map<Class<? extends Annotation>, Annotation> annotations, Method getter, Class<?> clazz) {
+        try {
+            Method overriddenGetter = clazz.getDeclaredMethod(getter.getName(), (Class<?>[]) getter.getParameterTypes());
+            for (Annotation annotation : overriddenGetter.getAnnotations()) {
+                // do not override a more specific version of the annotation type being scanned
+                if (!annotations.containsKey(annotation.annotationType()))
+                    annotations.put(annotation.annotationType(), annotation);
+            }
+        } catch (NoSuchMethodException e) {
+            //ok
+        }
+    }
+
+    private static <T extends AccessibleObject> T tryMakeAccessible(T object) {
+        if (object != null && !object.isAccessible()) {
+            try {
+                object.setAccessible(true);
+            } catch (SecurityException e) {
+                // ok
+            }
+        }
+        return object;
+    }
+
+}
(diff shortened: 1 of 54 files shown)


olim7t:
Let's finish this first and we'll focus on 1316 next.

olim7t:
The code is currently broken, removing the annotation-level exclusions creates an issue with `Object.getClass()`: even if it's excluded in the configuration we try to create a `PropertyMapper` for it, which fails because there is no corresponding setter.
We should not try to create property mappers for ignored fields, I'm trying to move the checks earlier.

adutra:
Ok I have completed the remaining tasks and fixed the tests.

A couple of remarks:

1. All classes moved to the child `config` package;
2. All classes are now immutable and use the builder pattern;
3. Hierarchy scan: we have previously agreed to remove/reduce `HierarchyScanStrategy`, but indeed we might have overlooked the inclusion/exclusion problem. I propose the following:
    1. Keep `HierarchyScanStrategy`, replace `deepestAllowedAncestor` with `nearestExcludedAncestor` (i.e. make this the actual stop class = inclusive);
    2. Remove `scanOnlyAnnotatedClasses`: I think I missed this parameter and it seems too much for me, and besides, we do not encourage users to inherit from annotated classes. If we want to re-introduce this parameter, then we will have to introduce a separate annotation (e.g. `@MappedSuperclass`).

adutra:
Also added one commit to disable mapper-level configuration in favor of manager-level configuration only. I honestly do not see a valid use case for mapping the same class in two different ways under the same application, but I'm open to discussion.

olim7t:
> Also added one commit to disable mapper-level configuration in favor of manager-level configuration only. I honestly do not see a valid use case for mapping the same class in two different ways under the same application, but I'm open to discussion.

What about mapping two different classes with a different set of property exclusions?

avivcarmis:
@olim7t cheers! sorry for not being available for a couple of days. everything looks great. let me know if there's anything else i can do regarding 1310.
At the meantime i'll try to sum everything up with 1316. Since it's based on my 1310 branch it might be a bit of a struggle to rebase it. I think before we start coding we might want to finish here, rebase and then go ahead.

adutra:
> What about mapping two different classes with a different set of property exclusions?

Hmm sorry but is this realistic? Even if it were, I don't think we have users willing to have this feature as of now, so why not add it later if need be? (it's not breaking)

olim7t:
I added a commit for the inclusive/exclusive hierarchy bound.

Two things bother me about transient properties:
1. you can't remove the defaults (this could be easily addressed with an additional builder method that replaces the set instead of adding to it)
2. the exclusions are shared across all mapped classes and UDTs.
    Per-class configuration is not a good answer to this, because UDT codecs share the configuration of the first table that required it, so there would be no deterministic way to know which one is used. And you couldn't have different exclusions for a table and its fields' UDTs either.
    Another approach would be a single configuration, but with dotted notation for excluded properties. This gets very complicated (`*` and `**` matching comes to mind), and there is currently no trivial way to pass the current prefix to a UDT codec (again because UDT codecs are shared across all mappers that need them).

I can't think of a good solution for 2. Maybe I'm overthinking this; if we're confident that shared exclusions are good enough, then we can keep the current implementation and I'm fine with a single, manager-wide configuration object.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyScanConfiguration.java
```diff
@@ -119,14 +54,14 @@ public Builder withPropertyAccessStrategy(PropertyAccessStrategy propertyAccessS
...
+         * Sets the {@link PropertyTransienceStrategy property mapping strategy} to use.
+         * The default is {@link OptOutPropertyTransienceStrategy}.
          *
-         * @param propertyMappingStrategy the {@link PropertyMappingStrategy property mapping strategy} to use.
+         * @param propertyTransienceStrategy the {@link PropertyTransienceStrategy property mapping strategy} to use.
          * @return this {@link Builder} instance (to allow for fluent builder pattern).
          */
-        public Builder withPropertyMappingStrategy(PropertyMappingStrategy propertyMappingStrategy) {
-            this.propertyMappingStrategy = propertyMappingStrategy;
+        public Builder withPropertyMappingStrategy(PropertyTransienceStrategy propertyTransienceStrategy) {
```
newkek: I think `withPropertyMappingStrategy` needs to be renamed

adutra: Good catch thanks!

adutra:
I am starting to have second thoughts about the usefulness of `PropertyScanConfiguration`. After my latest commits it's been reduced to nothing more than a wrapper for some strategies, so I would now be in favor of removing it altogether. Wdyt?
EDIT: besides, `HierarchyScanStrategy` doesn't really belong in there imo.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/OptInPropertyTransienceStrategy.java
```diff
@@ -0,0 +1,47 @@
...
+ * <li>{@link com.datastax.driver.mapping.annotations.FrozenValue FrozenValue}</li>
+ * <li>{@link com.datastax.driver.mapping.annotations.PartitionKey PartitionKey}</li>
+ * <li>{@link com.datastax.driver.mapping.annotations.Field Field}</li>
+ * </ol>
+ */
+public class OptInPropertyTransienceStrategy implements PropertyTransienceStrategy {
+
+    @Override
+    public boolean isTransient(String propertyName, Field field, Method getter, Method setter, Map<Class<? extends Annotation>, Annotation> annotations) {
+        return annotations.isEmpty() || annotations.containsKey(Transient.class);
```
olim7t: From what I can tell, `annotations` is not filtered to the driver's types, so that would include a field that has _any_ annotation except `Transient`.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/DefaultPropertyTransienceStrategy.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+
+    /**
+     * Returns a new {@link DefaultPropertyTransienceStrategy.Builder} instance.
+     *
+     * @return a new {@link DefaultPropertyTransienceStrategy.Builder} instance.
+     */
+    public static DefaultPropertyTransienceStrategy.Builder builder() {
+        return new DefaultPropertyTransienceStrategy.Builder();
```
olim7t: How about getting rid of the builder pattern for this class? It adds a lot of boilerplate, feels like if different behavior is needed it might take the form of alternate implementations.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyAccessStrategy.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+public interface PropertyAccessStrategy {
+
+    /**
+     * Returns whether or not the given {@link Field} should
+     * be used when accessing its underlying property.
+     *
+     * @return {@code true} if the given field should be considered, {@code false} otherwise.
+     */
+    boolean isFieldAccessAllowed();
```
olim7t: This (and `isGetterSetterAccessAllowed`) should be used in `ReflectionUtils.scanFieldsAndProperties` to completely skip scanning these members if they're not going to be used.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyAccessStrategy.java
```diff
@@ -0,0 +1,90 @@
...
+import java.lang.reflect.Method;
+
+/**
+ * A strategy to determine how mapped properties are read and written.
+ */
+public interface PropertyAccessStrategy {
+
+    /**
+     * Returns whether or not the given {@link Field} should
+     * be used when accessing its underlying property.
```
olim7t: This threw me off at first read, generally when we say "given" it refers to a parameter. But this method has no parameters (maybe it used to?).

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyAccessStrategy.java
```diff
@@ -0,0 +1,94 @@
...
+ *   limitations under the License.
+ */
+package com.datastax.driver.mapping.config;
+
+import java.beans.PropertyDescriptor;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+
+/**
+ * A strategy to determine how mapped properties are read and written.
```
olim7t: I think we should have a basic description of the scanning/mapping process detailing where each method is invoked, because it's hard to grasp from an outsider's perspective.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyAccessStrategy.java
```diff
@@ -0,0 +1,94 @@
...
+     * Locates a getter method for the given base class and given property.
+     * If this method returns {@code null} then no getter will be used to access the given property.
+     *
+     * @param mappedClass The mapped class; this is necessarily a class annotated with
+     *                  either {@link com.datastax.driver.mapping.annotations.Table @Table} or
+     *                  {@link com.datastax.driver.mapping.annotations.UDT @UDT}.
+     * @param property  The property to locate a getter for; never {@code null}.
+     * @return The getter method for the given base class and given property, or {@code null} if no getter was found.
+     */
+    Method locateGetter(Class<?> mappedClass, PropertyDescriptor property);
```
olim7t: I've tried to come up with use cases for this, the only one I can think of is boolean properties with verbal names, for example `hasAccount()` instead of `getHasAccount()` or worse `isHasAccount()`.
That is legitimate so I agree to have this method, but we should explain that in the javadoc to make it clear where it comes from.
Same thing from `locateSetter` and the fluent pattern.

Note that if you have both a non-standard getter and setter, then there's no way to use them. That's far-fetched so I don't think we need to support it.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyAccessStrategy.java
```diff
@@ -0,0 +1,94 @@
...
+     * @param entity       The instance to read the property from; never {@code null}.
+     * @param propertyName The property name, as inferred by the mapper; never {@code null}.
+     * @param field        The property field; may be {@code null}.
+     *                     May be {@code null}, but {@code field} and {@code getter} cannot be both {@code null}.
+     * @param getter       The property's getter method, as inferred by {@link #locateGetter(Class, PropertyDescriptor) this strategy}.
+     *                     May be {@code null}, but {@code field} and {@code getter} cannot be both {@code null}.
+     * @return The property value.
+     * @throws IllegalArgumentException if the property cannot be read.
+     */
+    Object getValue(Object entity, String propertyName, Field field, Method getter);
```
olim7t: Contrived example: you could have `isGetterSetterAccessAllowed` to true but then ignore the getter in this method. This leads me to believe that this might be too low-level.
Is there any other use case than controlling the order between getter and direct access? We could expose something simpler, like `boolean tryGetterFirst();`.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyAccessStrategy.java
```diff
@@ -0,0 +1,94 @@
...
+     */
+    boolean isFieldAccessAllowed();
+
+    /**
+     * Returns whether or not the given {@link PropertyDescriptor} should
+     * be used when accessing its underlying property.
+     *
+     * @return {@code true} if the given {@link PropertyDescriptor} should be considered, {@code false} otherwise.
+     */
+    boolean isGetterSetterAccessAllowed();
```
olim7t: It should be obvious that it is absurd to have both `isFieldAccessAllowed` and `isGetterSetterAccessAllowed` to false. But I'm mildly annoyed by the fact that the API allows it. That wasn't the case with the enum.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyTransienceStrategy.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param getter       The property getter, as inferred by the {@link PropertyAccessStrategy#locateGetter(Class, PropertyDescriptor) access strategy}.
+     *                     May be {@code null}, but {@code field}, {@code getter} and {@code setter} cannot be all {@code null}.
+     * @param setter       The property setter, as inferred by the {@link PropertyAccessStrategy#locateSetter(Class, PropertyDescriptor) access strategy}.
+     *                     May be {@code null}, but {@code field}, {@code getter} and {@code setter} cannot be all {@code null}.
+     * @param annotations  The property's annotations, as inferred by the mapper; may be empty, but never {@code null}.
+     *                     The map will include all annotations found on the field, the getter and the setter methods.
+     * @return {@code true} if the given property is transient (i.e., non-mapped), {@code false} otherwise.
+     */
+    boolean isTransient(String propertyName, Field field, Method getter, Method setter, Map<Class<? extends Annotation>, Annotation> annotations);
```
olim7t: I wonder if we should also include the mapped class here. It's not used right now but could prove useful. It cannot be derived from the field and getter/setter if those were inherited.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/DefaultHierarchyScanStrategy.java
```diff
@@ -0,0 +1,103 @@
...
+     * @return a new {@link DefaultHierarchyScanStrategy.Builder} instance.
+     */
+    public static DefaultHierarchyScanStrategy.Builder builder() {
+        return new DefaultHierarchyScanStrategy.Builder();
+    }
+
+    /**
+     * Builder for {@link DefaultHierarchyScanStrategy} instances.
+     */
+    public static class Builder {
```
olim7t: When I added a builder here it was mainly for consistency with other configuration classes. Since I'm now advocating for removing the builders elsewhere, this one could go too.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/MappingConfiguration.java
```diff
@@ -0,0 +1,136 @@
...
+     * @return a new {@link Builder} instance.
+     */
+    public static MappingConfiguration.Builder builder() {
+        return new MappingConfiguration.Builder();
+    }
+
+    /**
+     * Builder for {@link MappingConfiguration} instances.
+     */
+    public static class Builder {
```
olim7t: Just for clarity, here it's justified to have a builder.

driver-mapping/src/main/java/com/datastax/driver/mapping/ReflectionUtils.java
```diff
@@ -63,13 +58,15 @@
     // for each key representing a property name,
     // value[0] contains a Field object, value[1] contains a PropertyDescriptor object;
     // they cannot be both null at the same time
-    static <T> Map<String, Object[]> scanFieldsAndProperties(Class<T> baseClass) {
+    static <T> Map<String, Object[]> scanFieldsAndProperties(Class<T> baseClass, HierarchyScanStrategy hierarchyScanStrategy) {
         Map<String, Object[]> fieldsAndProperties = new HashMap<String, Object[]>();
-        Map<String, Field> fields = scanFields(baseClass);
+        List<Class<?>> classesToScan = Lists.<Class<?>>newArrayList(baseClass);
+        classesToScan.addAll(hierarchyScanStrategy.filterClassHierarchy(baseClass));
```
olim7t: Why not require that the strategy include the base class in the list? Then we could avoid that copy.

clirr-ignores.xml
```diff
@@ -147,4 +147,18 @@
...
+        <justification>False positive, it's an annotation and the new method has a default value</justification>
+    </difference>
+
+    <difference>
+        <differenceType>7012</differenceType> <!-- method added to interface -->
+        <className>com/datastax/driver/mapping/annotations/UDT</className>
+        <method>java.lang.String[] transientProperties()</method>
+        <justification>False positive, it's an annotation and the new method has a default value</justification>
+    </difference>
+
```
olim7t: I forgot to remove this, it's not needed anymore.

olim7t (commented):
(can only comment since I opened the PR)

Looks good to me with the latest changes to the strategies. `PropertyAccessStrategy` was the hardest to follow for me, but I think with a few changes and a bit of extra documentation it should be better.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyAccessStrategy.java
adutra: Sorry the docs are stale and referring to a previous version of this class.

driver-mapping/src/main/java/com/datastax/driver/mapping/ReflectionUtils.java
adutra: Indeed, but wouldn't we have to check that the strategy did comply with this rule? It would make little sense to annotate a class with `@Table` and exclude it from mapping.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyAccessStrategy.java
adutra: > Note that if you have both a non-standard getter and setter, then there's no way to use them. That's far-fetched so I don't think we need to support it.

Why can't they be used if they are both non-standard? The minimal requirement for a getter is to have a return type  assignable to (and covariant with) the property's type, and for a setter, is to have one single parameter type that is contravariant with the property's type.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyAccessStrategy.java
adutra: Yes, let's reinstate the enum.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyAccessStrategy.java
adutra: Initially I placed these methods here to allow users to completely bypass the field/getter mechanism and grasp their property values from virtually anywhere. So the meaning of the signature was to say "here is what the mapper knows about this property; give me (or write for me) its value". 
EDIT: I think it is useful to keep `getValue` and `setValue` here. I believe the initial authors of the mapper have kept in mind that some applications might be able to avoid reflection completely, that's why we used to have `ReflectionMapper` as a concrete subclass of `ColumnMapper` implementing precisely these two methods.

adutra: And by "virtually anywhere" I was thinking of applications capable of determine getters and setters without reflection, e.g. an application might be able to determine that to read property `foo` of class `MyTable`, one should simply call `((MyTable)entity).getFoo()`.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyAccessStrategy.java
olim7t: If both are non-standard, `Introspector.getBeanInfo` won't detect them as a `PropertyDescriptor`.

driver-mapping/src/main/java/com/datastax/driver/mapping/ReflectionUtils.java
olim7t: If the contract says that it must be included, and an implementation doesn't, then it's a bug in the implementation and there's little we can do about it.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyAccessStrategy.java
olim7t: I'm confused, the methods are removed in the latest commit.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyAccessStrategy.java
adutra: Yes, right, of course. That's why I've been working on yet another version of `PropertyAccessStrategy` that _really_ gives full control on what is a mapped property in the first place, and how to access it. It basically boils down to make everything or almost in `PropertyMapper` public, call it stg like `MappedProperty`, and have the `PropertyAccessStrategy` return something like `Set<MappedProperty<?>> mapProperties(List<Class<?>> classHierarchy)`.

Problem is, the more it goes, the more I think all of this is so over-engineered. I'm not even sure I will push the code I have so far.

And no matter what is the outcome of this PR and what the final API looks like, I think it would be wiser to mark it as `@Beta` – otherwise I fear we will regret some, if not all, of the choices made here.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/PropertyAccessStrategy.java
adutra: Sorry, I edited my previous comment to mention that I was leaning towards the idea ofkeeping `getValue` and `setValue`. 

I am still working on it, but cf. my previous comment, this might be too far-fetched. To let users really bypass reflection altogether – which is a plausible goal after all – we need a lot more refactoring. That's why I think it goes beyond the scope of this PR. But if I end up with something that works and passes the tests during the weekend, I might open a separate PR just to let you reviewers compare and judge  by yourselves.

In the meanwhile feel free to revert my commits or add yours on top of them as you deem useful.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/DefaultPropertyTransienceStrategy.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * A {@link PropertyTransienceStrategy} that adopts a permissive, opt-out approach that
+ * will consider a property to be non-transient by default, unless:
+ * <ol>
+ * <li>The property is annotated with {@link com.datastax.driver.mapping.annotations.Transient @Transient};</li>
+ * <li>The corresponding field is non-null and is marked with the keyword {@code transient};</li>
+ * <li>The property name has been explicitly black-listed.</li>
```
tolbertam: What does it mean for a property name to be explicitly blacklisted?

olim7t: Its name was passed to the constructor. This comment needs to explain it in more detail.

tolbertam: Makes sense 👍

driver-mapping/src/main/java/com/datastax/driver/mapping/config/DefaultPropertyTransienceStrategy.java
```diff
@@ -0,0 +1,111 @@
...
+            PartitionKey.class,
+            ClusteringColumn.class,
+            com.datastax.driver.mapping.annotations.Field.class,
+            Computed.class,
+            Frozen.class,
+            FrozenKey.class,
+            FrozenValue.class
+    );
+
+    private static final HashSet<String> DEFAULT_TRANSIENT_PROPERTIES = Sets.newHashSet(
```
tolbertam: wdyt about making this public?  Although I don't expect it to be used very often, if one passes in `transientProperties` into the constructor they may be able to want to include these properties in addition to the ones they pass in.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/MappingConfiguration.java
```diff
@@ -0,0 +1,136 @@
...
+
+        private PropertyAccessStrategy propertyAccessStrategy = new DefaultPropertyAccessStrategy();
```
tolbertam: This creates new default instances every time you construct a `MappingConfiguration`.  This may be an overoptimization, but you could have a static singleton for each strategies default implementation so we don't need to construct defaults every time.  On the other hand, creating `MappingConfiguration` should not be very frequent, so it might not be worth it.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/MappingConfiguration.java
```diff
@@ -0,0 +1,136 @@
...
+
+        private PropertyTransienceStrategy propertyTransienceStrategy = new DefaultPropertyTransienceStrategy();
+
+        private HierarchyScanStrategy hierarchyScanStrategy = new DefaultHierarchyScanStrategy();
+
+        /**
+         * Sets the {@link PropertyAccessStrategy property access strategy} to use.
+         * The default is {@link DefaultPropertyAccessStrategy}.
```
tolbertam: It might be nice to briefly summarize what the default strategy does for each of these, that way the user can decide if it is worth diving in to customizing their own strategy.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperInvalidAnnotationsTest.java
```diff
@@ -261,27 +265,4 @@ public void setNotReadable(int i) {
 
     }
 
-    @Test(groups = "unit", expectedExceptions = IllegalArgumentException.class,
```
tolbertam: Why are these tests no longer applicable?  Shouldn't they still be invalid?  Also noticed that the `Invalid14` class still exists (but is no longer used)

tolbertam: It looks like in the case of unreadable it throws an NPE (on a Preconditions check) and in the case of unwritable it logs:

> [main] WARN  com.datastax.driver.mapping.config.DefaultPropertyAccessStrategy  - Property 'notWritable' is not writable and will not be mapped

but doesn't throw an exception and returns an `EntityMapper` without any columns.

driver-mapping/src/test/java/com/datastax/driver/mapping/config/MappingConfigurationHierarchyScanStrategyTest.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+
+        public void setK(int k) {
+            this.k = k;
+        }
+
+    }
+
+    @Test(groups = "short")
+    public void should_inherit_annotations_up_to_highest_ancestor_exluded() {
```
tolbertam: `exluded` -> `excluded`

driver-mapping/src/test/java/com/datastax/driver/mapping/config/MappingConfigurationPropertyAccessTest.java
```diff
@@ -0,0 +1,211 @@
...
+
+        public void setK(int k) {
+            this.k = k;
+        }
+    }
+
+    @Test(groups = "short")
+    public void should_ignore_getters() {
+        MappingConfiguration conf = MappingConfiguration.builder()
+                .withPropertyAccessStrategy(new FieldOnlyPropertyAccessStrategy())
```
tolbertam: Seeing this in use, it looks a lot less complicated.  User only has to pass in the strategy they are interested in, and the provided implementations seem pretty viable and should cover a majority of use cases I would hope 👍

driver-mapping/src/test/java/com/datastax/driver/mapping/config/MappingConfigurationPropertyAccessTest.java
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        // fluent setter
+        public Foo4 setV(int v) {
+            this.v = v;
+            return this;
+        }
+    }
+
+    public static class NoReflectionPropertyAccessStrategy implements PropertyAccessStrategy {
```
tolbertam: Neat!

driver-mapping/src/main/java/com/datastax/driver/mapping/CommonNamingConventions.java
```diff
@@ -0,0 +1,358 @@
...
+
+    /**
+     * @link "https://en.wikipedia.org/wiki/Snake_case"
+     * <p>
+     * Represents snake case naming convention, meaning all letters are lower cased,
+     * and words are separated by an underscore ("_"). E.g. "my_xml_parser".
+     */
+    public static class LowerSnakeCase extends CharDelimitedNamingConvention {
+
+        public LowerSnakeCase() {
```
tolbertam: Would be nice for there to be a static member for this instead of a custom class since there's only one way to configure it.

tolbertam: Same for the others that only have default constructors too.

driver-mapping/src/main/java/com/datastax/driver/mapping/CommonNamingConventions.java
```diff
@@ -0,0 +1,358 @@
...
+
+        @Override
+        public boolean equals(NamingConvention other) {
+            return other instanceof CamelCase &&
+                    ((CamelCase) other).ignorablePrefixPattern.equals(ignorablePrefixPattern);
+        }
+
+    }
+
+    abstract public static class CharDelimitedNamingConvention implements NamingConvention {
```
tolbertam: I don't think we should make this abstract since none of its members are abstract and might be useful to create anonymous naming conventions (although not sure why anyone would do that :D).

driver-mapping/src/main/java/com/datastax/driver/mapping/DefaultPropertyMapper.java
```diff
@@ -0,0 +1,496 @@
...
+     * @return The inferred Cassandra object name.
+     */
+    protected String inferMappedName(@SuppressWarnings("unused") Class<?> mappedClass, String propertyName, Map<Class<? extends Annotation>, Annotation> annotations) {
+        if (annotations.containsKey(Computed.class)) {
+            String expression = ((Computed) annotations.get(Computed.class)).value();
+            if (expression.isEmpty())
+                throw new IllegalArgumentException(String.format("Property '%s': attribute 'value' of annotation @Computed is mandatory for computed properties", propertyName));
+            return expression;
+        }
+        boolean caseSensitive = false;
```
tolbertam: One thing that surprised me was that if I defined a custom `NamingStrategy`, unless I have a `@Column` annotation on my field/method with `caseSensitive=true`, this automatically lowercases the property name.  In the case where you are using a naming strategy where case is important (like most are), you would think that case sensitivity would be implied.  What do you think about changing the logic such that if there is a `NamingStrategy` provided by the user that `caseSensitive=true` is implied?  Otherwise you'll have to `@Column(caseSensitive=true)` everything which seems less than ideal.

tolbertam: In addition, as you mentioned `MapperCaseSensitivityTest.should_handle_case_sensitive_identifiers_when_keyspace_specified` is failing.  The reason is because The udt field name `zipCode` is being converted to `zipcode` because the default `NamingStrategy` enforces lowercasing of field name when getting the cassandra name representation.

I think it makes sense that if a user doesn't provide a `NamingStrategy` explicitly instead we could have a default strategy that does nothing to the input fields when mapping to cassandra column (and vice versa).   We could then assume if a `NamingStrategy` *is* provided, we enforce case sensitivity, wdyt?

tolbertam:
The API seems really nice so far, will keep reviewing and will push some more tests tomorrow.

tolbertam:
can you rebase this?  Would be good to do now since it's lagging a little bit behind 3.x.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/MappingConfiguration.java
adutra: The problem is that I wanted to keep `DefaultPropertyMapper` really simple, so it's mutable, and not thread-safe. If we wanted to use a singleton pattern we would have to introduce a builder pattern as well and make the class immutable.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperInvalidAnnotationsTest.java
adutra: You are right, these exceptions have became log warnings, see [here](https://github.com/datastax/java-driver/blob/4eaec3f16bb07acf3bec06d9bd51952dc436e35f/driver-mapping/src/main/java/com/datastax/driver/mapping/DefaultPropertyMapper.java#L237-L244). Not sure though if we should throw an exception or just log a warning after all.

driver-mapping/src/main/java/com/datastax/driver/mapping/config/MappingConfiguration.java
tolbertam: I think it is fine the way it is then 👍

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperInvalidAnnotationsTest.java
tolbertam: I think throwing an Exception is probably a better solution as the user would likely want to be made aware that they are improperly mapping their entity.

driver-mapping/src/test/java/com/datastax/driver/mapping/MappingConfigurationNamingStrategyTest.java
```diff
@@ -0,0 +1,153 @@
...
+        public void setprimary_key(int k) {
+            this._primaryKey = k;
+        }
+
+        @Column(caseSensitive = true)
+        public int getmy_value() {
+            return _myValue;
+        }
+
+        public void setmy_value(int v) {
```
tolbertam, line 175: Is there a way to configure `UpperSnakeCase` such that it allows you to configure a prefix that gets ignored?   Sort of like how `CamelCase` allows you to pass ignorable prefixes, this way one could do `set_my_value` instead of `setmy_value`.  In fact it might be better to just make that a default behavior for `SnakeCase` implementations (make `_` an ignorable prefix) and `LispCase` as well (so you could do `set-my-value`).

tolbertam: Hrmm maybe this is taking things a bit too far (can go either way).  I guess if a user wants to do that they could make their own naming convention.   Also I expect users to more likely use field based entities instead of getter/setters when using custom naming strategies.

driver-mapping/src/main/java/com/datastax/driver/mapping/DefaultPropertyMapper.java
olim7t: `Column.caseSensitive` should only apply to `Column.name` if it's explicitly provided. And an explicit column name in the annotation should override the naming strategy for this property.

olim7t: I think initializing `caseSensitive` to true on this line would address the issue. If a naming strategy intends to ignore case, it should return lower-cased identifiers, and quoting them on line 378 has no effect.
Alternatively, we could change the flow to have two different return statements (name provided via annotation => consider caseSensitive, name provided via strategy => always quote), would make things a bit clearer I think.

driver-mapping/src/main/java/com/datastax/driver/mapping/PassThroughNamingStrategy.java
```diff
@@ -0,0 +1,25 @@
...
+ *
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ */
+package com.datastax.driver.mapping;
+
+public class PassThroughNamingStrategy implements NamingStrategy {
```
olim7t: This is not used anywhere, do we need to provide it? Looks simple enough to reimplement if someone needs it.

driver-mapping/src/main/java/com/datastax/driver/mapping/DefaultPropertyMapper.java
olim7t: Pushed a commit to implement my suggestion.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperCaseSensitivityTest.java
```diff
@@ -40,7 +40,7 @@
     static class User {
 
         @PartitionKey
-        @Column(caseSensitive = true)
+        @Column(name = "userId", caseSensitive = true)
```
olim7t, line 43: This is a behavioral change compared to the current version, but I think it's the reasonable thing to do. If someone uses case-sensitive Cassandra names with a strategy, they will likely have the strategy return mixed-case names rather than annotate all their fields.

tolbertam: I like it!  Also fixes my other complaint about having to annotate everything with `@Column(caseSensitive=true)`.

driver-mapping/src/main/java/com/datastax/driver/mapping/CommonNamingConventions.java
```diff
@@ -0,0 +1,328 @@
...
+import java.util.Arrays;
+import java.util.Comparator;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.regex.Pattern;
+
+/**
+ * Implementations of industry common naming conventions.
+ */
+public class CommonNamingConventions {
```
olim7t: `NamingConventions` would be sufficient IMO

driver-mapping/src/main/java/com/datastax/driver/mapping/MappingConfiguration.java
```diff
@@ -17,11 +17,6 @@
 
 /**
  * The configuration to use for the mappers.
- * <p/>
- * The following parts can be configured:
- * <ul>
- * <li>{@link PropertyMapper}: the mapper component that maps Java properties to table columns and UDT fields.</li>
- * </ul>
```
olim7t: This is a text description of the builder methods, I think the API is obvious enough that we can remove it, and avoid having to keep it in sync.

manual/object_mapper/using/README.md
```diff
@@ -293,3 +293,49 @@ public ListenableFuture<Result<User>> getAllAsync();
...
+To map to `user_name` instead, use the following:
+
+```java
+PropertyMapper propertyMapper = new DefaultPropertyMapper()
+        .setNamingStrategy(new DefaultNamingStrategy(
+                CommonNamingConventions.LOWER_CAMEL_CASE, 
+                CommonNamingConventions.LOWER_SNAKE_CASE));
+```
+
+There is more to `DefaultPropertyMapper`; see the Javadocs and implementation for details.
```
olim7t: I didn't want to cover every aspect, customizing the configuration is involved anyway, people who do it will likely look at the implementation.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperCaseSensitivityTest.java
adutra: Yes this is a behavioral change but it solves the problem elegantly, so 👍

driver-mapping/src/test/java/com/datastax/driver/mapping/MappingConfigurationNamingStrategyTest.java
adutra: Hmm not sure it's worth the hassle. I'd vote for leaving it the way it is.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperInvalidAnnotationsTest.java
adutra: Makes sense I guess. I'll fix that.

avivcortica:
Happy to see things kicking in...looking forward to using the new API.
Let me know if I can still help somehow.

avivcarmis:
Happy to see things kicking in...looking forward to using the new API.
Let me know if I can still help somehow.

adutra:
I think we are close to code-complete here. The only minor nit I would like to offer for discussion is: the builder pattern in `MapppingConfiguration`. Like I explained for `PropertyMapper`, these configuration classes are not expected to execute concurrently and therefore do no need to be strict about immutability or thread-safety. Under this perspective, the builder pattern seems to bit a bit overkill; in addition, a simple, mutable version of `MappingConfiguration` with a public default constructor would be easier to extend (if that ever makes sense one day).

olim7t:
rebased on 3.x

tolbertam:
A few test issues popped up after rebase, looking into it, probably an easy explanation.

tolbertam:
It looks like the issue has to do with fields being quoted (presumably for case sensitivity) in a UDT when they don't always need to be:

image

Will see where/why that could be happening.

driver-mapping/src/main/java/com/datastax/driver/mapping/DefaultPropertyMapper.java
```diff
@@ -0,0 +1,542 @@
...
+        }
+        if (mappedName != null) {
+            return caseSensitive ? Metadata.quote(mappedName) : mappedName.toLowerCase();
+        }
+
+        // Otherwise delegate to the naming strategy
+        mappedName = namingStrategy.toCassandraName(propertyName);
+        if (mappedName == null || mappedName.isEmpty())
+            throw new IllegalArgumentException(String.format("Property '%s': could not infer mapped name", propertyName));
+        return Metadata.quote(mappedName);
```
tolbertam: If the Field/Column doesn't have a name set, we depend on the field name,  this explains the tests failing for UDT mapper tests([see this comment](https://github.com/datastax/java-driver/pull/814#issuecomment-287156158)).  I think we could just use `handleId` in `MappedUDTCodec` when looking up `columnMappers` and it should fix it, wdyt?

tolbertam: Actually handleId does the opposite of what i want there (strips quotes).

tolbertam: Maybe what we should be doing here is using `Metadata.escapeId` (note: this method is not accessible from this scope) instead of `quote`, which would only quote the name if it needs to be quoted.  `MappedUDTCodec` would then work as is I believe.  All the `MapperUDTTest` tests pass with that change.

driver-mapping/src/main/java/com/datastax/driver/mapping/PropertyMapper.java
```diff
@@ -15,182 +15,28 @@
...
-        if (udtField != null)
-            return udtField.codec();
-        return Defaults.NoCodec.class;
-    }
-
-    @Override
-    public String toString() {
-        return propertyName;
-    }
+public interface PropertyMapper {
```
newkek, line 24: I don't understand the need of an interface here, could there be another kind of implementation? Also I don't understand why we expose those 2 methods considering their default (and only) implementation will call the same method internally?

driver-mapping/src/main/java/com/datastax/driver/mapping/DefaultPropertyMapper.java
tolbertam: It looks like doing something like that will fix all the test issues since they seem to be tied to quoting.

driver-mapping/src/main/java/com/datastax/driver/mapping/DefaultPropertyMapper.java
```diff
@@ -0,0 +1,542 @@
...
+     * to use the {@link Transient @Transient} annotation
+     * on a specific property.
+     * <p/>
+     * Subclasses can also override {@link #isTransient(String, Field, Method, Method, Map)} to gain
+     * complete control over which properties should be considered transient.
+     *
+     * @param transientPropertyNames a set of property names to exclude from mapping; may not be {@code null}. This
+     *                               will completely replace any names already configured for this object.
+     */
+    public DefaultPropertyMapper setTransientPropertyNames(Set<String> transientPropertyNames) {
```
newkek, line 179: Is it possible now to specify a white list of properties to map? So that you can either: specify a list of properties to exclude, **or** specify a list of only the properties that will be included in the mapping (not considering the annotations).
I thought the original fix was allowing that but I just checked because I remember the `WHITE_LIST` enum but it was actually allowing to do the same thing than now, but what do you think of the idea?

driver-mapping/src/main/java/com/datastax/driver/mapping/MappedClassesOnlyHierarchyScanStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * A {@link HierarchyScanStrategy} that excludes all ancestors of mapped classes, thus
+ * restricting class scan to the mapped classes themselves.
+ * <p>
+ * This strategy can be used instead of {@link DefaultHierarchyScanStrategy} to
+ * achieve pre-<a href="https://datastax-oss.atlassian.net/browse/JAVA-541">JAVA-541</a>
+ * behavior.
+ */
+public class MappedClassesOnlyHierarchyScanStrategy implements HierarchyScanStrategy {
```
newkek, line 29: It seems like the purpose of this class is the same than if the user had used `new DefaultHierarchyScanStrategy(TheMappedClass.class, true)`, is it ok to expose those 2 ways for doing the same thing?

newkek (commented):
lgtm, had some comments. Mostly I was wondering if we could get rid of exposing the `PropertyMapper` class because I'm not sure if anybody would want to implement a different `PropertyMapper` considering we have the different mapping Strategies options. Imo the strategies offer great configuration already, so I'd be in favor of allowing to set the strategies via the `MappingConfiguration` directly, and not expose the `PropertyMapper/DefaultPropertyMapper`. The rest looks good, I also may have missed a part of the conversation but I was wondering if there is really a use for the `javaConvention` part of the `NamingStrategy`, do we really expect that many people to explicitly **not** follow the naming conventions of Java, in a Java program? There are probably use cases I am not aware of.

driver-mapping/src/main/java/com/datastax/driver/mapping/DefaultPropertyMapper.java
olim7t: Yes, escapeId is the solution here. We can't always quote the name because it's compared against the schema metadata coming from Cassandra, where identifiers are only quoted if necessary.
I think we can make handleId public instead of duplicating the logic, it looks like something that might be useful for end users occasionally.

driver-mapping/src/main/java/com/datastax/driver/mapping/DefaultPropertyMapper.java
olim7t: That feels less commonly needed, and as a last resort you can override `isTransient`.

I think what was called `WHITE_LIST` at some point was actually the equivalent of `PropertyTransienceStrategy.OPT_IN`.

driver-mapping/src/main/java/com/datastax/driver/mapping/MappedClassesOnlyHierarchyScanStrategy.java
olim7t: But `TheMappedClass.class` is not constant, `filterClassHierarchy` gets called for each class.

driver-mapping/src/main/java/com/datastax/driver/mapping/PropertyMapper.java
olim7t: > could there be another kind of implementation?

There used to be a test that completely bypassed reflection in favor of hard-coded casting (in case the mapped classes are known in advance), but I can't find it anymore.

> Also I don't understand why we expose those 2 methods considering their default (and only) implementation will call the same method internally

They call the same method but the second argument (the annotations allowed on the properties) is different.

olim7t:
> I'm not sure if anybody would want to implement a different PropertyMapper considering we have the different mapping Strategies options

There are things that you can do by overriding protected methods that go beyond strategies (e.g. your comment about white list).

> if there is really a use for the javaConvention part of the NamingStrategy

One example was if you use field access and your naming convention includes a prefix, for example `private String _userName` or `private String mUserName` (personally I don't like it but you see it sometimes).

driver-mapping/src/main/java/com/datastax/driver/mapping/PropertyMapper.java
adutra: My intention was to distinguish two populations: "regular" and "advanced" users. Regular users would typically use `DefaultPropertyMapper` and just configure the strategies associated with it. Advanced ones would rather extend it or re-implement it completely.
The test that completely bypasses reflection has indeed been removed, not because it doesn't work anymore, but because it's imo something we should leave for the advanced users to do. But please trust me on this, I have tried, and it works (it's just a considerable amount of code to write).

adutra:
> if there is really a use for the javaConvention part of the NamingStrategy

Hungarian notation is still pretty much common, although I admit I've been seeing it less and less frequently.

driver-mapping/src/main/java/com/datastax/driver/mapping/PropertyMapper.java
adutra: To be more specific: to bypass reflection you need at a bare minimum to 1) override `DefaultPropertyMapper.createMappedProperty()` and 2) provide your own implementation of `MappedProperty`.

clirr-ignores.xml
```diff
@@ -147,4 +147,11 @@
         <justification>False positive, the enclosing class is package-private so this was never exposed</justification>
     </difference>
 
+    <difference>
```
tolbertam: I think this is ok?  The method was protected, so in theory classes that extend `BuiltStatement` could access it, but you can't extend `BuiltStatement` directly since its abstract method `buildQueryString` is package private.  I guess you could extend a subclass, but it isn't clear which one you could extend since they all seem to have package private constructors.

manual/object_mapper/using/README.md
```diff
@@ -297,3 +297,49 @@ public ListenableFuture<Result<User>> getAllAsync();
...
+will relate to database objects. The best way to plug in specific behavior is to create an instance of
+[DefaultPropertyMapper] and customize it.
+
+For example, the mapper's default behavior is to try to map all the properties of your Java objects.
+You might want to take the opposite approach and only map the ones that are specifically annotated
+with `@Column` or `@Field`:
+
+```java
+PropertyMapper propertyMapper = new DefaultPropertyMapper()
+        .setPropertyTransienceStrategy(PropertyTransienceStrategy.OPT_OUT);
```
tolbertam: Shouldn't this be `OPT_IN` instead of `OPT_OUT` based on the description above?

manual/object_mapper/using/README.md
```diff
@@ -297,3 +297,49 @@ public ListenableFuture<Result<User>> getAllAsync();
 ```
 
 [@QueryParameters]: http://docs.datastax.com/en/drivers/java/3.0/com/datastax/driver/mapping/annotations/QueryParameters.html
+
+
+### Mapping configuration
```
tolbertam, line 302: This is probably the right amount of documentation 👍 .  I think if the user wants to go beyond this they probably need to take time to understand the code.   We can always add more based on questions we get from community.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperCaseSensitivityTest.java
tolbertam: Added note about that to upgrade guide in 3b86cef (hope that looks ok, feel free to adjust)

driver-mapping/src/main/java/com/datastax/driver/mapping/PropertyMapper.java
newkek: Fair enough, I am not convinced for the target of "advanced" users but we'll see.

driver-mapping/src/main/java/com/datastax/driver/mapping/DefaultPropertyMapper.java
newkek: Seems it could be useful for when doing some sort of inheritance in the Java objects when having only one Cassandra table to map to. By providing a list of only the properties that are supposed to be mapped, you can map a table with a Java object, or its child, or its child's childs, etc, without having to explicitly exclude the children's properties via a black list each time. Or when you're denormalizing tables in C*, it could help selecting only some fields to map, without worrying about the other fields in the different denormalized tables, or other fields in the mapped object. Seems useful to me but that's a subjective opinion and we don't really have external feedback requesting this specifically so... ok.

driver-mapping/src/main/java/com/datastax/driver/mapping/MappedClassesOnlyHierarchyScanStrategy.java
newkek: Ok, fair enough.

manual/object_mapper/using/README.md
tolbertam: Fixed (assuming i'm right since default is OPT_OUT)

tolbertam:
Pushed up some more tests.  I think at this point we are pretty good on coverage (can always add more to test more permutations but I think what we have is adequate enough).   All of my comments/concerns have been addressed and things are working pretty well.   👍 'ing the PR, but will follow convo/commits as they come.

tolbertam (approved):

clirr-ignores.xml
olim7t: I've re-added the method and deprecated it. It doesn't hurt to keep it and it's the cleanest/safest way.

tolbertam: Sounds good 👍

tolbertam:
I introduced a test failure (not using frozen on UDT)...fixing.

olim7t:
@adutra
> these configuration classes are not expected to execute concurrently and therefore do no need to be strict about immutability or thread-safety

To put it differently, attempting to change the configuration at runtime would have unpredictable consequences.
`PropertyMapper` is pretty safe since the configuration exposes an interface with no mutating methods. As for `MappingConfiguration`, I'd rather keep it immutable to make it clear that it's not supposed to change; a builder is overkill for only one field, but it's better for future backward compatibility.
So I vote to keep things as is.

adutra (approved):

olim7t:
I used the "squash and merge" strategy and I just realized it doesn't preserve the original commit author 😞  Sorry @avivcarmis I would have mentioned you in the commit message if I had known that.

PR #815, opened 2017-02-24
Remove use of MathJax from javadocs, update guava javadoc link

tolbertam:
Motivation:

In Java 8 update 121 javadoc is now more strict about including script
tags in comments, headers, etc.  A new option --allow-script-in-comments
was added to allow this, but it does not support older versions of JDK
8.  Limitations in maven make it difficult to selectively inject options
based on JDK update versions.  Since we only use script tags for MathJax
in one method (LatencyAwarePolicy.Builder.withScale) we opted to remove
our use of MathJax and use `{@code}` instead.

Modifications:

Remove MathJax from javadoc header.

Remove javascript that changed link target based on URL.

Update LatencyAwarePolicy.Builder.withScale to use `{@code}` instead of
LateX.

Update guava javadoc link.

Result:

Build now works with Java 8 Update 121+.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/policies/LatencyAwarePolicy.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/policies/LatencyAwarePolicy.java
@@ -589,18 +589,16 @@ public class LatencyAwarePolicy implements ChainableLoadBalancingPolicy {
          * Sets the scale to use for the resulting latency aware policy.
          * <p/>
          * The {@code scale} provides control on how the weight given to older latencies
-         * decreases over time. For a given host, if a new latency \(l\) is received at
-         * time \(t\), and the previously calculated average is \(prev\) calculated at
-         * time \(t'\), then the newly calculated average \(avg\) for that host is calculated
+         * decreases over time. For a given host, if a new latency {@code l} is received at
+         * time {@code t}, and the previously calculated average is {@code prev} calculated at
+         * time {@code t'}, then the newly calculated average {@code avg} for that host is calculated
          * thusly:
-         * \[
-         * d = \frac{t - t'}{scale} \\
-         * \alpha = 1 - \left(\frac{\ln(d+1)}{d}\right) \\
-         * avg = \alpha * l + (1-\alpha) * prev
-         * \]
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -228,23 +228,9 @@
...
-                        for(var i in document.links) {
-                        var link = document.links[i];
-                        if (link.href.indexOf('is-external=true') != -1) link.target = '_blank';
-                        }
-                        &lt;/script&gt;
-                    </bottom>
                     <links>
                         <link>https://docs.oracle.com/javase/8/docs/api/</link>
-                        <link>http://docs.guava-libraries.googlecode.com/git-history/v16.0.1/javadoc/</link>
+                        <link>http://google.github.io/guava/releases/16.0.1/api/docs/</link>
```
tolbertam, line 233: Note when merging upstream that this should point to guava 19.0 since that is what we use in our depedencies for 3.x

driver-core/src/main/java/com/datastax/driver/core/policies/LatencyAwarePolicy.java
```diff
@@ -589,18 +589,16 @@ public Builder withExclusionThreshold(double exclusionThreshold) {
          * Sets the scale to use for the resulting latency aware policy.
          * <p/>
          * The {@code scale} provides control on how the weight given to older latencies
-         * decreases over time. For a given host, if a new latency \(l\) is received at
-         * time \(t\), and the previously calculated average is \(prev\) calculated at
-         * time \(t'\), then the newly calculated average \(avg\) for that host is calculated
+         * decreases over time. For a given host, if a new latency {@code l} is received at
+         * time {@code t}, and the previously calculated average is {@code rev} calculated at
```
olim7t: typo: rev => prev
I will fix that while merging.

olim7t (approved):

driver-core/src/main/java/com/datastax/driver/core/policies/LatencyAwarePolicy.java
tolbertam: I also messed up `vg` -> `avg` went ahead and fixed/force pushed that.

PR #817, opened 2017-02-27
Javadoc fixes

adutra:
Some minor fixes for javadocs and also for compiler warnings and tarball generation.


tolbertam:
Should this start at the 3.0.x branch?

tolbertam:
Oddly some tests (mostly around retries) are failing.  Looking into it.

driver-core/src/test/java/com/datastax/driver/core/policies/AbstractRetryPolicyIntegrationTest.java
```diff
@@ -113,18 +114,18 @@ protected void simulateError(int hostNumber, Result result) {
     protected void simulateError(int hostNumber, Result result, Config config) {
         PrimingRequestBuilder builder = PrimingRequest.queryBuilder()
                 .withQuery("mock query")
-                .withResult(result);
+                .withThen(then().withResult(result));
```
tolbertam: Guessing test failures have to do with this, which is odd since it should be functionaly the same.  Maybe an issue with scassandra...diving..

driver-core/src/test/java/com/datastax/driver/core/policies/AbstractRetryPolicyIntegrationTest.java
```diff
@@ -113,18 +114,18 @@ protected void simulateError(int hostNumber, Result result) {
...
 
         if (config != null)
-            builder = builder.withConfig(config);
+            builder = builder.withThen(then().withConfig(config));
```
tolbertam: Ah, it appears that when there is a config, this overrides the failure result above at 117

tolbertam: I think changing simulateError to this would fix it:

```java
    protected void simulateError(int hostNumber, Result result, Config config) {
        PrimingRequest.Then.ThenBuilder then = then().withResult(result);
        PrimingRequestBuilder builder = PrimingRequest.queryBuilder().withQuery("mock query");

        if (config != null)
            then = then.withConfig(config);

        builder = builder.withThen(then);

        scassandras.node(hostNumber).primingClient().prime(builder.build());
    }
```

tolbertam: All the failures were children of `AbstractRetryPolicyIntegrationTest` ran through them and it does appear that this fixes them all.

pom.xml
```diff
@@ -282,6 +283,49 @@
                         <link>http://fasterxml.github.io/jackson-databind/javadoc/2.6/</link>
                         <link>https://javaee-spec.java.net/nonav/javadocs/</link>
                     </links>
+                    <!-- optional dependencies from other modules (must be explicitly declared here in order to be correctly resolved) -->
+                    <additionalDependencies>
```
tolbertam, line 287: For my own benefit, what difference does moving this up make?  Is it just so that all the modules can benefit from these instead of just core?

adutra:
> Should this start at the 3.0.x branch?

Some fixes are specific to 3.x. I suggest we backport the relevant ones to 3.0.x?

driver-core/src/test/java/com/datastax/driver/core/policies/AbstractRetryPolicyIntegrationTest.java
adutra: Thanks I will amend!

pom.xml
adutra: Yes, it's only to centralize this rather complex configuration in one place. Granted some of these additional dependencies aren't really required in all child modules, but I think it is easier to maintain if everything is declared in one single place.

tolbertam:
> Some fixes are specific to 3.x. I suggest we backport the relevant ones to 3.0.x?

I think the merge process would be easier if we started from 3.0.x and worked upwards, but as long as the fixes get into 3.0.x as well i'll be happy.

adutra:
> I think the merge process would be easier if we started from 3.0.x and worked upwards, but as long as the fixes get into 3.0.x as well i'll be happy.

No strong opinions here, so if nobody disagrees I can rebase onto 3.0.x later today or tomorrow morning.

tolbertam:
I think it will be easier for us if we always bring in changes in one direction.  If we were to backport, when we merged upstream later we would run into conflicts and it might not be obvious how/what to resolve. If we bring changes upstream we only run into conflicts when there are changes on a newer branch that do not apply to the older branch, so it is easier to reconcile.

olim7t:
👍 things are much easier if we always merge downstream. This makes things a tad more complicated, but it's the downside of maintaining multiple branches.

adutra:
I still don't see why we should be caring about 3.0.x for anything else than bugfixes, but anyway, as you all wish, closing.

PR #818, opened 2017-02-28
Fix various javac and javadoc warnings

adutra:


tolbertam (approved):
lgtm 👍

driver-core/src/test/java/com/datastax/driver/core/ClusterInitTest.java
```diff
@@ -116,7 +117,9 @@ public void should_handle_failing_or_missing_contact_points() throws UnknownHost
             // - 0 or 1 for the missing host. We can't know for sure because contact points are randomized. If it's tried
             //   before the live host there will be a connection attempt, otherwise it will be removed directly because
             //   it's not in the live host's system.peers.
+            //noinspection ResultOfMethodCallIgnored
             verify(socketOptions, atLeast(6)).getKeepAlive();
+            //noinspection ResultOfMethodCallIgnored
             verify(socketOptions, atMost(7)).getKeepAlive();
```
olim7t: These are IntelliJ IDEA warnings, I'd rather avoid IDE-specific stuff in the source code as much as possible (I use IDEA as well, but if someone added random comments for another IDE I would find it annoying).

driver-core/src/test/java/com/datastax/driver/core/HostAssert.java
```diff
@@ -141,31 +142,37 @@ public void onDown(Host host) {
         return this;
     }
 
+    @SuppressWarnings("deprecation")
     public HostAssert hasWorkload(String workload) {
         assertThat(actual.getDseWorkload()).isNotNull().isEqualTo(workload);
         return this;
     }
```
olim7t, line 149: If the methods are deprecated, do we still need to test them?

driver-core/src/test/java/com/datastax/driver/core/TestUtils.java
```diff
@@ -129,13 +129,13 @@ public static void setValue(SettableByIndexData<?> data, int i, DataType type, O
                 data.setUUID(i, (UUID) value);
                 break;
             case LIST:
-                data.setList(i, (List) value);
+                data.setList(i, (List<?>) value);
```
olim7t, line 132: 👍

driver-core/src/test/java/com/datastax/driver/core/policies/IdempotenceAwareRetryPolicyIntegrationTest.java
```diff
@@ -34,6 +34,7 @@
 /**
  * Integration test with an IdempotenceAwareRetryPolicy.
  */
+@SuppressWarnings("deprecation")
 public class IdempotenceAwareRetryPolicyIntegrationTest extends AbstractRetryPolicyIntegrationTest {
```
olim7t: I don't get any deprecation warnings here on 3.0.x. They do happen in 3.1.x though, the policy was deprecated (should we still test it?).

driver-dist/pom.xml
```diff
@@ -70,44 +70,6 @@
...
-                                    <groupId>joda-time</groupId>
-                                    <artifactId>joda-time</artifactId>
-                                    <version>${joda.version}</version>
-                                </additionalDependency>
-                                <additionalDependency>
-                                    <groupId>javax.json</groupId>
-                                    <artifactId>javax.json-api</artifactId>
-                                    <version>${jsr353-api.version}</version>
-                                </additionalDependency>
-                            </additionalDependencies>
```
olim7t, line 110: For some reason this change never made it to the 3.1.x branch, will fix when we merge downstream.

olim7t (changes requested):
Looks good overall, the `mvn javadoc:javadoc` output looks much better without warnings.
A few comments/questions.

driver-core/src/test/java/com/datastax/driver/core/ClusterInitTest.java
adutra: I know but there is no standard equivalent for this warning, it's specific to IntelliJ. In hindsight, I think the right choice is to remove these comments altogether.

driver-core/src/test/java/com/datastax/driver/core/HostAssert.java
adutra: 1) They are only deprecated in OSS driver
2) Not sure if, because some method is deprecated, it should not be tested.

driver-core/src/test/java/com/datastax/driver/core/policies/IdempotenceAwareRetryPolicyIntegrationTest.java
adutra: Good catch, I will remove it from the 3.0.x and see if this test could be removed altogether in 3.1.x.

driver-dist/pom.xml
adutra: Sorry I don't get it, how could it have made it to the 3.1.x branch if it isn't merged in 3.0.x yet?

olim7t: Sorry I meant the initial addition of the `<additionalDependencies>` block (which you moved in this commit).

driver-core/src/test/java/com/datastax/driver/core/HostAssert.java
olim7t: OK, I'm fine leaving it as is for now.

olim7t (approved):

PR #819, opened 2017-02-28
Fix OSGi tests and simplify bundle version management

adutra:
This commit also changes the project packaging to jar.

This project does not generate any artifact, so setting the packaging
to bundle as before would trigger an error message generated by the
bundle plugin.


driver-tests/osgi/src/test/java/com/datastax/driver/osgi/BundleOptions.java
```diff
@@ -35,19 +35,19 @@ public static UrlProvisionOption driverBundle() {
 
     public static UrlProvisionOption driverBundle(boolean useShaded) {
         String classifier = useShaded ? "-shaded" : "";
-        return bundle("reference:file:" + PathUtils.getBaseDir() + "/../../driver-core/target/cassandra-driver-core-" + projectVersion() + classifier + ".jar");
+        return bundle("reference:file:" + PathUtils.getBaseDir() + "/../../driver-core/target/cassandra-driver-core-" + Cluster.getDriverVersion() + classifier + ".jar");
```
tolbertam, line 38: Nice, hadn't thought of at the time to use`getDriverVersion()`, definitely is much more elegant.

driver-tests/osgi/pom.xml
```diff
@@ -203,6 +197,16 @@
                     <groups>${test.groups}</groups>
                     <systemPropertyVariables>
                         <cassandra.version>${cassandra.version}</cassandra.version>
+                        <!-- pull in declared version properties, this ensures we test with the same versions used by the driver -->
+                        <guava.version>${guava.version}</guava.version>
```
tolbertam, line 201: This is very nice.  Wish I had thought of it from the start.  Much better than it being hardcoded in the source.

tolbertam (approved):
Ran the OSGi tests locally and this works good 👍

olim7t (approved):

PR #821, opened 2017-03-03
Allow Guava 21 to be imported in OSGi bundle

tolbertam:
While implementing [JAVA-1328](https://datastax-oss.atlassian.net/browse/JAVA-1328), Guava 21.0 was released.  We accounted for it everywhere except in our OSGi bundle definition, this PR addresses this.  Also added a test.

### DIFF
--- a/driver-core/pom.xml
+++ b/driver-core/pom.xml
@@ -200,7 +200,7 @@
                         <_include>-osgi.bnd</_include>
                         <Import-Package>
                             <!-- JNR does not provide OSGi bundles, so exclude it; the driver can live without it -->
-                            <![CDATA[com.google.common*;version="[16.0.1,21)",!jnr.*,*]]></Import-Package>
+                            <![CDATA[com.google.common.*;version="[16.0.1,22)",!jnr.*,*]]></Import-Package>
                     </instructions>
                     <supportedProjectTypes>
                         <supportedProjectType>jar</supportedProjectType>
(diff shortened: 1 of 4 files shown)


driver-core/pom.xml
```diff
@@ -224,7 +224,7 @@
                                     JNR does not provide OSGi bundles, so exclude it; the driver can live without it
                                     Explicitly import javax.security.cert because it's required by Netty, but Netty has been explicitly excluded
                                     -->
-                                    <![CDATA[com.google.common.*;version="[16.0.1,21)",!jnr.*,!io.netty.*,javax.security.cert,*]]></Import-Package>
+                                    <![CDATA[com.google.common.*;version="[16.0.1,22)",!jnr.*,!io.netty.*,javax.security.cert,*]]></Import-Package>
```
adutra, line 227: I see that sometimes we use `com.google.common.*` and sometimes `com.google.common*`. Both seem to work but for consistency I'd suggest using one or the other. (Disclaimer: I might be the culprit here.)

adutra (changes requested):

driver-core/pom.xml
tolbertam: 👍 on making it consistent, didn't even notice that,  which would you prefer?  I like `com.google.common.*` i think.

adutra: Same here.

tolbertam: Fixed

adutra (approved):

tolbertam:
Rebased on 3.x and squashed commits, should be ready to go.

PR #823, opened 2017-03-07
JAVA-1414: Optimize Metadata.escapeId and Metadata.handleId

adutra:
Some benchmark results (benchmark code [here](https://gist.github.com/adutra/03ec7e8046c7db2b92646b8a929d1e54)):

```
1) No Optimization
Benchmark                               Mode  Cnt    Score     Error  Units
MetadataBenchmark.benchmarkSerialize    avgt   20  236.909 ±  20.722  ns/op
MetadataBenchmark.benchmarkDeserialize  avgt   20  150.895 ±   3.527  ns/op

2) Optimized handleId and escapeId
Benchmark                               Mode  Cnt    Score   Error  Units
MetadataBenchmark.benchmarkSerialize    avgt   20  166.681 ± 2.491  ns/op
MetadataBenchmark.benchmarkDeserialize  avgt   20  128.973 ± 1.439  ns/op

3) Optimized handleId and escapeId + Metadata.quote on deserialization
Benchmark                               Mode  Cnt    Score   Error  Units
MetadataBenchmark.benchmarkSerialize    avgt   20  165.921 ± 3.636  ns/op
MetadataBenchmark.benchmarkDeserialize  avgt   20  106.071 ± 1.998  ns/op
```

The current PR implements option 2 above. As you can see there would be an extra improvement by using `Metadata.quote()` instead of `Metadata.escapeId()` when deserializing, but that could break existing applications (although, this is rather unlikely imo).

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Metadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Metadata.java
@@ -27,7 +27,6 @@ import java.util.*;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.locks.ReentrantLock;
-import java.util.regex.Pattern;
 
 /**
  * Keeps metadata on the connected cluster, including known nodes and schema definitions.
@@ -44,9 +43,6 @@ public class Metadata {
     private volatile TokenMap tokenMap;
 
     final ReentrantLock lock = new ReentrantLock();
-
-    private static final Pattern alphanumeric = Pattern.compile("\\w+"); // this includes _
-    private static final Pattern lowercaseAlphanumeric = Pattern.compile("[a-z][a-z0-9_]*");
 
     // See https://github.com/apache/cassandra/blob/trunk/doc/cql3/CQL.textile#appendixA
     private static final Set<String> RESERVED_KEYWORDS = ImmutableSet.of(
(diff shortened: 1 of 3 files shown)


tolbertam:
I wonder if I have a difference in environment, but running your benchmarks I observe a bigger difference when comparing the 3.x and java1414 branch:

```
3.x

Benchmark                               Mode  Cnt    Score    Error  Units
MetadataBenchmark.benchmarkDeserialize  avgt   20  516.940 ± 18.175  ns/op
MetadataBenchmark.benchmarkSerialize    avgt   20  553.485 ± 15.860  ns/op

java1414

Benchmark                               Mode  Cnt    Score   Error  Units
MetadataBenchmark.benchmarkDeserialize  avgt   20  136.443 ± 3.723  ns/op
MetadataBenchmark.benchmarkSerialize    avgt   20  170.706 ± 3.707  ns/op
```

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -131,22 +127,56 @@ static String handleId(String id) {
...
+                : ident;
+    }
+
+    /**
+     * We don't need to escape an identifier if it
+     * matches non-quoted CQL3 ids ([a-z][a-z0-9_]*),
+     * and if it's not a CQL reserved keyword.
+     */
+    private static boolean needsQuote(String s) {
+        char c = s.charAt(0);
```
tolbertam: It seems unlikely, but since `escapeId` is used in a ton of places I don't know for sure it is possible that an empty string would be passed into here and if it was wouldn't this throw a `StringIndexOutOfBoundsException`?

tolbertam: Just went through all methods that called `escapeId` and it's probably a pretty safe assumption that the string will not be empty, but would be nice to either check `s.length()` here or add an assertion in `escapeId`

tolbertam (approved):
lgtm 👍

olim7t (approved):

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
adutra: `escapeId` is only applied to identifiers coming from C*, so I don't think we should test for the string length, but I agree that an assertion could help clarify things.

PR #824, opened 2017-03-07
JAVA-1415: Correctly report if a UDT column is frozen

olim7t:
Notes:
* `isFrozen()` used to be hard-coded to true, but the default is now false. I think it makes more sense in a context where frozenness does not apply, for example when retrieving the UDT definition directly from the cluster metadata.
* I've not updated `equals` and `hashCode`. Whether a UDT is frozen or not, it's still the same type, I can't think of a situation where it would make sense to differentiate (the opposite seems more likely, e.g. you retrieve the type of a frozen column, it should be equal to the metadata one).

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/DataType.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/DataType.java
@@ -188,7 +188,7 @@ public abstract class DataType {
                     DataType fieldType = decode(buffer, protocolVersion, codecRegistry);
                     fields.add(new UserType.Field(fieldName, fieldType));
                 }
-                return new UserType(keyspace, type, fields, protocolVersion, codecRegistry);
+                return new UserType(keyspace, type, false, fields, protocolVersion, codecRegistry);
             case TUPLE:
                 nFields = buffer.readShort() & 0xffff;
                 List<DataType> types = new ArrayList<DataType>(nFields);
--- a/driver-core/src/main/java/com/datastax/driver/core/DataTypeClassNameParser.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/DataTypeClassNameParser.java
@@ -113,7 +113,8 @@ class DataTypeClassNameParser {
             List<UserType.Field> fields = new ArrayList<UserType.Field>(rawFields.size());
             for (Map.Entry<String, String> entry : rawFields.entrySet())
                 fields.add(new UserType.Field(entry.getKey(), parseOne(entry.getValue(), protocolVersion, codecRegistry)));
-            return new UserType(keyspace, typeName, fields, protocolVersion, codecRegistry);
+            // create a frozen UserType since C* 2.x UDTs are always frozen.
+            return new UserType(keyspace, typeName, true, fields, protocolVersion, codecRegistry);
         }
 
         if (isTupleType(next)) {
--- a/driver-core/src/main/java/com/datastax/driver/core/DataTypeCqlNameParser.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/DataTypeCqlNameParser.java
@@ -137,7 +137,7 @@ class DataTypeCqlNameParser {
         // We need to remove escaped double quotes within the type name as it is stored unescaped.
         // Otherwise it's a UDT. If we only want a shallow definition build it, otherwise search known definitions.
         if (shallowUserTypes)
-            return new UserType.Shallow(currentKeyspaceName, Metadata.handleId(type));
+            return new UserType.Shallow(currentKeyspaceName, Metadata.handleId(type), frozen);
 
         UserType userType = null;
         if (currentUserTypes != null)
@@ -148,7 +148,7 @@ class DataTypeCqlNameParser {
         if (userType == null)
             throw new UnresolvedUserTypeException(currentKeyspaceName, type);
         else
-            return userType;
+            return userType.copy(frozen);
     }
 
     private static class Parser {
--- a/driver-core/src/test/java/com/datastax/driver/core/UserTypesTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/UserTypesTest.java
@@ -30,11 +30,10 @@ import java.nio.ByteBuffer;
 import java.util.*;
 import java.util.concurrent.Callable;
 
+import static com.datastax.driver.core.Assertions.assertThat;
 import static com.datastax.driver.core.ConditionChecker.check;
 import static com.datastax.driver.core.Metadata.quote;
 import static java.util.concurrent.TimeUnit.MINUTES;
-import static org.testng.Assert.assertEquals;
-import static org.testng.Assert.assertNotEquals;
 
 @CassandraVersion("2.1.0")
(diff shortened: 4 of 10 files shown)


driver-core/src/main/java/com/datastax/driver/core/DataType.java
```diff
@@ -188,7 +188,7 @@ static DataType decode(ByteBuf buffer, ProtocolVersion protocolVersion, CodecReg
                     DataType fieldType = decode(buffer, protocolVersion, codecRegistry);
                     fields.add(new UserType.Field(fieldName, fieldType));
                 }
-                return new UserType(keyspace, type, fields, protocolVersion, codecRegistry);
+                return new UserType(keyspace, type, true, fields, protocolVersion, codecRegistry);
```
adutra: If the default frozenness is `false`, why are we setting it to `true` when reading resultset metadata? I know it doesn't matter much in this context, but for consistency I wonder if we shouldn't consider all user types as non-frozen when reading metadata from a result response.

adutra (approved):

driver-core/src/main/java/com/datastax/driver/core/DataType.java
tolbertam: I think the reason it is done this way is so the behavior doesn't change from the way it worked.  Previously a `UserType` would always be considered frozen, so in the case of parsing the type back from result metadata where there is no way to tell, returning true will provide similar behavior as it did in the past.

driver-core/src/test/java/com/datastax/driver/core/UserTypesTest.java
```diff
@@ -66,13 +65,8 @@ public void onTestContextInitialized() {
...
     }
 
-    /**
-     * Basic write read test to ensure UDTs are stored and retrieved correctly.
-     *
-     * @throws Exception
-     */
     @Test(groups = "short")
-    public void simpleWriteReadTest() throws Exception {
+    public void should_store_and_retrieve_with_prepared_statements() throws Exception {
```
tolbertam, line 70: Nice work, these tests definitely needed some love ;)

driver-core/src/main/java/com/datastax/driver/core/DataType.java
olim7t: Well actually that's a typo, I intended to return false (see my comment in the PR description). I know this changes the behavior but it's more logical to return false, and I doubt anyone relied on that in the past. I'll add a test and find a place to document it.

driver-core/src/test/java/com/datastax/driver/core/UserTypesTest.java
```diff
@@ -416,139 +353,137 @@ public void udtNestedTest() throws Exception {
...
+        assertThat(row.getInt("a")).isEqualTo(1);
 
         userType = userTypeDef.newValue().setList(0, new ArrayList<Object>());
-        assertEquals(row.getUDTValue("b"), userType);
+        assertThat(row.getUDTValue("b")).isEqualTo(userType);
+    }
+
+    @Test(groups = "short")
+    @CassandraVersion("3.6")
+    public void should_indicate_if_user_type_is_frozen() {
```
tolbertam: Minor nit, but what do you think about separating this into two separate tests, one that requires 2.1+ that creates a frozen udt and one that requires 3.6+ that creates a non-frozen UDT?  This will allow us to validate that the UDT is considered frozen for 2.1+.

driver-core/src/main/java/com/datastax/driver/core/DataTypeCqlNameParser.java
```diff
@@ -148,7 +148,7 @@ static DataType parse(String toParse, Cluster cluster, String currentKeyspaceNam
         if (userType == null)
             throw new UnresolvedUserTypeException(currentKeyspaceName, type);
         else
-            return userType;
+            return userType.copy(frozen);
```
tolbertam, line 151: Just curious, but if the UDT is already resolved/known, isn't it likely that the `frozen` field on that existing user type is already correct?

olim7t: `copy` handles that.

tolbertam: Makes sense, but for my understanding was wondering if there would ever be a case where the existing type's `frozen` flag doesn't match the input `frozen` flag

olim7t: I think it happens if, as part of a full keyspace refresh, you've already parsed the definition of `udt1` (it's in `currentUserTypes` but not frozen), and now you're parsing a table that has a `frozen<udt1>` column.

tolbertam: ah, that makes complete sense, thanks!

driver-core/src/test/java/com/datastax/driver/core/UserTypesTest.java
tolbertam: I did just that and apparently the `UserType` column (`v`) associated with `TableMetadata` is not marked frozen on C* 2.1.17 but it is for C* 3.6, debugging my way through it.

tolbertam: Looks like because the code in `DataTypeCqlNameParser` is used for C* 3.0+, where `DataTypeClassNameParser` is used otherwise, we could probably adjust that code to imply that UDTs are always frozen since that would be the case for C* 2.x

driver-core/src/main/java/com/datastax/driver/core/DataTypeClassNameParser.java
```diff
@@ -113,7 +113,7 @@ static DataType parseOne(String className, ProtocolVersion protocolVersion, Code
             List<UserType.Field> fields = new ArrayList<UserType.Field>(rawFields.size());
             for (Map.Entry<String, String> entry : rawFields.entrySet())
                 fields.add(new UserType.Field(entry.getKey(), parseOne(entry.getValue(), protocolVersion, codecRegistry)));
-            return new UserType(keyspace, typeName, fields, protocolVersion, codecRegistry);
+            return new UserType(keyspace, typeName, false, fields, protocolVersion, codecRegistry);
```
tolbertam: With regards to [this comment](https://github.com/datastax/java-driver/pull/824#discussion_r104987494) I think we should have this pass in true since this is only used for < C* 3.0, where udts are always frozen.  I'll push that up but feel free to back it out if you disagree.

tolbertam (approved):
Added c58d0bd which changes behavior of C* 2.x parsing to mark parsed UserTypes as frozen and split out test to test for that.  If that looks ok to you, 👍 for squash and merge.

driver-core/src/main/java/com/datastax/driver/core/DataTypeClassNameParser.java
olim7t: 👍  I agree with your changes

PR #825, opened 2017-03-08
JAVA-1418: Make Guava version detection more reliable

olim7t:


tolbertam:
looks good so far, running a test matrix against this and if all goes well will give 👍 later today.

tolbertam (approved):
lgtm 👍 only main test issue was guava 21 with OSGi which is fixed by #821

adutra (approved):

PR #826, opened 2017-03-09
JAVA-1174: Add ifNotExists option to mapper

olim7t:


driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
```diff
@@ -1028,29 +1090,37 @@ boolean isIncludedInQuery() {
             }
 
             @Override
-            void appendTo(Insert.Options usings) {
-                throw new UnsupportedOperationException("shouldn't be called");
+            void validate(QueryType qt, MappingManager manager) {
+                checkArgument(qt == QueryType.SAVE || qt == QueryType.DEL || qt == QueryType.GET,
```
adutra: Just noticed: this check is useless, these is no other `QueryType` at this moment.

driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
```diff
@@ -1064,30 +1134,39 @@ boolean isIncludedInQuery() {
...
```
adutra: Same here.

adutra (approved):
👍

driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
```diff
@@ -932,19 +924,65 @@ public static Option saveNullFields(boolean enabled) {
...
+        /**
+         * "Binds" the option before the request gets executed (might be a no-op).
+         * <p/>
+         * If the option's value is actually set as a bound variable, then this must be done at
+         * the given index, and the method should return the new position.
+         * <p/>
+         * Options may also affect the statement itself, not a bound variable; in that case, the
+         * method must return the index unchanged.
+         */
+        abstract int bind(BoundStatement bs, int currentIndex);
```
newkek: `bind` sounds a bit ambiguous to me since the implementations are going to either bind a parameter, or set an option on the statement. Couldn't we call it something like `applyOption`?

driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
```diff
@@ -932,19 +924,65 @@ public static Option saveNullFields(boolean enabled) {
...
+         * @param enabled whether to include an IF NOT EXISTS clause in queries.
+         * @return the option.
+         */
+        public static Option ifNotExists(boolean enabled) {
+            return new IfNotExists(enabled);
+        }
+
+        final Type type;
+
+        protected Option(Type type) {
```
newkek, line 942: Just noticed that I don't think `protected` is actually helpful here, I think it can go package-private. Since the other abstract methods on `Option` are package-private, I don't think we can expect anybody outside the driver to `extends Option` because of the package private abstract methods, and so the only implementations possible are in this package

driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
```diff
@@ -932,19 +924,65 @@ public static Option saveNullFields(boolean enabled) {
...
 
-        abstract void checkValidFor(QueryType qt, MappingManager manager) throws IllegalArgumentException;
+        /**
+         * Force subclasses to be explicit about equality.
+         * Options are used in the prepared statement cache key, so if the same option with different
+         * values generate different query strings, then the class must include its values in the comparison.
+         * Otherwise, comparing the class is good enough. See subclasses for examples.
+         */
+        @Override
+        public abstract boolean equals(Object other);
```
newkek: Isn't it misleading that we end up with different options with different values being equal? Meaning `Option.ttl(1).equals(Option.ttl(2)) // => true`

newkek (approved):
sgtm, had some minor comments

driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
olim7t: Fair enough, the option values are exposed to the end user after all. I've added an `asCacheKey` method to improve this.

tolbertam:
I think we need to add a rule in Clirr:

```
[ERROR] 7002: com.datastax.driver.mapping.Mapper$Option: Method 'public com.datastax.driver.mapping.Mapper$Option$Type getType()' has been removed
```

I don't think users would address Options separately (just always pass in Option / never look at it), so it is probably ok to add an exception, wdyt?

tolbertam:
On second thought, it seems like it wouldn't be a problem to add `getType()` back and that seems like the most low risk thing to do, thoughts?

driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
```diff
@@ -1101,49 +1194,110 @@ boolean isIncludedInQuery() {
...
+            boolean ifNotExists;
+
+            IfNotExists(boolean ifNotExists) {
+                super(Type.IF_NOT_EXISTS);
+                this.ifNotExists = ifNotExists;
+            }
+
+            @Override
+            void validate(QueryType qt, MappingManager manager) {
+                checkArgument(qt == QueryType.SAVE, "IfNotExists option is only allowed in save queries");
```
tolbertam, line 1253: Could technically protocol version check here like we do for TTLs, although LWTs aren't tied to protocol version but CQL/C* version instead ( so maybe it's best the way it is since we can't reliably check if the C* server will support it)

tolbertam:
Ah, type itself was protected to begin with, so even though the API method was present, you couldn't use it for comparison purposes...sounds like clirr rule is more appropriate after all.

tolbertam:
Went ahead and added the clirr exception, feel free to back out if disagree.  Also added a couple negative tests.

olim7t:
Good catch. The public getType was an oversight, but just to be safe I think it's better to keep it and deprecate it.

tolbertam (approved):
lgtm 👍

tolbertam:
Just remembered, it would be good to add the new option to the manual where we document the rest of the mapper options.

olim7t:
Done

PR #827, opened 2017-03-09
Fixes for miscellaneous unreliable tests

GregBestland:
If both the inserts occur in the same millisecond for the NullFieldsTest in C* <2.1.x the order of operations can be un predictable due to the use of server side timestamps.

We need to allow for the request have a connection allocated to it prior simulating the success response in the  Should_wait_on_connection_if_zero_core_connections() test.

The pool logic changed in such a way for  the should_adjust_connection_keyspace_on_dequeue_if_pool_state_is_different test needs to wait longer then the default pool timeout.

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/HostConnectionPoolTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/HostConnectionPoolTest.java
@@ -18,7 +18,7 @@ package com.datastax.driver.core;
 import com.codahale.metrics.Gauge;
 import com.datastax.driver.core.exceptions.*;
 import com.datastax.driver.core.policies.ConstantReconnectionPolicy;
-import com.datastax.driver.core.utils.MoreFutures;
+import com.google.common.base.Function;
 import com.google.common.base.Predicate;
 import com.google.common.base.Throwables;
 import com.google.common.util.concurrent.*;
@@ -290,7 +290,7 @@ public class HostConnectionPoolTest extends ScassandraTestBase.PerClassCluster {
             int count = 0;
             for (MockRequest queuedRequest : queuedRequests) {
                 try {
-                    Uninterruptibles.getUninterruptibly(queuedRequest.connectionFuture, 5, TimeUnit.SECONDS);
+                    Uninterruptibles.getUninterruptibly(queuedRequest.connectionFuture, 10, TimeUnit.SECONDS);
                     count++;
                 } catch (ExecutionException e) {
                     // 128th request should timeout since all in flight requests are used.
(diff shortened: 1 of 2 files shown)


driver-core/src/test/java/com/datastax/driver/core/HostConnectionPoolTest.java
```diff
@@ -1242,8 +1241,9 @@ public void should_wait_on_connection_if_zero_core_connections() throws Exceptio
 
             // Should create up to core connections.
             verify(factory, timeout(readTimeout).times(1)).open(any(HostConnectionPool.class));
-
+            Uninterruptibles.getUninterruptibly(request.connectionFuture, 5, TimeUnit.SECONDS);
```
tolbertam: This line shouldn't be needed anymore since the completion of requestInitialized depends on connectionFuture.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperSaveNullFieldsTest.java
```diff
@@ -40,6 +41,7 @@ public void setup() {
         mapper = new MappingManager(session()).mapper(User.class);
     }
 
+    @CassandraVersion("2.2.0")
```
tolbertam: Should be 2.1.0 right?  The problem only happens on 1.2/2.0 because client timestamps aren't used (happens in event when both inserts are done in same millisecond)

driver-core/src/test/java/com/datastax/driver/core/HostConnectionPoolTest.java
```diff
@@ -290,7 +290,7 @@ public void should_adjust_connection_keyspace_on_dequeue_if_pool_state_is_differ
             int count = 0;
             for (MockRequest queuedRequest : queuedRequests) {
                 try {
-                    Uninterruptibles.getUninterruptibly(queuedRequest.connectionFuture, 5, TimeUnit.SECONDS);
+                    Uninterruptibles.getUninterruptibly(queuedRequest.connectionFuture, 10, TimeUnit.SECONDS);
```
tolbertam, line 293: Just for context for anyone else looking, this was needed because of [JAVA-1371](https://datastax-oss.atlassian.net/browse/JAVA-1371).  Previously an exception would be thrown nearly right away, after 1371 the future now fails after 5 seconds.  With only waiting up to 5 seconds, this creates a very small window for waiting to timeout before the future fails.

adutra (changes requested):
lgtm like Andy stated, the only nit was the value for `@CassandraVersion`.

tolbertam:
another very small nit, can you adjust the first line of the commit description to be less < 72 characters so github doesn't truncate it?

tolbertam (approved):
lgtm 👍

olim7t:
> can you adjust the first line of the commit description to be less < 72 characters so github doesn't truncate it?

As an aside, we've not enforced that strictly in the past, and it can be hard especially if the message also contains a ticket reference. IMO it's fine to be a bit longer, within reason (i.e. not 2000 characters 😉)

PR #828, opened 2017-03-09
Support Duration type in PrimitiveTypeSamples.

tolbertam:
A number of tests use PrimitiveTypeSamples for getting sample values for
given DataTypes.  Duration was previously omitted from this because it
is not supported in Set or Map keys.  Special case tests that involve
using these samples for Map and Sets so we get better Duration coverage.

In addition, this fixes a bug in UserTypesTest where it attempts to
resolve a sample for Duration but one is not found.

Since `Duration` is part of protocol V5, this is not currently exercised in our main testing, but tried this out with V5 and it works.


adutra (approved):
👍 added a small commit to fix some comments but this is definitely good to go!

PR #829, opened 2017-03-10
JAVA-1415: UserType.toString should only include 'frozen<>' when frozen

tolbertam:
Slight addition to #824 to fix `UserType.toString` implementation.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/UserType.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/UserType.java
@@ -291,7 +291,10 @@ public class UserType extends DataType implements Iterable<UserType.Field> {
 
     @Override
     public String toString() {
-        return "frozen<" + Metadata.escapeId(getKeyspace()) + '.' + Metadata.escapeId(getTypeName()) + ">";
+        String str = Metadata.escapeId(getKeyspace()) + "." + Metadata.escapeId(getTypeName());
+        return isFrozen() ?
+                "frozen<" + str + ">" :
+                str;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


adutra (approved):

olim7t:
This change led me to wonder where our code uses `UserType.toString()`. That's a bit hard to find because it can be called indirectly/implicitly.
I ran the following structural searches in IDEA, where `$typeVar$` has the type `(UserType|DataType)`:
* `$typeVar$.toString()` (explicit call): found occurrences in **schema builder classes**
* `$arg0$ + $typeVar$ + $arg1$` (where args1 and args2 have min=0, max=unlimited): **ColumnMetadata.toString** (which is eventually used in TableMetadata.asCQLQuery) and **UserType.Field.toString** (used in UserType.asCQLQuery)
* `$otherVar$.append($typeVar$)` (stringbuilder):  **FunctionMetadata.asCQLQuery**, **querybuilder.Utils.appendName and appendValue**, and a few toString methods
* `String.format($args1$, $typeVar$, $args2$)` (where args1 and args2 have min=0, max=unlimited): a few toString methods and error messages

I think toString should be reserved for debugging, and make no guarantee about its format. So for the cases I highlighted above we should have something specific instead of assuming the toString output yields what we want.

olim7t:
After reviewing the different cases, I think we're fine: TableMetadata, FunctionMetadata and UserType have the types with the correct frozen flag, so their asCQLQuery method will produce the expected output (would be nice to add a few tests for that though).
For the schema/query builder stuff, the user will most likely provide a type that they retrieved from Cluster.getMetadata, so it will have frozen=false which is what we want. And otherwise it makes sense that they would have to call copy(false) to flip the flag.
(By the way, some of these methods were likely broken when frozen<> was hard-coded)

In the long run I still think we should avoid toString() in favor of specialized methods with a well-specified output, but that can be done outside of the scope of this change, I'll create a follow-up ticket.

PR #831, opened 2017-03-23
JAVA-1431: Improve error handling during pool initialization

adutra:


driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -370,14 +370,17 @@ public void onFailure(Throwable t) {
...
                             cluster.manager.triggerOnDown(host, false);
+                            future.set(false);
                         } else if (t instanceof ClusterNameMismatchException) {
                             ClusterNameMismatchException e = (ClusterNameMismatchException) t;
                             cluster.manager.logClusterNameMismatch(host, e.expectedClusterName, e.actualClusterName);
                             cluster.manager.triggerOnDown(host, false);
+                            future.set(false);
                         } else {
-                            logger.warn("Error creating pool to " + host, t);
+                            logger.error("Error creating pool to " + host, t);
```
tolbertam: Ah interesting, so previously for any Error (like OOME as documented in HostConnectionPool), we wouldn't propagate?

driver-core/src/test/java/com/datastax/driver/core/AuthenticationTest.java
```diff
@@ -132,4 +137,29 @@ public void run() {
...
+    @CCMConfig(dirtiesContext = true)
+    public void should_not_connect_with_wrong_credentials() throws InterruptedException {
+        PlainTextAuthProvider authProvider = new PlainTextAuthProvider("cassandra", "cassandra");
+        Cluster cluster = Cluster.builder()
+            .addContactPoints(getContactPoints())
+            .withPort(ccm().getBinaryPort())
+            .withAuthProvider(authProvider)
+            .build();
+        cluster.init();
+        authProvider.setPassword("wrong");
```
tolbertam: One concern, what if you have 1 node in the cluster which for whatever reason you can't authenticate with?  In this case, you wouldn't be able to establish a Session.

One possibility that I could think of is if for whatever reason a node you are trying to authenticate with can't access the replicas owning data for that user id in the system_auth table and thus can't authenticate the connection.   This seems very unlikely, but in such a case it seems like it could be a bigger problem than the one this ticket attempts to solve.

I think triggering the host down and starting reconnection and not propagating the exception up to the user might be a good happy medium as you can then establish connections to the rest of the nodes (in this case).  wdyt?

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
adutra: Correct, and the pool creation future would complete normally (although with `false` instead of `true`).

I suppose it was so by design, on the assumption that the failure is transient and the failed host/pool would eventually get created later, so no need to stop the whole session initialization process. 

The change I made here indeed modifies this behavior. My main intent was to account for `AuthenticationException` because when it happens, it usually happens for all hosts, and the session gets created with no pools at all and the user gets a cryptic `BusyPoolException` when he tries to use the session. 

But you are right concerning OOME: it was being swollen here as well.

adutra: I refactored this portion a bit to make things clearer wrt to what exception should be propagated and why.

driver-core/src/test/java/com/datastax/driver/core/AuthenticationTest.java
adutra: > I think triggering the host down and starting reconnection and not propagating the exception up to the user might be a good happy medium as you can then establish connections to the rest of the nodes (in this case). wdyt?

`AuthenticationException` is a strange kind of error, because it can be specific to a host in some cases (like in your example), or cluster-wise in some others; besides, it can be sometimes temporary (like when a user changes his password), sometimes permanent (wrong credentials).

But you might be right, I tested your suggestion and it seems a good comprise. On one side, it does not prevent the session from being created, which is good in your scenario. But on the other side, it also improves the user experience by having the session throw NHAE instead of `BusyPoolException` if the error happened cluster-wise.

So yes, I think you have a very good point, I will amend the code accordingly, thanks!

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -368,16 +368,22 @@ public void onSuccess(Void result) {
...
                             cluster.manager.logClusterNameMismatch(host, e.expectedClusterName, e.actualClusterName);
-                            cluster.manager.triggerOnDown(host, false);
                         } else {
-                            logger.warn("Error creating pool to " + host, t);
+                            logger.error("Error creating pool to " + host, t);
+                        }
+                        boolean fatal = t instanceof UnsupportedProtocolVersionException ||
+                                t instanceof ClusterNameMismatchException;
+                        cluster.manager.triggerOnDown(host, !fatal);
+                        if (t instanceof Error) {
```
adutra: This is not strictly related to JAVA-1431, but we agreed that errors should be propagated.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -431,7 +437,7 @@ CloseFuture removePool(Host host) {
         }
 
         // Wait pool creation before removing, so we don't lose connectivity
-        ListenableFuture<?> allPoolsCreatedFuture = Futures.successfulAsList(poolCreatedFutures);
+        ListenableFuture<?> allPoolsCreatedFuture = Futures.allAsList(poolCreatedFutures);
```
adutra, line 442: Again, required for errors to be propagated.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -377,7 +378,11 @@ public void onFailure(Throwable t) {
                         } else {
                             logger.warn("Error creating pool to " + host, t);
                         }
-                        future.set(false);
+                        if (t instanceof Error) {
```
adutra: Special-case  `java.lang.Error` otherwise it's not propagated.

adutra:
You will notice that I have broadened the scope of this ticket to include any kind of error, and in particular `java.lang.Error`, as I realized that they are not ideally handled either. I think it makes sense to address these and `AuthenticationException` in the same PR, but let me know if you disagree.

adutra:
I have played a little with `java.lang.Error`. Below are my findings so far:

1. Errors thrown from a Netty listener _are not propagated_. This is very concerning because it leaves the driver in an irrecoverable state. See `io.netty.util.concurrent.DefaultPromise#notifyListener0`.
2. Errors thrown from a Guava listener are propagated and afaict do not compromise the driver's stability. 
3. However the current code had several places where errors would not be correctly propagated. That's fixed now.

Currently only item 1 is unsolved. Other than wrapping all code in a try-catch block, I don't see how to improve things here.

One last thing: I have tests for `java.lang.Error` handling, but these require a dependency on the byteman library. Do you think it's acceptable? (License is LGPL).

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -193,8 +193,8 @@ public void operationComplete(ChannelFuture future) throws Exception {
                     future.setException(t);
                 } else {
                     // Defunct to ensure that the error will be signaled (marking the host down)
-                    Exception e = (t instanceof ConnectionException || t instanceof DriverException || t instanceof InterruptedException)
-                            ? (Exception) t
+                    Throwable e = (t instanceof ConnectionException || t instanceof DriverException || t instanceof InterruptedException || t instanceof Error)
```
adutra, line 196: Related to propagation of `java.lang.Error`.

driver-core/src/test/java/com/datastax/driver/core/MemoryAppender.java
```diff
@@ -95,4 +96,33 @@ public String getNext() {
         nextLogIdx += next.length();
         return next;
     }
+
+    public MemoryAppender enableFor(Class<?> logger) {
```
adutra: Unrelated, but a small refactoring I've been willing to do for a long time.

driver-core/pom.xml
```diff
@@ -161,6 +161,40 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>org.jboss.byteman</groupId>
+            <artifactId>byteman-install</artifactId>
+            <scope>test</scope>
+            <version>${byteman.version}</version>
+        </dependency>
+
+        <dependency>
```
tolbertam: small nit: bmunit depends on byteman-install, submit and version so you should only need to list the byteman-bmunit dependency,

driver-core/src/test/java/com/datastax/driver/core/SessionErrorTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.testng.annotations.Listeners;
+import org.testng.annotations.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+/**
+ * Simple test of the Sessions methods against a one node cluster.
+ */
+@BMUnitConfig(loadDirectory = "target/test-classes")
```
tolbertam, line 34: This is really cool, looking forward to finding more fun use cases for byteman.

driver-core/src/test/java/com/datastax/driver/core/SessionErrorTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @BeforeMethod
+    public void setUp() throws Exception {
+        cluster1 = register(createClusterBuilder()
+                .addContactPointsWithPorts(ccm().addressOfNode(1)).build()).init();
+        cluster2 = register(createClusterBuilder()
+                .addContactPointsWithPorts(ccm().addressOfNode(1)).build()).init();
+    }
+
+    @Test
```
tolbertam: Do we intend for these to run as part of our test harness?  Since the Test annotation lacks a group definition they aren't running.

driver-core/src/test/java/com/datastax/driver/core/SessionErrorTest.java
```diff
@@ -0,0 +1,89 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+/**
+ * Simple test of the Sessions methods against a one node cluster.
+ */
+@BMUnitConfig(loadDirectory = "target/test-classes")
+@Listeners(BMNGListener.class)
+@CCMConfig(createCluster = false)
+public class SessionErrorTest extends CCMTestsSupport {
```
tolbertam: It would be nice if this used Scassandra so we could run through it a bit faster (an actual cassandra node isn't needed i think).

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -376,8 +376,14 @@ public void onFailure(Throwable t) {
                             cluster.manager.triggerOnDown(host, false);
                         } else {
                             logger.warn("Error creating pool to " + host, t);
+                            cluster.manager.triggerOnDown(host, true);
```
olim7t: I don't think this is necessary. It should be the conviction policy's job to mark the node down if there are no more connections to it.
If there are still active connections (for example from another session, or the control connection) then the host should stay up.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
olim7t: There's no way to get a failed future from maybeAddPool, so the two are equivalent.

olim7t (changes requested):

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
adutra: Now there is one, see [here](https://github.com/datastax/java-driver/blob/71c23f8b33dc5bd5cc66f3220ee119f1437b106e/driver-core/src/main/java/com/datastax/driver/core/SessionManager.java#L383-L384).

adutra:
For some reason my comment on `Futures.allAsList` did not show up so here it is again:

> There's no way to get a failed future from maybeAddPool, so the two are equivalent.
>> Now there is one, see [here](https://github.com/datastax/java-driver/blob/71c23f8b33dc5bd5cc66f3220ee119f1437b106e/driver-core/src/main/java/com/datastax/driver/core/SessionManager.java#L383-L384).

tolbertam (approved):
👍

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -431,7 +439,7 @@ CloseFuture removePool(Host host) {
         }
 
         // Wait pool creation before removing, so we don't lose connectivity
-        ListenableFuture<?> allPoolsCreatedFuture = Futures.successfulAsList(poolCreatedFutures);
+        ListenableFuture<?> allPoolsCreatedFuture = Futures.allAsList(poolCreatedFutures);
```
newkek, line 442: I have noticed that `void updateCreatedPools(Host h)` catches `InterruptedException` and `ExecutionException` and [does things](https://github.com/datastax/java-driver/pull/831/files#diff-8ccd18d8cb39074bf83554358d002ceeR476) if those exceptions were to be thrown. And I wonder if now that we use `allAsList()` it is something to be worried about in this method, too if that makes any sense?

GregBestland (approved):
I love the byteman additions. This is extremely useful.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
adutra: There is nothing to worry about it. Only `java.lang.Error` instances can be propagated up here, all the rest is unchanged.

adutra:
@olim7t I think I addressed all your feedback already, can you approve this PR?

olim7t (approved):

PR #832, opened 2017-03-30
JAVA-1438: QueryBuilder check for empty orderings

podnov:
Added orderings.isEmpty check before QueryBuilder adds ORDER BY clause

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/querybuilder/Select.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/querybuilder/Select.java
@@ -154,6 +154,9 @@ public class Select extends BuiltStatement {
         if (this.orderings != null)
             throw new IllegalStateException("An ORDER BY clause has already been provided");
 
+        if (orderings.length == 0)
+            throw new IllegalArgumentException("Invalid ORDER BY argument, the orderings must not be empty.");
+
         this.orderings = Arrays.asList(orderings);
         for (Ordering ordering : orderings)
             checkForBindMarkers(ordering);
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -6,6 +6,7 @@
 - [bug] JAVA-1599: exportAsString improvements (sort, format, clustering order)
 - [improvement] JAVA-1587: Deterministic ordering of columns used in Mapper#saveQuery
 - [improvement] JAVA-1500: Add a metric to report number of in-flight requests.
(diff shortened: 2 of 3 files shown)


tolbertam:
This looks good to me, thanks for your contribution @podnov!   I changed the target branch to 3.2.x as I think this is straightforward enough to fix in a patch release.

tolbertam:
A side note for whoever merges this that we should add a changelog entry for this (I am forgoing doing this at the moment as it would inevitably create a merge conflict, so it makes sense to wait until we merge).

tolbertam (approved):

driver-core/src/main/java/com/datastax/driver/core/querybuilder/Select.java
```diff
@@ -95,7 +95,7 @@ StringBuilder buildQueryString(List<Object> variables, CodecRegistry codecRegist
             Utils.joinAndAppend(builder, codecRegistry, " AND ", where.clauses, variables);
         }
 
-        if (orderings != null) {
+        if (orderings != null && !orderings.isEmpty()) {
```
newkek: I don't think we should silently ignore if the user calls QueryBuilder methods that are not valid to create a CQL query. If the argument in the `orderBy` method is not correct for a CQL query we should throw an exception. As it is done for other query builder methods, like `limit()` if the argument is not valid for Cassandra (limit < 0), we should not ignore it when creating the query

newkek (changes requested):

newkek:
@podnov Thank you very much for the contribution! I mentioned a small change in the review, and we are hoping to get this merged soon, so I'll go ahead and do that change (will keep the commit attributed to you) and if that change does not suit your needs, please re-open the JIRA, thanks!

newkek (approved):

driver-core/src/main/java/com/datastax/driver/core/querybuilder/Select.java
podnov: You've certainly fixed the issue as reported in JIRA, thanks for that.  My only feedback would be that when the person writing code against the querybuilder doesn't know exactly what the end user is going to give them as a query, not accepting an empty list of orderings can make for some ugly builder code:

https://gist.github.com/podnov/b67f7e57cf702df579b68b67dde839c7

I'm sure there are other ways to write this, but the ones immediately apparent to me don't seem terribly straight forward.

PR #833, opened 2017-04-05
Fix compiler warnings in mapping module

adutra:


tolbertam (approved):
Nice!  Took a look at the build output on CI and there are no more deprecation warnings :)

olim7t (approved):

PR #834, opened 2017-04-05
Update license headers

adutra:


tolbertam (approved):
I like that the license header is inline, nice!

PR #837, opened 2017-04-11
JAVA-1407: KeyspaceMetadata exportAsString should export user types in topological sort order

adutra:


driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
```diff
@@ -43,7 +40,7 @@
 
     final Map<String, TableMetadata> tables = new ConcurrentHashMap<String, TableMetadata>();
     final Map<String, MaterializedViewMetadata> views = new ConcurrentHashMap<String, MaterializedViewMetadata>();
-    final Map<String, UserType> userTypes = new ConcurrentHashMap<String, UserType>();
+    final Map<String, UserType> userTypes = Collections.synchronizedMap(new LinkedHashMap<String, UserType>());
```
adutra, line 43: This will certainly come with a performance penalty. I also considered the following alternatives:

1. Maintain a separate `CopyOnWriteArrayList`.
2. Use `ReentrantLock` to access the map.
3. Use Caffeine's `ConcurrentLinkedHashMap`.

All have pros and cons so I went for the simplest solution.

tolbertam: 👍 I like your solution.  The performance penalty is likely to be negligible in the grand scheme of things as typically there are not a lot of user types on a keyspace, and even if there were (thousands, say) this still wouldn't be noticeable I'm assuming.

tolbertam (approved):

driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
olim7t: And metadata is not a "hot" part of the code anyways.

olim7t (approved):
So the fix relies on the fact that updating the types was already done in topological order, right?

adutra:
> So the fix relies on the fact that updating the types was already done in topological order, right?

Yes, correct.

PR #838, opened 2017-04-20
JAVA-1448: TokenAwarePolicy should respect child policy ordering

adutra:


GregBestland (approved):
This all looks good to me. I did a quick sniff test with regards to memory foot print, I'm not seeing any substantiated difference even in the case where no local nodes are found. The test coverage is sufficient.

Thanks for the quick turn around on this.

adutra:
Closing as JAVA-1448 is likely to be redefined in a different direction.

PR #839, opened 2017-04-26
JAVA-1455: Mapper support using unset for null values

fromanator:


fromanator:
Direct link to related issue https://datastax-oss.atlassian.net/browse/JAVA-1455

tolbertam:
Great idea @fromanator .  I wonder if we could possibly unify `saveNullFields` and `saveNullFieldsAsUnset`.  I.E. if `saveNullFields` is false and protocol version >= 4, use `unset`.  For that to work I think we would need to change the API of `Option` so `modifiesQueryString()` accepts the `MappingManager` (sort of like `validate` does) and then use `Statement.unset()` selectively based on protocol version as well.

fromanator:
@tolbertam I think that's a good idea. I originally made the change as a separate option for compatibility. However the more I think of it, it's not really required for compatibility, the end user doesn't really care about how nulls are not saved, just that they are not. So I do think it would make more sense to merge the feature into the saveNullFields option and use unset for >= V4, otherwise fallback to what it currently does.

fromanator:
@tolbertam I updated my PR to merge this functionality into the existing `saveNullFields`, so it will use unset if protocol v4 or greater is used.

It's not possible to use the `modifiesQueryString` method to without some serious rewriting/refactor. The reason being that when generating the prepared statement in `QueryType.makePreparedQueryString` you need the values for the columns so you can omit the columns with null values when generating the statement to prepare to support <ProtocolV4.

driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
```diff
@@ -861,7 +865,7 @@ public void resetDefaultDeleteOptions() {
      */
     public static abstract class Option {
 
-        enum Type {TTL, TIMESTAMP, CL, TRACING, SAVE_NULL_FIELDS, IF_NOT_EXISTS}
+        enum Type {TTL, TIMESTAMP, CL, TRACING, SAVE_NULL_FIELDS, SAVE_NULL_FIELDS_AS_UNSET, IF_NOT_EXISTS}
```
tolbertam: It looks like SAVE_NULL_FIELDS_AS_UNSET is no longer used so this can be removed

manual/object_mapper/using/README.md
```diff
@@ -106,13 +106,13 @@ mapper.save(new User(userId, "helloworld"),
...
 
 <table border="1" style="text-align:center; width:100%;margin-bottom:1em;">
-    <tr> <td><b>Option</b></td>    <td><b>save/saveQuery</b></td> <td><b>get/getQuery</b></td> <td><b>delete/deleteQuery</b></td></tr>
-    <tr> <td>Ttl</td>              <td>yes</td>                   <td>no</td>                  <td>no</td> </tr>
-    <tr> <td>Timestamp</td>        <td>yes</td>                   <td>no</td>                  <td>yes</td> </tr>
-    <tr> <td>ConsistencyLevel</td> <td>yes</td>                   <td>yes</td>                 <td>yes</td> </tr>
-    <tr> <td>Tracing</td>          <td>yes</td>                   <td>yes</td>                 <td>yes</td> </tr>
-    <tr> <td>SaveNullFields</td>   <td>yes</td>                   <td>no</td>                  <td>no</td> </tr>
-    <tr> <td>IfNotExists</td>      <td>yes</td>                   <td>no</td>                  <td>no</td> </tr>
+    <tr> <td><b>Option</b></td>         <td><b>save/saveQuery</b></td> <td><b>get/getQuery</b></td> <td><b>delete/deleteQuery</b></td></tr>
```
tolbertam: As far as I can tell, with SaveNullFieldsAsUnset no longer needed this file is unchanged (other than the spacing), so this file change should be backed out.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperSaveNullFields22Test.java
```diff
@@ -0,0 +1,154 @@
...
+import com.datastax.driver.mapping.Mapper.Option;
+import com.datastax.driver.mapping.annotations.PartitionKey;
+import com.datastax.driver.mapping.annotations.Table;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@SuppressWarnings("unused")
+public class MapperSaveNullFields22Test extends CCMTestsSupport {
```
tolbertam: `@CassandraVersion("2.2.0")` just sets the minimum version requirement of the test, so `MapperSaveNullFieldsTest` will fail when run against >= C* 2.2.  We have an internal ci test matrix that runs against all of the latest cassandra versions (1.2.19, 2.0.17, 2.1.17, 2.2.9, 3.0.13 and 3.10 at the moment).

I think what we could do instead is change `MapperSaveNullFieldsTest#should_save_null_fields` to change the criteria of the validation based on the protocol version in use, using something like this to get at the protocol version:

```java
        boolean unsetSupported = session().getCluster().getConfiguration().getProtocolOptions()
                .getProtocolVersion().compareTo(ProtocolVersion.V4) >= 0;
```

and then we could do something like this in `should_save_null_fields`:

```java
        if (nullName && !saveExpected) {
            if (unsetSupported) { // column should be present if unset supported and value should be unset
                assertThat(queryString).as(description).contains("name");
                assertThat(!bs.isSet("name"));
            } else {
                assertThat(queryString).as(description).doesNotContain("name");
            }
        } else {
            assertThat(queryString).as(description).contains("name");
        }

        if (nullPhone && !saveExpected) {
            if (unsetSupported) {
                assertThat(queryString).as(description).contains("phone");
                assertThat(!bs.isSet("phone"));
            } else {
                assertThat(queryString).as(description).doesNotContain("phone");
            }
        } else {
            assertThat(queryString).as(description).contains("phone");
        }
```

I concede that is a bit complex, but it'll make it so we can use the existing tests to validate both cases.

tolbertam (changes requested):
Nice work!  This seems much more straightforward and I like that if you were previously using this option, you get the benefits of unset without having to make an explicit change.  I had a few comments/suggestions, but other than that this is looking great 👍

driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
```diff
@@ -216,25 +217,26 @@ public Statement saveQuery(T entity, Option... options) {
     }
 
     private ListenableFuture<BoundStatement> saveQueryAsync(T entity, final EnumMap<Option.Type, Option> options) {
-        final Map<AliasedMappedProperty, Object> values = new HashMap<AliasedMappedProperty, Object>();
-        boolean saveNullFields = shouldSaveNullFields(options);
+        final Map<AliasedMappedProperty, Object> columnToValue = new HashMap<AliasedMappedProperty, Object>();
+        final boolean useUnsetForNullValue = !shouldSaveNullFields(options) && manager.protocolVersionAsInt >= 4;
+        final boolean includeColumnsWithNullValue = shouldSaveNullFields(options) || useUnsetForNullValue;
```
tolbertam: > It's not possible to use the modifiesQueryString method to without some serious rewriting/refactor. The reason being that when generating the prepared statement in QueryType.makePreparedQueryString you need the values for the columns so you can omit the columns with null values when generating the statement to prepare to support <ProtocolV4.

It seems like `modifiesQueryString` wasn't relevant anyways.  This is where the magic is happening, very clever!     👍

fromanator:
Good callouts, I wasn't as careful as I thought about removing the old stuff. (Although I did fix the docs stuff, but did a `commit --amend` with a force push to my remote which didn't show up in the PR, hopefully I got that sorted out now).

Good info on how the testing framework is orchestrated, I agree with you that it's better to just use the same test class, it's not too messy to add in the protocol version logic switch.

fromanator:
I didn't see anything specific in the CONTRIBUTING guide about if you prefer squashed commit PRs (or perhaps it's squashed on merge automatically). Anyway just let me know if I need to clean up the commit history before it can be merged.

tolbertam:
> I didn't see anything specific in the CONTRIBUTING guide about if you prefer squashed commit PRs (or perhaps it's squashed on merge automatically). Anyway just let me know if I need to clean up the commit history before it can be merged.

I think it is good the way it is now, we can squash it when we go to merge it.  I'll work with the team to see if we can target it for 3.2.1. 👍

tolbertam (approved):

adutra (approved):

adutra:
Good job @fromanator ! I propose however that we schedule this for the next minor, 3.3.0, as it is a new feature and introduces behavioral changes.

adutra:
Thank you @fromanator for your contribution. I tried to squash the commits and force-push to your own branch, but GitHub didn't like it. Anyway your work has been merged into our codebase as 19e10176d. Thanks!

alezx:
Hi,

sorry in release 3.6 I see Mapper.java at line 260
`    final boolean useUnsetForNullValue =
        !shouldSaveNullFields(options) && manager.protocolVersionAsInt >= 4;`

I don't want to set null values so I am using the mapper option saveNullFields=false.
but then useUnsetForNullValue  is evaluated to true. 
I think is a bug because it will set the values to be null in the statement.
Can you have a look please ?

thank you

fromanator:
The other part to this feature is that you must be using at least protocolV4 to use this feature (Cassandra 2.2.x or 3.x). So I would double check your driver config that you're either explicitly setting that version, or if you're not ensure that all the nodes in your cluster support protocolV4.

alezx:
hi @fromanator  thanks for your reply 
What I am missing is looking at the lines 260-261
`    

    final boolean useUnsetForNullValue =
        !shouldSaveNullFields(options) && manager.protocolVersionAsInt >= 4;`

    final boolean includeColumnsWithNullValue =
        shouldSaveNullFields(options) || useUnsetForNullValue;`

includeColumnsWithNullValue  will always be true for protocol>=4, no matter the value of saveNullFields.

I am testing with cassandra unit latest version 3.5.0,1, you think is that ? If the statement produced is setting the values to null, how will the driver figure out that should ignore that values ? I hope you can point me in the right direction, thank you

Ale

PR #841, opened 2017-05-16
Tuple docs.

newkek:


manual/tuples/README.md
```diff
@@ -1,5 +1,77 @@
...
+  ck1 text,
+  ck2 text,
+  v tuple<int, text, float>,
+  PRIMARY KEY (pk, ck1, ck2)
+);
+```
+
+### Fetching Tuples from Rows results
+
+The DataStax Java driver exposes a special `TupleValue` class to handle such columns. `TupleValue` extends `GettableByIndexData` class, which allows to call specific `get...(int)` methods on a returned `TupleValue`:
```
tolbertam: Would be good to link to `TupleValue` & `GettableByIndexData` javadoc on the first reference to them.

tolbertam (approved):
Looks good.  Had a small suggestion for javadoc links.

manual/tuples/README.md
```diff
@@ -1,5 +1,77 @@
-## Tuples
+## Using Tuples with the Java driver
 
-*Coming soon... In the meantime, see the javadoc for [TupleType].*
+Cassandra allows to use `tuple` data types [in tables](https://docs.datastax.com/en/cql/3.1/cql/cql_reference/tupleType.html):
```
adutra: They are also allowed in user types.

manual/tuples/README.md
```diff
@@ -1,5 +1,77 @@
...
+int firstValueInTuple = tupleValue.getInt(0);
+
+String secondValueInTuple = tupleValue.getString(1);
+
+Float thirdValueInTuple = tupleValue.getFloat(2);
+```
+
+### Using tuples as statement parameters
+
+A prepared statement may be containing a Tuple as a query parameter. In such case, users will need to create or gather a `TupleType` first, in order to be able to create a `TupleValue` to bind:
```
adutra: Nit: "may contain" ... "In such cases,..."

manual/tuples/README.md
```diff
@@ -1,5 +1,77 @@
...
+bs.setString("ck1", "1");
+bs.setString("ck2", "1");
+bs.setTupleValue("v", tupleType.newValue(1, "hello", 2.3f));
+
+session.execute(bs);
+```
+
+#### More use cases
+
+Users can also define single-usage tuples in _SELECT_ queries with the `IN` keyword, usually for tables with composite clustering keys, in that case a tuple will be usable the same way it was for prepared statements parameters:
```
adutra: This is called a "multi-column IN restriction" – maybe it's worth mentioning this term somewhere.

adutra (changes requested):

adutra (approved):

manual/tuples/README.md
```diff
@@ -1,5 +1,87 @@
...
+  ck2 text,
+  v tuple<int, text, float>,
+  PRIMARY KEY (pk, ck1, ck2)
+);
+```
+
+### Fetching Tuples from Rows results
+
+The DataStax Java driver exposes a special [TupleValue] class to handle such columns. 
+[TupleValue] extends [GettableByIndexData] class, which allows to call specific `get...(int)`
```
olim7t: nit: GettableByIndexData is an implementation detail, not sure if it's relevant to mention it here. All the user needs to know is that there are getter methods.

manual/tuples/README.md
```diff
@@ -1,5 +1,87 @@
...
+```java
+PreparedStatement ps = session.prepare("INSERT INTO ks.collect_things (pk, ck1, ck2, v) VALUES (:pk, :ck1, :ck2, :v)");
+
+TupleType tupleType = cluster.getMetadata().newTupleType(DataType.cint(), DataType.text(), DataType.cfloat());
+
+BoundStatement bs = ps.bind();
+bs.setInt("pk", 1);
+bs.setString("ck1", "1");
+bs.setString("ck2", "1");
+bs.setTupleValue("v", tupleType.newValue(1, "hello", 2.3f));
```
olim7t, line 48: Would be useful to mention that `newValue(Object...)` follows the same rules as `new SimpleStatement(String, Object...)`, i.e. type inference with some ambiguities (like numeric literals always interpreted as `int`).
Another way to create a tuple value -- and the workaround to solve those ambiguities -- is the parameterless `newValue()` followed by setter calls.

olim7t (changes requested):

olim7t (approved):

olim7t:
lgtm, but maybe include a code example of newValue() + setters as well?
(even though it's obvious)

PR #842, opened 2017-05-16
JAVA-1460: Add speculative execution number to ExecutionInfo

olim7t:


driver-core/src/main/java/com/datastax/driver/core/ExecutionInfo.java
```diff
@@ -84,9 +85,9 @@ ExecutionInfo withIncomingPayload(Map<String, ByteBuffer> incomingPayload) {
      * <li>if a host is tried by the driver but is dead or in
      * error, that host is recorded and the query is retried;</li>
      * <li>on a timeout or unavailable exception, some
-     * {@link com.datastax.driver.core.policies.RetryPolicy} may retry the
```
adutra: Nit: we usually do not add an import for a class that is only referenced through javadocs, to avoid creating a compile-time dependency between classes that are otherwise unrelated.

driver-core/src/main/java/com/datastax/driver/core/ExecutionInfo.java
```diff
@@ -112,8 +113,24 @@ public Host getQueriedHost() {
...
 
     /**
+     * The speculative execution that completed this query.
+     * <p>
+     * 1 represents the initial, regular execution of the query, 2 represents the first speculative
+     * execution, etc.
+     * <p>
+     * Note that this is different from the number of <em>started</em> executions. For example, if
+     * one speculative execution was triggered, but the initial execution eventually completed
+     * first, this will be 1.
```
adutra: Wouldn't users wish they could have a more complete vision? I.e know exactly how many spec execs were triggered, and which one completed and which ones were canceled)?

adutra (changes requested):

driver-core/src/main/java/com/datastax/driver/core/ExecutionInfo.java
olim7t: It's more complicated... The dilemma with ExecutionInfo is that it's updated concurrently, so it boils down to how much info we want to provide vs the overhead of tracking this info.

driver-core/src/main/java/com/datastax/driver/core/ExecutionInfo.java
GregBestland: I've removed this.

driver-core/src/main/java/com/datastax/driver/core/ExecutionInfo.java
GregBestland: I agree it would be nice to add the number some context onto at least which hosts were queried with speculative. I also understand why you avoided making this change as it's not trivial.  Mostly I think it's confusing to the user.

I think we need to be more explicit about the information we are returning on the ExecutionInfo. Without knowing how we handle speculative execution, I can see users being confused as to why speculative executions executed on hosts aren't listed in the triedhosts list for instance. Perhaps we could provide seperate specific speculative execution related information with regards to things like tried hosts.

olim7t: Tracking the total number of executions sounds reasonable and is not too hard to do, adding it.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -111,7 +111,7 @@ private void startNewExecution() {
             return;
 
         Message.Request request = callback.request();
-        int position = executionCount.incrementAndGet();
+        int position = executionIndex.getAndIncrement();
```
olim7t, line 114: Changed this because it's more consistent with the way things are exposed in ExecutionInfo.
The initial execution will now have a position of 0 instead of 1. This was only exposed in debug logs so I don't think it's an issue.

driver-core/src/main/java/com/datastax/driver/core/ExecutionInfo.java
```diff
@@ -112,6 +116,33 @@ public Host getQueriedHost() {
     }
```
adutra, line 116: The javadocs for `getQueriedHost()` should be amended: `... This is a shortcut for {@code getTriedHosts().get(getTriedHosts().size() - 1)}.`

driver-core/src/main/java/com/datastax/driver/core/ExecutionInfo.java
```diff
@@ -87,7 +88,10 @@ ExecutionInfo withIncomingPayload(Map<String, ByteBuffer> incomingPayload) {
      * {@link com.datastax.driver.core.policies.RetryPolicy} may retry the
      * query on the same host, so the same host might appear twice.</li>
      * <li>if {@link com.datastax.driver.core.policies.SpeculativeExecutionPolicy speculative executions}
-     * are enabled, other hosts might have been tried speculatively as well.</li>
+     * are enabled, this will also contain hosts that were tried by other executions (however, note that
+     * this only contains hosts for which a response was received; if an execution is waiting for a response
```
adutra: This is misleading: there is only ever at most one response received (either a result or an error); other tried hosts, if any, can only be listed here if they timed out with no response at all. I suggest: "this only contains the queried host – i.e. the host who replied with the final response – and hosts that failed to reply before the request timed out; if an execution is waiting..."

driver-core/src/main/java/com/datastax/driver/core/ExecutionInfo.java
```diff
@@ -87,7 +88,10 @@ ExecutionInfo withIncomingPayload(Map<String, ByteBuffer> incomingPayload) {
      * {@link com.datastax.driver.core.policies.RetryPolicy} may retry the
      * query on the same host, so the same host might appear twice.</li>
```
adutra, line 89: Just as a reminder, we might want to change the return type in version 4.0 to `Set<Host>`, because it doesn't really make sense to return a `List` since we introduced retries on the same host. Or better yet, some sort of `Map<SpeculativeExecution, Host>`. And even better yet, a custom structure `RequestTimeline` that would record all the relevant events that lead from the request submission to request completion (e.g. at instant T0 request was submitted, at instant T1 spec exec 1 was fired, at instant T2 normal exec timed out, etc...) – in other words, some sort of client-side query trace.

adutra (changes requested):

driver-core/src/main/java/com/datastax/driver/core/ExecutionInfo.java
adutra: Besides, the queried host, according to `getQueriedHost()`, is always the last one. We could stress this fact as well.

driver-core/src/main/java/com/datastax/driver/core/ExecutionInfo.java
newkek: > RequestTimeline

That would be **great** it's kind of what I was imagining when opening the ticket

newkek: However we'll need to consider the cost of this

olim7t: Yes, consider the cost and compare it to the value for users. The question is, what percentage of users are interested in a detailed timeline of executions and retries, and how would they use it in their application?
For debugging => maybe trace logs are good enough
To compute metrics => we can expose those metrics ourselves, in a less invasive way (make them optional)

driver-core/src/main/java/com/datastax/driver/core/ExecutionInfo.java
olim7t: There can also be hosts that replied with an error response for which the retry policy decided to retry.

newkek:
I understand now that `triedHosts` will contain:
- the host that completed the query successfully (and it may be the host of a speculative retry if the speculative retry completed before an initial request)
- the hosts that may have produced an error (timeout, server error) and triggered a retry

And it will **not** contain the hosts that have been tried for Speculative executions attempts *that have not completed*. I see 2 downsides to that:
1. the name `triedHosts` is clearly misleading
2. there is no way for users to know which hosts have been tried in speculative executions for a specific request/result set.

So adding the hosts from the "incomplete" speculative executions would solve both of these problems and imo would be nice to address here, however we end up with the problem described in JAVA-1459 and need to solve it too.

GregBestland:
We previously discussed the issues kevin brought up. One of the options that was brought up was changing tried hosts to something more explicit like "failedhosts". I understand that doing so now would be  a breaking change, which we want to avoid. I'm okay with this going in as is, with the idea being we will fix it when such breaking api changes can be incorporated.

newkek:
Although we could fix things now and without introducing breaking changes by including the pending speculative executions in the list of tried hosts. Sorry if those points were brought up already and I missed the conversation.

newkek:
So any feedback on adding the pending speculative executions's hosts to the `triedHosts`? Imo with a switch to turn it on and off as @tolbertam suggested would be wise too?

GregBestland (approved):

adutra (approved):

PR #844, opened 2017-05-16
Doc updates (JAVA-1463 & JAVA-1466)

olim7t:


manual/query_timestamps/README.md
```diff
@@ -116,6 +83,43 @@ When this happens, the built-in generators log a periodic warning message in the
...
+
+This can be a problem when the order of the writes matter: with unlucky
+timing (different coordinators, network latency, etc.), two successive
+requests from the same client might be processed in a different order
+server-side, and end up with out-of-order timestamps. This is why client-side
+generation is the default starting with protocol v3. With older versions, the
+only way to get client timestamps is to add a `USING TIMESTAMP` clause to your
+CQL queries.
+
+Note that, although client-side timestamps are recommended when available, you can still opt for
```
tolbertam, line 113: It could be good to indicate when Server-side generation might be more viable than client side, i.e. if you have many more clients than C* nodes, or where it is more difficult to ensure time is sync'd across clients vs. managing time on the C* nodes.

manual/speculative_execution/README.md
```diff
@@ -66,15 +66,15 @@ there is no way to guarantee that only one node will apply the mutation.
 ### Enabling speculative executions
 
 Speculative executions are controlled by an instance of
-[SpeculativeExecutionPolicy][sep] provided when initializing the
+[SpeculativeExecutionPolicy] provided when initializing the
```
tolbertam, line 72: Not entirely related to the changes here, but wanted to point out we use the term 'cancel' a few times, i.e.:

```
If that second node replies faster, we can send the response back to the client (we also cancel the first query)
```

We should clarify that by cancel we mean that we disregard/throw away any subsequent response received.  In the past people have perceived `ResultSetFuture.cancel()` as possibly cancelling the request, which it's really just discarding any response right?

manual/speculative_execution/README.md
```diff
@@ -121,14 +118,22 @@ explicitly depend on it:
...
 ```
 
-Then create an instance of [PerHostPercentileTracker][phpt] that will collect
-latency statistics for your `Cluster`:
+Then create a [PercentileTracker] that will collect latency histograms for your `Cluster`. Two
+implementations are provided out of the box:
+
+* [PerHostPercentileTracker]: maintains a histogram per host. This means queries to a host will
+  only be compared against previous queries to the same host.
+* [ClusterWidePercentileTracker]: maintains a single histogram for the whole cluster. This means
```
tolbertam: small nit: Since we recommend ClusterWidePercentileTracker we should list it first.

manual/speculative_execution/README.md
```diff
@@ -148,20 +153,15 @@ Cluster cluster = Cluster.builder()
...
 
-Finally, don't forget to register your tracker with the cluster (the
-policy does not do this itself):
-
-```java
-cluster.register(tracker);
-```
-
-Note that `PerHostPercentileTracker` may also be used with a slow query
+Note that `PercentileTracker` may also be used with a slow query
```
tolbertam, line 159: In the QueryLogger docs we reference `PerHostPercentileTracker` explicitly, should we change that to `PercentileTracker` too?

tolbertam (approved):
Looks great 👍 !

manual/speculative_execution/README.md
```diff
@@ -7,7 +7,10 @@ will experience bad latency.
 One thing we can do to improve that is pre-emptively start a second
 execution of the query against another node, before the first node has
 replied or errored out. If that second node replies faster, we can send
-the response back to the client (we also cancel the first query):
+the response back to the client (we also cancel the first execution --
+note that "cancelling" in this context simply means discarding the response
+when it arrives later, Cassandra does not support cancellation of in flight
```
adutra, line 12: Nit: "in-flight" takes a hyphen.

adutra (approved):

PR #845, opened 2017-05-16
JAVA-1469: Update LoggingRetryPolicy to deal with SLF4J-353

tolbertam:
For [JAVA-1469](https://datastax-oss.atlassian.net/browse/JAVA-1469)

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -2,6 +2,8 @@
 
 ### 3.3.0 (in progress)
 
+- [bug] JAVA-1469: Update LoggingRetryPolicy to deal with SLF4J-353.
+
 
 ### 3.2.0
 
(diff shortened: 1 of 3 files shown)


adutra (approved):

tolbertam:
Just had a thought, one noticeable change here is now the full stack trace is logged when an exception is encountered.  In the general case the stack trace probably isn't too useful (as the error will be a timeout exception and so on) especially as I'm guessing non-nominal exceptions will be logged with error or something else anyways, maybe I should change it back so it just logs the exception message, wdyt?  I'll add a commit for that and we can back it out if we want it to include the full stack trace.

tolbertam:
Changed it to simply pass `e.toString()` as a parameter to `log.info` this is essentially how SLF4J handled this before.  The output with this should now be the same as if you were using SLF4J prior to 1.7.15.

adutra (approved):

olim7t (approved):

PR #847, opened 2017-05-17
JAVA-1428: Upgrade logback and jackson dependencies

tolbertam:
While not required dependencies, both logback and jackson have newer
versions that fix bugs and security vulnerabilities.


adutra:
Added a couple of commits to fix small inconsistencies regarding SLF4J + minor fixes related to dependency management.

olim7t (approved):
👍 
Nit: in the last commit message, men->maven? ;-)

tolbertam:
good catches @adutra 👍

PR #848, opened 2017-05-29
JAVA-1443: Add groupBy method to Select statement

adutra:
Note to reviewers: I seized the opportunity to also

1. Enhance the docs and fix some typos;
2. Add `toJson` support to `select(Object...)` (probably an oversight, only `select()` was supported);
3. Add support for `ALLOW FILTERING` when a `WHERE` clause is present (probably an oversight);
4. Add support for built-in function calls `sum`, `avg`, `count`, `min`, `max`.


GregBestland (approved):
This looks really good. I tried very hard to find something you missed and could not. Great changes.

PR #849, opened 2017-05-29
JAVA-1446: Support 'DEFAULT UNSET' in Query Builder JSON Insert

adutra:


driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
```diff
@@ -76,8 +75,6 @@
  */
 public abstract class BuiltStatement extends RegularStatement {
 
-    private static final Pattern lowercaseAlphanumeric = Pattern.compile("[a-z][a-z0-9_]*");
```
adutra, line 79: Unused.

GregBestland (approved):
I have nothing to add here. LGTM.

PR #850, opened 2017-05-29
JAVA-1458: Check thread in mapper sync methods

adutra:


GregBestland (approved):
I added some tests for this. Probably went a little overboard, these changes look good.

adutra:
Excellent test! I just fixed a small issue where the test isn't really being repeated when the future is executed on the caller thread.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperTest.java
```diff
@@ -416,7 +417,7 @@ public Thread apply(Void v) {
         executeFunctionAndTestForException(u, mapper, new Function<Void, Thread>() {
             @Override
             public Thread apply(Void v) {
-                mapper.save(u, null);
```
GregBestland: The idea was to test every entry point where the thread check was added. In this case that would be save method with and without options. I might be missing something but this would be the exact same as the above, and would miss save method where options are passed in wouldn't it?

If so applies to the methods below as well.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperTest.java
```diff
@@ -476,30 +477,25 @@ public Thread apply(Void v) {
...
-        assertThat(isFailed(toTest, IllegalStateException.class, "Detected a synchronous call on an I/O thread")).isTrue();
-    }
-
-    private boolean isFailed(ListenableFuture<Thread> future, Class<?> expectedException, String expectedMessage) {
         try {
-            Thread executedThread = future.get();
-            if (executedThread != Thread.currentThread()) {
-                fail("Expected a failed future, callback was executed on " + executedThread);
-            } else {
+            Thread executedThread = toTest.get();
```
GregBestland: Thanks I missed that. Good catch!

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperTest.java
adutra: You are absolutely right, I'm sorry, I missed that detail. 
But in this case, let's use something else than `null` because `null` will eventually throw a `NullPointerException` – for example, if `checkNotInEventLoop()` succeeds because the callback is being executed on the calling thread – and make the test fail.

PR #851, opened 2017-05-30
JAVA-1266 & JAVA-1411: Stop using oss-parent and normalize versions across the build

adutra:
Summary of changes:

[JAVA-1266](https://datastax-oss.atlassian.net/browse/JAVA-1266):
- Removed reference to deprecated OSS Sonatype parent poms
- Adjusted release process to use the recommended method,
  as described in http://central.sonatype.org/pages/apache-maven.html
- Created new 'release' profile to contain release-specific plugins

[JAVA-1411](https://datastax-oss.atlassian.net/browse/JAVA-1411):
- Introduced dependencyManagement and pluginManagement sections in the parent pom
- Declared versions for all dependencies and plugins at parent pom
- Concentrated common plugin configuration items into the parent pom
- Revisited bundle and shade plugin configurations
- Updated &lt;url> tags
- Removed unused 'env' system property
- Profile declaration cleanup

PR #853, opened 2017-06-01
Fix failing tests due to table name clashes

adutra:


tolbertam:
I see, so this explains the test failures on #851?  That makes sense, the tests did seem to execute in different order when running the entire suite on my laptop so maybe that's why I didn't encounter it.

tolbertam (approved):

tolbertam:
Going to go ahead and merge this since I don't see anything controversial about this PR 👍

PR #854, opened 2017-06-01
Fix intermittent ClusterInitTest failure

tolbertam:
Nodes were being excluded from LoadBalancingPolicy because their
Cassandra version was too old.  Setting release_version to 3.0.13
should prevent this for time being.  If this works will make this an
explicit variable to future proof the test.


tolbertam:
Merging as I am relatively confident and this is just a test fix.

PR #855, opened 2017-06-13
Fix failing test after JAVA-1431

adutra:


driver-core/src/test/java/com/datastax/driver/core/SessionErrorTest.java
```diff
@@ -84,14 +84,13 @@ public void should_not_propagate_unchecked_exceptions() {
             Session session = cluster.connect();
             // Pool to host1 should be still open because host1 is the control host,
             // but its pool should have no active connection
-            // Pool to host2 should have been closed because host2 has no
-            // more active connections
+            // Pool to host2 may be either closed because host2 was marked down before initialization
+            // was complete, or open and empty, if host2 was marked down after.
             Session.State state = session.getState();
```
tolbertam: I think this is good, but alternatively it could be good to just wait for host2 to go down before checking the session state (`TestUtils.waitForDown("2", cluster);`).  This way we ensure that a side effect of the exception being raised is that the host is marked down and the pool is ultimately closed (the same assertions we had before can be maintained).

adutra: I tried exactly that, but it doesn't work, I'm still getting two connected hosts instead of one.

adutra: Oh, nvm, I was being stupid. I need to get the session state _after_ the node is marked down. That is why I was still getting two connected hosts.

adutra: Indeed that works provided that the state is obtained after the host is marked down. Thanks!

tolbertam (approved):
Looks great 👍

PR #856, opened 2017-06-13
Instant speculative retries for JAVA-1490

clohfink:
Allow 0ms for the constant speculative execution policy and make it schedule it immediately.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
@@ -119,29 +119,39 @@ class RequestHandler {
     }
 
     private void scheduleExecution(long delayMillis) {
-        if (isDone.get() || delayMillis <= 0)
+        if (isDone.get() || delayMillis < 0)
             return;
         if (logger.isTraceEnabled())
             logger.trace("[{}] Schedule next speculative execution in {} ms", id, delayMillis);
-        scheduledExecutions.add(scheduler.newTimeout(newExecutionTask, delayMillis, TimeUnit.MILLISECONDS));
+        if(delayMillis == 0) {
+            // kick off request immediately
+            scheduleExecutionImmediately();
+        } else {
+            scheduledExecutions.add(scheduler.newTimeout(newExecutionTask, delayMillis, TimeUnit.MILLISECONDS));
+        }
     }
 
     private final TimerTask newExecutionTask = new TimerTask() {
         @Override
         public void run(final Timeout timeout) throws Exception {
             scheduledExecutions.remove(timeout);
-            if (!isDone.get())
+            if (!isDone.get()) {
                 // We're on the timer thread so reschedule to another executor
                 manager.executor().execute(new Runnable() {
                     @Override
                     public void run() {
-                        if (metricsEnabled())
-                            metrics().getErrorMetrics().getSpeculativeExecutions().inc();
-                        startNewExecution();
+                        scheduleExecutionImmediately();
                     }
                 });
+            }
         }
     };
+
+    private void scheduleExecutionImmediately() {
+        if (metricsEnabled())
+            metrics().getErrorMetrics().getSpeculativeExecutions().inc();
+        startNewExecution();
+    }
 
     private void cancelPendingExecutions(SpeculativeExecution ignore) {
         for (SpeculativeExecution execution : runningExecutions)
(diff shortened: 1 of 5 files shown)


tolbertam (approved):
looks good to me, thanks!  👍

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -119,17 +119,28 @@ private void startNewExecution() {
...
             return;
         if (logger.isTraceEnabled())
             logger.trace("[{}] Schedule next speculative execution in {} ms", id, delayMillis);
-        scheduledExecutions.add(scheduler.newTimeout(newExecutionTask, delayMillis, TimeUnit.MILLISECONDS));
+        if(delayMillis == 0) {
+            // kick off request immediately
+            try {
+                newExecutionTask.run(null);
+            } catch (Exception e) {
+                // the newExecutionTask shouldnt be able to throw a non-runtime but pokecatch needed for netty api
```
adutra: Was wondering if we could improve this. Maybe if you extract some code out of the timeout task body:

```
private final TimerTask newExecutionTask = new TimerTask() {
    @Override
    public void run(final Timeout timeout) {
        scheduledExecutions.remove(timeout);
        scheduleExecutionImmediately();
    }
};

private void scheduleExecution(long delayMillis) {
    if (isDone.get() || delayMillis < 0)
        return;
    if (logger.isTraceEnabled())
        logger.trace("[{}] Schedule next speculative execution in {} ms", id, delayMillis);
    if (delayMillis == 0) {
        // kick off request immediately
        scheduleExecutionImmediately();
    } else {
        scheduledExecutions.add(scheduler.newTimeout(newExecutionTask, delayMillis, TimeUnit.MILLISECONDS));
    }
}

private void scheduleExecutionImmediately() {
    if (!isDone.get())
        // We could be on the timer thread, so reschedule to another executor
        manager.executor().execute(new Runnable() {
            @Override
            public void run() {
                if (metricsEnabled())
                    metrics().getErrorMetrics().getSpeculativeExecutions().inc();
                startNewExecution();
            }
        });
}

```
Besides I wonder if we could avoid using `manager.executor()` to execute the speculative attempts when the delay is zero, given that we are not on an internal thread. Although sending the speculative query on the same thread as the initial one would have a funny side-effect: the speculative request is likely to be sent before the first (non-speculative) request. Not sure if this is of a concern.

driver-core/src/main/java/com/datastax/driver/core/policies/SpeculativeExecutionPolicy.java
```diff
@@ -64,8 +64,8 @@
          * Returns the time before the next speculative query.
          *
          * @param lastQueried the host that was just queried.
-         * @return the time (in milliseconds) before a speculative query is sent to the next host. If zero or negative,
-         * no speculative query will be sent.
+         * @return the time (in milliseconds) before a speculative query is sent to the next host. If negative,
+         * no speculative query will be sent. If zero it will immediately send the retry.
```
adutra: Nit: let's use the term "speculative (execution|query|request)" and avoid the term "retry" so that users do not confuse this with retry policies :)

driver-core/src/test/java/com/datastax/driver/core/SpeculativeExecutionTest.java
```diff
@@ -168,6 +168,55 @@ public void should_start_speculative_execution_if_first_execution_takes_too_long
...
+                .build();
+        Session session = cluster.connect();
+        host1 = TestUtils.findHost(cluster, 1);
+        host2 = TestUtils.findHost(cluster, 2);
+        host3 = TestUtils.findHost(cluster, 3);
+        errors = cluster.getMetrics().getErrorMetrics();
+
+        scassandras.node(1).primingClient().prime(PrimingRequest.queryBuilder()
+                .withQuery("mock query")
+                .withThen(then().withRows(row("result", "result1")).withFixedDelay(5L))
```
adutra: I suggest raising your delays to a few seconds to make sure the test will pass even in our CI environment (crazy scheduling behaviors happen there :) ).

adutra (approved):
Very good 👍 Just had a few comments, none of them is blocking.

clohfink:
@adutra made changes mentioned, I did end up removing the manager.executor on the immediate execution like you mentioned. Even pushing to executor the execution to the others will be more or less at the "same time" depending on the connection callback, other queries on connections, what the netty threads are doing, and a fair dice roll.

adutra:
@clohfink thanks looks very good 👍

olim7t (approved):
👍 lgtm

PR #858, opened 2017-06-15
JAVA-1500: Add a metric to report number of in-flight requests

mfiguiere:
Exposes a new Gauge backed by the sum of all totalInFlight counts of all HostConnectionPools.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Metrics.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Metrics.java
@@ -79,6 +79,16 @@ public class Metrics {
             return value;
         }
     });
+    private final Gauge<Integer> inFlightRequests = registry.register("inflight-requests", new Gauge<Integer>() {
+        @Override
+        public Integer getValue() {
+            int value = 0;
+            for (SessionManager session : manager.sessions)
+                for (HostConnectionPool pool : session.pools.values())
+                    value += pool.totalInFlight.get();
+            return value;
+        }
+    });
 
     private final Gauge<Integer> executorQueueDepth;
     private final Gauge<Integer> blockingExecutorQueueDepth;
@@ -216,6 +226,15 @@ public class Metrics {
      */
     public Gauge<Integer> getTrashedConnections() {
         return trashedConnections;
+    }
+
+    /**
+     * Returns the total number of in flight requests to Cassandra hosts.
+     *
+     * @return The total number of in flight requests to Cassandra hosts.
+     */
+    public Gauge<Integer> getInFlightRequests() {
+        return inFlightRequests;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


driver-core/src/main/java/com/datastax/driver/core/Metrics.java
```diff
@@ -219,6 +229,15 @@ public Errors getErrorMetrics() {
     }
 
     /**
+     * Returns the total number of in flight requests to Cassandra hosts.
+     *
+     * @return The total number of in flight requests to Cassandra hosts.
+     */
+    public Gauge<Integer> getInFlightRequests() {
```
adutra, line 236: Isn't this a bit too coarse-grained? I wonder if it would be better to return a `MetricSet` instead, and return the number of inflight requests on a per-host basis. This would allow users to detect slow hosts (and besides, this is exactly what has been suggested recently as the best available metric to measure host performance – much better than request latencies for instance).

adutra (changes requested):
Very interesting addition. Was just wondering if it would be better to report in-flight requests on a per-host basis (but I might be overthinking this a bit).

driver-core/src/main/java/com/datastax/driver/core/Metrics.java
mfiguiere: It's a good suggestion that several folks actually prefer as well on our side here! Let me update the PR accordingly.

olim7t: +1, from my experience most metrics would be more useful if they were exposed per host, I'll keep this in mind for the next major revision.

driver-core/src/main/java/com/datastax/driver/core/Metrics.java
```diff
@@ -79,6 +79,16 @@ public Integer getValue() {
...
+    private final Gauge<Integer> inFlightRequests = registry.register("inflight-requests", new Gauge<Integer>() {
+        @Override
+        public Integer getValue() {
+            int value = 0;
+            for (SessionManager session : manager.sessions)
+                for (HostConnectionPool pool : session.pools.values())
+                    value += pool.totalInFlight.get();
+            return value;
+        }
+    });
```
olim7t, line 91: lgtm.
My only remark is that the sum is not done atomically, but there would be no reasonable way to do it, and this should provide a good enough approximation.

newkek (approved):
Lgtm. We would like to merge this soon, the only question left is whether to return the metrics per hosts or not. We agree that this should be done not only for this metric but others already existing, so applying it only to this one now would be inconsistent imo, I'd rather make the change of per-host metrics on all the metrics together for the next major driver version and merge this as is. Currently working on adding a simple test to this to assure the behaviour and it's good to merge for me. wdyt @mfiguiere @olim7t @adutra ?

newkek:
Pushed some very simple tests on the branch https://github.com/datastax/java-driver/tree/java1500-tests

newkek:
Would like to merge this soon. Had confirmation from @adutra off-band that merging in this state is ok, I created a follow-up ticket to expose more of the driver's metrics on a per-host basis: https://datastax-oss.atlassian.net/browse/JAVA-1627

PR #861, opened 2017-07-19
Adding zero copy string construction for JDK8.

TheIndifferent:
Extracted JDK-version specific code to separate classes, plus fallback to normal copy constructor.
This stuff is surprisingly hard to test reliably, so any suggestions on how to improve tests are welcome.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/utils/Bytes.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/utils/Bytes.java
@@ -48,36 +48,18 @@ public final class Bytes {
     }
 
     /*
-     * We use reflexion to get access to a String protected constructor 
+     * We use reflexion to get access to a String protected constructor
      * (if available) so we can build avoid copy when creating hex strings.
      * That's stolen from Cassandra's code.
      */
-    private static final Constructor<String> stringConstructor;
+    private static final StringConstructor stringConstructor = createStringConstructor();
 
-    static {
-        Constructor<String> c;
-        try {
-            c = String.class.getDeclaredConstructor(int.class, int.class, char[].class);
-            c.setAccessible(true);
-        } catch (Exception e) {
-            c = null;
-        }
-        stringConstructor = c;
-    }
 
     private static String wrapCharArray(char[] c) {
         if (c == null)
             return null;
 
-        String s = null;
-        if (stringConstructor != null) {
-            try {
-                s = stringConstructor.newInstance(0, c.length, c);
-            } catch (Exception e) {
-                // Swallowing as we'll just use a copying constructor
-            }
-        }
-        return s == null ? new String(c) : s;
+        return stringConstructor.construct(c);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


TheIndifferent:
Could someone please re-trigger AppVeyor build? Seems like unrelated failure.

tolbertam:
I think i understand the intent of this, but wanted to clarify.  Is the reasoning for this because there is no `String(int, int, char[])` constructor in Java 8 so it silently uses the `new String` implementation?

EDIT: That seems to be the case:

```
java.lang.NoSuchMethodException: java.lang.String.<init>(int, int, [C)
	at java.lang.Class.getConstructor0(Class.java:3082)
	at java.lang.Class.getDeclaredConstructor(Class.java:2178)
	at com.datastax.driver.core.utils.Bytes.<clinit>(Bytes.java:60)
	at com.datastax.driver.core.BytesTest.main(BytesTest.java:11)
```

tolbertam:
Since this code was originally pulled from Cassandra, it got me thinking if the same issue might exist there as well and that seems to be the case: https://github.com/apache/cassandra/blob/8b3a60b9a7dbefeecc06bace617279612ec7092d/src/java/org/apache/cassandra/utils/Hex.java#L28.    Should we consider addressing this there as well?

tolbertam:
Thinking about this some more, I wonder if we should just remove the optimization all together and just always use `new String(c)`.  This has been broken with java 8 forever and has gone unnoticed until now.  I don't anticipate that toHexString is on any hot path, unless there is something i'm missing?  I see it is used by:

* `format` for `CustomCodec` and `BlobCodec`, which is only used with the `QueryBuilder` when using custom types / blobs as parameters (which could go away if we just used positional values which is supported in C* 2.0+)
* `toString` implementations of `OPPToken`  and `Rows`
* Logging warn statement in `AggregateMetadata.build` when initcond value can't be serialized.

TheIndifferent:
@tolbertam yes, that method is gone in Java 8. Since the driver seems to be targeting java 6, I thought having both implementations would be good.
Personally I think it would be good to have efficient zero-copy conversion of byte array to hex strings, so if you'd like I can make the same modification in other places that use private String constructors.

TheIndifferent:
I don't get it, there were no additional changes, but this time there are 5 AppVeyor test failures instead of 1?...

tolbertam:
> I don't get it, there were no additional changes, but this time there are 5 AppVeyor test failures instead of 1?...

The tests that are failing are integration tests, which depend on firing up local cassandra clusters.  For some reason that has become less reliable recently.  I'll have to look into if there's been any recent changes in appveyor infra (like instance size changes which has happened in the past).  I would just ignore these for now.  The travis run only runs unit tests and that looks to be running fine.

TheIndifferent:
So, would you like me to improve it in any way for it to be merged? Or would you rather drop it completely?

olim7t:
> Thinking about this some more, I wonder if we should just remove the optimization all together and just always use new String(c). This has been broken with java 8 forever and has gone unnoticed until now. I don't anticipate that toHexString is on any hot path

💯 to that.
Not criticising the quality of this PR in any way, but the initial feature was premature optimization. I think we can get rid of it.

PR #862, opened 2017-07-19
JAVA-1555: Include VIEW and CDC in WriteType

tolbertam:


adutra (commented):
One question: does any of these new types warrant changes in `DefaultRetryPolicy.onWriteTimeout`?

tolbertam:
> One question: does any of these new types warrant changes in DefaultRetryPolicy.onWriteTimeout?

Good question, my intuition is it would be difficult to say one way or another and probably too soon to know.  It looks like that the CDC case would happen when you are overrunning your configured space.  In the VIEW case, it indicates contention on a key.  I think in general it is not likely to encounter either of these, but if you did, not retrying is sensible.

adutra (approved):

olim7t (approved):

newkek (approved):

PR #863, opened 2017-07-27
JAVA-1582, JAVA-1457, JAVA-1599 Consistent order in exportAsString and other fixes

tolbertam:
For [JAVA-1582](https://datastax-oss.atlassian.net/browse/JAVA-1582).  This could use an integration test, so I will add one.

-------

JAVA-1407 improved exportAsString to ensure that user types were
described in topological order.  However a side effect of this change is
that User Types are not always consistently ordered as it depended on
undefined Map ordering behavior which may vary based on JDK
implementation.

While this is not a functional issue, it would be preferable that the
same order were always returned.  This implementation ensures this by
ordering the adjacency list of a vertex by alphabetical name of the user
type.  If there are no dependencies between user types, the result of
this change is that user types are ordered alphabetically.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
@@ -33,6 +33,14 @@ public abstract class AbstractTableMetadata {
         }
     };
 
+    // comparator for ordering tables and views by name.
+    static final Comparator<AbstractTableMetadata> byNameComparator = new Comparator<AbstractTableMetadata>() {
+        @Override
+        public int compare(AbstractTableMetadata o1, AbstractTableMetadata o2) {
+            return o1.getName().compareTo(o2.getName());
+        }
+    };
+
     static final Predicate<ClusteringOrder> isAscending = new Predicate<ClusteringOrder>() {
         @Override
         public boolean apply(ClusteringOrder o) {
@@ -237,10 +245,10 @@ public abstract class AbstractTableMetadata {
 
     protected StringBuilder appendOptions(StringBuilder sb, boolean formatted) {
         // Options
-        sb.append(" WITH ");
+        sb.append("WITH ");
         if (options.isCompactStorage())
             and(sb.append("COMPACT STORAGE"), formatted);
-        if (!Iterables.all(clusteringOrder, isAscending))
+        if (!clusteringOrder.isEmpty())
             and(appendClusteringOrder(sb), formatted);
         sb.append("read_repair_chance = ").append(options.getReadRepairChance());
         and(sb, formatted).append("dclocal_read_repair_chance = ").append(options.getLocalReadRepairChance());
@@ -285,7 +293,7 @@ public abstract class AbstractTableMetadata {
         sb.append("CLUSTERING ORDER BY (");
         for (int i = 0; i < clusteringColumns.size(); i++) {
             if (i > 0) sb.append(", ");
-            sb.append(clusteringColumns.get(i).getName()).append(' ').append(clusteringOrder.get(i));
+            sb.append(Metadata.quoteIfNecessary(clusteringColumns.get(i).getName())).append(' ').append(clusteringOrder.get(i));
         }
         return sb.append(')');
     }
@@ -310,18 +318,7 @@ public abstract class AbstractTableMetadata {
     }
 
     private StringBuilder and(StringBuilder sb, boolean formatted) {
-        return newLine(sb, formatted).append(spaces(2, formatted)).append(" AND ");
-    }
-
-    static String spaces(int n, boolean formatted) {
-        if (!formatted)
-            return "";
-
-        StringBuilder sb = new StringBuilder();
-        for (int i = 0; i < n; i++)
-            sb.append(' ');
-
-        return sb.toString();
+        return spaceOrNewLine(sb, formatted).append("AND ");
     }
 
     static StringBuilder newLine(StringBuilder sb, boolean formatted) {
@@ -331,7 +328,7 @@ public abstract class AbstractTableMetadata {
     }
 
     static StringBuilder spaceOrNewLine(StringBuilder sb, boolean formatted) {
-        sb.append(formatted ? '\n' : ' ');
+        sb.append(formatted ? "\n    " : ' ');
         return sb;
     }
 
--- a/driver-core/src/main/java/com/datastax/driver/core/DirectedGraph.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/DirectedGraph.java
@@ -33,8 +33,10 @@ class DirectedGraph<V> {
     final Map<V, Integer> vertices;
     final Multimap<V, V> adjacencyList;
     boolean wasSorted;
+    final Comparator<V> comparator;
 
-    DirectedGraph(List<V> vertices) {
+    DirectedGraph(Comparator<V> comparator, List<V> vertices) {
+        this.comparator = comparator;
         this.vertices = Maps.newHashMapWithExpectedSize(vertices.size());
         this.adjacencyList = HashMultimap.create();
 
@@ -43,8 +45,8 @@ class DirectedGraph<V> {
         }
     }
 
-    DirectedGraph(V... vertices) {
-        this(Arrays.asList(vertices));
+    DirectedGraph(Comparator<V> comparator, V... vertices) {
+        this(comparator, Arrays.asList(vertices));
     }
 
     /**
@@ -65,16 +67,21 @@ class DirectedGraph<V> {
 
         Queue<V> queue = new LinkedList<V>();
 
-        for (Map.Entry<V, Integer> entry : vertices.entrySet()) {
-            if (entry.getValue() == 0)
-                queue.add(entry.getKey());
+        // Sort vertices so order of evaluation is always the same (instead of depending on undefined map order behavior)
+        List<V> orderedVertices = new ArrayList<V>(vertices.keySet());
+        Collections.sort(orderedVertices, comparator);
+        for (V v : orderedVertices) {
+            if (vertices.get(v) == 0)
+                queue.add(v);
         }
 
         List<V> result = Lists.newArrayList();
         while (!queue.isEmpty()) {
             V vertex = queue.remove();
             result.add(vertex);
-            for (V successor : adjacencyList.get(vertex)) {
+            List<V> adjacentVertices = new ArrayList<V>(adjacencyList.get(vertex));
+            Collections.sort(adjacentVertices, comparator);
+            for (V successor : adjacentVertices) {
                 if (decrementAndGetCount(successor) == 0)
                     queue.add(successor);
             }
--- a/driver-core/src/main/java/com/datastax/driver/core/FunctionMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/FunctionMetadata.java
@@ -174,11 +174,9 @@ public class FunctionMetadata {
                 first = false;
             else
                 sb.append(',');
-            TableMetadata.newLine(sb, formatted);
             String name = entry.getKey();
             DataType type = entry.getValue();
             sb
-                    .append(TableMetadata.spaces(4, formatted))
                     .append(Metadata.quoteIfNecessary(name))
                     .append(' ')
                     .append(type.asFunctionParameterString());
@@ -190,7 +188,7 @@ public class FunctionMetadata {
 
         TableMetadata.spaceOrNewLine(sb, formatted)
                 .append("RETURNS ")
-                .append(returnType);
+                .append(returnType.asFunctionParameterString());
 
         TableMetadata.spaceOrNewLine(sb, formatted)
                 .append("LANGUAGE ")
--- a/driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
@@ -16,6 +16,7 @@
 package com.datastax.driver.core;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableSortedSet;
 import com.google.common.collect.Lists;
 
 import java.util.*;
@@ -243,6 +244,29 @@ public class KeyspaceMetadata {
         return aggregates.remove(fullName);
     }
 
+    // comparators for ordering types in cqlsh output.
+
+    private static final Comparator<UserType> typeByName = new Comparator<UserType>() {
+        @Override
+        public int compare(UserType o1, UserType o2) {
+            return o1.getTypeName().compareTo(o2.getTypeName());
+        }
+    };
+
+    private static final Comparator<FunctionMetadata> functionByName = new Comparator<FunctionMetadata>() {
+        @Override
+        public int compare(FunctionMetadata o1, FunctionMetadata o2) {
+            return o1.getSimpleName().compareTo(o2.getSimpleName());
+        }
+    };
+
+    private static final Comparator<AggregateMetadata> aggregateByName = new Comparator<AggregateMetadata>() {
+        @Override
+        public int compare(AggregateMetadata o1, AggregateMetadata o2) {
+            return o1.getSimpleName().compareTo(o2.getSimpleName());
+        }
+    };
+
     /**
      * Returns a {@code String} containing CQL queries representing this
      * keyspace and the user types and tables it contains.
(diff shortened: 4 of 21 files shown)


driver-core/src/main/java/com/datastax/driver/core/DirectedGraph.java
```diff
@@ -65,16 +67,21 @@ void addEdge(V from, V to) {
 
         Queue<V> queue = new LinkedList<V>();
 
-        for (Map.Entry<V, Integer> entry : vertices.entrySet()) {
-            if (entry.getValue() == 0)
-                queue.add(entry.getKey());
+        // Sort vertices so order of evaluation is always the same (instead of depending on undefined map order behavior)
+        List<V> orderedVertices = new ArrayList<V>(vertices.keySet());
+        Collections.sort(orderedVertices, comparator);
```
tolbertam, line 72: Not too concerned about performance as number of UDTs should be rather small, and we only sort at top level and within adjacency lists.

driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
```diff
@@ -663,6 +663,24 @@ else if (targetType == AGGREGATE)
...
         }
 
+        // Used by maybeSortUdts to sort at each dependency group alphabetically.
+        private static final Comparator<Row> sortByTypeName = new Comparator<Row>() {
+            @Override
+            public int compare(Row o1, Row o2) {
+                String type1 = o1.getString(UserType.TYPE_NAME);
+                String type2 = o2.getString(UserType.TYPE_NAME);
+
+                if (type1 == null && type2 == null) {
```
tolbertam, line 673: Technically these should never be null, could remove check if you think it is being overcautious

driver-core/src/main/java/com/datastax/driver/core/DirectedGraph.java
```diff
@@ -33,8 +33,10 @@
     final Map<V, Integer> vertices;
     final Multimap<V, V> adjacencyList;
     boolean wasSorted;
+    final Comparator<V> comparator;
 
-    DirectedGraph(List<V> vertices) {
+    DirectedGraph(Comparator<V> comparator, List<V> vertices) {
```
tolbertam, line 38: I considered having the `Comparator` be optional, but since this is a internal class that is only used for this one purpose, figured i'd leave it as is.

driver-core/src/test/java/com/datastax/driver/core/DirectedGraphTest.java
```diff
@@ -58,24 +68,80 @@ public void should_sort_complex_graph() {
...
+        // Topological sort order should be : [7,6],[5],[10],[2,1],[0]
+        // There's no guarantee on the order within the same level, so we use sublists:
+        List<Integer> sorted = g.topologicalSort();
+        System.out.println(sorted);
         assertThat(sorted.subList(0, 2))
-                .contains("G", "H");
-        assertThat(sorted.subList(2, 4))
-                .contains("F", "E");
+                .containsExactly(7, 6);
+        // 5 comes before 10 even though they appear at the same depth.  This happens because 5's (7) dependency
```
tolbertam: Note this detail, since 7's dependents (5,9) are evaluated first, 5 comes before 10, even though they are the same depth.   I don't expect this to bother anyone, but I thought I'd expound on that detail.

driver-core/src/test/java/com/datastax/driver/core/DirectedGraphTest.java
```diff
@@ -58,24 +68,80 @@ public void should_sort_complex_graph() {
...
+        g.addEdge(6, 9);
+        g.addEdge(9, 2);
+        g.addEdge(10, 2);
+        g.addEdge(9, 1);
+        g.addEdge(1, 0);
+
+        // Topological sort order should be : [7,6],[5],[10],[2,1],[0]
+        // There's no guarantee on the order within the same level, so we use sublists:
+        List<Integer> sorted = g.topologicalSort();
+        System.out.println(sorted);
```
tolbertam: Need to remove these

tolbertam:
Another case that causes the order to be off is if UDTs are added on the fly.  The map is only topologically sorted when multiple UDTs are parsed from schema tables.  On schema refresh from adding a UDT, topological sort isn't required because we know that the existing UDTs are not dependent on a newly added UDT.

I guess to resolve that we could revaluate the Map every time a UDT is added.  That creates more work on evaluation, but this I don't think we need to worry too much about that as UDTs being added is not a heavily exercised path.

tolbertam:
Looking at the code more, I think it would be pretty difficult to hold to that invariant that the ordering of user types is always consistent in `KeyspaceMetadata`.   I think we could make the ordering consistent in `exportAsString`, by rebuilding the graph and doing another `topologicalSort`.  That would ensure that the ordering is the same.  

I'll commit that for now, but I don't think we ever stated that `exportAsString` has a defined ordering or behaves in an idempotent manner.  I think the main thing that's important is that the schema is valid, which is why it is ordered topologically.

I also noticed that all the other types on `KeyspaceMetadata` are `ConcurrentHashMap`, so their ordering may not be well defined/consistent.

tolbertam:
I pushed a commit that ensures user types are always in the same order, regardless of whether or not they were all created before the last schema refresh, or if they were created during the lifetime of the session, but i think it's a bit complex and maybe not worth the complexity.

There are three options:

1. Don't worry about consistent ordering with `exportAsString`, we should only be concerned that the schema provided is *correct*.  This is the existing implementation.

2. Try to maintain consistent ordering on full schema refresh.

3. Try to maintain consistent ordering always.

I think 3) is an ideal goal, but i'm not confident it is worth the complexity with our existing implementation.   Also, since other collections use ConcurrentHashMap which does not make order guarantees [1],  I don't think we can really surely assert this without fixing more code.

[1]: from javadoc: *...Because the elements of a ConcurrentHashMap are not ordered in any particular way, and may be processed in different orders in different parallel executions, the correctness of supplied functions should not depend on any ordering...*

tolbertam:
While testing I found some additional inadequacies in `exportAsString` that I attempted to fix in 0710956.  I'm going to leave this PR in 'work-in-progress' so I can add some tests as there are not currently very many around `exportAsString`

driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
```diff
@@ -262,21 +286,61 @@ public String exportAsString() {
...
         sb.append(asCQLQuery()).append('\n');
 
-        for (UserType udt : userTypes.values())
+        // include types, tables, views, functions and aggregates, each ordered by name, with one small exception
+        // being that user types are ordered topologically and then by name within same level.
+        for (UserType udt : getSortedUserTypes())
             sb.append('\n').append(udt.exportAsString()).append('\n');
 
-        for (TableMetadata tm : tables.values())
+        for (AbstractTableMetadata tm : ImmutableSortedSet.orderedBy(AbstractTableMetadata.byNameComparator).addAll(tables.values()).build())
```
tolbertam, line 294: ImmutableSortedSet is not necessary, but it's nice because it makes it easy to make a copy of a collection and sort it.

driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
```diff
@@ -262,21 +286,61 @@ public String exportAsString() {
...
+        for (UserType from : unsortedTypes) {
+            for (UserType to : unsortedTypes) {
+                if (from != to && dependsOn(to, from))
+                    graph.addEdge(from, to);
+            }
+        }
+        return graph.topologicalSort();
+    }
+
+    private boolean dependsOn(UserType udt1, UserType udt2) {
```
tolbertam, line 319: It is kind of annoying that we have to duplicate this code from `V3SchemaParser` but it's necessary because here we are working on the UDT's directly, where in the other code we're working on the Row.

driver-core/src/main/java/com/datastax/driver/core/TableMetadata.java
```diff
@@ -384,12 +385,30 @@ public String exportAsString() {
 
         sb.append(super.exportAsString());
 
-        for (IndexMetadata index : indexes.values()) {
-            sb.append('\n').append(index.asCQLQuery());
+        if (!indexes.isEmpty()) {
+            sb.append('\n');
```
tolbertam, line 389: We have to be stingy with newlines here because `KeyspaceMetadata` adds one after the tables total metadata, so we don't want duplicate newlines.

driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
```diff
@@ -240,7 +248,7 @@ protected StringBuilder appendOptions(StringBuilder sb, boolean formatted) {
         sb.append(" WITH ");
         if (options.isCompactStorage())
             and(sb.append("COMPACT STORAGE"), formatted);
-        if (!Iterables.all(clusteringOrder, isAscending))
+        if (!clusteringOrder.isEmpty())
```
tolbertam: This change isn't necessary, but I noticed this is what the python driver does, and we are already pretty aggressive about including options which the user wouldn't (usually) specify themselves.

driver-core/src/test/java/com/datastax/driver/core/ExportAsStringTest2x.java
```diff
@@ -0,0 +1,108 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+@CassandraVersion("2.0")
+public class ExportAsStringTest2x extends CCMTestsSupport {
+
+    /**
+     * A version of {@link ExportAsStringTest} for C* 2.0 and 2.1 clusters.
+     */
+    @Test(groups = "short")
+    public void create_schema_and_ensure_exported_cql_is_expected() {
```
tolbertam: This is a duplicate of `ExportAsStringTest`, but for 2.0 and 2.1 only.  Separated them out before `ExportAsStringTest` requires UDFs to be enabled, but i just realized we already have selective logic to remove that configuration for older versions..will unify.

driver-core/src/main/java/com/datastax/driver/core/FunctionMetadata.java
```diff
@@ -174,11 +174,9 @@ private String asCQLQuery(boolean formatted) {
                 first = false;
             else
                 sb.append(',');
-            TableMetadata.newLine(sb, formatted);
```
tolbertam, line 177: cqlsh puts all arguments on the same line as `CREATE FUNCTION`,

driver-core/src/main/java/com/datastax/driver/core/FunctionMetadata.java
```diff
@@ -190,7 +188,7 @@ private String asCQLQuery(boolean formatted) {
 
         TableMetadata.spaceOrNewLine(sb, formatted)
                 .append("RETURNS ")
-                .append(returnType);
+                .append(returnType.asFunctionParameterString());
```
tolbertam, line 191: Needed to remove `frozen` from return type (which isn't valid)

tolbertam:
Removing [wip] tag as this should be good to go pending tests.   When reviewed i'll clean up commit history and add changelog.

driver-core/src/test/java/com/datastax/driver/core/DirectedGraphTest.java
```diff
@@ -58,24 +68,78 @@ public void should_sort_complex_graph() {
         g.addEdge("D", "B");
         g.addEdge("B", "A");
 
-        // Topological sort order should be : GH,FE,D,CB,A
+        // Topological sort order should be : GH,E,F,D,BC,A
         // There's no guarantee on the order within the same level, so we use sublists:
```
olim7t: Is that still relevant?

driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
```diff
@@ -262,21 +286,61 @@ public String exportAsString() {
 
         sb.append(asCQLQuery()).append('\n');
 
-        for (UserType udt : userTypes.values())
+        // include types, tables, views, functions and aggregates, each ordered by name, with one small exception
+        // being that user types are ordered topologically and then by name within same level.
+        for (UserType udt : getSortedUserTypes())
```
olim7t, line 291: If we re-sort when exporting as CQL, do we still need the changes from the first commit? Having a consistent order matters mostly for `exportAsString`, for a programmatic traversal I don't think it's a big deal if it differs across VMs.

driver-core/src/test/java/com/datastax/driver/core/UnresolvedUserTypeTest.java
```diff
@@ -28,6 +28,42 @@
 @CassandraVersion("3.0")
 public class UnresolvedUserTypeTest extends CCMTestsSupport {
 
+    private static final String keyspace = "unresolved_user_type_test";
```
olim7t: Those are constants, the names should be capitalized.

driver-core/src/test/java/com/datastax/driver/core/UnresolvedUserTypeTest.java
```diff
@@ -28,6 +28,42 @@
...
+
+    private static final String expectedSchema = "CREATE KEYSPACE unresolved_user_type_test WITH REPLICATION = { 'class' : 'org.apache.cassandra.locator.SimpleStrategy', 'replication_factor': '1' } AND DURABLE_WRITES = true;\n" +
```
olim7t: Maybe reuse the keyspace name constant here for consistency.

olim7t (changes requested):

driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
tolbertam: I was unsure if maintaining the topographical order in `KeyspaceMetadata.userTypes` was important or not.  If it's not it would simplify some code / remove some semi-duplication, if agree I'll remove it.

driver-core/src/test/java/com/datastax/driver/core/DirectedGraphTest.java
tolbertam: You are right, there *is* a guarantee that it is ordered the same way.   I should update the test to no longer check on sublists and validate it as a whole, will do that.

driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
tolbertam: There already isn't a consistent ordering in the collection itself depending on the order of UDT creation (if created within the same Cluster instance), but they are always ordered topologically (just not consistently).  I think it is not important for it to be sorted at that time, but only when exporting it to cql.  I'll add a separate commit that backs that out, but we can always remove that commit if we want to keep it.   I think the order isn't significant (nor is it documented) for the actual collection members, but it is significant when exporting as a cql string because if it's not ordered correctly it's not valid.

tolbertam: Actually it does appear to be relevant.   When connecting and building metadata, if a UDT depends on another UDT, it needs to be built in order, otherwise:

```
com.datastax.driver.core.exceptions.UnresolvedUserTypeException: Cannot resolve user type unresolved_user_type_test."C"
	at com.datastax.driver.core.DataTypeCqlNameParser.parse(DataTypeCqlNameParser.java:149)
	at com.datastax.driver.core.DataTypeCqlNameParser.parse(DataTypeCqlNameParser.java:120)
	at com.datastax.driver.core.UserType.build(UserType.java:93)
	at com.datastax.driver.core.SchemaParser.buildUserTypes(SchemaParser.java:196)
	at com.datastax.driver.core.SchemaParser.buildKeyspaces(SchemaParser.java:127)
	at com.datastax.driver.core.SchemaParser.refresh(SchemaParser.java:64)
	at com.datastax.driver.core.ControlConnection.refreshSchema(ControlConnection.java:336)
	at com.datastax.driver.core.ControlConnection.tryConnect(ControlConnection.java:276)
	at com.datastax.driver.core.ControlConnection.reconnectInternal(ControlConnection.java:201)
	at com.datastax.driver.core.ControlConnection.connect(ControlConnection.java:79)
	at com.datastax.driver.core.Cluster$Manager.negotiateProtocolVersionAndConnect(Cluster.java:1600)
	at com.datastax.driver.core.Cluster$Manager.init(Cluster.java:1518)
	at com.datastax.driver.core.Cluster.getMetadata(Cluster.java:399)
	at com.datastax.driver.core.UnresolvedUserTypeTest.should_resolve_nested_user_types(UnresolvedUserTypeTest.java:114)
```

tolbertam: Going to leave that as is for now for the preceding comments reason, but let me know if you'd like me to adjust it.

PR #864, opened 2017-07-28
JAVA-1587: Deterministic ordering of columns used in mapper query

collinsauve:
See https://datastax-oss.atlassian.net/browse/JAVA-1587

(P.S. this is my first time contributing to java-driver, sorry if I make any obvious mistakes)

### DIFF
--- a/driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
+++ b/driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
@@ -220,8 +220,18 @@ public class Mapper<T> {
         }
     }
 
+    private static class AliasedMappedPropertyComparator implements Comparator<AliasedMappedProperty>
+    {
+        @Override
+        public int compare(AliasedMappedProperty o1, AliasedMappedProperty o2) {
+            String col1Name = o1.mappedProperty.getMappedName();
+            String col2Name = o2.mappedProperty.getMappedName();
+            return col1Name.compareTo(col2Name);
+        }
+    }
+
     private ListenableFuture<BoundStatement> saveQueryAsync(T entity, final EnumMap<Option.Type, Option> options) {
-        final Map<AliasedMappedProperty, Object> columnToValue = new HashMap<AliasedMappedProperty, Object>();
+        final Map<AliasedMappedProperty, Object> columnToValue = new TreeMap<AliasedMappedProperty, Object>(new AliasedMappedPropertyComparator());
         final boolean useUnsetForNullValue = !shouldSaveNullFields(options) && manager.protocolVersionAsInt >= 4;
         final boolean includeColumnsWithNullValue = shouldSaveNullFields(options) || useUnsetForNullValue;
 
--- a/driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
+++ b/driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
@@ -220,18 +220,8 @@ public class Mapper<T> {
         }
     }
 
-    private static class AliasedMappedPropertyComparator implements Comparator<AliasedMappedProperty>
-    {
-        @Override
-        public int compare(AliasedMappedProperty o1, AliasedMappedProperty o2) {
-            String col1Name = o1.mappedProperty.getMappedName();
-            String col2Name = o2.mappedProperty.getMappedName();
-            return col1Name.compareTo(col2Name);
-        }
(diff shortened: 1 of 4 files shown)


tolbertam:
Hi @collinsauve, at first glance this looks great.  Agree that it would be nice to consistently order the columns in the prepared statement.  Taking a deeper look and will provide some more feedback or my thumbs up later today.

tolbertam:
Just wrote a quick test, that creates 5 Sessions and Mappers for a User object with fields `login`, `phone`, and `name`:

Before change:
```
INSERT INTO ks_1.user (phone,login,name) VALUES (?,?,?);
INSERT INTO ks_1.user (login,name,phone) VALUES (?,?,?);
INSERT INTO ks_1.user (login,name,phone) VALUES (?,?,?);
INSERT INTO ks_1.user (phone,name,login) VALUES (?,?,?);
INSERT INTO ks_1.user (name,phone,login) VALUES (?,?,?);
```

With change:
```
INSERT INTO ks_1.user (login,name,phone) VALUES (?,?,?);
INSERT INTO ks_1.user (login,name,phone) VALUES (?,?,?);
INSERT INTO ks_1.user (login,name,phone) VALUES (?,?,?);
INSERT INTO ks_1.user (login,name,phone) VALUES (?,?,?);
INSERT INTO ks_1.user (login,name,phone) VALUES (?,?,?);
```

So this is working great!  Would you mind if I threw a test that validates this on your branch?

driver-mapping/src/main/java/com/datastax/driver/mapping/Mapper.java
```diff
@@ -220,8 +220,18 @@ public Statement saveQuery(T entity, Option... options) {
...
+        public int compare(AliasedMappedProperty o1, AliasedMappedProperty o2) {
+            String col1Name = o1.mappedProperty.getMappedName();
+            String col2Name = o2.mappedProperty.getMappedName();
+            return col1Name.compareTo(col2Name);
+        }
+    }
+
     private ListenableFuture<BoundStatement> saveQueryAsync(T entity, final EnumMap<Option.Type, Option> options) {
-        final Map<AliasedMappedProperty, Object> columnToValue = new HashMap<AliasedMappedProperty, Object>();
+        final Map<AliasedMappedProperty, Object> columnToValue = new TreeMap<AliasedMappedProperty, Object>(new AliasedMappedPropertyComparator());
```
tolbertam: Small nit, instead of creating a new comparator each time, you could just have 1 static comparator, i.e.:

```java
    private static final Comparator<AliasedMappedProperty> PROPERTY_COMPARATOR = new Comparator<AliasedMappedProperty>() {
        @Override
        public int compare(AliasedMappedProperty o1, AliasedMappedProperty o2) {
            String col1Name = o1.mappedProperty.getMappedName();
            String col2Name = o2.mappedProperty.getMappedName();
            return col1Name.compareTo(col2Name);
        }
    };
```

tolbertam:
[Here's a patch](https://gist.github.com/tolbertam/0fa9918858f79fd278763fc7864c20b0) that adds a test for validating the fix.  If it's ok with you I can add it to this branch.  Other than [my comment about making the comparator static](https://github.com/datastax/java-driver/pull/864#discussion_r130642148), this looks great to me 👍

collinsauve:
Those changes sound good to me, do I need to give you write access to my fork to let you modify the PR?

tolbertam:
I believe I have access, will try it out, thanks!

tolbertam:
Pushed the test and the changes.  After thinking about it more, I figured we might as well make `AliasedMappedProperty` `Comparable` which accomplishes the same thing.

I also just realized we weren't considering `alias`, but I don't think that will really be a problem unless you have the same column/selector twice and you want to differentiate with an alias,  I think while possible that is not practical and not sure if it is worth adding it to compareTo (but will leave that to the other devs to judge when they review).

tolbertam (approved):

adutra (approved):
👍 one small nit though: changelog entry is missing.

olim7t (approved):

DukeDai:
The issue for query is still there, 3.4 and 3.6. Considering that cassandra issue https://issues.apache.org/jira/browse/CASSANDRA-13641. Not sure whether later version has fixed similar issue for query or not. Could you please confirm?

tolbertam:
the issue should be fixed from driver 3.3.1+, can you share a test case where you observe that this isn't the case?

PR #866, opened 2017-08-01
Integration tests for config reloading

tolbertam:
Also includes fix to prevent `SimulacronRule.before` from being invoked twice.

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/api/core/config/DriverConfigProfileReloadIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/api/core/config/DriverConfigProfileReloadIT.java
@@ -1,0 +1,219 @@
+/*
+ * Copyright (C) 2017-2017 DataStax Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


integration-tests/src/test/java/com/datastax/oss/driver/api/core/config/DriverConfigProfileReloadIT.java
```diff
@@ -0,0 +1,160 @@
...
+      configSource.set("request.timeout = 10s");
+      TimeUnit.SECONDS.sleep(3);
+
+      // Execute again, should not timeout.
+      session.execute(query);
+    }
+  }
+
+  @Test
+  public void should_not_allow_dynamically_adding_profile() throws Exception {
```
olim7t, line 130: I need to check why I didn't allow this, in theory it would be possible.

olim7t: It's a performance consideration: profiles are stored in an immutable map (`TpyeSafeDriverConfig#profiles`). If we allow new profiles at runtime, we have to turn it into a concurrent map.
The use case sounds extremely marginal to me, so I'm not sure if it's worth it.

olim7t:
👍  I've made a few changes to use events to wait for reloads, and added a test for forcing a reload with an event.

PR #867, opened 2017-08-02
JAVA-1585: Add GenericType#where

olim7t:


tolbertam (commented):
Looks great! I thought making this work would be way more involved than it ended up being.  I'll play with this a bit, but I think this will solve the issue.

tolbertam (approved):
Added an integration test using this, seems to be working great!

integration-tests/src/test/java/com/datastax/oss/driver/api/core/type/codec/registry/CodecRegistryIT.java
```diff
@@ -170,4 +185,171 @@ public void should_be_able_to_register_and_use_custom_codec() {
       assertThat(row.getFloat(0)).isEqualTo(3.0f);
     }
   }
+
+  // TODO: consider moving this into source as it could be generally useful.
+  private abstract static class MappingCodec<O, I> implements TypeCodec<O> {
```
tolbertam, line 190: I followed `OptionalCodec`s implementation in java driver extras, which also used `MappingCodec` which is generally useful for going from one java type to another, not sure if we wanted something like this in the source path, I think at this time, probably not, but might be nice if we decide to add additional mapped codecs.

PR #868, opened 2017-08-03
JAVA-1568: Handle Reconnection#reconnectNow/stop while the current attempt is still in progress

olim7t:


core/src/test/java/com/datastax/oss/driver/internal/core/pool/ChannelPoolReconnectTest.java
```diff
@@ -133,4 +136,59 @@ public void should_reconnect_when_channel_starts_graceful_shutdown() throws Exce
 
     factoryHelper.verifyNoMoreCalls();
   }
+
+  @Test
+  public void should_let_current_attempt_complete_when_reconnecting_now()
```
olim7t, line 141: @tolbertam Do you remember the integration test that led to the ticket?
I think this unit test covers it, but it would be good to double check.

tolbertam: I can't quite recall which test was causing this, but I think it happened on suite tear down.  It happens pretty consistently when our integration tests run, so I'll see if I can pin that down.

tolbertam: I added a debug log to [JAVA-1568](https://datastax-oss.atlassian.net/browse/JAVA-1568).  The test where this happens is `HeartbeatIT#should_close_connection_when_heartbeat_times_out()`.  Although it's very hard to reproduce in isolation, but I see it frequently when running the entire short suite.

core/src/test/java/com/datastax/oss/driver/internal/core/pool/ChannelPoolReconnectTest.java
```diff
@@ -133,4 +136,59 @@ public void should_reconnect_when_channel_starts_graceful_shutdown() throws Exce
...
+      throws ExecutionException, InterruptedException {
+    Mockito.when(reconnectionSchedule.nextDelay()).thenReturn(Duration.ofNanos(1));
```
tolbertam, line 143: Awesome test 👍

tolbertam (approved):

core/src/test/java/com/datastax/oss/driver/internal/core/pool/ChannelPoolReconnectTest.java
olim7t: So this fixes it?

tolbertam: Ah yes, sorry forgot to mention when I approved.  I reran the suite repeatedly and could no longer reproduce the AssertionError on ths branch, where on 4.x it manifests ~60% of the time

PR #870, opened 2017-08-07
Node state fixes

olim7t:


tolbertam:
Wow, this looks amazing so far, since it fixes quite a bit will take me a little bit to process, will spend some time reviewing this today.

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
```diff
@@ -344,6 +380,31 @@ private void reconnectNow() {
...
     }
 
+    private void onDistanceEvent(DistanceEvent event) {
+      assert adminExecutor.inEventLoop();
+      this.lastDistanceEvents.put(event.node, event);
+      if (event.distance == NodeDistance.IGNORED
+          && channel != null
+          && !channel.closeFuture().isDone()
+          && event.node.getConnectAddress().equals(channel.remoteAddress())) {
+        LOG.debug("[{}] Control node {} became IGNORED, reconnecting", logPrefix, event.node);
```
tolbertam: Would be nice to make this more explicit and say "reconnecting to a different node".

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
```diff
@@ -344,6 +380,31 @@ private void reconnectNow() {
...
+
+    private void onStateEvent(NodeStateEvent event) {
+      assert adminExecutor.inEventLoop();
+      this.lastStateEvents.put(event.node, event);
+      if ((event.newState == null /*(removed)*/ || event.newState == NodeState.FORCED_DOWN)
+          && channel != null
+          && !channel.closeFuture().isDone()
+          && event.node.getConnectAddress().equals(channel.remoteAddress())) {
+        LOG.debug(
+            "[{}] Control node {} was removed or forced down, reconnecting", logPrefix, event.node);
```
tolbertam: Same as above (indicate trying different node)

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
```diff
@@ -337,39 +345,57 @@ private void processDistanceEvent(DistanceEvent event) {
...
+    }
+
+    private void onTopologyEvent(TopologyEvent event) {
+      assert adminExecutor.inEventLoop();
+      if (event.type == TopologyEvent.Type.SUGGEST_UP) {
+        Node node = context.metadataManager().getMetadata().getNodes().get(event.address);
+        if (node.getDistance() != NodeDistance.IGNORED) {
+          LOG.debug(
+              "[{}] Received a SUGGEST_UP event for {}, reconnecting pool now", logPrefix, node);
+          createOrReconnectPool(node);
```
tolbertam, line 380: If the node is already UP, will this cause it to recreate the connection pool?

core/src/test/java/com/datastax/oss/driver/internal/core/session/DefaultSessionTest.java
```diff
@@ -455,6 +455,36 @@ public void should_recreate_pool_if_node_is_forced_back_up() {
   }
 
   @Test
+  public void should_not_recreate_pool_if_node_is_forced_back_up_but_ignored() {
```
tolbertam, line 458: Really nice,  testing for behavior as result of Topology/Status events was difficult in the past, nice that this can be done easily at unit level.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/NodeStateIT.java
```diff
@@ -0,0 +1,381 @@
...
+                    && metadataRegularNode.isReconnecting())
+        .as("Reconnection started")
+        .before(10, TimeUnit.SECONDS)
+        .becomesTrue();
+    assertThat(metadataRegularNode).hasState(NodeState.UP);
+  }
+
+  @Test
+  public void should_mark_regular_node_down_when_no_more_connections() {
+    simulacronRegularNode.rejectConnections(0, RejectScope.UNBIND);
```
tolbertam: `simulacronRegularNode.stop()` can be used as a shortcut for these 2 methods.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/NodeStateIT.java
```diff
@@ -0,0 +1,381 @@
...
+    driverContext
+        .loadBalancingPolicyWrapper()
+        .setDistance(metadataRegularNode, NodeDistance.IGNORED);
+    ConditionChecker.checkThat(() -> metadataRegularNode.getOpenConnections() == 0)
+        .as("Driver closed all connections")
+        .before(10, TimeUnit.SECONDS)
+        .becomesTrue();
+    assertThat(metadataRegularNode).hasState(NodeState.UP).isNotReconnecting();
+
+    driverContext
```
tolbertam: So awesome that you can fire arbitrary events on the event bus 👍

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/NodeStateIT.java
```diff
@@ -0,0 +1,381 @@
...
+                    && metadataRegularNode.getOpenConnections() == 0
+                    && !metadataRegularNode.isReconnecting())
+        .as("SUGGEST_DOWN event applied")
+        .before(10, TimeUnit.SECONDS)
+        .becomesTrue();
+
+    driverContext.eventBus().fire(TopologyEvent.suggestUp(metadataRegularNode.getConnectAddress()));
+    ConditionChecker.checkThat(
+            () ->
+                metadataRegularNode.getState() == NodeState.UP
```
tolbertam: With assertj are mechanisms to do things like apply assertions over Optional values if present (using `hasValueSatisfying`), i.e. from `DriverConfigProfileIT`:

```java
      assertThat(log)
          .isPresent()
          .hasValueSatisfying(
              (l) -> {
                assertThat(l.getConsistency().toString()).isEqualTo("LOCAL_ONE");
                assertThat(l.getSerialConsistency().toString()).isEqualTo("SERIAL");
              });
```

It would be nice if we did something similar with ConditionChecker, that way you can make use of the `NodeMetadataAssertions` that you added.  Something like this:

```java
ConditionChecker.checkThat(() ->
  assertThat(n).hasState(NodeState.UP)
    .hasDistance(NodeDistance.IGNORED)
    .hasOpenConnections(0)
    .isReconnecting()
);
```

I think to make that work we'd have to alter `ConditionChecker` to also accept a `Runnable` where if when evaluated does not throw an `AssertionError` the condition is considered successful.  What do you think?

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/NodeStateIT.java
```diff
@@ -0,0 +1,381 @@
...
+        .as("Node forced back up")
+        .before(10, TimeUnit.SECONDS)
+        .becomesTrue();
+  }
+
+  @Test
+  public void should_force_down_when_ignored() throws InterruptedException {
+    driverContext
+        .loadBalancingPolicyWrapper()
+        .setDistance(metadataRegularNode, NodeDistance.IGNORED);
```
tolbertam, line 319: This persists between tests, so marking distance as `IGNORED` here causes subsequent tests to fail

tolbertam: Wonder if it might be better to make the rules `@Rule` instead of `@ClassRule` instead of marking everything static, that way we don't have to worry about state between tests.  it'll add some runtime to the tests, but it's very small (milliseconds)

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/NodeStateIT.java
```diff
@@ -0,0 +1,381 @@
...
+import java.util.concurrent.TimeUnit;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+
+import static com.datastax.oss.driver.assertions.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+public class NodeStateIT {
```
tolbertam, line 46: Super cool that this validates a bunch of non-trivial scenarios very quickly and the code is very easy to follow 👍

tolbertam (commented):
Looks great!  I'd like to do some more testing, but had some feedback that I thought I should commit more and i'll continue the review.

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
olim7t: Actually if a node is up it should never have a null pool, at worst an empty and reconnecting pool. But I figured it wouldn't hurt to recreate the pool if it's null.
Same thing in reverse in `processStateEvent` (if the node is forced back up, it should always create, never reconnect).

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/NodeStateIT.java
olim7t: Yes, that's a good idea.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/NodeStateIT.java
olim7t: Great idea, it's been really annoying to avoid interference between tests.

olim7t: 👍 This is much cleaner, and actually faster than waiting for the cluster to get back to its base state each time.

olim7t:
Note to self: the "Address feedback" commit mixes up changes for 1593 and NodeStateIT. I'll fix that when I rebase for the final merge.
Serves me right for putting too much stuff on the same PR.

olim7t:
There's something wrong with my latest commits, looking into it.

olim7t:
Fixed, I had to force-push, sorry.

tolbertam:
I'm not sure why appveyor felt compelled to add a status check `AppVeyor was unable to build non-mergeable pull request`.  I might disable it completely for now and only use it for adhoc windows testing.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/NodeStateIT.java
```diff
@@ -0,0 +1,369 @@
...
+  }
+
+  @Test
+  public void should_keep_regular_node_up_when_still_one_connection() {
+    simulacronRegularNode.rejectConnections(0, RejectScope.UNBIND);
+    NodeConnectionReport report = simulacronRegularNode.getConnections();
+    simulacron.cluster().closeConnection(report.getConnections().get(0), CloseType.DISCONNECT);
+
+    ConditionChecker.checkThat(
+            () -> assertThat(metadataRegularNode).isUp().hasOpenConnections(1).isReconnecting())
```
tolbertam, line 118: Awesome, change to allow asserts in `ConditionChecker` looks really nice, code is much more concise and nice that we can use custom assertions.

tolbertam (approved):
Looks great!  I'll put this through its paces in endurance test environment, but am 👍 for merge

PR #871, opened 2017-08-07
JAVA-1595: Don't use system.local.rpc_address when refreshing node list

olim7t:
Turns out the fix was trivial to implement.


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -129,7 +134,10 @@ public DefaultTopologyMonitor(InternalDriverContext context) {
         peersQuery,
         (controlNodeResult, peersResult) -> {
           List<NodeInfo> nodeInfos = new ArrayList<>();
-          nodeInfos.add(buildNodeInfo(controlNodeResult.iterator().next()));
+          // Don't rely on system.local.rpc_address for the control row, because it mistakenly
```
tolbertam, line 137: Great to document this 👍

tolbertam (approved):
I verified this in the environment I initially encountered this in this morning and the issue was resolved by this change, I think the unit level test is adequate in this case and don't think we need an explicit integration test for it.

PR #873, opened 2017-08-11
JAVA-1295: Auto-detect best protocol version in mixed cluster

olim7t:


core/src/main/java/com/datastax/oss/driver/internal/core/DefaultCluster.java
```diff
@@ -145,25 +149,63 @@ private void init() {
...
+      try {
+        boolean protocolWasForced =
+            context.config().getDefaultProfile().isDefined(CoreDriverOption.PROTOCOL_VERSION);
+        boolean needSchemaRefresh = true;
+        if (!protocolWasForced) {
+          ProtocolVersion currentVersion = context.protocolVersion();
+          ProtocolVersion bestVersion =
+              context
+                  .protocolVersionRegistry()
+                  .highestCommon(metadataManager.getMetadata().getNodes().values());
```
tolbertam, line 170: Should we consider only non-ignored nodes?  Use case I can think of is if you have another DC that hasn't been upgraded yet.   You won't connect to that DC, so you might not want to consider its supported protocol versions. 

One use case is if you are migrating your infra to another cloud provider and brought up another DC with a newer version that you are connecting to but keeping up the old DC in the meantime.  That is probably an edge case, and even in that event you probably don't want to be using a new protocol version anyways.

core/src/main/java/com/datastax/oss/driver/internal/core/CassandraProtocolVersionRegistry.java
```diff
@@ -0,0 +1,180 @@
...
+        LOG.warn(
+            "[{}] Node {} reports null Cassandra version, "
+                + "ignoring it from optimal protocol version computation",
+            logPrefix,
+            node.getConnectAddress());
+        continue;
+      }
+      cassandraVersion = cassandraVersion.nextStable();
+      if (cassandraVersion.compareTo(CASSANDRA_210) < 0) {
+        throw new IllegalArgumentException(
```
tolbertam: Should we create a specific exception type for this case?  In previous driver we had `UnsupportedProtocolVersionException` for this case.

core/src/main/resources/reference.conf
```diff
@@ -32,14 +32,21 @@ datastax-java-driver {
...
     #
-    # This option is not required. If it is absent, the driver will negotiate it with the *first*
-    # node it tries to connect to. More precisely, it will try with the highest supported version,
-    # and if not supported fallback to the second highest and so on.
-    # Once the version is set, it will be used for the lifetime of the driver instance.
-    # Auto detection can be problematic with mixed-version clusters: if the driver connects first
-    # to one of the higher-version nodes, it will negotiate a version that might not work when
-    # connecting to lower-version nodes later. You should force the lowest common protocol version
-    # in that case.
+    # This option is not required. If it is absent, the driver looks up the versions of the nodes at
```
tolbertam, line 35: Nice description 👍

core/src/test/java/com/datastax/oss/driver/internal/core/CassandraProtocolVersionRegistryHighestCommonTest.java
```diff
@@ -0,0 +1,105 @@
...
+        .isEqualTo(CoreProtocolVersion.V3);
+    assertThat(
+            registry.highestCommon(
+                ImmutableList.of(mockNode("2.2.0-rc2"), mockNode("2.2.1"), mockNode("3.1.9"))))
+        .isEqualTo(CoreProtocolVersion.V4);
+  }
+
+  @Test
+  public void should_skip_nodes_that_report_null_version() {
+    assertThat(
```
tolbertam, line 68: Good idea to have a test for this,  the system.peers table has been unpredictable in the past and this is definitely possible.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/ProtocolVersionMixedClusterIT.java
```diff
@@ -0,0 +1,186 @@
...
+              "SELECT cluster_name FROM system.local",
+              "SELECT * FROM system.local",
+              "SELECT * FROM system.peers");
+
+      // Note: the 2.0.0 would be forced down if we try to open a connection to it. We can't check
+      // that here because Simulacron can't prime STARTUP requests.
+    }
+  }
+
+  private BoundCluster mixedVersions(String... versions) {
```
tolbertam, line 148: Nice!  The simulacron cluster builders are kinda ugly to work with, nice to simplify this with a method.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/ProtocolVersionMixedClusterIT.java
```diff
@@ -0,0 +1,186 @@
...
+ * control connection immediately.
+ */
+public class ProtocolVersionMixedClusterIT {
+  @Rule public ExpectedException thrown = ExpectedException.none();
+
+  private static Server server;
+
+  @BeforeClass
+  public static void setup() {
+    server = Server.builder().withAddressResolver(new AddressResolver.Inet4Resolver(9043)).build();
```
tolbertam: It would be nice to reuse `SimulacronRule.server` or find a way to share that globally, that way we only ever create 1 set of event loops for that, but because it's being appropriately closed here it's not too big of a concern.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/ProtocolVersionMixedClusterIT.java
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core;
+
+import com.datastax.oss.driver.api.core.metadata.Node;
```
tolbertam: Two unused imports here

integration-tests/src/test/java/com/datastax/oss/driver/api/core/ProtocolVersionMixedClusterIT.java
tolbertam: Looks like we getting bind exceptions in CI, my guess is because since this in run in parallel with other tests that use `SimulacronRule.server` and at time they check for used addresses they aren't used (maybe small window).

integration-tests/src/test/java/com/datastax/oss/driver/api/core/ProtocolVersionMixedClusterIT.java
```diff
@@ -0,0 +1,186 @@
...
+
+import static com.datastax.oss.driver.assertions.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+/**
+ * Covers protocol re-negotiation with a mixed cluster: if, after the initial connection and the
+ * first node list refresh, we find out that some nodes only support a lower version, reconnect the
+ * control connection immediately.
+ */
+public class ProtocolVersionMixedClusterIT {
```
tolbertam, line 42: Awesome tests!

tolbertam (commented):
Looks great!  Had a few comments

core/src/main/java/com/datastax/oss/driver/internal/core/DefaultCluster.java
olim7t: It would be great in theory, but the problem is that the LBP has not been initialized yet, so we don't know distances.
So it sounds reasonable to file that as an edge case where you would have to force the version.

core/src/main/java/com/datastax/oss/driver/internal/core/CassandraProtocolVersionRegistry.java
olim7t: Yes, actually it's probably better to throw UPVE all the time, I'll look into it.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/ProtocolVersionMixedClusterIT.java
```diff
@@ -0,0 +1,186 @@
...
+    server = Server.builder().withAddressResolver(new AddressResolver.Inet4Resolver(9043)).build();
+  }
+
+  @AfterClass
+  public static void teardown() {
+    server.close();
+  }
+
+  @Test
+  public void should_downgrade_if_peer_does_not_support_negotiated_version() {
```
tolbertam: It's peculiar that `should_downgrade_if_peer_does_not_support_negotiated_version` fails if run after `should_not_downgrade_and_force_down_old_nodes_if_version_forced`  I don't see any shared state between tests other than simulacron itself, will look into it.

Do you mind if I update these tests?  I have a change pending on simulacron to expose the incoming `Frame` in `QueryLog` (that way we can interact with that if we need more info, including protocol version), and I will also seek to consolidate the simulacron use between tests.

tolbertam: It looks like it its a timing thing that has to do with the initialization future for `DefaultCluster` completing before the reconnection for the control connection is complete ([see here](https://github.com/datastax/java-driver/pull/873/files#r133275224))

core/src/main/java/com/datastax/oss/driver/internal/core/DefaultCluster.java
```diff
@@ -145,25 +149,63 @@ private void init() {
...
+              controlConnection.reconnectNow();
+              // Reconnection already triggers a full schema refresh
+              needSchemaRefresh = false;
+            }
+          }
+        }
+        if (needSchemaRefresh) {
+          metadataManager.refreshSchema(SchemaElementKind.WHOLE_SCHEMA, null, null, null);
+        }
+        metadataManager.firstSchemaRefreshFuture().thenAccept(this::afterInitialSchemaRefresh);
```
tolbertam, line 191: Not sure if this is a problem or not, but from [this comment](https://github.com/datastax/java-driver/pull/873/files#r133274225), the `DefaultCluster` init future completes before we have established a downgraded control connection because we only wait on the `firstSchemaRefreshFuture` to complete (which is immediate at the moment).  I think this won't be a problem once the schema code is implemented because the future won't be complete at this point (wont be refreshed until we have a downgraded reconnection)

olim7t: I've changed the code to complete the future from `refreshSchema`, that should do a better job of simulating it until the real schema code is in.

tolbertam: Nice 👍 that seems to make things work, thanks!

tolbertam: There is technically another very small possible timing issue (encountered more rarely):

In [`ControlConnection.onSuccessfulReconnect()`](https://github.com/datastax/java-driver/blob/java1295/core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java#L371), schemaRefresh is scheduled immediately, so it could happen before `refreshNodes` is completed. 

Wouldn't it be better to defer completion on cluster initialization until after `refreshNodes` completes and the LBP is initialized?

olim7t: I still see an intermittent error in this test (query log contains 5 entries instead of 6), I'm trying to reproduce it right now.

tolbertam: Yep, that's exactly what i'm seeing too, it's because the schema refresh future completes before `refreshNodes()` completes, so the peers query is not sent/done by time that cluster init completes.

olim7t: Fixed, it was a race condition again because the node list refresh and schema refresh were running concurrently.

olim7t: Ha, I should have refreshed this page sooner 😃

integration-tests/src/test/java/com/datastax/oss/driver/api/core/ProtocolVersionMixedClusterIT.java
```diff
@@ -0,0 +1,172 @@
...
+
+  private BoundCluster mixedVersions(String... versions) {
+    ClusterSpec clusterSpec = ClusterSpec.builder().withCassandraVersion(versions[0]).build();
+    DataCenterSpec dc0 = clusterSpec.addDataCenter().build();
+    // inherits versions[0]
+    dc0.addNode().build();
+    for (int i = 1; i < versions.length; i++) {
+      dc0.addNode().withCassandraVersion(versions[i]).build();
+    }
+    return SimulacronRule.server.register(clusterSpec);
```
tolbertam, line 156: Bit of a hack for now until I come up with a better way to share that.

tolbertam (approved):
looks great 👍 !

PR #876, opened 2017-09-06
JAVA-1350: Send keyspace option per-query

tolbertam:
Implementation of [JAVA-1350](https://datastax-oss.atlassian.net/browse/JAVA-1350).  Needs more testing.

### DIFF
--- a/clirr-ignores.xml
+++ b/clirr-ignores.xml
@@ -167,11 +167,4 @@
         <justification>Protected method, only used internally</justification>
     </difference>
 
-    <difference>
-        <differenceType>7012</differenceType>
-        <className>com/datastax/driver/core/PreparedStatement</className>
-        <method>java.lang.String getKeyspace()</method>
-        <justification>Only expect to be extend internally</justification>
-    </difference>
-
 </differences>
--- a/clirr-ignores.xml
+++ b/clirr-ignores.xml
@@ -159,4 +159,19 @@
         <justification>Relaxed parameters from List to Iterable for in, lt, lte, eq, gt, and gte</justification>
     </difference>
 
+    <difference>
+        <differenceType>7004</differenceType>
+        <className>com/datastax/driver/core/AbstractSession</className>
+        <method>com.google.common.util.concurrent.ListenableFuture prepareAsync(java.lang.String, java.util.Map)</method>
+        <to>com.google.common.util.concurrent.ListenableFuture prepareAsync(java.lang.String, java.lang.String, java.util.Map)</to>
+        <justification>Protected method, only used internally</justification>
+    </difference>
+
+    <difference>
+        <differenceType>7012</differenceType>
+        <className>com/datastax/driver/core/PreparedStatement</className>
+        <method>java.lang.String getKeyspace()</method>
+        <justification>Only expect to be extend internally</justification>
+    </difference>
+
 </differences>
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -2289,7 +2289,7 @@ public class Cluster implements Closeable {
                     // Empty string mean no particular keyspace to set
                     // Optimization: Only change keyspace for older protocol versions as newer protocols allow
                     // specifying keyspace on prepared statement.
-                    if (protocolVersion().compareTo(ProtocolVersion.V5) >= 0 && !keyspace.isEmpty())
+                    if (protocolVersion().compareTo(ProtocolVersion.V5) < 0 && !keyspace.isEmpty())
                         connection.setKeyspace(keyspace);
 
                     List<Connection.Future> futures = new ArrayList<Connection.Future>(preparedQueries.size());
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -2277,12 +2277,12 @@ public class Cluster implements Closeable {
                 // used for preparing it. However, since we are likely that all prepared query belong to only a handful
                 // of different keyspace (possibly only one), and to avoid setting the current keyspace more than needed,
                 // we first sort the query per keyspace.
-                SetMultimap<String, String> perKeyspace = HashMultimap.create();
+                SetMultimap<String, PreparedStatement> perKeyspace = HashMultimap.create();
                 for (PreparedStatement ps : preparedQueries.values()) {
                     // It's possible for a query to not have a current keyspace. But since null doesn't work well as
                     // map keys, we use the empty string instead (that is not a valid keyspace name).
                     String keyspace = ps.getQueryKeyspace() == null ? "" : ps.getQueryKeyspace();
-                    perKeyspace.put(keyspace, ps.getQueryString());
+                    perKeyspace.put(keyspace, ps);
                 }
 
                 for (String keyspace : perKeyspace.keySet()) {
@@ -2291,8 +2291,8 @@ public class Cluster implements Closeable {
                         connection.setKeyspace(keyspace);
 
                     List<Connection.Future> futures = new ArrayList<Connection.Future>(preparedQueries.size());
-                    for (String query : perKeyspace.get(keyspace)) {
-                        futures.add(connection.write(new Requests.Prepare(query)));
+                    for (PreparedStatement prepared : perKeyspace.get(keyspace)) {
+                        futures.add(connection.write(new Requests.Prepare(prepared.getQueryString(), prepared.getKeyspace())));
                     }
                     for (Connection.Future future : futures) {
                         try {
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -17,6 +17,7 @@ package com.datastax.driver.core;
 
 import com.datastax.driver.core.exceptions.*;
 import com.datastax.driver.core.policies.*;
+import com.datastax.driver.core.ProtocolVersion.ProtocolFeature;
 import com.datastax.driver.core.utils.MoreFutures;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Functions;
@@ -2289,7 +2290,7 @@ public class Cluster implements Closeable {
                     // Empty string mean no particular keyspace to set
                     // Optimization: Only change keyspace for older protocol versions as newer protocols allow
                     // specifying keyspace on prepared statement.
-                    if (!protocolVersion().supportsKeyspaceOnQuery() && !keyspace.isEmpty())
+                    if (!protocolVersion().supports(ProtocolFeature.KEYSPACE_ON_QUERY) && !keyspace.isEmpty())
                         connection.setKeyspace(keyspace);
 
                     List<Connection.Future> futures = new ArrayList<Connection.Future>(preparedQueries.size());
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -17,7 +17,6 @@ package com.datastax.driver.core;
 
 import com.datastax.driver.core.exceptions.*;
 import com.datastax.driver.core.policies.*;
-import com.datastax.driver.core.ProtocolVersion.ProtocolFeature;
 import com.datastax.driver.core.utils.MoreFutures;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Functions;
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -2290,7 +2290,7 @@ public class Cluster implements Closeable {
                     // Empty string mean no particular keyspace to set
                     // Optimization: Only change keyspace for older protocol versions as newer protocols allow
                     // specifying keyspace on prepared statement.
-                    if (!protocolVersion().supports(ProtocolFeature.KEYSPACE_ON_QUERY) && !keyspace.isEmpty())
+                    if (!ProtocolFeature.KEYSPACE_ON_QUERY.isSupportedBy(protocolVersion()) && !keyspace.isEmpty())
                         connection.setKeyspace(keyspace);
 
                     List<Connection.Future> futures = new ArrayList<Connection.Future>(preparedQueries.size());
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -2277,22 +2277,24 @@ public class Cluster implements Closeable {
                 // used for preparing it. However, since we are likely that all prepared query belong to only a handful
                 // of different keyspace (possibly only one), and to avoid setting the current keyspace more than needed,
                 // we first sort the query per keyspace.
-                SetMultimap<String, PreparedStatement> perKeyspace = HashMultimap.create();
+                SetMultimap<String, String> perKeyspace = HashMultimap.create();
                 for (PreparedStatement ps : preparedQueries.values()) {
                     // It's possible for a query to not have a current keyspace. But since null doesn't work well as
                     // map keys, we use the empty string instead (that is not a valid keyspace name).
                     String keyspace = ps.getQueryKeyspace() == null ? "" : ps.getQueryKeyspace();
-                    perKeyspace.put(keyspace, ps);
+                    perKeyspace.put(keyspace, ps.getQueryString());
                 }
 
                 for (String keyspace : perKeyspace.keySet()) {
                     // Empty string mean no particular keyspace to set
-                    if (!keyspace.isEmpty())
+                    // Optimization: Only change keyspace for older protocol versions as newer protocols allow
+                    // specifying keyspace on prepared statement.
+                    if (protocolVersion().compareTo(ProtocolVersion.V5) >= 0 && !keyspace.isEmpty())
                         connection.setKeyspace(keyspace);
 
                     List<Connection.Future> futures = new ArrayList<Connection.Future>(preparedQueries.size());
-                    for (PreparedStatement prepared : perKeyspace.get(keyspace)) {
-                        futures.add(connection.write(new Requests.Prepare(prepared.getQueryString(), prepared.getKeyspace())));
+                    for (String query : perKeyspace.get(keyspace)) {
+                        futures.add(connection.write(new Requests.Prepare(query, keyspace)));
                     }
                     for (Connection.Future future : futures) {
                         try {
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -2289,7 +2289,7 @@ public class Cluster implements Closeable {
                     // Empty string mean no particular keyspace to set
                     // Optimization: Only change keyspace for older protocol versions as newer protocols allow
                     // specifying keyspace on prepared statement.
-                    if (protocolVersion().compareTo(ProtocolVersion.V5) < 0 && !keyspace.isEmpty())
+                    if (!protocolVersion().supportsKeyspaceOnQuery() && !keyspace.isEmpty())
                         connection.setKeyspace(keyspace);
 
                     List<Connection.Future> futures = new ArrayList<Connection.Future>(preparedQueries.size());
--- a/driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
@@ -190,6 +190,43 @@ public abstract class RegularStatement extends Statement {
     }
 
     /**
+     * Returns the keyspace this query operates on.
+     * <p/>
+     * Unless the keyspace has been explicitly set through {@link #setKeyspace},
+     * this method will return {@code null} to avoid having to parse the query
+     * string.
+     *
+     * @return the keyspace set through {@link #setKeyspace} if such keyspace was
+     * set, {@code null} otherwise.
+     * @see Statement#getKeyspace
+     */
+    @Override
+    public abstract String getKeyspace();
+
+    /**
+     * Sets the keyspace this query operates on.
+     * <p/>
+     * This method allows you to manually provide a keyspace for this query.  It is used for the following:
+     * <p>
+     * <ol>
+     * <li>To indicate to cassandra what keyspace the statement is applicable to (protocol V5+ only).  This is useful
+     * when the query does not provide an explicit keyspace and you want to override the session's keyspace.</li>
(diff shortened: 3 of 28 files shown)


clirr-ignores.xml
```diff
@@ -147,4 +147,19 @@
...
 
+    <difference>
+        <differenceType>7004</differenceType>
+        <className>com/datastax/driver/core/AbstractSession</className>
+        <method>com.google.common.util.concurrent.ListenableFuture prepareAsync(java.lang.String, java.util.Map)</method>
+        <to>com.google.common.util.concurrent.ListenableFuture prepareAsync(java.lang.String, java.lang.String, java.util.Map)</to>
+        <justification>Protected method, only used internally</justification>
+    </difference>
+
+    <difference>
```
tolbertam: Figured this would be ok, but thought i'd point it out just in case.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -2262,12 +2262,12 @@ private Connection prepareAllQueries(Host host, Connection reusedConnection) thr
                 // used for preparing it. However, since we are likely that all prepared query belong to only a handful
                 // of different keyspace (possibly only one), and to avoid setting the current keyspace more than needed,
                 // we first sort the query per keyspace.
-                SetMultimap<String, String> perKeyspace = HashMultimap.create();
+                SetMultimap<String, PreparedStatement> perKeyspace = HashMultimap.create();
```
tolbertam: Could also change this such that if a PreparedStatement has a keyspace set explicitly, prepare it with default session since it doesn't matter what the session keyspace state is in this case.

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
```diff
@@ -190,6 +189,47 @@ public boolean hasValues() {
...
+     * Unless the keyspace has been explicitly set through {@link #setKeyspace},
+     * this method will return {@code null} to avoid having to parse the query
+     * string.
+     *
+     * @return the keyspace set through {@link #setKeyspace} if such keyspace was
+     * set, {@code null} otherwise.
+     * @see Statement#getKeyspace
+     */
+    @Override
+    public String getKeyspace() {
```
tolbertam: Added default implementation to avoid clirr errors, we could also consider simply defining keyspace in `RegularStatement` since I think most implementations do something similar and we can handle the few cases that don't.

driver-core/src/main/java/com/datastax/driver/core/Requests.java
```diff
@@ -240,9 +241,13 @@ public String toString() {
 
         static void serialize(EnumSet<QueryFlag> flags, ByteBuf dest, ProtocolVersion version) {
             int i = 0;
-            for (QueryFlag flag : flags)
-                i |= 1 << flag.ordinal();
-            if (version.compareTo(ProtocolVersion.V5) >= 0) {
+            boolean isV5 = version.compareTo(ProtocolVersion.V5) >= 0;
+            for (QueryFlag flag : flags) {
+                // don't include flag if would make value width greater than int and protocol < 5
+                if (flag.ordinal() < QueryFlag.KEYSPACE.ordinal() || isV5)
```
tolbertam, line 262: One oddity, if you set the keyspace flag and using protocol v4, C* 4.0+ will fail since it will still attempt to parse the keyspace.   Because of this figured i'd add a check here to not add the flag if protocol is less than v5.

driver-core/src/main/java/com/datastax/driver/core/Requests.java
```diff
@@ -569,27 +591,42 @@ public void encode(Prepare msg, ByteBuf dest, ProtocolVersion version) {
...
-                return CBUtil.sizeOfLongString(msg.query);
+                int size = CBUtil.sizeOfLongString(msg.query);
+
+                if (version.compareTo(ProtocolVersion.V5) >= 0 && msg.flags.contains(QueryFlag.KEYSPACE))
+                    size += CBUtil.sizeOfString(msg.keyspace);
+                return size;
             }
         };
 
+        private final EnumSet<QueryFlag> flags = EnumSet.noneOf(QueryFlag.class);
```
tolbertam: Will get rid of this since it's unused.   Originally i used it, but later noticed that QueryFlag.KEYSPACE (0x80) != 0x1 which is what prepare uses for keyspace

driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
```diff
@@ -77,7 +77,7 @@
 
     private final List<ColumnMetadata> partitionKey;
     private final List<Object> routingKeyValues;
-    final String keyspace;
+    String keyspace;
```
tolbertam, line 80: I wonder if we should consider differentiating between keyspace provided with `BuildStatement` (i.e. what comes from `insertInto(keyspace, table)` and otherwise in some way, as it is redundant to send the keyspace with the statement if it's already set in the query.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -2272,12 +2272,14 @@ private Connection prepareAllQueries(Host host, Connection reusedConnection) thr
...
-                    if (!keyspace.isEmpty())
+                    // Optimization: Only change keyspace for older protocol versions as newer protocols allow
+                    // specifying keyspace on prepared statement.
+                    if (protocolVersion().compareTo(ProtocolVersion.V5) < 0 && !keyspace.isEmpty())
                         connection.setKeyspace(keyspace);
 
                     List<Connection.Future> futures = new ArrayList<Connection.Future>(preparedQueries.size());
                     for (String query : perKeyspace.get(keyspace)) {
-                        futures.add(connection.write(new Requests.Prepare(query)));
+                        futures.add(connection.write(new Requests.Prepare(query, keyspace)));
```
tolbertam, line 2297: One thing here, since we can't distinguish whether the keyspace was established via the session keyspace or set on the statement explicitly we always pass it here.  I don't consider this that big of a deal due to the infrequency of prepared statement preparations.

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: java
 jdk:
-- openjdk6
-- oraclejdk7
+- openjdk7
```
tolbertam: not related to pr, but new travis default image (ubuntu trusty) does not have openjdk6 and oraclejdk7 was replaced by openjdk7.   Since we have jenkins and appveyor builds running for java 6 i think it's ok to rely on them for that.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/BuiltStatement.java
tolbertam: Had separate discussion that we were ok with this.

tolbertam:
Removing WIP tag as I consider this in a reviewable state now.  I'd still like to add some integration tests for:

* [x] reprepare (when UNPREPARED returned for node)
* [x] prepare on node up

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -637,7 +637,7 @@ public void onSet(Connection connection, Message.Response response, long latency
                                                 + "Seeing this message a few times is fine, but seeing it a lot may be source of performance problems",
```
tolbertam, line 670: Need to add special case at line 626 to allow case where keyspaces don't match if protocol version >= 5

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
```diff
@@ -190,6 +190,44 @@ public boolean hasValues() {
...
+     * Do note that if the query does not use a fully qualified keyspace, then
+     * you do not need to set the keyspace through this method as the
+     * currently logged in keyspace will be used if it is set.
+     *
+     * @param keyspace the name of the keyspace this query operates on.
+     * @return this {@code SimpleStatement} object.
+     * @see Statement#getKeyspace
+     */
+    public RegularStatement setKeyspace(String keyspace) {
+        throw new UnsupportedOperationException("No concrete implementation of setKeyspace defined for " + this.getClass().getName());
```
olim7t: It sucks that we have to resort to that. I understand that the goal is to avoid a breaking API change, but we're only moving the issue from compile time to run time. On the one hand, it seems reasonable to assume that custom `RegularStatement` implementations should be rare in the field, on the other we could push down the method to each subclass, WDYT?

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -202,7 +202,12 @@ public Cluster getCluster() {
                         switch (rm.kind) {
                             case PREPARED:
                                 Responses.Result.Prepared pmsg = (Responses.Result.Prepared) rm;
-                                PreparedStatement stmt = DefaultPreparedStatement.fromMessage(pmsg, cluster, query, poolsState.keyspace);
+                                String keyspaceToUse = poolsState.keyspace;
+                                if (keyspace != null) {
+                                    // TODO: possibly check protocol version here and if not supported throw an exception.
+                                    keyspaceToUse = keyspace;
```
olim7t: Yes, we should do that check. Otherwise we might end up storing a corrupted prepared statement if you prepare with an old protocol version and a routing keyspace that doesn't match the query string.

driver-core/src/test/java/com/datastax/driver/core/PreparedStatementMultiNodeTest.java
```diff
@@ -0,0 +1,230 @@
...
+    public void executeAfterNodeBroughtBackUp(boolean reprepareOnUp, boolean setKeyspace) {
+        QueryOptions queryOptions = new QueryOptions().setReprepareOnUp(reprepareOnUp).setPrepareOnAllHosts(false);
+        Cluster.Builder builder = createClusterBuilderNoDebouncing()
+                .withNettyOptions(TestUtils.nonQuietClusterCloseOptions)
+                .addContactPointsWithPorts(getContactPointsWithPorts())
+                .withPort(ccm().getBinaryPort())
+                .withQueryOptions(queryOptions)
+                .withLoadBalancingPolicy(new SortingLoadBalancingPolicy());
+
+        // TODO, remove this when V5 is no longer beta.
```
olim7t: You can add an assertion so that we get a reminder:
```
assert ProtocolVersion.NEWEST_BETA == ProtocolVersion.V5
```

driver-core/src/test/java/com/datastax/driver/core/PreparedStatementMultiNodeTest.java
```diff
@@ -0,0 +1,230 @@
...
+                prep2 = session.prepare(new SimpleStatement("SELECT * from users"));
+                session.execute("USE " + keyspace3);
+                prep3 = session.prepare(new SimpleStatement("SELECT * from users"));
+            }
+
+            Host host1 = TestUtils.findHost(cluster, 1);
+            Host host2 = TestUtils.findHost(cluster, 2);
+
+            // Execute queries 10 times, should always hit host1 due to sorting policy.
+            for (int i = 0; i < 10; i++) {
```
olim7t: Why 10 times?

driver-core/src/test/java/com/datastax/driver/core/PreparedStatementMultiNodeTest.java
```diff
@@ -0,0 +1,230 @@
...
+                String.format("INSERT INTO %s.users(id, id2, name) VALUES (2, 2, 'test2')", keyspace2),
+                String.format(TestUtils.CREATE_KEYSPACE_SIMPLE_FORMAT, keyspace3, 2),
+                String.format("CREATE TABLE %s.users(id int, id2 int, name text, primary key (id, id2))", keyspace3),
+                String.format("INSERT INTO %s.users(id, id2, name) VALUES (3, 3, 'test3')", keyspace3)
+        );
+    }
+
+    @Test(groups = "long")
+    @CassandraVersion("4.0.0")
+    public void should_be_able_to_execute_statement_on_restarted_node_reprepare_on_up_set_keyspace() {
```
olim7t: The method names are really hard to understand. "be able to" can generally be removed without losing anything. Also, the fact that reprepare on up is enabled or not is a detail, especially with recent Cassandra version where it will make no difference; the two cases could be done in the same method.

So my suggestion would be:
```java
    @Test(groups = "long")
    @CassandraVersion("4.0.0")
    public void should_execute_statement_with_per_query_keyspace_on_restarted_node() {
        executeAfterNodeBroughtBackUp(true, true);
        executeAfterNodeBroughtBackUp(false, true);
    }

    @Test(groups = "long")
    public void should_execute_statement_with_session_keyspace_on_restarted_node() {
        executeAfterNodeBroughtBackUp(true, false);
        executeAfterNodeBroughtBackUp(false, false);
    }
```

olim7t (changes requested):
Looks really good overall, see a few comments inline.

driver-core/src/test/java/com/datastax/driver/core/PreparedStatementMultiNodeTest.java
tolbertam: arbitrary, just wanted to hit all hosts.  Will reduce.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
tolbertam: will do

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
tolbertam: i had kind of waffled back and forth on that.  I ultimately did this to avoid adding a clirr exception, but I do like the idea of making it abstract to force implementation.  Will change it.

olim7t: To be clear, I think we must avoid the Clirr exception. By "push down" I meant removing it entirely from the abstract class, not making it abstract.

tolbertam:
Thanks for reviewing!  I'll process your feedback and push a commit either later today or monday morning.

driver-core/src/test/java/com/datastax/driver/core/PreparedStatementMultiNodeTest.java
olim7t: But doesn't the sorting policy guarantee that you'll always hit host1?

tolbertam: ah that's right.  I'm probably using the policy incorrectly then (or not, i need to revisit the tests). I'll look into adjusting one way or another.

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
tolbertam: I see what you are saying now, that makes complete sense to me.  `setKeyspace` is not really something used within the driver itself (afaict), so there's really no need to be part of `RegularStatement`.  Instead it is only used by client APIs, who have a more concrete implementation, like `BuiltStatement` for example.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
tolbertam: One thing about this is that for this scenario to happen, the statement has to be successfully prepared by the server, which means the table and query criteria etc have to match for this query for the target keyspace, but not necessarily completely, so I agree we should still have this check.

tolbertam: Just had a thought, we could handle this better by checking the poolState before sending the request to the server to ensure the keyspaces match ahead of time.  I think we should still check here too just in case for whatever reason the pool state changes during the request (which is unlikely).

tolbertam: we could do this in general (not just for prepared statements) so at least we get the error before the query is sent and with a more appropriate error.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -149,12 +149,25 @@ public void run() {
...
+        Requests.Prepare request = new Requests.Prepare(query, keyspace);
         request.setCustomPayload(customPayload);
         Connection.Future future = new Connection.Future(request);
         execute(future, Statement.DEFAULT);
-        return toPreparedStatement(query, future);
+        return toPreparedStatement(query, keyspace, future);
+    }
+
+    private void checkCanSetKeyspace(String keyspace) {
+        if (keyspace != null && !keyspace.equals(poolsState.keyspace) && cluster.manager.protocolVersion().compareTo(ProtocolVersion.V5) < 0) {
```
tolbertam: Should I remove the condition that checks if the keyspaces are different?  Maybe we should also throw an exception when using `setKeyspace` at < V5?

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -202,7 +215,16 @@ public Cluster getCluster() {
                         switch (rm.kind) {
                             case PREPARED:
                                 Responses.Result.Prepared pmsg = (Responses.Result.Prepared) rm;
-                                PreparedStatement stmt = DefaultPreparedStatement.fromMessage(pmsg, cluster, query, poolsState.keyspace);
+                                String keyspaceToUse = poolsState.keyspace;
+                                if (keyspace != null && !keyspace.equals(keyspaceToUse)) {
```
tolbertam, line 216: This seems very unlikely now that we check before sending the query.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
tolbertam: Just remembered that this is imperfect because it's possible `setKeyspace` was used for something else (like routing).  I think i'll remove this check then

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
olim7t: 👍  for the change.

As to:
> we could handle this better by checking the poolState before sending the request to the server to ensure the keyspaces match ahead of time

It's not possible if the query string uses a qualified table name, since we don't parse it.
```
session.execute("USE ks1");
// session's keyspace does not match the statement's keyspace, yet the query is valid
session.execute(new SimpleStatement("SELECT * FROM ks2.bar").setKeyspace(ks2));
```

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
olim7t: Yes I think that's in line with my previous comment about qualified table names.

olim7t (approved):

olim7t (approved):
I made ProtocolFeature top-level.
Looks good to me.

tolbertam:
Added 6e66730 and rebased on 3.x.  Leaving commits in tact for now (can always squash on merge)

driver-core/src/main/java/com/datastax/driver/core/Requests.java
newkek: Should we open a C* ticket about it? Also maybe we can add a warning here in case the user sets a keyspace and expects it to work, but for some reason (let's say protocol negotiation) the driver ends up using protocol V4.

newkek (commented):
This looks like an awesome feature. Should we offer a setKeyspace() option on `QueryOptions` so that people can completely remove their "USE keyspace" in their application and not have to worry about setting the keyspace on each statement?

driver-core/src/main/java/com/datastax/driver/core/Requests.java
tolbertam: I didn't realize this at first, but apparently this is the way cassandra handles all query flags: 

https://github.com/apache/cassandra/blob/6bea7de50418a7a7b9e2b8dcf2446701d9cc5b69/src/java/org/apache/cassandra/cql3/QueryOptions.java#L442

If a flag is present, it will attempt to parse it, regardless of protocol version.  I think it leaves it up to the driver to behave appropriately and only set flags and values that are supported by the protocol version in use.  Not sure if that should be considered a bug, and if it was I doubt it would be fixed.

The main issue is we can't really determine whether or not the keyspace was set for routing, or to be used on the query (since they use the same method `setKeyspace`).  With other query features, we can preemptively throw an error if a user tries using a feature that isn't supported by their protocol version.   Because we can't tell why `setKeyspace` was used I think adding a warning might be more annoying for users than not.  Having queries fail is just as good of a solution I think.

tolbertam:
> Should we offer a setKeyspace() option on QueryOptions so that people can completely remove their "USE keyspace" in their application and not have to worry about setting the keyspace on each statement?

That's a good idea and I did consider it but I ultimately decided against it because while `"USE keyspace"` is really troublesome for us to deal with internally, it's more of a technical detail than a user level detail.  Allowing setting keyspace at the `Session` level and globally in `QueryOptions` is redundant at the user level.   However, I think with Execution Profiles in java driver 4, such capability will be nice to have since you can say "profile A uses this keyspace, profile B uses that keyspace" so we should consider that for version 4.

newkek:
True it's an internal simplification, and indeed the Execution profile use case would make life easier in multi-tenancy situations for exmaple. The only reason I would see the keyspace `QueryOptions` being added is if we make it available now and start recommending it loudly in docs, in order to ease the transition for later. In terms of API it also feels strange to have to execute a query to set a global configuration, but have a programmatic way to set something "locally" (i.e. to a statement), like for most of the options we have on `Statement` it would have been consistent to have both settings available in the statement and `QueryOptions`. But I don't know it's maybe not a priority now.

driver-core/src/main/java/com/datastax/driver/core/Requests.java
newkek: > Having queries fail is just as good of a solution I think.

Wdym having queries fail? It sounded like here the keyspace would just be ignored in the query flags if protocol_version < 5. Would the driver fail if protocol < 5 and the keyspace is set or you did you mean the query would fail because supposedly the keyspace wouldn't exist and cassandra would error out?

tolbertam: For < protocol V5, the keyspace flag (and thus keyspace) isn't set at all.  So the query would be executed and fail since the wrong keyspace is used (unless there is an equivalent enough schema).

newkek (approved):
Good to merge imo. I would rather add a setting on `QueryOptions` for consistency and to anticipate for next versions but waiting for 4.0 to do it also makes sense.

tolbertam:
Rebased this on 3.x since it was very behind.  Also reran a bunch of tests against latest C* trunk and everything checked out.

tolbertam:
This basically needs to be completely re-evaluated.  It's been a while since the work was done and we've changed a bunch since then.  We'll want it for C* 4.0, so we should keep this on our radar.

tolbertam:
Closing this for now since there are quite a bit of conflicts and this needs to be rebased.  This is already implemented in the 4.x driver, so we may opt to not implement it in 3.x since it's a involved change.  We may revisit if there is demand for it.

PR #877, opened 2017-09-15
JAVA-1619: Update QueryBuilder.in, lt, lte, eq, gt, & gte to support Iterable input

tolbertam:
For [JAVA-1619](https://datastax-oss.atlassian.net/browse/JAVA-1619)

### DIFF
--- a/clirr-ignores.xml
+++ b/clirr-ignores.xml
@@ -147,4 +147,16 @@
         <justification>False positive, the enclosing class is package-private so this was never exposed</justification>
     </difference>
 
+    <difference>
+        <differenceType>7005</differenceType>
+        <className>com/datastax/driver/core/querybuilder/QueryBuilder</className>
+        <!--
+            A Clirr bug prevents us from excluding the methods individually (errors re-appear as soon as there are more
+            than one method). Excluding every method is too coarse, but it's the only solution.
+        -->
+        <method>*</method>
+        <to>*</to>
+        <justification>Relaxed parameters from List to Iterable for in, lt, lte, eq, gt, and gte</justification>
+    </difference>
+
 </differences>
--- a/driver-core/src/main/java/com/datastax/driver/core/querybuilder/Clause.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/querybuilder/Clause.java
@@ -16,7 +16,7 @@
 package com.datastax.driver.core.querybuilder;
 
 import com.datastax.driver.core.CodecRegistry;
-
+import com.google.common.collect.Lists;
 import java.util.List;
 
 public abstract class Clause extends Utils.Appendeable {
@@ -70,13 +70,14 @@ public abstract class Clause extends Utils.Appendeable {
 
         private final List<?> values;
 
-        InClause(String name, List<?> values) {
+        InClause(String name, Iterable<?> values) {
             super(name);
-            this.values = values;
-
             if (values == null)
                 throw new IllegalArgumentException("Missing values for IN clause");
-            if (values.size() > 65535)
+
+            this.values = Lists.newArrayList(values);
+
+            if (this.values.size() > 65535)
                 throw new IllegalArgumentException("Too many values for IN clause, the maximum allowed is 65535");
         }
 
@@ -91,7 +92,7 @@ public abstract class Clause extends Utils.Appendeable {
             //    ... IN ? ...
             // which binds the variable to the full list the IN is on.
             if (values.size() == 1 && values.get(0) instanceof BindMarker) {
-                Utils.appendName(name, sb).append(" IN ").append(values.get(0));
+                Utils.appendName(name, sb).append(" IN ").append(values.iterator().next());
                 return;
             }
 
@@ -178,11 +179,12 @@ public abstract class Clause extends Utils.Appendeable {
         private final List<String> names;
         private final List<?> values;
 
-        CompoundClause(List<String> names, String op, List<?> values) {
-            assert names.size() == values.size();
+        CompoundClause(Iterable<String> names, String op, Iterable<?> values) {
             this.op = op;
-            this.names = names;
-            this.values = values;
+            this.names = Lists.newArrayList(names);
+            this.values = Lists.newArrayList(values);
+            if (this.names.size() != this.values.size())
+                throw new IllegalArgumentException(String.format("The number of names (%d) and values (%d) don't match", this.names.size(), this.values.size()));
         }
 
         @Override
@@ -227,25 +229,32 @@ public abstract class Clause extends Utils.Appendeable {
 
     static class CompoundInClause extends Clause {
         private final List<String> names;
-        private final List<?> valueLists;
+        private final List<Object> valueLists;
 
-        public CompoundInClause(List<String> names, List<?> valueLists) {
+        public CompoundInClause(Iterable<String> names, Iterable<?> valueLists) {
             if (valueLists == null)
                 throw new IllegalArgumentException("Missing values for IN clause");
-            if (valueLists.size() > 65535)
-                throw new IllegalArgumentException("Too many values for IN clause, the maximum allowed is 65535");
+            if (names == null)
+                throw new IllegalArgumentException("Missing names for IN clause");
+
+            this.names = Lists.newArrayList(names);
+            this.valueLists = Lists.newArrayList();
+
             for (Object value : valueLists) {
-                if (value instanceof List) {
-                    List<?> tuple = (List<?>) value;
-                    if (tuple.size() != names.size()) {
-                        throw new IllegalArgumentException(String.format("The number of names (%d) and values (%d) don't match", names.size(), tuple.size()));
+                if (value instanceof Iterable) {
+                    List<?> tuple = Lists.newArrayList((Iterable<?>) value);
+                    if (tuple.size() != this.names.size()) {
+                        throw new IllegalArgumentException(String.format("The number of names (%d) and values (%d) don't match", this.names.size(), tuple.size()));
                     }
+                    this.valueLists.add(tuple);
                 } else if (!(value instanceof BindMarker)) {
                     throw new IllegalArgumentException(String.format("Wrong element type for values list, expected List or BindMarker, got %s", value.getClass().getName()));
+                } else {
+                    this.valueLists.add(value);
                 }
             }
-            this.names = names;
-            this.valueLists = valueLists;
+            if (this.valueLists.size() > 65535)
+                throw new IllegalArgumentException("Too many values for IN clause, the maximum allowed is 65535");
         }
 
         @Override
--- a/driver-core/src/main/java/com/datastax/driver/core/querybuilder/QueryBuilder.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/querybuilder/QueryBuilder.java
@@ -313,10 +313,7 @@ public final class QueryBuilder {
      * @return the corresponding where clause.
      * @throws IllegalArgumentException if {@code names.size() != values.size()}.
      */
-    public static Clause eq(List<String> names, List<?> values) {
-        if (names.size() != values.size())
-            throw new IllegalArgumentException(String.format("The number of names (%d) and values (%d) don't match", names.size(), values.size()));
-
+    public static Clause eq(Iterable<String> names, Iterable<?> values) {
         return new Clause.CompoundClause(names, "=", values);
     }
 
@@ -351,7 +348,7 @@ public final class QueryBuilder {
      * @param values the values
      * @return the corresponding where clause.
      */
-    public static Clause in(String name, List<?> values) {
+    public static Clause in(String name, Iterable<?> values) {
         return new Clause.InClause(name, values);
     }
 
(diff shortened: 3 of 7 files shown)


driver-core/src/main/java/com/datastax/driver/core/querybuilder/QueryBuilder.java
```diff
@@ -351,6 +351,19 @@ public static Clause in(String name, Object... values) {
      * @param values the values
      * @return the corresponding where clause.
      */
+    public static Clause in(String name, Collection<?> values) {
+        return new Clause.InClause(name, values);
+    }
+
+    /**
```
tolbertam: I considered removing this method, but per the binary compatibility guide [13.4.14](https://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.14):

> Changing [...] the type of a formal parameter [...] creates a method or constructor with a new signature, and has the combined effect of deleting the method or constructor with the old signature and adding a method or constructor with the new signature (§13.4.12).

Clirr throws an error as well, so figured would just deprecate it instead.

tolbertam:
Will change this to take `Iterable` since no reason why that couldn't work as well (unless it is non-terminating).

tolbertam:
On second thought.  We have validation to check that the size of the collection is less than 65535 elements.  Using `Iterable` would require iterating over that many elements, which may consume whatever you are iterating over.  It might be better to keep it at `Collection` for that reason.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/QueryBuilder.java
```diff
@@ -351,6 +351,19 @@ public static Clause in(String name, Object... values) {
...
+     * Create an "in" {@code WHERE} clause stating the provided column must be equal
+     * to one of the provided values.
+     *
+     * @param name   the column name
+     * @param values the values
+     * @return the corresponding where clause.
+     * @deprecated Use {@link #in(String, Collection)} instead.
+     */
     public static Clause in(String name, List<?> values) {
```
newkek: I think this needs the `@Deprecated` annotation too if we deprecate the method.

newkek: Kind of annoying though to deprecate because people using a list are going to have to manually cast their list to `Collection` in order to not use a deprecated method and avoid a warning. Imo we could remove it and add a clirr exception since the `QueryBuilder` class is final and therefore should not be extended. It would also not impact users just using because List is a child of Collection.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/QueryBuilder.java
```diff
@@ -351,6 +351,19 @@ public static Clause in(String name, Object... values) {
      * @param values the values
      * @return the corresponding where clause.
      */
+    public static Clause in(String name, Collection<?> values) {
```
newkek: Should we also modify/overload the method `QueryBuilder.in(List<String> names, List<?> values)`?

newkek: As a side note, I wonder why the wildcards are necessary in those methods parameters for our use case (they are in many other methods parameters of the query builder). Was looking around and couldn't find a reason why we can't just put `Collection` instead of `Collection<?>`, but maybe it's a Java < 8 thing, or a best practice / usual things to do that I'm not aware of (happy to be explained then).

newkek (commented):
Looking good but I had a few suggestions/questions

driver-core/src/main/java/com/datastax/driver/core/querybuilder/QueryBuilder.java
tolbertam: that's a cool idea.  I'll do that too

tolbertam: It's more of a best practice than anything else, I think they are functionally the same.  I believe some IDEs may warn when using collections with no generic type.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/QueryBuilder.java
tolbertam: > I think this needs the @Deprecated annotation too if we deprecate the method.

D'oh, good catch.  

> Kind of annoying though to deprecate because people using a list are going to have to manually cast their list to Collection in order to not use a deprecated method and avoid a warning. Imo we could remove it and add a clirr exception since the QueryBuilder class is final and therefore should not be extended. It would also not impact users just using because List is a child of Collection.

I think getting a compiler warning is a better tradeoff than possibly breaking user code.  In general case it won't, but if a user is using reflection or some other high level language there could be unknown consequences of changing the method signature.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/QueryBuilder.java
tolbertam: Hrmm, there are so many places to do that.  Also for gt, lt, lte, gte.  I'm starting to think that maybe the deprecated route is not the best route since we'll create a lot of duplicate methods.  Since this change is to fix a minor inconvenience, maybe it's best to wait until java driver 4,  before adjusting the signatures, wdyt?

olim7t: A long time ago, before Java had generics, all collections were raw. You would receive a collection in a method and cast the elements out of it. You could pass a collection of the wrong type and it would only fail at runtime. It sucked.
Nowadays if you use `Collection` in your API, the client is left wondering: did you live in a cave since Java 1.4 or is it an oversight? Will the method try to cast elements to a particular type?
If you use `Collection<?>`, you clearly document the fact that any element type works.
Admittedly this is a subtle difference and it's all the same after erasure, but I think clear contracts matter.

olim7t:
> On second thought. We have validation to check that the size of the collection is less than 65535 elements. Using Iterable would require iterating over that many elements, which may consume whatever you are iterating over. It might be better to keep it at Collection for that reason.

Given how much of an edge case that is, I think we can use `Iterables.size` and be fine.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/QueryBuilder.java
newkek: I'd vote to ~~modify~~ take care of the current one and the `QueryBuilder.in(List<String> names, List<?> values)` to be consistent, and the rest later indeed wdyt?

newkek: > A long time ago,

in a galaxy far far away...

> it's all the same after erasure

Yes, especially because in the rest of the codebase of the query builder we just declare the objects that we get from this collection as `Object`s. 

Although, the idea of expressing explicitly that you accept objects from any type, instead of leaving it empty and leaving the uncertainty of not knowing whether the fact that there's no type means that "any type" would work, or it means that it expects one specific type but didn't mention it makes sense. Thanks for the clarifications.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/QueryBuilder.java
newkek: Hmm I can't say about higher level tools that use reflection and so on on which it would potentially break so I don't know how common it is. However deprecating would lead _every_ existing user to use a deprecated method and get a warning, and lead all of them to do that `(Collection<?>)` casting to avoid it. I'd rather affect less people but would be ok to go the "deprecation" way

tolbertam:
Based on feedback I am making the following changes:

1. Use `Iterable` instead of `Collection`
2. Update all of the other methods in `QueryBuilder` that take lists to take `Iterable` instead (i.e. `lte`, `lt`, `gt`, `gte`, etc.).
3. Add clirr exceptions for everything.
4. Add upgrade guide entry.

tolbertam:
> Given how much of an edge case that is, I think we can use Iterables.size and be fine.

`Iterables.size` consumes the the `Iterable` so to handle this I copy the input into an `ArrayList` beforehand.

tolbertam:
Addressed feedback and rebased.

clirr-ignores.xml
```diff
@@ -147,4 +147,12 @@
         <justification>False positive, the enclosing class is package-private so this was never exposed</justification>
     </difference>
 
+    <difference>
+        <differenceType>7005</differenceType>
+        <className>com/datastax/driver/core/querybuilder/QueryBuilder</className>
+        <method>*</method>
```
tolbertam: Unfortunately, having such a broad exception is necessary otherwise clirr repeats errors in `QueryBuilder` over and over:

```
[ERROR] 7005: com.datastax.driver.core.querybuilder.QueryBuilder: Parameter 1 of 'public com.datastax.driver.core.querybuilder.Clause gt(java.util.List, java.util.List)' has changed its type to java.lang.Iterable
[ERROR] 7005: com.datastax.driver.core.querybuilder.QueryBuilder: Parameter 2 of 'public com.datastax.driver.core.querybuilder.Clause gt(java.util.List, java.util.List)' has changed its type to java.lang.Iterable
[ERROR] 7005: com.datastax.driver.core.querybuilder.QueryBuilder: Parameter 1 of 'public com.datastax.driver.core.querybuilder.Clause gt(java.util.List, java.util.List)' has changed its type to java.lang.Iterable
[ERROR] 7005: com.datastax.driver.core.querybuilder.QueryBuilder: Parameter 2 of 'public com.datastax.driver.core.querybuilder.Clause gt(java.util.List, java.util.List)' has changed its type to java.lang.Iterable
[ERROR] 7005: com.datastax.driver.core.querybuilder.QueryBuilder: Parameter 1 of 'public com.datastax.driver.core.querybuilder.Clause gte(java.util.List, java.util.List)' has changed its type to java.lang.Iterable
[ERROR] 7005: com.datastax.driver.core.querybuilder.QueryBuilder: Parameter 2 of 'public com.datastax.driver.core.querybuilder.Clause gte(java.util.List, java.util.List)' has changed its type to java.lang.Iterable
[ERROR] 7005: com.datastax.driver.core.querybuilder.QueryBuilder: Parameter 1 of 'public com.datastax.driver.core.querybuilder.Clause gt(java.util.List, java.util.List)' has changed its type to java.lang.Iterable
[ERROR] 7005: com.datastax.driver.core.querybuilder.QueryBuilder: Parameter 2 of 'public com.datastax.driver.core.querybuilder.Clause gt(java.util.List, java.util.List)' has changed its type to java.lang.Iterable
[ERROR] 7005: com.datastax.driver.core.querybuilder.QueryBuilder: Parameter 1 of 'public com.datastax.driver.core.querybuilder.Clause gte(java.util.List, java.util.List)' has changed its type to java.lang.Iterable
[ERROR] 7005: com.datastax.driver.core.querybuilder.QueryBuilder: Parameter 2 of 'public com.datastax.driver.core.querybuilder.Clause gte(java.util.List, java.util.List)' has changed its type to java.lang.Iterable
[ERROR] 7005: com.datastax.driver.core.querybuilder.QueryBuilder: Parameter 1 of 'public com.datastax.driver.core.querybuilder.Clause gt(java.util.List, java.util.List)' has changed its type to java.lang.Iterable
[ERROR] 7005: com.datastax.driver.core.querybuilder.QueryBuilder: Parameter 2 of 'public com.datastax.driver.core.querybuilder.Clause gt(java.util.List, java.util.List)' has changed its type to java.lang.Iterable
[ERROR] 7005: com.datastax.driver.core.querybuilder.QueryBuilder: Parameter 1 of 'public com.datastax.driver.core.querybuilder.Clause gte(java.util.List, java.util.List)' has changed its type to java.lang.Iterable
[ERROR] 7005: com.datastax.driver.core.querybuilder.QueryBuilder: Parameter 2 of 'public com.datastax.driver.core.querybuilder.Clause gte(java.util.List, java.util.List)' has changed its type to java.lang.Iterable
```

While probably too coarse grained, I don't imagine this becoming much of an issue as I doubt we will make much changes in `QueryBuilder` on the 3.x branch in the future.

olim7t:
> Iterables.size consumes the the Iterable 

Mmm no it doesn't:
```java
ImmutableList<Integer> list = ImmutableList.of(1, 2, 3);
System.out.println(Iterables.size(list)); // 3
System.out.println(Iterables.size(list)); // 3
```
Except if the iterable returns a consuming iterator, but that's an edge case so we can let the caller handle that.

driver-core/src/main/java/com/datastax/driver/core/querybuilder/Clause.java
```diff
@@ -227,25 +229,32 @@ void appendTo(StringBuilder sb, List<Object> variables, CodecRegistry codecRegis
...
+        public CompoundInClause(Iterable<String> names, Iterable<?> valueLists) {
             if (valueLists == null)
                 throw new IllegalArgumentException("Missing values for IN clause");
-            if (valueLists.size() > 65535)
-                throw new IllegalArgumentException("Too many values for IN clause, the maximum allowed is 65535");
+            if (names == null)
+                throw new IllegalArgumentException("Missing names for IN clause");
+
+            this.names = Lists.newArrayList(names);
+            this.valueLists = Lists.newArrayListWithExpectedSize(this.names.size());
```
olim7t: Nit: should be `valueLists.size()`. The size of the list of tuples is not related to the number of values, for example `(a,b) IN ((1,2),(3,4),(5,6))` => 2 values, 3 tuples.
I'm taking care of this while merging.

olim7t (approved):

PR #878, opened 2017-09-18
Fix number of executions parameter to match documented behavior.

tolbertam:
The documentation shows the behavior for 3 executions, but only 2 were configured.


tolbertam:
This change is technically wrong.  This is probably the third or fourth time I've been confused about whether the number of executions includes the first, or if they are additional beyond the first.

PR #879, opened 2017-09-19
JAVA-1493: Handle schema metadata

olim7t:
Note to reviewers: this is a big pull request! I've kept the commits separate for now, but they will be squashed eventually (I'm not sure if it's better to review them individually or as a whole).

The main point of entry is `MetadataManager.refreshSchema`. Unlike 3.x, we only do full schema refreshes. The schema can be restricted to a subset of keyspaces. Refreshes can be enabled via the config and programmatically, and triggered manually.

The implementation is split in 3 phases (see subpackages in `internal.metadata.schema`):
1. *querying* the appropriate system tables (asynchronously and with paging)
2. *parsing* the resulting rows into `XxxMetadata` objects
3. *refreshing* the metadata, that is, publishing a copy with the new schema (reminder: metadata is immutable), and computing the events to send.


olim7t:
There are docs for this on the `metadata_docs` branch if that helps.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/schema/AggregateMetadata.java
```diff
@@ -0,0 +1,126 @@
...
+
+  @Override
+  default String describeWithChildren(boolean pretty) {
+    // An aggregate has no children
+    return describe(pretty);
+  }
+
+  @Override
+  default String describe(boolean pretty) {
+    ScriptBuilder builder = new ScriptBuilder(pretty);
```
tolbertam, line 89: 👍 nice concept to push away all the common behaviors around pretty printing.

core/src/main/java/com/datastax/oss/driver/api/core/type/DataType.java
```diff
@@ -26,4 +27,16 @@
...
+   * Builds an appropriate representation for use in a CQL query.
+   *
+   * @param includeFrozen whether to include the {@code frozen<...>} keyword if applicable. This
+   *     will need to be set depending on where the result is used: for example, {@code CREATE
+   *     TABLE} statements use the frozen keyword, whereas it should never appear in {@code CREATE
+   *     FUNCTION}.
+   * @param pretty whether to pretty-print UDT names (as described in {@link
+   *     CqlIdentifier#asCql(boolean)}.
+   */
+  String asCql(boolean includeFrozen, boolean pretty);
```
tolbertam, line 41: Well documented 👍 .  When I initially saw `includeFrozen` in `CustomType` i scratched my head a bit, but now I understand why (for UDFs)

core/src/main/java/com/datastax/oss/driver/api/core/metadata/schema/RelationMetadata.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+package com.datastax.oss.driver.api.core.metadata.schema;
+
+import com.datastax.oss.driver.api.core.CqlIdentifier;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+/** A table or materialized view in the schema metadata. */
+public interface RelationMetadata extends Describable {
```
tolbertam, line 24: Better name than `TableOrView` for sure 👍

core/src/main/java/com/datastax/oss/driver/api/core/CassandraVersion.java
```diff
@@ -38,6 +38,10 @@
       "(\\d+)\\.(\\d+)(\\.\\d+)?(\\.\\d+)?([~\\-]\\w[.\\w]*(?:\\-\\w[.\\w]*)*)?(\\+[.\\w]+)?";
   private static final Pattern pattern = Pattern.compile(VERSION_REGEXP);
 
+  public static final CassandraVersion V2_1_0 = parse("2.1.0");
```
tolbertam, line 41: Thank you for doing this 😄 , few places in tests where was constantly recreating same versions over and over.

core/src/main/java/com/datastax/oss/driver/api/core/Cluster.java
```diff
@@ -68,6 +69,44 @@ static String getDriverVersion() {
...
+  void setSchemaMetadataEnabled(Boolean newValue);
+
+  /**
+   * Force an immediate refresh of the schema metadata, even if it is currently disabled (either in
+   * the configuration or via {@link #setSchemaMetadataEnabled(Boolean)}).
+   *
+   * <p>The new metadata is returned in the resulting future (and will also be reflected by {@link
+   * #getMetadata()} when that future completes).
+   */
+  CompletionStage<Metadata> refreshSchemaAsync();
```
tolbertam, line 107: Love that you can do this manually now :)

core/src/main/java/com/datastax/oss/driver/api/core/Cluster.java
```diff
@@ -68,6 +69,44 @@ static String getDriverVersion() {
...
+   * statements.
+   *
+   * <p>If calling this method re-enables the metadata (that is, {@link #isSchemaMetadataEnabled()}
+   * was false before, and becomes true as a result of the call), a refresh is also triggered.
+   *
+   * @param newValue a boolean value to enable or disable schema metadata programmatically, or
+   *     {@code null} to use the driver's configuration.
+   * @see CoreDriverOption#METADATA_SCHEMA_ENABLED
+   */
+  void setSchemaMetadataEnabled(Boolean newValue);
```
tolbertam: What is the interaction like with a dynamic config?  I.E. if someone changes `METADATA_SCHEMA_ENABLED` afterwords does it override whatever was done programatically?  (Edit:  I think i got the answer from [this](https://github.com/datastax/java-driver/pull/879#discussion_r144078738)).

tolbertam: Should this also return `CompletionStage<Metadata>` so the user can track the completion of the refresh?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
```diff
@@ -115,10 +159,40 @@ public void removeNode(InetSocketAddress address) {
     RunOrSchedule.on(adminExecutor, () -> singleThreaded.removeNode(address));
   }
 
-  public void refreshSchema(
-      SchemaElementKind kind, String keyspace, String object, List<String> arguments) {
-    // TODO refresh schema metadata, only complete the future once it's done
-    singleThreaded.firstSchemaRefreshFuture.complete(null);
+  /**
+   * @param keyspace if this refresh was triggered by an event, that event's keyspace, otherwise
```
tolbertam, line 161: Would be good to clarify that if null it refreshes everything.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
```diff
@@ -115,10 +159,40 @@ public void removeNode(InetSocketAddress address) {
...
+      RunOrSchedule.on(adminExecutor, () -> singleThreaded.acceptSchemaRequest(future, flushNow));
+    } else {
+      future.complete(metadata);
+      singleThreaded.firstSchemaRefreshFuture.complete(null);
+    }
+    return future;
+  }
+
+  public boolean isSchemaEnabled() {
+    return (schemaEnabledProgrammatically != null)
```
tolbertam, line 178: I think this answers one of my previous questions, if schema is enabled programmatically it overrides the config completely (even if the config changes?)

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
```diff
@@ -190,7 +283,96 @@ private void addNode(InetSocketAddress address, Optional<NodeInfo> maybeInfo) {
...
+    }
+
+    // The debouncer has flushed, start the actual work.
+    private void startSchemaRequest(CompletableFuture<Metadata> future) {
+      assert adminExecutor.inEventLoop();
+      if (closeWasCalled) {
+        future.completeExceptionally(new IllegalStateException("Cluster is closed"));
+        return;
+      }
+      if (currentSchemaRefresh == null) {
```
tolbertam: Nice, it's a lot easier to reconcile the current / queued schema refresh coordination knowing that everything is done in the same thread.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/AggregateParser.java
```diff
@@ -0,0 +1,121 @@
...
+  AggregateParser(DataTypeParser dataTypeParser, InternalDriverContext context) {
+    this.dataTypeParser = dataTypeParser;
+    this.context = context;
+  }
+
+  AggregateMetadata parseAggregate(
+      AdminRow row,
+      CqlIdentifier keyspaceId,
+      Map<CqlIdentifier, UserDefinedType> userDefinedTypes) {
+    // Cassandra < 3.0:
```
tolbertam, line 45: Great idea to capture this here for reference 👍

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/AggregateParser.java
```diff
@@ -0,0 +1,121 @@
...
+    FunctionSignature finalFuncSignature =
+        (finalFuncSimpleName == null)
+            ? null
+            : new FunctionSignature(CqlIdentifier.fromInternal(finalFuncSimpleName), stateType);
+
+    DataType returnType =
+        dataTypeParser.parse(keyspaceId, row.getString("return_type"), userDefinedTypes, context);
+
+    Object initCond;
+    if (row.isString("initcond")) { // Cassandra 3
```
tolbertam, line 101: 👍 on checking the types instead of doing explicit version checks.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/DataTypeClassNameCompositeParser.java
```diff
@@ -0,0 +1,99 @@
...
+import com.datastax.oss.driver.api.core.type.DataType;
+import com.datastax.oss.driver.api.core.type.UserDefinedType;
+import com.datastax.oss.driver.internal.core.context.InternalDriverContext;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+class DataTypeClassNameCompositeParser extends DataTypeClassNameParser {
```
tolbertam, line 28: Kind of a downer that we need to support both the class and type name parsing, but that's the consequence of supporting < C* 3.0 :)

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/Cassandra21SchemaQueries.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+package com.datastax.oss.driver.internal.core.metadata.schema.queries;
+
+import com.datastax.oss.driver.api.core.config.DriverConfigProfile;
+import com.datastax.oss.driver.api.core.metadata.Metadata;
+import com.datastax.oss.driver.internal.core.channel.DriverChannel;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+class Cassandra21SchemaQueries extends SchemaQueries {
```
tolbertam, line 24: 👍  I like how the different queries are broken up among classes.

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/AggregateParserTest.java
```diff
@@ -0,0 +1,110 @@
...
+          "0");
+
+  @Before
+  public void setup() {
+    Mockito.when(context.codecRegistry()).thenReturn(new DefaultCodecRegistry("test"));
+    Mockito.when(context.protocolVersion()).thenReturn(ProtocolVersion.DEFAULT);
+  }
+
+  @Test
+  public void should_parse_modern_table() {
```
tolbertam, line 64: Great to have these kinds of tests at unit level

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SchemaChangesIT.java
```diff
@@ -0,0 +1,554 @@
...
+                verifyNewMetadata.accept(extract.apply(cluster2.getMetadata()));
+                verifyListener.accept(listener2, oldElement, newElement);
+              })
+          .becomesTrue();
+    }
+  }
+
+  // Some element types don't have an ALTER command, but we can still observe an update if they get
+  // dropped and recreated while schema metadata is disabled
+  private <T> void should_handle_update_via_drop_and_recreate(
```
tolbertam, line 534: Nice, I like how you consolidated the logic here so it isn't repeated all over.

tolbertam (commented):
This looks really great so far.  I skimmed over some classes that I assume were partially ported over from 3.x, but have gone through all of it otherwise.  I'll work on adding some more integration level tests, and if all goes well i should be able to approve this today or tomorrow morning.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/schema/RelationMetadata.java
olim7t: I'm still not 100% happy with it because it kind of rings like "relational db", but that's the best I found so far.

core/src/main/java/com/datastax/oss/driver/api/core/Cluster.java
olim7t: re interaction: yes, the programmatic approach always takes precedence over the configuration.

re returning CompletionStage: yes, that's a good idea, I'll change it.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
olim7t: Actually it always refreshes everything. `keyspace` is only used to ignore events for keyspaces that are excluded via the config.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
olim7t: Yes, however long the current refresh is taking, we know that at most one other refresh gets queued up. Whereas in 3.x multiple refreshes could wait on the metadata lock (if things went really bad and refreshes started taking longer than the debouncing interval).

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/AggregateParser.java
olim7t: I tried to reference versions as little as possible in the parsing code.

tolbertam:
It looks like the topological sort logic isn't working for UDTs in <= 2.2:

```
CREATE TYPE ks_0.atype (
    c frozen<ks_0.ctype>
);

CREATE TYPE ks_0.btype (
    a text
);

CREATE TYPE ks_0.ctype (
    z frozen<ks_0.ztype>,
    x frozen<ks_0.xtype>
);

CREATE TYPE ks_0.xtype (
    d text
);

CREATE TYPE ks_0.ztype (
    c text,
    a int
);
```

It does on the other hand work for 3.0+.

tolbertam:
I see the problem, `UserDefinedTypeParser` only checks for `ShallowUserDefinedType`, not `DefaultUserDefinedType`s which they can be with 2.x.  I'll try adding that and see if it helps.

tolbertam:
That seems to fix the ordering, the only problem that remains is that the UserType fields are not marked as frozen in 2.x, should be easy to fix:

```
CREATE TYPE ks_0.ctype (
    z ks_0.ztype,
    x ks_0.xtype
);

CREATE TYPE ks_0.atype (
    c ks_0.ctype
);
```

tolbertam:
w/r/t udt type fields not being frozen - it was an easy fix.  Will push my change in a little bit.

tolbertam:
I made some changes to make the `describe` output more consistent with cqlsh, however there are a few small differences between this implementation and 3.x that I think are acceptable:

1. The order of table/view options is ordered alphabetically in 4.x vs. fixed arbitrary order in 3.x.

    In 3.x, certain table/view options were omitted if their values were default values (i.e.: memtable_flush_period_in_ms = 0), this is no longer the case in 4.x unless the value is null (which typically means the column is not supported by the C* version or a feature is disabled).

2. The spacing on table/view map options are subtly different:

    - 3.x: `compaction = { 'class' : 'org.apache.cassandra.db.compaction.SizeTieredCompactionStrategy' }`
    - 4.x: `compaction = {‘class':'org.apache.cassandra.db.compaction.SizeTieredCompactionStrategy'}`

3. In 3.x, indices on a table are always sorted by name.

    In 4.x, indices on a table are sorted depending on the cassandra version:
      - C* 2.X:  Sorted by column indexed on
      - C* 3.X+: Sorted by index name.

    This is because indices are on columns in 2.x, and at table level on 3.x.  It makes sense to keep it this way to be consistent with the encounter order for the given cassandra version.

tolbertam:
There's a few integration tests that need to be updated to accommodate more queries made to capture schema metadata,  I'll update those shortly.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/schema/KeyspaceMetadata.java
```diff
@@ -98,7 +98,7 @@ default String describe(boolean pretty) {
         new ScriptBuilder(pretty)
             .append("CREATE KEYSPACE ")
             .append(getName())
-            .append(" WITH replication = { 'class' : '")
+            .append(" WITH REPLICATION = { 'class' : '")
```
olim7t: cqlsh lower-cases it (as well as durable_writes).

tolbertam: you are right about that.  Not sure how I got that impression.  Will revert that then.

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/Debouncer.java
```diff
@@ -107,8 +107,10 @@ public void flushNow() {
     assert adminExecutor.inEventLoop();
     LOG.debug("Flushing now");
     cancelNextFlush();
-    onFlush.accept(coalescer.apply(currentBatch));
-    currentBatch = new ArrayList<>();
+    if (!currentBatch.isEmpty()) {
```
olim7t, line 110: In which scenario did you observe that? I'm surprised this can happen, since a flush is only scheduled when the first element is received.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/DescribeIT.java
```diff
@@ -0,0 +1,247 @@
...
+                    + "WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1}",
+                keyspace));
+
+    // create session from this keyspace.
+    Session session = clusterRule.cluster().connect(keyspace);
+
+    KeyspaceMetadata originalKsMeta = clusterRule.cluster().getMetadata().getKeyspace(keyspace);
+
+    // udts require 2.1+
+    if (ccmRule.getCassandraVersion().compareTo(CassandraVersion.V2_1_0) >= 0) {
```
olim7t: The driver does not support lower versions, we don't need this test.

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/Debouncer.java
tolbertam: I noticed that when I set `metadata.schema.debouncer.window = 0 seconds` in one of my tests that an assertion was failing in `MetadataManager$SingleThreaded.coalesceSchemaRequests`.  I think this is because the code in `acceptSchemaRequest` isn't aware that the debouncer window is 0 and in that case the flush happens right away.

```java      
      schemaRefreshDebouncer.receive(future);
      if (flushNow) {
        schemaRefreshDebouncer.flushNow();
      }
```

in receive:

```java
    if (window.isZero() || maxEvents == 1) {
      LOG.debug(
          "Received {}, flushing immediately (window = {}, maxEvents = {})",
          element,
          window,
          maxEvents);
      onFlush.accept(coalescer.apply(ImmutableList.of(element)));
    }
```
```
java.lang.AssertionError
	at com.datastax.oss.driver.internal.core.metadata.MetadataManager$SingleThreaded.coalesceSchemaRequests(MetadataManager.java:304)
	at com.datastax.oss.driver.internal.core.util.concurrent.Debouncer.flushNow(Debouncer.java:110)
	at com.datastax.oss.driver.internal.core.metadata.MetadataManager$SingleThreaded.acceptSchemaRequest(MetadataManager.java:296)
	at com.datastax.oss.driver.internal.core.metadata.MetadataManager$SingleThreaded.access$1100(MetadataManager.java:224)
	at com.datastax.oss.driver.internal.core.metadata.MetadataManager.lambda$refreshSchema$5(MetadataManager.java:176)
	at com.datastax.oss.driver.internal.core.util.concurrent.RunOrSchedule.on(RunOrSchedule.java:55)
	at com.datastax.oss.driver.internal.core.metadata.MetadataManager.refreshSchema(MetadataManager.java:176)
	at com.datastax.oss.driver.internal.core.DefaultCluster$SingleThreaded.afterInitialNodeListRefresh(DefaultCluster.java:219)
```

olim7t: I see, the problem comes from this method:
```java
    private void acceptSchemaRequest(CompletableFuture<Metadata> future, boolean flushNow) {
      schemaRefreshDebouncer.receive(future);
      if (flushNow) {
        schemaRefreshDebouncer.flushNow();
      }
    }
```
Since the window is 0, `receive` has already flushed, so when we do the `flushNow` there is nothing left. I think it would be better to test the batch size in `flushNow`.

olim7t: > I think it would be better to test the batch size in flushNow.

Oops, your fix *is* in flushNow, ignore this :man_facepalming:

tolbertam (approved):
Looks great 👍 !

PR #881, opened 2017-09-27
JAVA-1631 Publish a sources jar for driver-core-tests

mfiguiere:
Over the past years, a lot of investment has been made in sophisticated testing API for the driver internals to make it easy to use CCM and Scassandra. These tools are available as a test-jar extra artifact of driver-core. This can be very convenient for users that need to implement some extensions to the DataStax Driver. Unfortunately the sources for this test-jar aren't published.

### DIFF
--- a/driver-core/pom.xml
+++ b/driver-core/pom.xml
@@ -185,6 +185,18 @@
                 </executions>
             </plugin>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-source-plugin</artifactId>
+                <executions>
+                    <execution>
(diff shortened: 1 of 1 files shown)


mfiguiere:
IntelliJ seems to find the sources automatically once installed on the local repo.

tolbertam (approved):
👍 this seems reasonable to me.  java-driver 4.x introduces a new module [`test-infra`](https://github.com/datastax/java-driver/tree/4.x/test-infra) where we intend to put code that drives our testing (CCM bridges, test utils, and so on) going forward, which will not require the `test` classifier to use.

PR #882, opened 2017-09-28
JAVA-1632: Add a withIpPrefix(String) method to CCMBridge.Builder

mfiguiere:
Currently, the IP prefix of the CCM cluster is obtained through TestUtils.IP_PREFIX which can be configured with a system property. This doesn't make it possible to create multiple CCM clusters at the same time as they would each need a different prefix. This is something we've needed in order to test a dual writes feature, for which we currently have to fork the class on our side.
Not of any immediate use for the driver itself, but note that ScassandraClusterBuilder does currently offer a withIpPrefix method anyway.


tolbertam (approved):
Looks great, thanks! 👍

PR #883, opened 2017-09-28
JAVA-1605: Refactor request execution model

olim7t:
Motivation:

Request types encapsulate their sync and async result types. This is too
restrictive: not all requests will have exactly those two types of
results. For example, if we add a CQL reactive processor, it would be
nice to reuse the Statement classes, not create new ones.

Modifications:

Make Request non generic. Change the generic Session execution method
to: execute(RequestT request, GenericType<ResultT> resultType)

Result:

Each request processor now handles exactly one (RequestT, ResponseT)
pair, so it is possible to add new response types for existing requests.

As a consequence, the number of built-in processor has doubled
(sync/async variants for each).


core/src/main/java/com/datastax/oss/driver/api/core/session/Session.java
```diff
@@ -54,101 +60,76 @@
...
-
-  /**
-   * Executes a CQL statement synchronously.
-   *
-   * <p>This is a convenience method that does the exact same thing as {@link #execute(Request)},
-   * but exposes a more user-friendly signature reminiscent of the 3.x API.
+   * Executes a CQL statement synchronously (the calling thread blocks until the result becomes
+   * available).
    */
   default ResultSet execute(Statement<?> statement) {
```
tolbertam: This is not completely related to the PR, but I had a thought about these `default` methods.  If I'm implementing my own `Session` implementation that simply wraps `Session` and provides some extra sugar methods wouldn't they clash with these?

Say for example I wanted to make a `GuavaSession` which provides async methods for returning `ListenableFuture`, the `executeAsync` methods from this interface come along with it, so I try to define `ListenableFuture<ResultSet> executeAsync(Statement<?> statement)`.  This would not work because the signature would clash with `CompletionStage<ResultSet> executeAsync(Statement<?> statement)`.

Do you see this as being a problem?  One possible solution would be to break Session into two interfaces.  A top level one (still named `Session`) that only has `getKeyspace` and the `execute` method that requires `Request` and then another one that extends `Session` and provides these specialized default methods (`DefaultSession` would extend this).

I mostly started thinking about this due to the general `execute` method taking a `Request`.  I figure library implementors will want to write customized sessions that provide sugar methods like these for the applicable types, but maybe the approach would be to make something that isn't a `Session`, but uses one and calls this method on the behalf of the user.

If the approach is to have implementors come up with their own `Session`.  It would be nice to generalize `Cluster` with its `connect` methods returning a type of `Session`.

I think I will just need to try implementing something to wrap my head around it.  I think this is the best approach to reviewing this (try implementing something that uses it).

tolbertam: I was revisiting some of our previous discussion and I think our thought was not to make specific `Session` implementations as things can get out of hand quickly, so maybe the *make something that isn't a Session, but uses one and calls this execute method on the behalf of the user* is the best way to go.

tolbertam (commented):
This looks like a great improvement over the previous implementation so far, I'd like to play with it a little bit to see how well it works to implement my own processors and use them.  Will start on that now and provide feedback sometime Monday.

Main things for me to explore/play with:

1. what does it look like/take to wire in my own processors (need to override `requestProcessorRegistry` in `InternalDriverContext`).
2. how easy is it to compose a processor from another (i.e. I just want to transform types from an underlying processor).
3. how would I go about exposing my own syntactic sugar that utilizes my processor (so users don't have to call execute(Request, GenericType) themselves).

tolbertam:
What do you think about the idea of being able to register `RequestProcessor`s into `RequestProcessorRegistry` on the fly?   Say for example, I want to take an existing `Cluster`  and decorate in functionality via processors.  Or would you prefer that is all spec'd out beforehand?  

The nice benefit of being able to do it on the fly:

1. Don't need to implement custom `InternalDriverContext` `requestProcessorRegistry()`, you can just wire in functionality post-hoc.
2. Might make it easier to add separate things that provide functionality (Reactive support, DSE support, etc. can all be wired in separately).

Although I can't see a reason why that all could be done beforehand, so maybe this isn't the best idea, just brainstorming :)

olim7t:
> What do you think about the idea of being able to register RequestProcessors into RequestProcessorRegistry on the fly?

The biggest drawback is that this would require some form of synchronization, and processor selection is done on the hot path. So I'd rather keep an immutable structure.

core/src/main/java/com/datastax/oss/driver/api/core/session/Session.java
olim7t: I can't think of a good solution to this in Java. I really think wrapping is not the solution, the main problem is that it doesn't compose.

What we would need is a way to dynamically mix in the different capabilities, returning some kind of union type:
```
session = cluster.connect() with GuavaCqlMethods with ReactiveCqlMethods
```
Obviously in Java this is science fiction.

All things considered, passing a type token to every `Session.execute` call is not that bad once you got used to it.

core/src/main/java/com/datastax/oss/driver/api/core/ClusterBuilder.java
```diff
@@ -135,14 +160,15 @@ public ClusterBuilder addTypeCodecs(TypeCodec<?>... typeCodecs) {
...
 
-  /** Convenience method to call {@link #buildAsync()} and block on the result. */
-  public Cluster<CqlSession> build() {
-    BlockingOperation.checkNotDriverThread();
-    return CompletableFutures.getUninterruptibly(buildAsync());
+  /**
+   * This <b>must</b> return an instance of {@code InternalDriverContext} (it's not expressed
+   * directly in the signature to avoid leaking that type through the protected API).
+   */
+  protected DriverContext buildContext(DriverConfigLoader configLoader) {
```
olim7t: This is a bit disingenuous: if a class using the protected API requires a type, then that type should be in the public API. But we're providing a convenience for custom driver extensions here, which use the internal APIs anyway, so I think the compromise is acceptable.
I've tried extracting a `ClusterBuilder` interface to avoid this problem, but it doesn't work well with self-types (you can't extend the interface and get covariant returns automatically, you'd have to override all methods).

core/src/main/java/com/datastax/oss/driver/api/core/Cluster.java
```diff
@@ -33,9 +33,10 @@
  *     CqlSession}.
  */
 public interface Cluster<SessionT extends Session> extends AsyncAutoCloseable {
+
   /** Returns a builder to create a new instance of the default implementation. */
-  static ClusterBuilder builder() {
-    return new ClusterBuilder();
+  static DefaultClusterBuilder builder() {
```
olim7t, line 38: We could return the generic type here, but it's a bit intimidating with the self type. So I chose to return the implementation (which means it has to be in the public API, but that's not a problem since it contains almost nothing).

tolbertam:
need to do a little work on the maven side to facilitate the use of non-shaded guava in the tests.

tolbertam (approved):
Looks great 👍

integration-tests/src/test/java/com/datastax/oss/driver/api/core/session/RequestProcessorIT.java
```diff
@@ -0,0 +1,146 @@
...
+ * implementations for returning {@link ListenableFuture}s rather than {@link
+ * java.util.concurrent.CompletionStage}s in async method responses.
+ *
+ * <p>{@link GuavaSession} provides execute method implementation shortcuts that mimics {@link
+ * CqlSession}'s async methods.
+ *
+ * <p>{@link KeyRequestProcessor} is also registered for handling {@link KeyRequest}s which
+ * simplifies a certain query down to 1 parameter.
+ */
+public class RequestProcessorIT {
```
olim7t, line 53: Great tests 👍

integration-tests/src/test/java/com/datastax/oss/driver/api/core/session/GuavaSession.java
```diff
@@ -0,0 +1,54 @@
...
+import com.datastax.oss.driver.api.core.cql.Statement;
+import com.datastax.oss.driver.api.core.type.reflect.GenericType;
+import com.datastax.oss.driver.internal.core.cql.DefaultPrepareRequest;
+import com.datastax.oss.driver.internal.core.session.SessionWrapper;
+import com.google.common.util.concurrent.ListenableFuture;
+
+public class GuavaSession extends SessionWrapper {
+
+  static final GenericType<ListenableFuture<AsyncResultSet>> ASYNC =
+      new GenericType<ListenableFuture<AsyncResultSet>>() {};
```
olim7t, line 30: Not that we care for this test, but for real extensions I think it will be a good practice to make the type tokens public just in case.
And also add the convenience methods via an interface.

PR #884, opened 2017-09-30
JAVA-1494: Implement Snappy and LZ4 compression

olim7t:
This was scheduled for alpha3 so no pressure, but if time permits we could include it in alpha2.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufCompressor.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufCompressor.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright (C) 2017-2017 DataStax Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.internal.core.protocol;
+
+import com.datastax.oss.protocol.internal.Compressor;
+import io.netty.buffer.ByteBuf;
+import java.nio.ByteBuffer;
+
+public abstract class ByteBufCompressor implements Compressor<ByteBuf> {
+
+  @Override
+  public ByteBuf compress(ByteBuf uncompressed) {
+    return uncompressed.isDirect() ? compressDirect(uncompressed) : compressHeap(uncompressed);
+  }
+
+  protected abstract ByteBuf compressDirect(ByteBuf input);
+
+  protected abstract ByteBuf compressHeap(ByteBuf input);
+
+  @Override
+  public ByteBuf decompress(ByteBuf compressed) {
+    return compressed.isDirect() ? decompressDirect(compressed) : decompressHeap(compressed);
+  }
+
+  protected abstract ByteBuf decompressDirect(ByteBuf input);
+
+  protected abstract ByteBuf decompressHeap(ByteBuf input);
+
+  protected static ByteBuffer inputNioBuffer(ByteBuf buf) {
+    // Using internalNioBuffer(...) as we only hold the reference in this method and so can
+    // reduce Object allocations.
+    int index = buf.readerIndex();
+    int len = buf.readableBytes();
+    return buf.nioBufferCount() == 1
+        ? buf.internalNioBuffer(index, len)
+        : buf.nioBuffer(index, len);
+  }
+
+  protected static ByteBuffer outputNioBuffer(ByteBuf buf) {
+    int index = buf.writerIndex();
+    int len = buf.writableBytes();
+    return buf.nioBufferCount() == 1
+        ? buf.internalNioBuffer(index, len)
+        : buf.nioBuffer(index, len);
+  }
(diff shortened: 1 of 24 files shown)


pom.xml
```diff
@@ -50,7 +50,7 @@
       <dependency>
         <groupId>com.datastax.oss</groupId>
         <artifactId>native-protocol</artifactId>
-        <version>1.4.0</version>
+        <version>1.4.1-SNAPSHOT</version>
```
olim7t, line 53: This requires a local install of native-protocol. I'll do the release at the last minute in case other changes come up.

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufCompressor.java
```diff
@@ -0,0 +1,59 @@
...
+
+import com.datastax.oss.protocol.internal.Compressor;
+import io.netty.buffer.ByteBuf;
+import java.nio.ByteBuffer;
+
+public abstract class ByteBufCompressor implements Compressor<ByteBuf> {
+
+  @Override
+  public ByteBuf compress(ByteBuf uncompressed) {
+    return uncompressed.isDirect() ? compressDirect(uncompressed) : compressHeap(uncompressed);
```
tolbertam, line 26: Nice to break up the direct/heap implementations, that made the code a bit hard to follow on 3.x. 👍

tolbertam:
I had to go back to `HEAD~2` on native-protocol to get a snapshot that works (I think recent changes on native-protocol aren't compatible with this branch).

integration-tests/src/test/java/com/datastax/oss/driver/api/core/compression/DirectCompressionIT.java
```diff
@@ -0,0 +1,99 @@
...
+  public void should_execute_queries_with_lz4_compression() throws Exception {
+    createAndCheckCluster(
+        "protocol.compressor.class = "
+            + "com.datastax.oss.driver.internal.core.protocol.Lz4Compressor");
+  }
+
+  private void createAndCheckCluster(String compressorOption) {
+
+    try (Cluster cluster =
+        ClusterUtils.newCluster(ccmRule, compressorOption, "request.timeout = 30 seconds")) {
```
tolbertam: The request.timeout isn't needed here right?  That's more needed for making schema changes I had thought.

tolbertam (commented):
Looks great so far, the tests follow what was done on 3.x so I consider that adequate.  I'm doing some quick local testing with large throughput to see if the memory profiles look like what I'm used to seeing with direct/heap use and then will approve.  I pushed a small change to tests to remove the additional use of larger timeout which I don't think is needed.

tolbertam:
Unsure if this is related to this PR, but i noticed that the netty leak detection fired off when running with Snappy and LZ4 compression ~(not sure if isolated to snappy, checking lz4)~:

```
ERROR [2017-10-13 10:15:22,489] [c0-io-13            ] io.netty.util.ResourceLeakDetector - LEAK: ByteBuf.release() was not called before it's garbage-collected. Enable advanced leak reporting to find out where the leak occurred. To enable advanced leak reporting, specify the JVM option '-Dio.netty.leakDetection.level=advanced' or call ResourceLeakDetector.setLevel() See http://netty.io/wiki/reference-counted-objects.html for more information.
```

Will turn on advanced detection and see what I find.

Note that the warning does not pop up if compression is not enabled.

tolbertam:
Leak detection output:  https://gist.github.com/tolbertam/1269be8a4efd899d1a82220e2637fee8

tolbertam:
Wonder if it's because we're not releasing `compressedMessage` in the native-protocol code in `FrameCodec.encode`?  Will try it out

```
      // 2) Compress and measure size, discard uncompressed buffer
      B compressedMessage = compressor.compress(uncompressedMessage);
      primitiveCodec.release(uncompressedMessage);
      int messageSize = primitiveCodec.sizeOf(compressedMessage);

      // 3) Encode final frame
      B header = primitiveCodec.allocate(headerSize);
      encodeHeader(frame, flags, messageSize, header);
      return primitiveCodec.concat(header, compressedMessage);
```

tolbertam:
I was on the wrong track, the problem is in decompress, not compress.  Looking into it.

tolbertam:
I think I have a handle on this, simply just needed to add a `primitiveCodec.release(source)` in `FrameCodec.decode` and the tracker no longer detects any leaks:

```java

    if (flags.contains(Flag.COMPRESSED)) {
      source = compressor.decompress(source);
    }

    boolean isTracing = flags.contains(Flag.TRACING);
    UUID tracingId = (isResponse && isTracing) ? primitiveCodec.readUuid(source) : null;

    Map<String, ByteBuffer> customPayload =
        (flags.contains(Flag.CUSTOM_PAYLOAD))
            ? primitiveCodec.readBytesMap(source)
            : Collections.emptyMap();

    List<String> warnings =
        (isResponse && flags.contains(Flag.WARNING))
            ? primitiveCodec.readStringList(source)
            : Collections.emptyList();

    Message.Codec decoder = decoders.get(protocolVersion, opcode);
    ProtocolErrors.check(
        decoder != null, "Unsupported request opcode: %s in protocol %d", opcode, protocolVersion);
    Message response = decoder.decode(source, primitiveCodec);

    primitiveCodec.release(source); // <----HERE

    return new Frame(
        protocolVersion, beta, streamId, isTracing, tracingId, customPayload, warnings, response);
```

However, this causes issues with compression disabled, (source is released upstream otherwise).  So maybe we should change the code to only release if the reference returned by `decompress` is different than the input.

tolbertam (approved):
Working great 👍 .  Ran ~2 minute duration tests with uncompressed, Snappy and LZ4 compressors in both direct and heap mode.  Everything worked as detected and profile showed correct paths being used with no leaks detected with advanced detection enabled.

PR #885, opened 2017-10-01
JAVA-1514: Port Uuids utility class

olim7t:
Same comment as #884: scheduled for alpha3, but we could include it in alpha2 if time permits.


tolbertam (approved):
👍

PR #886, opened 2017-10-04
JAVA-1520: Add node state listeners

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/ClusterBuilder.java
```diff
@@ -116,6 +119,11 @@ public ClusterBuilder addTypeCodecs(TypeCodec<?>... typeCodecs) {
     return this;
   }
 
+  public ClusterBuilder addNodeStateListeners(NodeStateListener... newListeners) {
```
tolbertam: I noticed that `onRegister` is not called for listeners passed in this way (although `onUnregister` is), should it be?

tolbertam: Added a commit for that as I think it makes sense, but feel free to back out if you disagree.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/NodeStateIT.java
```diff
@@ -119,6 +144,7 @@ public void should_keep_regular_node_up_when_still_one_connection() {
         .as("Reconnection started")
         .before(10, TimeUnit.SECONDS)
         .becomesTrue();
+    inOrder.verify(nodeStateListener, never()).onDown(metadataRegularNode);
```
tolbertam, line 148: Awesome 👍 .  Like that could use existing tests to validate the state listener behavior.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/NodeStateIT.java
```diff
@@ -352,10 +393,137 @@ public void should_force_down_when_ignored() throws InterruptedException {
...
+
+      // Non-contact point now has a connection opened => up
+      Mockito.verify(localNodeStateListener, timeout(100)).onUp(localMetadataNode2);
+    }
+  }
+
+  @Test
+  public void should_remove_invalid_contact_point() {
+    // Initialize the driver with 1 wrong address and 1 valid address
+    InetSocketAddress wrongContactPoint = unusedAddress();
```
tolbertam, line 439: Looks like a great way to validate the onRemove event (tricky to do otherwise).

tolbertam (approved):
Looks great!  I added a suggested change a couple of tests around it.  This looks good to me 👍

core/src/main/java/com/datastax/oss/driver/api/core/ClusterBuilder.java
olim7t: Yes, I forgot.

olim7t: I'm going to move this to `init()`, it's functionally equivalent but will be a bit more concise since we're already on the admin thread.

PR #887, opened 2017-10-12
JAVA-1525: Handle token metadata

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/cql/ResultSet.java
```diff
@@ -65,6 +69,38 @@ default ExecutionInfo getExecutionInfo() {
   List<ExecutionInfo> getExecutionInfos();
 
   /**
+   * Returns the next row, or {@code null} if the result set is exhausted.
+   *
+   * <p>This is convenient for queries that are known to return exactly one row, for example count
+   * queries.
+   */
+  default Row one() {
```
tolbertam, line 77: happy to see these methods back :)

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Metadata.java
```diff
@@ -38,9 +40,27 @@
...
   Map<CqlIdentifier, KeyspaceMetadata> getKeyspaces();
 
   default KeyspaceMetadata getKeyspace(CqlIdentifier keyspaceId) {
     return getKeyspaces().get(keyspaceId);
   }
+
+  /**
+   * The token map for this cluster.
+   *
+   * <p>Note that this property might be absent if there was a runtime error while computing the map
```
tolbertam: should note it can also be absent if disabled via configuration

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Node.java
```diff
@@ -19,10 +19,12 @@
 import com.datastax.oss.driver.api.core.addresstranslation.AddressTranslator;
 import com.datastax.oss.driver.api.core.loadbalancing.LoadBalancingPolicy;
 import com.datastax.oss.driver.api.core.loadbalancing.NodeDistance;
+import com.datastax.oss.driver.api.core.metadata.token.Token;
```
tolbertam: Are these extra imports accidental?  Since no code was added seems like these should be unused.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultMetadata.java
```diff
@@ -56,13 +70,90 @@ private DefaultMetadata(
...
+  @Override
+  public Optional<TokenMap> getTokenMap() {
+    return tokenMap;
+  }
+
+  /**
+   * @param tokenMapEnabled
+   * @param tokensChanged whether we observed a change of tokens for at least one node. This will
+   *     require a full rebuild of the token map.
+   * @param tokenFactory only needed for the initial refresh, afterwards the existing one in the
```
tolbertam, line 82: looks like this comment was cut off / is incomplete.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultMetadata.java
```diff
@@ -56,13 +70,90 @@ private DefaultMetadata(
...
+   */
+  public DefaultMetadata withNodes(
```
tolbertam: Nice, so whenever topology of schema changes, the token map is recomputed?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultNodeInfo.java
```diff
@@ -79,6 +81,11 @@ public String getCassandraVersion() {
   }
 
   @Override
+  public String getPartitioner() {
```
tolbertam, line 84: Will the partitioner ever vary at a node level?  Maybe makes more sense to put this on `Cluster` or `Metadata`

tolbertam: I see this is a column in the peers table, that is sort of strange.  Might be done that way because there is no easier way to expose it.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/InitContactPointsRefresh.java
```diff
@@ -36,15 +35,15 @@
   }
 
   @Override
-  public Result compute(DefaultMetadata oldMetadata) {
+  public Result compute(DefaultMetadata oldMetadata, boolean tokenMapEnabled) {
```
tolbertam, line 38: tokenMapEnabled doesn't seem to be used here.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/TokenMap.java
```diff
@@ -0,0 +1,92 @@
...
+  Set<Node> getReplicas(CqlIdentifier keyspace, ByteBuffer partitionKey);
+
+  /**
+   * The replicas for a given range in the given keyspace.
+   *
+   * <p>It is assumed that the input range does not overlap across multiple node ranges. If the
+   * range extends over multiple nodes, it only returns the nodes that are replicas for the last
+   * token of the range.
+   */
+  Set<Node> getReplicas(CqlIdentifier keyspace, TokenRange range);
```
tolbertam: A `getReplicas(CqlIdentifier keyspace, Token token)` would be nice to have too

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/DefaultTokenMap.java
```diff
@@ -0,0 +1,266 @@
...
+    return (keyspaceMap == null) ? Collections.emptySet() : keyspaceMap.getReplicas(range);
+  }
+
+  private KeyspaceTokenMap getKeyspaceMap(CqlIdentifier keyspace) {
+    Map<String, String> config = replicationConfigs.get(keyspace);
+    return (config == null) ? null : keyspaceMaps.get(config);
+  }
+
+  /** Called when only the schema has changed. */
+  public DefaultTokenMap refresh(Collection<Node> nodes, Collection<KeyspaceMetadata> keyspaces) {
```
tolbertam, line 173: Should this be public?  I noticed it's only called from a private method (`DefaultMetadata.rebuildTokenMap`)

tolbertam (commented):
Looks great so far and the tests are pretty complete, will do some more testing but it is likely this is ready to go as is 👍 .  Completing review to share comments so far

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/ByteOrderedTokenIT.java
```diff
@@ -0,0 +1,58 @@
...
+import com.datastax.oss.driver.api.testinfra.ccm.CustomCcmRule;
+import com.datastax.oss.driver.api.testinfra.cluster.ClusterRule;
+import com.datastax.oss.driver.categories.IsolatedTests;
+import com.datastax.oss.driver.internal.core.metadata.token.ByteOrderedToken;
+import com.datastax.oss.driver.internal.core.metadata.token.RandomToken;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.experimental.categories.Category;
+
+@Category(IsolatedTests.class)
```
tolbertam: I don't think these need to be isolated (could just be long), unless i'm missing something,  however in hindsight the difference between long and isolated isn't really that much, maybe we should consider joining the two.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Node.java
olim7t: Yes, they were accidental.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultMetadata.java
olim7t: Wrt to nodes, it's whenever nodes were added/removed, or the tokens of a node changed.
Wrt schema, it's whenever (non-filtered) keyspaces were added/removed, or the replication settings of a keyspace changed.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultNodeInfo.java
olim7t: I hadn't thought about this, but it's indeed a direct consequence of it being in the local/peers tables. It's the same for all nodes, and only the value of the control node during the first node list refresh is really used.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/InitContactPointsRefresh.java
olim7t: No, it's only here because it's inherited from the interface.
This refresh is applied before we're even connected to the first node, we don't know any tokens yet so the map can't be computed.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/DefaultTokenMap.java
olim7t: The two classes are in different packages so it has to be public. This is the internal API, so it's not really critical to hide it.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/ByteOrderedTokenIT.java
olim7t: They require a specific configuration of the CCM cluster (3 nodes, custom partitioner), doesn't that require isolation?

core/src/main/java/com/datastax/oss/driver/api/core/metadata/TokenMap.java
olim7t: Added it.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/ByteOrderedTokenIT.java
tolbertam: the `IsolatedTests` category was made for tests that need their own JVM (for setting up system properties and such for SSL among other things).   The `CustomCcmRule` is set up such that there can only be one going at a given time (and they are shut down) so you can have multiple tests using `CustomCcmRule` in the same JVM.

However, the long tests by their nature are...long, and running each test class in its own JVM doesn't create all too much overhead, so we could consider merging those two categories if that makes sense?

tolbertam (approved):
Looks really great!  The test coverage looks similar or better than 3.x as is, so I don't really have anything to add. 👍

PR #888, opened 2017-10-12
 JAVA-1638: Check schema agreement

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/Cluster.java
```diff
@@ -116,6 +117,40 @@ default Metadata refreshSchema() {
...
+   * Checks if all nodes in the cluster agree on a common schema version.
+   *
+   * <p>The goal of schema agreement is to ensure that a schema change has been propagated to all
+   * nodes before proceeding with the next queries (which might go to a different coordinator).
+   * Normally, you should not have to do that manually, because it's handled by the driver after all
+   * schema-altering queries. However, this method is provided in case schema agreement failed
+   * earlier and you want to check it again.
+   *
+   * <p>To check agreement, this method queries system tables to find out the schema version of all
+   * nodes that are currently {@code {@link NodeState#UP UP}}, and compares them. If the comparison
```
tolbertam: I know we did this in 3.x as well, but is there any reason why don't we validate against all nodes, not just ones that we consider `UP`?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/SchemaAgreementChecker.java
```diff
@@ -0,0 +1,200 @@
...
+      sendQueries();
+    }
+    return result;
+  }
+
+  private void sendQueries() {
+    long elapsedNs = System.nanoTime() - start;
+    if (elapsedNs > timeoutNs) {
+      LOG.warn("[{}] Schema agreement not reached after {}", logPrefix, NanoTime.format(elapsedNs));
+      result.complete(false);
```
tolbertam: Noticed while testing that the checker keeps trying even after the future completes,  shouldn't the code also return here?  Otherwise it keeps trying to verify whether or not the schema agrees.

I.e., default timeout set to 10s, but i noticed logs well after:

```
14:45:27.949 [c0-io-0] WARN  c.d.o.d.i.c.m.SchemaAgreementChecker - [c0] Schema agreement not reached after 22 s
```

tolbertam (commented):
Looks great with one slight exception that the schema check keeps running even after timeout.

Also added a 'long' integration test.  It doesn't validate as much as the unit test but provides a good sanity check with a ccm cluster.

core/src/main/java/com/datastax/oss/driver/api/core/Cluster.java
tolbertam: To expound on that.  Just because the driver thinks a node is down, doesn't mean it's actually down (the driver could be actively ignoring it too).  Maybe it is best to ensure all nodes agree, not just the ones we can connect to.

olim7t: What if a node is really down? Schema agreement would have no chance of succeeding, as I imagine `system.peers` keeps the last known version.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/SchemaAgreementChecker.java
olim7t: Right that's a mistake, the rest of the method should be in an `else` block.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SchemaAgreementIT.java
```diff
@@ -0,0 +1,115 @@
...
+  @Test
+  public void should_succeed_when_all_nodes_agree() {
+    ResultSet result = createTable();
+
+    assertThat(result.getExecutionInfo().isSchemaInAgreement()).isTrue();
+    assertThat(clusterRule.cluster().checkSchemaAgreement()).isTrue();
+  }
+
+  @Test
+  public void should_fail_on_timeout() {
```
olim7t, line 60: How about lowering `schema-agreement.timeout` to make this test a bit faster?

tolbertam: good idea, will do

core/src/main/java/com/datastax/oss/driver/api/core/Cluster.java
tolbertam: That's a good point, but we don't really have a way to know if a host is actually 'down' right?   This is probably too paranoid, but if you have some kind of network partition, and a schema change is made on one set of nodes, and also on another set, you could get a disagreement that way and i'm not sure how cassandra resolves these situations (it probably varies from release to release).   Since the 3.x driver already worked this way, I don't think we need to consider changing it here now (maybe some kind of future debate, i'll think on it some more).

It probably isn't a good idea to make schema changes if nodes are down/unavailable.  You could also argue that we should check for schema agreement before making a schema change (although it's not really possible on our end to know if a query is a schema change ahead of time).

tolbertam (approved):
👍

core/src/main/java/com/datastax/oss/driver/api/core/Cluster.java
olim7t: The goal of schema agreement is to protect *the client* from inconsistencies, e.g.:
* you create a table in one request and insert a row into it in the next, but the latter fails because it went to a different coordinator that hasn't sync'd yet,
* a new node comes up, you try to reprepare statements on it but it fails because it doesn't have all the tables,
* a schema change event comes in, you refresh but miss the update (admittedly harder because most of the time the refresh is done on the same control connection that received the event).

If the client thinks a node is down, it's not interacting with it, so none of these issues can happen. If the node is actually up and out of sync, it does not matter.

PR #890, opened 2017-10-16
JAVA-1527: Expose host_id and schema_version in Host metadata.

newkek:


driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
```diff
@@ -516,6 +516,8 @@ private static void updateInfo(Host host, Row row, Cluster.Manager cluster, bool
             String dseVersion = row.getString("dse_version");
             host.setDseVersion(dseVersion);
         }
+        host.setHostId(row.getUUID("host_id"));
```
tolbertam, line 519: I was skeptical that these would always be present in the column metadata, but as far as I know these values are always present going back to 1.2, CI seems to agree with that.

driver-core/src/main/java/com/datastax/driver/core/Host.java
```diff
@@ -304,6 +317,31 @@ public boolean isDseGraphEnabled() {
...
+     * @return the node's host id value.
+     */
+    public UUID getHostId() {
+        return hostId;
+    }
+
+    /**
+     * Return the current schema version for the host.
+     * <p/>
+     * Schema versions in Cassandra are used to insure all the nodes agree on the current
```
tolbertam: Small nit:  `insure` should be `ensure`

tolbertam (approved):
Looks great, just a small spelling nit, but otherwise this is good to go 👍

PR #894, opened 2017-10-19
JAVA-1524: Add query trace API

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/config/CoreDriverOption.java
```diff
@@ -49,6 +49,8 @@
   SPECULATIVE_EXECUTION_POLICY_ROOT("request.speculative-execution-policy", true),
   RELATIVE_SPECULATIVE_EXECUTION_MAX("max-executions", false),
   RELATIVE_SPECULATIVE_EXECUTION_DELAY("delay", false),
+  REQUEST_TRACE_ATTEMPTS("request.trace.attempts", true),
```
tolbertam, line 52: Was just remembering that query trace can be unreliable in multi-DC scenarios ([JAVA-1473](https://datastax-oss.atlassian.net/browse/JAVA-1473)).  If we use the default consistency level (LOCAL_ONE), it's possible trace queries will fail if the replicas of the session_id are in a remote DC.  Should we considering adding a parameter to configure the trace CL, or just set it to ONE?

olim7t: Good observation, I pushed a commit to address that.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/QueryTraceFetcher.java
```diff
@@ -50,6 +52,17 @@
...
+
+    ConsistencyLevel regularConsistency =
+        configProfile.getConsistencyLevel(CoreDriverOption.REQUEST_CONSISTENCY);
+    ConsistencyLevel traceConsistency =
+        configProfile.getConsistencyLevel(CoreDriverOption.REQUEST_TRACE_CONSISTENCY);
+    this.configProfile =
+        (traceConsistency == regularConsistency)
+            ? configProfile
+            : configProfile.withConsistencyLevel(
+                CoreDriverOption.REQUEST_CONSISTENCY, traceConsistency);
```
olim7t, line 64: This creates a new profile for every trace retrieval. It should be fine since tracing incurs a performance hit anyway, you're not supposed to do it for every request in high-performance environments.

PR #895, opened 2017-10-20
MixSpeculativeExecutionPolicy

Kamil-Benedykcinski:
A policy that triggers speculative executions when the request time to the current host is above a given max delay or when request time is in given range (min, max) and above given percentile.


tolbertam:
Hi @Kamil-Benedykcinski, thank you for your contribution!  My intuition is that while this policy could be useful, I think it is too specialized to be included in the core driver.  The nice thing about policies is that if others find this useful, they can bring it into their project and use it, and if we get good feedback from the community that this is useful, we can bring it into core.

tolbertam:
Closing this for now, but if anyone has a strong motivation, please reopen this.  Since policies are pluggable, you can extract this code and add it to your application.

PR #896, opened 2017-10-20
ConnectionMetrics

Kamil-Benedykcinski:
Connection metrics per connected hosts + StateListener + other metrics.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/ConnectionMetrics.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ConnectionMetrics.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright (C) 2012-2017 DataStax Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


tolbertam (commented):
Will this can be useful, I think users will have different opinions on how they want to expose and log `Metrics`, `Session.State`, and host state changes.  Since these are already publicly available, my preference is not to include this.

newkek:
If this PR is about reporting Metrics **per host**, it has been mentioned before and we have a ticket for it: https://datastax-oss.atlassian.net/browse/JAVA-1627. However it is likely that this will be in the next major version of the driver (4.0). Hence we can't accept this PR here but a contribution on that JAVA-1627 ticket would be much appreciated!

olim7t:
Mmm actually it's a bit early for a contribution on JAVA-1627, 4.0 is still in alpha version and the metrics architecture is not yet fully defined.

tolbertam:
4.0.0-alpha3 was released earlier this week includes the option to surface host level metrics (see [manual](https://docs.datastax.com/en/developer/java-driver/4.0-alpha/manual/core/metrics/)), including # of open connections and in flight requests which was originally suggested this by PR, so closing.

PR #899, opened 2017-10-27
JAVA-1526: Provide a single load balancing policy implementation

olim7t:


olim7t:
Rebased on current 4.x.

olim7t:
~TODO revisit all integration tests, most of them should require an explicit local DC now.~

PR #900, opened 2017-10-27
JAVA-1584: Validate that no bound values are unset in protocol v3

olim7t:


olim7t:
Merged this early since this is low risk and I need `ProtocolFeature` for another ticket.

PR #902, opened 2017-10-30
[JAVA-1659] Propagate fix from CASSANDRA-13651 to reduce CPU usage by eventLoop.schedule

netudima:
Introduce new tuning options:
com.datastax.driver.FLUSHER_SCHEDULE_PERIOD_NS
com.datastax.driver.FLUSHER_RUN_WITHOUT_WORK_TIMES
Old behavior is used as default


olim7t (approved):
LGTM, minimal impact since the default values match the previously hard-coded ones.

tolbertam (approved):
👍

olim7t:
Rebased on 3.3.x, squashed and merged.

We don't update dependencies on patch versions, so if you want the Netty patch that supports microsecond resolution (https://github.com/netty/netty/pull/7042), you'll have to override the driver's dependency in your application. The 4.0 version that has it is 4.0.51.Final.

PR #903, opened 2017-10-30
JAVA-1660: Support netty-transport-native-epoll in OSGi container

netudima:


driver-core/pom.xml
```diff
@@ -229,6 +229,8 @@
                             <instructions>
                                 <!-- JNR does not provide OSGi bundles, so exclude it; the driver can live without it -->
                                 <Import-Package><![CDATA[com.google.common.*;version="[16.0.1,22)",!jnr.*,*]]></Import-Package>
+                                <!-- Give the ability to discover and load netty-transport-native-epoll classes -->
+                                <DynamicImport-Package>*</DynamicImport-Package>
```
olim7t: I have limited knowledge of OSGI, but from what I can see in the [Netty PR](https://github.com/netty/netty/pull/5326/files), this is necessary to load native libraries. Is there any way we could target this more accurately? The `*` seems very generic (although I can't really tell the consequences).

netudima: Native library is a corner case. DynamicImport-Package has the following main meaning: within OSGi container you have multiple bundles (JARs), each one is loaded by it's own classloader. OSGi controls visibility of java packages between bundles based on declaration in bundles manifests. Bundle can declare some packages as exported (Export-Package), it means that they can be used by other bundles. From consumer side - bundle declares what packages it needs using Import-Package. Usually import packages are generated automatically by maven-bungle-plugin based on imports in code. If a class is not a part of bundle itself, not a part of base JDK packages (in OSGI there is a boot delegation logic) and not declared as imported - it will not be resolved. 
There are cases when plain import-export logic is not enough - classes are not declared in code explicitly, the most common cause is Class.forName usage when class is loaded by name (as we have in com.datastax.driver.core.NettyUtil class). One possible way to deal with such dynamic class loading is to add DynamicImport-Package declaration. It allows to resolve class from another bundle if it is not declared explicitly as imported. Only exported packages can be imported via dynamic imports, so it does not break encapsulation.
Import-Package and DynamicImport-Package are also resolved at different moments of bundle lifecycle. Import-Package is resolved when a bundle is installed (and if a package is not present and not declared as optional - the resolving will fail). DynamicImport-Package is used in runtime, when an actual code tries to load classes and other ways do not work.
In theory we can try to limit DynamicImport-Package scope to  `io.netty.channel.epoll.*` but I did not test such option to be sure. Regarding consequences of using `*`  - actually there is no large difference because DynamicImport is used only as a last chance to load a class if other mechanisms do not work (the resolving order is a part of OSGi specification), so for existing working logic there is no large difference.

adutra (changes requested):
Thanks for bringing this up! 

I'm not convinced however that we need dynamic imports.

I just made the following experiment:

1. Bumped Netty's version to 4.1.17 because in 4.0 the epoll package is not OSGi-ready;
2. Added epoll to the regular `Import-Package` directive instead of `DynamicImport-Package`:
```xml
<Import-Package><![CDATA[com.google.common.*;version="[16.0.1,22)",!jnr.*,io.netty.channel.epoll,*]]></Import-Package>
```
3. Verified that the manifest contains the following `Import-Package` entry:
```
io.netty.channel.epoll;version="[4.1,5)";resolution:=optional
```
4. Modified our OSGi tests by adding a few more bundles in `BundleOptions.java`:
```java
public static CompositeOption nettyBundles() {
    final String nettyVersion = getVersion("netty.version");
    return new CompositeOption() {

        @Override
        public Option[] getOptions() {
        return options(
            mavenBundle("io.netty", "netty-buffer", nettyVersion),
            mavenBundle("io.netty", "netty-codec", nettyVersion),
            mavenBundle("io.netty", "netty-common", nettyVersion),
            mavenBundle("io.netty", "netty-handler", nettyVersion),
            mavenBundle("io.netty", "netty-resolver", nettyVersion),
            mavenBundle("io.netty", "netty-transport", nettyVersion),
            mavenBundle("io.netty", "netty-transport-native-epoll", nettyVersion),
            mavenBundle("io.netty", "netty-transport-native-unix-common", nettyVersion)
        );
        }
    };
}
```
5. Launched `com.datastax.driver.osgi.MailboxServiceDefaultIT`;
6. Verified that the logs contain the following messages (I'm on Mac OS):
```
WARN  com.datastax.driver.core.NettyUtil - Found Netty's native epoll transport, but not running on linux-based operating system. Using NIO instead.
DEBUG c.d.driver.osgi.MailboxServiceTests - Loaded bundle: io.netty.transport-native-epoll 4.1.17.Final
DEBUG c.d.driver.osgi.MailboxServiceTests - Loaded bundle: io.netty.transport-native-unix-common 4.1.17.Final
```

So my take away here is that we could use a regular `Import-Package`; no need for `DynamicImport-Package`. (I'm not familiar with dynamic imports tbh, but from what I read, its usage is fairly discouraged, and if we could avoid it, that would make me more comfortable accepting this PR).

BTW it would be nice to add a test to our OSGi test suite to validate that Epoll gets loaded in *Nix systems.

netudima:
Thank you for the comments.

>  Bumped Netty's version to 4.1.17 because in 4.0 the epoll package is not OSGi-ready;

Actually it was added in 4.0.x version as well: https://github.com/netty/netty/issues/5119, so I think it is not mandatory.
For example: netty-transport-native-epoll-4.0.52.Final-linux-x86_64.jar which I used to test contains required OSGi directives in manifest and works normally in OSGi container.
I used 4.0.x version to reduce the scope of changes because I am not aware about plans to switch from netty 4.0.x to 4.1.x versions in the driver dependencies and possible related concerns
So, I am ok with 4.0.x or 4.1.x as well.

Regarding Import vs Dynamic-Import, yes, usual Import is more explicit in declaring dependencies, so we can try to use optional Import (optional directive is important here to not fail resolving if the optional dependency is not available). I will check what this option actually works and there are no hidden problems.

olim7t:
I tested Alexandre's solution on a Linux box:
* update `<Import-Package>` as he indicated
* update `BundleOptions`, but with only one additional bundle:
    ```
    mavenBundle()
        .groupId("io.netty")
        .artifactId("netty-transport-native-epoll")
        .version(nettyVersion)
        .classifier("linux-x86_64")
    ```
* run the tests and check that I see the message `Found Netty's native epoll transport in the classpath, using it` in the logs. Add a log of the channel class and check that it's the epoll one.
* revert the change in `BundleOptions` and check that the test falls back to NIO.

Everything worked fine, so I'll proceed with this approach.

olim7t:
Merged.
Like for #902, you'll need to override Netty version in your app.

netudima:
Thank you!, I have run a smoke test on my environment (linux, Apache Karaf) using 3.3.2 version and it works fine.

PR #905, opened 2017-10-30
JAVA-1661: avoid toLowerCase if it is possible in Metadata

netudima:
Avoid toLowerCase invocation if it is possible to reduce CPU usage and GC overheads in Metadata class. The operation is quite simple but invokes very frequently.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Metadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Metadata.java
@@ -18,6 +18,7 @@ package com.datastax.driver.core;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
+import io.netty.util.collection.IntObjectHashMap;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -45,7 +46,7 @@ public class Metadata {
     final ReentrantLock lock = new ReentrantLock();
 
     // See https://github.com/apache/cassandra/blob/trunk/doc/cql3/CQL.textile#appendixA
-    private static final Set<String> RESERVED_KEYWORDS = ImmutableSet.of(
+    private static final IntObjectHashMap<List<char[]>> RESERVED_KEYWORDS = indexByCaseInsensitiveHash(
             "add", "allow", "alter", "and", "any", "apply", "asc", "authorize", "batch", "begin", "by",
             "columnfamily", "create", "delete", "desc", "drop", "each_quorum", "from", "grant", "in",
             "index", "inet", "infinity", "insert", "into", "keyspace", "keyspaces", "limit", "local_one",
@@ -125,25 +126,32 @@ public class Metadata {
         if (id == null)
             return null;
 
-        if (isAlphanumeric(id))
+        boolean isAlphanumericLowCase = true;
+        boolean isAlphanumeric = true;
+        for (int i = 0; i < id.length(); i++) {
+            char c = id.charAt(i);
+            if (c >= 65 && c <= 90) { // A-Z
+                isAlphanumericLowCase = false;
+            } else if (!(
+                    (c >= 48 && c <= 57) // 0-9
+                            || (c == 95) // _ (underscore)
+                            || (c >= 97 && c <= 122) // a-z
+            )) {
+                isAlphanumeric = false;
+                isAlphanumericLowCase = false;
+                break;
+            }
+        }
+
+        if (isAlphanumericLowCase) {
+            return id;
+        }
+        if (isAlphanumeric) {
             return id.toLowerCase();
+        }
 
         // Check if it's enclosed in quotes. If it is, remove them and unescape internal double quotes
         return ParseUtils.unDoubleQuote(id);
-    }
-
-    private static boolean isAlphanumeric(String s) {
-        for (int i = 0; i < s.length(); i++) {
-            char c = s.charAt(i);
-            if (!(
-                    (c >= 48 && c <= 57) // 0-9
-                            || (c >= 65 && c <= 90) // A-Z
-                            || (c == 95) // _ (underscore)
-                            || (c >= 97 && c <= 122) // a-z
-            ))
-                return false;
-        }
-        return true;
     }
 
     /**
@@ -273,8 +281,73 @@ public class Metadata {
      * @return {@code true} if the given identifier is a known reserved
      * CQL keyword, {@code false} otherwise.
      */
+
     public static boolean isReservedCqlKeyword(String id) {
-        return id != null && RESERVED_KEYWORDS.contains(id.toLowerCase());
+        if (id == null) {
+            return false;
+        }
+        int hash = caseInsensitiveHash(id);
+        List<char[]> keywords = RESERVED_KEYWORDS.get(hash);
+        if (keywords == null) {
+            return false;
+        } else {
+            for (char[] keyword : keywords) {
+                if (equalsIgnoreCaseAscii(id, keyword)) {
+                    return true;
+                }
+            }
+            return false;
+        }
+    }
+
+    private static int caseInsensitiveHash(String str) {
+        int hashCode = 17;
+        for (int i = 0; i < str.length(); i++) {
+            char c = toLowerCaseAscii(str.charAt(i));
+            hashCode = 31 * hashCode + c;
+        }
+        return hashCode;
+    }
+
+    // keyword is expected as a second argument always in low case
+    private static boolean equalsIgnoreCaseAscii(String str1, char[] str2LowCase) {
+        if (str1.length() != str2LowCase.length) return false;
+
+        for (int i = 0; i < str1.length(); i++) {
+            char c1 = str1.charAt(i);
+            char c2Low = str2LowCase[i];
+            if (c1 == c2Low) {
+                continue;
+            }
+            char low1 = toLowerCaseAscii(c1);
+            if (low1 == c2Low) {
+                continue;
+            }
+            return false;
+        }
+        return true;
+    }
+
+    private static char toLowerCaseAscii(char c) {
+        if (c >= 65 && c <= 90) { // A-Z
+            c ^= 0x20; // convert to low case
+        }
+        return c;
+    }
+
+    private static IntObjectHashMap<List<char[]>> indexByCaseInsensitiveHash(String... words) {
+        IntObjectHashMap<List<char[]>> result = new IntObjectHashMap<List<char[]>>();
+        for (String word : words) {
+            char[] wordAsCharArray = word.toLowerCase().toCharArray();
+            int hash = caseInsensitiveHash(word);
+            List<char[]> list = result.get(hash);
+            if (list == null) {
+                list = new ArrayList<char[]>();
+                result.put(hash, list);
+            }
+            list.add(wordAsCharArray);
+        }
+        return result;
     }
 
     /**
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -5,6 +5,7 @@
 - [bug] JAVA-1666: Fix keyspace export when a UDT has case-sensitive field names.
 - [improvement] JAVA-1196: Include hash of result set metadata in prepared statement id.
 - [improvement] JAVA-1670: Support user-provided JMX ports for CCMBridge.
+- [improvement] JAVA-1661: Avoid String.toLowerCase if possible in Metadata.
 
 
 ### 3.3.1
--- a/driver-core/src/test/java/com/datastax/driver/core/MetadataTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/MetadataTest.java
@@ -125,6 +125,7 @@ public class MetadataTest extends CCMTestsSupport {
     public void handleId_should_lowercase_unquoted_alphanumeric_identifiers() {
         assertThat(Metadata.handleId("FooBar1")).isEqualTo("foobar1");
         assertThat(Metadata.handleId("Foo_Bar_1")).isEqualTo("foo_bar_1");
+        assertThat(Metadata.handleId("foo_bar_1")).isEqualTo("foo_bar_1");
     }
 
     @Test(groups = "unit")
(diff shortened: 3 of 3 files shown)


driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -125,27 +132,36 @@ static String handleId(String id) {
...
 
-        if (isAlphanumeric(id))
+        boolean isAlphanumericLowCase = true;
+        boolean isAlphanumeric = true;
+        for (int i = 0; i < id.length(); i++) {
+            char c = id.charAt(i);
+            if (c >= 65 && c <= 90) { // A-Z
+                isAlphanumericLowCase = false;
+            } else {
+                if (!(
```
olim7t: Nit: you can use `else if` here.

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -274,7 +290,7 @@ public static String quote(String id) {
      * CQL keyword, {@code false} otherwise.
      */
     public static boolean isReservedCqlKeyword(String id) {
-        return id != null && RESERVED_KEYWORDS.contains(id.toLowerCase());
+        return id != null && RESERVED_KEYWORDS.contains(id);
```
olim7t: Unfortunately this won't work for mixed-case keywords, as in `Local_Quorum`. I know this is a silly edge case, but it's possible so we have to handle it all the way.

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -125,27 +132,36 @@ static String handleId(String id) {
...
+                    isAlphanumeric = false;
+                    isAlphanumericLowCase = false;
+                    break;
+                }
+            }
+        }
+
+        if (isAlphanumericLowCase) {
+            return id;
+        }
```
olim7t: 👍  agreed, nice improvement.
Especially since hopefully 99% of people use lower case.

olim7t (changes requested):
+1 for the first change, but the second (reserved keywords) won't work with mixed case.

Maybe we could use a custom case-insensitive hashcode to do a first quick check before doing the whole comparison, I'm not convinced it will improve things but we can benchmark it.

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
netudima: yes, agree, will update with the change

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
netudima: yes, agree, missed the case, will think how is better to deal with it

olim7t: To expand on my suggestion in the review summary, we can try using this method:
```java
private static int caseInsensitiveHash(String s) {
    int hash = 17;
    for (int i = 0; i < s.length(); i++) {
        char c = s.charAt(i);
        if (c >= 65 && c <= 90) { // A-Z
            c += 32;
        }
        hash = 31 * hash + c;
    }
    return hash;
```
We turn `RESERVED_KEYWORDS` into a map of caseInsensitiveHash => lower-case keyword.
Then `isReservedKeywords` can be implemented as:
```java
if (id == null) {
  return false;
}
int hash = caseInsensitiveHash(id);
String keyword = RESERVED_KEYWORDS.get(hash);
return keyword != null && id.toLowerCase().equals(keyword);
```
Again I'm not sure if this will improve performance so we'll have to benchmark it.

netudima: Thank you for the details, I got the idea. Checking different options with JMH, will publish results soon.

netudima: Some initial results:

https://gist.github.com/netudima/6c60455a370de4d44bb0879f36f4641b 
```
# JMH version: 1.19
# VM version: JDK 1.8.0_151, VM 25.151-b12
# VM invoker: C:\Program Files\Java\jre1.8.0_151\bin\java.exe
# VM options: <none>
# Warmup: 10 iterations, 1 s each
# Measurement: 30 iterations, 1 s each
# Timeout: 10 min per iteration
# Threads: 2 threads, will synchronize iterations
# Benchmark mode: Average time, time/op

Benchmark                                                                      Mode  Cnt    Score   Error  Units
MetadataBenchmark.equalsWrapper_NonReservedKeywordMixedCase                    avgt   30   35.298 ± 1.840  ns/op
MetadataBenchmark.equalsWrapper_NonReservedKeywordUpperCase                    avgt   30   39.861 ± 0.864  ns/op
MetadataBenchmark.equalsWrapper_ReservedKeywordMixedCase                       avgt   30   66.610 ± 1.391  ns/op
MetadataBenchmark.equalsWrapper_reservedKeywordUpperCase                       avgt   30   92.498 ± 1.685  ns/op
MetadataBenchmark.fullyUpperOrFullyLowerCaseCheck_NonReservedKeywordMixedCase  avgt   30   40.995 ± 0.725  ns/op
MetadataBenchmark.fullyUpperOrFullyLowerCaseCheck_NonReservedKeywordUpperCase  avgt   30   30.344 ± 0.538  ns/op
MetadataBenchmark.fullyUpperOrFullyLowerCaseCheck_ReservedKeywordMixedCase     avgt   30   70.420 ± 3.180  ns/op
MetadataBenchmark.fullyUpperOrFullyLowerCaseCheck_ReservedKeywordUpperCase     avgt   30   20.327 ± 0.635  ns/op
MetadataBenchmark.toLowCase_NonReservedKeywordMixedCase                        avgt   30  130.456 ± 2.344  ns/op
MetadataBenchmark.toLowCase_NonReservedKeywordUpperCase                        avgt   30  161.837 ± 2.163  ns/op
MetadataBenchmark.toLowCase_ReservedKeywordMixedCase                           avgt   30  128.490 ± 4.897  ns/op
MetadataBenchmark.toLowCase_reservedKeywordUpperCase                           avgt   30  142.862 ± 2.836  ns/op
```

Where:

1) equalsWrapper - wrap String with case-insensitive equals and hashcode
```java

private static final Set<IgnoreCaseWrapper> RESERVED_KEYWORDS_WITH_CUSTOM_EQUAL = ignoreCaseImmutableSet(RESERVED_KEYWORDS);

private static IgnoreCaseWrapper ignoreCase(String str) {
    return new IgnoreCaseWrapper(str);
}

private static class IgnoreCaseWrapper {
    private final String str;

    private IgnoreCaseWrapper(String str) {
        this.str = str;
    }

    @Override
        if (this == o) return true;
        if (!(o instanceof IgnoreCaseWrapper)) return false;
        IgnoreCaseWrapper anotherWrapper = (IgnoreCaseWrapper) o;

        if (str.length() != anotherWrapper.str.length()) return false;

        String anotherString = ((IgnoreCaseWrapper) o).str;

        for (int i = 0; i < str.length(); i++) {
            char c1 = str.charAt(i);
            char c2 = anotherString.charAt(i);
            if (c1 == c2) {
                continue;
            }
            char low1 = toLowerCaseAscii(c1);
            char low2 = toLowerCaseAscii(c2);
            if (low1 == low2) {
                continue;
            }
            return false;
        }
        return true;
    }

    private static char toLowerCaseAscii(char c) {
        if (c >= 65 && c <= 90) { // A-Z
            c += 32; // convert to low case
        }
        return c;
    }

    @Override
    public int hashCode() {
        int hashCode = 17;
        for (int i = 0; i < str.length(); i++) {
            char c = toLowerCaseAscii(str.charAt(i));
            hashCode = 31 * hashCode + c;
        }
        return hashCode;
    }
}
```

2) fullyUpperOrFullyLowerCaseCheck - check does String contain only Upper case or only Lower case symbols and use plain set contains check in this case, use equalsWrapper otherwise.
```java
public static boolean isReservedCqlKeyword(String id) {
    if(id == null) return false;
    if (isFullyUpperOrFullyLowerCase(id)) {
        return RESERVED_KEYWORDS.contains(id);
    }
    return RESERVED_KEYWORDS_WITH_CUSTOM_EQUAL.contains(ignoreCase(id));
}

private static boolean isFullyUpperOrFullyLowerCase(String s) {
    boolean isAlphanumericLowCase = true;
    boolean isAlphanumericUpperCase = true;
    for (int i = 0; i < s.length(); i++) {
        char c = s.charAt(i);
        if (c >= 65 && c <= 90) { // A-Z
            isAlphanumericLowCase = false;
            if (!isAlphanumericUpperCase) {
                break;
            }
        } else if (c >= 97 && c <= 122) { // a-z
            isAlphanumericUpperCase = false;
            if (!isAlphanumericLowCase) {
                break;
            }
        } else if (!(
                (c >= 48 && c <= 57) // 0-9
                        || (c == 95) // _ (underscore)
        )) {
            isAlphanumericLowCase = false;
            isAlphanumericUpperCase = false;
            break;
        }
    }

    return isAlphanumericLowCase || isAlphanumericUpperCase;
}
```
3) toLowCase - the current behavior with toLowerCase conversion
```java
public static boolean isReservedCqlKeyword(String id) {
    return id != null && RESERVED_KEYWORDS_OLD.contains(id.toLowerCase());
}
```

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -289,8 +350,9 @@ public static String quote(String id) {
      * @return {@code true} if the given identifier is a known reserved
      * CQL keyword, {@code false} otherwise.
      */
+
     public static boolean isReservedCqlKeyword(String id) {
-        return id != null && RESERVED_KEYWORDS.contains(id);
+        return id != null && IGNORE_CASE_RESERVED_KEYWORDS.contains(ignoreCase(id));
```
olim7t: I agree that a wrapper is cleaner, but since we're trying to squeeze the last bit of performance, I wonder if we should avoid this extra object instantiation for each method call. That's why I suggested a separate method for the hash and a map for the reserved keywords. WDYT?

netudima: yes, I thought about the extra object allocation. The problem is what in this case:
```java
int hash = caseInsensitiveHash(id);
String keyword = RESERVED_KEYWORDS.get(hash);
```
we also have a allocation - auto-boxing of primitive int. There is a pool of pre-allocated objects in Integer.valueOf ([-128 ; 127]) but in cause of hashes when values are spread across int range there is no large benefit from it. We can use a special map with primitive int keys to avoid it - will be ok to use https://netty.io/4.0/api/io/netty/util/collection/IntObjectMap.html in the driver code (there are other libraries like Trove but Netty is already present as a dependency)?

Regarding:
```java 
String keyword = RESERVED_KEYWORDS.get(hash)
```
Do you suggest here to use 'perfect hash functions' without collisions on the constant set of keywords?

olim7t: > we also have a allocation - auto-boxing of primitive int

Good point.

> Do you suggest here to use 'perfect hash functions' without collisions on the constant set of keywords?

I hadn't completely thought this through, we'd need a multimap (map of lists) to handle collisions. At this point I'm not sure if this is worth exploring.

netudima: Have implemented a non-wrap option using io.netty.util.collection.IntObjectHashMap.
Also because there is a granular control on comparison logic for this way - used char[] to avoid boundary checks in equals logic (saves about 10 ns).

Results:

```
# JMH version: 1.19
# VM version: JDK 1.8.0_151, VM 25.151-b12
# VM invoker: C:\Program Files\Java\jre1.8.0_151\bin\java.exe
# VM options: <none>
# Warmup: 5 iterations, 1 s each
# Measurement: 30 iterations, 1 s each
# Timeout: 10 min per iteration
# Threads: 2 threads, will synchronize iterations
# Benchmark mode: Average time, time/op

Benchmark                                                    Mode  Cnt    Score   Error  Units
MetadataBenchmark.equalsWrapper_NonReservedKeywordMixedCase  avgt   30   49.738 ± 1.313  ns/op
MetadataBenchmark.equalsWrapper_NonReservedKeywordUpperCase  avgt   30   59.430 ± 3.230  ns/op
MetadataBenchmark.equalsWrapper_ReservedKeywordMixedCase     avgt   30   75.837 ± 4.706  ns/op
MetadataBenchmark.equalsWrapper_reservedKeywordUpperCase     avgt   30  100.812 ± 2.402  ns/op

MetadataBenchmark.hashCheck_NonReservedKeywordMixedCase      avgt   30   45.817 ± 1.289  ns/op
MetadataBenchmark.hashCheck_NonReservedKeywordUpperCase      avgt   30   47.084 ± 1.760  ns/op
MetadataBenchmark.hashCheck_ReservedKeywordMixedCase         avgt   30   72.546 ± 1.198  ns/op
MetadataBenchmark.hashCheck_reservedKeywordUpperCase         avgt   30   93.308 ± 1.769  ns/op
```

Code:

```java
import io.netty.util.collection.IntObjectHashMap;
import io.netty.util.collection.IntObjectMap;

private static final IntObjectMap<List<char[]>> KEYWORD_HASHES = new IntObjectHashMap<List<char[]>>();
static {
    for (String keyword : RESERVED_KEYWORDS) {
        char[] keywordAsCharArray = keyword.toLowerCase().toCharArray();
        int hash = ignoreCaseHashCode(keyword);
        List<char[]> keywordsPerHash = KEYWORD_HASHES.get(hash);
        if (keywordsPerHash == null) {
            keywordsPerHash = new ArrayList<char[]>();
            keywordsPerHash.add(keywordAsCharArray);
            KEYWORD_HASHES.put(hash, keywordsPerHash);
        } else {
            keywordsPerHash.add(keywordAsCharArray);
        }
    }
}
    
// keyword is expected as a second argument always in low case
private static boolean equalsIgnoreCaseAscii(String str1, char[] str2LowCase) {
    if (str1.length() != str2LowCase.length) return false;

    for (int i = 0; i < str1.length(); i++) {
        char c1 = str1.charAt(i);
        char c2Low = str2LowCase[i];
        if (c1 == c2Low) {
            continue;
        }
        char low1 = toLowerCaseAscii(c1);
        if (low1 == c2Low) {
            continue;
        }
        return false;
    }
    return true;
}

private static char toLowerCaseAscii(char c) {
    if (c >= 65 && c <= 90) { // A-Z
        c = (char) (c ^ 0x20); // convert to lower case
    }
    return c;
}

private static int ignoreCaseHashCode(String str) {
    int hashCode = 17;
    for (int i = 0; i < str.length(); i++) {
        char c = toLowerCaseAscii(str.charAt(i));
        hashCode = 31 * hashCode + c;
    }
    return hashCode;
}

public static boolean isReservedCqlKeyword(String id) {
    if(id == null) {
        return false;
    }
    int hash = ignoreCaseHashCode(id);
    List<char[]> keywords = KEYWORD_HASHES.get(hash);
    if (keywords == null) {
        return false;
    } else {
        for (char[] keyword : keywords) {
            if (equalsIgnoreCaseAscii(id, keyword)) {
                return true;
            }
        }
        return false;
    }
}
```

netudima: That is the preference for the last 2 options (ignore case wrapper vs hashes in primitive int map from netty-common)?

olim7t: The primitive int map seems a bit faster, let's go with that.
> used char[] to avoid boundary checks in equals logic

Do you mean the checks made by `charAt(i)`?

netudima: yes, that logic
ok, thank you, I will update the patch with the primitive int map option.

olim7t:
Rebased on 3.3.x. I also renamed the test methods and reordered the methods in `Metadata`. This is good to go, thanks 👍

netudima:
Thank you :-)

PR #906, opened 2017-11-03
JAVA-1662: Raise default request timeout

olim7t:


core/src/main/resources/reference.conf
```diff
@@ -239,7 +239,13 @@ datastax-java-driver {
...
     # This option can be changed at runtime, the new value will be used for requests issued after
     # the change. It can be overridden in a profile.
-    timeout = 500 milliseconds
+    #
+    # By default, this value is set pretty high to ensure that DDL queries don't time out, in order
+    # to provide the best experience for new users trying the driver with the out-of-the-box
+    # configuration.
+    # For any serious deployment, we recommend that you use separate configuration profiles for DDL
+    # and DML; you can then set the DML timeout much lower (down to a few milliseconds if needed).
+    timeout = 2 seconds
```
olim7t, line 248: I tested against a local 3-node cluster, 2 seconds seems good enough.

PR #909, opened 2017-11-09
Upgrade to simulacron 0.6.0

tolbertam:
Includes update to jackson-databind 2.9.2 which has a fix for [CVE-2017-15095](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-15095).  Since jackson is only used by our test code this is not really a concern.


tolbertam:
failing unit test looks unrelated to change.

olim7t:
> failing unit test looks unrelated to change.

Agreed, it will probably get fixed by #906

PR #910, opened 2017-11-11
JAVA-1670 Added support for configuring and getting JmxPort per node

jolynch:
This makes it possible to test multi node clusters and send the
jmx commands to specific nodes for testing purposes (e.g. compaction,
flushing, monitoring, repairs, etc ...)

Per the contributing guidelines I included changelog, tests, and code changes. Let me know if I missed something :-)

I also made CCMAccess's constructor protected so that subclasses can exist (that way folks can extend that class without copy pasting it).


tolbertam (approved):
This looks great, thanks! 👍

olim7t:
Merged for 3.3.2, thank you 👍

PR #913, opened 2017-11-28
Update Manual and API links version to 3.3

marcogalluzzi:
Both Manual and API links point at version 3.2. I've changed to the correct 3.3 version.


olim7t (approved):
Good catch, thanks.

PR #914, opened 2017-12-03
JAVA-1633: Handle per-request keyspace in protocol v5

olim7t:


core/src/main/java/com/datastax/oss/driver/internal/core/ProtocolFeature.java
```diff
@@ -24,5 +24,12 @@
    * @see <a href="https://issues.apache.org/jira/browse/CASSANDRA-7304">CASSANDRA-7304</a>
    */
   UNSET_BOUND_VALUES,
+
+  /**
+   * The ability to override the keyspace on a per-request basis.
+   *
+   * @see <a href="https://issues.apache.org/jira/browse/CASSANDRA-10145">CASSANDRA-10145</a>
```
tolbertam, line 31: Really nice to put the C* tickets here.  I imagine as time goes on and these protocol features build up, it will be really nice to have a mapping of features to tickets.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -107,14 +108,19 @@ static Message toMessage(
...
+    CqlIdentifier keyspace = statement.getKeyspace();
     if (statement instanceof SimpleStatement) {
       SimpleStatement simpleStatement = (SimpleStatement) statement;
-
       if (!simpleStatement.getPositionalValues().isEmpty()
           && !simpleStatement.getNamedValues().isEmpty()) {
         throw new IllegalArgumentException(
             "Can't have both positional and named values in a statement.");
       }
+      if (keyspace != null
```
tolbertam, line 119: This is a nice improvement over the 3.x implementation as since we previously overloaded `setKeyspace` for routing and per-query keyspace we didn't know for sure how the method was used.  Since there are two explicit methods now it's nice that we can emit an error instead of relying on the server (which wasn't fool proof).

tolbertam (commented):
This looks great so far!  I noticed there's a todo for repreparing so will hold off on approving review.

olim7t:
> I noticed there's a todo for repreparing

Good catch! I had actually missed it.

tolbertam (approved):
Looks great! 👍

PR #915, opened 2017-12-08
JAVA-1651: Add NO_COMPACT startup option

tolbertam:
For [JAVA-1651](https://datastax-oss.atlassian.net/browse/JAVA-1651) in support of [CASSANDRA-10857](https://issues.apache.org/jira/browse/CASSANDRA-10857).

Since this requires creating a table with the thrift API, i've excluded tests, but I did test it manually in the following manner:

1. Create a cluster running 2.1:

    ```bash
    ccm create -n 1 -v 2.1.17 10857_test -s

2. Define the schema and insert some data using cassandra-cli:

    ```bash
    create keyspace simple;
    use simple;
    create column family foo 
        with comparator = UTF8Type
        and default_validation_class = UTF8Type
        and column_metadata = [
            {column_name: bar, validation_class: Int32Type, index_type: KEYS},
            {column_name: baz, validation_class: UUIDType, index_type: KEYS}
        ];
    set foo['c0']['bar'] = 10;
    set foo['c0']['baz'] = 33cb65d4-6721-4ca8-854f-1f020c5353cb;
    set foo['c0']['yak'] = cafedead;
    ```

    Both `bar` and `baz` are part of column metadata and will show up as static columns in cql tables,  `yak` should not unless `NO_COMPACT` is provided on startup.

3.  upgrade to 2.2:

    ```bash
    ccm stop
    ccm node1 setdir -v 2.2.11
    ccm node1 updateconf start_rpc:true
    ccm start --wait-for-binary-proto
    ccm node1 nodetool upgradesstables
    ```

4. Upgrade to 3.0 (local install dir of cassandra-3.0):

    ```bash
    ccm stop
    ccm node1 setdir --install-dir=/Users/atolbert/Documents/Projects/cassandra-3.0
    ccm start --wait-for-binary-proto
    ccm node1 nodetool upgradesstables
    ```

5. To test with cqlsh first, try without `--no_compact` to observe `yak` is missing:

    ```bash
    ccm node1 cqlsh
    cqlsh> select * from simple.foo where key=0xc0;

     key  | bar | baz
    ------+-----+--------------------------------------
     0xc0 |  10 | 33cb65d4-6721-4ca8-854f-1f020c5353cb
    ```

6.  Try again with `--no_compact` and observe the precense of `yak`:

    ```bash
    ccm node1 cqlsh --no_compact
    cqlsh> select * from simple.foo where key=0xc0;

     key  | column1 | bar | baz                                  | value
    ------+---------+-----+--------------------------------------+----------
     0xc0 |     yak |  10 | 33cb65d4-6721-4ca8-854f-1f020c5353cb | cafedead
    ```

7. Verify the same behavior with java driver:

    ```java
    Cluster cluster = Cluster.builder().withNoCompact(true).addContactPoint("127.0.0.1").build();
    try {
        Row row = session.execute("select * from simple.foo where key=0xc0").one();
        System.out.println(row);
    } finally {
        cluster.close();
    }
    ```

    Should print with `withNoCompact()`:

    > Row[java.nio.HeapByteBuffer[pos=0 lim=1 cap=1], yak, 10, 33cb65d4-6721-4ca8-854f-1f020c5353cb, cafedead]

    Should print without `withNoCompact()`:

    > Row[java.nio.HeapByteBuffer[pos=0 lim=1 cap=1], 10, 33cb65d4-6721-4ca8-854f-1f020c5353cb]

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -687,6 +687,7 @@ public class Cluster implements Closeable {
         private boolean metricsEnabled = true;
         private boolean jmxEnabled = true;
         private boolean allowBetaProtocolVersion = false;
+        private boolean noCompact = false;
 
         private Collection<Host.StateListener> listeners;
 
@@ -1296,6 +1297,24 @@ public class Cluster implements Closeable {
         }
 
         /**
+         * Enables the <code>NO_COMPACT</code> startup option.
+         * <p>
+         * When this option is supplied, <code>SELECT</code>, <code>UPDATE</code>, <code>DELETE</code> and
+         * <code>BATCH</code> statements on <code>COMPACT STORAGE</code> tables function in "compatibility" mode which
+         * allows seeing these tables as if they were "regular" CQL tables.
+         * <p>
+         * This option only effects interactions with tables using <code>COMPACT STORAGE<code> and is only supported by
+         * C* 4.0+ and DSE 6.0+.
+         *
+         * @return this builder.
+         * @see <a href="https://issues.apache.org/jira/browse/CASSANDRA-10857">CASSANDRA-10857</a>
+         */
+        public Builder withNoCompact() {
+            this.noCompact = true;
+            return this;
+        }
+
+        /**
          * The configuration that will be used for the new cluster.
          * <p/>
          * You <b>should not</b> modify this object directly because changes made
@@ -1306,7 +1325,7 @@ public class Cluster implements Closeable {
          */
         @Override
         public Configuration getConfiguration() {
-            ProtocolOptions protocolOptions = new ProtocolOptions(port, protocolVersion, maxSchemaAgreementWaitSeconds, sslOptions, authProvider)
+            ProtocolOptions protocolOptions = new ProtocolOptions(port, protocolVersion, maxSchemaAgreementWaitSeconds, sslOptions, authProvider, noCompact)
                     .setCompression(compression);
 
             MetricsOptions metricsOptions = new MetricsOptions(metricsEnabled, jmxEnabled);
--- a/driver-core/src/main/java/com/datastax/driver/core/ProtocolOptions.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ProtocolOptions.java
@@ -99,6 +99,8 @@ public class ProtocolOptions {
     private final SSLOptions sslOptions; // null if no SSL
     private final AuthProvider authProvider;
 
+    private final boolean noCompact;
+
     private volatile Compression compression = Compression.NONE;
 
     /**
@@ -118,7 +120,7 @@ public class ProtocolOptions {
      * @param port the port to use for the binary protocol.
      */
     public ProtocolOptions(int port) {
-        this(port, null, DEFAULT_MAX_SCHEMA_AGREEMENT_WAIT_SECONDS, null, AuthProvider.NONE);
+        this(port, null, DEFAULT_MAX_SCHEMA_AGREEMENT_WAIT_SECONDS, null, AuthProvider.NONE, false);
     }
 
     /**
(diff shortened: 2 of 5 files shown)


driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1295,6 +1296,24 @@ public Builder withNettyOptions(NettyOptions nettyOptions) {
...
+         * <code>BATCH</code> statements on <code>COMPACT STORAGE</code> tables function in "compatibility" mode which
+         * allows seeing these tables as if they were "regular" CQL tables.
+         * <p>
+         * This option only effects interactions with tables using <code>COMPACT STORAGE<code>.
+         *
+         * @param noCompact Whether or not NO_COMPACT should be configured on connections.
+         * @return this builder.
+         * @see <a href="https://issues.apache.org/jira/browse/CASSANDRA-10857">CASSANDRA-10857</a>
+         */
+        public Builder withNoCompact(boolean noCompact) {
```
tolbertam: Wasn't sure if I should have done it this way or to simply have `withNoCompact()` which enables the flag.  The advantage of doing it this way if you can disable the option after the fact, but the downside is you have to provide the boolean value.

driver-core/src/main/java/com/datastax/driver/core/ProtocolOptions.java
```diff
@@ -135,11 +137,30 @@ public ProtocolOptions(int port) {
...
+     * @param protocolVersion the protocol version to use. This can be {@code null}, in which case the
+     *                        version used will be the biggest version supported by the <em>first</em> node the driver connects to.
+     *                        See {@link Cluster.Builder#withProtocolVersion} for more details.
+     * @param sslOptions      the SSL options to use. Use {@code null} if SSL is not
+     *                        to be used.
+     * @param authProvider    the {@code AuthProvider} to use for authentication against
+     *                        the Cassandra nodes.
+     * @param noCompact       whether or not to include the NO_COMPACT startup option.
+     */
+    public ProtocolOptions(int port, ProtocolVersion protocolVersion, int maxSchemaAgreementWaitSeconds, SSLOptions sslOptions, AuthProvider authProvider, boolean noCompact) {
```
tolbertam, line 157: Created additional constructor for backwards compatibility,  but may be overkill.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1295,6 +1296,24 @@ public Builder withNettyOptions(NettyOptions nettyOptions) {
             return this;
         }
 
+        /**
```
tolbertam, line 1299: Following the [python driver implementation](https://github.com/datastax/python-driver/pull/846), NO_COMPACT is not enabled by default.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
olim7t: I'd prefer the no-argument version, in order to follow the same pattern as `withoutMetrics`.

driver-core/src/main/java/com/datastax/driver/core/ProtocolOptions.java
olim7t: lgtm 👍

olim7t (changes requested):

tolbertam:
Thanks for reviewing!  I adjusted the `withNoCompact` option to take no arguments as requested.

olim7t (approved):

ifesdjeen:
:+1: sorry for the delay, was not on top of my stack. The change looks great! 

The only concern I have is that there are no integration tests for this option neither on driver side nor on Cassandra side. Should we port the dtest ones? https://github.com/apache/cassandra-dtest/compare/master...ifesdjeen:10857#diff-f3024f0514c899682b2cd78bae80e75bR1

tolbertam:
Thanks for taking a look @ifesdjeen!

> The only concern I have is that there are no integration tests...

Ah, I guess I didn't realize that I'd be able to populate data without needing a thrift client.  In that case I should be able to add some integration tests, will give it a go within the next week or so 👍

ifesdjeen:
Yes, you can just use "compact storage". It won't cover all the cases, but I at least in the current version of code SuperColumns and thrift tables have to work the same way. In the end, we can add tests on Cassandra side if we see it's not enough.

tolbertam:
> Yes, you can just use "compact storage". It won't cover all the cases, but I at least in the current version of code SuperColumns and thrift tables have to work the same way. In the end, we can add tests on Cassandra side if we see it's not enough.

Took me a bit to come back around to this and now i'm also implementing the nodejs-driver implementation.  I think I understand what you mean.  Since this is supported in C* 3.0.16 and 3.11.2, I should be able to write tests against those versions that use `compact storage` tables and insert rows including values for `column1` and `value`.   I tried that out and it seems to work as I expect, so i'll write some tests around that, thanks!

PR #917, opened 2017-12-12
JAVA-1647: Handle metadata_changed flag in protocol v5

olim7t:
(Has the fix for CASSANDRA-13992 already)

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/cql/PreparedStatement.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/cql/PreparedStatement.java
@@ -15,6 +15,7 @@
  */
 package com.datastax.oss.driver.api.core.cql;
 
+import com.datastax.oss.driver.api.core.CoreProtocolVersion;
 import java.nio.ByteBuffer;
 import java.util.List;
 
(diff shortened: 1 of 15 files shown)


pom.xml
```diff
@@ -50,7 +50,7 @@
       <dependency>
         <groupId>com.datastax.oss</groupId>
         <artifactId>native-protocol</artifactId>
-        <version>1.4.1</version>
+        <version>1.4.2-SNAPSHOT</version>
```
olim7t, line 53: Update required for datastax/native-protocol#19

core/src/main/java/com/datastax/oss/driver/api/core/cql/PreparedStatement.java
```diff
@@ -68,12 +69,40 @@
...
+   * <p>This information is mostly for internal use: with protocol {@link CoreProtocolVersion#V5} or
+   * higher, the driver sends it with every execution of the prepared statement, to validate that
+   * its result metadata is still up-to-date.
+   *
+   * <p>Note: this method returns null for protocol {@link CoreProtocolVersion#V4} or lower;
+   * otherwise, the returned buffer is read-only.
+   *
+   * @see <a href="https://issues.apache.org/jira/browse/CASSANDRA-10786">CASSANDRA-10786</a>
+   */
+  ByteBuffer getResultMetadataId();
```
olim7t, line 85: Users can wrap prepared statements (and pass us a bound statement that references a wrapped prepared statement), so this has to be on the public API. Same for `setResultMetadata`.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -284,13 +289,39 @@ static AsyncResultSet toResultSet(
     }
   }
 
+  private static ColumnDefinitions getResultDefinitions(
+      Rows rows, Statement<?> statement, InternalDriverContext context) {
+    RowsMetadata rowsMetadata = rows.getMetadata();
+    if (rowsMetadata.columnSpecs.isEmpty()) {
+      // If the response has no metadata, it means the request had SKIP_METADATA set, the driver
+      // only ever does that for bound statements.
+      BoundStatement boundStatement = (BoundStatement) statement;
```
tolbertam, line 298: I noticed this causes problems for some tests, but it appears its caused by simulacron not returning metadata for queries that weren't primed.  Instead it should probably return a `Void` result, i'll make that update.

tolbertam: Oh, apparently simulacron already does this, it's just that our tests are priming `noRows()` instead of `void`.  Will fix, although I do wonder if in general it should throw some result metadata on there to be considered a valid select response.

tolbertam: Updated simulacron, when noRows is specified we fake out some result metadata to match C* behavior.

tolbertam (approved):
Looks great 👍

olim7t:
Build fails because it depends on a snapshot version of `native-protocol`.

PR #918, opened 2017-12-21
JAVA-1694: Upgrade to jackson-databind 2.7.9.2 to address CVE-2015-15095 

tolbertam:
For [JAVA-1694](https://datastax-oss.atlassian.net/browse/JAVA-1694).  This doesn't have much impact because jackson-databind is only used in extras and examples, but would be good to update anyways.


olim7t (approved):
👍

PR #919, opened 2017-12-21
JAVA-1707: Add test infrastructure for running DSE clusters with CCM

tolbertam:
Includes the following changes:

* Updates `CcmBridge` so it supports running DSE clusters.
* Add `@DseRequirement` for checing against DSE versions for test requirements.
* Rename ccm.cassandraVersion and ccm.cassandraDirectory to ccm.version and ccm.directory respecitvely.
* Adds ccm.dse, if set assumes DSE cluster.
* Adds support for EverywhereStrategy ([JAVA-1707](https://datastax-oss.atlassian.net/browse/JAVA-1708)), which is currently only supported by DSE, but thought it was simple enough to include.
* Changes integration-tests pom to deploy the tests jar so it may be used in other projects.
* Use generic version of `Cluster<CqlSession>` in tests and introduce system property `cluster.builder` for specifying what class to use to create initial cluster builder instances.   See `DefaultClusterBuilderInstantiator` for base implementation.  Will be useful for running all tests against custom Cluster/Session implementations.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -4,6 +4,7 @@
 
 ### 4.0.0-alpha3 (in progress)
 
+- [improvement] JAVA-1707: Add test infrastructure for running DSE clusters with CCM
 - [bug] JAVA-1715: Propagate unchecked exceptions to CompletableFuture in SyncAuthenticator methods
 - [improvement] JAVA-1714: Make replication strategies pluggable
 - [new feature] JAVA-1647: Handle metadata_changed flag in protocol v5
@@ -19,6 +20,7 @@
 - [improvement] JAVA-1662: Raise default request timeout
 - [improvement] JAVA-1566: Enforce API rules automatically
 - [bug] JAVA-1584: Validate that no bound values are unset in protocol v3
+
 
 ### 4.0.0-alpha2
 
--- a/integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/PerRequestKeyspaceIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/PerRequestKeyspaceIT.java
@@ -35,7 +35,7 @@ import static org.assertj.core.api.Assertions.assertThat;
  * test against a local build, run with
  *
  * <pre>
- *   -Dccm.cassandraVersion=4.0.0 -Dccm.cassandraDirectory=/path/to/cassandra -Ddatastax-java-driver.protocol.version=V5
+ *   -Dccm.version=4.0.0 -Dccm.directory=/path/to/cassandra -Ddatastax-java-driver.protocol.version=V5
  * </pre>
  */
 @Category(ParallelizableTests.class)
--- a/test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/BaseCcmRule.java
+++ b/test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/BaseCcmRule.java
@@ -20,20 +20,19 @@ import com.datastax.oss.driver.api.core.CoreProtocolVersion;
 import com.datastax.oss.driver.api.core.ProtocolVersion;
 import com.datastax.oss.driver.api.testinfra.CassandraRequirement;
 import com.datastax.oss.driver.api.testinfra.CassandraResourceRule;
-import com.datastax.oss.driver.api.testinfra.ccm.CcmBridge;
+import com.datastax.oss.driver.api.testinfra.DseRequirement;
 import org.junit.AssumptionViolatedException;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
+
+import java.util.Optional;
 
 public abstract class BaseCcmRule extends CassandraResourceRule {
 
   protected final CcmBridge ccmBridge;
 
-  private final CassandraVersion cassandraVersion;
-
-  public BaseCcmRule(CcmBridge ccmBridge) {
+  BaseCcmRule(CcmBridge ccmBridge) {
     this.ccmBridge = ccmBridge;
-    this.cassandraVersion = ccmBridge.getCassandraVersion();
     Runtime.getRuntime()
         .addShutdownHook(
             new Thread(
@@ -57,8 +56,28 @@ public abstract class BaseCcmRule extends CassandraResourceRule {
     ccmBridge.remove();
   }
 
+  private Statement buildErrorStatement(
+      CassandraVersion requirement, String description, boolean lessThan, boolean dse) {
+    return new Statement() {
+
+      @Override
+      public void evaluate() {
+        throw new AssumptionViolatedException(
+            String.format(
+                "Test requires %s %s %s but %s is configured.  Description: %s",
+                lessThan ? "less than" : "at least",
+                dse ? "DSE" : "C*",
+                requirement,
+                dse ? ccmBridge.getDseVersion().orElse(null) : ccmBridge.getCassandraVersion(),
+                description));
+      }
+    };
+  }
+
   @Override
   public Statement apply(Statement base, Description description) {
+    // If test is annotated with CassandraRequirement or DseRequirement, ensure configured CCM
+    // cluster meets those requirements.
     CassandraRequirement cassandraRequirement =
         description.getAnnotation(CassandraRequirement.class);
 
@@ -66,21 +85,8 @@ public abstract class BaseCcmRule extends CassandraResourceRule {
       // if the configured cassandra cassandraRequirement exceeds the one being used skip this test.
       if (!cassandraRequirement.min().isEmpty()) {
         CassandraVersion minVersion = CassandraVersion.parse(cassandraRequirement.min());
-        if (minVersion.compareTo(cassandraVersion) > 0) {
-          // Create a statement which simply indicates that the configured cassandra cassandraRequirement is too old for this test.
-          return new Statement() {
-
-            @Override
-            public void evaluate() throws Throwable {
-              throw new AssumptionViolatedException(
-                  "Test requires C* "
-                      + minVersion
-                      + " but "
-                      + cassandraVersion
-                      + " is configured.  Description: "
-                      + cassandraRequirement.description());
-            }
-          };
+        if (minVersion.compareTo(ccmBridge.getCassandraVersion()) > 0) {
+          return buildErrorStatement(minVersion, cassandraRequirement.description(), false, false);
         }
       }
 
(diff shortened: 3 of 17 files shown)


integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/PerRequestKeyspaceIT.java
```diff
@@ -35,7 +35,7 @@
  * test against a local build, run with
  *
  * <pre>
- *   -Dccm.cassandraVersion=4.0.0 -Dccm.cassandraDirectory=/path/to/cassandra -Ddatastax-java-driver.protocol.version=V5
+ *   -Dccm.version=4.0.0 -Dccm.directory=/path/to/cassandra -Ddatastax-java-driver.protocol.version=V5
```
olim7t, line 38: Kudos for thinking of updating this 👍 
There is a similar comment in `PreparedStatementInvalidationIT`.

changelog/README.md
```diff
@@ -17,6 +17,8 @@
 - [improvement] JAVA-1662: Raise default request timeout
 - [improvement] JAVA-1566: Enforce API rules automatically
 - [bug] JAVA-1584: Validate that no bound values are unset in protocol v3
+- [improvement] JAVA-1707: Test infrastructure for running DSE clusters with CCM
+
```
olim7t: Could you move this to the beginning of the changelog, and also change the commit message to the imperative form (see [contribution guidelines](https://github.com/datastax/java-driver/blob/4.x/CONTRIBUTING.md#commits))?

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
```diff
@@ -278,17 +332,20 @@ public static Builder builder() {
   public static class Builder {
     private int[] nodes = {1};
     private final Map<String, Object> cassandraConfiguration = new LinkedHashMap<>();
+    private final Map<String, Object> dseConfiguration = new LinkedHashMap<>();
     private final List<String> jvmArgs = new ArrayList<>();
     private String ipPrefix = "127.0.0.";
-    private CassandraVersion cassandraVersion = CcmBridge.DEFAULT_CASSANDRA_VERSION;
-    private String cassandraDirectory = CcmBridge.DEFAULT_CASSANDRA_DIRECTORY;
+    private CassandraVersion version = CcmBridge.DEFAULT_CASSANDRA_VERSION;
+    private String directory = CcmBridge.DEFAULT_CASSANDRA_DIRECTORY;
```
olim7t: The builder does not provide any way to change those two fields.
If there is a case for controlling them programatically, we should have builder methods. If system properties are enough, we should get rid of all the fields and rely exclusively on the two constants.
No strong feeling but I would say the latter, wdyt?

EDIT: same for `dse` actually

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
```diff
@@ -49,29 +50,39 @@
...
-      System.getProperty("ccm.cassandraDirectory");
+  public static final String DEFAULT_CASSANDRA_DIRECTORY = System.getProperty("ccm.directory");
+
+  public static final Boolean DEFAULT_DSE_ENABLEMENT;
+
+  static {
+    String dseProperty = System.getProperty("ccm.dse");
+    DEFAULT_DSE_ENABLEMENT =
+        dseProperty != null && (dseProperty.isEmpty() || Boolean.parseBoolean(dseProperty));
+  }
```
olim7t: I think this can be simplified as:
```java
public static final boolean DEFAULT_DSE_ENABLEMENT = Boolean.getBoolean("ccm.dse");
```

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
```diff
@@ -139,18 +155,45 @@ private CcmBridge(
...
+  public Optional<CassandraVersion> getDseVersion() {
+    return dse ? Optional.of(version) : Optional.empty();
   }
 
   public CassandraVersion getCassandraVersion() {
-    return cassandraVersion;
+    if (!dse) {
+      return version;
+    } else {
+      CassandraVersion cassandraVersion = version;
```
olim7t: Why the extra local variable?

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
```diff
@@ -139,18 +155,45 @@ private CcmBridge(
...
+          // 5.0 = 3.0.x
+          return CassandraVersion.parse("3.0.15");
+        } else {
+          // 5.1 == 3.10
+          return CassandraVersion.parse("3.10");
+        }
+      } else {
+        // assume anything older is 2.1.19
+        return CassandraVersion.parse("2.1.19");
+      }
```
olim7t: No big deal but I think it would be slightly more readable using constants and comparisons:

```java
CassandraVersion stableVersion = version.nextStable();
if (stableVersion.compareTo(V6_0_0) >= 0) {
  return CassandraVersion.parse("4.0.0");
} else if (stableVersion.compareTo(V5_1_0) >= 0) {
  return CassandraVersion.parse("3.10");
} else if (stableVersion.compareTo(V5_0_0) >= 0) {
  return CassandraVersion.parse("3.0.15");
} else {
  return CassandraVersion.parse("2.1.19");
}
```
And then the comments are not needed.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/BaseCcmRule.java
```diff
@@ -20,20 +20,18 @@
...
+import com.datastax.oss.driver.api.testinfra.DseRequirement;
+import java.util.Optional;
 import org.junit.AssumptionViolatedException;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
 
 public abstract class BaseCcmRule extends CassandraResourceRule {
 
-  protected final CcmBridge ccmBridge;
+  final CcmBridge ccmBridge;
```
olim7t: why?

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/BaseCcmRule.java
```diff
@@ -88,33 +89,86 @@ public void evaluate() throws Throwable {
...
           };
         }
       }
     }
+
+    DseRequirement dseRequirement = description.getAnnotation(DseRequirement.class);
+    if (dseRequirement != null) {
+      Optional<CassandraVersion> dseVersionOption = ccmBridge.getDseVersion();
+      if (!dseVersionOption.isPresent()) {
+        throw new AssumptionViolatedException("Test Requires DSE but C* is configured.");
```
olim7t: Why can we throw directly here, but in the other cases we have to build a statement?

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/BaseCcmRule.java
```diff
@@ -59,24 +57,27 @@ protected void after() {
...
+            public void evaluate() {
               throw new AssumptionViolatedException(
                   "Test requires C* "
                       + minVersion
                       + " but "
-                      + cassandraVersion
+                      + ccmBridge.getCassandraVersion()
                       + " is configured.  Description: "
                       + cassandraRequirement.description());
             }
```
olim7t: This is repeated 4 times, maybe we should extract a method that creates the Statement.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/cluster/ClusterRuleBuilder.java
```diff
@@ -15,12 +15,13 @@
...
 package com.datastax.oss.driver.api.testinfra.cluster;
 
+import com.datastax.oss.driver.api.core.cql.CqlSession;
 import com.datastax.oss.driver.api.core.metadata.NodeStateListener;
 import com.datastax.oss.driver.api.testinfra.CassandraResourceRule;
 import com.datastax.oss.driver.api.testinfra.ccm.CcmRule;
 import com.datastax.oss.driver.api.testinfra.simulacron.SimulacronRule;
 
-public class ClusterRuleBuilder {
+public class ClusterRuleBuilder<T extends CqlSession> {
```
olim7t: If we plan to extend this, we should introduce a "self" type to use as the return of the *with* methods.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationStrategy.java
```diff
@@ -34,6 +34,8 @@ static ReplicationStrategy newInstance(Map<String, String> replicationConfig, St
         return new SimpleReplicationStrategy(replicationConfig);
       case "org.apache.cassandra.locator.NetworkTopologyStrategy":
         return new NetworkTopologyReplicationStrategy(replicationConfig, logPrefix);
+      case "org.apache.cassandra.locator.EverywhereStrategy":
+        return new EverywhereStrategy();
```
olim7t: All things considered, I think it would be cleaner to introduce an overridable factory. I can do it in a separate pull request if you want.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/cluster/ClusterUtils.java
```diff
@@ -55,25 +57,57 @@
...
+
+  @SuppressWarnings("unchecked")
+  public static <T extends CqlSession> ClusterBuilder<?, ? extends Cluster<T>> baseBuilder() {
+    try {
+      Class<?> clazz = Class.forName(CLUSTER_BUILDER_CLASS);
+      Method m = clazz.getMethod("builder");
+      return (ClusterBuilder<?, ? extends Cluster<T>>) m.invoke(null);
+    } catch (Exception e) {
+      e.printStackTrace();
+      // TODO: Logger
```
olim7t: Is it just a matter of declaring an SLF4J logger?

olim7t (changes requested):
👍  makes sense!
A few comments inline.

tolbertam:
Thanks for reviewing!  I'll make the suggested changes today

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/cluster/ClusterUtils.java
tolbertam: Err woops, yes, it's pretty simple just forgot to do it, will fix.

changelog/README.md
tolbertam: Forgot about that change, will do!

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationStrategy.java
tolbertam: :+1: to doing that in a separate PR,  just did this here to get things working in the near term.  Will back it out of my final commits after I finish testing everything.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/PerRequestKeyspaceIT.java
tolbertam: Ah yes, will adjust that too.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/BaseCcmRule.java
tolbertam: This is an artifact of my initial stab (which required accessing ccm bridge in the subclasses), that's no longer needed so I should revert this change.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/BaseCcmRule.java
tolbertam: Good idea, this code is a big repetitive.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
tolbertam: :+1: on the latter.  Originally I wanted to make the version/directory/dse enablement API configurable, but I think it would be better to avoid tests that use and rely on specifying versions, and rather just only run them when that version is configured.    Will fix.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/BaseCcmRule.java
tolbertam: ah yes, this looks like a bug, will fix.

tolbertam:
Processed feedback, still TODO:

* [ ] Update commit message (will be taken care when ready to merge)
* [x] Remove EverywhereStrategy (will remove after rebasing when [JAVA-1714](https://datastax-oss.atlassian.net/browse/JAVA-1714 is merged)
* [x] Some more testing with dse (will do that today)

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/cluster/ClusterUtils.java
```diff
@@ -53,27 +59,62 @@
...
+  }
+
+  public static String getConfigPath() {
+    try {
+      Class<?> clazz = Class.forName(CLUSTER_BUILDER_CLASS);
+      Method m = clazz.getMethod("configPath");
+      return (String) m.invoke(null);
+    } catch (Exception e) {
+      e.printStackTrace();
+      // TODO: Logger
```
olim7t: There's one more here ;-)

tolbertam: :man_facepalming: , good catch, will fix.

tolbertam:
Rebased on 4.x and removed `EverywhereStrategy`, doing a little more testing.

tolbertam:
Finished testing and this should be good to go 👍

olim7t (approved):

PR #922, opened 2018-01-02
JAVA-1677: Warn if auth is configured on the client but not the server

olim7t:


tolbertam (approved):
Looks great!

>     25960  [cluster2-nio-worker-0] WARN  com.datastax.driver.core.Connection - /127.0.1.1:38267 did not send an authentication challenge; This is suspicious because the driver expects authentication (configured auth provider = com.datastax.driver.core.PlainTextAuthProvider)

I imagine this could get pretty annoying as it would happen for every connection, but that's probably a good thing as the user should be aware of this.

olim7t:
> I imagine this could get pretty annoying as it would happen for every connection, but that's probably a good thing as the user should be aware of this.

Yes, I can't think of a legitimate reason to do it on purpose, so it shouldn't be a problem if the logs are a bit verbose. We can always add a way to disable the warning (like in 4.0) later.

PR #923, opened 2018-01-02
JAVA-1672: Remove schema agreement check when repreparing on up

olim7t:


tolbertam (approved):
The rationale on the jira ticket makes sense :+1:

PR #925, opened 2018-01-02
JAVA-1683: Add metrics to track writes to nodes

olim7t:


tolbertam:
Some thoughts:

* Per-host metrics could be a problem if you have a lot of cassandra nodes.  If you have JMX metrics enabled, it creates additional attributes which could have some overhead.  Also if reporting to a metrics server like graphite, this could increase the storage requirements.  We could consider adding a new configuration flag for enabling/disabling host specific metrics, but not sure if I'm over-worrying there.

* `Meter` might be a nice enhancement over `Counter` so you could also additionally monitor rates, that might create some additional overhead to my first point though.

PR #927, opened 2018-01-04
JAVA-1229: Allow specifying the keyspace for individual queries

olim7t:
Contains the changes of #697, rebased on the 3.x branch, with additional changes:
* given that the keyspace can now be specified at the last moment (when invoking the mapper operation), init time schema validation is not possible anymore. For example, it's possible to have an entity with no keyspace defined no logged keyspace on the session either, and only target specific keyspaces later. Removed the validations and corresponding tests.
* in the same vein, it's not possible to reliably react to schema changes anymore (JAVA-1126), because the declared keyspace could be different from the one that is used at runtime. Reverted this ticket, it should be pretty obvious that things won't work if you change the schema at runtime.
* actually, even exposing `Mapper.getTableMetadata` is questionable now, but we can't remove it since it's part of the public API. I've left it and it returns a "default" metadata if we have enough information to retrieve it (using the keyspace from the annotation or the logged session).
* extracted the tests from `MapperOptionTest` to a dedicated class.


tolbertam:
> in the same vein, it's not possible to reliably react to schema changes anymore (JAVA-1126), because the declared keyspace could be different from the one that is used at runtime. Reverted this ticket, it should be pretty obvious that things won't work if you change the schema at runtime.

Since UDTs are still keyspace bound (we throw an exception if `@UDT` `keyspace` annotation is empty and no keyspace tied to session), is there any reason that logic couldn't still exist?  It looks like `udtCodecs` are always still present, whereas table metadata is not be present if no keyspace is specified on the annotation.
  
I think leaving the schema change logic there in general is still a good thing, to warn when these changes happen if metadata happens to have been present.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperKeyspaceOptionTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+    private MappingManager mappingManager;
+    private Mapper<User> mapper;
+    private String otherKeyspace;
+
+    @Override
+    public void onTestContextInitialized() {
+        otherKeyspace = TestUtils.generateIdentifier("ks_");
+        execute("CREATE TABLE user (key int primary key, v text)",
+                String.format("CREATE KEYSPACE IF NOT EXISTS %s WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1}", otherKeyspace),
```
tolbertam: `IF NOT EXISTS` requires 2.0, so would be good to annotate the class with `CassandraVersion(major="2.0")`

olim7t:
> I think leaving the schema change logic there in general is still a good thing, to warn when these changes happen if metadata happens to have been present.

The problem is that we don't know if the metadata that happens to be present is the one (or the only one) that is actually used. So we might end up detecting changes to the wrong schema, or conversely not detecting changes to a schema that will get used at runtime. I think it shows that we were trying to go too far with 1126, for a use case that is frankly a non-issue (it's pretty obvious that things can go bad if you change the schema at runtime, much like with regular CQL queries).

> Since UDTs are still keyspace bound

They shouldn't be. Looking at how we can make that work.

olim7t:
Just force-pushed a different approach: the keyspace is specified when creating the `Mapper`, not for each operation. This is much cleaner and allows us to keep all the validations.

This forces client code to have a `Mapper` instance for each keyspace, but that's not a big issue IMO.

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -111,17 +111,19 @@ Host getHost(InetSocketAddress address) {
...
      * We expect client-facing APIs to behave like cqlsh, that is:
-     * - identifiers that are mixed-case or contain special characters should be quoted.
-     * - unquoted identifiers will be lowercased: getKeyspace("Foo") will look for a keyspace named "foo"
+     * <ul>
+     * <li>identifiers that are mixed-case or contain special characters should be quoted.</li>
+     * <li>unquoted identifiers will be lowercased: getKeyspace("Foo") will look for a keyspace named "foo".</li>
+     * </ul>
      */
-    static String handleId(String id) {
+    public static String handleId(String id) {
```
olim7t: This was already duplicated in the query builder; now that it's used in 3 places, I think it makes sense to make it public.

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperKeyspaceTest.java
```diff
@@ -0,0 +1,273 @@
...
+    public static class UserWithKeyspace extends User<AddressWithKeyspace> {
+        public UserWithKeyspace() {
+        }
+
+        public UserWithKeyspace(String name, AddressWithKeyspace address) {
+            super(name, address);
+        }
+
+        @Override
+        public AddressWithKeyspace getAddress() {
```
olim7t, line 253: This getter has to be overridden, otherwise the mapper tries to map to `AddressT`. That's an edge case but I'll open an other ticket.

olim7t: FYI: https://datastax-oss.atlassian.net/browse/JAVA-1719

driver-mapping/src/test/java/com/datastax/driver/mapping/MapperKeyspaceTest.java
```diff
@@ -0,0 +1,273 @@
...
+
+@CassandraVersion("2.0")
+public class MapperKeyspaceTest extends CCMTestsSupport {
+
+    @Override
+    public void onTestContextInitialized() {
+        execute(
+                "CREATE KEYSPACE IF NOT EXISTS MapperKeyspaceTest WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1}",
+                "CREATE TYPE IF NOT EXISTS MapperKeyspaceTest.address(street text)",
+                "CREATE TABLE IF NOT EXISTS MapperKeyspaceTest.user(name text PRIMARY KEY, address address)",
```
tolbertam: Would be good to make the UDT frozen in both tables so it works with all versions 2.1+

tolbertam (approved):
Other than 1 small comment about using frozen UDTs this looks great.  Definitely like expressing the keyspace as input to `mapper` instead of making it an Option.  The implementation looks a lot simpler and we don't lose some of the existing validations.

olim7t:
Merged with the frozen keyword as suggested 👍

PR #928, opened 2018-01-05
JAVA-1714: Make replication strategies pluggable

olim7t:


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/DefaultReplicationStrategyFactory.java
```diff
@@ -0,0 +1,46 @@
...
+public class DefaultReplicationStrategyFactory implements ReplicationStrategyFactory {
+
+  private final String logPrefix;
+
+  public DefaultReplicationStrategyFactory(InternalDriverContext context) {
+    this.logPrefix = context.clusterName();
+  }
+
+  @Override
+  public ReplicationStrategy newInstance(Map<String, String> replicationConfig) {
```
tolbertam, line 31: It'd be nice to have some implementation of this that doesn't throw an IllegalArgumentException when no match is found, that way one could reuse this logic to evaluate these strategies and than add their additional code for additional strategies they implement.

tolbertam (approved):
Had one small comment, but otherwise this lgtm :+1:

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/DefaultReplicationStrategyFactory.java
olim7t: If you test your strategies first and only delegate to `super()` if none matches, it should work no?

tolbertam: Oh hah, can't believe I didn't think of that, yeah that does work great :+1:

PR #929, opened 2018-01-05
JAVA-1715: Propagate unchecked exceptions to future in SyncAuthenticator

tolbertam:
For [JAVA-1715](https://datastax-oss.atlassian.net/browse/JAVA-1715)

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/CompletableFutures.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/CompletableFutures.java
@@ -18,13 +18,14 @@ package com.datastax.oss.driver.internal.core.util.concurrent;
 import com.datastax.oss.driver.api.core.DriverException;
 import com.datastax.oss.driver.api.core.DriverExecutionException;
 import com.google.common.base.Preconditions;
-import com.google.common.base.Throwables;
+
 import java.util.List;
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/CompletableFutures.java
```diff
@@ -120,4 +121,22 @@ public static Throwable getFailed(CompletableFuture<?> future) {
...
+   * Executes a function on the calling thread and returns result in a {@link CompletableFuture}.
+   *
+   * <p>Similar to {@link CompletableFuture#completedFuture} except takes a {@link Supplier} and if
+   * the supplier throws an unchecked exception, the returning future fails with that exception.
+   *
+   * @param supplier Function to execute
+   * @param <T> Type of result
+   * @return result of function wrapped in future
+   */
+  public static <T> CompletableFuture<T> completedFuture(Supplier<T> supplier) {
```
tolbertam: Alternatively could have used `CompletableFuture.supplyAsync` with `MoreExectuors.directExecutor` but this seems more simple.

olim7t (approved):
LGTM, except maybe we should give the method a different name. How about `wrap`?

tolbertam:
:+1: on wrap,  will make that change on Monday morning.

PR #930, opened 2018-01-08
 JAVA-1717: Add metrics to latency-aware policy

olim7t:


driver-core/src/main/java/com/datastax/driver/core/policies/LatencyAwarePolicy.java
```diff
@@ -433,7 +476,29 @@ public TimestampedAverage latencyOf(Host host) {
...
         }
 
+        public void addHost(final Host host) {
+            logger.debug("Adding tracker for {}", host);
+            HostLatencyTracker old = latencies.putIfAbsent(host,
+                    new HostLatencyTracker(scale, (30L * minMeasure) / 100L));
+            if (old == null && hostMetricsEnabled()) {
+                String metricName = MetricsUtil.hostMetricName("LatencyAwarePolicy.latencies.", host);
+                if (!metrics.getRegistry().getNames().contains(metricName)) {
+                    logger.info("Adding gauge " + metricName);
```
tolbertam: Minor nit, shouldn't this also be at debug?

tolbertam (approved):
👍

PR #932, opened 2018-01-12
JAVA-1720: Merge Cluster and Session into a single interface

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/session/Session.java
```diff
@@ -16,25 +16,153 @@
...
+    BlockingOperation.checkNotDriverThread();
+    return CompletableFutures.getUninterruptibly(checkSchemaAgreementAsync());
+  }
+
+  /** Returns a context that provides access to all the policies used by this driver instance. */
+  DriverContext getContext();
+
   /**
    * The keyspace that this session is currently connected to.
    *
```
adutra, line 169: The javadocs for this method contain a few broken links to former `Cluster` class.

core/src/main/java/com/datastax/oss/driver/api/core/cql/CqlSessionBuilder.java
```diff
@@ -13,16 +13,16 @@
  * See the License for the specific language governing permissions and
```
adutra: So, `CqlSession` now resides in `com.datastax.oss.driver.api.core` but `CqlSessionBuilder` is still in `com.datastax.oss.driver.api.core.cql`. Is this intentional?

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/api/core/cql/CqlSessionBuilder.java
olim7t: `CqlSession` used to be in the `cql` package, but since it's now the main entry point to the driver, I thought it would be more appropriate for it to be in the root package. I should at least add an implementation comment to explain that.
As for `CqlSessionBuilder`, it's true that it seems a bit weird to keep it in another package, I don't have strong feelings either way.

olim7t:
I ended up moving `CqlSessionBuilder` to the root package too.

PR #933, opened 2018-01-12
JAVA-1725: Add a getNodeCount method to CCMAccess for easier automation

mfiguiere:
The int[] with the node count isn't accessible easily, whereas it can be useful when manipulating CCMAccess objects in a generic way.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -12,6 +12,7 @@
 - [documentation] JAVA-1380: Add FAQ entry for errors arising from incompatibilities.
 - [improvement] JAVA-1748: Support IS NOT NULL and != in query builder.
 - [documentation] JAVA-1740: Mention C*2.2/3.0 incompatibilities in paging state manual.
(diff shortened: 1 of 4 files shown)


adutra (changes requested):
The reason why we didn't expose this is because nodes can be added or removed dynamically during the lifecycle of a CCM cluster, so the `nodes` field only reflects the initial number of nodes. I think at least you should add a note to the javadocs to warn users about the fact that the numbers returned by this method might not be accurate if nodes were added or removed. Alternatively we could keep track of every node added or removed and update this property accordingly.

olim7t:
I rebased on 3.x, added a note in the javadoc (as suggested by Alexandre), and added a defensive copy of the array (it's not essential because it's never used internally after init, but at least we avoid returning false information if someone mutates it).

olim7t (approved):

tolbertam (approved):

adutra (approved):

PR #935, opened 2018-01-16
3.3.x driver with Java8 + Guava21 compatibility

RussellRC:


adutra:
Hi, thanks for this PR but I don't get the meaning of it. The Java driver has to support all versions of Guava from 16.0.1 to 23 inclusive. Did you notice something wrong when using Guava 21? In theory Guava 21 is fully supported and tested, so I'm a bit surprised.

RussellRC:
Hi @adutra, I am using guava 21 with spring-data-cassandra 2.0.2, which uses cassandra-driver-core 3.3.2, but when I run my application I get a ClassNotFoundException com/google/common/util/concurrent/FutureFallback when the Cluster is being created.

I've searched around and this is a known issue, but going down to guava 19 is not an option for me, neither using the alpha version of cassandra driver 4.

Going through the source code it seems the best option is to fork the code of the driver and do the changes to support guava 21.

tolbertam:
The driver itself doesn't enforce a particular version of guava.  It tries to detect which version of guava you have ony our classpath and uses it.  It's possible that if you have multiple versions of guava on your classpath that this may cause problems, can you double check?

RussellRC:
Hi @tolbertam and @adutra ,

From what I see, the driver does have explicit compile and runtime dependency on guava 19.0 ([pom.xml](https://github.com/datastax/java-driver/blob/3.3.x/pom.xml#L53)).
Confirmed by [compile dependencies on maven repository](https://mvnrepository.com/artifact/com.datastax.cassandra/cassandra-driver-core/3.3.2) and also on the artifact/module dependencies on my project.

As I also mentioned, it seems this has been acknowledged and it's a known issue, which is being addressed on the next java driver: https://datastax-oss.atlassian.net/browse/JAVA-1328

Again, my code has a strong dependency on guava 21, and getting rid of it is looking quite tougher than the forked version of the driver, which I am already testing on my local environment and is working fine at first glance.

I understand approving this PR and creating an official new version of the cassandra driver that supports guava 20+ is probably asking for too much, but so far this is looking as my best option, although I will keep exploring other possibilities.

Thanks, and I really appreciate your feedback.

tolbertam:
> From what I see, the driver does have explicit compile and runtime dependency on guava 19.0 (pom.xml).
Confirmed by compile dependencies on maven repository and also on the artifact/module dependencies on my project.

We need guava 19.0 to compile the project as guava 19.0 has methods present in both older and newer versions of guava.   At runtime, you should be be able to use any guava version between 16.0.1 and 23.

> As I also mentioned, it seems this has been acknowledged and it's a known issue, which is being addressed on the next java driver: https://datastax-oss.atlassian.net/browse/JAVA-1328

JAVA-1328 was fixed in release 3.2.0, and this PR targets the 3.3.x branch which already has the fix.

RussellRC:
I changed something that made my project not pull the latest cassadra driver.  I've fixed that and seems to be working fine now.

Thanks a lot for your feedback @tolbertam, and apologies for the trouble.

tolbertam:
Awesome, glad you were able to get it working, thanks for letting us know and no trouble at all!

PR #937, opened 2018-01-22
JAVA-1518: Expose metrics

olim7t:
Design goals:
* metrics can be enabled or disabled individually.
* disabled metrics are transparent for the driver code (no `if (metricEnabled)` tests) and have minimal runtime overhead.
* metrics can be node-level or session-level.
* the driver simply exposes a `MetricRegistry` and makes no assumption about how they will be exported; JMX is trivial (as shown below), so I don't think we need to build it in. In the same vein, individual metric objects are not exposed directly through the API (I doubt many people used that in 3.x, 99% of the time they will be exported to a monitoring tool with a reporter).

Basic manual test:
```java
try (CqlSession session = CqlSession.builder().build();
    JmxReporter reporter =
        JmxReporter.forRegistry(session.getMetricRegistry())
            .inDomain("com.datastax.oss.driver")
            .build()) {

  reporter.start();
  for (int i = 0; i < 1000; i++) {
    session.execute("select release_version from system.local");
  }
  TimeUnit.MINUTES.sleep(10);
}
```
Then connect to the process with jvisualvm or jconsole to observe the JMX MBeans.

The first commit is a PoC with a minimal set of metrics. TODO:
- [x] add remaining metrics (figure out what we want per node vs. per session, if we allow redundancy...)
- [x] switch `Timer` to an HDRHistogram-based implementation (either 3rd-party or custom).
- [x] tests

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -15,7 +15,9 @@
     limitations under the License.
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
   <parent>
@@ -74,6 +76,14 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.dropwizard.metrics</groupId>
+      <artifactId>metrics-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.hdrhistogram</groupId>
+      <artifactId>HdrHistogram</artifactId>
     </dependency>
     <dependency>
       <groupId>ch.qos.logback</groupId>
--- a/core/src/main/java/com/datastax/oss/driver/api/core/config/CoreDriverOption.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/config/CoreDriverOption.java
@@ -37,6 +37,7 @@ public enum CoreDriverOption implements DriverOption {
   CONNECTION_HEARTBEAT_INTERVAL("connection.heartbeat.interval", true),
   CONNECTION_HEARTBEAT_TIMEOUT("connection.heartbeat.timeout", true),
   CONNECTION_MAX_ORPHAN_REQUESTS("connection.max-orphan-requests", true),
+  CONNECTION_WARN_INIT_ERROR("connection.warn-on-init-error", true),
   CONNECTION_POOL_LOCAL_SIZE("connection.pool.local.size", true),
   CONNECTION_POOL_REMOTE_SIZE("connection.pool.remote.size", true),
 
--- a/core/src/main/java/com/datastax/oss/driver/api/core/config/CoreDriverOption.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/config/CoreDriverOption.java
@@ -109,6 +109,15 @@ public enum CoreDriverOption implements DriverOption {
   TIMESTAMP_GENERATOR_DRIFT_WARNING_INTERVAL(
       "request.timestamp-generator.drift-warning.interval", false),
 
+  METRICS_SESSION_ENABLED("metrics.session.enabled", false),
+  METRICS_NODE_ENABLED("metrics.node.enabled", false),
+  METRICS_SESSION_CQL_REQUESTS_HIGHEST("metrics.session.cql-requests.highest-latency", false),
+  METRICS_SESSION_CQL_REQUESTS_DIGITS("metrics.session.cql-requests.significant-digits", false),
+  METRICS_SESSION_CQL_REQUESTS_INTERVAL("metrics.session.cql-requests.refresh-interval", false),
+  METRICS_NODE_CQL_MESSAGES_HIGHEST("metrics.node.cql-messages.highest-latency", false),
+  METRICS_NODE_CQL_MESSAGES_DIGITS("metrics.node.cql-messages.significant-digits", false),
+  METRICS_NODE_CQL_MESSAGES_INTERVAL("metrics.node.cql-messages.refresh-interval", false),
+
   NETTY_IO_SIZE("netty.io-group.size", false),
   NETTY_IO_SHUTDOWN_QUIET_PERIOD("netty.io-group.shutdown.quiet-period", false),
   NETTY_IO_SHUTDOWN_TIMEOUT("netty.io-group.shutdown.timeout", false),
--- a/core/src/main/java/com/datastax/oss/driver/api/core/metrics/CoreNodeMetric.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/metrics/CoreNodeMetric.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core.metrics;
+
+import com.google.common.collect.ImmutableMap;
+import java.util.Map;
+
+/** See {@code reference.conf} for a description of each metric. */
+public enum CoreNodeMetric implements NodeMetric {
+  OPEN_CONNECTIONS("pool.open-connections"),
+  AVAILABLE_STREAMS("pool.available-streams"),
+  IN_FLIGHT("pool.in-flight"),
+  ORPHANED_STREAMS("pool.orphaned-streams"),
+  CQL_MESSAGES("cql-messages"),
+  UNSENT_REQUESTS("errors.request.unsent"),
+  ABORTED_REQUESTS("errors.request.aborted"),
+  WRITE_TIMEOUTS("errors.request.write-timeouts"),
+  READ_TIMEOUTS("errors.request.read-timeouts"),
+  UNAVAILABLES("errors.request.unavailables"),
+  OTHER_ERRORS("errors.request.others"),
+  RETRIES("retries.total"),
+  RETRIES_ON_ABORTED("retries.aborted"),
+  RETRIES_ON_READ_TIMEOUT("retries.read-timeout"),
+  RETRIES_ON_WRITE_TIMEOUT("retries.write-timeout"),
+  RETRIES_ON_UNAVAILABLE("retries.unavailable"),
+  RETRIES_ON_OTHER_ERROR("retries.other"),
+  IGNORES("ignores.total"),
+  IGNORES_ON_ABORTED("ignores.aborted"),
+  IGNORES_ON_READ_TIMEOUT("ignores.read-timeout"),
+  IGNORES_ON_WRITE_TIMEOUT("ignores.write-timeout"),
+  IGNORES_ON_UNAVAILABLE("ignores.unavailable"),
+  IGNORES_ON_OTHER_ERROR("ignores.other"),
+  SPECULATIVE_EXECUTIONS("speculative-executions"),
+  CONNECTION_INIT_ERRORS("errors.connection.init"),
+  AUTHENTICATION_ERRORS("errors.connection.auth"),
+  ;
+
+  private static final Map<String, CoreNodeMetric> BY_PATH = sortByPath();
+
+  private final String path;
+
+  CoreNodeMetric(String path) {
+    this.path = path;
+  }
+
+  @Override
+  public String getPath() {
+    return path;
+  }
+
+  public static CoreNodeMetric fromPath(String path) {
+    CoreNodeMetric metric = BY_PATH.get(path);
+    if (metric == null) {
+      throw new IllegalArgumentException("Unknown node metric path " + path);
+    }
+    return metric;
+  }
+
+  private static Map<String, CoreNodeMetric> sortByPath() {
+    ImmutableMap.Builder<String, CoreNodeMetric> result = ImmutableMap.builder();
+    for (CoreNodeMetric value : values()) {
+      result.put(value.getPath(), value);
+    }
+    return result.build();
+  }
+}
--- a/core/src/main/java/com/datastax/oss/driver/api/core/metrics/NodeMetric.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/metrics/NodeMetric.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core.metrics;
+
+import com.datastax.oss.driver.api.core.session.Session;
+import java.net.InetAddress;
+
+/**
+ * A node-level metric exposed through {@link Session#getMetricRegistry()}.
+ *
+ * <p>Note that the actual key in the registry is composed of the {@link Session#getName() name of
+ * the session}, followed by "nodes", followed by a textual representation of the address of the
+ * node, followed by the path of the metric. IPv4 addresses are represented as the decimal
+ * components followed by the port, separated with underscores; IPv6 addresses are represented as
+ * the result of {@link InetAddress#getHostAddress()}, followed by an underscore, followed by the
+ * port. For example:
+ *
+ * <pre>
+ * // Retrieve the `retries.total` metric for node 127.0.0.1:9042 in session `s0`:
+ * Meter retriesMeter = session.getMetricRegistry().meter("s0.nodes.127_0_0_1_9042.retries.total");
+ *
+ * // Retrieve the `retries.total` metric for node ::1:9042 in session `s0`:
+ * Meter retriesMeter = session.getMetricRegistry().meter("s0.nodes.0:0:0:0:0:0:0:1_9042.retries.total");
+ * </pre>
+ *
+ * <p>All metrics exposed out of the box by the driver are instances of {@link CoreNodeMetric} (this
+ * interface only exists to allow custom metrics in driver extensions).
+ *
+ * @see SessionMetric
+ */
+public interface NodeMetric {
+  String getPath();
+}
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
@@ -26,6 +26,8 @@ import com.datastax.oss.driver.api.core.cql.AsyncResultSet;
 import com.datastax.oss.driver.api.core.cql.ExecutionInfo;
 import com.datastax.oss.driver.api.core.cql.Statement;
 import com.datastax.oss.driver.api.core.metadata.Node;
+import com.datastax.oss.driver.api.core.metrics.CoreNodeMetric;
+import com.datastax.oss.driver.api.core.metrics.CoreSessionMetric;
 import com.datastax.oss.driver.api.core.retry.RetryDecision;
 import com.datastax.oss.driver.api.core.retry.RetryPolicy;
 import com.datastax.oss.driver.api.core.servererrors.BootstrappingException;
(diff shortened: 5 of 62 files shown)


core/pom.xml
```diff
@@ -75,6 +77,15 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.dropwizard.metrics</groupId>
+      <artifactId>metrics-core</artifactId>
+    </dependency>
+    <dependency>
```
adutra, line 84: A few remarks about JMX:

1) `metrics-jmx` does not seem used at all, not even in tests. Is this intentional? 
2) More broadly speaking, are you leaning towards the idea of not providing JMX support inside the driver and leave that for users?
3) You seem to be creating hierarchical metric names such as `s0.nodes.127_0_0_1_9042.retries`. If you provide out-of-the-box JMX support, the nice thing is that you could customize your MBeans by creating hierarchical beans that reflect your internal hierarchy of metrics.

core/src/main/java/com/datastax/oss/driver/api/core/metrics/CoreNodeMetric.java
```diff
@@ -0,0 +1,46 @@
...
+public enum CoreNodeMetric implements NodeMetric {
+
+  /**
+   * The number of connections open to this node for regular requests (exposed as a {@link Gauge
+   * Gauge&lt;Integer&gt;}.
+   *
+   * <p>This does not include the control connection (which uses at most one extra connection to a
+   * random node in the cluster).
+   */
+  pooled_connection_count,
```
adutra: Can't we stick to standards and use upper-case enum constant names?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultNode.java
```diff
@@ -50,8 +53,11 @@
 
   volatile NodeDistance distance;
 
-  public DefaultNode(InetSocketAddress connectAddress) {
+  public DefaultNode(InetSocketAddress connectAddress, InternalDriverContext context) {
     this.connectAddress = connectAddress;
+    // We leak a reference to a partially constructed object (this), but in practice this won't be a
+    // problem because the node updater only needs the connect address to initialize.
+    this.metricUpdater = context.metricUpdaterFactory().newNodeUpdater(this);
```
adutra: it seems a little worse than that, I see that the constructor of `DefaultNodeUpdater` also tries to do the following:
```
ChannelPool pool = context.poolManager().getPools().get(node);
```

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DefaultNodeMetricUpdater.java
```diff
@@ -0,0 +1,86 @@
...
+                ChannelPool pool = context.poolManager().getPools().get(node);
+                return (pool == null) ? 0 : pool.getAvailableIds();
+              });
+    }
+  }
+
+  @Override
+  public void markMeter(NodeMetric metric) {
+    if (enabledMetrics.contains(metric)) {
+      metricRegistry.meter(metricNamePrefix + metric.name()).mark();
```
adutra: Hmm what if the metric in question is not a `Meter`? Would you create a new method here? I guess what i'm getting at is:  maybe `NodeMetric` should somehow convey the metric type, and not only the metric name, so that we can update the right metric without risking a mistake here? It could maybe even update itself, stg like:
```
NodeMetric metric = ...
metric.update(context); // get the registry and the session name from here and update the relevant metric
```
... in which case, the logic from `NodeMetricUpdater` could be merged into `NodeMetric`.

core/src/test/resources/logback-test.xml
```diff
@@ -25,4 +25,5 @@
         <appender-ref ref="STDOUT"/>
     </root>
     <logger name="com.datastax.oss.driver" level= "${driverLevel:-WARN}"/>
+    <logger name="Tmp" level= "DEBUG"/>
```
adutra: What is this logger for?

pom.xml
```diff
@@ -43,6 +43,7 @@
     <format.validateOnly>true</format.validateOnly>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+    <metrics.version>4.0.0</metrics.version>
```
adutra: 4.0.2 is the latest version.

adutra (changes requested):

core/pom.xml
olim7t: 1. True, I added it to write my manual test class, but it's not used elsewhere and might not even be used in future tests. Removing it.
2. Yes. I don't see any need to have it out of the box because a) Adding a JMX reporter is trivial (two lines of code) and b) it's not recommended in production so users are likely to use something else.
3. I don't understand, as far as I can tell the JMX reporter does already create a hierarchy based on the dot separators in the metric names. When I look up the MBeans in jvisualvm, `retries` is a child of `127_0_0_1_9042`, etc.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultNode.java
olim7t: That's actually done in a lambda that isn't invoked until you read the corresponding gauge (and anyway the hashcode of `DefaultNode` also happens to only use the connect address).

pom.xml
olim7t: :+1: Fixed

core/src/test/resources/logback-test.xml
olim7t: Hehe, that's an accidental commit, fixing.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DefaultNodeMetricUpdater.java
olim7t: Yes, there will be a method for each metric type: `markMeter`,  `updateTimer`, `incrementCounter`, etc.
The caller is expected to know what type of metric each enum constant corresponds to. A mistake is very unlikely because it's all driver code, and the same person will add a metric and add the code to update it. And if a mistake does happen the error should be pretty obvious.

The problem with something like `NodeMetric#update` is that `NodeMetric` couldn't be an enum anymore (because it would need a reference to the `Meter`, which is created at runtime, and also because there could be multiple instances for separate sessions). Then the list of metric names wouldn't be known statically, we'd probably need a `NodeMetricRegistry` in order to validate the configuration, etc. The (very low) risk of a mistake does not justify all that complexity IMO.

core/src/main/java/com/datastax/oss/driver/api/core/metrics/CoreNodeMetric.java
olim7t: I deliberately broke the convention because metric-land seems to favor lower case (from what I saw by looking at a few monitoring tools), and I find it nice to have a direct correspondence between the metric name and constant name.

If we want to respect the convention at all cost I could add a `getMetricName` that returns the lower-case version, but then it's a bit of a pity to have to explain that sometimes it's upper-case and sometimes lower.

core/src/main/java/com/datastax/oss/driver/api/core/config/CoreDriverOption.java
```diff
@@ -108,6 +108,9 @@
   TIMESTAMP_GENERATOR_DRIFT_WARNING_INTERVAL(
       "request.timestamp-generator.drift-warning.interval", false),
 
+  METRICS_SESSION_ENABLED("metrics.session.enabled", false),
```
tolbertam, line 112: 💯 for separating session and node level metrics!

core/src/main/java/com/datastax/oss/driver/api/core/metrics/NodeMetric.java
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core.metrics;
+
+import com.datastax.oss.driver.api.core.session.Session;
+import java.net.InetAddress;
+
+/**
```
tolbertam, line 21: Excellent docs 👍

core/src/main/resources/reference.conf
```diff
@@ -550,6 +550,21 @@ datastax-java-driver {
     token-map.enabled = true
   }
 
+  metrics {
```
tolbertam, line 569: I think it's cool that you can enable/disable specific metrics.  If you have a metrics db (graphite, influx, etc.) that a lot of clients are feeding into, it's pretty easy to overwhelm it, so to be able to be selective about what you record is nice.

tolbertam (commented):
Looks great so far 👍

core/pom.xml
adutra: > as far as I can tell the JMX reporter does already create a hierarchy based on the dot separators in the metric names

Was this available since metrics 3? I didn't know. OK then.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DefaultNodeMetricUpdater.java
adutra: Sounds reasonable. 👍

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/HdrReservoir.java
```diff
@@ -0,0 +1,240 @@
...
+    throw new UnsupportedOperationException("HdrReservoir does not implement size()");
+  }
+
+  @Override
+  public Snapshot getSnapshot() {
+    long now = System.nanoTime();
+
+    cacheLock.readLock().lock();
+    try {
+      if (now < cachedHistogramTimestampNanos + refreshIntervalNanos) {
```
adutra: This can overflow so I think the recommended approach is to use `(now - cachedHistogramTimestampNanos) > refreshIntervalNanos`

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/HdrReservoir.java
```diff
@@ -0,0 +1,240 @@
...
+        return cachedSnapshot;
+      }
+    } finally {
+      cacheLock.readLock().unlock();
+    }
+
+    cacheLock.writeLock().lock();
+    try {
+      // Might have raced with another writer => re-check the timestamp
+      if (now >= cachedHistogramTimestampNanos + refreshIntervalNanos) {
```
adutra: Same here

core/src/main/resources/reference.conf
```diff
@@ -553,16 +553,48 @@ datastax-java-driver {
...
+        cql_requests,
+      ]
+
+      # Extra configuration (for the metrics that need it)
+      cql_requests {
+        # The largest latency that we expect to record (should be consistent with request.timeout).
+        #
+        # This is used to scale internal data structures. If a higher recording is encountered at
+        # runtime, it is discarded and a warning is logged.
+        highest_latency = 2 seconds
```
adutra: Shouldn't we default then to `${request.timeout}`?

adutra (changes requested):

core/src/main/resources/reference.conf
olim7t: 👍 
I wonder if there is a risk of observing a higher value because of some processing overhead though. With significant_digits = 3 it would have to be a few milliseconds more, so I think we're safe. At worst it will come up as a warning and we can advise users to configure it a bit higher.

core/pom.xml
olim7t: I double-checked and actually you're right, I stand corrected.

But there is a way to do it from outside the driver: build the `JmxReporter` with an `ObjectNameFactory` that turns every dot-separated component into a subdirectory:

```java
ObjectNameFactory objectNameFactory = (type, domain, name) -> {
  StringBuilder objectName = new StringBuilder(domain).append(':');
  List<String> nameParts = Splitter.on('.').splitToList(name);
  int i = 0;
  for (String namePart : nameParts) {
    boolean isLast = (i == nameParts.size() - 1);
    String key =
        isLast ? "name" : Strings.padStart(Integer.toString(i), 2, '0');
    objectName.append(key).append('=').append(namePart);
    if (!isLast) {
      objectName.append(',');
    }
    i += 1;
  }
  try {
    return new ObjectName(objectName.toString());
  } catch (MalformedObjectNameException e) {
    throw new RuntimeException(e);
  }
};

JmxReporter reporter =
    JmxReporter.forRegistry(session.getMetricRegistry())
        .inDomain("com.datastax.oss.driver")
        .createsObjectNamesWith(objectNameFactory)
        .build()
```

adutra (commented):
Looking excellent so far 👍

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/HdrReservoir.java
```diff
@@ -0,0 +1,241 @@
...
+   * <p>This reservoir implementation is intended for use with a {@link
+   * com.codahale.metrics.Histogram}, which doesn't use this method.
+   */
+  @Override
+  public int size() {
+    throw new UnsupportedOperationException("HdrReservoir does not implement size()");
+  }
+
+  @Override
+  public Snapshot getSnapshot() {
```
tolbertam: Since live and cached are only switched out on call to `getSnapshot` is there a risk that there can be more data in live and the cached is extra stale if `getSnapshot` is called on intervals > `refreshInterval`?  Since the refreshInterval is configurable I think it's fine, since we can just tell users to set it to the same duration they are using in their Reporters.

olim7t:
I added the remaining metrics compared to what was available in 3.0.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
```diff
@@ -304,6 +270,12 @@ private void setFinalResult(
           Conversions.toResultSet(resultMessage, executionInfo, session, context);
       if (result.complete(resultSet)) {
         cancelScheduledTasks();
+        session
+            .getMetricUpdater()
+            .updateTimer(
+                CoreSessionMetric.cql_requests,
+                System.nanoTime() - startTimeNanos,
```
tolbertam: Is the reason `nanoTime` is used instead of `Timer.Context` because we can't be sure that a `Timer` is actually in use as metrics can be disabled?

tolbertam (commented):
Looks really good so far.  Going to play with this a little bit in an environment and will report back

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
olim7t: Yes, it is to avoid the extra object allocation, especially if the metric is disabled.

core/pom.xml
adutra: Yes that's basically how we configured the JMX reporter in DSBulk. 👍

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DefaultNodeMetricUpdater.java
```diff
@@ -0,0 +1,126 @@
...
+import org.slf4j.LoggerFactory;
+
+public class DefaultNodeMetricUpdater extends MetricUpdaterBase<NodeMetric>
+    implements NodeMetricUpdater {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultNodeMetricUpdater.class);
+
+  private final String metricNamePrefix;
+
+  public DefaultNodeMetricUpdater(
```
tolbertam: I noticed that a lot of metrics were not showing up, it looks like its because they aren't being registered here.  I.E. at the node level I just see `pooled_connection_count`, `available_stream_count` and `cql_messages`, but there's a lot more in `CoreNodeMetric`.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DefaultSessionMetricUpdater.java
```diff
@@ -0,0 +1,81 @@
...
+
+public class DefaultSessionMetricUpdater extends MetricUpdaterBase<SessionMetric>
+    implements SessionMetricUpdater {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultSessionMetricUpdater.class);
+
+  private final String metricNamePrefix;
+
+  public DefaultSessionMetricUpdater(
+      Set<SessionMetric> enabledMetrics, InternalDriverContext context) {
```
tolbertam, line 32: Missing registration of `cql_client_timeouts`

core/src/main/resources/reference.conf
```diff
@@ -550,6 +566,183 @@ datastax-java-driver {
...
+        # The number of errors encountered while trying to establish a connection to this node
+        # (exposed as a Counter).
+        #
+        # Connection errors are not a fatal issue for the driver, failed connections will be retried
+        # periodically according to the reconnection policy. You can choose whether or not to log
+        # those errors at WARN level with the connection.warn-on-init-error option.
+        #
+        # Authentication errors are not included in this counter, they are tracked separately in
+        # authentication_errors.
+        // connection_errors,
```
tolbertam: Should `connection_errors` and `authentication_errors` be under `errors.`?

olim7t: They relate to connections while the others are more specifically for requests, so maybe I should have two sub-categories.

tolbertam: good point, maybe differentiating into `connection_errors` and `request_errors` or something, and then renaming the current `connection_errors` to something like more specific like `startup_errors` or something like that.

core/src/main/resources/reference.conf
```diff
@@ -550,6 +566,183 @@ datastax-java-driver {
...
+    }
+    # The node-level metrics (all disabled by default).
+    node {
+      enabled = [
+        # The number of connections open to this node for regular requests (exposed as a
+        # Gauge<Integer>).
+        #
+        # This does not include the control connection (which uses at most one extra connection to a
+        # random node in the cluster).
+        // pooled_connection_count,
```
tolbertam: Just brainstorming (maybe i'm over thinking it).  A side effect of the granularity of being able to turn individual metrics on/off is that there are a few downsides:

1. If you want to include all metrics, you have to extract them all from this file and put them into your application.conf.  It's a bit tedious to have to copy the individual ones, but It's tolerable.

2. If a new metric is added, the user has to know about it and add it to their application.conf.

One possible solution would be to make the configuration more coarse and allow enabling metrics by category, i.e.:

```
enabled = [
  pooled_connection_count,
  available_stream_count,
  errors,
  retries,
  ignores
]
```

I wouldn't expect someone to selectively enable/disable specific retries for example, either they'd enable all of them or disable all of them.

you could also categorize `pooled_connection_count` , `available_stream_count` under a `pool` group or something.

core/src/main/java/com/datastax/oss/driver/api/core/metrics/CoreNodeMetric.java
```diff
@@ -23,12 +23,12 @@
   POOLED_CONNECTION_COUNT("pooled_connection_count"),
   AVAILABLE_STREAM_COUNT("available_stream_count"),
   CQL_MESSAGES("cql_messages"),
-  WRITE_ERRORS("errors.write_errors"),
-  ABORTED_REQUESTS("errors.aborted_requests"),
-  WRITE_TIMEOUTS("errors.write_timeouts"),
-  READ_TIMEOUTS("errors.read_timeouts"),
-  UNAVAILABLES("errors.unavailables"),
-  OTHER_ERRORS("errors.others"),
+  UNSENT_REQUESTS("errors.request.unsent"),
```
olim7t: Renamed this because "write errors" was too similar to "write timeouts".

tolbertam: 👍 i also like what you did with the hierarchy (`errors.request` / `errors.connection`)

core/src/main/resources/reference.conf
olim7t: > If a new metric is added, the user has to know about it and add it to their application.conf

Yes, but on the flip side if metrics are enabled by default, then you risk introducing new (potentially expensive) metrics in existing environments, without the user's notice. I've been pondering on this "enabled vs disabled by default" question since the beginning of this PR, I think I prefer being conservative. We'll advertise new metrics in release notes and the upgrade guide.

> If you want to include all metrics, you have to extract them all from this file and put them into your application.conf. It's a bit tedious

At least that encourages you to read all the descriptions 😉

tolbertam: > At least that encourages you to read all the descriptions 😉

Haha fair enough!  There isn't much of a downside in educating what the metrics **actually mean** (especially when it comes to the timer-based ones) and it will be a one time task.   I think i'm 👍  on the current configuration.

tolbertam:
small nit:  The formatter doesn't like core/src/test/java/com/datastax/oss/driver/internal/core/pool/ChannelPoolInitTest.java

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DefaultNodeMetricUpdater.java
```diff
@@ -44,11 +44,7 @@ public DefaultNodeMetricUpdater(
     if (enabledMetrics.contains(CoreNodeMetric.OPEN_CONNECTIONS)) {
       metricRegistry.register(
           buildFullName(CoreNodeMetric.OPEN_CONNECTIONS),
-          (Gauge<Integer>)
-              () -> {
-                ChannelPool pool = context.poolManager().getPools().get(node);
-                return (pool == null) ? 0 : pool.size();
-              });
+          (Gauge<Integer>) node::getOpenConnections);
```
olim7t: duh

core/src/main/resources/reference.conf
```diff
@@ -550,6 +566,198 @@ datastax-java-driver {
     token-map.enabled = true
   }
 
+  metrics {
+    # The session-level metrics (all disabled by default).
+    session {
+      enabled = [
+        # The number of nodes to which the driver has at least one active connection (exposed as a
+        # Gauge<Integer>).
+        //connected-nodes,
```
tolbertam: Very tiny nit, no space between `//` and `connected-nodes` where other metrics have a space.

tolbertam (approved):
This looks and is working great!  Much improved over the 3.x implementation.  👍

PR #938, opened 2018-01-23
Run ErrorProne as part of the build

olim7t:
http://errorprone.info

This is a proposal. As with any automatic checker there are pros and cons, but I've found this one to be very good at detecting some serious bugs (see "errors" commit -- except the switch/case comments).

The most problematic warning is [FutureReturnValueIgnored](http://errorprone.info/bugpattern/FutureReturnValueIgnored). Swallowing exceptions in futures is a real issue, but it's already addressed correctly in the driver, most of the occurrences are either:
* false positives, such as Netty's `addListener`.
* cases where the action is "fire and forget", we have no way to propagate the error upstream. I've elected to either ignore the error (e.g. when closing a channel), or log unexpected errors so that users would report a bug (see usages of `UncaughtException#log`).

What also sucks is that ErrorProne uses `@SuppressWarnings` as a suppression mechanism, and annotations do not apply to individual statements. So most of the time we would have to annotate the enclosing element, which casts our net too wide. So maybe we could explicitly exclude `FutureReturnValueIgnored` and rely on good development/review practices.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/CassandraVersion.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/CassandraVersion.java
@@ -271,7 +271,7 @@ public class CassandraVersion implements Comparable<CassandraVersion> {
 
   @Override
   public int hashCode() {
-    return Objects.hash(major, minor, patch, dsePatch, preReleases, build);
+    return Objects.hash(major, minor, patch, dsePatch, Arrays.hashCode(preReleases), build);
   }
 
   @Override
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/DefaultCodecRegistry.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/DefaultCodecRegistry.java
@@ -67,21 +67,25 @@ public class DefaultCodecRegistry extends CachingCodecRegistry {
     if (cacheWeigher != null) {
       cacheBuilder.weigher(cacheWeigher::apply).maximumWeight(maximumCacheWeight);
     }
+    CacheLoader<CacheKey, TypeCodec<?>> cacheLoader =
+        new CacheLoader<CacheKey, TypeCodec<?>>() {
+          @Override
+          public TypeCodec<?> load(CacheKey key) throws Exception {
+            return createCodec(key.cqlType, key.javaType);
+          }
+        };
     if (cacheRemovalListener != null) {
-      //noinspection ResultOfMethodCallIgnored
-      cacheBuilder.removalListener(
-          (RemovalListener<CacheKey, TypeCodec<?>>)
-              notification ->
-                  cacheRemovalListener.accept(notification.getKey(), notification.getValue()));
+      this.cache =
+          cacheBuilder
+              .removalListener(
+                  (RemovalListener<CacheKey, TypeCodec<?>>)
+                      notification ->
+                          cacheRemovalListener.accept(
+                              notification.getKey(), notification.getValue()))
+              .build(cacheLoader);
+    } else {
+      this.cache = cacheBuilder.build(cacheLoader);
     }
-    this.cache =
-        cacheBuilder.build(
-            new CacheLoader<CacheKey, TypeCodec<?>>() {
-              @Override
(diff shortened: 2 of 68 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/CassandraVersion.java
```diff
@@ -271,7 +271,7 @@ public boolean equals(Object other) {
 
   @Override
   public int hashCode() {
-    return Objects.hash(major, minor, patch, dsePatch, preReleases, build);
+    return Objects.hash(major, minor, patch, dsePatch, Arrays.hashCode(preReleases), build);
```
olim7t, line 274: JAVA-1639, I had forgotten to port it.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/DefaultCodecRegistry.java
```diff
@@ -67,21 +67,25 @@ public DefaultCodecRegistry(
...
+          public TypeCodec<?> load(CacheKey key) throws Exception {
+            return createCodec(key.cqlType, key.javaType);
+          }
+        };
     if (cacheRemovalListener != null) {
-      //noinspection ResultOfMethodCallIgnored
-      cacheBuilder.removalListener(
-          (RemovalListener<CacheKey, TypeCodec<?>>)
-              notification ->
-                  cacheRemovalListener.accept(notification.getKey(), notification.getValue()));
```
olim7t, line 75: Another serious bug, the removal listener would have been ignored.

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/CycleDetector.java
```diff
@@ -74,7 +74,7 @@ void onReleaseLock(LazyReference<?> reference) {
       synchronized (this) {
         Thread me = Thread.currentThread();
         LOG.debug("{} is done initializing {}", me, reference.getName());
-        graph.removeEdge(reference.getName(), me);
+        graph.removeEdge(reference.getName(), me.getName());
```
olim7t, line 77: The removal would fail before.

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/TupleCodecTest.java
```diff
@@ -150,7 +150,7 @@ public void should_parse_tuple() {
     TupleValue tuple = parse("(1,NULL,'a')");
 
     assertThat(tuple.getInt(0)).isEqualTo(1);
-    assertThat(tuple.isNull(1));
+    assertThat(tuple.isNull(1)).isTrue();
```
olim7t, line 153: assertion did nothing

tolbertam:
This seems like a good idea and now seems like a good time to do it.  I think my only concern is that there doesn't seem to be a lot tooling around error-prone (compared to findbugs / sonar / coverity / etc.).  Do you see us getting to the point where we could consider failing the maven build if there are any warnings present that aren't being suppressed?   I think that would go far in making this useful, otherwise I can see the list of warnings continue to grow.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/NodeStateIT.java
```diff
@@ -586,22 +586,12 @@ private InetSocketAddress withUnusedPort(InetSocketAddress address) {
...
-    ServerSocket ss = null;
-    try {
-      // let the system pick an ephemeral port
-      ss = new ServerSocket(0);
+    // let the system pick an ephemeral port
+    try (ServerSocket ss = new ServerSocket(0)) {
       ss.setReuseAddress(true);
       return ss.getLocalPort();
-    } catch (IOException e) {
+    }catch (IOException e) {
```
tolbertam: Small Nit, spacing throws off our formatting rules.

olim7t:
We've decided to try out ErrorProne as a prerequisite for a successful build. If this becomes too cumbersome to maintain, we can revert it. I've moved the `SuppressWarnings` annotations to a dedicated commit to make that easier.

PR #939, opened 2018-01-23
JAVA-1727 Add equals to DefaultUdtValue

GregBestland:


core/src/main/java/com/datastax/oss/driver/internal/core/data/DefaultUdtValue.java
```diff
@@ -91,6 +92,39 @@ public ProtocolVersion protocolVersion() {
     return type.getAttachmentPoint().protocolVersion();
   }
 
+  @Override
```
tolbertam, line 99: It might be better to use IntelliJ generated `equals` and `hashCode` methods which are more straight forward and depends on `ByteBuffer.equals()`:

```java
  @Override
    if (this == o) return true;
    if (o == null || getClass() != o.getClass()) return false;
    DefaultUdtValue that = (DefaultUdtValue) o;
    return Objects.equals(type, that.type) && Arrays.equals(values, that.values);
  }

  @Override
  public int hashCode() {

    int result = Objects.hash(type);
    result = 31 * result + Arrays.hashCode(values);
    return result;
  }
```

core/src/main/java/com/datastax/oss/driver/internal/core/data/DefaultUdtValue.java
```diff
@@ -91,6 +92,39 @@ public ProtocolVersion protocolVersion() {
...
+            return false;
+          }
+        } else if (!getBytesUnsafe(i).equals(that.getBytesUnsafe(i))) return false;
+      }
+    }
+    return true;
+  }
+
+  @Override
+  public int hashCode() {
```
tolbertam, line 140: I think this is the same as not implementing `hashCode` isn't it?

core/src/main/java/com/datastax/oss/driver/internal/core/data/DefaultUdtValue.java
olim7t: 👍  for doing an instance equality check first (`this == o`).

About comparing the byte buffers directly, there is a comment in the 3.x implementation that explains why we might want to avoid that (it might be a good idea to duplicate it here):

> Deserializing each value is slightly inefficient, but comparing
> the bytes could in theory be wrong (for varint for instance, 2 values
> can have different binary representation but be the same value due to
> leading zeros). So we don't take any risk.

Finally, rather than comparing `getClass()`, we should do an `instanceof` check, but against the interface, not the implementation. That way if someone writes their own `UdtValue` implementation, ours compares to it nicely. See `DefaultUserDefinedType.equals` for example.

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/UdtCodecTest.java
```diff
@@ -128,6 +128,15 @@ public void should_decode_udt() {
     Mockito.verify(textCodec).decode(Bytes.fromHexString("0x61"), ProtocolVersion.DEFAULT);
   }
 
+  @Test
+  public void should_evaluate_eqaulity() {
```
olim7t: nit: eqaulity => equality

core/src/main/java/com/datastax/oss/driver/internal/core/data/DefaultUdtValue.java
```diff
@@ -91,6 +92,39 @@ public ProtocolVersion protocolVersion() {
     return type.getAttachmentPoint().protocolVersion();
   }
 
+  @Override
+  public boolean equals(Object o) {
+    if (!(o instanceof DefaultUdtValue)) return false;
```
olim7t: nit: our style guide requires braces for if statements, even when there is only one statement. I used to omit braces too, but in hindsight I find it slightly less readable.

olim7t (changes requested):

PR #940, opened 2018-01-29
JAVA-1729 Adding equals to tuple

GregBestland:


core/src/main/java/com/datastax/oss/driver/internal/core/data/DefaultTupleValue.java
```diff
@@ -93,6 +95,44 @@ private void readObject(ObjectInputStream stream) throws InvalidObjectException
     throw new InvalidObjectException("Proxy required");
   }
 
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
```
olim7t: Nit: can you add braces, even if there is only one statement? That's part of the Google style guide that we adhere to.

core/src/main/java/com/datastax/oss/driver/internal/core/data/DefaultTupleValue.java
```diff
@@ -93,6 +95,44 @@ private void readObject(ObjectInputStream stream) throws InvalidObjectException
...
+
+    if (o == null || !(o instanceof TupleValue)) return false;
```
olim7t: The `instanceof` check fails if o is null, so you don't need the explicit null check.

core/src/main/java/com/datastax/oss/driver/internal/core/data/DefaultTupleValue.java
```diff
@@ -93,6 +95,44 @@ private void readObject(ObjectInputStream stream) throws InvalidObjectException
...
+    DefaultTupleValue that = (DefaultTupleValue) o;
+
+    if (this.protocolVersion() != that.protocolVersion()) return false;
```
olim7t: In practice all instances of `ProtocolVersion` are enum values so this works, but since in theory there could be other implementations, this should use `equals()`.

Unrelated: I've noticed that `hashCode` doesn't use the protocol version, but that's not necessarily an issue: it doesn't break the rule that equal objects must have the same hash code, and in practice different protocol versions will be rare, so I'm ok with it.

core/src/main/java/com/datastax/oss/driver/internal/core/data/DefaultTupleValue.java
```diff
@@ -93,6 +95,44 @@ private void readObject(ObjectInputStream stream) throws InvalidObjectException
...
+      }
+      Object thisValue =
+          this.codecRegistry()
+              .codecFor(innerThatType)
+              .decode(this.values[i], this.protocolVersion());
+      Object thatValue =
+          that.codecRegistry()
+              .codecFor(innerThatType)
+              .decode(that.values[i], that.protocolVersion());
+      if (!((thisValue == thatValue) || (thisValue != null && thisValue.equals(thatValue)))) {
```
olim7t: Java 8 has a method that handles the null check, so this can be simplified as `if (!Objects.equals(thisValue, thatValue))`.

core/src/main/java/com/datastax/oss/driver/internal/core/data/DefaultTupleValue.java
```diff
@@ -93,6 +95,44 @@ private void readObject(ObjectInputStream stream) throws InvalidObjectException
...
+    for (int i = 0; i < values.length; i++) {
+      DataType innerThisType = type.getComponentTypes().get(i);
+      DataType innerThatType = type.getComponentTypes().get(i);
+      if (!innerThisType.equals(innerThatType)) {
+        return false;
+      }
+      Object thisValue =
+          this.codecRegistry()
+              .codecFor(innerThatType)
+              .decode(this.values[i], this.protocolVersion());
```
olim7t: I wonder if we should handle potential errors here: this can happen if the tuple value contains custom types but became detached from the session, so the codec lookup will fail. Maybe we should log a warning and fall back to byte array comparisons in that specific case. Or is it better to simply let the method throw?

core/src/main/java/com/datastax/oss/driver/internal/core/data/DefaultTupleValue.java
```diff
@@ -93,6 +95,44 @@ private void readObject(ObjectInputStream stream) throws InvalidObjectException
...
+    }
+    return true;
+  }
+
+  @Override
+  public int hashCode() {
+
+    int result = Objects.hash(type);
+    result = 31 * result + Arrays.hashCode(values);
+    return result;
```
olim7t: If we decode in `equals()`, we must also decode here, otherwise it violates the rule that equal objects must have the same hash codes. For example two equal `varint` values that have different binary representations will be considered equal, but produce different hash codes.

olim7t (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/data/DefaultTupleValue.java
```diff
@@ -93,6 +98,78 @@ private void readObject(ObjectInputStream stream) throws InvalidObjectException
...
+        return false;
+      }
+    }
+    return true;
+  }
+
+  @Override
+  public int hashCode() {
+
+    int result = Objects.hash(type);
```
olim7t: Since only one object is getting hashed here, this can be replaced by `type.hashCode()`.
Same at line 161.

core/src/main/java/com/datastax/oss/driver/internal/core/data/DefaultTupleValue.java
```diff
@@ -93,6 +98,78 @@ private void readObject(ObjectInputStream stream) throws InvalidObjectException
...
+
+    for (int i = 0; i < values.length; i++) {
+      DataType innerThisType = type.getComponentTypes().get(i);
+      TypeCodec thisCodec = this.codecRegistry().codecFor(innerThisType);
+      if (thisCodec == null) {
+        LOG.warn(
+            "No codec found for tuple type "
+                + innerThisType.toString()
+                + " falling bytes for hashcode");
```
olim7t: Nit: I think this should say "falling *back to* bytes".

olim7t (approved):
LGTM 👍 
Just 2 trivial comments.

core/src/main/java/com/datastax/oss/driver/internal/core/data/DefaultTupleValue.java
```diff
@@ -93,6 +98,78 @@ private void readObject(ObjectInputStream stream) throws InvalidObjectException
...
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+
+    if (!(o instanceof TupleValue)) {
+      return false;
+    }
+    DefaultTupleValue that = (DefaultTupleValue) o;
```
olim7t: Oops, I had missed this: it should cast to `TupleValue` here, and use the interface methods to access the fields. Otherwise it might throw if the other object is a custom implementation.

olim7t:
So this is a bit embarrassing: I just realised that there is no way that the field codecs can be missing, since we're dealing with CQL types (which any decent `CodecRegistry` implementation must support out of the box -- I've added a javadoc to that effect). Sorry for misleading you, I've removed the unnecessary checks.

 I've also moved the tests to `DefaultTupleValueTest`.

PR #941, opened 2018-01-31
JAVA-1738: Convert enums to allow extensibility

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigProfile.java
```diff
@@ -61,8 +60,4 @@
   Duration getDuration(DriverOption option);
 
   DriverConfigProfile withDuration(DriverOption option, Duration value);
-
-  ConsistencyLevel getConsistencyLevel(DriverOption option);
```
tolbertam, line 65: Did you decide these that these didn't merit having their own methods for resolving CL by option?  It looks like the changes don't require removing this, but having this here might have just been considered extraneous.

core/src/main/java/com/datastax/oss/driver/api/core/CoreConsistencyLevel.java
```diff
@@ -0,0 +1,66 @@
...
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core;
+
+import com.datastax.oss.protocol.internal.ProtocolConstants;
+import com.google.common.collect.ImmutableMap;
+import java.util.Map;
+
+/** A default consistency level supported by the driver out of the box. */
+public enum CoreConsistencyLevel implements ConsistencyLevel {
```
tolbertam, line 23: Some possible downsides of this approach:

1. You can't really iterate over all possible ConsistencyLevels without checking the concrete enum types.  What about adding a `values()` method to `ConsistencyLevelRegistry` (and the others) so someone could ask what all the possibly supported ones are?

2. Other enum features like `switch` statements only work with the concrete types, so you would have to cast to evaluate in that way, but that doesn't seem like too huge of a deal.

tolbertam (approved):
Looks good 👍

adutra (approved):

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigProfile.java
olim7t: Parsing CLs now requires `ConsistencyLevelRegistry`, which is in the context. The config can't have a reference to the context because it's created first.

But yeah apart from the technical difficulties it also feels extraneous to have a special case for CLs.

core/src/main/java/com/datastax/oss/driver/api/core/CoreConsistencyLevel.java
olim7t: For 1 yeah why not. Registries are internal components and we don't have the need at the moment, but it seems reasonable to add `values()`.

PR #942, opened 2018-02-02
Check that tests using CcmRule are properly annotated

olim7t:
Do this check automatically because it's easy to forget and hard to debug.
```
org.junit.AssumptionViolatedException: Tests using CcmRule must be annotated with `@Category(ParallelizableTests.class)`. Description: com.datastax.oss.driver.api.core.cql.SimpleStatementIT
```
I had to move back the category classes into the test-infra module.


tolbertam (approved):
Looks great!  👍

PR #944, opened 2018-02-07
JAVA-1739 Add host_id and schema_version to node metadata

GregBestland:


olim7t:
👍 
I amended the commit to use `schemaVersion` everywhere (was `schemaId` in a few places) and moved the integration test to a dedicated class (it does not relate directly to schema metadata).

PR #948, opened 2018-02-13
Change signature of AsyncResultSet.fetchNextPage()

adutra:
Motivation:
Its return type should reflect that AsyncResultSet is an interface,
so what is actually returned is a future of some concrete implementation,
hence the correct return type should be CompletionStage<? extends AsyncResultSet>.


olim7t (approved):
👍 
This is transparent when iterating the results (as evidenced by the absence of changes in `MultiPageResultSet`). The only tricky part is when trying to complete the future, but that should never be useful outside of tests.

The changes in `Conversions` seemed unrelated and they're covered by your other PR, so I removed them. I also extracted a utility method in the test to make things more readable.

PR #950, opened 2018-02-14
JAVA-1751: Include defaultTimestamp length in encodedSize for >= V3

tolbertam:
QueryProtocolOptions.encodedSize previously incorrectly only accounted
for defaultTimestamp length if protocol version was equal to V3.  It
should be included for all protocol versions >= V3.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/CBUtil.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/CBUtil.java
@@ -113,22 +113,17 @@ abstract class CBUtil { // TODO rename
         }
     }
 
-    public static void writeBytes(byte[] bytes, ByteBuf cb) {
+    public static void writeShortBytes(byte[] bytes, ByteBuf cb) {
         cb.writeShort(bytes.length);
         cb.writeBytes(bytes);
     }
 
-    public static void writeBytes(ByteBuffer bytes, ByteBuf cb) {
-        cb.writeShort(bytes.remaining());
-        cb.writeBytes(bytes.duplicate());
-    }
-
-    public static int sizeOfBytes(byte[] bytes) {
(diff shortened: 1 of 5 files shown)


newkek (approved):

olim7t (approved):

olim7t:
If there are other bugs in `encodedSize`, we want to fix them in order to allocate the optimal size and avoid resizing.
However blowing up if the size is wrong is probably too aggressive, at least for driver 3, since the error is recoverable. So I suggest adding a warning that will alert us if this happens elsewhere (it's the only warn log in this class, so easy to disable). I'm going to trigger our integration suite with it to check that it doesn't pop up.

tolbertam:
👍 on adding the warning, thanks!

tolbertam:
The warning identified at least one other issue, [JAVA-1770](https://datastax-oss.atlassian.net/browse/JAVA-1770) which I pushed a fix for in a89b30a.  Will run through suite again to see if any other issues are identified.

driver-core/src/main/java/com/datastax/driver/core/CBUtil.java
```diff
@@ -113,22 +113,17 @@ public static int sizeOfLongString(String str) {
         }
     }
 
-    public static void writeBytes(byte[] bytes, ByteBuf cb) {
+    public static void writeShortBytes(byte[] bytes, ByteBuf cb) {
         cb.writeShort(bytes.length);
         cb.writeBytes(bytes);
     }
 
-    public static void writeBytes(ByteBuffer bytes, ByteBuf cb) {
```
tolbertam, line 121: Removed unused methods, while they might be useful in future, would be easy to reimplement them.

tolbertam:
I don't see any more incidences of the added warning, so I think we got them all 🎉

PR #951, opened 2018-02-16
JAVA-1537: Add remaining socket options

adutra:


olim7t (approved):
LGTM 👍 
I amended the comments to be a bit more specific on default values.

PR #954, opened 2018-02-20
Fix eclipse problems

alexott:
There were 2 problems with importing Java driver into Eclipse:

1.  no `groupId` specified, so m2e plugin couldn't import the project correctly
2. no m2e connectors for 2 plugins  - they were just marked as ignored in m2e configuration

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -843,6 +843,7 @@ limitations under the License.
                             <pluginExecutions>
                                 <pluginExecution>
                                     <pluginExecutionFilter>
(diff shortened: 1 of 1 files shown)


tolbertam (approved):
Looks good to me.  Since changes are isolated to `org.eclipse.m2e:lifecycle-mapping` it is a low risk change 👍

olim7t (approved):
👍 since the m2eclipse configuration was already present, but on a side note I'm not a big fan of polluting the POM with IDE-specific configuration.

PR #955, opened 2018-02-20
JAVA-1380: Add FAQ entry for errors arising from incompatibilities

tolbertam:
For [JAVA-1380](https://datastax-oss.atlassian.net/browse/JAVA-1380).

Often users run into issues that are ultimately caused by using incompatible tools or libraries.   This adds a FAQ entry that gives further guidance (but doesn't go into too much detail as this is not a driver specific problem).

### DIFF
--- a/faq/README.md
+++ b/faq/README.md
@@ -238,7 +238,7 @@ requests when connections are over-utilized.
 
 ### What is Netty's native epoll transport and how do I enable or disable it?
 
-Netty provides [native transport libraries](http://netty.io/wiki/native-transports.html) which generally generate less 
+Netty provides [native transport libraries](http://netty.io/wiki/native-transports.html) which generally generate less
 garbage and improve performance when compared to the default NIO-based transport.
 By default if the driver detects the  `netty-transport-native-epoll` library in its classpath it will attempt to use
 [`EpollEventLoopGroup`](https://netty.io/4.0/api/io/netty/channel/epoll/EpollEventLoopGroup.html) for its underlying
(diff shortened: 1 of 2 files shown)


faq/README.md
```diff
@@ -258,10 +258,32 @@ If properly used, the following log message will be logged at INFO on startup:
...
+Some monitoring and agent-based tools such as [DynaTrace] offer solutions that instrument the driver to observe
+and record useful metrics such as request rates, latencies and more.  It is possible that the tool
+you are using is not compatible with the version of the java driver you are using.  In this case, check to
+see if a newer version of that tool is available that works with this version of the driver.  If no such
+version is available, you may want to reach out to the maintainer of that tool to request that they provide
+an update with compatibility to the driver version you are using.
+
+Another possibility could simply be that another library you are using depends on a different version of the
+driver.  In this case, observe the stacktrace of the exception to see which library is attempting to use
+the driver.  To identify compatible versions, check that libraries' dependency on the driver to understand
```
adutra: nit: "check that library's dependency"

faq/README.md
```diff
@@ -258,10 +258,32 @@ If properly used, the following log message will be logged at INFO on startup:
...
+what version is compatible.
+
+Finally, it could be that an older or newer version of a library that the driver depends on, such as Netty
+or Guava, may be present in your application's classpath.  If you are using maven or another dependency
+management tool, the tool should offer a command, such as `mvn dependency:tree` to identify the dependecies
+in your project to help you understand the dependent versions across the various libraries you use in your
+project.  You may also want to evaluate your classpath to see if there are multiple jars present for a library,
+but with different versions, which could cause compatibility issues.
```
adutra: I think that here you could also mention the possibility of using the shaded jar, if the library causing problems is Netty.

adutra (changes requested):
Very nice! 

Apart from a few nits, I wonder if we couldn't inverse the order of the 3 topics in this section:

1. Different versions of the same library in the classpath (netty, guava, etc.)
2. Another library depends on a different version of the driver
3. An instrumentation library depends on a different version of the driver (a special case of 2?)

I think that this order would better reflect the frequency with which these issues appear.

tolbertam:
Thanks for reviewing!  I agree with all of your comments, I'll push some changes shortly.

tolbertam:
Updated with feedback.  I also added a sentence about including packaging data in stacktraces using logback, which is very helpful for debugging incompatibilities.

faq/README.md
```diff
@@ -258,10 +258,44 @@ If properly used, the following log message will be logged at INFO on startup:
...
+It could be that an older or newer version of a library that the driver depends on, such as Netty
+or Guava, may be present in your application's classpath.  If you are using maven or another dependency
+management tool, the tool should offer a command, such as `mvn dependency:tree` to identify the dependecies
+in your project to help you understand the dependent versions across the various libraries you use in your
+project.  You may also want to evaluate your classpath to see if there are multiple jars present for a library,
+but with different versions, which could cause compatibility issues.  In addition, consider evaluating
+using the logback logging framework, which provides the capability to include [packaging data] for classes
+in stack traces.
+
+For Netty in particular, the driver offers an alternative artifact that shades its netty dependency,
```
adutra: Nit: "netty" appears twice without capital N.

adutra (approved):
👍 Fix the small nit below and :shipit:

faq/README.md
tolbertam: Good catch, fixed!

faq/README.md
```diff
@@ -258,10 +258,44 @@ If properly used, the following log message will be logged at INFO on startup:
...
 > Found Netty's native epoll transport in the classpath, but NIO was forced through the FORCE_NIO system property.
 
+### Why am I encountering `NoSuchMethodFoundException`, `NoClassDefFoundError`, or `VerifyError`s and how do I avoid them?
+
+Incompatibilities between the java driver and other libraries may cause these exceptions to surface in your
+application at runtime.  
+
+It could be that an older or newer version of a library that the driver depends on, such as Netty
+or Guava, may be present in your application's classpath.  If you are using maven or another dependency
+management tool, the tool should offer a command, such as `mvn dependency:tree` to identify the dependecies
```
olim7t: typo: dependecies => dependencies

faq/README.md
```diff
@@ -258,10 +258,44 @@ If properly used, the following log message will be logged at INFO on startup:
 
 > Found Netty's native epoll transport in the classpath, but NIO was forced through the FORCE_NIO system property.
 
+### Why am I encountering `NoSuchMethodFoundException`, `NoClassDefFoundError`, or `VerifyError`s and how do I avoid them?
+
+Incompatibilities between the java driver and other libraries may cause these exceptions to surface in your
+application at runtime.  
+
+It could be that an older or newer version of a library that the driver depends on, such as Netty
+or Guava, may be present in your application's classpath.  If you are using maven or another dependency
```
olim7t: nit: capitalize Maven?
(same for jar=>JAR, Logback, Netty elsewhere in the paragraph)

faq/README.md
```diff
@@ -258,10 +258,44 @@ If properly used, the following log message will be logged at INFO on startup:
...
+driver.  In this case, observe the stacktrace of the exception to see which library is attempting to use
+the driver.  To identify compatible versions, check that library's dependency on the driver to understand
+what version is compatible.
+
+Finally, some monitoring and agent-based tools such as [DynaTrace] offer solutions that instrument the driver to
+observe and record useful metrics such as request rates, latencies and more.  It is possible that the tool
+you are using is not compatible with the version of the java driver you are using.  In this case, check to
+see if a newer version of that tool is available that works with this version of the driver.  If no such
+version is available, you may want to reach out to the maintainer of that tool to request that they provide
+an update with compatibility to the driver version you are using.
```
olim7t: nit: a bit of repetition of "you are using" in this paragraph

olim7t (changes requested):

olim7t (approved):

PR #956, opened 2018-02-21
JAVA-1448: TokenAwarePolicy should respect child policy ordering

adutra:


driver-core/src/test/java/com/datastax/driver/core/policies/TokenAwarePolicyTest.java
```diff
@@ -255,58 +307,60 @@ public void should_use_other_nodes_when_replicas_having_token_are_down() {
                     .setRoutingKey(routingKey)
                     .setKeyspace("keyspace");
 
+            QueryTracker queryTracker = new QueryTracker();
             queryTracker.query(session, 10, statement);
 
-            // then: The node that is the primary for that key's hash is chosen.
-            queryTracker.assertQueried(sCluster, 1, 1, 0);
+            // then: primary replica is 4, secondary is 1; since the child policy returns [1,2,3,4], the
```
adutra, line 313: The figures changed because the child policy changed.

driver-core/src/main/java/com/datastax/driver/core/policies/TokenAwarePolicy.java
```diff
@@ -23,31 +23,70 @@
...
+     */
+    public enum ReplicaOrdering {
+
+        /**
+         * Order replicas by token ring topology, i.e. always return the "primary" replica first,
+         * then the second, etc., according to the placement of replicas around the token ring.
+         * <p/>
+         * This strategy is the only one guaranteed to order replicas in a deterministic and
+         * constant way. This increases the effectiveness of server-side row caching (especially
+         * at consistency level ONE), but may create hotspots, since the primary replica is always
```
tolbertam: nit: Would change `but may create hotspots` to `but is more heavily impacted by hotspots`.

driver-core/src/main/java/com/datastax/driver/core/policies/TokenAwarePolicy.java
```diff
@@ -56,28 +95,36 @@
...
+    }
+
+    /**
+     * Creates a new {@code TokenAware} policy with {@link ReplicaOrdering#TOPOLOGICAL TOPOLOGICAL} replica ordering.
+     *
+     * @param childPolicy the load balancing policy to wrap with token awareness.
      */
     public TokenAwarePolicy(LoadBalancingPolicy childPolicy) {
-        this(childPolicy, true);
+        this(childPolicy, ReplicaOrdering.TOPOLOGICAL);
```
tolbertam: For consistency with the existing (3.4.0) behavior, shouldn't this be `ReplicaOrdering.RANDOM`?   Previously we shuffled by default, this would change it so it no longer does.

driver-core/src/main/java/com/datastax/driver/core/policies/TokenAwarePolicy.java
```diff
@@ -23,31 +23,70 @@
...
  * token aware policy, replicas from remote data centers may only be
- * returned after all the host of the local data center.
+ * returned after all the hosts of the local data center.
  */
 public class TokenAwarePolicy implements ChainableLoadBalancingPolicy {
 
+    /**
+     * Strategies for replica ordering.
+     */
+    public enum ReplicaOrdering {
```
tolbertam, line 57: Big 👍 on using the enum instead of having 2 booleans (shuffle, and whether to regard child policy).

driver-core/src/main/java/com/datastax/driver/core/policies/TokenAwarePolicy.java
```diff
@@ -94,26 +141,23 @@ public void init(Cluster cluster, Collection<Host> hosts) {
...
-     * them (i.e. mainly if {@code statement.getRoutingKey()} is not {@code null}).
-     * Following what it will return the plan of the child policy.
-     *
-     * @param statement the query for which to build the plan.
-     * @return the new query plan.
+     * The returned plan will first return local replicas for the query (i.e.
+     * replicas whose {@link HostDistance distance} according to the child policy is {@code LOCAL}),
+     * if it can determine them (i.e. mainly if the statement's
+     * {@link Statement#getRoutingKey(ProtocolVersion, CodecRegistry)} routing key}
+     * is not {@code null}); following what it will return the rest of the child policy's original query plan.
```
tolbertam: It would also be good to add that the replicas will be ordered by the configured `ReplicaOrdering`.

driver-core/src/test/java/com/datastax/driver/core/policies/TokenAwarePolicyTest.java
```diff
@@ -58,20 +125,14 @@ public void setUp() {
      * @test_category load_balancing:token_aware
      */
     @Test(groups = "short", dataProvider = "shuffleProvider")
-    public void should_shuffle_replicas_based_on_configuration(Boolean shuffleReplicas) {
+    public void should_order_replicas_based_on_configuration(TokenAwarePolicy.ReplicaOrdering ordering) {
```
tolbertam, line 128: Nice 👍

tolbertam (changes requested):
Looks really great!  I like what you did with using an enum.  I had a few small nits in the comments, and I think we should also retain random ordering with the default constructor to be consistent with previous releases, other than that I think this is good to go.

tolbertam (approved):
changes address all my feedback, looks great 👍

olim7t (approved):

PR #957, opened 2018-02-22
 Create a test-jar for the core module

adutra:


tolbertam:
What do you think about creating a separate module in the same vein as `test-infra`?  I suspect this is being done to access some of the utilities provided in these tests?

On the other hand i'm not sure how that would work, as that test module would likely require code from the core source, so it would create a circular dependency, so maybe what is done here is for the best.

adutra:
> On the other hand i'm not sure how that would work, as that test module would likely require code from the core source, so it would create a circular dependency, so maybe what is done here is for the best.

Unfortunately classes like `RequestHandlerTestHarness` indeed depend on the core classes so that would create a circular dependency. Unless we move all the unit tests to a separate module as well, but that seems overkill.

tolbertam:
> Unfortunately classes like RequestHandlerTestHarness indeed depend on the core classes so that would create a circular dependency. Unless we move all the unit tests to a separate module as well, but that seems overkill.

💯 agreed with the approach you are using in that case then.

tolbertam:
👍, on deploying test-jar.  there are some failing unit tests in travis, are those caused by this?

adutra:
> there are some failing unit tests in travis, are those caused by this?

No these were transient failures, I triggered a new build and it passed.

tolbertam:
> No these were transient failures, I triggered a new build and it passed.

Ah ok, i'll look into that test.  I could have sworn we've had issues with this particular test in the past and made it more resilient, but perhaps there is something more we can do there.

tolbertam (approved):

core/src/test/java/com/datastax/oss/driver/internal/core/CompletionStageAssert.java
```diff
@@ -66,8 +66,31 @@ public CompletionStageAssert(CompletionStage<V> actual) {
     return isFailed(f -> {});
   }
 
+  public CompletionStageAssert<V> isCancelled() {
+    assertThat(actual.toCompletableFuture().isCancelled())
+        .overridingErrorMessage("Expected completion stage to be cancelled")
+        .isTrue();
```
olim7t: It might be good to have a small delay in here, I've had similar assertions fail in CI because of timing issues (see 82d0d280078cb10ec043c544c9cd7e3d8dea6b02).

olim7t (approved):
👍 , reusing RequestHandlerTestHarness outside of the project is really nice to have.
See my comment about future assertions though.

PR #958, opened 2018-02-22
JAVA-1760: Add metrics documentation

tolbertam:
Also includes instructions on how to use Metrics 4.x with the java-driver.

I limited the documentation to how to access the metrics, how to disable them, and how to use Metrics 4.x with the driver.   I think the javadoc (which are linked) does a pretty good job describing what the metrics are, and the Dropwizard metrics documentation provide guidance that I didn't want to duplicate here.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### 3.5.0 (In progress)
+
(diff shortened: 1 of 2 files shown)


olim7t (approved):
👍

PR #960, opened 2018-02-23
JAVA-1536: Add request throttling

olim7t:
Entry points: `RequestThrottler` interface and `request.throttler` in the config.
General design goals: requests register themselves with the throttler, and wait for its signal to start. This is so that various request types (including custom ones) can be all throttled together (since we don't know in advance what the result types are, it would have been complicated to do something generic in `Session.execute`).

TODO:
- [x] more tests
- [x] manual (make sure to explain that throttling is included in client timeout)
- ~~4th implementation that only starts enqueuing when we get `BusyConnectionException` on all nodes (maximizing stream id usage in a way)~~

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/metrics/DefaultSessionMetric.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/metrics/DefaultSessionMetric.java
@@ -23,6 +23,9 @@ public enum DefaultSessionMetric implements SessionMetric {
   CONNECTED_NODES("connected-nodes"),
   CQL_REQUESTS("cql-requests"),
   CQL_CLIENT_TIMEOUTS("cql-client-timeouts"),
+  THROTTLING_DELAY("throttling.delay"),
+  THROTTLING_QUEUE_SIZE("throttling.queue-size"),
+  THROTTLING_ERRORS("throttling.errors"),
   ;
 
   private static final Map<String, DefaultSessionMetric> BY_PATH = sortByPath();
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRequestHandler.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRequestHandler.java
@@ -77,7 +77,7 @@ public class AdminRequestHandler implements ResponseCallback {
   private final Duration timeout;
   private final String logPrefix;
   private final String debugString;
-  private final CompletableFuture<AdminResult> result = new CompletableFuture<>();
+  protected final CompletableFuture<AdminResult> result = new CompletableFuture<>();
 
   // This is only ever accessed on the channel's event loop, so it doesn't need to be volatile
   private ScheduledFuture<?> timeoutFuture;
@@ -110,12 +110,12 @@ public class AdminRequestHandler implements ResponseCallback {
           channel.eventLoop().schedule(this::fireTimeout, timeout.toNanos(), TimeUnit.NANOSECONDS);
       timeoutFuture.addListener(UncaughtExceptions::log);
     } else {
-      result.completeExceptionally(future.cause());
+      setFinalError(future.cause());
     }
   }
 
   private void fireTimeout() {
-    result.completeExceptionally(
+    setFinalError(
         new DriverTimeoutException(String.format("%s timed out after %s", debugString, timeout)));
     if (!channel.closeFuture().isDone()) {
       channel.cancel(this);
@@ -127,7 +127,7 @@ public class AdminRequestHandler implements ResponseCallback {
     if (timeoutFuture != null) {
       timeoutFuture.cancel(true);
     }
-    result.completeExceptionally(error);
+    setFinalError(error);
   }
 
   @Override
@@ -141,14 +141,22 @@ public class AdminRequestHandler implements ResponseCallback {
       Rows rows = (Rows) message;
       ByteBuffer pagingState = rows.getMetadata().pagingState;
       AdminRequestHandler nextHandler = (pagingState == null) ? null : this.copy(pagingState);
-      result.complete(new AdminResult(rows, nextHandler, channel.protocolVersion()));
+      setFinalResult(new AdminResult(rows, nextHandler, channel.protocolVersion()));
     } else if (message instanceof Prepared) {
       // Internal prepares are only "reprepare on up" types of queries, where we only care about
       // success, not the actual result, so this is good enough:
-      result.complete(null);
+      setFinalResult(null);
     } else {
-      result.completeExceptionally(new UnexpectedResponseException(debugString, message));
+      setFinalError(new UnexpectedResponseException(debugString, message));
     }
+  }
+
+  protected boolean setFinalResult(AdminResult result) {
+    return this.result.complete(result);
+  }
+
+  protected boolean setFinalError(Throwable error) {
+    return result.completeExceptionally(error);
   }
 
   private AdminRequestHandler copy(ByteBuffer pagingState) {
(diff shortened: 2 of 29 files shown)


core/src/main/resources/reference.conf
```diff
@@ -475,6 +475,58 @@ datastax-java-driver {
...
+    # throttler has determined that it can neither allow the request to proceed now, nor enqueue it;
+    # this indicates that your session is overloaded.
+    throttler {
+      # The following implementations are provided out of the box (all located in the package
+      # com.datastax.oss.driver.internal.core.session.throttling):
+      #
+      # - PassThroughRequestThrottler: does not perform any kind of throttling, all requests are
+      #   allowed to proceed immediately. Required options: none.
+      #
+      # - ConcurrencyLimitingRequestThrottler: limits the number of requests that can be executed
```
tolbertam, line 501: Did you consider allowing a means of limiting both the concurrency and the rate?  Or do you think that is too complex / doesn't offer more value compared to choosing one over the other?

core/src/main/java/com/datastax/oss/driver/internal/core/session/throttling/ConcurrencyLimitingRequestThrottler.java
```diff
@@ -0,0 +1,209 @@
...
+    final boolean closed;
+
+    State(int concurrentRequests, int queueSize, boolean closed) {
+      this.concurrentRequests = concurrentRequests;
+      this.queueSize = queueSize;
+      this.closed = closed;
+    }
+
+    State acquirePermit() {
+      assert !closed && queueSize == 0;
```
tolbertam: Took me a bit to make sense of these assertions, but eventually wrapped my head around it (queueSize should always be 0 here because caller only calls `acquirePermit` when queue is empty, otherwise it calls enqueue).   Might be helpful to add comments describing reason for assertions.

core/src/main/java/com/datastax/oss/driver/internal/core/session/throttling/RateLimitingRequestThrottler.java
```diff
@@ -0,0 +1,322 @@
...
+
+  @VisibleForTesting final AtomicReference<State> stateRef;
+  @VisibleForTesting final Deque<Throttled> queue = new ConcurrentLinkedDeque<>();
+
+  public RateLimitingRequestThrottler(DriverContext context) {
+    this(context, System::nanoTime);
+  }
+
+  @VisibleForTesting
+  RateLimitingRequestThrottler(DriverContext context, NanoClock clock) {
```
tolbertam, line 62: Nice idea to use a `Clock` implementation for testing 👍

core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRequestHandler.java
```diff
@@ -110,12 +110,12 @@ private void onWriteComplete(Future<? super Void> future) {
...
       timeoutFuture.addListener(UncaughtExceptions::log);
     } else {
-      result.completeExceptionally(future.cause());
+      setFinalError(future.cause());
     }
   }
 
   private void fireTimeout() {
-    result.completeExceptionally(
+    setFinalError(
```
tolbertam, line 118: Does the time a request spend throttled factor into the client timeout?  From what I can tell I think it is.  I think depending on how you are using the throttle (i.e. hammering hard and using driver to limit your throughput), it's like a user would also want to raise the client timeout.   I think in documentation we should probably want to make that clear that the time a request spends throttled factors in client timeout.

tolbertam (commented):
Looking really good!  The RequestThrottler implementations were really easy to follow.  Setting up an environment to play with this and will provide feedback likely later today.

core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRequestHandler.java
olim7t: Yes, it does factor into the client timeout. I think it's mentioned somewhere in the configuration comments, I'll make sure to explain it in the manual too.

tolbertam: ah yes!  I do see that, it's explained pretty well there too 👍

core/src/main/resources/reference.conf
```diff
@@ -475,6 +475,58 @@ datastax-java-driver {
...
+      #   in parallel. Required options: max-concurrent-requests, max-queue-size.
+      #
+      # - RateLimitingRequestThrottler: limits the request rate per second. Required options:
+      #   max-requests-per-second, max-queue-size, drain-interval.
+      class = com.datastax.oss.driver.internal.core.session.throttling.PassThroughRequestThrottler
```
tolbertam, line 506: Was just thinking that other classes, like `ConstantSpeculativeExecutionPolicy` are in the `api.core` package rather than `internal`.  What should the policy be on this?  Should implementations we offer for use in config files be in the api package or internal?

Another thing that wasn't sure if we discussed or not in the past:  Should we consider having a default package namespace when an instance is called for?  Similar to cassandra.yaml's authenticator/authorizer settings, instead of providing `org.apache.cassandra.auth.AllowAllAuthenticator`, one could simply specify `AllowAllAuthenticator`.

olim7t: Agreed on both points. I've logged [JAVA-1774](https://datastax-oss.atlassian.net/browse/JAVA-1774) and [JAVA-1775](https://datastax-oss.atlassian.net/browse/JAVA-1775).

core/src/main/java/com/datastax/oss/driver/api/core/metrics/DefaultSessionMetric.java
```diff
@@ -23,6 +23,9 @@
   CONNECTED_NODES("connected-nodes"),
   CQL_REQUESTS("cql-requests"),
   CQL_CLIENT_TIMEOUTS("cql-client-timeouts"),
+  THROTTLING_DELAY("throttling.delay"),
```
tolbertam, line 26: I was thinking that it'd also be a nice to have a way to measure how many requests are being throttled vs. not throttled, but I think it can already be inferred from cql-requests.count - throttling.delay.count, so no need for that 👍

manual/core/throttling/README.md
```diff
@@ -0,0 +1,143 @@
+## Request throttling
```
tolbertam, line 1: Excellent docs!

tolbertam (commented):
Looks great!  I'd like to do another pass experimenting with this, which will also gave me a chance to play with the new metrics.  Will do that tomorrow morning.

tolbertam (approved):
Experimented with latest changes on Friday afternoon and today and this looks to be working great!

The new metrics are really nice, here's a snapshot comparing throttling delay vs. session request times (and including the other throttling metrics):

metrics

Since my scenario only 2048 requests at a time, the queue size should never exceed that (it never did) and should have encountered 0 throttle errors.  My configuration used rate limiting at 20k requests/sec:

```
throttler {
  class = com.datastax.oss.driver.internal.core.session.throttling.RateLimitingRequestThrottler
  max-queue-size = 100000
  max-requests-per-second = 20000
  drain-interval = 10 milliseconds
}
```

I also ran a few scenarios where exceeding the queue was expected.  In this case I did encounter some throttling errors (and the metric incremented):

throttle-errors

tolbertam:
I don't observe much of a discernible difference with the nanoTime optimization, but it also seems to be working well so I'm 👍 on including it if you like.

PR #962, opened 2018-02-26
JAVA-1752: Deprecate DowngradingConsistencyRetryPolicy

adutra:


driver-examples/src/main/java/com/datastax/driver/examples/retry/DowngradingRetry.java
```diff
@@ -0,0 +1,467 @@
...
+                // then retry with the initial consistency level, we might be luckier next time
+                // and get the data back.
+                if (!readTimeout.wasDataRetrieved()) {
+
+                    return read(cl, retryCount + 1);
+
+                }
+
+                // Otherwise, abort since the read timeout is probably a more serious problem.
+                throw e;
```
adutra: A question came to my mind while writing this example: in which circumstances can we get here? If we got the data _and_ enough acks, how can the read timeout?

driver-examples/src/main/java/com/datastax/driver/examples/retry/DowngradingRetry.java
```diff
@@ -0,0 +1,470 @@
...
+                // then retry with the initial consistency level, we might be luckier next time
+                // and get the data back.
+                if (!readTimeout.wasDataRetrieved()) {
+
+                    return read(cl, retryCount + 1);
+
+                }
+
+                // Otherwise, abort since the read timeout is unlikely to be solved by a retry.
+                throw e;
```
adutra, line 338: A question came to my mind while writing this example: in which circumstances can we get here? If we got the data and enough acks, how can the read timeout?

upgrade_guide/README.md
```diff
@@ -3,6 +3,35 @@
...
+
+`DowngradingConsistencyRetryPolicy` is now deprecated, see [JAVA-1752]. 
+It will also be removed in the next major release of the driver (4.0.0), 
+see [JAVA-1376].
+
+The main motivation is the agreement that `DowngradingConsistencyRetryPolicy`'s 
+behavior should be the application's concern, not the driver's. APIs provided by 
+the driver should instead encourage idiomatic use of a distributed system like 
+Apache Cassandra, and a downgrading policy works against this. It suggests that 
+an anti-pattern such as "try to read at ALL, but fall back to QUORUM if that fails"
```
tolbertam: nit: `try to read at QUORUM, but fall back to ONE`, would be probably the most likely scenario so would be better to document that.

upgrade_guide/README.md
```diff
@@ -3,6 +3,35 @@
 The purpose of this guide is to detail changes made by successive
 versions of the Java driver.
 
+### 3.5.0
+
+`DowngradingConsistencyRetryPolicy` is now deprecated, see [JAVA-1752].
```
tolbertam: Great explanation

driver-examples/src/main/java/com/datastax/driver/examples/retry/DowngradingRetry.java
tolbertam: I don't think it's possible, except maybe for an odd timing issue on the C* side, where it times out waiting for replicas, but then acks/data comes in after the timeout.  I seem to recall seeing that before and being puzzled.

driver-examples/src/main/java/com/datastax/driver/examples/retry/DowngradingRetry.java
```diff
@@ -0,0 +1,470 @@
...
+    }
+
+    /**
+     * Inserts data, retrying if necessary with a downgraded CL.
+     *
+     * @param cl         the consistency level to apply.
+     * @param retryCount the current retry count.
+     * @throws DriverException if the current consistency level cannot be downgraded.
+     */
+    private void write(ConsistencyLevel cl, int retryCount) {
```
tolbertam, line 154: The use case you describe in the docs:

>  We recognize that there are use cases where downgrading is 
good – for instance, a dashboard application would present the latest information 
by reading at QUORUM, but it's acceptable for it to display stale information 
by reading at ONE sometimes. Users will now have to implement this logic directly
in their application, or alternatively, implement their own retry policy.

Wouldn't that only apply to reads?  Should we only show examples for application-level retries for reads?

upgrade_guide/README.md
```diff
@@ -3,6 +3,35 @@
...
+Instead, users should choose upfront the consistency level that works best for 
+their use cases. We recognize that there are use cases where downgrading is 
+good – for instance, a dashboard application would present the latest information 
+by reading at QUORUM, but it's acceptable for it to display stale information 
+by reading at ONE sometimes. Users will now have to implement this logic directly
+in their application, or alternatively, implement their own retry policy.
+
+To help users migrate existing applications that rely on 
+`DowngradingConsistencyRetryPolicy`, see this 
+[online example](../driver-examples/src/main/java/com/datastax/driver/examples/retry/DowngradingRetry.java).
```
olim7t: This link won't work in the generated docs on docs.datastax.com. I think the best alternative is to change it to:
https://github.com/datastax/java-driver/blob/3.x/driver-examples/src/main/java/com/datastax/driver/examples/retry/DowngradingRetry.java

driver-examples/src/main/java/com/datastax/driver/examples/retry/DowngradingRetry.java
```diff
@@ -0,0 +1,470 @@
...
+import com.datastax.driver.core.exceptions.UnavailableException;
+import com.datastax.driver.core.exceptions.WriteTimeoutException;
+import java.text.SimpleDateFormat;
+
+import static com.datastax.driver.core.BatchStatement.Type.UNLOGGED;
+import static com.datastax.driver.core.ConsistencyLevel.QUORUM;
+
+/**
+ * This example illustrates how to replace the deprecated
+ * {@link com.datastax.driver.core.policies.DowngradingConsistencyRetryPolicy DowngradingConsistencyRetryPolicy}
```
olim7t: nit: is it really that bad to import classes that are only linked from javadocs? I've been doing it in 4.x because I can't find a good reason not to...

olim7t (approved):

driver-examples/src/main/java/com/datastax/driver/examples/retry/DowngradingRetry.java
adutra: It is said that some static analysis tools (e.g. Checkstyle) tend to report such imports as "unused". But OK for replacing them with imports.

driver-examples/src/main/java/com/datastax/driver/examples/retry/DowngradingRetry.java
adutra: It's indeed imo harder to find valid examples of downgrading writes than it is for reads, but I think we should strive to give a complete example that works for both reads and writes.

PR #963, opened 2018-02-28
JAVA-1735: Log driver version on first use (3.x)

adutra:


driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -201,6 +202,17 @@ public static String getDriverVersion() {
...
 
+    /**
+     * Logs the driver version to the console.
+     * <p>
+     * This method logs the version using the logger {@code com.datastax.driver.core}
+     * and level {@code INFO}.
+     */
+    public static void logDriverVersion() {
+        Logger core = LoggerFactory.getLogger("com.datastax.driver.core");
+        core.info("DataStax Java driver for Apache Cassandra version {}", getDriverVersion());
```
mbulman: The wording of this might be easy to misunderstand as the DataStax Java driver for: "Apache Cassandra version X".  In other words, mistaking the version logged as being for Apache Cassandra and not the java driver.  Does something like "DataStax Java driver 1.2.3 for Apache Cassandra" satisfy legal requirements?

adutra: Very good point, I will amend, thanks! 👍

tolbertam (approved):

olim7t (approved):

PR #964, opened 2018-02-28
JAVA-1803 (1735): Log driver version on first use (4.x)

adutra:


core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -200,6 +199,17 @@ protected DriverContext buildContext(
     return new DefaultDriverContext(configLoader, typeCodecs);
   }
 
+  /**
+   * This <b>must</b> return an instance of {@code DefaultSession} (it's not expressed directly in
+   * the signature to avoid leaking that type through the protected API).
+   */
+  protected Session buildDefaultSession(
```
adutra: I saw this warning in the comments of `buildContext` and had to do the same for this method. I think it is a code smell that indicates that we should rather extract an interface from this class and keep the implementation details in an internal class.

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
```diff
@@ -74,14 +79,8 @@
 public class DefaultSession implements CqlSession {
 
   private static final Logger LOG = LoggerFactory.getLogger(DefaultSession.class);
-
-  public static CompletionStage<CqlSession> init(
```
adutra: This factory method wasn't really bringing any value, this class is internal so the constructor can be public.

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
```diff
@@ -107,11 +106,50 @@ private DefaultSession(
...
+      if (Native.isGetProcessIdAvailable()) {
+        processId = Native.getProcessId();
+      }
+      long start = System.nanoTime();
+      STARTUP_LOG.info("[{}] {} initializing (PID: {})", logPrefix, driverInfo, processId);
+      singleThreaded.initFuture.thenRun(
+          () -> {
+            Duration elapsed =
+                Duration.ofMillis(TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start));
+            String elapsedStr =
```
adutra: Just a pretty-print for durations. It is a common enough problem that I think we should have an utility method somewhere to print durations in good ol' English. Currently our log messages print durations in ISO-8601 format, which is not very user-friendly – but that's a different issue.

core/src/main/resources/reference.conf
```diff
@@ -102,6 +102,32 @@ datastax-java-driver {
...
+
+    # An optional custom banner to be printed at startup. This can be useful to differentiate
+    # between two applications using the driver.
+    #
+    # Tip 1: this option works best with multi-line strings, as in the example below.
+    #
+    # Tip 2: you can include system properties, environment variables and/or references to any
+    # other configuration option, as in the example below. Beware that any reference must point
+    # to an existing value, otherwise an error will be raised.
+    //custom-banner = """
```
adutra: This really shows all the potential of TypeSafe Config imo.

core/src/main/java/com/datastax/oss/driver/internal/core/DefaultDriverInfo.java
```diff
@@ -0,0 +1,83 @@
...
+        new InputStreamReader(resource.openStream(), StandardCharsets.UTF_8)) {
+      Properties props = new Properties();
+      props.load(reader);
+      String name = props.getProperty("driver.name");
+      String groupId = props.getProperty("driver.groupId");
+      String artifactId = props.getProperty("driver.artifactId");
+      String version = props.getProperty("driver.version");
+      return new DefaultDriverInfo(name, groupId, artifactId, Version.parse(version));
+    } catch (IOException e) {
+      throw new UncheckedIOException(e);
```
olim7t, line 53: Nice, didn't know this exception.

core/src/main/java/com/datastax/oss/driver/api/core/Version.java
```diff
@@ -23,24 +23,41 @@
...
+ * the DataStax Java driver.
  *
  * <p>Version numbers compare the usual way, the major number (X) is compared first, then the minor
  * one (Y) and then the patch level one (Z). Lastly, versions with pre-release sorts before the
  * versions that don't have one, and labels are sorted alphabetically if necessary. Build metadata
- * are ignored for sorting versions.
+ * are ignored when sorting versions.
  */
-public class CassandraVersion implements Comparable<CassandraVersion> {
+public class Version implements Comparable<Version> {
```
olim7t: Not a huge deal, but it would have been nice to isolate the name change in a dedicated commit. It's both trivial and the most impacting in terms of changed files, so it would make the other changes much easier to review.

Just a thing to keep in mind for future refactorings like this.

core/src/main/java/com/datastax/oss/driver/api/core/Version.java
```diff
@@ -23,24 +23,41 @@
...
 
   private static final String VERSION_REGEXP =
-      "(\\d+)\\.(\\d+)(\\.\\d+)?(\\.\\d+)?([~\\-]\\w[.\\w]*(?:\\-\\w[.\\w]*)*)?(\\+[.\\w]+)?";
+      "(\\d+)\\.(\\d+)(\\.\\d+)?(\\.\\d+)?([~\\-]\\w[.\\w]*(?:-\\w[.\\w]*)*)?(\\+[.\\w]+)?";
```
olim7t: I think I just blew a few neurons trying to parse that change 😉 

If I'm following correctly, the dash separating pre-release labels &mdash; as in "rc1-SNAPSHOT" &mdash; does not need to be escaped. Is it just unnecessary or was it failing?

core/src/main/java/com/datastax/oss/driver/api/core/Version.java
```diff
@@ -23,24 +23,41 @@
...
-  public static final CassandraVersion V2_2_0 = parse("2.2.0");
-  public static final CassandraVersion V3_0_0 = parse("3.0.0");
+  /** Apache Cassandra's version 2.1.0. */
+  public static final Version CASSANDRA_2_1_0 = parse("2.1.0");
+
+  /** Apache Cassandra's version 2.2.0. */
+  public static final Version CASSANDRA_2_2_0 = parse("2.2.0");
+
+  /** Apache Cassandra's version 3.0.0. */
+  public static final Version CASSANDRA_3_0_0 = parse("3.0.0");
```
olim7t: Why use "Cassandra" in the name if the version is now generic? This could apply to any tool having a 3.0.0 version.

core/src/main/resources/reference.conf
olim7t: 👍 Nice, especially considering that in a typical deployment this can be changed without recompiling.

core/src/main/resources/reference.conf
```diff
@@ -102,6 +102,32 @@ datastax-java-driver {
...
 
+  # A set of customizable actions that are performed when the driver starts up, i.e.,
+  # each time a new session is initialized.
+  # Actions in this section that involve printing log messages will always use the special logger
+  # 'com.datastax.oss.driver.api.core.Startup' at level INFO.
+  startup {
+
+    # Whether or not to print basic information about the initialization process,
+    # such as the driver name, its version number, and the process PID, if available.
+    print-basic-info = true
```
olim7t: Why not combine the two options? We could have a single `startup-log` that defaults to the basic info, and then you can customize it, or comment it out to disable the log.

Artifact coordinates can't use regular interpolation, but we could come up with a custom one and use a simple search and replace:
```
startup-log = "Starting #{driver.groupId}:#{driver.artifactId}:#{driver.version}"
```

core/src/main/java/com/datastax/oss/driver/api/core/DriverInfo.java
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core;
+
+/** Exposes information about the driver. */
+public interface DriverInfo {
```
olim7t, line 19: "Info" is a bit generic. What this really is is the artifact coordinates.

core/src/main/java/com/datastax/oss/driver/api/core/DriverInfo.java
```diff
@@ -0,0 +1,32 @@
...
+
+  /** @return the driver full name. */
+  String getName();
```
olim7t, line 22: Not sure if this is really needed, it doesn't bring any additional info than the artifactId.

On the other hand, it would be nice to know if the driver is shaded. We can't do it through `Driver.properties` because it's generated at compile time, but we could do a runtime check on one of the Netty classes.

core/src/main/java/com/datastax/oss/driver/api/core/session/Session.java
```diff
@@ -51,15 +52,14 @@
...
    *
    * <p>This is intended for products that wrap or extend the driver, as a way to check
    * compatibility if end-users override the driver version in their application.
    */
-  static String getCoreDriverVersion() {
-    // Note: getBundle caches its result
-    return ResourceBundle.getBundle("com.datastax.oss.driver.Driver").getString("driver.version");
+  default DriverInfo getDriverInfo() {
+    return DefaultDriverInfo.buildFromResource(
+        getClass().getResource("/com/datastax/oss/driver/Driver.properties"));
```
olim7t: 👎 on making this an instance method.
There is some value in being able to test the driver version before creating a session. I could imagine some 3rd-party tool needing that.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfigProfile.java
```diff
@@ -181,7 +181,7 @@ void register(Derived derivedProfile) {
   }
 
   /**
-   * A profile that was copied from another profile programatically using {@code withXxx} methods.
+   * A profile that was copied from another profile programmatically using {@code withXxx} methods.
```
olim7t, line 232: 👍 , but it feels a bit weird in this PR since this class is not impacted by any other change (don't remove it, but let's fix unrelated typos in separate commits in the future).

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
```diff
@@ -74,14 +79,8 @@
...
-
-  public static CompletionStage<CqlSession> init(
-      InternalDriverContext context,
-      Set<InetSocketAddress> contactPoints,
-      CqlIdentifier keyspace,
-      Set<NodeStateListener> nodeStateListeners) {
-    return new DefaultSession(context, contactPoints, nodeStateListeners).init(keyspace);
-  }
+  private static final Logger STARTUP_LOG =
+      LoggerFactory.getLogger("com.datastax.oss.driver.api.core.Startup");
```
olim7t: Not sure if we need a dedicated log, the message can already be disabled through the configuration if it's unwanted.

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
olim7t: I wrote `NanoTime.format` to that effect for token map logs.

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
```diff
@@ -107,11 +106,50 @@ private DefaultSession(
...
+                Duration.ofMillis(TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - start));
+            String elapsedStr =
+                elapsed.toString().substring(2).replaceAll("(\\d[HMS])(?!$)", "$1 ").toLowerCase();
+            int size = getPools().size();
+            STARTUP_LOG.info(
+                "[{}] Session successfully initialized in {}, connected to {} node{}",
+                logPrefix,
+                elapsedStr,
+                size,
+                size > 1 ? "s" : "");
```
olim7t: I would prefer to have a single message (so the second one after the session is initialized).

Initialization time sounds nice to have, not sure if the PID will be that useful but why not. If we go with my suggestion on the single configuration option, these would have to be added to the list of custom interpolation variables.

olim7t (changes requested):
Interesting suggestions in this PR 👍 

I like the extended driver information class, and love the customizable message.

However I would keep the getter method static and avoid all the impacts on the session class.

core/src/main/java/com/datastax/oss/driver/api/core/Version.java
adutra: Just unnecessary (according to IntelliJ).

core/src/main/java/com/datastax/oss/driver/api/core/Version.java
adutra: Yes but the constant names would look weird: why have `_2_1_0` and `_2_2_0` but not `_2_3_0`, `_2_4_0`, etc.?
I think a more elegant approach would be to declare these constants in a dedicated class, will try that.

core/src/main/java/com/datastax/oss/driver/api/core/DriverInfo.java
adutra: As you pointed out below, we might want to enrich this interface with more info than just the GAV coordinates, which is why I chose the word "Info".

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
adutra: My idea was that users might want to attach special appenders to this logger and/or switch off its additivity to give these messages a special treatment (think monitoring tools that are only interested in knowing how many sessions were successfully created, and when).

core/src/main/resources/reference.conf
adutra: 1. I like the idea of combining everything into one banner.
2. I don't like mixing two interpolation levels, I think users will get confused.

I did try today to come up with a solution entirely handled internally by TypeSafe Config, and using the new `ConfigResolver` API. It looked very promising, e.g. the following could have been the default banner:

```
banner = ${oss-driver.name} (${oss-driver.groupId}":"${oss-driver.artifactId}) version ${oss-driver.version} (PID ${pid})
```

But unfortunately it seems that there is a bug (or feature) in this new API, see https://github.com/lightbend/config/issues/553.

In the meanwhile, I'm having second thoughts about these banners. An even better solution would be to implement a listener interface for `Session` lifecycle events (`onSessionInitializing`, `onSessionInitialized`, `onSessionClosing`, etc.), and provide by default an implementation that logs basic driver info + time elapsed when the session is fully initialized.

Unless anyone disagrees, I will implement the idea of a listener.

core/src/main/java/com/datastax/oss/driver/api/core/session/Session.java
adutra: Fair enough, will revert to static. But in this case I think a more explicit method name is better, e.g. `getOssDriverInfo()`.

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
adutra: Ok to everything, but will try to go forward with my idea of a Session lifecycle listener interface.

core/src/main/resources/reference.conf
adutra: It turns out the bug is a feature :) so I suggest that we abandon the idea of a custom banner in favor of a listener.

olim7t:
> An even better solution would be to implement a listener interface for Session lifecycle events (onSessionInitializing, onSessionInitialized, onSessionClosing, etc.)

Mmm, that exists already:
* onSessionInitializing => when you call build/buildAsync on CqlSession.builder()
* onSessionInitialized => use buildAsync and add a whenComplete listener on the returned future
* onSessionClosed => add a whenComplete listener on session.closeFuture()

If we abandon the idea of a custom banner, how about going back to the original intent of this ticket: "Log driver version on first use"

olim7t:
FYI I've created a separate ticket for driver 4: [JAVA-1803](https://datastax-oss.atlassian.net/browse/JAVA-1803). No need to rename the branch at this point.

adutra:
OK I updated this PR and went back to something much simpler. No more fancy startup banners, just a (static) method that exposes the driver info.

olim7t:
Closing in favor of #1002. I couldn't find a simple way to rebase this, in particular because of the weird merge commit in the middle (which also reverts part of the lifecycle listener stuff...)

I've cherry-picked the two unrelated changes (bc8d368 and 3bdab03) directly on 4.x. The `Version` rename is on the other PR, but I didn't take all of your other changes (javadoc fixes etc.), feel free to open another PR.

I'm going to update the contribution guidelines wrt to merging and rebasing, please read them, we need to find a way to make this work better.

PR #967, opened 2018-03-02
Java1702

GregBestland:


core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigProfile.java
```diff
@@ -42,6 +43,8 @@
 
   int getInt(DriverOption option);
 
+  Map<String, String> getMap(DriverOption option);
```
olim7t: 1. Could you call it `getStringMap` instead?
2. For completeness, we should have a `withStringMap` as well. It can't reuse `with(DriverOption, Object)` like the others because it's not a primitive type, but I think it should be pretty easy to adapt (basically, we need to convert the other way Map => Config).
3. nit: could you move the `getStringMap/withStringMap` pair under `withStringList`, and use the same order in the implementation class?

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfigProfile.java
```diff
@@ -72,6 +76,19 @@ public Duration getDuration(DriverOption option) {
     return getCached(option.getPath(), getEffectiveOptions()::getDuration);
   }
 
+  @Override
+  public Map<String, String> getMap(DriverOption option) {
+    Config subConfig = getCached(option.getPath(), getEffectiveOptions()::getConfig);
+    Map<String, String> map = new HashMap<>();
+    Set<Map.Entry<String, ConfigValue>> entrySet = subConfig.entrySet();
+    for (Map.Entry<String, ConfigValue> entry : entrySet) {
```
olim7t: nit: maybe inline `entrySet`?

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfigProfile.java
```diff
@@ -72,6 +76,19 @@ public Duration getDuration(DriverOption option) {
...
+      if (entry.getValue().valueType().equals(ConfigValueType.STRING)) {
+        map.put(entry.getKey(), (String) entry.getValue().unwrapped());
```
olim7t: 👍 I agree that ignoring non-string fields is probably the best thing to do.

core/src/test/java/com/datastax/oss/driver/internal/core/config/typesafe/TypeSafeDriverConfigTest.java
```diff
@@ -94,6 +95,20 @@ public void should_create_derived_profile_overriding_option() {
     assertThat(derived.getInt(MockOptions.REQUIRED_INT)).isEqualTo(43);
   }
 
+  @Test
+  public void should_be_able_to_fecth_maph() {
```
olim7t: spelling: `fetch_map`?

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ProtocolInitHandler.java
```diff
@@ -292,8 +292,14 @@ void onResponse(Message response) {
...
-      } catch (AuthenticationException e) {
-        fail(e);
+      } catch (Exception e) {
+        if (e instanceof AuthenticationException) {
+          fail(e);
+        } else {
+          fail(
+              new AuthenticationException(
+                  channel.remoteAddress(), "Unexpected exception in encoutered in response ", e));
+        }
```
olim7t: I think two separate catch blocks would be more idiomatic:
```java
} catch (AuthenticationException e) {
  fail(e);
} catch (Throwable t) {
  fail(
      new AuthenticationException(
          channel.remoteAddress(), "Unexpected exception in encoutered in response ", t));
}
```
(also, using Throwable for the generic one)

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfigProfile.java
```diff
@@ -72,6 +76,19 @@ public Duration getDuration(DriverOption option) {
     return getCached(option.getPath(), getEffectiveOptions()::getDuration);
   }
 
+  @Override
+  public Map<String, String> getMap(DriverOption option) {
+    Config subConfig = getCached(option.getPath(), getEffectiveOptions()::getConfig);
+    Map<String, String> map = new HashMap<>();
```
olim7t: Maybe I'm being paranoid, but I think it would be safer to return an immutable map. See `DefaultConsistencyLevel.mapByCode` for an example of how to build it.

olim7t (changes requested):

core/src/test/java/com/datastax/oss/driver/internal/core/config/typesafe/TypeSafeDriverConfigTest.java
```diff
@@ -96,19 +97,33 @@ public void should_create_derived_profile_overriding_option() {
...
+
+    TypeSafeDriverConfig config = parse("required_int = 42");
+    Map<String, String> authThingMap = new HashMap<>();
+    authThingMap.put("auth_thing_one", "one");
+    authThingMap.put("auth_thing_two", "two");
+    authThingMap.put("auth_thing_three", "three");
+    DriverConfigProfile base = config.getDefaultProfile();
+    DriverConfigProfile mapBase = base.withStringMap(MockOptions.OPTIONAL_AUTH, authThingMap);
+    Map<String, String> fetchedMap = mapBase.getStringMap(MockOptions.OPTIONAL_AUTH);
+    assertThat(fetchedMap).isEqualTo(authThingMap);
```
olim7t, line 123: :+1:

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfigProfile.java
```diff
@@ -109,11 +96,38 @@ public DriverConfigProfile withString(DriverOption option, String value) {
...
 
+  @Override
+  public DriverConfigProfile withStringMap(DriverOption option, Map<String, String> map) {
+    Base base = getBaseProfile();
+    // Add the new option to any already derived options
+    Config newAdded = getAddedOptions();
+    for (String key : map.keySet()) {
+      newAdded =
+          newAdded.withValue(
+              option.getPath() + "." + key, ConfigValueFactory.fromAnyRef(map.get(key)));
```
olim7t, line 124: Nice, even simpler than I thought.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ProtocolInitHandler.java
```diff
@@ -294,6 +294,8 @@ void onResponse(Message response) {
         }
       } catch (AuthenticationException e) {
         fail(e);
+      } catch (Throwable t) {
+        fail("Unexpected exception at step " + step, t);
```
olim7t, line 298: I changed this at the last minute because it could in theory be something else than an authentication exception. I've checked that it's properly surfaced to the thread that creates the session.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfig.java
```diff
@@ -33,9 +33,9 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class TypeSafeDriverConfig implements DriverConfig {
+public class TypesafeDriverConfig implements DriverConfig {
```
olim7t, line 36: I noticed that I had been mixing different capitalizations, fixing in a separate commit.

olim7t (approved):

PR #969, opened 2018-03-03
Update TokenRange toString “]%s, %s]” to "[%s, %s]"

johnyannj:
a little mistake?

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/TokenRange.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/TokenRange.java
@@ -322,7 +322,7 @@ public final class TokenRange implements Comparable<TokenRange> {
 
     @Override
     public String toString() {
-        return String.format("]%s, %s]", start, end);
+        return String.format("[%s, %s]", start, end);
     }
 
     @Override


tolbertam:
Hi @johnyannj , thank you for your contribution!

I will admit it does look odd, but this is actually by design.  token ranges in cassandra are start-token exclusive and end-token inclusive, so when you see a range with a format of `]10, 20]` it means "all tokens from 10 to 20, excluding 10, but including 20.

One common notation of intervals is to use square-brackets to indicate membership.  From the wikipedia article for [Brackets in Mathematics under the Intervals section](https://en.wikipedia.org/wiki/Bracket_(mathematics)#Intervals):

> Both parentheses, `( )`, and square brackets, `[ ]`, can also be used to denote an interval. The notation `[a, c)` is used to indicate an interval from `a` to `c` that is inclusive of `a` but exclusive of `c`. That is, `[5,12)` would be the set of all real numbers between 5 and 12, including 5 but not 12. The numbers may come as close as they like to 12, including 11.999 and so forth (with any finite number of 9s), but 12.0 is not included. **In some European countries, the notation `[5,12[` is also used for this**.

johnyannj:
@tolbertam 

Thank you for your explaining
I learned more~~

olim7t:
Ha, same here, I didn't know this was a european thing 😃

PR #970, opened 2018-03-05
JAVA-1777: Backport JAVA-444 to 3.0.x

pkolaczk:
Fixes duplicate UUIDs generation.


tolbertam (approved):
Looks great 👍 .  Some of our CI failed because this is an older branch that we haven't maintained for a while, but don't think we need to worry about that as this change is fairly isolated.

PR #971, opened 2018-03-05
JAVA-1769: Allocate exact buffer size for outgoing requests

olim7t:
I re-ran the whole integration suite, and the native-protocol unit tests already cover message sizes.


core/src/test/java/com/datastax/oss/driver/internal/core/protocol/ByteBufPrimitiveCodecTest.java
```diff
@@ -240,7 +240,7 @@ public void should_fail_to_read_long_string_if_not_enough_characters() {
     expectedException.expect(IllegalArgumentException.class);
     expectedException.expectMessage(
         "Not enough bytes to read an UTF-8 serialized string of size 4");
-    ByteBuf source = codec.allocate(2);
+    ByteBuf source = codec.allocate(4);
```
olim7t, line 243: This was wrong, we need space to write at least the size.

adutra (approved):

PR #973, opened 2018-03-09
3.x add update query mapper

vagifzeynalov:
Here is the implementation of **UPDATE** query mapper. Especially it is very useful with **IfExists** option.


vagifzeynalov:
Found some minor bugs... will fix then and open another PR.

PR #974, opened 2018-03-09
JAVA-2042: Add mapper support for UPDATE queries

vagifzeynalov:
Here is the implementation of **UPDATE** query mapper. Especially it is very useful together with **IfExists** option.


vagifzeynalov:
is there any chance someone will review it?

adutra:

@vagifzeynalov I like the idea though so I went ahead and created JAVA-2042 so that we don't forget it when the new mapper API will arrive. Thanks for your time and effort!

vagifzeynalov:
@adutra no problem. Looking forward for 4.0 release!

PR #975, opened 2018-03-11
fix example to match `TypeCodec`s documentation

alexott:
- for `parse`:  Null values and empty Strings should be accepted, as well as the string
  `NULL`; in most cases, implementations should interpret these inputs has equivalent to
  a null reference.
- for `format`: Null values should be accepted; in most cases, implementations should
  return the CQL keyword `NULL` for null inputs.

### DIFF
--- a/manual/custom_codecs/README.md
+++ b/manual/custom_codecs/README.md
@@ -296,12 +296,12 @@ public class AddressCodec extends TypeCodec<Address> {
 
     @Override
     public Address parse(String value) throws InvalidTypeException {
-        return value == null || value.isEmpty() || value.equals(NULL) ? null : toAddress(innerCodec.parse(value));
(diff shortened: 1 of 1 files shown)


manual/custom_codecs/README.md
```diff
@@ -296,12 +296,12 @@ public class AddressCodec extends TypeCodec<Address> {
 
     @Override
     public Address parse(String value) throws InvalidTypeException {
-        return value == null || value.isEmpty() || value.equals(NULL) ? null : toAddress(innerCodec.parse(value));
+        return value == null || value.isEmpty() || value.equalsIgnoreCase("NULL") ? null : toAddress(innerCodec.parse(value));
```
olim7t: Nit: could you split that into 3 lines (before `?` and `:`)?
With the default Github markdown rendering, the code overflows on the right.

olim7t (approved):

PR #978, opened 2018-03-19
 JAVA-708: Query request size information

adutra:
TODO list:
- [ ] Add unit tests
- [ ] Add integration tests

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Connection.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Connection.java
@@ -142,7 +142,11 @@ class Connection {
                     new Initializer(this, protocolVersion, protocolOptions.getCompression().compressor(), protocolOptions.getSSLOptions(),
                             factory.configuration.getPoolingOptions().getHeartbeatIntervalSeconds(),
                             factory.configuration.getNettyOptions(),
-                            factory.configuration.getCodecRegistry()));
+                            factory.configuration.getCodecRegistry(),
+                            factory.configuration.getMetricsOptions().isEnabled() ?
+                                    factory.manager.metrics :
+                                    null
+                    ));
 
             ChannelFuture future = bootstrap.connect(address);
 
@@ -1419,8 +1423,9 @@ class Connection {
         private final NettyOptions nettyOptions;
         private final ChannelHandler idleStateHandler;
         private final CodecRegistry codecRegistry;
+        private final Metrics metrics;
 
-        Initializer(Connection connection, ProtocolVersion protocolVersion, FrameCompressor compressor, SSLOptions sslOptions, int heartBeatIntervalSeconds, NettyOptions nettyOptions, CodecRegistry codecRegistry) {
+        Initializer(Connection connection, ProtocolVersion protocolVersion, FrameCompressor compressor, SSLOptions sslOptions, int heartBeatIntervalSeconds, NettyOptions nettyOptions, CodecRegistry codecRegistry, Metrics metrics) {
             this.connection = connection;
             this.protocolVersion = protocolVersion;
             this.compressor = compressor;
@@ -1428,6 +1433,7 @@ class Connection {
             this.nettyOptions = nettyOptions;
             this.codecRegistry = codecRegistry;
             this.idleStateHandler = new IdleStateHandler(heartBeatIntervalSeconds, 0, 0);
+            this.metrics = metrics;
         }
 
         @Override
@@ -1450,6 +1456,11 @@ class Connection {
             }
 
             // pipeline.addLast("debug", new LoggingHandler(LogLevel.INFO));
+
+            if (metrics != null) {
+                pipeline.addLast("inboundTrafficMeter", new InboundTrafficMeter(metrics.getBytesReceived()));
+                pipeline.addLast("outboundTrafficMeter", new OutboundTrafficMeter(metrics.getBytesSent()));
+            }
 
             pipeline.addLast("frameDecoder", new Frame.Decoder());
             pipeline.addLast("frameEncoder", frameEncoder);
--- a/driver-core/src/main/java/com/datastax/driver/core/InboundTrafficMeter.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/InboundTrafficMeter.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.driver.core;
+
+import com.codahale.metrics.Meter;
+import io.netty.buffer.ByteBuf;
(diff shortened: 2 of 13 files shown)


driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
```diff
@@ -189,6 +192,52 @@ public boolean hasValues() {
         return hasValues(CodecRegistry.DEFAULT_INSTANCE);
     }
 
+    @Override
+    public int requestSizeInBytes(ProtocolVersion protocolVersion, CodecRegistry codecRegistry) {
+        int size = Header.lengthFor(protocolVersion);
+        try {
```
olim7t, line 198: I think the query string is missing:
```java
size += CBUtil.sizeOfLongString(getQueryString(codecRegistry));
```

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
```diff
@@ -189,6 +192,52 @@ public boolean hasValues() {
...
+                    size += QueryFlag.serializedSize(protocolVersion);
+                    if (hasValues()) {
+                        if (usesNamedValues()) {
+                            size += CBUtil.sizeOfNamedValueList(getNamedValues(protocolVersion, codecRegistry));
+                        } else {
+                            size += CBUtil.sizeOfValueList(Arrays.asList(getValues(protocolVersion, codecRegistry)));
+                        }
+                    }
+                    if (getFetchSize() > 0) {
+                        size += 4;
```
olim7t: `fetchSize` and `serialConsistencyLevel` also depend on the session-wide defaults. This is annoying because we don't have a reference to the session here, and it would be overkill to inject it as a parameter.

I think it's fine if the method is a few bytes off (we should document that), and I'd rather overestimate than underestimate (currently the method underestimates if the session defaults are not the protocol defaults). So I am in favor of always counting those two parameters.

driver-core/src/main/java/com/datastax/driver/core/Statement.java
```diff
@@ -584,6 +584,17 @@ public Statement setOutgoingPayload(Map<String, ByteBuffer> payload) {
...
     }
 
+    /**
+     * Returns the number of bytes required to encode this statement.
+     * <p/>
+     * If the size cannot be accurately calculated, this method returns -1.
+     *
+     * @return the number of bytes required to encode this statement.
+     */
+    public int requestSizeInBytes(ProtocolVersion protocolVersion, CodecRegistry codecRegistry) {
```
olim7t: 👍 for not naming that `get`, to make it clear that it is recomputed every time. We should also mention it in the javadoc.

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
```diff
@@ -189,6 +192,52 @@ public boolean hasValues() {
...
+                    }
+                    if (getPagingState() != null) {
+                        size += CBUtil.sizeOfValue(getPagingState());
+                    }
+                    if (getSerialConsistencyLevel() != null) {
+                        size += CBUtil.sizeOfConsistencyLevel(getSerialConsistencyLevel());
+                    }
+                    if (getDefaultTimestamp() > Long.MIN_VALUE) {
+                        size += 8;
```
olim7t: Nit: the default timestamp is only sent for v3 and above

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -1451,6 +1457,11 @@ protected void initChannel(SocketChannel channel) throws Exception {
 
             // pipeline.addLast("debug", new LoggingHandler(LogLevel.INFO));
 
+            if (metrics != null) {
+                pipeline.addLast("inboundTrafficMeter", new InboundTrafficMeter(metrics.getBytesReceived()));
+                pipeline.addLast("outboundTrafficMeter", new OutboundTrafficMeter(metrics.getBytesSent()));
+            }
+
```
olim7t, line 1464: Open question: should we include SSL traffic in those metrics?

On the one hand, we're interested about the impact of our data model and queries on the traffic, so SSL is a bit off topic. On the other, I can see how some people would expect this to be the overall traffic.

driver-core/src/main/java/com/datastax/driver/core/InboundTrafficMeter.java
```diff
@@ -0,0 +1,41 @@
...
+        this.meter = meter;
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        if (msg instanceof ByteBuf) {
+            meter.mark(((ByteBuf) msg).readableBytes());
+        }
+        super.channelRead(ctx, msg);
+    }
```
olim7t, line 39: Nice, straightforward way to implement this 💯 

I was wondering if we could use a single `ChannelDuplexHandler`, but the two methods don't share any state so it doesn't really make any difference.

olim7t (changes requested):
A few minor comments but overall this looks really great 👍 
I'll work on the tests to keep it moving.

driver-core/src/main/java/com/datastax/driver/core/Statement.java
olim7t: Mmm actually I read "request" as a verb, but you probably meant "the request".

driver-core/src/test/java/com/datastax/driver/core/StatementSizeTest.java
```diff
@@ -0,0 +1,200 @@
...
+        assertThat(v5SizeOf(statementWithNamedValue))
+                .isEqualTo(expectedSize
+                        + 2 // size of number of values
+                        + (2 + "key".getBytes(Charsets.UTF_8).length) // key
+                        + (4 + "local".getBytes(Charsets.UTF_8).length) // value
+                );
+
+        statement.setFetchSize(10);
+        expectedSize += 4;
+        assertThat(v5SizeOf(statement)).isEqualTo(expectedSize);
```
olim7t: To adapt if we decide to overestimate for session-wide defaults.

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
adutra: It is included in the line below (199).

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
adutra: In this case the rule also applies to `defaultTimestamp` in V3+.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
adutra: According to [this SO question](https://stackoverflow.com/questions/548029/how-much-overhead-does-ssl-impose) the overhead in bytes sent/received of SSL vs non SSL is basically confined to the sole handshake phase (of course there is a lot of CPU overhead, but that's irrelevant here). Since these meters are never reset, I'd say that this overhead would tend to zero over time, so it doesn't really matter much.

driver-core/src/main/java/com/datastax/driver/core/RegularStatement.java
olim7t: Indeed. They only apply per-message, not inside of the statements in a batch, so we'll overestimate at most 12 bytes.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
olim7t: ok

olim7t (approved):

olim7t:
lgtm 👍 
Do you want to keep the two commits separate?

adutra:
> Do you want to keep the two commits separate?

I think it makes sense to keep them separate here.

PR #982, opened 2018-03-23
JAVA-1789: Account for flags in Prepare encodedSize

tolbertam:
For [JAVA-1789](https://datastax-oss.atlassian.net/browse/JAVA-1789)

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Requests.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Requests.java
@@ -591,7 +591,12 @@ class Requests {
 
             @Override
             public int encodedSize(Prepare msg, ProtocolVersion version) {
-                return CBUtil.sizeOfLongString(msg.query);
+                int size = CBUtil.sizeOfLongString(msg.query);
+
+                if (version.compareTo(ProtocolVersion.V5) >= 0) {
+                    size += 4; //flags
+                }
+                return size;
             }
         };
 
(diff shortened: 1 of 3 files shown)


driver-core/src/main/java/com/datastax/driver/core/Requests.java
```diff
@@ -591,7 +591,12 @@ public void encode(Prepare msg, ByteBuf dest, ProtocolVersion version) {
 
             @Override
             public int encodedSize(Prepare msg, ProtocolVersion version) {
-                return CBUtil.sizeOfLongString(msg.query);
+                int size = CBUtil.sizeOfLongString(msg.query);
+
+                if (version.compareTo(ProtocolVersion.V5) >= 0) {
+                    size += 4; //flags
+                }
```
olim7t, line 598: It would be better to use `PREPARED_METADATA_CHANGES.isSupportedBy(version)` for this test.

olim7t (changes requested):

driver-core/src/main/java/com/datastax/driver/core/Requests.java
tolbertam: was following the logic in `encode` which checks the protocol version, should we change that to?

olim7t: Mmm actually the flag does not relate directly to this feature, but instead to per-query keyspaces, which are not supported at all in the driver at this point. Keep the version test, my bad.

olim7t (approved):

PR #983, opened 2018-03-29
JAVA-1759: Revisit metrics API

olim7t:
- allow excluding Dropwizard completely if metrics are disabled
- expose methods to access individual metrics programatically (this was feedback from @adutra that I forgot to address in the original PR)
- open the door to switch metrics framework

TODO:
- [x] add concurrency annotations

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/metrics/Metrics.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/metrics/Metrics.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core.metrics;
+
+import com.codahale.metrics.Metric;
+import com.codahale.metrics.MetricRegistry;
+import com.datastax.oss.driver.api.core.metadata.Node;
+
+/**
+ * A wrapper around a {@link MetricRegistry} to expose the driver's metrics.
+ *
+ * <p>This type exists mainly to avoid a hard dependency to Dropwizard Metrics (that is, the JAR can
+ * be completely removed from the classpath if metrics are disabled). It also provides convenience
+ * methods to access individual metrics programatically.
+ */
+public interface Metrics {
+
+  /**
+   * Returns the underlying Dropwizard registry.
+   *
+   * <p>Typically, this can be used to configure a reporter.
+   *
+   * @see <a href="http://metrics.dropwizard.io/4.0.0/manual/core.html#reporters">Reporters
+   *     (Dropwizard Metrics manual)</a>
+   */
+  MetricRegistry getRegistry();
+
+  /**
+   * Retrieves a session-level metric from the registry.
+   *
+   * <p>To determine the type of each metric, refer to the comments in the default {@code
+   * reference.conf} (included in the driver's codebase and JAR file). Note that the method does not
+   * check that this type is correct (there is no way to do this at runtime because some metrics are
+   * generic); if you use the wrong type, you will get a {@code ClassCastException} in your code:
+   *
+   * <pre>{@code
+   * // Correct:
+   * Gauge<Integer> connectedNodes = getNodeMetric(node, DefaultSessionMetric.CONNECTED_NODES);
+   *
+   * // Wrong, will throw CCE:
+   * Counter connectedNodes = getNodeMetric(node, DefaultSessionMetric.CONNECTED_NODES);
+   * }</pre>
+   *
+   * @return the metric, or {@code null} if it is disabled.
+   */
+  @SuppressWarnings("TypeParameterUnusedInFormals")
+  <T extends Metric> T getSessionMetric(SessionMetric metric);
+
+  /**
+   * Retrieves a node-level metric for a given node from the registry.
+   *
+   * <p>To determine the type of each metric, refer to the comments in the default {@code
+   * reference.conf} (included in the driver's codebase and JAR file). Note that the method does not
+   * check that this type is correct (there is no way to do this at runtime because some metrics are
+   * generic); if you use the wrong type, you will get a {@code ClassCastException} in your code:
+   *
+   * <pre>{@code
+   * // Correct:
+   * Gauge<Integer> openConnections = getNodeMetric(node, DefaultNodeMetric.OPEN_CONNECTIONS);
+   *
+   * // Wrong, will throw CCE:
+   * Counter openConnections = getNodeMetric(node, DefaultNodeMetric.OPEN_CONNECTIONS);
+   * }</pre>
+   *
+   * @return the metric, or {@code null} if it is disabled.
+   */
+  @SuppressWarnings("TypeParameterUnusedInFormals")
+  <T extends Metric> T getNodeMetric(Node node, NodeMetric metric);
+}
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricUpdater.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricUpdater.java
@@ -15,6 +15,7 @@
  */
 package com.datastax.oss.driver.internal.core.metrics;
 
+import com.codahale.metrics.Metric;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import com.datastax.oss.driver.api.core.config.DefaultDriverOption;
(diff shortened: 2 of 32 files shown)


integration-tests/src/test/java/com/datastax/oss/driver/api/core/metrics/MetricsIT.java
```diff
@@ -40,37 +39,30 @@ public void should_expose_metrics() {
...
-        if (entry.getKey().endsWith(DefaultSessionMetric.CQL_REQUESTS.getPath())) {
-          requestsTimer = entry.getValue();
-        }
-      }
-      assertThat(requestsTimer).isNotNull();
+      assertThat(session.getMetrics())
+          .hasValueSatisfying(
+              metrics -> {
+                Timer requestsTimer =
+                    metrics.getSessionMetric(DefaultSessionMetric.CQL_REQUESTS, Timer.class);
```
olim7t: This is indeed much nicer than the previous code.

core/src/main/java/com/datastax/oss/driver/api/core/metrics/Metrics.java
```diff
@@ -0,0 +1,79 @@
...
+
+  /**
+   * Returns the underlying Dropwizard registry.
+   *
+   * <p>Typically, this can be used to configure a reporter.
+   *
+   * @see <a href="http://metrics.dropwizard.io/4.0.0/manual/core.html#reporters">Reporters
+   *     (Dropwizard Metrics manual)</a>
+   */
+  MetricRegistry getRegistry();
```
olim7t, line 39: This references Dropwizard classes. This means that if someone were to integrate another metrics framework, they would have to make `session.getMetrics` return `Optional.empty()`,  and find another way to expose that framework's equivalent of the registry.
I don't want to go further than that, because any other solution would have to manifest at a higher level in the API (e.g. `CqlSession<DropwizardMetrics>`, `CqlSession<MicrometerMetrics>`, etc). Ultimately I think switching metrics framework is a very rare edge case, so I'd rather keep the API simpler for the 99.9% regular ones.

tolbertam: I think that in the general case, if you were plugging in a metrics framework, at some level that integration would provide access to the equivalent of a registry, so I think what you are doing here is fine.  Most users will be content with dropwizard metrics (I'd guess) and direct access to the registry here seems good.

tolbertam (approved):
lgtm 👍

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricUpdater.java
```diff
@@ -15,62 +15,69 @@
...
   @Override
   public void updateTimer(MetricT metric, long duration, TimeUnit unit) {
     if (enabledMetrics.contains(metric)) {
-      metricRegistry.timer(buildFullName(metric)).update(duration, unit);
+      registry.timer(buildFullName(metric)).update(duration, unit);
     }
   }
 
+  @SuppressWarnings("unchecked")
+  public <T extends Metric> T getMetric(MetricT metric, GenericType<T> expectedType) {
```
adutra: `expectedType` is unused, shouldn't we at least check that the returned metric is an instance of it?

core/src/main/java/com/datastax/oss/driver/api/core/metrics/Metrics.java
adutra: There might be a more generic solution.

This interface could stay agnostic of Dropwizard metrics:

1. Remove the `getRegistry()` method (it is not used by the driver anyway)
2. instead of `<T extends Metric>`, use simply `<T>`

Granted, people would have to cast to `DefaultMetrics` (maybe renamed to `DropwizardMetrics`?) to access the registry, but that looks cleaner to me: if someone wants to change for another metrics framework, it's just a matter of implementing another `MetricsFactory` (unless I'm missing something).

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/MetricsFactory.java
```diff
@@ -16,8 +16,12 @@
 package com.datastax.oss.driver.internal.core.metrics;
 
 import com.datastax.oss.driver.api.core.metadata.Node;
+import com.datastax.oss.driver.api.core.metrics.Metrics;
+import java.util.Optional;
 
-public interface MetricUpdaterFactory {
+public interface MetricsFactory {
```
adutra, line 22: Slightly better imo:

```
public interface MetricsFactory<T extends Metrics> {

  Optional<? extends T> getMetrics();
```

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/api/core/metrics/Metrics.java
olim7t: > people would have to cast to DefaultMetrics

That's what I meant by "would have to manifest at a higher level in the API". Requiring a cast for a default feature is really bad IMHO. I'd rather not punish the 99.9%, and instead cause a minor inconvenience to the 0.1% that might want to change the metrics framework. Frankly the fact that this is even possible is a cherry on top of the cake already.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricUpdater.java
olim7t: Guava's type token doesn't have any method to do that. Understandably so, because it could only compare to the erased type of the object.
Which leads me to believe that `expectedType` is redundant here; I put it thinking of the `Class<T>`-based type token pattern, but really it doesn't help at all in this case, since we're going to suppress a warning anyway.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/MetricsFactory.java
olim7t: Based on my reply to your other comment, I think we should avoid specialized `Metrics` subinterfaces, so this wouldn't apply.

core/src/main/java/com/datastax/oss/driver/api/core/metrics/Metrics.java
adutra: I slightly disagree. A cast is still better imo, and I still like the idea that there is no "default" metric library in the driver, it just happens that we ship with one impl based on Dropwizard. But that's a detail and I don't want to spend more time on it, I'm fine if we leave things as they are.

core/src/main/java/com/datastax/oss/driver/api/core/metrics/Metrics.java
```diff
@@ -42,38 +41,22 @@
...
-  <T extends Metric> T getNodeMetric(Node node, NodeMetric metric, GenericType<T> expectedType);
-
-  /**
-   * Shortcut for {@link #getNodeMetric(Node, NodeMetric, GenericType) getNodeMetric(node, metric,
-   * GenericType.of(expectedClass))}.
-   */
-  default <T extends Metric> T getNodeMetric(Node node, NodeMetric metric, Class<T> expectedClass) {
-    return getNodeMetric(node, metric, GenericType.of(expectedClass));
-  }
+  <T extends Metric> T getNodeMetric(Node node, NodeMetric metric);
```
adutra: That's worse imo, this `T` is not bounded by anything. Granted we can only check against the erased type here so `GenericType` is not an option, but why not check against a `Class<T>`? To me the ideal signature would be:

```
<T extends Metric> T getNodeMetric(Node node, NodeMetric metric, Class<T> expectedType);
```

And the implementation would enforce that the returned object is an instance of `expectedType`.

core/src/main/java/com/datastax/oss/driver/api/core/metrics/Metrics.java
```diff
@@ -42,38 +41,22 @@
...
    * Retrieves a node-level metric for a given node from the registry.
    *
-   * @param expectedType the type of the metric (refer to the comments in the default {@code
-   *     reference.conf} included in the driver's codebase or JAR file). If it is not parameterized,
-   *     you can use {@link #getSessionMetric(SessionMetric, Class)} as a shortcut.
+   * <p>To determine the type of each metric, refer to the comments in the default {@code
+   * reference.conf} (included in the driver's codebase and JAR file).
+   *
    * @return the metric, or {@code null} if it is disabled.
    * @throws ClassCastException if the type does not match.
```
adutra: With the current signature the `ClassCastException` will not be thrown from within the method, but in client code (when assigning to a variable of the wrong type), so technically this `@throws` clause should be removed.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/MetricsFactory.java
adutra: Sorry, I don't get it. I'm just saying that `Optional<? extends Metrics> getMetrics()` is more correct than `Optional<Metrics> getMetrics()`.

core/src/main/java/com/datastax/oss/driver/api/core/metrics/Metrics.java
olim7t: Mmm, but then it doesn't work for generic metrics:

```java
// unchecked assignment
Gauge<Integer> nodeMetric = getNodeMetric(node, DefaultNodeMetric.OPEN_CONNECTIONS, Gauge.class);
```

Forcing the client to suppress a warning is not better than not enforcing T with a type token argument. Actually I see it as two sides of the same problem: that Java's type system is not powerful enough to enforce generic types at runtime. There's nothing we can do about it, so I'd rather go with the solution that doesn't generate a warning.
(to put it another way, we're moving the suppression of the warning into our code)

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/MetricsFactory.java
olim7t: The only goal of this method is to be exposed as `Session.getMetrics()`. In what scenario would `? extends Metric` be more useful for a client application?

core/src/main/java/com/datastax/oss/driver/api/core/metrics/Metrics.java
adutra: Indeed,  my suggestion wouldn't work well will Gauges. Sounds good.

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/MetricsFactory.java
adutra: If someone extends `Session` and `Metrics`, then if this method has the return type `Optional<? extends Metrics>`  it would be possible to use covariance and override it with a more specialized `Optional<MySpecialMetrics>`. That is not possible if the original return type is `Optional<Metrics>`. We had a similar problem with `AsyncResultSet.fetchNextPage()` recently.

olim7t: OK that makes sense, I will change it.

PR #984, opened 2018-03-30
JAVA-1787: Provide a standalone shaded Guava artifact

olim7t:
~~Requires https://github.com/datastax/java-driver-shaded-guava in the local repository~~ => now published in Central.

I didn't change the Guava session example in the integration tests, I felt it would have been weird to use the shaded classes there.


tolbertam (approved):
Looks great!

PR #985, opened 2018-03-30
JAVA-1797: Use jnr-ffi version required by jnr-posix

tolbertam:
For [JAVA-1797](https://datastax-oss.atlassian.net/browse/JAVA-1797).

### DIFF
--- a/driver-core/pom.xml
+++ b/driver-core/pom.xml
@@ -56,11 +56,6 @@
 
         <dependency>
             <groupId>com.github.jnr</groupId>
-            <artifactId>jnr-ffi</artifactId>
-        </dependency>
-
-        <dependency>
-            <groupId>com.github.jnr</groupId>
             <artifactId>jnr-posix</artifactId>
         </dependency>
 
--- a/driver-core/pom.xml
+++ b/driver-core/pom.xml
@@ -56,6 +56,11 @@
 
         <dependency>
             <groupId>com.github.jnr</groupId>
+            <artifactId>jnr-ffi</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.github.jnr</groupId>
             <artifactId>jnr-posix</artifactId>
         </dependency>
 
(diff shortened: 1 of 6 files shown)


tolbertam:
We may also want to consider having `UUIDs.getProcessPiece()` skip using `Native.processId()` if it returns 0, wdyt?

driver-core/pom.xml
```diff
@@ -54,11 +54,6 @@
             <artifactId>slf4j-api</artifactId>
         </dependency>
 
-        <dependency>
```
adutra: When a project uses some dependency directly, it is usually recommended that it declares that dependency explicitly instead of relying on transitive dependencies. I would prefer that we keep the dependency declared here.

driver-core/src/main/java/com/datastax/driver/core/utils/UUIDs.java
```diff
@@ -153,7 +153,12 @@ private static String getProcessPiece() {
         if (pid == null && Native.isGetpidAvailable()) {
             try {
                 pid = Native.processId();
-                LOGGER.info("PID obtained through native call to getpid(): {}", pid);
+                if (pid == 0) {
```
adutra, line 156: 👍

pom.xml
```diff
@@ -63,7 +63,6 @@
         <joda.version>2.9.9</joda.version>
         <jsr353-api.version>1.0</jsr353-api.version>
         <jsr353-ri.version>1.0.4</jsr353-ri.version>
-        <jnr-ffi.version>2.0.9</jnr-ffi.version>
```
adutra, line 66: I would suggest that we declare both dependencies for the reasons outlined above, with maybe a comment here to remind us that jnr-posix and jnr-ffi must have compatible versions.

adutra (changes requested):

driver-core/pom.xml
tolbertam: I considered that, but I was sort of put off by how jnr-posix upgraded minor version dependency of jnr-ffi in a patch release.  Since we now have a test to verify the functionality of jnr-posix retrieving pids, we could add the direct dependency back and feel relatively safe, so i'll do that.

adutra (approved):

olim7t (approved):

PR #987, opened 2018-03-30
JAVA-1776: Use concurrency annotations

olim7t:
I documented almost everything except utility classes.
There are a lot of changes in this PR, so pointing out the ones that are not just adding the annotation in comments below.


core/pom.xml
```diff
@@ -83,6 +83,11 @@
       <groupId>org.hdrhistogram</groupId>
       <artifactId>HdrHistogram</artifactId>
     </dependency>
+    <dependency>
+      <groupId>net.jcip</groupId>
+      <artifactId>jcip-annotations</artifactId>
+      <optional>true</optional>
+    </dependency>
```
olim7t, line 90: I went with this artifact for simplicity, it's the only one that has everything and doesn't pull a bunch of other stuff.

Optional so that it's not pulled in client applications. As far as I can tell, the only consequence is that the annotations appear in error if you browse the sources from the dependent project (this happens with Guava too).

pom.xml
```diff
@@ -238,6 +243,8 @@
           <target>1.8</target>
           <compilerArgs combine.children="override">
             <compilerArg>-Xep:FutureReturnValueIgnored:OFF</compilerArg>
+            <!-- Too limited, does not detect locks acquired by a calling method: -->
+            <compilerArg>-Xep:GuardedBy:OFF</compilerArg>
```
olim7t, line 248: For example, `RateLimitingRequestThrottler` would cause errors in `acquire` (even though we assert that the lock is held at the beginning of the method).

tolbertam (approved):
Looks good 👍

pom.xml
adutra: Can't we instead reacquire the lock in these cases? According to the javadocs of `ReentrantLock`, if the thread already has the lock, acquiring it again is a no-op, and `isHeldByCurrentThread()` should be used for debugging purposes only.

olim7t: Reacquiring the lock is not the same as expecting it to be held. The javadocs explicitly mention assertions in callees as a typical use case.

PR #988, opened 2018-04-04
JAVA-1778: Port more documentation from 3.x

olim7t:


manual/core/retries/README.md
```diff
@@ -0,0 +1,153 @@
...
+client with a `READ_TIMEOUT` error.
+
+This could be due to a temporary overloading of these replicas, or even that they just failed or
+were turned off. During reads, Cassandra doesn't request data from every replica to minimize
+internal network traffic; instead, some replicas are only asked for a checksum of the data. A read
+timeout may occur even if enough replicas responded to fulfill the consistency level, but only
+checksum responses were received (the method's `dataPresent` parameter allow you to check if you're
+in that situation).
+
+If the policy rethrows the error, the user code will get a [ReadTimeoutException] (do not confuse
```
tolbertam: This doesn't seem to be a problem with github markdown, but our tooling that generates documentation interprets `(do not confuse...)` as a link so it doesn't show up.

faq/README.md
```diff
@@ -26,4 +26,23 @@ Also, `CompletionStage` does not expose a `get()` method; one can view that as a
 use a fully asynchronous programming model (chaining callbacks instead of blocking for a result).
 
 At any rate, `CompletionStage` has a `toCompletableFuture()` method. In current JDK versions, every
-`CompletionStage` is a `CompletableFuture`, so the conversion has no performance overhead.
\ No newline at end of file
+`CompletionStage` is a `CompletableFuture`, so the conversion has no performance overhead.
+
+### Where is `DowngradingConsistencyRetryPolicy`?
```
tolbertam, line 31: Well stated 👍

manual/core/load_balancing/README.md
```diff
@@ -0,0 +1,218 @@
...
+}
+```
+
+### Concepts
+
+#### Node distance
+
+For each node, the policy computes a *distance* that determines how connections will be established:
+
+* `LOCAL` and `REMOTE` are "active" distances, meaning that the driver will keep open connections to
```
tolbertam, line 23: Was just thinking, is `REMOTE` relevant any more now that we don't have a policy that allows connection to remote DCs?   Is connecting to 'remote' DCs considered a viable use case?

manual/core/load_balancing/README.md
```diff
@@ -0,0 +1,218 @@
...
+that we consider the best choice for most cases. You can still write a
+[custom implementation](#custom-implementation) if you have special requirements.
+
+#### Local only
+
+The default policy **only connects to a single datacenter**. The rationale is that a typical
+multi-region deployment will collocate one or more application instances with each Cassandra
+datacenter:
+
+```ditaa
```
tolbertam, line 61: Awesome diagram, looks great when generated.

manual/core/load_balancing/README.md
```diff
@@ -0,0 +1,218 @@
...
+  local-datacenter = datacenter1
+}
+```
+
+This option is required, except when you didn't specify any contact points and let the driver
+default to 127.0.0.1:9042 (this is mostly for convenience during the development phase).
+
+#### Token-aware
+
+The default policy is **token-aware** by default: requests will be routed in priority to the
```
tolbertam, line 110: I think we should make this configurable, with it being enabled by default.

manual/core/load_balancing/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+// Set the keyspace manually (skip this if using a per-query keyspace):
+statement.setRoutingKeyspace("testKs");
+
+// Set the routing key manually: serialize each partition key component to its target CQL type
+statement.setRoutingKey(
+    TypeCodecs.INT.encodePrimitive(1, session.getContext().protocolVersion()),
+    TypeCodecs.INT.encodePrimitive(2016, session.getContext().protocolVersion()));
+
+session.execute(statement);
```
tolbertam: since `Statement` is immutable, shouldn't we be reassigning it in the invocations to `setRoutingKeyspace` and `setRoutingKey`?

manual/core/pooling/README.md
```diff
@@ -0,0 +1,150 @@
...
+
+In particular, it's a good idea to keep an eye on those two metrics:
+
+* `pool.open-connections`: if this doesn't match your configured pool size, something is preventing
+  connections from opening (either configuration or network issues, or a server-side limitation --
+  see [CASSANDRA-8086]);
+* `pool.available-streams`: if this is often close to 0, it's a sign that the pool is getting
+  saturated. Maybe `max-requests-per-connection` is too low, or more connections should be added.
+
+### Tuning
```
tolbertam, line 123: Nice explanation here 👍

manual/core/query_timestamps/README.md
```diff
@@ -0,0 +1,137 @@
...
+the future. This happens if timestamps are generated at a rate of more than one per microsecond, or
+more likely in the event of a system clock skew. When this happens, the generator logs a warning
+message in the category `com.datastax.oss.driver.api.core.time.MonotonicTimestampGenerator`:
+
+```
+Clock skew detected: current tick (...) was ... microseconds behind the last generated timestamp (...),
+returned timestamps will be artificially incremented to guarantee monotonicity.
+```
+
+You can control that message with those options:
```
tolbertam: 'with those' > 'with these'

manual/core/query_timestamps/README.md
```diff
@@ -0,0 +1,137 @@
...
+    # How far in the future timestamps are allowed to drift before the warning is logged.
+    # If it is undefined or set to 0, warnings are disabled.
+    threshold = 1 second
+    # How often the warning will be logged if timestamps keep drifting above the threshold.
+    interval = 10 seconds
+  }
+}
+```
+
+The generator strives to achieve microsecond resolution on a best-effort basis. But in practice, the
```
tolbertam: 'The generator' -> 'This generator'

manual/core/query_timestamps/README.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+### Timestamp generator
+
+The driver has a timestamp generator that gets invoked for every outgoing request; it either assigns
+a client-side timestamp to the request, or indicates that the server should assign it. 
+
+The timestamp generator is defined in the [configuration](../configuration/).
+
+#### `AtomicTimestampGenerator`
```
tolbertam: In our generated docs, sections headings using backticks look a bit awkward, it might be better to not use them (although opinion on that is subjective).

manual/core/query_timestamps/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+To check what the driver is currently using, turn on `INFO` logs for the category
+`com.datastax.oss.driver.internal.core.time`, and look for one of the following messages at
+initialization:
+
+* `Using Java system clock because this was explicitly required in the configuration`
+* `Could not access native clock (see debug logs for details), falling back to Java system clock`
+* `Using native clock for microsecond precision`
+
+#### `ThreadLocalTimestampGenerator`
```
tolbertam: I wonder how useful `ThreadLocalTimestampGenerator` is.  Do we expect it to be that much of a performance improvement?  I can see it doing more harm than good, so maybe it's worth getting rid of but I don't feel too strongly about it.   Also we should make it clear which thread the timestamp is generated on the calling thread.

manual/core/query_timestamps/README.md
```diff
@@ -0,0 +1,137 @@
+## Query timestamps
```
tolbertam, line 1: One 'gothca' with Client timestamps that most aren't aware of is that client timestamps are not used by Lightweight transactions.  We should indicate that here.

manual/core/retries/README.md
```diff
@@ -0,0 +1,153 @@
...
+### onWriteTimeout
+
+This is similar to `onReadTimeout`, but for write operations. The reason reads and writes are
+handled separately is because a read is obviously a non mutating operation, whereas a write is
+likely to be. If a write times out at the coordinator level, there is no way to know whether the
+mutation was applied or not on the non-answering replica.
+
+If the policy rethrows the error, the user code will get a [WriteTimeoutException].
+
+This method is only invoked for [idempotent](../idempotence/) statements. Otherwise, the driver
```
tolbertam, line 74: We should also port the idempotence docs

manual/core/speculative_execution/README.md
```diff
@@ -0,0 +1,209 @@
...
+high rate.
+
+The best way to monitor this is to compare the `pool.orphaned-streams` [metric](../metrics/) to the
+total number of available stream ids (which can be computed from the configuration:
+`pool.local.size * max-requests-per-connection`). The `pool.available-streams` and `pool.in-flight`
+metrics will also give you an idea of how many stream ids are left for active queries.
+
+#### Request ordering
+
+Note: ordering issues are only a problem with [server-side timestamps](../query_timestamps/), which
```
tolbertam, line 183: Technically client timestamps have been the default since 2.1 (I think)

tolbertam (commented):
Looking great!  The documentation for 4.x is really starting to round out nicely!  Had a few comments, mostly small nits.

manual/core/speculative_execution/README.md
tolbertam: It's 3.0 actually.

manual/core/retries/README.md
olim7t: They're already in #972 because I also needed them there.

tolbertam: Ah, didn't catch that, nice!

manual/core/retries/README.md
olim7t: Ah yeah this gets me every time, the opening parenthesis must be escaped with a backslash.

faq/README.md
olim7t: Shamelessly stolen from the 3.x upgrade guide 😄

manual/core/load_balancing/README.md
olim7t: That's right, the driver doesn't use `REMOTE` out of the box anymore. But we still allow custom policies, so I feel that removing it completely would be too much.

manual/core/load_balancing/README.md
olim7t: Good catch! Ironically I wrote a FAQ entry about that.

manual/core/query_timestamps/README.md
olim7t: I agree, I used plain text in the retries page, I'll do the  same here.

manual/core/speculative_execution/README.md
olim7t: Client timestamps are available in all the protocol versions supported by driver 4 (v3 and above).

tolbertam: To clarify, my comment was more to the statement of `which are not the default anymore in driver 4+`.  This is technically true for driver 3+ (if using protocol v3+) right?  Probably better to leave it as is (4+) to avoid confusing the user :D

tolbertam (approved):
Changes look great 👍

PR #989, opened 2018-04-07
JAVA-1523: Add query logger

olim7t:
TODO:
- [x] close the `RequestTracker` when the session shuts down
- [x] add concurrency annotations

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
@@ -23,6 +23,7 @@ import com.datastax.oss.driver.api.core.config.DriverConfigProfile;
 import com.datastax.oss.driver.api.core.context.DriverContext;
 import com.datastax.oss.driver.api.core.metadata.NodeStateListener;
 import com.datastax.oss.driver.api.core.metadata.schema.SchemaChangeListener;
+import com.datastax.oss.driver.api.core.tracker.RequestTracker;
 import com.datastax.oss.driver.api.core.type.codec.TypeCodec;
 import com.datastax.oss.driver.internal.core.ContactPoints;
 import com.datastax.oss.driver.internal.core.config.typesafe.DefaultDriverConfigLoader;
@@ -59,6 +60,7 @@ public abstract class SessionBuilder<SelfT extends SessionBuilder, SessionT> {
   protected List<TypeCodec<?>> typeCodecs = new ArrayList<>();
   private NodeStateListener nodeStateListener;
   private SchemaChangeListener schemaChangeListener;
+  protected RequestTracker requestTracker;
   protected CqlIdentifier keyspace;
 
   /**
@@ -158,6 +160,17 @@ public abstract class SessionBuilder<SelfT extends SessionBuilder, SessionT> {
   }
 
   /**
+   * Register a request tracker to use with the session.
+   *
+   * <p>If the tracker is specified programmatically with this method, it overrides the
+   * configuration (that is, the {@code request.tracker.class} option will be ignored).
+   */
+  public SelfT withRequestTracker(RequestTracker requestTracker) {
+    this.requestTracker = requestTracker;
+    return self;
+  }
+
+  /**
    * Sets the keyspace to connect the session to.
    *
    * <p>Note that this can also be provided by the configuration; if both are defined, this method
@@ -216,7 +229,8 @@ public abstract class SessionBuilder<SelfT extends SessionBuilder, SessionT> {
 
     return DefaultSession.init(
         (InternalDriverContext)
-            buildContext(configLoader, typeCodecs, nodeStateListener, schemaChangeListener),
+            buildContext(
+                configLoader, typeCodecs, nodeStateListener, schemaChangeListener, requestTracker),
         contactPoints,
         keyspace);
   }
@@ -229,9 +243,10 @@ public abstract class SessionBuilder<SelfT extends SessionBuilder, SessionT> {
       DriverConfigLoader configLoader,
       List<TypeCodec<?>> typeCodecs,
       NodeStateListener nodeStateListener,
-      SchemaChangeListener schemaChangeListener) {
+      SchemaChangeListener schemaChangeListener,
+      RequestTracker requestTracker) {
     return new DefaultDriverContext(
-        configLoader, typeCodecs, nodeStateListener, schemaChangeListener);
+        configLoader, typeCodecs, nodeStateListener, schemaChangeListener, requestTracker);
   }
 
   private static <T> T buildIfNull(T value, Supplier<T> builder) {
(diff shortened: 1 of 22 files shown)


manual/core/logging/README.md
```diff
@@ -0,0 +1,212 @@
...
+
+Check your logging framework documentation for more information about how to properly configure it.
+You can also find some configuration examples at the end of this page.
+
+Performance tips:
+
+* Use asynchronous appenders; both
+  [Log4J](http://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/AsyncAppender.html) and
+  [Logback](http://logback.qos.ch/manual/appenders.html#AsyncAppender) provide asynchronous
+  appenders that can significantly boost latencies when writing log messages.
```
tolbertam: This particular section appears to come from 3.x documentation, but had one small nit: would reword this to something like

> [...] provide asynchronous appenders which reduce the impact of logging in latency-sensitive applications.

manual/core/logging/README.md
```diff
@@ -0,0 +1,212 @@
...
+* While the driver does not provide such capability, it is possible for client applications to
+  hot-reload the log configuration without stopping the application. This usually involves JMX and
+  is available for [Logback](http://logback.qos.ch/manual/jmxConfig.html); Log4J provides a
+  `configureAndWatch()` method but it is not recommended to use it inside J2EE containers (see
+  [FAQ](https://logging.apache.org/log4j/1.2/faq.html#a3.6)).
+
+### Taxonomy of driver logs
```
tolbertam, line 55: EXCELLENT to have this.  I'm constantly overthinking "Should this be warn or info?" and so on.  Having a nice guideline for this is immensely useful.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -135,6 +137,17 @@ public SelfT addNodeStateListeners(NodeStateListener... newListeners) {
     return self;
   }
 
+  /**
+   * Register a request tracker to use with the session.
+   *
+   * <p>If the tracker is specified programmatically with this method, it overrides the
+   * configuration (that is, the {@code request.tracker.class} option will be ignored).
+   */
+  public SelfT withRequestTracker(RequestTracker requestTracker) {
```
tolbertam, line 168: Question about this.  We have several configurations where you pass in a class name:

* auth-provider
* compressor
* load-balancing-policy
* reconnection-policy
* coalescer
* timestamp-generator
* retry-policy
* speculative-execution-policy
* throttler
* address-translator
* ssl-engine-factory

Is there a hard and fast rule for which we would want to provide a programmatic API method for passing in custom implementations?  Where does it makes sense to have a programmatic API method and where doesn't it?

manual/core/logging/README.md
```diff
@@ -0,0 +1,212 @@
...
+These levels are intended primarily for driver developers; we might ask you to enable them to
+investigate an issue.
+
+Keep in mind that they are quite verbose, in particular TRACE. It's a good idea to only enable them
+on a limited set of categories.
+
+### Logging request latencies
+
+The driver provides a built-in component to log the latency and outcome of every application
+request. See the [request tracker](../request_tracker/#request-logger) page for more details.
```
tolbertam, line 105: Good idea having a separate document for this, it's a big enough topic to merit its own standalone page 👍

integration-tests/src/test/java/com/datastax/oss/driver/api/core/tracker/RequestLoggerIT.java
```diff
@@ -0,0 +1,190 @@
...
+              "request.tracker.class = com.datastax.oss.driver.internal.core.tracker.RequestLogger",
+              "request.tracker.logs.success.enabled = true",
+              "request.tracker.logs.slow.enabled = true",
+              "request.tracker.logs.error.enabled = true",
+              "request.tracker.logs.max-query-length = 500",
+              "request.tracker.logs.show-values = true",
+              "request.tracker.logs.max-value-length = 50",
+              "request.tracker.logs.max-values = 50",
+              "request.tracker.logs.show-stack-traces = true",
+              "profiles.low-threshold.request.tracker.logs.slow.threshold = 1 nanosecond",
```
tolbertam, line 72: Awesome that you can configure this at a profile level.

core/src/main/java/com/datastax/oss/driver/internal/core/tracker/RequestLogger.java
```diff
@@ -0,0 +1,156 @@
...
+
+    StringBuilder builder = formatter.logBuilder(logPrefix, node);
+    formatter.appendErrorDescription(builder);
+    formatter.appendLatency(latencyNanos, builder);
+    formatter.appendRequest(
+        request, maxQueryLength, showValues, maxValues, maxValueLength, builder);
+    if (showStackTraces) {
+      LOG.error(builder.toString(), error);
+    } else {
+      LOG.error("{} {}", builder.toString(), error.toString());
```
tolbertam: Nit: would be nice to have some kind of obvious indicator/separator demarcating the builder message and the error string since it looks kinda awkward currently:

```
Error (2 ms) [0 values] SELECT release_version FROM system.local  com.datastax.oss.driver.api.core.servererrors.ServerError: test
```

tolbertam (approved):
Looks great 👍

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
olim7t: Good question. I would say any time there's a reasonably high chance that they would interact with some other component that is not easily accessed statically. For example if you want to send request info somewhere else than the logs:

```java
MyCustomSink sink = ...
RequestTracker tracker = new MyCustomTracker(sink);
CqlSession session = CqlSession.builder().withRequestTracker(tracker).build();
```

I think NodeStateListener and SchemaChangeListener fall into that category (see JAVA-1801).

For other policies, I'd be more conservative. We don't want to have everything in the builder and go back to the driver 3 API-based configuration. Also, it's still possible to plug your runtime components with a bit more work (extend the context and the builder).

Of course we can come up with contrived examples for any of those interfaces, but we need to figure out what is *likely*. Time will also tell, we can add a particular interface later if there's high demand for it.

manual/core/logging/README.md
olim7t: It was also in the contribution guidelines ;-)

Interestingly, we have no ERROR logs yet. When the driver is completely unusable, it's usually at startup and it throws instead.

core/src/main/java/com/datastax/oss/driver/internal/core/tracker/RequestLogger.java
olim7t: I'll also fix the double-space issue.

PR #990, opened 2018-04-10
JAVA-1799: Use CqlIdentifier for simple statement named values

olim7t:


core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class CqlIdentifiers {
+
+  public static <V> Map<CqlIdentifier, V> wrapKeys(Map<String, V> in) {
+    ImmutableMap.Builder<CqlIdentifier, V> out = ImmutableMap.builder();
+    for (Map.Entry<String, V> entry : in.entrySet()) {
+      out.put(CqlIdentifier.fromCql(entry.getKey()), entry.getValue());
+    }
+    return out.build();
+  }
```
olim7t, line 39: There should be a few places in the QueryBuilder DSL where we can reuse this, we'll address it when they both get merged.

tolbertam (approved):
looks good 👍

PR #991, opened 2018-04-11
JAVA-1801: Revisit NodeStateListener and SchemaChangeListener APIs

olim7t:


integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/NodeStateIT.java
```diff
@@ -81,7 +81,7 @@
               String.format(
                   "load-balancing-policy.filter.class = \"%s$CustomNodeFilter\"",
                   NodeStateIT.class.getName()))
-          .withNodeStateListeners(nodeStateListener)
+          .withNodeStateListener(nodeStateListener)
```
tolbertam, line 82: The formatter fails for this class, I think because `Session` and `SessionWrapper` are listed as imports that are no longer needed.  Did not know the formatter validated unused imports, cool!

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -130,8 +132,26 @@ public SelfT addTypeCodecs(TypeCodec<?>... typeCodecs) {
...
 
-  public SelfT addNodeStateListeners(NodeStateListener... newListeners) {
-    Collections.addAll(this.nodeStateListeners, newListeners);
+  /**
+   * Registers a node state listener to use with the session.
+   *
+   * <p>If the listener is specified programmatically with this method, it overrides the
+   * configuration (that is, the {@code metadata.node-state-listener.class} option will be ignored.
+   */
+  public SelfT withNodeStateListener(NodeStateListener nodeStateListener) {
```
tolbertam, line 143: 👍 makes sense to have builder methods for these based on our [discussion](https://github.com/datastax/java-driver/pull/989#discussion_r180159691) in #989.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeStateListenerBase.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+package com.datastax.oss.driver.internal.core.metadata;
+
+import com.datastax.oss.driver.api.core.metadata.Node;
+import com.datastax.oss.driver.api.core.metadata.NodeStateListener;
+
+/**
+ * Convenience class for node state listener implementations that only need to override one method.
+ */
+public class NodeStateListenerBase implements NodeStateListener {
```
tolbertam: Should this and `SchemaChangeListenerBase` really be in the internal package?  It sort of makes sense given that it's not an interface, but I would also expect this to be used more often than not by users, and also wouldn't expect it to change, so it seems safe to be part of API, but was just wondering your thoughts on this.

tolbertam (approved):
big 👍 .  I like that this removes some methods from the `Session` API, and also the simplicity that comes from only having 1 listener instance to deal with.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeStateListenerBase.java
olim7t: I was just wondering about that too. I think it makes sense to keep it in the API package after all.

PR #992, opened 2018-04-11
CodecRegistry & GenericType API enhancements

adutra:


core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -367,8 +341,7 @@ protected CachingCodecRegistry(String logPrefix, TypeCodec<?>... userCodecs) {
     return builder.build();
   }
 
-  // We call this after validating the types, so we know the cast will never fail.
-  private static <T, U> TypeCodec<T> safeCast(TypeCodec<U> codec) {
+  private static <T, U> TypeCodec<T> uncheckedCast(TypeCodec<U> codec) {
```
adutra: This is far from "safe", i.e. the following compiles and runs just fine:

```
  TypeCodec<Number> codec = unsafeCast(TypeCodecs.TEXT);
  System.out.println(codec);
  // prints com.datastax.oss.driver.internal.core.type.codec.StringCodec@ea30797
``` 

The only benefit of having this method imo si to reduce the scope of the `@SuppressWarnings` annotation to the minimum, but the name is misleading.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -302,9 +288,27 @@ protected CachingCodecRegistry(String logPrefix, TypeCodec<?>... userCodecs) {
...
+    }
+    for (TypeCodec<?> userCodec : userCodecs) {
+      if (userCodec.getJavaType().isSupertypeOf(javaType)) {
+        LOG.trace("[{}] Found matching user codec {}", logPrefix, userCodec);
+        return uncheckedCast(userCodec);
+      }
+    }
+    return uncheckedCast(getCachedCodec(null, javaType));
+  }
+
```
olim7t: Please don't move stuff around for no reason, this is really hard to review. There are no structural changes to this method, yet I had to do a line-by-line comparison.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/registry/CodecRegistry.java
```diff
@@ -35,11 +36,39 @@
...
+        TypeCodecs.BOOLEAN,
+        TypeCodecs.SMALLINT,
+        TypeCodecs.TINYINT,
+        TypeCodecs.DATE,
+        TypeCodecs.TIME,
+        TypeCodecs.DURATION,
+        TypeCodecs.COUNTER,
+        TypeCodecs.ASCII
+      };
+
```
olim7t: 👍 , but I would move it to `DefaultCodecRegistry`: this is an implementation detail and I don't see a strong case for making it part of the public API.

olim7t (changes requested):
LGTM, see minor comments inline.

Once again I'd like to call your attention to the "reviewability" of PRs: avoid cosmetic changes, isolate trivial refactorings in separate commits. We really need to ingrain those habits to streamline reviews.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/registry/CodecRegistry.java
adutra: Believe me or not, It doesn't work. If `DEFAULT_PRIMITIVE_CODECS` is declared in `DefaultCodecRegistry` or `CachingCodecRegistry`,  then when you instantiate `DefaultCodecRegistry` from `CodecRegistry.DEFAULT`, if the class is not loaded yet the constructor will be invoked _before_ the static initializers, and as a consequence, `DEFAULT_PRIMITIVE_CODECS` will be null. That's why I moved `DEFAULT_PRIMITIVE_CODECS` to `CodecRegistry`. There must be some circularity going on. The only alternative would be to declare `DEFAULT_PRIMITIVE_CODECS` in a separate class.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
adutra: Hmm don't you think it's easier to follow now that this method is declared below the method where it is invoked from?

olim7t: Not necessarily, one might prefer to keep all the "codecFor" methods together because they do the same thing (this was probably my train of thought at the time).

But that's not the point, it's about minimising changes to streamline reviews.

adutra: Fair enough, my apologies. And besides, I need this method exposed publicly so my argument in favor of moving it doesn't hold anymore.

olim7t (approved):

olim7t:
Rebased on 4.x and cleaned up the history.

I had to restart from scratch and apply all the changes manually because things were too entangled. I've made a diff with your old branch (a9a99eb154c31b8bff71b78ec594d7dba7cd1d45) and everything seems to be here, please double-check.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/registry/CodecRegistry.java
```diff
@@ -70,6 +70,26 @@
...
   <T> TypeCodec<T> codecFor(DataType cqlType);
 
+  /**
+   * Returns a codec to convert the given Java type to the CQL type deemed most appropriate to
+   * represent it.
+   *
+   * <p>The definition of "most appropriate" is unspecified, and left to the appreciation of the
+   * registry implementor.
+   */
+  <T> TypeCodec<T> codecFor(GenericType<T> javaType);
```
olim7t: I still have doubts about this, in particular with the fact that the implementation _happens to be_ covariant. Is this an implementation detail, or should we make it part of the contract? Or should it actually be invariant, and require a dedicated implementation instead of reusing what we had previously?

It doesn't help that we have no identified use case for this method. You mentioned people in the community requested it, do you remember where that conversation happened?

I just don't want this to become a "it sounded like a good idea at the time" moment 😉

adutra: The discussion has been captured in [JAVA-1015](https://datastax-oss.atlassian.net/browse/JAVA-1015).

The original pull request had an invariant implementation (i.e. it used `codec.accepts(javaType)`, not `codec.getJavaType().isSuperTypeOf(javaType)`).

Indeed I'd rather keep it invariant. Suppose that it's covariant, that we have `B extends A` and there is a codec for `A`. The following would happen:

```
BoundStatement s = ...
s.set(0, new B(), codecFor(B.class)); // ok, will return TypeCodec<A> which can encode B instances
B b = s.get(0, codecFor(B.class)); // TypeCodec<A> will decode B to an A instance -> ClassCastException
```

But that's going to be far from trivial. 

Because there is something new in 4.x: previously lookups by value wouldn't be cached, but now they are. 

What was your motivation to change that?

The practical consequence is that lookups by java type now _have_ to be covariant, whether or not the initial entry point was a lookup by value or by java type:

1. `codecFor(Object)` or `codecFor(GenericType<?>)` may both call
2. `getCachedCodec(DataType, GenericType<?>)` which may call
3. `createCodec(DataType, GenericType<?>)` which may call (since `cqlType` is null)
4. `createCovariantCodec(GenericType<?> javaType)`, which may call
5. `codecFor(GenericType<T> javaType)` (previously named `covariantCodecFor(GenericType<T> javaType)`)

... which must be covariant because the entry point above might have been `codecFor(Object)`, which requires covariance.

I only see two ways to fix this issue:

1. Do not cache lookups by value, use covariant lookups only for these lookups.
2. Keep caching lookups by value, but store them in a separate cache that creates covariant codecs.

I will provide a solution using approach 1 above, as `codecFor(Object)` is only used in `SimpleStatement`s and I don't think that performance is paramount.

adutra:
Pushed a few more commits that bypass the cache for lookups by value, and that only creates covariant codecs for these lookups.

We'll have to squash and rebase again, but let's first agree on the solution.

adutra:
One more thing: `CachingCodecRegistry` does not hold a cache itself, but `DefaultCodecRegistry` does. Their names are imo slightly confusing.

adutra:
4093733 add a missing constructor after your rebase.
e80a183 brings in some recent changes that we were missing.

olim7t:
> The discussion has been captured in JAVA-1015.

Thanks. After reading #557 again, I see the case for schema generation, and I agree that the method should be invariant (the Java classes that they generate the schema from will declare the exact codec types like `List<String>`, not `ArrayList<String>`).

> Because there is something new in 4.x: previously lookups by value wouldn't be cached, but now they are. What was your motivation to change that?

The same as for other methods: cache stuff so that we don't recompute it every time.

> I only see two ways to fix this issue

Another way is to capture the kind of lookup (object vs. type) in the cache key, I'll give it a try.

> CachingCodecRegistry does not hold a cache itself, but DefaultCodecRegistry does. Their names are imo slightly confusing.

`CachingCodecRegistry` is the base infrastructure for a registry that's backed by a cache. The cache itself is an implementation detail, that's why it's not in the class. If you wrote a registry that doesn't cache, you would extend the interface directly, not this class. Therefore every subclass "is a" caching registry; you name the abstract type after the common functionality shared by subclasses.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodec.java
```diff
@@ -86,7 +86,7 @@ default boolean accepts(Class<?> javaClass) {
    */
   default boolean accepts(Object value) {
     Preconditions.checkNotNull(value);
-    return getJavaType().__getToken().isSupertypeOf(TypeToken.of(value.getClass()));
+    return getJavaType().isSupertypeOf(GenericType.of(value.getClass()));
```
olim7t, line 89: 👍 for isolating that in separate commits, that makes it easy to squash with the "Enhance GenericType..." one.

adutra:
> CachingCodecRegistry is the base infrastructure for a registry that's backed by a cache. [..] If you wrote a registry that doesn't cache, you would extend the interface directly, not this class. 

Not sure, `CachingCodecRegistry` has useful plumbery to inspect primitive and user codecs, so it's a useful base class, regardless of whether you are caching or not.

And my point is precisely that `CachingCodecRegistry` has no way to force implementors to  use a cache, one could implement `getCachedCodec` without caching anything, so stating that this class is a caching codec registry is merely a wishful thinking.

Which led me to think that 1) this base class would be better named `AbstractCodecRegistry` and 2) its method `getCachedCodec` would be better named `newCodec` or `lookupCodec` or stg similar, and 3) the decision to cache or not should be left to implementors/subtypes.

But all this is just me nitpicking, it's fine if the names stay as they are now.

olim7t:
To expand on the schema generation scenario: if a class exposes a `B` field and we only have an `A<=>int` codec in the registry with`B extends A`, making `codecFor(B.class)` covariant would be bad because it would make it look like a `B<=>int` mapping is valid, whereas it will fail for decoding.

olim7t:
I implemented my suggestion, it's squashed with "Add ability to lookup codec by Java type only".
I rebased again, hoping that this will be the final version...

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/registry/CodecRegistry.java
```diff
@@ -65,26 +77,40 @@
...
    *
-   * <p>The definition of "most appropriate" is unspecified, and left to the appreciation of the
-   * registry implementor.
+   * <p>The driver does not use this method. It is provided as a convenience for third-party usage,
+   * for example if you were to generate a schema based on a set of Java classes.
+   *
+   * <p>The driver's default registry implementation is <em>invariant</em> with regard to the Java
+   * type: for example, if {@code B extends A} and an {@code A<=>int} codec is registered, {@code
+   * codecFor(DataTypes.INT, B.class)} <b>will not</b> find that codec. This is because this method
+   * we don't know whether this method will be used for encoding, decoding, or both.
```
adutra: Small nit here: "this is because this method we don't know..."

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/registry/CodecRegistry.java
```diff
@@ -94,8 +120,18 @@
    * Returns a codec to convert the given Java object to the CQL type deemed most appropriate to
    * represent it.
    *
-   * <p>The definition of "most appropriate" is unspecified, and left to the appreciation of the
-   * registry implementor.
+   * <p>This is used internally by the driver, in cases where the CQL type is unknown, for example
+   * for example for {@link SimpleStatement#setPositionalValues(List) simple statement variables}
```
adutra: Small nit here, "for example" is repeated twice.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -64,122 +64,146 @@
...
+        LOG.trace("[{}] Found matching user codec {}", logPrefix, userCodec);
+        return uncheckedCast(userCodec);
+      }
+    }
+    return uncheckedCast(getCachedCodec(null, javaType, false));
   }
 
   // Not exposed publicly, this is only used for the recursion from
   // createCovariantCodec(GenericType)
   private TypeCodec<?> covariantCodecFor(GenericType<?> javaType) {
```
adutra: This has to be protected, subclasses may need to call this method directly if they override `createCovariantCodec`.

adutra:
Went ahead and fixed the issues I had, squashed and rebased. Please have a last look if you wish, then feel free to merge.

olim7t:
🎉

PR #995, opened 2018-04-17
 JAVA-1798: Allow passing the default LBP filter as a session builder argument

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -150,13 +154,32 @@ public SelfT withNodeStateListener(NodeStateListener nodeStateListener) {
...
 
+  /**
+   * A custom filter to include/exclude nodes.
+   *
+   * <p>The predicate's {@link Predicate#test(Object) test()} method will be invoked each time the
+   * {@link LoadBalancingPolicy} processes a topology or state change: if it returns false, the node
+   * will be set at distance IGNORED (meaning the driver won't ever connect to it), and never
+   * included in any query plan.
+   *
+   * <p>Note that this behavior is implemented in the default load balancing policy. If you use a
```
adutra, line 185: Alternatively, couldn't we add a `Predicate<Node>` argument to `LoadBalancingPolicy.init()` and inject the predicate into the policy there? This way users don't need to reinvent the wheel if they have a custom LBP.

adutra (approved):

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
olim7t: That would be a way to factor the "combine configuration and builder" thing, but custom LBPs would still have to apply the filter in every onXxx method. Also, injecting things in `init` is not the best because you can't store them in final fields.

But your comment makes me realize that the filter does not have to be part of the policy either. We could make it a top-level component and apply it before calling the policy (in `LoadBalancingPolicyWrapper`). In that case it doesn't really make a lot of sense to have it in the configuration, it could be builder-only.

PR #997, opened 2018-04-17
JAVA-1762: Build alternate core artifact with Netty shaded

olim7t:


core/pom.xml
```diff
@@ -15,7 +15,9 @@
     limitations under the License.
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
adutra, line 18: This change seems spurious.

adutra:
I'm wondering: with JAVA-1787 we provided an external shaded artifact for Guava, while here we are taking the old approach of creating an alternate artifact during the build.

How is the DSE driver supposed to behave in the presence of a Netty-shaded OSS driver? I don't see how this could work since the DSE driver exposes Netty classes in internal types, e.g. `DseDriverContext.buildFrameCodec()`. Are we going to create an alternate artifact for the DSE driver too?

tolbertam:
I think you are onto something @adutra, I think it would make the most sense to take a similar route to how we shade guava, otherwise we would run into the same problem with the shaded DSE driver where netty is shaded for each library.

The shaded classifier for each library could rewrite the package imports but not bring in the netty classes, but instead reference the netty shaded jar.

olim7t:
> The shaded classifier for each library could rewrite the package imports but not bring in the netty classes, but instead reference the netty shaded jar.

If we can coax the shade plugin into doing that, that does indeed sound like a viable solution. But from a quick look at the doc it won't be trivial, more investigation needed.

olim7t:
Closing in favor of #1019

PR #998, opened 2018-04-20
JAVA-1569: Allow null to be used in positional and named values in statements

adutra:


olim7t:
`Collections.unmodifiableList` is just a view of the underlying collection, so it does not guarantee thread safety.
Also, I've noticed that `setPositionalValues` and `setNamedValuesWithIds` don't wrap their argument (my bad, this is a shortcoming of my original implementation). `setNamedValuesWith` would also need to use a specialised version of `CqlIdentifiers.wrapKeys`.

I was going to suggest sticking to immutable collections and using a special value for null, but it's going to leak through the public API via `getPositionalValues` and `getNamedValues` (and we can't really rewrap in those methods, because that's what the processor calls internally).

adutra:
> Collections.unmodifiableList is just a view of the underlying collection, so it does not guarantee thread safety.

But are simple statements required to be thread-safe? I didn't know it was the case (it's not annotated with `@ThreadSafe`).

olim7t:
Actually, is it that bad to leak a special value through the API? We already did it for custom payloads in 3.x (and I should have ported that as well btw).
We could catch any NPE thrown by immutable collections if the user tries to insert `null`, and rethrow it as an `IllegalArgumentException` that says "use `SimpleStatement.NULL_VALUE` instead".

adutra:
> Actually, is it that bad to leak a special value through the API? 

I personally think that the cure would be worse than the disease. I would still be more in favor of leaving `SimpleStatement` implementations non-thread-safe.

However if you think thread safety is paramount for simple statements, then I don't think we absolutely need to expose a special value: I pushed an attempt to illustrate this idea in a separate branch: https://github.com/datastax/java-driver/tree/java1569b.

olim7t:
I took the plunge and implemented custom collections that support null elements. Beyond this ticket, I think we should move them to native-protocol and use them for all message properties.

tolbertam (approved):
👍 initially I had an aversion to creating our own immutable collections that mirror the guava ones, except accepting nulls, but I think if we limit our use of them to where we need them, it's fine.  It also doesn't look like too much code and I can't imagine problems arising out of it.

With regards to moving to native-protocol, I think it would be fine, I think as long as we keep 'internal' as part of the package structure to make it clear we don't to support its use outside our own needs.

olim7t:
Requires a local install of datastax/native-protocol@1261d355a146865dadce3ac5692d7236a4867d35

PR #1000, opened 2018-04-24
JAVA-1779: Add dynamic port discovery for system.peers_v2

GregBestland:


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeInfo.java
```diff
@@ -44,29 +43,27 @@
...
    *
    * <p>This is currently not used anywhere in the driver. If you write a custom topology monitor
    * and don't need this information, you can leave it empty.
    */
-  Optional<InetAddress> getListenAddress();
-
+  Optional<InetSocketAddress> getListenAddress();
   /**
-   * The data center that this node belongs to, according to the Cassandra snitch.
+   * The node's listen port. That is, the port that the Cassandra process binds to.
```
tolbertam: Accidental comment change?  This was made to `getDatacenter`

GregBestland:
Just an FYI, some IT tests are failing due to a simulacron issue, I will sort this out and push a new version of simulacron. We want to make sure we do this prior to merging.

core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOption.java
```diff
@@ -79,6 +79,8 @@
   REQUEST_LOGGER_STACK_TRACES("request.tracker.logs.show-stack-traces", false),
 
   CONTROL_CONNECTION_TIMEOUT("connection.control-connection.timeout", true),
+  CONTROL_CONNECTION_ALLOW_PORT_DISCOVERY(
+      "connection.control-connection.allow-port-discovery", false),
```
olim7t: Why make this configurable?
The only use I can think of for this option is "the cluster has CASSANDRA-7544, but I still want to force the same port for all nodes". But why would we want to do that? If some nodes actually use different ports, the driver will fail to connect to them. If they all use the same port, I assume that it's present in peers_v2 anyway, so we don't need to handle this as a special case.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -188,13 +230,19 @@ public DefaultTopologyMonitor(InternalDriverContext context) {
     return query(channel, queryString, Collections.emptyMap());
   }
 
+  private String fetchPeersTable() {
```
olim7t: Nit: I find the name a bit misleading, to me "fetch" implies that we retrieve it for somewhere (like a database query).

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -215,9 +263,17 @@ private NodeInfo asNodeInfo(AdminRow row) {
     if (broadcastAddress == null) {
       broadcastAddress = row.getInetAddress("peer"); // in system.peers
     }
-    builder.withBroadcastAddress(broadcastAddress);
-
-    builder.withListenAddress(row.getInetAddress("listen_address"));
+    int broadcastPort = 0;
+    if (row.contains("peer_port")) {
+      broadcastPort = row.getInteger("peer_port");
+    }
```
olim7t, line 264: So we use 0 as a special value for "port unknown". This bothers me a bit because 0 already has a different meaning in the JDK: "A port number of zero will let the system pick up an ephemeral port in a bind operation".
On the other hand:
* the alternative would be to keep the two values separate and use `Optional`, which sounds a bit overkill
* this will only happen for `node.getBroadcastAddress()` and `node.getListenAddress()`, which are never used by a typical enterprise application (and even an advanced tool will never bind to them)

So overall I'm fine keeping it this way, but let's document it in `Node`'s javadoc.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -252,4 +305,19 @@ private void savePort(DriverChannel channel) {
       port = ((InetSocketAddress) channel.remoteAddress()).getPort();
     }
   }
+
+  private InetSocketAddress getNativeAddressFromPeers(AdminRow row) {
+    InetAddress native_address = row.getInetAddress("native_address");
+    if (native_address == null) {
```
olim7t: Nit: could you use camel case for variable names?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -133,10 +156,29 @@ public DefaultTopologyMonitor(InternalDriverContext context) {
     savePort(channel);
 
     CompletionStage<AdminResult> localQuery = query(channel, "SELECT * FROM system.local");
-    CompletionStage<AdminResult> peersQuery = query(channel, "SELECT * FROM system.peers");
+    CompletionStage<AdminResult> peersV2Query = query(channel, "SELECT * FROM system.peers_v2");
+    CompletableFuture<AdminResult> peersQ = new CompletableFuture<>();
```
olim7t: Nit: could you name this `Query` like the other two variables?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -133,10 +156,29 @@ public DefaultTopologyMonitor(InternalDriverContext context) {
...
+            // The query to system.peers_v2 failed, whe should not attempt this query in the future.
+            // Since this method should be run as part of the control connection initialization, we
+            // shouldn't
+            // need this fallback anywhere else.
+            this.isSchemaV2 = false;
+            query(channel, "SELECT * FROM system.peers")
+                .whenComplete(
+                    (r2, t2) -> {
+                      peersQ.complete(r2);
+                    });
```
olim7t: I wrote a utility method for that, you can replace lines 170-174 by:
```java
CompletableFutures.completeFrom(query(channel, "SELECT * FROM system.peers"), peersQ);
```
As a bonus, it will also propagate the error if the query fails.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
olim7t: It saddens me that we have to use an intermediary future for that, but indeed I don't see any other way. We would need a mix of `thenCompose` and `exceptionally` (in other words, chain the second async call only when an exception happened), there is no such method in the API.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -133,10 +156,29 @@ public DefaultTopologyMonitor(InternalDriverContext context) {
...
+          } else {
+            peersQ.complete(r);
+          }
+        });
```
olim7t: Could you chain `.exceptionally(UncaughtExceptions::log)` here?
It's very unlikely that an exception would be thrown from your `whenComplete` callback, but if that happens it would be silently ignored with the current code (because we discard the resulting future). With the log we at least detect it.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -133,10 +156,29 @@ public DefaultTopologyMonitor(InternalDriverContext context) {
...
+    CompletableFuture<AdminResult> peersQ = new CompletableFuture<>();
+
+    peersV2Query.whenComplete(
+        (r, t) -> {
+          if (t != null) {
+            // The query to system.peers_v2 failed, whe should not attempt this query in the future.
+            // Since this method should be run as part of the control connection initialization, we
+            // shouldn't
+            // need this fallback anywhere else.
+            this.isSchemaV2 = false;
```
olim7t: There is an edge case where it would be done after connection initialization: if we're in a mixed 3.x/4.0 cluster and the first contact point was a 4.0 node, `isSchemaV2` is set when we first connect.
If the control connection later drops and reconnects to a 3.x node, it will reach this line and flip the flag back to false. But the driver recovers correctly: the first thing we do on a control connection reconnect is a `refreshNodeList`, so there's no risk of running a `refreshNode` with the flag still set to true. And if we're in a mixed cluster, all nodes should still be using the same ports, so it doesn't make any difference whether we use `peers` or `peers_v2` after the reconnection.

So I don't think we need to change anything to handle that case.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeInfo.java
```diff
@@ -44,29 +43,27 @@
    * which the request was made, otherwise the new node will be ignored.
    */
   InetSocketAddress getConnectAddress();
-
```
olim7t, line 47: Nit: probably unintended I guess

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeInfo.java
```diff
@@ -44,29 +43,27 @@
...
+  Optional<InetSocketAddress> getListenAddress();
   /**
-   * The data center that this node belongs to, according to the Cassandra snitch.
+   * The node's listen port. That is, the port that the Cassandra process binds to.
    *
-   * <p>This is used by some {@link LoadBalancingPolicy} implementations to compute the {@link
-   * NodeDistance}.
+   * <p>This is currently not used anywhere in the driver. If you write a custom topology monitor
+   * and don't need this information, you can leave it empty.
```
olim7t: Mmm I disagree, `getDatacenter()` is still used by the default LBP. Not exactly for distance, but non-local nodes are ignored based on this information.

olim7t (changes requested):
Looks pretty good, no major remarks.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
tolbertam: :+1:.  The 3.x PR creates a new type [`InetAddressOptPort`](https://github.com/aweisberg/java-driver/blob/648fb67fd14ef0b263b90d50a912c724576a924a/driver-core/src/main/java/com/datastax/driver/core/utils/InetAddressOptPort.java).  Greg and I talked about it and agreed it would be better to just assign the port to 0 in the case it can't be discerned instead of creating a new type.  I agree that we should document these methods and indicate the port can be 0 if it can't be determiend.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
tolbertam: agree, also in a mixed version cluster, the user shouldn't be setting up multiple nodes on the same interface, so I agree that it's ok to drop back to the `peers` table permanently.

tolbertam (commented):
This seems to be working great for me, 👍 when the suggestions Olivier recommended are implemented.

core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOption.java
newkek: I wondered the same, there is more info on that on JAVA-1388 (btw why is there 2 tickets?):
```
By default the client won't using the discovered port numbers. When building the client you have to specify allowing
port discovery. This nit is necessary for transitioning from a cluster where discovered port numbers are for the 
non-SSL port. Only the native port that supports both SSL and non-SSL traffic is propagated. In the future people
will only use one port for both SSL and non-SSL, but during the transition the client needs to now if it should trust
the discovered port number or use the one that is supplied in the client configuration.
```
Although that explanation is a little unclear to me

olim7t: re separate tickets: it makes it easier to track progress across the two codebases. Otherwise we mark the JIRA ticket as "fixed" when it's merged to driver 3 but not yet to driver 4, or vice-versa.

newkek: Agree although I thought both tickets were targeted to driver 4 but maybe I was confused because on JAVA-1338 it says "Affects Version 4.0.0"

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
GregBestland: Sorry man I this is what I get when I constantly switch from python to java.

PR #1001, opened 2018-05-01
Fix links in Upgrade guide

alexott:
Links aren't correctly formatted in 3.5 section of Upgrade guide (https://docs.datastax.com/en/developer/java-driver/3.5/upgrade_guide/). This PR fixes this small problem

P.S. Although I'm not sure that it makes sense to move links that are used only once onto
separate lines - this usually makes sense when the same link is reused.


tolbertam (approved):
👍 Good catch and thanks for fixing!

PR #1005, opened 2018-05-14
JAVA-1592: Expose statement's total Frame size through API.

newkek:
## Notes for reviewers
- requires some changes from the `native-protocol` project that I [pushed on a `java1592` branch on the repo](https://github.com/datastax/native-protocol/tree/java1592) - these aren't hugely important, but it seemed to make more sense to have these changes there. To compile this PR you'd need to compile and install the native-protocol project of that branch.
- the method to compute a statement's size requires a `DriverContext`. It is the only way I have found to be able to use `protocolVersionRegistry.supports(ProtocolVersion, ProtocolFeature)` feature.
- It is a complete re-write from the original JAVA-708 PR since most of the codebase has changed and wasn't compatible.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/cql/Statement.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/cql/Statement.java
@@ -18,6 +18,7 @@ package com.datastax.oss.driver.api.core.cql;
 import com.datastax.oss.driver.api.core.CqlIdentifier;
 import com.datastax.oss.driver.api.core.CqlSession;
 import com.datastax.oss.driver.api.core.config.DriverConfigProfile;
+import com.datastax.oss.driver.api.core.context.DriverContext;
 import com.datastax.oss.driver.api.core.metadata.token.Token;
 import com.datastax.oss.driver.api.core.session.Request;
 import com.datastax.oss.driver.api.core.session.Session;
@@ -164,6 +165,20 @@ public interface Statement<T extends Statement<T>> extends Request {
    * if you do so, you must override {@link #copy(ByteBuffer)}.
    */
   T setPagingState(ByteBuffer newPagingState);
+
+  /**
+   * Calculates the approximate size in bytes that the statement will have when encoded.
+   *
+   * <p>The size might be over-estimated by a few bytes due to global options that may be defined on
+   * a {@link Session} but not explicitly set on the statement itself.
+   *
+   * <p>The result of this method is not cached, calling it will cause some encoding to be done in
+   * order to determine some of the statement's attributes sizes. Therefore, use this method
+   * sparingly in order to avoid unnecessary computation.
+   *
+   * @return the approximate number of bytes this statement will take when encoded.
+   */
+  int computeSizeInBytes(DriverContext context);
 
   /**
    * Creates a <b>new instance</b> with a different paging state.
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -4,6 +4,7 @@
 
 ### 4.0.0-alpha4 (in progress)
 
+- [new feature] JAVA-1592: Expose request's total Frame size through API
 - [new feature] JAVA-1829: Add metrics for bytes-sent and bytes-received 
 - [improvement] JAVA-1755: Normalize usage of DEBUG/TRACE log levels
 - [improvement] JAVA-1803: Log driver version on first use
--- a/core/src/main/java/com/datastax/oss/driver/api/core/cql/BatchStatement.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/cql/BatchStatement.java
@@ -17,9 +17,13 @@ package com.datastax.oss.driver.api.core.cql;
 
 import com.datastax.oss.driver.api.core.CqlIdentifier;
 import com.datastax.oss.driver.api.core.DefaultProtocolVersion;
+import com.datastax.oss.driver.api.core.context.DriverContext;
 import com.datastax.oss.driver.api.core.session.Request;
 import com.datastax.oss.driver.internal.core.cql.DefaultBatchStatement;
+import com.datastax.oss.driver.internal.core.time.ServerSideTimestampGenerator;
+import com.datastax.oss.driver.internal.core.util.Sizes;
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
+import com.datastax.oss.protocol.internal.PrimitiveSizes;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -155,4 +159,41 @@ public interface BatchStatement extends Statement<BatchStatement>, Iterable<Batc
    * method. However custom implementations may choose to be mutable and return the same instance.
    */
   BatchStatement clear();
+
+  @Override
+  default int computeSizeInBytes(DriverContext context) {
+    int size = Sizes.minimumStatementSize(this, context);
+
+    // BatchStatement's additional elements to take into account are:
+    // - batch type
+    // - inner statements (simple or bound)
+    // - per-query keyspace
+    // - timestamp
+
+    // batch type
+    size += PrimitiveSizes.BYTE;
+
+    // inner statements
+    size += PrimitiveSizes.SHORT; // number of statements
+
+    for (BatchableStatement<?> batchableStatement : this) {
+      size +=
+          Sizes.sizeOfInnerBatchStatementInBytes(
+              batchableStatement, context.protocolVersion(), context.codecRegistry());
+    }
+
+    // per-query keyspace
+    if (getKeyspace() != null) {
+      size += PrimitiveSizes.sizeOfString(getKeyspace().asInternal());
+    }
+
+    // timestamp
+    if (!(context.timestampGenerator() instanceof ServerSideTimestampGenerator)
+        || getTimestamp() != Long.MIN_VALUE) {
+
+      size += PrimitiveSizes.LONG;
+    }
+
+    return size;
+  }
 }
--- a/core/src/main/java/com/datastax/oss/driver/api/core/cql/BoundStatement.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/cql/BoundStatement.java
@@ -16,6 +16,11 @@
 package com.datastax.oss.driver.api.core.cql;
 
 import com.datastax.oss.driver.api.core.CqlIdentifier;
+import com.datastax.oss.driver.api.core.context.DriverContext;
+import com.datastax.oss.driver.internal.core.time.ServerSideTimestampGenerator;
+import com.datastax.oss.driver.internal.core.util.Sizes;
+import com.datastax.oss.protocol.internal.PrimitiveSizes;
+import com.datastax.oss.protocol.internal.request.query.Values;
 import java.nio.ByteBuffer;
 import java.util.List;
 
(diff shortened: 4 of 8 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -441,4 +450,77 @@ public static CoordinatorException toThrowable(
         return new ProtocolError(node, "Unknown error code: " + errorMessage.code);
     }
   }
+
+  /** Returns a common size for all kinds of Request implementations. */
```
newkek: I didn't find better class than this to expose these utility methods..

newkek:
Travis builds fail because of the dependency to native-protocol's changes.

core/src/main/java/com/datastax/oss/driver/api/core/cql/Statement.java
```diff
@@ -165,6 +166,20 @@ default T setRoutingKey(ByteBuffer... newRoutingKeyComponents) {
...
+   * <p>The size might be over-estimated by a few bytes due to global options that may be defined on
+   * a {@link Session} but not explicitly set on the statement itself.
+   *
+   * <p>The result of this method is not cached, calling it will cause some encoding to be done in
+   * order to determine some of the statement's attributes sizes. Therefore, use this method
+   * sparingly in order to avoid unnecessary computation.
+   *
+   * @return the approximate number of bytes this statement will take when encoded.
+   */
+  int computeSizeInBytes(DriverContext context);
```
tolbertam, line 181: The 3.x implementation used the name `requestSizeInBytes` for this method, where we use `computeSizeInBytes`  here.  I think I like the new name better (since `compute` seems more explicit), but just wanted to point out they were different.

core/src/main/java/com/datastax/oss/driver/api/core/cql/Statement.java
```diff
@@ -18,6 +18,7 @@
 import com.datastax.oss.driver.api.core.CqlIdentifier;
 import com.datastax.oss.driver.api.core.CqlSession;
 import com.datastax.oss.driver.api.core.config.DriverConfigProfile;
+import com.datastax.oss.driver.api.core.context.DriverContext;
```
tolbertam, line 21: The JAVA-708 PR (#978) also introduced new metrics for measuring the number of bytes/received sent (via Inbound/OutboundTrafficMeter channel handlers),  we should make sure to add metrics for that, but it doesn't need to be done in this PR.  I created [JAVA-1829](https://datastax-oss.atlassian.net/browse/JAVA-1829) for that.

core/src/main/java/com/datastax/oss/driver/internal/core/ProtocolFeature.java
```diff
@@ -31,5 +31,19 @@
...
+
+  /**
+   * The ability to add a custom set of key/value pairs associated to each request.
+   *
+   * @see <a href="https://issues.apache.org/jira/browse/CASSANDRA-8553">CASSANDRA-8553</a>
+   */
+  CUSTOM_PAYLOAD,
+
+  /**
+   * The ability at the protocol level to detect when a prepared statement has had its columns altered.
```
tolbertam: This line throws off the formatter as it's too long I think.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -441,4 +450,77 @@ public static CoordinatorException toThrowable(
...
+      size += PrimitiveSizes.sizeOfBytesMap(statement.getCustomPayload());
+    }
+
+    /* These are options in the protocol inside a frame that are common to all Statements */
+
+    size += queryFlagsSize(context.protocolVersion().getCode());
+
+    size += PrimitiveSizes.SHORT; // size of consistency level
+    size += PrimitiveSizes.SHORT; // size of serial consistency level
+    size += PrimitiveSizes.LONG; // size of default timestamp
```
tolbertam: It's possible if using a server timestamp generator and timestamp is not explicitly set that we won't be sending a timestamp with the request.   I noticed that the 3.x driver also assumes timestamp is always present.  Do you think we should evaluate the statement and timestamp generator from `DriverContext` to determine if the timestamp will be provided?  On the other hand, using server side timestamps is less common these days so maybe not worth it.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -441,4 +450,77 @@ public static CoordinatorException toThrowable(
...
+      SimpleStatement simpleStatement,
+      ProtocolVersion protocolVersion,
+      CodecRegistry codecRegistry) {
+    int size = 0;
+    if (!simpleStatement.getPositionalValues().isEmpty()) {
+      int totalValuesSize =
+          Values.sizeOfPositionalValues(
+              simpleStatement
+                  .getPositionalValues()
+                  .stream()
```
tolbertam: In general we are trying to avoid using streams if possible (reasoning outlined [here](https://github.com/datastax/java-driver/blob/4.x/CONTRIBUTING.md#no-stream-api)).

tolbertam (changes requested):
Looking great so far!  Had a few comments but for most part I think this is close to good to go.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
newkek: Hm ok. Will modify to follow the guidelines.

core/src/main/java/com/datastax/oss/driver/api/core/cql/Statement.java
tolbertam: I'll make an attempt at implementing that (looks easy) and will open up a separate PR.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
newkek: Yes initially there wasn't a DriverContext in the methods parameters, only protocol version and codec registry, then I realized it wasn't good without the DriverContext / ProtocolVersionRegistry so I added it. Didn't think it was worth it to make that check but I guess since DriverContext is there now why not

core/src/main/java/com/datastax/oss/driver/internal/core/ProtocolFeature.java
newkek: Right, made this change last minute and didn't re-run `mvn install` thanks

newkek:
Processed feedback from @tolbertam, thanks for reviewing

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -489,32 +488,27 @@ static int sizeOfSimpleStatementValues(
...
-          Values.sizeOfPositionalValues(
-              simpleStatement
-                  .getPositionalValues()
-                  .stream()
-                  // need to convert Object into ByteBuffer
-                  .map(val -> Conversions.encode(val, codecRegistry, protocolVersion))
-                  .collect(Collectors.toList()));
-      size += totalValuesSize;
+
+      List<ByteBuffer> positionalValues =
```
tolbertam: Instead of building up a new List (or Map in the named value case), you could create a running count in place, using `Values.sizeOfValue` for each value and then adding `PrimtiveSizes.SHORT` at the end for the positional value case . That would prevent need for building a new collection, but on the other hand it requires this code to have knowledge about how value collections are sized, so maybe it's better for this to be the done the way you've currently done it as i don't think this code needs to be particularly performant since its not on the hot path (user has to call `computeSizeInBytes`)

tolbertam (approved):
Looks great!  In terms of merging, not sure if we should wait for next native protocol release or if we can merge in with the snapshot requirement.  Will delegate that decision to @olim7t 👍

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
newkek: Indeed the new map/list instances are not necessary and I could have done it manually, I'd prefer however to calculate the size of values by re-using the utility function that's already there to do it so that we don't duplicate that logic in multiple places if possible.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultBatchStatement.java
```diff
@@ -467,4 +473,68 @@ public BatchStatement setTimestamp(long newTimestamp) {
         newTimestamp,
         pagingState);
   }
+
+  @Override
+  public int computeSizeInBytes(DriverContext context) {
```
olim7t: I think this could be defined as a default method on `BatchStatement` instead. That way potential 3rd-party implementations won't need to reimplement it. Private helpers can be moved to an internal utility class (either `Conversions` or create a new one and move the methods you had added to `Conversions` there as well).

Same for other statement subtypes.

core/src/main/java/com/datastax/oss/driver/internal/core/ProtocolFeature.java
```diff
@@ -31,5 +31,20 @@
...
+   */
+  CUSTOM_PAYLOAD,
+
+  /**
+   * The ability at the protocol level to detect when a prepared statement has had its columns
+   * altered.
+   *
+   * @see <a href="https://issues.apache.org/jira/browse/CASSANDRA-10786">CASSANDRA-10786</a>
+   */
+  PREPARED_RESULT_METADATA_ID,
```
olim7t: With legacy protocols, `PreparedStatement.getResultMetadataId()` returns null, so you can check that and avoid adding a new `ProtocolFeature`.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -15,11 +15,15 @@
  */
 package com.datastax.oss.driver.internal.core.cql;
 
+import static com.datastax.oss.protocol.internal.FrameCodec.headerEncodedSize;
+import static com.datastax.oss.protocol.internal.request.query.QueryOptions.queryFlagsSize;
```
olim7t: Nit: could you avoid static imports? They make the code harder to follow.
There are some exceptions for test files (see the contributing guidelines).

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -441,4 +450,71 @@ public static CoordinatorException toThrowable(
...
     }
   }
+
+  /** Returns a common size for all kinds of Request implementations. */
+  static int minimumRequestSize(Request statement, DriverContext context) {
+    Preconditions.checkArgument(
+        context instanceof InternalDriverContext,
+        "DriverContext provided cannot be used to calculate statement's size");
+
+    /* Header and payload are common inside a Frame at the protocol level */
```
olim7t: Nit: could you avoid block-style implementation comments? They are annoying because they don't nest, e.g. if I try to enclose the whole method in a block-style comment, this one closes it and I get a compile error.

core/src/main/java/com/datastax/oss/driver/api/core/cql/Statement.java
olim7t, line 181: I was thinking of ways to avoid the `DriverContext` argument (assume the latest protocol version, overestimate more things, etc), in order to avoid the need for an active session to estimate sizes.
But in the end I think this is futile: if you're working with bound statements (which should be the most common), you need an active session anyway.

olim7t (changes requested):

olim7t:
Just a general thought about this feature (not specific to this PR, this would also apply to driver 3): we know the *actual* frame length in `FrameEncoder`, it saddens me a bit to do this extra work in `computeRequestSize`. Also this is specific to `Statement`, it will have to be reimplemented for other request types.

On the other hand:
* there's no trivial way to bubble state out of the Netty handler on the write path, we would have to rely on a callback or a write to a mutable field.
* you would have to execute the query to get the size, with the current approach you estimate without executing anything. I don't see a strong use case for that, but it might be useful.
* this feature should be used mainly for testing anyway, so the extra work should not be too much of an issue.

The *response* size is easy to surface, and I think we should have it in the execution info. I'll create a ticket for it.

PS: depending on a snapshot of native-protocol is fine, that's how we've handled it previously. I'll merge the native-protocol changes.

newkek:
I think I processed all feedback, thanks for reviewing.

olim7t (approved):

olim7t:
I removed the protocol version check for the custom payload, this allows us to skip the cast to `InternalDriverContext`. If it's set with protocol v3, the request will fail to execute anyway. I think it's more likely that someone would wrap their context, than trying to measure invalid requests.

PR #1006, opened 2018-05-15
JAVA-1394: Add request-queue-depth metric

kevinconaway:
@tolbertam This PR addresses https://datastax-oss.atlassian.net/browse/JAVA-1394 by adding a `getRequestQueueDepth(Host)` method to `Session.State`

The build fails locally because the _clirr_ check so I'm not quite sure what the proper path is for adding this in.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -1,5 +1,10 @@
 ## Changelog
 
+### 3.6.0 (In progress)
+
+- [improvement] JAVA-1394: Add request-queue-depth metric.
+
+
 ### 3.5.0
 
 - [improvement] JAVA-1448: TokenAwarePolicy should respect child policy ordering.
--- a/driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
@@ -59,7 +59,7 @@ class HostConnectionPool implements Connection.Owner {
     final Set<Connection> trash = new CopyOnWriteArraySet<Connection>();
 
     private final Queue<PendingBorrow> pendingBorrows = new ConcurrentLinkedQueue<PendingBorrow>();
-    private final AtomicInteger pendingBorrowCount = new AtomicInteger();
+    final AtomicInteger pendingBorrowCount = new AtomicInteger();
 
     private final Runnable newConnectionTask;
 
--- a/driver-core/src/main/java/com/datastax/driver/core/Metrics.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Metrics.java
@@ -92,6 +92,17 @@ public class Metrics {
         }
     });
 
+    private final Gauge<Integer> requestQueueDepth = registry.register("request-queue-depth", new Gauge<Integer>() {
+        @Override
+        public Integer getValue() {
+            int value = 0;
+            for (SessionManager session : manager.sessions)
+                for (HostConnectionPool pool : session.pools.values())
+                    value += pool.pendingBorrowCount.get();
+            return value;
+        }
+    });
+
     private final Gauge<Integer> executorQueueDepth;
     private final Gauge<Integer> blockingExecutorQueueDepth;
     private final Gauge<Integer> reconnectionSchedulerQueueSize;
@@ -237,6 +248,16 @@ public class Metrics {
      */
     public Gauge<Integer> getInFlightRequests() {
         return inFlightRequests;
+    }
+
+    /**
+     * Returns the total number of enqueued requests on all Cassandra hosts.
+     *
+     * @see Session.State#getRequestQueueDepth(Host)
+     * @return The total number of enqueued requests on all Cassandra hosts.
+     */
+    public Gauge<Integer> getRequestQueueDepth() {
+        return requestQueueDepth;
     }
 
     /**
--- a/driver-core/src/main/java/com/datastax/driver/core/Session.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Session.java
@@ -463,5 +463,20 @@ public interface Session extends Closeable {
          * queries to {@code host}.
          */
         int getInFlightQueries(Host host);
+
+        /**
+         * The number of queries that are queued to be executed on a given host.
+         * <p/>
+         *
+         * This corresponds to the number of queries that are waiting to be executed.  The number of requests
+         * that could be waiting can be affected by a combination of the {@link PoolingOptions#getMaxQueueSize()},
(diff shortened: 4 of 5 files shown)


tolbertam:
Hi @kevinconaway, this looks great to me, thanks!

> The build fails locally because the clirr check so I'm not quite sure what the proper path is for adding this in.

Since `Session.State` is a public interface, adding methods will cause a breaking API change.   However I think we can probably be pragmatic about this and add a clirr exception here as it's doubtful users are creating their own `Session.State` implementations.  Will run this by the team and make sure it's alright.

Another thing that might be good to do is to add a `Gauge` to `Metrics` similar to `Metrics.inFlightRequests` that summarizes the total number of queued requests there are among all hosts:

https://github.com/datastax/java-driver/blob/c8902fd7d16794bd47fe59b2371e4ac04d948544/driver-core/src/main/java/com/datastax/driver/core/Metrics.java#L84-L93

I can make those changes if you like, otherwise feel free to add that.

kevinconaway:
> However I think we can probably be pragmatic about this and add a clirr exception here as it's doubtful users are creating their own Session.State implementations. 

That was my thinking as well

> Another thing that might be good to do is to add a Gauge to Metrics similar to Metrics.inFlightRequests that summarizes the total number of queued requests there are among all hosts

Done, thanks for the suggestion.

tolbertam:
@kevinconaway, I spoke with the team about the `Session.State` breaking change, and we decided that we need to be really strict on API compatibility after all.  While it is pragmatic to expect that no one will create a custom implementation of `Session.State`, we also need to be able to safely ensure that users can upgrade between minor versions and not have to worry about API changes breaking code.   The only cases we've added exceptions for this in the past is where an API was marked beta, relaxing parameter types, and false positives flagged by clirr.

With Java 8, we could avoid this all together with default methods on interfaces (and just return an `UnsupportedOperationException` by default or something), but we need to keep Java 6 support (for now..).

So, there are essentially two routes we could take with this:

1. Remove `Session.State#getReqeuestQueueDepth(Host)`, but keep `Metrics.getRequestQueueDepth()` (since that's not a breaking change)

or

2. Add a new public interface `Session.StateWithRequestQueueDepth` that extends `Session.State` and provides this method.  You could then explicitly cast to that type if you want the value:

```java
Session.StateWithRequestQueueDepth state = (Session.StateWithRequestQueueDepth)session.getState();
```

The 2nd solution is kinda ugly, but it would work, what do you think?

kevinconaway:
Thanks for following up on this Andrew.  I appreciate the explanation

I'm fine with removing `getReqeuestQueueDepth` from `Session.State` for the time being.  I think that having an overall view of the number of pending requests will be good enough.

I wouldn't want to inflict `Session.StateWithRequestQueueDepth` on you all :)  I fear that you could end up supporting a bunch of those marker interfaces if new metrics are requested before you move to supporting only Java8+

The only other thought that I had would be to register (and unregister as needed) the host state metrics as Gauges in the `MetricRegistry` in addition to providing the ability to query them via `Session.State`. That would allow metrics to be added on the fly without breaking API compatibility.  I personally find it somewhat cumbersome to use `Session.State` and the ability to just dump everything in the `MetricRegistry` would be nice.

kevinconaway:
>  I fear that you could end up supporting a bunch of those marker interfaces if new metrics are requested before you move to supporting only Java8+

Unless, perhaps, you called this something like `Session.ExtendedState` and marked it as "beta" and then you could feel free to extend it as needed.

tolbertam:
> I'm fine with removing getReqeuestQueueDepth from Session.State for the time being. I think that having an overall view of the number of pending requests will be good enough.

If you are fine with it, I think this is probably the best way to go.  I agree that knowing that there is any queuing happening is valuable enough on its own, and using `MetricRegistry` and hooking into some kind of reporter is easier than introspecting `Session.State`.

> The only other thought that I had would be to register (and unregister as needed) the host state metrics as Gauges in the MetricRegistry in addition to providing the ability to query them via Session.State.

Providing host level metrics is something that we recently discussed (#858).  In the next major version of the java driver, we are introducing metrics at this level ([docs](https://github.com/datastax/java-driver/tree/4.x/manual/core/metrics#metrics)).  The design of how pool selection works has changed in that driver version so we don't have the concept of queuing requests on host pools there.

kevinconaway:
Thanks for your time on this Andrew.  I've updated the PR to remove the metric from `Session.State`

kevinconaway:
> The design of how pool selection works has changed in that driver version so we don't have the concept of queuing requests on host pools there.

As an aside, will you please point me to the docs on how this will work going forward?

tolbertam (approved):
This looks great, thanks 👍 !

tolbertam:
> As an aside, will you please point me to the docs on how this will work going forward?

Sure thing!  The approach with java driver 4.x with regards to how queued up requests is slightly different.  I somewhat muddied the water a bit when I said `we don't have the concept of queueing requests on host pools`, hopefully this clears things up.

While there is no longer a queue on the host pools, you can still use the driver to delay (and therefore queue) requests when there is too much load by using [request throttling](https://github.com/datastax/java-driver/tree/4.x/manual/core/throttling).

When throttling is enabled, you configure a `max-queue-size` which defines how many enqueued requests you may have until you starting receiving `RequestThrottlingException`.  This is similar to the behavior in driver 3.x where one would receive a `BusyPoolException` when you fill the queue, except now this is not done at a `Host` level, but at a `Session` level.

Without throttling enabled, there is no queue, and instead you will see a `AllNodesFailedException` wrapping `BusyConnectionException`s for each host if all connections are out of stream ids.

For monitoring throttling there are three metrics:

* `throttling.delay`: The amount of time requests spent throttled.
* `throttling.queue-size`: Current size of the throttling queue (this is similar to request queue depth)
* `throttling.errors`: Number of times `RequestThrottlingException` was thrown.

kevinconaway:
Sounds great.  

My initial reaction is that throughput will be improved because rather than a request being enqueued on a potentially busy host, as it is now, the pending request could be stolen by a host that is idle, or perhaps less busy.  Is that accurate?

tolbertam:
> My initial reaction is that throughput will be improved because rather than a request being enqueued on a potentially busy host, as it is now, the pending request could be stolen by a host that is idle, or perhaps less busy. Is that accurate?

That is a good assessment 👍 .  Instead of enqueuing a request on a host pool that has no free connections and waiting for one to become available, we can now move on to the next host immediately.  So the new implementation should behave better in scenarios where a host has a bunch of requests piled up.

olim7t (approved):

tolbertam:
Merged, thanks again 👍

PR #1007, opened 2018-05-17
JAVA-1824: Make policies overridable in profiles

olim7t:
For LBP, retry and speculative execution policy, and timestamp generators.

Profiles can have separate configuration. The driver traverses all profiles at startup, and initializes all distinct policies.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigProfile.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigProfile.java
@@ -36,6 +36,22 @@ import java.util.SortedSet;
  * @see DriverConfig
  */
 public interface DriverConfigProfile {
+
+  /**
+   * The name of the default profile (the string {@value}).
+   *
+   * <p>Named profiles can't use this name. If you try to declare such a profile, a runtime error
+   * will be thrown.
+   */
+  String DEFAULT_NAME = "default";
+
+  /**
+   * The name of the profile in the configuration.
+   *
+   * <p>Derived profiles inherit the name of their parent.
+   */
+  String getName();
+
   boolean isDefined(DriverOption option);
 
   boolean getBoolean(DriverOption option);
@@ -66,6 +82,15 @@ public interface DriverConfigProfile {
   Duration getDuration(DriverOption option);
 
   DriverConfigProfile withDuration(DriverOption option, Duration value);
+
+  /**
+   * Returns a representation of all the child options under a given option.
+   *
+   * <p>This is only used to compare configuration sections across profiles, so the actual
+   * implementation does not matter, as long as identical sections (same options with same values,
+   * regardless of order) compare as equal and have the same {@code hashCode()}.
+   */
+  Object getComparisonKey(DriverOption option);
 
   /**
    * Enumerates all the entries in this profile, including those that were inherited from another
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfig.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfig.java
@@ -130,7 +130,7 @@ public class TypesafeDriverConfig implements DriverConfig {
   }
 
   @Override
-  public DriverConfigProfile getNamedProfile(String profileName) {
+  public DriverConfigProfile getProfile(String profileName) {
     Preconditions.checkArgument(
         profiles.containsKey(profileName),
         "Unknown profile '%s'. Check your configuration.",
@@ -139,7 +139,7 @@ public class TypesafeDriverConfig implements DriverConfig {
   }
 
   @Override
-  public Map<String, DriverConfigProfile> getNamedProfiles() {
+  public Map<String, DriverConfigProfile> getProfiles() {
     return ImmutableMap.copyOf(profiles);
   }
 
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfig.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfig.java
@@ -38,7 +38,6 @@ import org.slf4j.LoggerFactory;
 public class TypesafeDriverConfig implements DriverConfig {
 
   private static final Logger LOG = LoggerFactory.getLogger(TypesafeDriverConfig.class);
-  private static final String DEFAULT_PROFILE_KEY = "__default_internal__";
 
   private final Collection<DriverOption> options;
   private final TypesafeDriverConfigProfile.Base defaultProfile;
@@ -52,7 +51,8 @@ public class TypesafeDriverConfig implements DriverConfig {
 
     Map<String, Config> profileConfigs = extractProfiles(config);
     this.defaultProfile =
-        new TypesafeDriverConfigProfile.Base(profileConfigs.get(DEFAULT_PROFILE_KEY));
+        new TypesafeDriverConfigProfile.Base(
+            DriverConfigProfile.DEFAULT_NAME, profileConfigs.get(DriverConfigProfile.DEFAULT_NAME));
 
     if (profileConfigs.size() == 1) {
       this.profiles = Collections.emptyMap();
@@ -61,8 +61,9 @@ public class TypesafeDriverConfig implements DriverConfig {
           ImmutableMap.builder();
       for (Map.Entry<String, Config> entry : profileConfigs.entrySet()) {
         String profileName = entry.getKey();
-        if (!profileName.equals(DEFAULT_PROFILE_KEY)) {
-          builder.put(profileName, new TypesafeDriverConfigProfile.Base(entry.getValue()));
+        if (!profileName.equals(DriverConfigProfile.DEFAULT_NAME)) {
+          builder.put(
+              profileName, new TypesafeDriverConfigProfile.Base(profileName, entry.getValue()));
         }
       }
       this.profiles = builder.build();
@@ -77,11 +78,11 @@ public class TypesafeDriverConfig implements DriverConfig {
       lastLoadedConfig = config;
       try {
         Map<String, Config> profileConfigs = extractProfiles(config);
-        this.defaultProfile.refresh(profileConfigs.get(DEFAULT_PROFILE_KEY));
+        this.defaultProfile.refresh(profileConfigs.get(DriverConfigProfile.DEFAULT_NAME));
         if (profileConfigs.size() > 1) {
           for (Map.Entry<String, Config> entry : profileConfigs.entrySet()) {
             String profileName = entry.getKey();
-            if (!profileName.equals(DEFAULT_PROFILE_KEY)) {
+            if (!profileName.equals(DriverConfigProfile.DEFAULT_NAME)) {
               TypesafeDriverConfigProfile.Base profile = this.profiles.get(profileName);
               if (profile == null) {
                 LOG.warn(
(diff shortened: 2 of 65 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigProfile.java
```diff
@@ -67,6 +83,15 @@
...
   DriverConfigProfile withDuration(DriverOption option, Duration value);
 
+  /**
+   * Returns a representation of all the child options under a given option.
+   *
+   * <p>This is only used to compare configuration sections across profiles, so the actual
+   * implementation does not matter, as long as identical sections (same options with same values,
+   * regardless of order) compare as equal and have the same {@code hashCode()}.
+   */
+  Object getComparisonKey(DriverOption option);
```
olim7t, line 93: Not super happy about this. But we need a way to know when two profiles have the same configuration, in order to create a single policy instance that will serve them both. This also applies to the default profile and its children, because we have no way of knowing which options were inherited or redeclared.

core/src/main/resources/reference.conf
```diff
@@ -122,14 +122,25 @@ datastax-java-driver {
...
   # coordinators, and in which order.
+  # It can be overridden in a profile. Note that the driver creates as few instances as possible:
+  # if a named profile inherits from the default profile, or if two sibling profiles have the
+  # exact same configuration, they will share a single policy instance at runtime.
+  #
+  # If there are multiple load balancing policies in a single driver instance, they work together in
+  # the following way:
+  # - each request gets a query plan from its profile's policy (or the default policy if the request
+  #   has no profile, or the profile does not override the policy).
+  # - when the policies assign distances to nodes, the driver uses the smallest assigned distance
```
tolbertam: Maybe `closest` is a better way of phrasing this instead of `smallest`?   It's more clear to me that LOCAL is "closer" instead of "smaller" than REMOTE for example.

core/src/main/resources/reference.conf
```diff
@@ -122,14 +122,25 @@ datastax-java-driver {
 
   # The policy that decides the "query plan" for each query; that is, which nodes to try as
   # coordinators, and in which order.
+  # It can be overridden in a profile. Note that the driver creates as few instances as possible:
+  # if a named profile inherits from the default profile, or if two sibling profiles have the
+  # exact same configuration, they will share a single policy instance at runtime.
+  #
+  # If there are multiple load balancing policies in a single driver instance, they work together in
```
tolbertam, line 129: Great description!

core/src/main/resources/reference.conf
```diff
@@ -122,14 +122,25 @@ datastax-java-driver {
...
+  #   for any given node.
   load-balancing-policy {
     # The class of the policy. If it is not qualified, the driver assumes that it resides in the
     # package com.datastax.oss.driver.internal.core.loadbalancing.
     #
     # The driver provides a single implementation out of the box: DefaultLoadBalancingPolicy.
     #
     # You can also specify a custom class that implements LoadBalancingPolicy and has a public
-    # constructor with a DriverContext argument.
+    # constructor with two arguments: the DriverContext and a String representing the profile
```
tolbertam, line 142: I was thinking at first that requiring this constructor signature was be too restrictive...however you could add additional settings to a a profile or context, so it makes sense.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfig.java
```diff
@@ -130,7 +113,7 @@ public boolean reload(Config config) {
     if (rootObject.containsKey("profiles") && rootObject.get("profiles").valueType() == OBJECT) {
       ConfigObject profilesObject = (ConfigObject) rootObject.get("profiles");
       for (String profileName : profilesObject.keySet()) {
-        if (profileName.equals(DEFAULT_PROFILE_KEY)) {
+        if (profileName.equals(DriverConfigProfile.DEFAULT_NAME)) {
```
tolbertam, line 116: Was thinking about this exact case "What if a user actually defined a profile named default?"  Glad this is covered here.

manual/core/load_balancing/README.md
```diff
@@ -212,6 +212,43 @@ You can use your own implementation by specifying its fully-qualified name in th
...
+The `custom-lbp` profile uses a dedicated policy. The `slow` profile inherits the default profile's.
+Note that this goes beyond configuration inheritance: the driver only creates a single
+`DefaultLoadBalancingPolicy` instance and reuses it (this also occurs if two sibling profiles have
+the same configuration).
+
+For query plans, each request uses its declared profile's policy. If it doesn't declare any profile,
+or if the profile doesn't have a dedicated policy, then the default profile's policy is used.
+
+For node distances, the driver remembers the last distance suggested by each policy for each node.
+Then it uses the "shortest" distance for any given node. For example:
```
tolbertam: Similar comment here about shortest vs. closest.

tolbertam (commented):
Very well done!  It looks like this brings us to parity with the nodejs and python driver implementations and takes it even further.  i.e. you can't define speculative execution or timestamp generation on either, but you probably should be able to (especially spec ex).   I think I need to play with this a bit before I formally approve it, but I think this looks very good.  I iterated through our configuration to see if there is anything else we could tie to a profile.  The only things I could see possibly being requested:

* throttling:  What if I wanted to throttle requests by profile? (i.e. only allow 1000 requests per sec for profile X, 5000 for profile Y and so on)

* query logging:  what might be considered a "slow query" may vary from profile to profile.

* metrics:  a mechanism to categorize metrics by profile.  "How many requests are being sent by profile x, profile y?" and so on.

In each case, i see some value, but I'm not sure if it's worth the implementation and configuration complexity.

I'll play with this monday and if all goes well 👍 it.

core/src/main/resources/reference.conf
olim7t: Yeah, I asked myself the same thing while writing it. Weirdly, "smaller" is the most logical (in its original definition, a distance is a quantity), but I agree that "closer" might be easier to understand 😄

core/src/main/resources/reference.conf
olim7t: And it's how the implementation will retrieve the configuration options, see `ConstantSpecExPolicy` for example.

olim7t:
Timestamp generation is a bit of a stretch, but I decided to include it because it was cheap once the mechanism was in place.
Throttling and query logging fall into the same category, they can definitely be added as well.

But let's not forget that you can always write a custom policy that tests the request's profile name too, so we're not completely closing the door if we don't do it. Maybe we should stick to just LBP, retries and specex.

As for the metrics, I find the idea compelling on paper, but indeed it adds a dimension in the metric space and will probably complicate things. I'll take a look.

csplinter:
My opinion on the additional options ( throttling, query logging, and metrics ) - I would like to keep this to the minimal set at first ( I also question whether we need timestamp generation here ). We can always add these later if we see use cases requiring them.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -64,8 +64,11 @@
   private volatile DistanceReporter distanceReporter;
   @VisibleForTesting volatile String localDc;
 
-  public DefaultLoadBalancingPolicy(DriverContext context) {
-    this(getLocalDcFromConfig(context), getFilterFromConfig(context), context);
+  public DefaultLoadBalancingPolicy(DriverContext context, String profileName) {
```
tolbertam, line 68: Small slightly unrelated issue with `DefaultLoadBalancingPolicy`, the logic at line 113 assumes that if no contact points are provided that the localDC be inferred from 127.0.0.1.  We should probably get rid of that logic if the localDC is explicitly set.  In my case I had a default profile using `dc1`, and another profile using `dc2`, and in both cases `dc1` was chosen because no contact points were provided.

tolbertam: To clarify, i'm suggesting changing:

```java
    if (contactPoints.isEmpty()) {
```

to

```java
    if (localDc == null && contactPoints.isEmpty()) {
```

manual/core/load_balancing/README.md
```diff
@@ -212,6 +212,43 @@ You can use your own implementation by specifying its fully-qualified name in th
...
+
+```
+datastax-java-driver {
+  load-balancing-policy {
+    class = DefaultLoadBalancingPolicy
+  }
+  profiles {
+    custom-lbp {
+      load-balancing-policy {
+        class = CustomLoadBalancingPolicy
```
tolbertam, line 227: One thing I found really nice about this was that if all I wanted to do was override a lbp setting and still use `DefaultLoadBalancingPolicy`, I didn't actually have to specify the class name, just the name of the setting I wanted to change.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
tolbertam: went ahead and pushed d62853d which does that

tolbertam:
I added a couple integration tests and also plan on adding some more validation on the speculative execution tests and a LBP specific test soon.

tolbertam:
27977c4: Added some additional validation in `DefaultLoadBalancingPolicy` to ensure we don't log

> [{}] You specified {} as the local DC, but some contact points are from a different DC ({})

when a profile-based LBP is used.

Also added some profile tests for load balancing.

I'm 👍 on this PR, but I know you mentioned there was some additional work you were doing as it pertained to passing around profile information to metrics, so will hold my PR approval until then.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
olim7t: Agreed 👍 If there are no contact points, the local DC is optional, but it can still be specified.
The code can be improved further for readability because we check `localDc!=null` in the next branch, I'm pushing a small commit to change it.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/loadbalancing/PerProfileLoadBalancingPolicyIT.java
```diff
@@ -0,0 +1,134 @@
...
+import com.datastax.oss.driver.internal.core.loadbalancing.DefaultLoadBalancingPolicy;
+import com.datastax.oss.simulacron.common.cluster.ClusterSpec;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(ParallelizableTests.class)
+public class PerProfileLoadBalancingPolicyIT {
```
olim7t, line 40: Great job on those integration tests 👍

olim7t:
Rebased and force-pushed.
The Travis build fails because we depend on a native-protocol snapshot since JAVA-1592, but I've run the full integration suite locally.
The last commit passes profile information to the metric updaters when relevant/practical. The default updaters ignore it at this point, but that leaves a door open if we decide to do per-profile metrics later.

tolbertam (approved):
:+1: to the recent changes, and 👍 overall!  :shipit:

PR #1008, opened 2018-05-17
JAVA-1829: Add metrics for bytes-sent and bytes-received

tolbertam:
Motivation:

JAVA-708 added metrics to java driver 3.x for tracking the number of
bytes added and received for a Cluster instance.  For feature parity,
these metrics should also be tracked at a Session and Node level in java
driver 4.x.

Modifications:

- Add 'pools.bytes-sent' and 'pools-bytes-received' to node-level
  metrics and 'bytes-sent' and 'bytes-received' to session-level
  metrics.

- Add InboundTrafficMeter and OutboundTrafficMeter to pipeline in
  ChannelFactory.initializer.

- Update ChannelFactory.connect methods to accept Node instead of
  SocketAddress.  Previous connect method retained for unit testing.

- Updated unit tests to reason with Node instead of SocketAddress where
  connect is involved and a Node instance is available.

- Added static singleton instance of NoopNodeMetricUpdater.

Result:

Additional metrics added for tracking bytes sent and received at both
Session and Node level.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
@@ -19,8 +19,15 @@ import com.datastax.oss.driver.api.core.ProtocolVersion;
 import com.datastax.oss.driver.api.core.UnsupportedProtocolVersionException;
 import com.datastax.oss.driver.api.core.config.DefaultDriverOption;
 import com.datastax.oss.driver.api.core.config.DriverConfigProfile;
+import com.datastax.oss.driver.api.core.metadata.Node;
+import com.datastax.oss.driver.api.core.metrics.DefaultNodeMetric;
+import com.datastax.oss.driver.api.core.metrics.DefaultSessionMetric;
 import com.datastax.oss.driver.internal.core.context.InternalDriverContext;
 import com.datastax.oss.driver.internal.core.context.NettyOptions;
+import com.datastax.oss.driver.internal.core.metadata.DefaultNode;
+import com.datastax.oss.driver.internal.core.metrics.NodeMetricUpdater;
+import com.datastax.oss.driver.internal.core.metrics.NoopNodeMetricUpdater;
+import com.datastax.oss.driver.internal.core.metrics.SessionMetricUpdater;
 import com.datastax.oss.driver.internal.core.protocol.FrameDecoder;
 import com.datastax.oss.driver.internal.core.protocol.FrameEncoder;
 import com.datastax.oss.driver.shaded.guava.common.annotations.VisibleForTesting;
@@ -84,8 +91,19 @@ public class ChannelFactory {
     this.protocolVersion = newVersion;
   }
 
-  public CompletionStage<DriverChannel> connect(
-      final SocketAddress address, DriverChannelOptions options) {
+  public CompletionStage<DriverChannel> connect(Node node, DriverChannelOptions options) {
+    NodeMetricUpdater nodeMetricUpdater;
+    if (node instanceof DefaultNode) {
+      nodeMetricUpdater = ((DefaultNode) node).getMetricUpdater();
+    } else {
+      nodeMetricUpdater = NoopNodeMetricUpdater.INSTANCE;
+    }
+    return connect(node.getConnectAddress(), options, nodeMetricUpdater);
+  }
+
+  @VisibleForTesting
+  CompletionStage<DriverChannel> connect(
+      SocketAddress address, DriverChannelOptions options, NodeMetricUpdater nodeMetricUpdater) {
     CompletableFuture<DriverChannel> resultFuture = new CompletableFuture<>();
 
     ProtocolVersion currentVersion;
@@ -99,14 +117,22 @@ public class ChannelFactory {
       isNegotiating = true;
     }
 
-    connect(address, options, currentVersion, isNegotiating, attemptedVersions, resultFuture);
+    connect(
+        address,
+        options,
+        nodeMetricUpdater,
+        currentVersion,
+        isNegotiating,
+        attemptedVersions,
+        resultFuture);
     return resultFuture;
   }
 
   private void connect(
       SocketAddress address,
       DriverChannelOptions options,
-      final ProtocolVersion currentVersion,
+      NodeMetricUpdater nodeMetricUpdater,
+      ProtocolVersion currentVersion,
       boolean isNegotiating,
       List<ProtocolVersion> attemptedVersions,
       CompletableFuture<DriverChannel> resultFuture) {
@@ -118,7 +144,8 @@ public class ChannelFactory {
             .group(nettyOptions.ioEventLoopGroup())
             .channel(nettyOptions.channelClass())
             .option(ChannelOption.ALLOCATOR, nettyOptions.allocator())
-            .handler(initializer(address, currentVersion, options, resultFuture));
+            .handler(
+                initializer(address, currentVersion, options, nodeMetricUpdater, resultFuture));
 
     DriverConfigProfile config = context.config().getDefaultProfile();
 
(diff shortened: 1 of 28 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
```diff
@@ -84,8 +88,18 @@ public void setProtocolVersion(ProtocolVersion newVersion) {
...
   }
 
-  public CompletionStage<DriverChannel> connect(
+  @VisibleForTesting
+  CompletionStage<DriverChannel> connect(
       final SocketAddress address, DriverChannelOptions options) {
+    return connect(null, address, options);
+  }
+
+  public CompletionStage<DriverChannel> connect(final Node node, DriverChannelOptions options) {
```
tolbertam: Needed a way to get at the `NodeMetricUpdater` instance,  couldn't find a means other than passing in the `Node`.  I figured this was ok since outside of unit tests we have access to the `Node` instance, but if there is a better way of getting at the `NodeMetricUpdater` instance might be nice.   This change triggered a lot of unit test code changes, which is why there are a bunch of changes in this PR that don't look related.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
```diff
@@ -236,7 +259,22 @@ protected void initChannel(Channel channel) {
               .sslHandlerFactory()
               .map(f -> f.newSslHandler(channel, address))
               .map(h -> pipeline.addLast("ssl", h));
+
+          NodeMetricUpdater nodeMetricUpdater;
+          if (node instanceof DefaultNode) {
```
tolbertam: node could also be null (particularly in unit tests), `instanceof` check covers that for us.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
```diff
@@ -236,7 +259,22 @@ protected void initChannel(Channel channel) {
...
+            nodeMetricUpdater = ((DefaultNode) node).getMetricUpdater();
+          } else {
+            nodeMetricUpdater = NoopNodeMetricUpdater.INSTANCE;
+          }
           pipeline
+              .addLast(
```
tolbertam: driver 3.x selectively excludes initializing these if metrics are disabled, could do something similar here but wasn't sure if it was worth it (and also there isn't a general flag I can check).

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/NoopNodeMetricUpdater.java
```diff
@@ -22,6 +22,10 @@
 @ThreadSafe
 public class NoopNodeMetricUpdater implements NodeMetricUpdater {
 
+  public static NoopNodeMetricUpdater INSTANCE = new NoopNodeMetricUpdater();
```
tolbertam, line 25: perhaps this is not necessary, however previously we were creating 1 of these per node (if metric registry was null).  With the change to create this in `ChannelFactory` we'd create one per connection so thought it might be good to make this a singleton.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
```diff
@@ -99,11 +113,12 @@ public void setProtocolVersion(ProtocolVersion newVersion) {
       isNegotiating = true;
     }
 
-    connect(address, options, currentVersion, isNegotiating, attemptedVersions, resultFuture);
+    connect(node, address, options, currentVersion, isNegotiating, attemptedVersions, resultFuture);
     return resultFuture;
   }
 
   private void connect(
+      Node node,
```
newkek: Do we need both `node` and `address`?

newkek: Do we need both `node` and `address`? Could we just use node all around? Also I saw that except for tests, the `ChannelFactory.connect(SocketAddress address, options)` was only used once and the method is called with `channelFactory.connect(node.getConnectAddress(), ..)` in https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/pool/ChannelPool.java#L278 , so maybe we can also remove the `ChannelFactory.connect(SocketAddress address, options)` now too?

newkek (approved):
Had a minor comment about the connect() method, otherwise sounds good to me. _mini 👍 _ cause I may be missing stuff specific driver 4 :)

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
newkek: Oh I missed the "previous method retained for unit testing" in the initial comment. Not sure it is worth to keep this public method out there just for unit testing, but since it's the internal package, I guess it's debatable. Not a strong opinion but also as mentioned above the private method could maybe be simplified too.

tolbertam: There might be a way for me to adjust the existing unit tests that use an address to use `Node` instead.   The problem was that they used a `LocalAddress` instead of `SocketAddress` so I couldn't simply wrap it in a mock `Node`, i'll look into it more though.

tolbertam: Hrmm, looking at it again, I think this is a decent compromise.  `connect(SocketAddress address, DriverChannelOptions options)` uses default visibility (not public) and the class is in the internal package structure.  The tests that use `LocalAddress` use it for a reason so I think it may be better to leave them that way.

newkek: Good point I hadn't seen that the visibility of `connect(SocketAddress, options)` had changed from public to package-local, sgtm for that then.

olim7t: 👍 
Indeed the factory used a raw `SocketAddress` to allow testing with `LocalAddress` (also, `Node` did not exist yet at the time it was written).
The way you addressed that is fine IMHO. Alternatively, you could have only two methods: the public one that takes a `Node`, and a package-private one that takes a `SocketAddress` and the two metric updaters. This way you could cover the invocation of the metric updaters from unit tests.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
olim7t: We could add `MetricUpdater.isEnabled(MetricT)`. It's trivial to implement, the updaters already need this check internally.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/NoopNodeMetricUpdater.java
olim7t: 👍 wfm

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
olim7t: On a side note, if it's not null it's always an instance of `DefaultNode`. That is not meant to be customizable, the interface/implementation separation is just for hiding stuff from the public API.

olim7t (approved):
LGTM, but see my comments inline.

tolbertam:
Thanks for reviewing, I've pushed a commit based on feedback and if that looks good this should be ready to go 👍

olim7t:
I did a small change to use the noop reporter from tests, felt cleaner than allowing null. I realised that my idea of testing the metrics in unit tests doesn't work, because they don't encode all the way to bytes.

This is good to go, I'll rebase and merge tomorrow.

olim7t:
I removed the `pool` prefix for node-level metrics.

PR #1009, opened 2018-05-25
JAVA-1846 Add ConstantReconnectionPolicy

GregBestland:


olim7t (approved):
👍 
I changed the implementation slightly: we can build the schedule at construction time and always return the same instance.

PR #1010, opened 2018-05-31
JAVA-1845: Include reference.conf in the manual

olim7t:
Prerequisites for local generation:
```
pip install docutils
pip install pygments
```

### DIFF
--- a/docs.yaml
+++ b/docs.yaml
@@ -8,6 +8,11 @@ sections:
     sources:
       - type:  markdown
         files: 'manual/**/*.md'
+  - title:     Reference configuration
+    prefix:    /manual/core/configuration/reference
+    sources:
+      - type: rst
+        files: 'manual/core/configuration/reference/*.rst'
   - title:     Changelog
     prefix:    /changelog
     sources:
--- a/manual/core/README.md
+++ b/manual/core/README.md
@@ -50,8 +50,8 @@ variants that return a `CompletionStage`).
 customization is done through the driver configuration (refer to the
 [corresponding section](configuration/) of this manual for full details).
 
-We recommend that you take a look at the `reference.conf` file bundled with the driver for the list
-of available options, and cross-reference with the sub-sections in this manual for more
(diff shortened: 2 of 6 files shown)


docs.yaml
```diff
@@ -18,6 +18,11 @@ sections:
     sources:
       - type:  markdown
         files: 'upgrade_guide/**/*.md'
+  - title:     Reference configuration
+    prefix:    /reference_conf
+    sources:
+      - type:  rst
+        files: 'reference_conf/**/*.md'
```
olim7t: For some reason this doesn't work if the file uses the extension `rst`, I'll look into it further tomorrow.

reference_conf/README.md
```diff
@@ -0,0 +1,12 @@
+Reference configuration
+-----------------------
```
olim7t: This is reStructuredText, I had to resort to that since markdown doesn't support file inclusions.
As explained in my `docs.yaml` comment, this needs to use the extension `.md` right now (probably a bug), which means it's not rendered correctly on GitHub.

I hope the include will work on GitHub once we use the right extension, but at worst it's not hard to figure it out.

reference_conf/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+The following is a copy of the ``reference.conf`` file matching the version of this documentation.
+It is packaged in the ``java-driver-core`` JAR artifact, and used at runtime to provide the default
+values for all configuration options (in the sources, it can be found under
+``core/src/main/resources``).
+
+See the `configuration page <../manual/core/configuration/>`_ for more explanations.
+
+.. include:: core/src/main/resources/reference.conf
+   :code: properties
```
olim7t: `properties` seems to be the closest syntax for HOCON, as a bare minimum comments should be highlighted.

The code in the generated HTML is correctly lexed into various `span`s, but there are no visible styles, I think we need to add a CSS file.

tolbertam (approved):
👍

PR #1011, opened 2018-05-31
JAVA-1815: Reorganize configuration into basic/advanced categories

olim7t:
Please no debates about what should or should not be considered basic. @csplinter and I agreed on the current layout.

TODO:
- [x] update example in address translator EC2 section (JAVA-1832)

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
@@ -149,31 +149,30 @@ public class ChannelFactory {
 
     DriverConfigProfile config = context.config().getDefaultProfile();
 
-    boolean tcpNoDelay = config.getBoolean(DefaultDriverOption.CONNECTION_SOCKET_TCP_NODELAY);
+    boolean tcpNoDelay = config.getBoolean(DefaultDriverOption.SOCKET_TCP_NODELAY);
     bootstrap = bootstrap.option(ChannelOption.TCP_NODELAY, tcpNoDelay);
-    if (config.isDefined(DefaultDriverOption.CONNECTION_SOCKET_KEEP_ALIVE)) {
-      boolean keepAlive = config.getBoolean(DefaultDriverOption.CONNECTION_SOCKET_KEEP_ALIVE);
+    if (config.isDefined(DefaultDriverOption.SOCKET_KEEP_ALIVE)) {
+      boolean keepAlive = config.getBoolean(DefaultDriverOption.SOCKET_KEEP_ALIVE);
       bootstrap = bootstrap.option(ChannelOption.SO_KEEPALIVE, keepAlive);
     }
-    if (config.isDefined(DefaultDriverOption.CONNECTION_SOCKET_REUSE_ADDRESS)) {
-      boolean reuseAddress = config.getBoolean(DefaultDriverOption.CONNECTION_SOCKET_REUSE_ADDRESS);
+    if (config.isDefined(DefaultDriverOption.SOCKET_REUSE_ADDRESS)) {
+      boolean reuseAddress = config.getBoolean(DefaultDriverOption.SOCKET_REUSE_ADDRESS);
       bootstrap = bootstrap.option(ChannelOption.SO_REUSEADDR, reuseAddress);
     }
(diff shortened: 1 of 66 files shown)


core/src/main/resources/reference.conf
```diff
@@ -107,40 +57,101 @@ datastax-java-driver {
...
   # To disable periodic reloading, set this to 0.
-  config-reload-interval = 5 minutes
+  #
+  # Required: yes (unless you pass a different ConfigLoader to the session builder).
+  # Modifiable at runtime: yes, the new value will be used after the next time the configuration
+  #   gets reloaded.
+  # Overridable in a profile: no
+  basic.config-reload-interval = 5 minutes
+
+  basic.request {
```
olim7t, line 81: While working on #1010 I noticed that it was hard to figure out the level of nesting when you're in the middle of the file. So instead of nesting everything into `basic { ... }`, I repeat the prefix for each section. Also, I reduced the level of nesting in some places (e.g. move `control-connection` out of `connection`).

tolbertam: This was my first question as it initially seems odd.  It does make it more obvious so 👍 on doing it this way.

olim7t: We're spoiled in IDEA because it recognizes HOCON and displays a breadcrumbs-style widget when the caret is on an option.
But I think users will most often look at this file in a less friendly environment, such as GitHub web or the HTML documentation.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
```diff
@@ -149,31 +149,30 @@ private void connect(
 
     DriverConfigProfile config = context.config().getDefaultProfile();
 
-    boolean tcpNoDelay = config.getBoolean(DefaultDriverOption.CONNECTION_SOCKET_TCP_NODELAY);
```
tolbertam, line 152: Noticed that some settings were moved up from `connection`, but others weren't, what was the thinking there?  Just to keep the full config path more brief or just to separate them from the rest of the connection options?  👍 in any case :)

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
```diff
@@ -314,7 +314,7 @@ protected EventBus buildEventBus() {
     return (Compressor<ByteBuf>)
         Reflection.buildFromConfig(
                 this,
-                DefaultDriverOption.PROTOCOL_COMPRESSOR_CLASS,
+                DefaultDriverOption.COMPRESSOR_CLASS,
```
tolbertam: Another case where a section (`protocol`) was removed for one option, but not others.  Although this really makes sense as compression isn't really tied to the protocol from user's perspective.

core/src/main/resources/reference.conf
```diff
@@ -216,625 +255,463 @@ datastax-java-driver {
...
+  # The policy that controls if the driver pre-emptively tries other nodes if a node takes too long
+  # to respond.
+  #
+  # Required: yes
+  # Modifiable at runtime: no (but custom implementations may elect to watch configuration changes
+  #   and allow child options to be changed at runtime).
+  # Overridable in a profile: yes. Note that the driver creates as few instances as possible: if a
+  #   named profile inherits from the default profile, or if two sibling profiles have the exact
+  #   same configuration, they will share a single policy instance at runtime.
+  advanced.speculative-execution-policy {
```
tolbertam, line 324: Makes sense to de-nest this from `request` 👍

tolbertam (approved):
lgtm 👍 !  Tests all seemed to run good for me locally so must have taken care of making all the appropriate changes there too.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
olim7t: I moved `socket` out of `connection` to reduce the levels of nesting. `connection` is now only about the pooling / stream management aspects.

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
olim7t: I moved up all "policies" to the top level.

PR #1013, opened 2018-06-01
JAVA-1825: Add remaining Typesafe config primitive types to DriverConfigProfile

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigProfile.java
```diff
@@ -79,10 +103,21 @@
 
   DriverConfigProfile withBytes(DriverOption option, long value);
 
+  List<Long> getBytesList(DriverOption option);
```
tolbertam: This confused me initially (`bytes` represented as `long`).  But i see the comment on `getBytes` represents a size in bytes, not the `byte` type.

tolbertam (approved):
lgtm 👍

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigProfile.java
olim7t: I added a link to `getBytes`, and more explanations there.

PR #1015, opened 2018-06-01
Clone and build native protocol in ci build

tolbertam:
There are times where we are developing against an unreleased version of native protocol.  Due to this, our CI builds may fail as this version is not in maven central yet.   To fix this, simply clone and build native-protocol as part of our CI processes.


olim7t:
🎉

PR #1016, opened 2018-06-03
JAVA-1532: Add Codec support for Java 8's LocalDateTime and ZoneId

jubobs:
Better late than never -_-'

### DIFF
--- a/driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/LocalDateTimeCodec.java
+++ b/driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/LocalDateTimeCodec.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.driver.extras.codecs.jdk8;
+
+import com.datastax.driver.core.*;
+import com.datastax.driver.core.exceptions.InvalidTypeException;
+
+import java.nio.ByteBuffer;
+
+import static com.datastax.driver.core.ParseUtils.isLongLiteral;
+import static com.datastax.driver.core.ParseUtils.quote;
+
+/**
+ * {@link TypeCodec} that maps
+ * {@link java.time.LocalDateTime} to CQL {@code timestamp},
+ * allowing the setting and retrieval of {@code timestamp}
+ * columns as {@link java.time.LocalDateTime} instances.
+ * <p>
+ * <strong>IMPORTANT</strong>
+ * <p>
+ * 1) The default timestamp formatter used by this codec produces CQL literals
+ * that may include milliseconds.
+ * <strong>This literal format is incompatible with Cassandra < 2.0.9.</strong>
+ * <p>
+ * 2) Even if the ISO-8601 standard accepts timestamps with nanosecond precision,
+ * Cassandra timestamps have millisecond precision; therefore, any sub-millisecond
+ * value set on a {@link java.time.LocalDateTime} will be lost when persisted to Cassandra.
+ *
+ * @see <a href="https://cassandra.apache.org/doc/cql3/CQL-2.2.html#usingtimestamps">'Working with timestamps' section of CQL specification</a>
+ */
+@IgnoreJDK6Requirement
+@SuppressWarnings("Since15")
+public class LocalDateTimeCodec extends TypeCodec<java.time.LocalDateTime> {
+
+    private static final java.time.format.DateTimeFormatter FORMATTER = java.time.format.DateTimeFormatter.ISO_LOCAL_DATE_TIME;
+
+    public static final LocalDateTimeCodec instance = new LocalDateTimeCodec();
+
+    private LocalDateTimeCodec() {
+        super(DataType.timestamp(), java.time.LocalDateTime.class);
+    }
+
+    @Override
+    public ByteBuffer serialize(java.time.LocalDateTime value, ProtocolVersion protocolVersion) {
+        if (value == null) {
+            return null;
+        }
+        long millis = value.atZone(java.time.ZoneOffset.UTC).toInstant().toEpochMilli();
+        return bigint().serializeNoBoxing(millis, protocolVersion);
+    }
+
+    @Override
+    public java.time.LocalDateTime deserialize(ByteBuffer bytes, ProtocolVersion protocolVersion) {
+        if (bytes == null || bytes.remaining() == 0) {
+            return null;
+        }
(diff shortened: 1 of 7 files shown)


clirr-ignores.xml
```diff
@@ -44,6 +44,12 @@
         <justification>This class is only present if the project was compiled with JDK 8+</justification>
     </difference>
 
+    <difference>
```
adutra: This seems a bit odd, you shouldn't see Clirr warnings when you add new classes. I don't think you need these exceptions.

driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/LocalDateTimeCodec.java
```diff
@@ -0,0 +1,105 @@
...
+ */
+@IgnoreJDK6Requirement
+@SuppressWarnings("Since15")
+public class LocalDateTimeCodec extends TypeCodec<java.time.LocalDateTime> {
+
+    private static final java.time.format.DateTimeFormatter FORMATTER = java.time.format.DateTimeFormatter.ISO_LOCAL_DATE_TIME;
+
+    public static final LocalDateTimeCodec instance = new LocalDateTimeCodec();
+
+    private LocalDateTimeCodec() {
```
adutra, line 52: The code is nice, but very similar to `InstantCodec` – why don't you delegate to it instead?

driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/LocalDateTimeCodec.java
```diff
@@ -0,0 +1,105 @@
...
+    public java.time.LocalDateTime deserialize(ByteBuffer bytes, ProtocolVersion protocolVersion) {
+        if (bytes == null || bytes.remaining() == 0) {
+            return null;
+        }
+        long millis = bigint().deserializeNoBoxing(bytes, protocolVersion);
+        return java.time.LocalDateTime.ofInstant(java.time.Instant.ofEpochMilli(millis), java.time.ZoneOffset.UTC);
+    }
+
+    @Override
+    public String format(java.time.LocalDateTime value) {
```
adutra, line 75: Parse() and format() must produce valid CQL literals. A local date-time is not a valid CQL literal. What you need here is to produce output identical to `InstantCodec`, using UTC. Which is another good reason to re-use `InstantCodec` :)

driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/ZoneIdCodec.java
```diff
@@ -0,0 +1,83 @@
...
+    @Override
+    public java.time.ZoneId deserialize(ByteBuffer bytes, ProtocolVersion protocolVersion) {
+        if (bytes == null || bytes.remaining() == 0) {
+            return null;
+        }
+        return java.time.ZoneId.of(varchar().deserialize(bytes, protocolVersion));
+    }
+
+    @Override
+    public String format(java.time.ZoneId value) {
```
adutra, line 57: Again, here you could leverage `VarcharCodec` to do the actual parse/format just as you did for serialize/deserialize.

adutra (changes requested):
Good work, thanks for contributing!

driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/LocalDateTimeCodec.java
adutra: After giving this some more thought, I think you can leave it that way. Local date-times in ISO-8601 _are_ valid CQL timestamp literals after all, the only drawback is if they were interpreted by C* as full timestamps it would use its own default time zone, and not necessarily UTC.

driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/LocalDateTimeCodec.java
jubobs: Does your [earlier comment](https://github.com/datastax/java-driver/pull/1016#discussion_r192770854) make this one obsolete?

clirr-ignores.xml
jubobs: I got rid of those changes.

jubobs:
@adutra Hmmm... The [AppVeyor build](https://ci.appveyor.com/project/DataStax/java-driver/build/2201.3.x/job/e6vt7qtuj0uj58qt) seems to be in trouble:

```
2063   [main] INFO  com.datastax.driver.core.CCMBridge - Tests requiring CCM will by default use Cassandra version 3.11.0 (install arguments: [-v 3.11.0])
2078   [main] INFO  com.datastax.driver.core.CCMCache - Maximum number of running CCM nodes: 1
7469   [Exec Stream Pumper] ERROR com.datastax.driver.core.CCMBridge - ccmerr> Started: node1 with pid: 2924
```

Any guidance about that?

Edit: The AppVeyor build appears to have failed in the past for PRs that nonetheless got merged. That's confusing, esp. for first-time contributors like myself. I couldn't find any Jira ticket to address those flaky builds, but perhaps there should be one...

driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/ZoneIdCodec.java
jubobs: Done.

jubobs:
@adutra I've made the requested changes and rebase my branch to `datastax:3.x`. Keep me posted.

driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/LocalDateTimeCodec.java
adutra: Yes. Sorry for the noise.

adutra (approved):
Nice work thanks. We now need to figure the timeline, this could certainly make it into the next minor release.

jubobs:
Great! Thanks.

driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/LocalDateTimeCodec.java
```diff
@@ -0,0 +1,105 @@
...
+    private LocalDateTimeCodec() {
+        super(DataType.timestamp(), java.time.LocalDateTime.class);
+    }
+
+    @Override
+    public ByteBuffer serialize(java.time.LocalDateTime value, ProtocolVersion protocolVersion) {
+        if (value == null) {
+            return null;
+        }
+        long millis = value.atZone(java.time.ZoneOffset.UTC).toInstant().toEpochMilli();
```
olim7t, line 61: Would it make sense to make the `ZoneOffset` a final field, and expose a constructor that allows to change it? That way you could build a codec instance that matches your server's timezone.

olim7t (approved):
LGTM, but one small suggestion to make the time zone customizable in the `LocalDateTime` codec.

driver-extras/src/main/java/com/datastax/driver/extras/codecs/jdk8/LocalDateTimeCodec.java
tolbertam: Since `LocalDateTime` is meant to be a description of time, and not an actual zoned instant of time, I think it makes sense to make it completely devoid of time zone information and use UTC internally like it currently does.

tolbertam (approved):
Excellent work :+1:

tolbertam:
Rebased on 3.x and updated changelog, will merge on ci pass.

PR #1018, opened 2018-06-06
JAVA-1853: Add newValue(Object...) to TupleType and UserDefinedType

olim7t:
Requires `codecFor(cqlType, value)` (1st commit) to properly handle custom codecs. This was also wrong in the existing bound statement method (2nd commit).

TODO:
- [x] JAVA-1863

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -4,6 +4,7 @@
 
 ### 4.0.0-alpha4 (in progress)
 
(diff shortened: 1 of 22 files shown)


tolbertam (approved):
Looks good!  I added a small commit for `DataTypeIT` to add samples for UDTs & Tuples created using `newValue`.

PR #1019, opened 2018-06-07
JAVA-1761: Add OSGi descriptors and JAVA-1762: Shading

tolbertam:
Motivation:

To enable the driver to be used in OSGi containers.

Modifications:

Use maven-bundle-plugin to add OSGi information to the generated
MANIFEST.MF.

Add jctools as a dependency instead of using netty-shaded
MpscLinkedAtomicQueue in DefaultWriteCoalescer.  This was required
because netty does not export its shaded jctools packages in its
Export-Package definition.

Change Reflection.buildFromConfig to use the expectedSuperType's
ClassLoader instead of the current Thread's ClassLoader when loading a
class.  This enables the driver to load classes from the core bundle in
OSGi, but may be problematic in other situations so this may be an
incomplete solution.

Add OSGi integration tests.

Make all dependency versions properties so they may be passed to
failsafe plugin for use in OSGi tests.

Result:

The driver now supports OSGi.

### DIFF
--- a/core-shaded/pom.xml
+++ b/core-shaded/pom.xml
@@ -1,0 +1,128 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright DataStax, Inc.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.datastax.oss</groupId>
+    <artifactId>java-driver-parent</artifactId>
+    <version>4.0.0-alpha4-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>java-driver-core-shaded</artifactId>
+  <packaging>bundle</packaging>
+
+  <name>DataStax Java driver for Apache Cassandra(R) - core with netty shaded</name>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.datastax.oss</groupId>
+      <artifactId>java-driver-core</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <!-- extract shaded manifest from core jar -->
+      <plugin>
+        <artifactId>maven-dependency-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>unpack-manifest</id>
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>unpack</goal>
+            </goals>
+            <configuration>
+              <artifactItems>
+                <artifactItem>
+                  <groupId>com.datastax.oss</groupId>
+                  <artifactId>java-driver-core</artifactId>
+                  <version>${project.version}</version>
+                  <includes>META-INF-shaded/MANIFEST.MF</includes>
+                  <outputDirectory>${project.build.directory}</outputDirectory>
+                </artifactItem>
+              </artifactItems>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <artifactId>maven-shade-plugin</artifactId>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <artifactSet>
+                <includes>
+                  <include>com.datastax.oss:java-driver-core</include>
+                  <include>io.netty:*</include>
+                </includes>
+              </artifactSet>
+              <relocations>
+                <relocation>
+                  <pattern>io.netty</pattern>
+                  <shadedPattern>com.datastax.oss.driver.shaded.netty</shadedPattern>
+                </relocation>
+              </relocations>
+              <transformers>
+                <transformer implementation="org.apache.maven.plugins.shade.resource.DontIncludeResourceTransformer">
+                  <resources>
+                    <resource>META-INF/MANIFEST.MF</resource>
+                    <resource>META-INF-shaded/MANIFEST.MF</resource>
+                    <resource>META-INF/maven/com.datastax.oss/java-driver-core/pom.xml</resource>
+                    <resource>META-INF/maven/com.datastax.oss/java-driver-core/pom.properties</resource>
+                    <resource>META-INF/maven/io.netty/netty-handler/pom.properties</resource>
+                    <resource>META-INF/maven/io.netty/netty-handler/pom.xml</resource>
+                    <resource>META-INF/maven/io.netty/netty-buffer/pom.properties</resource>
+                    <resource>META-INF/maven/io.netty/netty-buffer/pom.xml</resource>
+                    <resource>META-INF/maven/io.netty/netty-common/pom.properties</resource>
+                    <resource>META-INF/maven/io.netty/netty-common/pom.xml</resource>
+                    <resource>META-INF/maven/io.netty/netty-transport/pom.properties</resource>
+                    <resource>META-INF/maven/io.netty/netty-transport/pom.xml</resource>
+                    <resource>META-INF/maven/io.netty/netty-resolver/pom.properties</resource>
+                    <resource>META-INF/maven/io.netty/netty-resolver/pom.xml</resource>
+                    <resource>META-INF/maven/io.netty/netty-codec/pom.properties</resource>
+                    <resource>META-INF/maven/io.netty/netty-codec/pom.xml</resource>
+                    <!-- netty's shading of jctools does not remove its pom files -->
+                    <resource>META-INF/maven/org.jctools/jctools-core/pom.properties</resource>
+                    <resource>META-INF/maven/org.jctools/jctools-core/pom.xml</resource>
+                  </resources>
+                </transformer>
+                <!-- Pick up the alternate manifest that was extracted from the core module -->
+                <transformer implementation="org.apache.maven.plugins.shade.resource.IncludeResourceTransformer">
+                  <resource>META-INF/MANIFEST.MF</resource>
+                  <file>${project.build.directory}/META-INF-shaded/MANIFEST.MF</file>
+                </transformer>
+              </transformers>
+              <!-- Keep the dependencies of driver-core -->
+              <promoteTransitiveDependencies>true</promoteTransitiveDependencies>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+</project>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -27,7 +27,6 @@
   </parent>
 
   <artifactId>java-driver-core</artifactId>
-  <packaging>bundle</packaging>
 
   <name>DataStax Java driver for Apache Cassandra(R) - core</name>
 
@@ -140,6 +139,11 @@
     <plugins>
       <plugin>
         <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+          </archive>
+        </configuration>
         <executions>
           <execution>
             <id>test-jar</id>
@@ -160,30 +164,16 @@
           </properties>
         </configuration>
       </plugin>
+      <!--
+      We avoid packaging bundle because it does not play nicely with the shade plugin, see
+      https://stackoverflow.com/questions/31262032/maven-shade-plugin-and-custom-packaging-type
+      -->
       <plugin>
         <groupId>org.apache.felix</groupId>
         <artifactId>maven-bundle-plugin</artifactId>
         <configuration>
-          <manifestLocation>${project.build.outputDirectory}/META-INF</manifestLocation>
           <instructions>
             <Bundle-SymbolicName>com.datastax.oss.driver.core</Bundle-SymbolicName>
-            <!-- Allow importing code from other packages (so reflection-based loading of policies works) -->
-            <DynamicImport-Package>*</DynamicImport-Package>
-            <!-- Don't include JNR packages since some JNR modules are not OSGi bundles,
-                 jcip because its not an OSGi bundle, jctools because its shaded.
-                 Import sun.misc as jctools does this. -->
-            <Import-Package>
-              !net.jcip.annotations,!jnr.*,!org.jctools.*,sun.misc;resolution:=optional,*
-            </Import-Package>
-            <!-- Explicitly declare shaded jctools packages since this isn't covered by shade plugin -->
-            <Export-Package>
-              com.datastax.oss.driver.*.core.*,
-              com.datastax.oss.driver.shaded.jctools.util;version="${project.version}";uses:="sun.misc",
-              com.datastax.oss.driver.shaded.jctools.queues;version="${project.version}";uses:="com.datastax.oss.driver.shaded.jctools.queues.spec",
-              com.datastax.oss.driver.shaded.jctools.queues.spec;version="${project.version}",
-              com.datastax.oss.driver.shaded.jctools.queues.atomic;version="${project.version}";uses:="com.datastax.oss.driver.shaded.jctools.queues,com.datastax.oss.driver.shaded.jctools.queues.spec",
-              com.datastax.oss.driver.shaded.jctools.maps;version="${project.version}"
-            </Export-Package>
           </instructions>
           <!--
           Prevent customized manifest entries from the project's maven-jar-plugin configuration from being read, see
@@ -193,6 +183,64 @@
             <forced>true</forced>
           </archive>
         </configuration>
+        <executions>
+          <execution>
+            <id>bundle-manifest</id>
+            <phase>process-classes</phase>
+            <goals>
+              <goal>manifest</goal>
+            </goals>
+            <configuration>
+              <manifestLocation>${project.build.outputDirectory}/META-INF</manifestLocation>
+              <instructions>
+                <!-- Allow importing code from other packages (so reflection-based loading of policies works) -->
+                <DynamicImport-Package>*</DynamicImport-Package>
+                <!-- Don't include JNR packages since some JNR modules are not OSGi bundles,
+                     jcip because its not an OSGi bundle, jctools because its shaded.
+                     Import sun.misc as jctools does this. -->
+                <Import-Package>
+                  !net.jcip.annotations,!jnr.*,!org.jctools.*,sun.misc;resolution:=optional,*
+                </Import-Package>
+                <!-- Explicitly declare shaded jctools packages since this isn't covered by shade plugin -->
+                <Export-Package>
+                  com.datastax.oss.driver.*.core.*,
+                  com.datastax.oss.driver.shaded.jctools.util;version="${project.version}";uses:="sun.misc",
+                  com.datastax.oss.driver.shaded.jctools.queues;version="${project.version}";uses:="com.datastax.oss.driver.shaded.jctools.queues.spec",
+                  com.datastax.oss.driver.shaded.jctools.queues.spec;version="${project.version}",
+                  com.datastax.oss.driver.shaded.jctools.queues.atomic;version="${project.version}";uses:="com.datastax.oss.driver.shaded.jctools.queues,com.datastax.oss.driver.shaded.jctools.queues.spec",
+                  com.datastax.oss.driver.shaded.jctools.maps;version="${project.version}"
+                </Export-Package>
+              </instructions>
+            </configuration>
+          </execution>
+          <!-- Alternate execution to generate the manifest to be used by the java-driver-core-shaded module -->
+          <execution>
+            <id>bundle-manifest-shaded</id>
+            <phase>process-classes</phase>
+            <goals>
+              <goal>manifest</goal>
+            </goals>
+            <configuration>
+              <manifestLocation>${project.build.directory}/classes/META-INF-shaded</manifestLocation>
+              <instructions>
+                <Bundle-Name>DataStax Java driver for Apache Cassandra(R) - core netty shaded</Bundle-Name>
+                <DynamicImport-Package>*</DynamicImport-Package>
+                <!-- Don't include netty since it will be shaded, import javax.security.cert as needed by netty ssl -->
+                <Import-Package>
+                  !net.jcip.annotations,!jnr.*,!org.jctools.*,sun.misc;resolution:=optional,!io.netty.*,javax.security.cert,*
+                </Import-Package>
+                <Export-Package>
+                  com.datastax.oss.driver.*.core.*,
+                  com.datastax.oss.driver.shaded.jctools.util;version="${project.version}";uses:="sun.misc",
+                  com.datastax.oss.driver.shaded.jctools.queues;version="${project.version}";uses:="com.datastax.oss.driver.shaded.jctools.queues.spec",
+                  com.datastax.oss.driver.shaded.jctools.queues.spec;version="${project.version}",
+                  com.datastax.oss.driver.shaded.jctools.queues.atomic;version="${project.version}";uses:="com.datastax.oss.driver.shaded.jctools.queues,com.datastax.oss.driver.shaded.jctools.queues.spec",
+                  com.datastax.oss.driver.shaded.jctools.maps;version="${project.version}"
+                </Export-Package>
+              </instructions>
+            </configuration>
+          </execution>
+        </executions>
       </plugin>
       <!-- shade jctools -->
       <plugin>
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -15,7 +15,9 @@
     limitations under the License.
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
   <parent>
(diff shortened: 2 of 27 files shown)


core/pom.xml
```diff
@@ -75,6 +75,11 @@
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.jctools</groupId>
+      <artifactId>jctools-core</artifactId>
+      <version>2.1.2</version>
```
tolbertam: TODO: need to move definition to parent pom, remove version here.

integration-tests/pom.xml
```diff
@@ -87,9 +113,55 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <!-- Not needed for maven build but useful for testing from intellij -->
+          <systemPropertyVariables>
```
tolbertam, line 121: Couldn't find an easy way to not duplicate this.  Putting this on surefire plugin enables IntelliJ to automatically pick it up.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/DefaultWriteCoalescer.java
```diff
@@ -31,6 +30,7 @@
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import net.jcip.annotations.ThreadSafe;
+import org.jctools.queues.atomic.MpscLinkedAtomicQueue;
```
tolbertam, line 33: Needed because netty shade's jctools and shaded packages are not automatically exported by maven-bundle-plugin, so just added jctools as a dependency, although maybe its better to shade it ourselves since its a tiny dependency.

core/src/main/java/com/datastax/oss/driver/internal/core/util/Reflection.java
```diff
@@ -45,11 +45,10 @@
    *
    * @return null if the class does not exist.
    */
-  public static Class<?> loadClass(String className) {
+  public static Class<?> loadClass(ClassLoader classLoader, String className) {
```
tolbertam, line 49: With OSGi, `Thread.currentThread().getContextClassLoader()` is able to load the class (i.e. `ExpontentialReconnectionPolicy`) just fine, but its a different class loader than what was used to initially load the super type (i.e. `ReconnectionPolicy`) we are checking against which was loaded with the bundle it belongs to (the driver core bundle).   To work around that, I changed this to pass in the `ClassLoader` of the super type (see line 180, 186), but this will not work for custom implementations that are not part of the core driver bundle.

In the general case, none of this is an issue for users unless they are running in an embedded runtime (like osgi or a web app container) or doing something tricky with classloaders.

I think if we were to provide a programmatic builder option, such as `Session.builder().withClassLoader(..)` that was only used for classes that aren't part of the core driver, that could be a solution.

core-shaded/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+                    <resource>META-INF/maven/io.netty/netty-codec/pom.xml</resource>
+                    <!-- netty's shading of jctools does not remove its pom files -->
+                    <resource>META-INF/maven/org.jctools/jctools-core/pom.properties</resource>
+                    <resource>META-INF/maven/org.jctools/jctools-core/pom.xml</resource>
+                  </resources>
+                </transformer>
+                <!-- Pick up the alternate manifest that was generated by the core module -->
+                <transformer implementation="org.apache.maven.plugins.shade.resource.IncludeResourceTransformer">
+                  <resource>META-INF/MANIFEST.MF</resource>
+                  <file>${project.parent.basedir}/core/target/META-INF-shaded/MANIFEST.MF</file>
```
tolbertam: This might be a bit suspect/non-ideal, what do you think?  We rely on the java-driver-core module to generate the shaded manifest for us, and then pull it in here.  I found this a better alternative than explicitly defining the `Export-Package` and `Import-Package` sections which is quite complex.

tolbertam:
TODO:

* [x] OSGi doc
* [x] Shaded doc

manual/core/shaded_jar/README.md
```diff
@@ -0,0 +1,39 @@
...
+This explicit dependency can be a problem if your application already
+uses another Netty version. To avoid conflicts, we provide a "shaded"
+version of the JAR, which bundles the Netty classes under a different
+package name:
+
+```xml
+<dependency>
+  <groupId>com.datastax.oss</groupId>
+  <artifactId>java-driver-core-shaded</artifactId>
+  <version>4.0.0-alpha3</version>
```
tolbertam, line 15: Since it's a completely different artifact, we don't need to exclude netty as a dependency anymore which is nice.

manual/core/shaded_jar/README.md
```diff
@@ -0,0 +1,39 @@
...
+  <artifactId>java-driver-query-builder</artifactId>
+  <version>4.0.0-alpha3</version>
+  <exclusions>
+    <exclusion>
+      <groupId>com.datastax.oss</groupId>
+      <artifactId>java-driver-core</artifactId>
+    </exclusion>
+  </exclusions>
+</dependency>
+```
```
tolbertam, line 39: I omitted the [Limitations](https://docs.datastax.com/en/developer/java-driver/3.5/manual/shaded_jar/#limitations) section we had in the original docs because:

1. You can override netty settings in reference the config (under `advanced.netty`).  So a user would no longer need to provide their own `NettyOptions` implementation in most cases.
2. The only real impact from what I can tell is if you want to override `Compressor<ByteBuf> compressor()`, `FrameCodec<ByteBuf> frameCode()`, or `NettyOptions nettyOptions()` in `InternalDriverContext`.  I consider that a somewhat advanced topic and probably extremely atypical, so I think having documentation for that may just add more noise/confusion than it does any good.

manual/osgi/README.md
```diff
@@ -0,0 +1,36 @@
+# OSGi
```
tolbertam, line 1: The OSGi docs for 3.x provide a FAQ section which is pretty extensive and includes steps for logging, compression and so on.  It also had some framework specific stuff for tycho and linked to an examples library.

I don't think we need to go so far other than to indicate that we support OSGi and the other sections i've included here.   But let me know if you'd like me to add more detail.

pom.xml
```diff
@@ -35,119 +35,171 @@
...
+    <assertj.version>3.6.2</assertj.version>
+    <commons-exec.version>1.3</commons-exec.version>
+    <felix.version>5.6.10</felix.version>
+    <jackson.version>2.9.5</jackson.version>
+    <junit.version>4.12</junit.version>
+    <junit-dataprovider.version>1.12.0</junit-dataprovider.version>
+    <logback.version>1.2.3</logback.version>
+    <mockito.version>2.7.19</mockito.version>
+    <pax-exam.version>4.11.0</pax-exam.version>
+    <simulacron.version>0.8.3</simulacron.version>
```
olim7t: Is this only to inject the versions in the `integration-tests` POM?

If so, I'd rather limit it to the strictly minimal set. We did it for every dependency in 3.x, but I find that annoying, especially for an outsider who is navigating the POMs from somewhere else than an IDE (say, on search.maven.org).

I'm ok with capturing things as properties if there's a good reason (another one is if the version is shared across multiple dependencies), but not by default.

core/src/main/java/com/datastax/oss/driver/internal/core/util/Reflection.java
olim7t: > With OSGi, Thread.currentThread().getContextClassLoader() is able to load the class (i.e. ExpontentialReconnectionPolicy) just fine, but its a different class loader than what was used to initially load the super type (i.e. ReconnectionPolicy) we are checking against which was loaded with the bundle it belongs to (the driver core bundle).

So what's happening then? The two classes are loaded, does the `isAssignableFrom` call fail?

If OSGI is too complicated (and after reading a bit on it, it seems like it is), then I agree with your suggestion of adding the "user classloader" as a builder argument. But if it's not specified, we should keep the current behavior (context class loader, then `Reflection`'s loader), because I believe it's the right thing to do in a JEE container, if the driver is in the boot classpath. I'll run a test in Tomcat to confirm this.

core-shaded/pom.xml
olim7t: As discussed privately, the relative reference bothers me a bit out of principle: in theory, you should be able to check out a module as an independent Maven project and build it (provided that the parent POM and all upstream modules are in your local repo).

But, all things considered, there's little chance of that happening, so that's the most pragmatic approach here.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/DefaultWriteCoalescer.java
olim7t: Yeah, I think it would be a good idea to shade it, both in the regular and in the Netty-shaded core.
The package should be exported, in case a client of the shaded core writes their own flusher.

pom.xml
tolbertam: 👍 i'll change this to only reflect the versions we need for the tests.  Alternatively, I could just hardcode the version numbers in the failsafe configuration as well, but i can see us forgetting to update those when we update dependencies.

There's also [a maven plugin](https://gitlab.com/josh.cain/dependencyversion-maven-plugin) out there that surfaces dependency versions as properties automatically. It doesn't seem too heavily used, but we could try it out, wdyt?

core/src/main/java/com/datastax/oss/driver/internal/core/util/Reflection.java
tolbertam: exactly, `isAssignable` from fails because the classloaders used between the two classes is different.   I'll revert it back and add a classloader builder option and we can document that detail in the manual somewhere as well.

core-shaded/pom.xml
tolbertam: The other possible viable alternative is to add the shaded MANIFEST to the core jar somwhere, and then pull that MANIFEST from the jar when we shade and put it into `META-INF/MANIFEST.MF`.  That would work around the problem of relative referencing, but will add a unrelated file to the core jar.  Is that worth doing or would you rather not add that extra file to the core jar?

core/src/main/java/com/datastax/oss/driver/internal/core/channel/DefaultWriteCoalescer.java
tolbertam: Sounds good, i'll add that as it's pretty easy to do.

core/src/main/java/com/datastax/oss/driver/internal/core/util/Reflection.java
tolbertam: Hrmm, thinking about this some more, I think you are right.

If you use `Thread.currentThread.getContextClassLoader()` by default, I expect it to work in the general case and also in a JEE container if the driver is in the boot classpath or loaded in the same webapp.

You could do the same thing with OSGi by loading the driver bundle in the boot classpath as well, which I think could be reasonable, but I think it also means that all libraries the driver depends on also have to be loaded in the boot classpath.  That's a little different than a webapp, where you can have a ton of jars loaded in the webapp.

I'll do a little more research on how typical this is in OSGi land, but maybe that is the right approach.

tolbertam: Oh, i got something working and it didn't require something crazy like that (well at least it doesn't seem as crazy to me).  Will push up shortly.

tolbertam: For now, I just changed the code to use `Class.forName` with no `ClassLoader` input provided to using the calling class' `ClassLoader` (45f9d32), which seems to work fine as well.  However, I think it may not work in the case where the policy comes from another bundle, which may be solved by using `DynamicImport-Package: *` (see: [this](http://billhiggins.us/blog/2011/02/24/using-reflection-in-osgi/).  Playing with that now.

tolbertam: Confirmed, `Dynamic-ImportPackage: *` was needed to allow the core driver to import code from other bundles via reflection.  I added that and a test to exercise it (10be2cd).  I think that will solve our problem for OSGi, but we may have more to do with `ClassLoader`, maybe we can attack that in a future ticket, or see if it becomes a problem for users?  We may be ok doing nothing.

pom.xml
tolbertam: Unfortunately the maven plugin doesn't work with intellij, best not to use it and only have properties for the things referenced.  Will make that change.

core/src/main/java/com/datastax/oss/driver/internal/core/util/Reflection.java
olim7t: > isAssignableFrom fails because the classloaders used between the two classes is different

I don't understand, this is with a built-in policy implementation right? How can two types that belong to the same JAR get loaded by different classloaders?

Regarding `Dynamic-ImportPackage: *`, as far as I understand this is the nuclear option, so I still see a value in allowing users to pass their classloader to the builder.

core-shaded/pom.xml
olim7t: What does it take to pull the file from the other JAR, just a different transformer? If it's a similar amount of lines in the POM, I don't mind adding an unrelated file to the core jar. If it's more than that, we can live with the relative reference.

tolbertam: It's just as easy, and more straightforward (as doesn't rely on location of generating file in other module).  I can make that change.

core/src/main/java/com/datastax/oss/driver/internal/core/util/Reflection.java
tolbertam: > How can two types that belong to the same JAR get loaded by different classloaders?

They are loaded by different classloaders because one of them is loaded via reflection using the given `ClassLoader` and the other isn't.   

i.e.:

`ReconnectionPolicy.class` that gets passed to `Reflection.buildFromConfig` is referenced in code in `DefaultDriverContext.buildReconnectionPolicy()`.  Therefore it uses the classloader for the bundle of java-driver-core.

`ExpontentialReconnectionPolicy` gets loaded by the `ClassLoader` via reflection using the `ClassLoader` specified as an input to `Class.forName`.

* `Thread.currentThread().contextClassLoader()` is an `Launcher$AppClassLoader`, its associated with the main thread of the app, which can load classes from bundles.
* `Class.forName` without a `ClassLoader` uses the `ClassLoader` of the caller, `Reflection` (the caller) which was loaded with the bundle, so its the same `ClassLoader` as `ExponentialReconnectionPolicy`, but it does not have access to load code from outside bundles, unless using `DynamicImport-Package: *`.

One solution I just realized is that you could just load the supertype using reflection too using the same `ClassLoader` and then do the `isAssignableFrom` check.  `Thread.currentThread().contexClassLoader()` would actually work for that and may not require dynamic import.  I'll look into that.   That might not be a 100% solution (which is why we may need a way for user to provide ClassLoader), but makes the issue less pressing.

tolbertam: Hah, no that doesn't work either, because the resolution of the constructor fails because the `ClassLoader` used to load the class is different than that of the arguments being checked against:

> java.lang.IllegalArgumentException: Expected class ExponentialReconnectionPolicy (specified by advanced.reconnection-policy.class) to have an accessible constructor with arguments (interface com.datastax.oss.driver.api.core.context.DriverContext)

I think the rules we need to work with are:

1. The same `ClassLoader` used to load the super type, (in this case the bundle classloader) must be used to load the input class via reflection.  (So `Class.forName(className)` makes the most sense)

2. The `ClassLoader` loading the class via reflection needs to have access to the class its trying to load.  My current way around this was to use `DynamicImport-Package: *`.

So I really think the best option is to not provide a `ClassLoader` to `Class.forName` (so we use the calling class' `ClassLoader` and use `DynamicImport-Package`.

We could try loading everything in `buildFromConfig` using Reflection (including `DriverContext` for constructor resolution), and even then it may not work and it just feels wrong.  In any case i'm trying it to see what happens.

tolbertam: > We could try loading everything in buildFromConfig using Reflection (including DriverContext for constructor resolution), and even then it may not work and it just feels wrong. In any case i'm trying it to see what happens.

That doesn't work.

core-shaded/pom.xml
tolbertam: Done: 3d32d0017 . Wasn't as easy as I anticipated but no longer relies on relative pathing.

core/src/main/java/com/datastax/oss/driver/internal/core/util/Reflection.java
tolbertam: I realized that Akka would likely have the same issue and came across [this](https://doc.akka.io/docs/akka/2.5/general/configuration.html#a-word-about-classloaders):

> In several places of the configuration file it is possible to specify the fully-qualified class name of something to be instantiated by Akka. This is done using Java reflection, which in turn uses a ClassLoader. Getting the right one in challenging environments like application containers or OSGi bundles is not always trivial, the current approach of Akka is that each ActorSystem implementation stores the current thread’s context class loader (if available, otherwise just its own loader as in this.getClass.getClassLoader) and uses that for all reflective accesses. This implies that putting Akka on the boot class path will yield NullPointerException from strange places: this is not supported.

I noticed that their manifest also uses `DynamicImport-Package: *` ([sbt config](https://github.com/akka/akka/blob/master/project/OSGi.scala#L36)).

tolbertam: Looked more into Akka's class loading behavior today.  Here's how it resolves the `ClassLoader` by default, which matches the documentation from the previous comment:

https://github.com/akka/akka/blob/75f4e2dae297b3e04b309de8709d2dc0abf92be9/akka-actor/src/main/scala/akka/util/Reflect.scala#L162-L178

The gist is, default to `Thread.currentThread().getContextClassLoader()` and if that is null, use the ClassLoader of the caller class (with some exceptions).  In the general case, it will use `getContextClassLoader()`, but that can be overridden when creating the `ActorSystem`, which is exactly what akka-osgi does:

https://github.com/akka/akka/blob/75f4e2dae297b3e04b309de8709d2dc0abf92be9/akka-osgi/src/main/scala/akka/osgi/OsgiActorSystemFactory.scala#L53-L63

It creates a [`BundleDelegatingClassLoader`](https://github.com/akka/akka/blob/75f4e2dae297b3e04b309de8709d2dc0abf92be9/akka-osgi/src/main/scala/akka/osgi/BundleDelegatingClassLoader.scala) that falls back on `ActorSystem`s `ClassLoader`.

This is all very complex, but I think we should probably follow suit on the idea of allowing the user to pass in the `ClassLoader` as a builder option, and otherwise using `Thread.currentThread().getContextClassLoader()` by default.  I don't think we need to go as far as creating a custom OSGi module for using the driver with OSGi, but we can document how to provide a class loader that works.

tolbertam: I added `SessionBuilder.withClassLoader(..)` in e0ba792.  I also added a documentation section to our OSGi docs similar to the akka documentation.  I think this is probably the best solution, the default `ClassLoader` resolution is the same as it was from the beginning, and now the user has the option to override it if they need to.  This behavioris also very close to how to akka handles things, short of providing a `driver-osgi` module, which I think might be overkill.

tolbertam:
I think i've addressed everything but let me know if there's anything I missed.   If it's ok, I rebase this against 4.x locally and there were a few changes I needed to account for, but didn't have to change anything fundamental (didn't squash or anything either).  Ok if I force push that up?

olim7t:
👍 to rebase

tolbertam:
> +1 to rebase

Thanks, done!

manual/osgi/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+The driver is available as an [OSGi] bundle.  More specifically, the following maven artifacts are valid OSGi bundles:
```
olim7t: Nit: could you respect the 100-character line limit in mardown files? You can show the right margin in IDEA to make this easier, there's also a "wrap to column" plugin.

tolbertam:
Need to do a little work now that jcip annotations dependency is required, will fix that shortly.

manual/osgi/README.md
```diff
@@ -0,0 +1,66 @@
...
+In several places of the [driver configuration] it is possible to specify the class name of something to be
+instantiated by the driver such as the reconnection policy. This is accomplished using reflection, which uses a
+ClassLoader.  By default, the driver uses `Thread.currentThread.getContextClassLoader()` if available,
+otherwise it uses its own `ClassLoader`.  This is typically adequate except in environments like application containers
+or OSGi frameworks where class loading logic is much more deliberate and libraries are isolated from each other.
+
+If the chosen `ClassLoader` is not able to ascertain whether a loaded class is the same instance as its expected parent
+type, you may encounter an error such as:
+
+    java.lang.IllegalArgumentException: Expected class ExponentialReconnectionPolicy (specified by advanced.reconnection-policy.class) to be a subtype of com.datastax.oss.driver.api.core.connection.ReconnectionPolicy
```
olim7t: It's also a good idea to wrap long code lines (even if they don't wrap in reality), I think our HTML generated doc doesn't deal very well with them.

core/src/main/java/com/datastax/oss/driver/internal/core/util/Reflection.java
```diff
@@ -45,11 +46,14 @@
...
-      if (contextClassLoader != null) {
-        return Class.forName(className, true, contextClassLoader);
+      // If input classLoader is null, use current thread's ClassLoader, if that is null, use
+      // default (calling class') ClassLoader.
+      ClassLoader cl =
+          classLoader != null ? classLoader : Thread.currentThread().getContextClassLoader();
+      if (cl != null) {
+        return Class.forName(className, true, cl);
       } else {
         return Class.forName(className);
```
olim7t, line 58: 👍

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -222,6 +223,16 @@ public SelfT withKeyspace(String keyspaceName) {
     return withKeyspace(CqlIdentifier.fromCql(keyspaceName));
   }
 
+  /**
+   * The {@link ClassLoader} to use to reflectively load class names defined in configuration. If
+   * null, the driver attempts to use {@link Thread#getContextClassLoader()} of the current thread
+   * or {@link com.datastax.oss.driver.internal.core.util.Reflection}'s {@link ClassLoader}.
```
olim7t: Maybe add a comment that this is for complex environment like OSGi, and in other case user doesn't need to worry about this.

olim7t: Referencing an internal class feels a bit too specific, how about "the classloader that loaded the core driver classes"?

integration-tests/src/test/java/com/datastax/oss/driver/osgi/OsgiCustomLoadBalancingPolicyIT.java
```diff
@@ -0,0 +1,37 @@
...
+package com.datastax.oss.driver.osgi;
+
+import org.ops4j.pax.exam.Option;
+
+/**
+ * Test that uses a policy from a separate bundle from the core driver to ensure that the driver is
+ * able to load that policy via Reflection. To support this, the driver uses <code>
+ * DynamicImport-Package: *</code>.
+ */
+public class OsgiCustomLoadBalancingPolicyIT extends OsgiBaseIT {
```
olim7t, line 25: 💯 for covering this

pom.xml
```diff
@@ -35,54 +35,75 @@
...
   </modules>
 
   <properties>
     <format.validateOnly>true</format.validateOnly>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+    <config.version>1.3.1</config.version>
+    <guava.version>25.1-jre-SNAPSHOT</guava.version>
+    <hdrhistogram.version>2.1.10</hdrhistogram.version>
+    <jctools.version>2.1.2</jctools.version>
```
olim7t: I think we don't need to extract this as a property anymore, we don't need it in the OSGi tests since it's shaded.

integration-tests/src/test/java/com/datastax/oss/driver/osgi/OsgiBaseIT.java
```diff
@@ -0,0 +1,88 @@
...
+import org.ops4j.pax.exam.Configuration;
+import org.ops4j.pax.exam.Option;
+import org.ops4j.pax.exam.junit.PaxExam;
+import org.ops4j.pax.exam.spi.reactors.ExamReactorStrategy;
+import org.ops4j.pax.exam.spi.reactors.PerClass;
+
+@RunWith(PaxExam.class)
+@ExamReactorStrategy(PerClass.class)
+@Category(IsolatedTests.class)
+public abstract class OsgiBaseIT {
```
olim7t: OSGi tests require at least a `mvn package` in the other modules so that the JARs are available. Which is fine given the nature of these tests, but we should mention it here, it's easy to forget and run into errors when you try to execute from the IDE.

olim7t (approved):
LGTM, glad that we're done with this 🎉 
See a few minor remarks inline.

One last thing: I would *really* like to have the two tickets as separate commits -- 1762 first sounds most logical -- it makes things much clearer when you `git blame` the POMs later. It won't be trivial because everything is intertwined right now. I have a bit of experience doing that so I can take care of it, lmk.

manual/osgi/README.md
tolbertam: Ah, I thought I was.  Apparently my IDE is using a 120 line margin, will fix.

pom.xml
tolbertam: good catch - will fix.

integration-tests/src/test/java/com/datastax/oss/driver/osgi/OsgiBaseIT.java
tolbertam: :+1: will do

tolbertam:
Thanks for reviewing!  I'll process your feedback quickly and push up my changes.

> One last thing: I would really like to have the two tickets as separate commits -- 1762 first sounds most logical -- it makes things much clearer when you git blame the POMs later. It won't be trivial because everything is intertwined right now. I have a bit of experience doing that so I can take care of it, lmk.

Good point.  It actually doesn't seem too bad so I can give that a try if you are ok with that.  My approach would be evaluate each commit and split it into separate commits to divide shading / osgi commits, and then bring everything together in separate commits.

olim7t:
> My approach would be evaluate each commit and split it into separate commits to divide shading / osgi commits, and then bring everything together in separate commits.

Sounds good to me 👍

tolbertam:
Updated to separate changes into commits and tested each commit individually to ensure everything was working.  Then I extracted the java-driver-shaded-guava snapshot out into a separate commit.

Forced pushed, kept the old commits around as [java1761old](https://github.com/datastax/java-driver/tree/java1761old) if needed to look at most recent commit (e8d3ee9).

After we merge and release datastax/java-driver-shaded-guava#1, we can chop off the last commit, retest and this should be good to go!  :crossed_fingers: .

tolbertam:
25.1-jre of java-driver-shaded-guava is now on central, dropped the temporary commit to use the snapshot.  Waiting for CI to run and then should be good to go.  Let me know if see any issues in the two separate commits, if looks good :+1:  to merge.

tolbertam:
CI ran well

tolbertam:
Sorry for the noise, I just happened to notice an issue with the JAVA-1762 commit message:

> java-driver-core-shaded module is not generated as part of the build.

i changed it to

> java-driver-core-shaded module is now generated as part of the build.

olim7t:
Forced-push to address a minor issue in manual (one instance of `ClassLoader` was not in monospace). All set 👍

PR #1020, opened 2018-06-08
Add various manual sections

olim7t:
Note: I deliberately didn't list them in the changelog, I don't think it's very useful to list documentation tasks there.

### DIFF
--- a/manual/core/async/README.md
+++ b/manual/core/async/README.md
@@ -1,0 +1,196 @@
+## Asynchronous programming
+
+The driver exposes an asynchronous API that allows you to write fully non-blocking programs.
+Asynchronous methods return instances of the JDK's [CompletionStage], that can be conveniently
+chained and composed.
+
+Here is a short example that opens a session and runs a query asynchronously:
+
+```java
+CompletionStage<CqlSession> sessionStage = CqlSession.builder().buildAsync();
+
+// Chain one async operation after another:
+CompletionStage<AsyncResultSet> responseStage =
+    sessionStage.thenCompose(
+        session -> session.executeAsync("SELECT release_version FROM system.local"));
+
+// Apply a synchronous computation:
+CompletionStage<String> resultStage =
+    responseStage.thenApply(resultSet -> resultSet.one().getString("release_version"));
+
+// Perform an action once a stage is complete:
+resultStage.whenComplete(
+    (version, error) -> {
+      if (error != null) {
+        System.out.printf("Failed to retrieve the version: %s%n", error.getMessage());
+      } else {
+        System.out.printf("Server version: %s%n", version);
+      }
+      sessionStage.thenAccept(CqlSession::closeAsync);
+    });
+```
+
+### Threading model
+
+The driver uses two internal thread pools: one for request I/O and one for administrative tasks
+(such as metadata refreshes, schema agreement or processing server events). Note that you can
+control the size of these pools with the `advanced.netty` options in the
+[configuration](../configuration).
+
+When you register a callback on a completion stage, it will execute on a thread in the corresponding
+pool:
+
+```java
+CompletionStage<CqlSession> sessionStage = CqlSession.builder().buildAsync();
+sessionStage.thenAccept(session -> System.out.println(Thread.currentThread().getName()));
+// prints s0-admin-n (admin pool thread)
+
+CompletionStage<AsyncResultSet> resultStage =
+    session.executeAsync("SELECT release_version FROM system.local");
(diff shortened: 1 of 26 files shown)


manual/core/async/README.md
```diff
@@ -0,0 +1,196 @@
...
+CompletionStage<Integer> resultStage =
+    responseStage.thenApplyAsync( // note: thenApplyAsync instead of thenApply
+        resultSet -> someVeryExpensiveComputation(resultSet),
+        computeExecutor);
+```
+
+Note that an alternate executor can also be used to allow synchronous driver API calls in callbacks,
+but the recommended approach is to fully commit to the asynchronous model described above. 
+
+### Error propagation
```
tolbertam, line 118: Excellent idea including this, I also like that you propose several mechanisms for error handling 👍

manual/core/tuples/README.md
```diff
@@ -0,0 +1,131 @@
...
+    Note that the resulting type is [detached](../detachable_types).
+  
+Once you have the type, call `newValue()` and set the fields:
+
+```java
+TupleValue tupleValue =
+    tupleType.newValue().setInt(0, 1).setString(1, "hello").setFloat(2, 2.3f);
+
+// Or as a one-liner for convenience:
+TupleValue tupleValue = tupleType.newValue(1, "hello", 2.3f);
```
olim7t, line 83: This example requires #1018 (same in the UDTs page)

manual/core/compression/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+### LZ4
+
+Dependency:
+
+```xml
+<dependency>
+  <groupId>net.jpountz.lz4</groupId>
+  <artifactId>lz4</artifactId>
+  <version>1.3.0</version>
```
tolbertam: I just realized, aren't we using the old coordinates for LZ4?  It looks like it has since moved to:

```xml
<dependency>
    <groupId>org.lz4</groupId>
    <artifactId>lz4-java</artifactId>
    <version>1.4.1</version>
</dependency>
```

I remembered we updated this in the 3.x driver, we should probably do it for 4.x too.

manual/core/native_protocol/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+```
+INFO DefaultSession - Negotiated protocol version v4 for the initial contact point, but other nodes
+                      only support v3, downgrading
+```
+
+If you want to see the details of mixed cluster negotiation, enable `DEBUG` level for the category
+`com.datastax.oss.driver.internal.core.CassandraProtocolVersionRegistry`.
+
+### New features by protocol version
```
tolbertam, line 92: Great idea to document new features by version

manual/core/ssl/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+```
+datastax-java-driver {
+  advanced.ssl-engine-factory {
+    class = com.mycompany.HostnameVerificationSslEngineFactory
+  }
+}
+```
+
+#### Netty
```
tolbertam, line 136: I wonder if we should consider providing something for this in the form of a separate module, what do you think?  I think providing it in the core module was problematic for us in 3.x, but adding a separate module for it might be a nice compromise as the Netty SSL implementation is much more object allocation efficient / performant.

manual/core/statements/prepared/README.md
```diff
@@ -131,6 +131,8 @@ This can be ambiguous if the query uses the same column multiple times, like in
 where sku = ? and date > ? and date < ?`. In these situations, use positional setters or named
 parameters.
 
+#### Unset values
+
 With native protocol V3, all variables must be bound. With native protocol V4 or above, variables
```
tolbertam: Would be good to link to native protocol section here

tolbertam (approved):
Excellent work on these docs!  I had a few comments, but overall 👍 .

manual/core/compression/README.md
olim7t: Right, I missed that. I will do it in a separate commit.

olim7t:
I pushed a couple of extra commits for docs-related items I had on my todo list.

manual/core/statements/prepared/README.md
olim7t: 👍 
I also updated the per-query keyspace page.

tolbertam:
recent changes look good too 👍

manual/core/ssl/README.md
tolbertam: Created: https://datastax-oss.atlassian.net/browse/JAVA-1859

PR #1021, opened 2018-06-08
JAVA-1830: Surface response frame size in ExecutionInfo

olim7t:
Requires datastax/native-protocol#23


core/src/main/java/com/datastax/oss/driver/api/core/cql/ExecutionInfo.java
```diff
@@ -147,4 +148,31 @@ default QueryTrace getQueryTrace() {
...
+   * <p>This is the size of the protocol-level frame (including the frame header) as it came in the
+   * TCP response, before decompression and decoding by the driver.
+   *
+   * <p>If compression is disabled, or if the information is not available (for example if this
+   * execution info comes from an {@link RetryDecision#IGNORE IGNORE} decision of the retry policy),
+   * this method returns -1.
+   *
+   * @see #getResponseSizeInBytes()
+   */
+  int getCompressedResponseSizeInBytes();
```
tolbertam, line 177: Nice idea including compressed byte size as well 👍

tolbertam (approved):
👍

PR #1022, opened 2018-06-10
JAVA-1767: better message when column not in result set

detinho:


core/src/main/java/com/datastax/oss/driver/api/core/data/AccessibleByName.java
```diff
@@ -56,4 +56,18 @@
...
+   * @see AccessibleByName#firstIndexOf
+   * @throws IllegalArgumentException if name was not found
+   */
+  default int tryFirstIndexOf(String name) {
+    final int indexOf = firstIndexOf(name);
+    if (indexOf == -1) {
+      throw new IllegalArgumentException(name + " is not a column defined in this metadata");
+    }
+    return indexOf;
+  }
```
olim7t: This looks good to me, but since `firstIndexOf` is not called from anywhere else anymore (actually it is from `DefaultRow.getType`, but we would benefit from the improved message there as well), we might as well modify the contract of `firstIndexOf` directly,  instead of adding a separate method: specify in the javadoc that an `IllegalArgumentException` should be thrown, and change the 4 existing implementations. This duplicates the check in each implementation, but we can also give an even better message: "is not a [column|field|variable] in this [row|UDT|bound statement]".

olim7t: Also, the same should be done for `AccessibleById`.

olim7t (changes requested):

detinho:
@olim7t Thanks for the review. 
I've done the changes you requested.

olim7t (approved):

olim7t:
👍 looks great to me, thanks! We'll merge this for the next release.

olim7t:
Force-pushed to:
- rebase on latest 4.x
- add a changelog entry
- fix a few trivial issues (change commit title, use camelCase for local variable in test, fix typo in javadoc)

PR #1023, opened 2018-06-11
JAVA-1794: Driver tries to create a connection array of size -1

detinho:
I'll look on branch 4.x to see if it suffers from a similar issue

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
@@ -115,7 +115,7 @@ class HostConnectionPool implements Connection.Owner {
 
         int toCreate = coreSize;
 
-        if (reusedConnection != null && reusedConnection.setOwner(this)) {
+        if (reusedConnection != null && toCreate > 0 && reusedConnection.setOwner(this)) {
             toCreate -= 1;
             connections.add(reusedConnection);
             connectionFutures.add(MoreFutures.VOID_SUCCESS);
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -5,6 +5,7 @@
 - [improvement] JAVA-1394: Add request-queue-depth metric.
 - [improvement] JAVA-1857: Add Statement.setHost.
 - [bug] JAVA-1920: Use nanosecond precision in LocalTimeCodec#format().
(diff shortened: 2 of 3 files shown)


driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
```diff
@@ -115,7 +115,7 @@ public void run() {
 
         int toCreate = coreSize;
 
-        if (reusedConnection != null && reusedConnection.setOwner(this)) {
+        if (reusedConnection != null && reusedConnection.setOwner(this) && toCreate > 0) {
```
tolbertam: I think we should change this slighty to:

> if (reusedConnection != null && toCreate > 0 && reusedConnection.setOwner(this)) {

This way we don't invoke `setOwner` on the connection to this pool unless `toCreate` is greater than 0.  As we don't want to assign ownership unless we will actually do it because then the connection will not get closed in `Cluster.onUp` where we check if the connection has an owner and if not close it.

tolbertam (changes requested):
Looks good thanks! I had one small comment though.

detinho:
@tolbertam as this is a small change I just made an amend.

tolbertam (approved):
looks good to me, 👍 to merge, pending signing the contributors license agreement.

detinho:
I've already signed the CLA. Ping me if I need to take any further action!

tolbertam:
> I've already signed the CLA. Ping me if I need to take any further action!

Hrm, something must be wrong with our CLA bot, we're looking into it, thanks

driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java
```diff
@@ -115,7 +115,7 @@ public void run() {
 
         int toCreate = coreSize;
 
-        if (reusedConnection != null && reusedConnection.setOwner(this)) {
+        if (reusedConnection != null && toCreate > 0 && reusedConnection.setOwner(this)) {
```
olim7t, line 118: I was wondering if the connection was getting closed if we don't enter this condition; it's handled in the `finally` block in `Cluster.Manager.onUp` so 👍 .

olim7t (approved):

tolbertam:
Rebased on 3.x and updated changelog and force pushed, will merge if travis build goes well :+1:

PR #1025, opened 2018-06-12
JAVA-1860: Allow reconnection at startup if no contact point is available

olim7t:


core/src/main/resources/reference.conf
```diff
@@ -268,6 +268,21 @@ datastax-java-driver {
...
+  # `SessionBuilder.build()` call -- or the future returned by `SessionBuilder.buildAsync()` --
+  # won't complete until a contact point has been reached.
+  #
+  # If this is false and no contact points are available, the driver will fail with an
+  # AllNodesFailedException.
+  #
+  # Required: yes
+  # Modifiable at runtime: no
+  # Overridable in a profile: no
+  advanced.reconnect-on-init = false
```
tolbertam, line 284: This is a good idea.  I think in typical case we want to fail fast and users can opt in to this feature if they want.

core/src/main/resources/reference.conf
```diff
@@ -268,6 +268,21 @@ datastax-java-driver {
     warn-on-init-error = true
   }
 
+  # Whether to schedule reconnection attempts if the driver fails to reach any contact point on the
```
tolbertam: I think it would be less vague if it said "... if the driver fails to reach all contact points ... " to imply that it will try all contact points before reconnecting.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/ConnectKeyspaceIT.java
```diff
@@ -0,0 +1,55 @@
...
+import com.datastax.oss.driver.api.testinfra.ccm.CcmRule;
+import com.datastax.oss.driver.api.testinfra.session.SessionRule;
+import com.datastax.oss.driver.api.testinfra.session.SessionUtils;
+import com.datastax.oss.driver.categories.ParallelizableTests;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(ParallelizableTests.class)
+public class ConnectKeyspaceIT {
```
tolbertam, line 30: At first I thought this test was not related, but I think the idea is, if the keyspace we are trying to connect to doesn't exist, we should fail instead of scheduling reconnects, is that right?

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -228,7 +228,11 @@ public SelfT withKeyspace(String keyspaceName) {
    * @return a completion stage that completes with the session when it is fully initialized.
    */
   public CompletionStage<SessionT> buildAsync() {
-    return buildDefaultSessionAsync().thenApply(this::wrap);
+    CompletionStage<CqlSession> buildStage = buildDefaultSessionAsync();
+    CompletionStage<SessionT> wrapStage = buildStage.thenApply(this::wrap);
+    // thenApply does not propagate cancellation (!)
```
tolbertam, line 248: Huh, did not know this, good to document 👍

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -83,7 +85,12 @@ public DefaultTopologyMonitor(InternalDriverContext context) {
     if (closeFuture.isDone()) {
       return CompletableFutures.failedFuture(new IllegalStateException("closed"));
     }
-    return controlConnection.init(true, false);
+    return controlConnection.init(true, reconnectOnInit);
```
tolbertam, line 88: Brilliant.  I was staring at the code for minutes trying to figure out where you added logic for this, nice to see it really fits in to the existing implementation and doesn't take a lot of code.

tolbertam (approved):
This looks great!  💯

integration-tests/src/test/java/com/datastax/oss/driver/api/core/ConnectKeyspaceIT.java
olim7t: Actually, I just renamed this test because it was called `ConnectIT` before. But now that you mention it, I could also check that a keyspace error does trigger retries even if the option is on.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
olim7t: Yeah that was a big surprise for me, I'm pretty sure the Guava futures did.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
olim7t: The existing implementation was for the case when you use a custom topology monitor that doesn't initialize the control connection. If you do a schema refresh later, you need to initialize it, and have it reconnect if it fails. The only difference is that the returned future represented the outcome of the first attempt only (metadata refreshes are not vital, so no need to wait). I changed that to only complete the future when the reconnection is complete. Metadata refreshes now wait on `ControlConnection.firstConnectionAttemptFuture()` instead.

PR #1027, opened 2018-06-13
JAVA-1858: Implement Serializable in implementations, not interfaces

olim7t:
Motivation:

A few of our types are serializable (row, data types, etc.), this is
mainly for integration with Spark. By making our interfaces extend
Serializable, we signal that all possible implementations should do it,
this is not necessarily the case if someone writes their own version in
an application that doesn't use Spark.

Modifications:

Remove `extends Serializable` from the interfaces, have the default
classes implement it instead.
Document in the interfaces that serializability is recommended, but not
strictly mandatory.

Result:

Less burden on implementors if they're not interested in Spark
integration.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -4,6 +4,7 @@
 
 ### 4.0.0-alpha4 (in progress)
 
+- [improvement] JAVA-1858: Implement Serializable in implementations, not interfaces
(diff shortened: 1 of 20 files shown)


tolbertam (approved):
Looks good, only thing in api that I see as `Serializable` now is `CqlIdentifier` which makes sense since it's class implementation.

PR #1028, opened 2018-06-13
JAVA-1847: Node level request tracking

GregBestland:


olim7t:
As discussed privately, I think it would be beneficial to make node-level callbacks separate methods, to give the VM a better chance of optimizing them when they are empty.

I wrote a throwaway method to run a bunch of queries in a loop:
```java
  @Test
  public void run() {
    logger.setLevel(Level.WARN); // not needed here, avoid lengthy output
    for (int i = 0; i < 100000; i++) {
      sessionRuleRequest.session().execute(QUERY);
    }
  }
```
And ran it with VM options to log inlining info:
```
-XX:+PrintCompilation -XX:+UnlockDiagnosticVMOptions -XX:+PrintInlining -XX:+LogVMOutput -XX:LogFile=/tmp/hotspot.log -XX:-DisplayVMOutput
```

I see this a lot (first call is the node-level callback, that does nothing in this case):
```
   2990 3412   !   3       com.datastax.oss.driver.internal.core.cql.CqlRequestHandlerBase::setFinalResult (173 bytes)
                              [...]
                              @ 90   com.datastax.oss.driver.internal.core.tracker.RequestLogger::onSuccess (213 bytes)   callee is too large
                              [...]
                              @ 122   com.datastax.oss.driver.internal.core.tracker.RequestLogger::onSuccess (213 bytes)   callee is too large
```
I think the node-level callback should be a separate method, and that it should be empty (most users probably expect session-level when they think of query logging, so let's not provide it out of the box). Then the VM would probably be able to elide the call altogether.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/tracker/RequestLoggerIT.java
```diff
@@ -189,4 +215,63 @@ public void should_not_log_when_disabled() throws InterruptedException {
     TimeUnit.MILLISECONDS.sleep(500);
     Mockito.verify(appender, never()).doAppend(any(LoggingEvent.class));
   }
+
+  @Test
+  public void should_log_failed_nodes_on_succesfull_request() {
```
tolbertam: nit: successfull -> successful

tolbertam (commented):
Looks good so far, will give another look when Olivier's change is addressed.  Just one small nit with spelling.

GregBestland:
I think this should solve the issues discussed in the PR. I'd like someone to give it a once over prior to squash and merge.

core/src/main/java/com/datastax/oss/driver/api/core/tracker/RequestTracker.java
```diff
@@ -48,4 +48,32 @@ void onError(
...
+   */
+  void onNodeError(
+      Request request,
+      Throwable error,
+      long latencyNanos,
+      DriverConfigProfile configProfile,
+      Node node);
+
+  /**
+   * Invoked each time a request succeeds at the node level. Similar to on Success but at per Node
```
tolbertam: `on Success` -> `onSuccess` and since referring to another method it would be useful to use `{@link #onSuccess(...)}` so you can click to it in javadoc.  (same for `onError` above).

core/src/main/java/com/datastax/oss/driver/api/core/tracker/RequestTracker.java
```diff
@@ -48,4 +48,32 @@ void onError(
...
       Node node);
+
+  /**
+   * Invoked each time a request fails at the node level. Similar to onError but at a per node
+   * level.
+   *
+   * @param latencyNanos the overall execution time (from the {@link Session#execute(Request,
+   *     GenericType) session.execute} call until the error is propagated to the client).
+   * @param configProfile the configuration profile that this request was executed with.
+   * @param node the node that returned the error response, or {@code null} if the error occurred
```
tolbertam: This part was copied over and I think it may technically be incorrect for this new method:

> node the node that returned the error response, or `{@code null}` if the error occurred

Since it's invoked for a specific node, won't it always be non-null?

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
```diff
@@ -339,9 +347,13 @@ public void onThrottleFailure(RequestThrottlingException error) {
   }
 
   private void setFinalError(Throwable error, Node node) {
+
     if (result.completeExceptionally(error)) {
       cancelScheduledTasks();
       long latencyNanos = System.nanoTime() - startTimeNanos;
+      if (node != null) {
+        context.requestTracker().onNodeError(statement, error, latencyNanos, configProfile, node);
```
tolbertam: Couldn't this also use `trackNodeError`?

tolbertam (approved):
Some very tiny nits, but looks good otherwise :+1:

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
tolbertam: I see, this is explained by eac3282b, nevermind :)

tolbertam:
ah forgot to mention, this needs a changelog entry

core/src/main/java/com/datastax/oss/driver/api/core/tracker/RequestTracker.java
```diff
@@ -48,4 +48,32 @@ void onError(
       long latencyNanos,
       DriverConfigProfile configProfile,
       Node node);
+
+  /**
+   * Invoked each time a request fails at the node level. Similar to onError but at a per node
+   * level.
+   *
+   * @param latencyNanos the overall execution time (from the {@link Session#execute(Request,
+   *     GenericType) session.execute} call until the error is propagated to the client).
```
olim7t: I completely missed this on my first review: the latency passed to node-level methods should be measured from the time we sent the request to that node, not from the beginning of the whole request cycle. Right now it includes all tries on previous nodes.

There is already a `NodeResponseCallback.start` field for metrics, you can just use that instead of `startTimeNanos` (the name `start` is not great by the way, maybe we could rename it to `nodeStartTimeNanos` for clarity).

That also means we won't be able to handle the final node in `setFinalError`, so instead just add a call to `trackNodeError` everywhere `setFinalError` is called in the `NodeResponseCallback`.

olim7t (changes requested):

GregBestland:
I addressed your change @olim7t  also updated for new annotation will squash and merge on your approval

olim7t (approved):

olim7t:
I pushed a few additional nits. If you agree, OK to rebase, squash and merge at your convenience.

olim7t:
PS -- with a changelog entry please!

PR #1030, opened 2018-06-13
JAVA-1865: Add RelationMetadata.getPrimaryKey()

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/metadata/schema/RelationMetadata.java
```diff
@@ -30,6 +31,22 @@
...
+   *
+   * <p>Note that this creates a new list instance on each call.
+   *
+   * @see #getPartitionKey()
+   * @see #getClusteringColumns()
+   */
+  default List<ColumnMetadata> getPrimaryKey() {
+    return ImmutableList.<ColumnMetadata>builder()
+        .addAll(getPartitionKey())
+        .addAll(getClusteringColumns().keySet())
```
olim7t, line 46: Obviously this doesn't return the clustering order. The 3.x version didn't either, so I guess it's not a problem.

tolbertam (approved):

PR #1032, opened 2018-06-14
JAVA-1869: Add DefaultDriverConfigLoaderBuilder

tolbertam:
Note:  This is currently a proof of concept to ~illicit~ solicit some feedback before pursuing further.

Motivation:

To enable a convenient way to configure a Driver Session
programmatically without being part of the top-level API.

Modifications:

Added ProgrammaticDriverConfigLoader which provides a builder for
passing in configuration programmatically.

Added class DriverOption back to they may be configured.

Updated SessionUtils to provide a way of constructing a Session
with the given ProgrammaticDriverConfigLoader.Builder so it can
be easily used in integration tests.  Updated SpeculativeExecutionIT
to use it.

Result:

ProgrammaticDriverConfigLoader.builder() may be used to construct
a DriverConfigLoader which may be passed in via
SessionBuilder.withConfigLoader.

TODOs:

* [x] Use `@Nullable` / `@NonNullable` annotations where applicable
* [x] Javadoc
* [x] Update tests to use this and remove `TestConfigLoader`


integration-tests/src/test/java/com/datastax/oss/driver/api/core/specex/SpeculativeExecutionIT.java
```diff
@@ -327,14 +329,19 @@ public void should_not_speculatively_execute_when_defined_in_profile() {
...
         simulacron,
-        String.format("basic.request.timeout = %d milliseconds", SPECULATIVE_DELAY * 10),
-        "basic.request.default-idempotence = true",
-        "basic.load-balancing-policy.class = com.datastax.oss.driver.api.testinfra.loadbalancing.SortingLoadBalancingPolicy",
-        "advanced.speculative-execution-policy.class = ConstantSpeculativeExecutionPolicy",
-        String.format(
-            "advanced.speculative-execution-policy.max-executions = %d", maxSpeculativeExecutions),
-        String.format(
-            "advanced.speculative-execution-policy.delay = %d milliseconds", speculativeDelayMs));
+        ProgrammaticDriverConfigLoader.builder()
```
tolbertam: Example of using this

integration-tests/src/test/java/com/datastax/oss/driver/api/core/specex/SpeculativeExecutionIT.java
```diff
@@ -343,54 +350,67 @@ private CqlSession buildSessionWithProfile(
       int profile1MaxSpeculativeExecutions,
       long profile1SpeculativeDelayMs) {
 
-    List<String> config = Lists.newArrayList();
-    config.add(String.format("basic.request.timeout = %d milliseconds", SPECULATIVE_DELAY * 10));
-    config.add("basic.request.default-idempotence = true");
-    config.add(
-        "basic.load-balancing-policy.class = com.datastax.oss.driver.api.testinfra.loadbalancing.SortingLoadBalancingPolicy");
+    ProgrammaticDriverConfigLoader.Builder builder =
```
tolbertam: Example of using this with profiles.

core/src/main/java/com/datastax/oss/driver/api/core/config/ProgrammaticDriverConfigLoader.java
```diff
@@ -0,0 +1,217 @@
...
+
+    private Supplier<Config> fallbackSupplier = DefaultDriverConfigLoader.DEFAULT_CONFIG_SUPPLIER;
+
+    private Builder() {}
+
+    public Builder withBoolean(DriverOption option, boolean value) {
+      return with(option, value);
+    }
+
+    public Builder withBoolean(String profileName, DriverOption option, boolean value) {
```
tolbertam: One idea I had for profiles which would be feasible would be to have a way for user a to create a `DriverConfigProfile` and then have a `withProfile(DriverConfigProfile profile)` method in here.  That would reduce the duplicated methods.   The one downside of that is that you could not use these`DriverConfigProfile` objects when working with the driver, but we could probably find a way to throw an early runtime exception if someone tries that.    What do you think of this idea?

newkek:
I know naming is subjective, but there's no notion of "Programmatic" VS "file-based" so far in the driver in the naming iiuc, hence I find the name a little obscure, I would suggest something along the lines of `SimpleOverrideDriverConfigLoader`, wdyt?

tolbertam:
> I know naming is subjective, but there's no notion of "Programmatic" VS "file-based" so far in the driver in the naming iiuc, hence I find the name a little obscure, I would suggest something along the lines of SimpleOverrideDriverConfigLoader, wdyt?

I've never been good with naming things and I agree that the current naming is awkward, so I'd be open to renaming it.  To be honest I'm not a fan of `SimpleOverrideDriverConfigLoader` either, but we can probably think of something good.

tolbertam:
I was looking to akka for inspiration and I do see they have a way of [loading configuration programmatically](https://doc.akka.io/docs/akka/2.5/general/configuration.html#configuring-multiple-actorsystem) by just allowing you to pass in a `Config` object which is quite similar to how `TestConfigLoader` in `java-driver-test-infra` works:

```scala
import akka.actor.ActorSystem
import com.typesafe.config.ConfigFactory
val customConf = ConfigFactory.parseString("""
  akka.actor.deployment {
    /my-service {
      router = round-robin-pool
      nr-of-instances = 3
    }
  }
  """)
// ConfigFactory.load sandwiches customConfig between default reference
// config and default overrides, and then resolves it.
val system = ActorSystem("MySystem", ConfigFactory.load(customConf))
```

I think it might be nice to add a `withConfig` option to this config loader so you could do something similar, but I still think having a programmatic API is nice, especially since we have `DriverOption` exposing all of our properties.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/specex/SpeculativeExecutionIT.java
stamhankar999: I see no profile ref here or below. What am I missing?

integration-tests/src/test/java/com/datastax/oss/driver/api/core/specex/SpeculativeExecutionIT.java
```diff
@@ -343,54 +350,67 @@ private CqlSession buildSessionWithProfile(
...
+              DefaultDriverOption.SPECULATIVE_EXECUTION_POLICY_CLASS,
+              NoSpeculativeExecutionPolicy.class);
     }
 
     if (profile1MaxSpeculativeExecutions != -1 || profile1SpeculativeDelayMs != -1) {
-      config.add(
-          "profiles.profile1.advanced.speculative-execution-policy.class = ConstantSpeculativeExecutionPolicy");
+      builder =
+          builder.withClass(
+              "profile1",
```
stamhankar999: This one uses a profile; is this what you were referring to?

stamhankar999 (approved):
Looks great to me; just found that one comment you made in the PR confusing.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/specex/SpeculativeExecutionIT.java
tolbertam: Yes, sorry if it was confusing.

core/src/main/java/com/datastax/oss/driver/api/core/config/ProgrammaticDriverConfigLoader.java
```diff
@@ -0,0 +1,217 @@
...
+    /** Unsets an option. */
+    public Builder without(DriverOption option) {
+      return with(option, null);
+    }
+
+    public Builder without(String profileName, DriverOption option) {
+      return with(profileName, option, null);
+    }
+
+    public Builder withFallback(Supplier<Config> configSupplier) {
```
olim7t: This leaks Typesafe Config through the public API (parameter of a public method in an `api.*` package).

More generally, `ProgrammaticDriverConfigLoader` only works with Typesafe config, so I think we should either:

a. extract an interface that has the `withXxx` methods and works with any `DriverConfigLoader`, but that's kind of hard given how generic the latter is.
b. make programmatic configuration an internal-only thing.

I'm leaning towards b, having the ability to change options programmatically is an implementation-dependent thing.

core/src/main/java/com/datastax/oss/driver/api/core/config/ProgrammaticDriverConfigLoader.java
```diff
@@ -0,0 +1,217 @@
...
+    public Builder withBoolean(DriverOption option, boolean value) {
+      return with(option, value);
+    }
+
+    public Builder withBoolean(String profileName, DriverOption option, boolean value) {
+      return with(profileName, option, value);
+    }
+
+    public Builder withBooleanList(DriverOption option, List<Boolean> value) {
+      return with(option, value);
```
olim7t: This is the same as in `TypesafeDriverConfigProfile`. You could extract an internal interface with default methods and only have to implement `with()`. As a bonus, this would help validate that we're not missing any method here.
(provided that the loader becomes an internal class as suggested in my other comment)

core/src/main/java/com/datastax/oss/driver/api/core/config/ProgrammaticDriverConfigLoader.java
olim7t: Mmm not a fan...
How about a "sub-builder":
```java
withProfile(
    "slow",
    profileBuilder()
        .withDuration(DefaultDriverOption.REQUEST_TIMEOUT, Duration.ofSeconds(30))
        .build()
)
```
It could reuse the same internal interface that I'm suggesting below.

olim7t (changes requested):
Wrt to naming, I think "programmatic" is pretty good, since it provides a way to override options programmatically, as opposed to in the configuration file.

core/src/main/java/com/datastax/oss/driver/api/core/config/ProgrammaticDriverConfigLoader.java
tolbertam: I like it :+1: will pursue this.

core/src/main/java/com/datastax/oss/driver/api/core/config/ProgrammaticDriverConfigLoader.java
tolbertam: Hrm, i'll think some more about this.  Agree that we shouldn't leak typesafe config.  In addition I still need to think a little bit more about how this will work with the dse driver.  I think preferably this could be part of api and not internal, but maybe that is the right thing to do if there isn't a good way to allow overriding the backing config in some way.

tolbertam:
I force pushed just to rebase (very small merge conflicts dealt with in first commit).  I also pushed a change based on feedback and added something to the configuration docs page.

core/src/main/java/com/datastax/oss/driver/api/core/config/ProgrammaticDriverConfigLoader.java
```diff
@@ -0,0 +1,152 @@
...
+    private ImmutableMap.Builder<String, Object> values = ImmutableMap.builder();
+
+    private Builder() {}
+
+    /**
+     * @return Constructs a config loader based on the configuration values provided to this builder
+     *     and falls back to the default configuration mechanism of the driver.
+     */
+    public DriverConfigLoader build() {
+      ProgrammaticDriverConfigLoader.Builder configBuilder =
```
tolbertam: woops, accidental commit, will remove this.

tolbertam:
I rebased on 4.x, but still have some work to do (tracking via TODOs in original PR comment).

tolbertam:
I was thinking about this API some more and I have some more ideas that I want to experiment with, so hold off on further review until then.

tolbertam:
I refactored this idea quite a bit in 1439054 .  The main change is that `ProgrammaticDriverConfigLoaderBuilder` was changed to an abstract `DriverConfigLoaderBuilder` for which there is concrete implementation `CqlDriverConfigLoaderBuilder` which can be accessed via `CqlSession.configLoaderBuilder`.   The idea is we can implement alternative `DriverConfigLoaderBuilder` implementations which factor in the provided configuration to `withXXX` methods to generating `DriverConfigLoader` implementations.  Here's an example of how it's used:

```java
import com.datastax.oss.driver.api.core.CqlDriverConfigLoaderBuilder;
import com.datastax.oss.driver.api.core.DefaultConsistencyLevel;
import com.datastax.oss.driver.api.core.config.DefaultDriverOption;
import com.datastax.oss.driver.api.core.config.DriverConfigLoaderBuilder;
import java.time.Duration;

CqlDriverConfigLoaderBuilder configBuilder =
    CqlSession.configLoaderBuilder()
        .withDuration(DefaultDriverOption.REQUEST_TIMEOUT, Duration.ofMillis(500))
        .withProfile(
            "profile1",
            DriverConfigLoaderBuilder.profileBuilder()
                .withString(
                    DefaultDriverOption.REQUEST_CONSISTENCY,
                    DefaultConsistencyLevel.EACH_QUORUM.name())
                .build());

CqlSession.builder()
  .withConfigLoader(configBuilder.build())
  .build();
```

What do you think about this idea?  I'll add additional comments to further explain.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/session/DefaultSessionBuilderInstantiator.java
```diff
@@ -16,13 +16,18 @@
...
 import com.datastax.oss.driver.api.core.CqlSession;
+import com.datastax.oss.driver.api.core.config.DriverConfigLoaderBuilder;
 import com.datastax.oss.driver.api.core.session.SessionBuilder;
 
 public class DefaultSessionBuilderInstantiator {
   public static SessionBuilder<?, ?> builder() {
     return CqlSession.builder();
   }
 
+  public static DriverConfigLoaderBuilder configLoaderBuilder() {
```
tolbertam: Changing our tests to use this instead of `TestConfigLoader`, we can get rid of `configPath()`, and then this class starts to look at a lot like `CqlSession`.  In fact, we can get rid of `DefaultSessionBuilderInstantiator` completely and just point to `CqlSession` instead.

core/src/main/java/com/datastax/oss/driver/api/core/CqlSession.java
```diff
@@ -36,6 +37,18 @@ static CqlSessionBuilder builder() {
...
 
+  /**
+   * Returns a builder to create a new {@link DriverConfigLoader} that uses the default config
+   * loader as a basis, but allows providing additional configuration programmatically.
+   *
+   * <p>The built config loader may then be passed to {@link
+   * com.datastax.oss.driver.api.core.session.SessionBuilder#withConfigLoader(DriverConfigLoader)}.
+   */
+  @NonNull
+  static CqlDriverConfigLoaderBuilder configLoaderBuilder() {
```
tolbertam: I think providing `configLoaderBuilder` at the `CqlSession` level is a good happy medium.  It makes it easy enough to pass in programmatic configuration, but not so obvious that it's something that users will want to by default.  So I think this should continue to encourage use of configuration files, with providing programmatic config a fallback for where that isn't viable.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/session/SessionUtils.java
```diff
@@ -110,14 +125,26 @@ public static String getConfigPath() {
     return newSession(cassandraResource, keyspace, null, null, null, options);
   }
 
-  @SuppressWarnings({"unchecked", "TypeParameterUnusedInFormals"})
+  @SuppressWarnings("TypeParameterUnusedInFormals")
   public static <SessionT extends Session> SessionT newSession(
+      CassandraResourceRule cassandraResourceRule, DriverConfigLoaderBuilder loader) {
```
tolbertam: This should take `DriverConfigLoader` instead of `DriverConfigLoaderBuilder`, will update.

tolbertam: Actually, I think this will need to take a builder, since we need to do further customization in some places

tolbertam: nevermind, found a way around it.

tolbertam:
Rebased and pushed d24b724 which is a large commit that updates all the tests to use this new loader in favor of `TestConfigLoader`.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoaderBuilder.java
```diff
@@ -0,0 +1,116 @@
...
+  @Override
+  public SelfT with(@NonNull String path, @Nullable Object value) {
+    values.put(path, value);
+    return self;
+  }
+
+  /** Adds configuration for a profile constructed using {@link #profileBuilder()} by name. */
+  @NonNull
+  public SelfT withProfile(String profileName, Profile profile) {
+    String prefix = "profiles." + profileName + ".";
```
olim7t: This is kind of leaky: it assumes every config implementation will prepend profiles in the path. Which is probably likely, but technically nothing in the `DriverConfig` API mandates it.

core/src/main/java/com/datastax/oss/driver/api/core/CqlDriverConfigLoaderBuilder.java
```diff
@@ -0,0 +1,54 @@
...
+ * Provides a mechanism for constructing a {@link DriverConfigLoader} programmatically that uses
+ * {@link com.datastax.oss.driver.api.core.CqlSession}'s default config loader with the values of
+ * {@code withXXX(...)} methods overriding the configuration defined in configuration files.
+ *
+ * <p>The built {@link DriverConfigLoader} provided by {@link #build()} can be passed to {@link
+ * com.datastax.oss.driver.api.core.session.SessionBuilder#withConfigLoader(DriverConfigLoader)}.
+ */
+@NotThreadSafe
+public class CqlDriverConfigLoaderBuilder
+    extends DriverConfigLoaderBuilder<CqlDriverConfigLoaderBuilder> {
```
olim7t: This should not be called "Cql". It could be used with any kind of session so "Default" or "Typesafe" would be better. Also it should be in an internal package.

olim7t (changes requested):
I feel uneasy about exposing this in the public API:

* the concept of "config loader builder" is not clear, maybe because the public API types didn't retain "programmatic" in their name. Looking at `CqlSession.configLoaderBuilder()` from an outsider's perspective, I don't understand what it does until I read the javadoc.
* we are pushing file-based configuration as the default in driver 4. I'm fine providing workarounds in the internal API, but I think making this user-facing is going too far.
* the shared infrastructure doesn't provide a lot of value. The main point of interaction between `DriverConfigLoaderBuilder` and its implementations is `entrySet()`, so it's basically a map builder. What would be really useful is something that takes an arbitrary `DriverConfigLoader` and adds programmatic options on top of it, but as we've seen from previous attempts, that's too complex to do without leaking implementation details.
* as commented inline, the generic builder makes assumptions about how profiles are represented in the underlying source.

For all those reasons, I would prefer not to surface the concept in the public API, and only expose a typesafe config-specific implementation in internal packages as we initially intended.

On the other hand, big 👍 on switching all of our integration tests to it.

tolbertam:
Since this has already been reviewed and rebasing is looking challenging, i'm going to squash all commits into one as part of rebase for changes up to this point (so I don't have to fix conflicts at each commit).  For further changes, I'll add those as separate commits.

tolbertam:
I finished rebasing and force pushed, now will focus on processing feedback.

tolbertam:
I've processed feedback as fc862b3.  There were a few points I wasn't confident I understood completely:

> * [`withProfile` ... ] is kind of leaky: it assumes every config implementation will prepend profiles in the path. Which is probably likely, but technically nothing in the DriverConfig API mandates it.

I changed `withProfile` to be abstract in `DriverConfigLoaderBuilder` to deal with that concern, hopefully that was the right approach.

> the shared infrastructure doesn't provide a lot of value. The main point of interaction between DriverConfigLoaderBuilder and its implementations is entrySet(), so it's basically a map builder. What would be really useful is something that takes an arbitrary DriverConfigLoader and adds programmatic options on top of it, but as we've seen from previous attempts, that's too complex to do without leaking implementation details.

I may have not understood what is meant by this, 

Another thing I'm not confident about is the `ProgrammaticConfigBuilder` interface, which defines `withXXX` methods used by `DriverExecutionProfile` and `DriverConfigLoaderBuilder` and remains in tact, but only `DriverExecutionProfile` is API facing so to an outside observer it it could seem superfluous.  I could always get rid of that interface entirely and duplicate the method in each interface, but thought I'd ask to be sure before going through the effort of doing that.  What would you prefer?

olim7t:
To sum up the first two points: I wonder if abstracting out `DriverConfigLoaderBuilder` is worth it at all. I can see how it factors some boilerplate if someone ever wants to do the same thing for another `DriverConfig` implementation, but that feels a bit premature.

However this is more a matter of personal taste, and less of a concern now that it's internal, so it's fine for me either way. If you decide to keep the abstraction, I think `withProfile` can stay concrete like before, again because it's internal and we impose a few more restrictions on implementations.

About `ProgrammaticConfigBuilder`: it doesn't bother me that it's implemented only once in the public API. The name feels a bit off though, maybe because I think of a traditional builder as something with a `build()` method. I'm trying to think of another suggestion.

tolbertam:
I keep on iterating on this, but I think things are in a good place now as of fc84786 and would appreciate another look.

The end result is that the PR no longer touches `DriverExecutionProfile` at all, nor are there any outward changes at an API level (other than adding some missing `DefaultDriverOption`):

```
git diff 4.x --stat core/src/
 core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOption.java                 |   3 +
 core/src/main/java/com/datastax/oss/driver/internal/core/config/DriverOptionConfigBuilder.java      | 152 +++++++++++++++++++++++++++++++++++++++++++++++++
 .../main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java  |   5 ++
 .../com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoaderBuilder.java     | 133 +++++++++++++++++++++++++++++++++++++++++++
 4 files changed, 293 insertions(+)
```

One thing:  It may be of value to add `withClass` to `ExecutionProfile` as you can do things like change the `RetryPolicy` and `LoadBalancingPolicy` at a profile level, but didn't want to add that without confirmation.

olim7t:
OK, I will take another look.

> It may be of value to add withClass to ExecutionProfile as you can do things like change the RetryPolicy and LoadBalancingPolicy at a profile level

Unfortunately that won't work, policies are only resolved once at startup, changes in dynamic profiles will be ignored.

tolbertam:
> Unfortunately that won't work, policies are only resolved once at startup, changes in dynamic profiles will be ignored.

Ah, in that case we definitely don't want that then :)

olim7t (approved):

olim7t:
LGTM.

Leaving `DriverExecutionProfile` unchanged sounds reasonable to avoid impacting the public API.

Could you just add a quick javadoc to `DefaultDriverConfigLoader.builder()`? To explain that it's a way to override options programmatically.

olim7t:
PS: feel free to squash & merge at your convenience.

tolbertam:
Thanks for reviewing!

> Could you just add a quick javadoc to DefaultDriverConfigLoader.builder()? To explain that it's a way to override options programmatically.

Will do!

I'll squash and amend change log/commit message to match what was done here and will then merge.

PR #1033, opened 2018-06-14
JAVA-1819: Propagate more attributes to bound statements

olim7t:


core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
```diff
@@ -65,7 +65,7 @@ public DefaultDriverConfigLoader() {
 
   /**
    * Builds an instance with custom arguments, if you want to load the configuration from somewhere
-   * else or have custom options.
+   * else or have cusAwtom options.
```
tolbertam: Accidental?

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/BoundStatementIT.java
```diff
@@ -151,6 +164,68 @@ public void should_allow_custom_codecs_when_setting_values_in_bulk() {
     }
   }
 
+  @Test
+  public void should_propagate_attributes_when_preparing_a_simple_statement() {
```
tolbertam: Should add `@CassandraRequirement` for 2.2 since Custom Payload won't work with C* 2.1.  Either that or don't include custom payload if not supported by protocol version (similar to `supportsPerRequestKeyspace`)

tolbertam (approved):
Looks good!  Two small comments, but otherwise :+1:

olim7t:
Force-pushed to rebase on latest 4.x and take into account recent changes (most notably #1052).

tolbertam:
Awesome, taking a look after rebase.

core/src/main/java/com/datastax/oss/driver/api/core/CqlSession.java
```diff
@@ -77,6 +78,46 @@ default ResultSet execute(@NonNull String query) {
...
+   *         <li>{@link Request#getConfigProfileName() boundStatement.getConfigProfileName()}
+   *         <li>{@link Request#getConfigProfile() boundStatement.getConfigProfile()}
+   *         <li>{@link Statement#getPagingState() boundStatement.getPagingState()}
+   *         <li>{@link Request#getRoutingKey() boundStatement.getRoutingKey()}
+   *         <li>{@link Request#getRoutingToken() boundStatement.getRoutingToken()}
+   *         <li>{@link Request#getCustomPayload() boundStatement.getCustomPayload()}
+   *         <li>{@link Request#isIdempotent() boundStatement.isIdempotent()}
+   *         <li>{@link Request#getTimeout() boundStatement.getTimeout()}
+   *         <li>{@link Statement#getPagingState() boundStatement.getPagingState()}
+   *         <li>{@link Statement#getPageSize()} boundStatement.getPageSize()}
```
tolbertam: Nit: for these next 4, there is an extraneous `}` after `()` which i think would cause formatting issues.

core/src/main/java/com/datastax/oss/driver/api/core/CqlSession.java
```diff
@@ -110,11 +180,16 @@ default PreparedStatement prepare(@NonNull String query) {
...
    * sent, generally before the statement is prepared).
+   *
+   * <p>This variant is exposed in case you use an ad hoc {@link PrepareRequest} implementation to
+   * customize how attributes are propagated when you prepare a {@link SimpleStatement} (see {@link
+   * #prepare(SimpleStatement)} for more explanations). Otherwise, you should rarely have to deal
+   * with {@link PrepareRequest} directly.
    */
   @NonNull
-  default CompletionStage<PreparedStatement> prepareAsync(@NonNull SimpleStatement statement) {
+  default CompletionStage<PreparedStatement> prepareAsync(PrepareRequest request) {
```
tolbertam, line 193: Should add a `@NonNull` on `PrepareRequest request` right?

core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultPrepareRequest.java
```diff
@@ -127,7 +111,7 @@ public Token getRoutingToken() {
   @Nullable
   @Override
   public Duration getTimeout() {
-    return statement.getTimeout();
+    return null;
```
tolbertam, line 114: When I originally implemented this, I wasn't sure if it made sense to use the timeout from the `Request`.  I decided to since it currently grabs the timeout from the execution profile, so I figured that made sense here, but i agree that it's somewhat weird because typically prepare requests are fast / don't take any more time based on request.

tolbertam (approved):
A couple small nits, otherwise :+1:

PR #1035, opened 2018-06-15
JAVA-1872: Retain table's views when processing table update

tolbertam:
Motivation:

When a table update schema change event is received, the driver
creates a new TableMetadata instance for it and replaces the
existing one.

If that table has associated views, the driver currently does not
retain those views on the newly created TableMetadata instance.  This
change fixes that.

Modifications:

When a table update is received, update the base table on the views
tied to the old TableMetadata instance to the new one.  Added
MaterializedViewMetadata.setBaseTable to handle this.

Add SchemaChangesTest.should_retain_view_on_table_update to verify
this fix.

Result:

Views are now retained on table update schema events.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -1,5 +1,10 @@
 ## Changelog
 
+### 3.5.1
+
+- [bug] JAVA-1872: Retain table's views when processing table update.
+
+
 ### 3.5.0
 
 - [improvement] JAVA-1448: TokenAwarePolicy should respect child policy ordering.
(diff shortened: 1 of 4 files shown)


olim7t (approved):

tolbertam:
Thanks for reviewing, I just realized I should create a 3.5.x branch and change the base for his and the changelog.  Will do and then merge :+1:

tolbertam:
Done, the 3.5.x branch basically contains everything except c65e715 as all other recent changes have been documentation / CI related.

PR #1036, opened 2018-06-15
JAVA-1766: Annotate and document nullability in driver API

adutra:
TODO:
- [x] check vararg parameters (use double annotation?)


core/src/main/java/com/datastax/oss/driver/api/core/CqlSession.java
```diff
@@ -38,6 +38,9 @@ static CqlSessionBuilder builder() {
   /**
    * Executes a CQL statement synchronously (the calling thread blocks until the result becomes
    * available).
+   *
+   * @param statement The statement to execute; never {@code null}.
```
tolbertam: If we annotate the method as `@Nonnull`, is it necessary to document it?  Doesn't the annotation show up in the javadocs?

core/src/main/java/com/datastax/oss/driver/api/core/CqlSession.java
```diff
@@ -47,6 +50,9 @@ default ResultSet execute(@Nonnull Statement<?> statement) {
   /**
    * Executes a CQL statement synchronously (the calling thread blocks until the result becomes
    * available).
+   *
+   * @param query The query to execute; never {@code null}.
+   * @return The {@linkplain ResultSet results} returned by the server; never {@code null}.
```
tolbertam: I think we tend to omit `@param` / `@return` definitions if it is clear/obvious what the values mean as a way to reduce documentation noise.

core/src/main/java/com/datastax/oss/driver/api/core/data/GettableByIndex.java
```diff
@@ -160,9 +170,12 @@ default Object getObject(int i) {
   default boolean getBoolean(int i) {
     DataType cqlType = getType(i);
     TypeCodec<Boolean> codec = codecRegistry().codecFor(cqlType, Boolean.class);
-    return (codec instanceof PrimitiveBooleanCodec)
-        ? ((PrimitiveBooleanCodec) codec).decodePrimitive(getBytesUnsafe(i), protocolVersion())
-        : get(i, codec);
+    if (codec instanceof PrimitiveBooleanCodec) {
```
tolbertam, line 170: why change this?  It doesn't seem to change the behavior (nulls get changed to false in boolean conversion)

core/src/main/java/com/datastax/oss/driver/api/core/data/GettableByIndex.java
```diff
@@ -427,10 +469,13 @@ default Token getToken(int i) {
...
    * more complex list types, use {@link #get(int, GenericType)}.
    *
+   * @return the {@code i}th value as a Java map; may be empty but never {@code null}.
    * @throws IndexOutOfBoundsException if the index is invalid.
    */
+  @Nonnull
   default <T> List<T> getList(int i, Class<T> elementsClass) {
-    return get(i, GenericType.listOf(elementsClass));
+    List<T> value = get(i, GenericType.listOf(elementsClass));
+    return value == null ? new ArrayList<>(0) : value;
```
tolbertam: Hrmm, is this a hard requirement that it returns an empty list?  Shouldn't it be dependent on the registered codec?  The list codec will return nulls as empty collections, but I see that the 3.x code defers to the codec and doesn't make any assumptions otherwise:

https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/GettableByIndexData.java#L306-L309

tolbertam (commented):
:+1: in to the choice of what is marked `Optional`.  

I'm a bit conflicted on the use of the annotations. Having to mark parameters and return types as nullable / nonnullable seems tedious and creates a lot of noise in the code.

core/src/main/java/com/datastax/oss/driver/api/core/CqlSession.java
adutra: Yes it does show up. This is a leftover from a previous state where I was duplicating the information everywhere in the docs. I will remove, thanks for pointing that out.

core/src/main/java/com/datastax/oss/driver/api/core/data/GettableByIndex.java
adutra: There was a risk of NPE in the previous version:
```
    return (codec instanceof PrimitiveByteCodec)
        ? ((PrimitiveByteCodec) codec).decodePrimitive(getBytesUnsafe(i), protocolVersion())
        : get(i, codec);
```
The unboxing of `get(i, codec)` will throw NPE if that method returns `null` (which can happen depending on which codec has been selected).

core/src/main/java/com/datastax/oss/driver/api/core/data/GettableByIndex.java
adutra: We discussed yesterday about collections and I was under the impression that the general policy is to never return null, but rather empty collections for List, Set and Map. This change is required to comply with the `@Nonnull` annotations. But I don't have strong opinions here, changing the contract to `@Nullable` is fine too. Also, I'm returning regular java.util collections rather than immutable structures as I saw that the built-in codecs do the same (which I find a bit surprising btw).

olim7t: When we talked of empty collections, I was thinking more of session-level methods, such as `Metadata.getKeyspaces()`: return an empty collection even when schema metadata is disabled, instead of wrapping the collection in an `Optional`.

For row-level getters I think we should keep the current behavior of depending on the registered codec.

I kept the returned collections mutable because I felt it would trip users too much if you can't manipulate the result directly, and the risk of thread safety issues when you're dealing with a simple column value is pretty minimal.

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
```diff
@@ -75,6 +76,7 @@ private static String buildMessage(Map<Node, Throwable> errors) {
     return errors;
   }
 
+  @Nonnull
```
olim7t: Out of curiosity: how did you determine where to add the annotations? Is there any tooling, or did you just go over the whole codebase?
We want to keep using them consistently in new code, it would be nice to have a way to check automatically (ideally as part of the build).

core/src/main/java/com/datastax/oss/driver/api/core/CqlIdentifier.java
```diff
@@ -66,7 +67,7 @@
   // One exception is named getters, where we keep raw strings with the 3.x rules.
 
   /** Creates an identifier from its {@link CqlIdentifier CQL form}. */
-  public static CqlIdentifier fromCql(String cql) {
+  public static @Nonnull CqlIdentifier fromCql(@Nonnull String cql) {
```
olim7t: I know we've not normalized that yet, but I'm not a big fan of "sandwiching" the annotation between the modifiers and the return type. I find the following more readable, maybe because I'm more used to it:
```java
  @Nonnull
  public static CqlIdentifier fromCql(@Nonnull String cql) {
```
Any other opinions?

core/src/main/java/com/datastax/oss/driver/api/core/metadata/schema/IndexMetadata.java
```diff
@@ -19,25 +19,33 @@
...
+  @Nonnull
   String getTarget();
 
   /**
    * If this index is custom, the name of the server-side implementation. Otherwise, {@code null}.
    */
-  default String getClassName() {
-    return getOptions().get("class_name");
+  @Nonnull
+  default Optional<String> getClassName() {
```
olim7t: The javadoc should be updated too, it still says null.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/DefaultAggregateMetadata.java
```diff
@@ -114,7 +128,7 @@ public boolean equals(Object other) {
       AggregateMetadata that = (AggregateMetadata) other;
       return Objects.equals(this.keyspace, that.getKeyspace())
           && Objects.equals(this.signature, that.getSignature())
-          && Objects.equals(this.finalFuncSignature, that.getFinalFuncSignature())
+          && Objects.equals(this.finalFuncSignature, that.getFinalFuncSignature().orElse(null))
```
olim7t: Good catch. Should be the same for `initCond` on the next line.
It's a bit unnerving that this is not a compile-time error :(

EDIT -- I checked with a structural search: `Objects.equals($o1$, $o2$)` with a constraint on `o2`'s expression type `java\.util\.Optional`. This is the only occurrence.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricsFactory.java
```diff
@@ -60,19 +61,19 @@ public DropwizardMetricsFactory(InternalDriverContext context) {
...
       this.sessionUpdater = dropwizardSessionUpdater;
-      this.metrics = Optional.of(new DefaultMetrics(registry, dropwizardSessionUpdater));
+      this.metrics = new DefaultMetrics(registry, dropwizardSessionUpdater);
     }
   }
 
   @Override
   public Optional<Metrics> getMetrics() {
-    return metrics;
+    return Optional.ofNullable(metrics);
```
olim7t, line 76: Why not, I don't care either way since this method should not be called very often.
But for my understanding: any rationale for wrapping at the last minute rather of storing the optional as a field?

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/AggregateParserTest.java
```diff
@@ -76,12 +77,13 @@ public void should_parse_modern_table() {
         .containsExactly(DataTypes.INT, DataTypes.INT);
     assertThat(aggregate.getStateType()).isEqualTo(DataTypes.INT);
 
-    FunctionSignature finalFuncSignature = aggregate.getFinalFuncSignature();
-    assertThat(finalFuncSignature.getName().asInternal()).isEqualTo("to_string");
-    assertThat(finalFuncSignature.getParameterTypes()).containsExactly(DataTypes.INT);
+    Optional<FunctionSignature> finalFuncSignature = aggregate.getFinalFuncSignature();
+    assertThat(finalFuncSignature).isPresent();
+    assertThat(finalFuncSignature.get().getName().asInternal()).isEqualTo("to_string");
+    assertThat(finalFuncSignature.get().getParameterTypes()).containsExactly(DataTypes.INT);
```
olim7t: You can rewrite this in a slightly prettier way with `assertThat(finalFuncSignature).hasValueSatisfying(signature -> { ... })`.

olim7t (changes requested):
Thanks for separating the commits, it made it easy to review. Minor comments inline, only the one in `DefaultAggregateMetadata` is really serious.

I understand your concerns about verbosity @tolbertam, but on the other hand this is really valuable for developers with a modern IDE, so I'm willing to put in the extra effort in our code.

Similarly, `Optional` is annoying to deal with in client code, but it does force you to handle missing values correctly. I think we're striking a good middle ground in that PR.

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
adutra: Thee is no automatic tool to generate or suggest where to place annotations to my knowledge. IntelliJ can however suggest to propagate annotations from interfaces to implementations, but otherwise I've been crawling the entire codebase.

core/src/main/java/com/datastax/oss/driver/api/core/CqlIdentifier.java
adutra: I agree with you actually, this line somewhat was annotated differently from all the others.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/DefaultAggregateMetadata.java
adutra: On my side IntelliJ warns me that an equals comparison is being performed on incompatible types in such situations.

adutra: I guess the problem with `initCond` is that `initCond` is an `Object` so I had no warnings from IntelliJ.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricsFactory.java
adutra: Yes, it is not advised to store `Optional` instances as instance fields, mainly because `Optional` is not serializable.

adutra: I found a few other occurrences of `Optional` as instance fields. I will fix them in a separate commit, but if you prefer to keep things the way they are, we can just drop that commit later.

olim7t: I think if the field is final, the enclosing type is not meant to be serializable, and it is on the hot path, then we should store the optional as a field. Otherwise, I'm fine recreating it each time.
(so this particular example falls into the "rewrap on every call" category)

core/src/main/java/com/datastax/oss/driver/api/core/cql/ExecutionInfo.java
```diff
@@ -34,9 +36,11 @@
 public interface ExecutionInfo {
 
   /** The statement that was executed. */
+  @NonNull
   Statement<?> getStatement();
 
   /** The node that was used as a coordinator to successfully complete the query. */
+  @NonNull
   Node getCoordinator();
```
olim7t, line 44: This can actually be null under certain circumstances. My bad because I should have documented it in the first place, I will push a commit to address it.

olim7t: Nevermind, that's only after my changes in #1038

core/src/main/java/com/datastax/oss/driver/api/core/cql/Statement.java
```diff
@@ -66,54 +68,72 @@
    * <p>All the driver's built-in implementations are immutable, and return a new instance from this
    * method. However custom implementations may choose to be mutable and return the same instance.
    */
-  T setConfigProfileName(String newConfigProfileName);
+  @NonNull
+  T setConfigProfileName(@NonNull String newConfigProfileName);
```
olim7t: `newConfigProfileName` is nullable, it could be used to remove an existing name. Same for `setConfigProfile`.

core/src/main/java/com/datastax/oss/driver/api/core/cql/StatementBuilder.java
```diff
@@ -86,24 +91,28 @@ public T withRoutingKeyspace(CqlIdentifier routingKeyspace) {
    * Shortcut for {@link #withRoutingKeyspace(CqlIdentifier)
    * withRoutingKeyspace(CqlIdentifier.fromCql(routingKeyspaceName))}.
    */
-  public T withRoutingKeyspace(String routingKeyspaceName) {
+  @NonNull
+  public T withRoutingKeyspace(@NonNull String routingKeyspaceName) {
```
olim7t: Most arguments are nullable if you initialized the builder by copying another statement, and need to clear some of the attributes.

core/src/main/java/com/datastax/oss/driver/api/core/cql/TraceEvent.java
```diff
@@ -15,18 +15,21 @@
...
 
+import edu.umd.cs.findbugs.annotations.Nullable;
 import java.net.InetAddress;
 
 /** An event in a {@link QueryTrace}. */
 public interface TraceEvent {
 
   /** Which activity this event corresponds to. */
+  @Nullable
   String getActivity();
```
olim7t, line 26: This is read directly from Cassandra tables. In theory it shouldn't be null, but the data could be corrupted (it's possible to alter it manually), so I agree with marking it nullable.

core/src/main/java/com/datastax/oss/driver/api/core/loadbalancing/LoadBalancingPolicy.java
```diff
@@ -64,7 +65,8 @@ void init(
    * @return the list of coordinators to try. <b>This must be a concurrent queue</b>; {@link
    *     java.util.concurrent.ConcurrentLinkedQueue} is a good choice.
    */
-  Queue<Node> newQueryPlan(Request request, Session session);
+  @NonNull
+  Queue<Node> newQueryPlan(@NonNull Request request, @NonNull Session session);
```
olim7t: Both arguments nullable per the javadoc.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Node.java
```diff
@@ -57,12 +60,16 @@
...
   Optional<InetSocketAddress> getListenAddress();
 
+  @NonNull
   String getDatacenter();
 
+  @NonNull
   String getRack();
 
+  @NonNull
   Version getCassandraVersion();
```
olim7t: Datacenter, rack and version should be non-null in a healthy deployment, but the driver would still work if the data was corrupted. I'm going to mark them as nullable and add explanatory javadocs.

edit -- same for host id and schema version

core/src/main/java/com/datastax/oss/driver/api/core/metadata/schema/FunctionSignature.java
```diff
@@ -30,38 +32,45 @@
...
   }
 
-  public FunctionSignature(CqlIdentifier name, DataType... parameterTypes) {
-    this(name, ImmutableList.<DataType>builder().add(parameterTypes).build());
+  public FunctionSignature(@NonNull CqlIdentifier name, @Nullable DataType... parameterTypes) {
+    this(
+        name,
+        parameterTypes == null
+            ? ImmutableList.of()
+            : ImmutableList.<DataType>builder().add(parameterTypes).build());
```
olim7t: Small glitch here but this is more an issue in the initial implementation: if the constructor is called with no varargs, `parameterTypes` is empty, not null. I think it's impossible to get null, I'll fix the method and mark it as non nullable.

olim7t: > I think it's impossible to get null

Mmm actually it is possible 🙁 :
```java
public static void test(Object... args) {
  System.out.println(args == null ? "null" : args.length);
}
test((Object) null); // prints "1" (non-null array containing 1 null element)
test(null); // prints "null" (null array)
test((Object[]) null); // prints "null" (null array)
```

[This answer](https://stackoverflow.com/a/32327819/3775443) suggests a way to differentiate the two. Sounds a bit overkill but I'll check all vararg parameters in our codebase.

olim7t: The Spotbugs annotations do not declare `@Target(ElementType.TYPE_USE)`, which is needed for the "double annotation".
I'll just annotate element nullability and document array nullability in the javadocs.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -209,7 +219,8 @@ public SelfT withNodeFilter(Predicate<Node> nodeFilter) {
    * <p>Note that this can also be provided by the configuration; if both are defined, this method
    * takes precedence.
    */
-  public SelfT withKeyspace(CqlIdentifier keyspace) {
+  @NonNull
+  public SelfT withKeyspace(@NonNull CqlIdentifier keyspace) {
```
olim7t: Actually our integration tests rely on the keyspace being nullable 😃 
And it makes sense if you're injecting the values from some other part of your program where they can be null. It avoids having to surround every builder call with a null check:
```java
if (keyspace != null) {
  builder = builder.withKeyspace(keyspace);
}
```
I'll change this and a few other attributes in the class.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodec.java
```diff
@@ -184,7 +190,8 @@ default boolean accepts(DataType cqlType) {
    * If you choose not to implement this method, don't throw an exception but instead return a
    * constant string (for example "XxxCodec.format not implemented").
    */
-  String format(T value);
+  @Nullable
+  String format(@Nullable T value);
```
olim7t: All built-in codecs return a non-null value, and the request logger currently fails if this method returns null. I think it's fair to make it `@NonNull`.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/schema/FunctionSignature.java
olim7t: IDEA validates it the other way around compared to what's explained in the stackoverflow answer: `void foo(@NonNull Object... objects)` means that the vararg array itself can't be null. There is no way to indicate if array elements are nullable, so that has to go in the javadoc.

olim7t (approved):

olim7t:
Rebased on latest 4.x and squashed.

I've decided to squash everything together for this one: the initial commits were well separated (annotations vs `Optional`), but with the fixup commits on top of each other, it would have been hard to put everything back in its place. Since the two topics are somehow related (nullability), I don't feel too bad committing them together.

If we want to keep things separate in the future, ultimately the best way is to open separate PRs.

PR #1037, opened 2018-06-15
JAVA-1864: Add DataType.getTypeArguments()

olim7t:
As simple as it is, I'm having seconds thoughts about this. The same arguments as for other methods in the original ticket apply: it doesn't add any functionality that isn't already available, the client can do it on their side in about 20 lines of code and a few `instanceof` checks (and arguably more efficiently if they don't need to allocate collections). So why extend our API surface with it?

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -4,6 +4,7 @@
 
 ### 4.0.0-alpha4 (in progress)
 
+- [improvement] JAVA-1864: Add DataType.getTypeArguments()
 - [bug] JAVA-1560: Correctly propagate policy initialization errors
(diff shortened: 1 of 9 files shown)


tolbertam:
Hrmm, yeah you are probably right about that.  Another alternative would be to put the collection types (`ListType`, `SetType`, `UserType`) etc. under a `CollectionType` and then expose this method, but that may not be worth it.  I'm ok with whatever you think is best :+1:.

olim7t:
Closing. From the jira ticket:

These methods  are not very commonly needed, and they can easily be implemented from the client side with a few `instanceof` checks.
So in that case we choose API simplicity over providing a very minor convenience.

PR #1038, opened 2018-06-17
JAVA-1624: Expose ExecutionInfo on exceptions where applicable

olim7t:
New field: `DriverException.getExecutionInfo()`.

The challenge is that exceptions are sometimes thrown from deeper into the stack, where we don't have all the information to build the `ExecutionInfo`. To solve this, the field is set later, in `CqlRequestHandlerBase.setFinalError()`. This requires making it volatile which is not particularly elegant, but since we're dealing with exceptional conditions I think that's a fair compromise.

TODO
- [x] update nullability annotations (from #1036) on `ExecutionInfo` if needed

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/DriverException.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/DriverException.java
@@ -15,6 +15,8 @@
  */
 package com.datastax.oss.driver.api.core;
 
+import com.datastax.oss.driver.api.core.cql.ExecutionInfo;
+import com.datastax.oss.driver.api.core.cql.ResultSet;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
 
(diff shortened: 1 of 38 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/DriverException.java
```diff
@@ -33,8 +36,49 @@
...
+   *   <li>{@link ExecutionInfo#getPagingState()} and {@link ExecutionInfo#getTracingId()} will
+   *       always be null;
+   *   <li>{@link ExecutionInfo#getWarnings()} and {@link ExecutionInfo#getIncomingPayload()} will
+   *       always be empty;
+   *   <li>{@link ExecutionInfo#isSchemaInAgreement()} will always be true;
+   *   <li>{@link ExecutionInfo#getResponseSizeInBytes()} and {@link
+   *       ExecutionInfo#getCompressedResponseSizeInBytes()} will always be -1.
+   * </ul>
+   *
+   * <p>Note that this is only set for exceptions that are rethrown directly to the client from a
```
tolbertam, line 75: To make more clear that this means the value may be null, like

> do not contain their own execution info and therefore this method returns null in these cases.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
```diff
@@ -396,7 +415,7 @@ public void operationComplete(Future<java.lang.Void> future) throws Exception {
         Throwable error = future.cause();
         if (error instanceof EncoderException
             && error.getCause() instanceof FrameTooLongException) {
-          setFinalError(error.getCause(), node);
+          setFinalError(error.getCause(), node, execution);
```
tolbertam, line 420: Nice that this also tracks the execution # even in the failure case.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/session/ExceptionIT.java
```diff
@@ -0,0 +1,114 @@
...
+
+  private static String QUERY_STRING = "select * from foo";
+
+  @Before
+  public void clear() {
+    simulacron.cluster().clearLogs();
+  }
+
+  @Test
+  public void should_expose_execution_info_on_exceptions() {
```
tolbertam, line 63: Nice :+1:

tolbertam (approved):
Looks great :+1:

PR #1039, opened 2018-06-18
JAVA-1852: Document driver integration

olim7t:
TODO:
- [x] document dependencies: which can be excluded and under which conditions, why we made the annotations optional, etc.

### DIFF
--- a/manual/core/integration/README.md
+++ b/manual/core/integration/README.md
@@ -1,0 +1,418 @@
+## Integration
+
+This page contains various information on how to integrate the driver in your application.
+
+### Minimal project structure
+
+We publish the driver to [Maven central][central_oss]. Most modern build tools can download the
+dependency automatically.
+
+#### Maven
+
+Create the following 4 files:
+
+```
+$ find . -type f
+./pom.xml
+./src/main/resources/application.conf
+./src/main/resources/logback.xml
+./src/main/java/Main.java
+```
+
+##### Project descriptor
+
+`pom.xml` is the [Project Object Model][maven_pom] that describes your application. We declare the
+dependencies, and tell Maven that we're going to use Java 8:
+
+```xml
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.example.yourcompany</groupId>
+  <artifactId>yourapp</artifactId>
+  <version>1.0.0-SNAPSHOT</version>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.datastax.oss</groupId>
+      <artifactId>java-driver-core</artifactId>
+      <version>4.0.0-beta1</version>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
+      <version>1.2.3</version>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>1.8</source>
+          <target>1.8</target>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+</project>
+```
+
+##### Application configuration
+
+`application.conf` is not stricly necessary, but it illustrates an important point about the
+driver's [configuration](../configuration/): you override any of the driver's default options here.
+
+```
+datastax-java-driver {
+  basic.session-name = poc
+}
+```
+
+In this case, we just specify a custom name for our session, it will appear in the logs.
+
+##### Logging configuration
+
+For this example, we choose Logback as our [logging framework](../logging/) (we added the dependency
+in `pom.xml`). `logback.xml` configures it to send the driver's `INFO` logs to the console.
+
+```xml
+<configuration>
+  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+    <encoder>
+      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+    </encoder>
+  </appender>
+  <root level="WARN">
+    <appender-ref ref="STDOUT"/>
+  </root>
+  <logger name="com.datastax.oss.driver" level= "INFO"/>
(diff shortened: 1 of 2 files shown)


manual/core/integration/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+#### Manually (from the binary tarball)
+
+If your build tool can't fetch dependencies from Maven central, we publish a binary tarball on the 
+[DataStax download server][downloads].
+
+The driver and its dependencies must be in the compile-time classpath. Application resources, such
+as `application.conf` and `logback.xml` in our previous examples, must be in the runtime classpath.
+
+### Component lifecycles
```
olim7t: I was going to put this under a "best practices" header, but I don't have a lot of other topics to cover. Any idea?

manual/core/integration/README.md
```diff
@@ -0,0 +1,287 @@
+## Integration
```
tolbertam: Should add a link to beginning of manual/core/README.md that links here as that currently offers how to add dependency to maven, but would also be good to point here for more information.

tolbertam, line 1: Should add a link to beginning of manual/core/README.md that links here as that currently offers how to add dependency to maven, but would also be good to point here for more information.

manual/core/integration/README.md
```diff
@@ -0,0 +1,287 @@
...
+This page contains various information on how to integrate the driver in your application.
+
+### Minimal project structure
+
+We publish the driver to [Maven central][central_oss]. Most modern build tools can download the
+dependency automatically.
+
+#### Maven
+
+Create the following 4 files:
```
tolbertam, line 12: Maybe we should consider creating and publish maven archetype of our own to kickstart a project which creates all the relevant files via `mvn archetype:generate`.

manual/core/integration/README.md
```diff
@@ -0,0 +1,287 @@
...
+    mavenCentral()
+}
+
+dependencies {
+    compile group: 'com.datastax.oss', name: 'java-driver-core', version: '4.0.0-alpha4-SNAPSHOT'
+    compile group: 'ch.qos.logback', name: 'logback-classic', version: '1.2.3'
+}
+```
+
+Then place [application.conf](#application-conf), [logback.xml](#logback-xml) and
```
tolbertam: It looks like github markdown anchor links truncate the `.` instead of adding `-`, i.e. `#applicationconf`.  Is this a difference with our documentation tooling?

manual/core/integration/README.md
```diff
@@ -0,0 +1,287 @@
...
+:execute
+13:32:25.339 [main] INFO  c.d.o.d.i.c.DefaultMavenCoordinates - DataStax Java driver for Apache Cassandra(R) (com.datastax.oss:java-driver-core) version 4.0.0-alpha4-SNAPSHOT
+13:32:25.682 [poc-admin-0] INFO  c.d.o.d.internal.core.time.Clock - Using native clock for microsecond precision
+13:32:25.683 [poc-admin-0] INFO  c.d.o.d.i.c.metadata.MetadataManager - [poc] No contact points provided, defaulting to /127.0.0.1:9042
+3.11.2
+
+BUILD SUCCESSFUL
+```
+
+#### Manually (from the binary tarball)
```
tolbertam, line 225: Good you kept this section short.  Encourages the use of a build tool (maven/gradle) :+1:

manual/core/integration/README.md
tolbertam: We could probably expand this into a separate manual page for Best Practices if we have more we want to add, but I agree it makes sense to just leave this here as is for now.

tolbertam (approved):
Looks good to me!  Had a few small comments but :+1: overall.

manual/core/integration/README.md
```diff
@@ -0,0 +1,287 @@
...
+
+```java
+import com.datastax.oss.driver.api.core.CqlSession;
+import com.datastax.oss.driver.api.core.cql.ResultSet;
+
+public class Main {
+  public static void main(String[] args) {
+    try (CqlSession session = CqlSession.builder().build()) {
+      ResultSet rs = session.execute("SELECT release_version FROM system.local");
+      System.out.println(rs.one().getString(0));
```
adutra, line 113: Since you showed above how to configure Logback, maybe we could replace the call to `System.out` with a logger?

manual/core/integration/README.md
```diff
@@ -0,0 +1,287 @@
...
+in `pom.xml`). This file configures it to send the driver's `INFO` logs to the console.
+
+```xml
+<configuration>
+  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+    <encoder>
+      <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+    </encoder>
+  </appender>
+  <root level="ERROR">
```
adutra: Not sure if we should encourage users to adopt `ERROR` as the root log level.

adutra (approved):

manual/core/integration/README.md
olim7t: It is ☹️ 
I'll use description of the files as titles, instead of the names.

manual/core/integration/README.md
olim7t: Changed to WARN

manual/core/integration/README.md
olim7t: This is just a dummy class to illustrate the point, keeping it as simple as possible will be fine.

manual/core/integration/README.md
olim7t: Maybe, it might be cool to have generation options for the different setups: shaded or not, pick a logging framework, etc. On the other hand, that's another thing to maintain....

Feel free to open a ticket if you want to pursue the idea.

manual/core/integration/README.md
olim7t: I ended up removing this section, everything is covered already in other parts of the manual.

manual/core/integration/README.md
```diff
@@ -0,0 +1,418 @@
...
+#### Mandatory dependencies
+
+The remaining core driver dependencies are the only ones that are truly mandatory:
+
+* the [native protocol](https://github.com/datastax/native-protocol) layer. This is essentially part
+  of the driver code, but was externalized for reuse in other projects;
+* `java-driver-shaded-guava`, a shaded version of [Guava](https://github.com/google/guava). It is
+  relocated to a different package, and only used by internal driver code, so it should be
+  completely transparent to third-party code;
+* the [SLF4J](https://www.slf4j.org/) API for [logging](../logging/).
```
olim7t, line 408: I had never thought specifically about this, but it's pretty cool: you can run the driver with a minimal set of 3 dependencies, two of them being our internal stuff. SLF4J is the only one that could clash with 3rd-party code.

PR #1040, opened 2018-06-19
JAVA-1870: Add Component level configuration validation to the RequestLogger

GregBestland:


core/src/main/java/com/datastax/oss/driver/internal/core/tracker/RequestLogger.java
```diff
@@ -155,4 +168,21 @@ protected void logError(
       LOG.error("{} [{}]", builder.toString(), error.toString());
     }
   }
+
```
GregBestland: My only question here is should these be moved to a general utility for re-use?

tolbertam: Maybe this is too clever but we could generalize this to one method on `DriverConfigProfile` using a function variable for the method to call, i.e.:

```java
public interface DriverConfigProfile {
  // ...
  default <T> Optional<T> get(
      DriverOption option, BiFunction<DriverConfigProfile, DriverOption, T> getter) {
    return isDefined(option) ? Optional.of(getter.apply(this, option)) : Optional.empty();
  }
  // ...
}
```

And to use it:

```java
boolean successEnabled =
    configProfile
        .get(
            DefaultDriverOption.REQUEST_LOGGER_SUCCESS_ENABLED, DriverConfigProfile::getBoolean)
        .orElse(false);
```

tolbertam (commented):
Was reading the [jira ticket](https://datastax-oss.atlassian.net/browse/JAVA-1870) and it mentions configuration validation, where this change instead just falls back on defaults if they aren't specified.  I think I agree with that change but just wanted to make sure that is the route we want to go.  :+1: if that is the case, although had a suggestion/idea to generalize retrieving configuration with a fallback value.

olim7t:
> this change instead just falls back on defaults if they aren't specified. I think I agree with that change but just wanted to make sure that is the route we want to go.

Yes, that's what we decided to do eventually, because these options can be changed at runtime.
Only doing validation raised the question of how to handle the error when we start with a good configuration, but a required option is then removed at runtime. Falling back to a sane default is the simpler approach, and good enough for this use case.

core/src/main/java/com/datastax/oss/driver/internal/core/tracker/RequestLogger.java
olim7t: I don't like how this creates an `Optional` for every call. This is on the hot path because these options are re-read for each request.
Maybe we could have `DriverConfigOption.getBoolean(DriverOption option, boolean default)`, but then it makes sense to have it for every type. It can be done as default methods as you suggest, so why not.

tolbertam: To remove optional, tt could be redefined as:

```java
default <T> T get(
      DriverOption option, BiFunction<DriverConfigProfile, DriverOption, T> getter, T defaultValue) {
  return isDefined(option) ? getter.apply(this, option) : defaultvalue;
}
```

But that is more of an effort to reduce adding another method for every possible return type.  Really wish java supported default arguments :rage:

core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOptionUtil.java
```diff
@@ -0,0 +1,227 @@
...
+   * Fetch a boolean from the {@link DriverConfigProfile} if it exists, otherwise return provided
+   * default.
+   *
+   * @param configProfile to fetch from.
+   * @param option to fetch.
+   * @param defaultValue fallback value if the option doesn't exist.
+   * @return the specified boolean option, or the default if it's not defined.
+   */
+  public static boolean getConfigBooleanIfDefined(
+      DriverConfigProfile configProfile, DefaultDriverOption option, boolean defaultValue) {
```
olim7t: My suggestion was to add those directly to `DriverConfigProfile`, as default methods: `getBoolean(DriverOption option, boolean default)`, etc.

olim7t (changes requested):

GregBestland:
@olim7t I think your suggested changes should be addressed. I also rebased. If this looks correct I will sqaush and commit.

olim7t (approved):

olim7t:
👍 merged. I changed a few details directly:
* define the new getters as default methods on the interface
* declare the default value -- and therefore the result of the method -- `@Nullable`. This makes sense for scalar values (the driver does it for `String` in one place), not so much for collections but I did it everywhere for consistency.

PR #1041, opened 2018-06-19
 JAVA-1763: Generate a binary tarball as part of the build process

olim7t:
To test:
```
mvn package -DskipTests -Prelease
```

1) There are a few non-fatal error during the build:
    ```
    [ERROR] no module descriptor for com.datastax.oss:java-driver-distribution
    [ERROR] no module descriptor for com.datastax.oss:java-driver-core
    [ERROR] no module descriptor for com.datastax.oss:java-driver-query-builder
    ```
    I don't understand because this seems related to Jigsaw, but I'm building with Java 8.

    Plus a few javadoc warnings related to optional / shaded dependencies, I haven't spent time investigating them either.

2) As crazy as it sounds, I wasn't able to reuse the same configuration as driver 3. If I exclude the module from its own `dependencySet` (as done in driver 3), transitive dependencies -- like `netty-buffer` -- are missing. If I leave it, transitive dependencies work but the module's JAR is also duplicated (e.g. `driver-core.jar` both at the root level and in `lib/core`). I was using the exact same plugin definition and assembly descriptor, so maybe I'm missing an extra detail somewhere else.

    I ended up putting all the binaries for each module in the same directory, including the module itself. That is, `driver-core.jar` is in `lib/core` mixed with all its dependencies, instead of being at the root. It's not as pretty, but frankly I've reached the limit of the effort I want to put into this, especially considering that the audience for the tarball is probably a minimal percentage of our users.

    If someone wants to give it a shot, feel free to try and good luck.

TODO:
- [ ] include API docs in tar
- [x] create LICENSE

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -1,0 +1,160 @@
+<!--
+
+    Copyright DataStax, Inc.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.datastax.oss</groupId>
+    <artifactId>java-driver-parent</artifactId>
+    <version>4.0.0-beta1-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>java-driver-distribution</artifactId>
+  <!-- Should be pom but Javadoc generation requires a "classpath-capable" package -->
+  <packaging>jar</packaging>
+
+  <name>DataStax Java driver for Apache Cassandra(R) - binary distribution</name>
+
+  <!--
+    These dependencies are only here to ensure proper build order and proper inclusion of binaries
+    in the final tarball
+  -->
+  <dependencies>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>java-driver-core</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>java-driver-query-builder</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <!--
+      The annotations are optional dependencies, but it's nice to show annotations in the javadocs
+      so redeclare them.
+    -->
+    <dependency>
+      <groupId>com.github.stephenc.jcip</groupId>
+      <artifactId>jcip-annotations</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.github.spotbugs</groupId>
+      <artifactId>spotbugs-annotations</artifactId>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <finalName>datastax-java-driver-${project.version}</finalName>
+    <plugins>
+      <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
+        <!-- http://stackoverflow.com/questions/13218313/unable-to-disable-generation-of-empty-jar-maven-jar-plugin -->
+        <executions>
+          <execution>
+            <id>default-jar</id>
+            <phase>none</phase>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <artifactId>maven-source-plugin</artifactId>
+        <configuration>
+          <skipSource>true</skipSource>
+        </configuration>
+      </plugin>
+      <plugin>
+        <artifactId>maven-install-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
+        </configuration>
+      </plugin>
+      <plugin>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
+  <profiles>
+    <profile>
+      <id>release</id>
+      <build>
+        <plugins>
+          <plugin>
+            <artifactId>maven-javadoc-plugin</artifactId>
+            <executions>
+              <execution>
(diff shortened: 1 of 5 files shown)


distribution/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.datastax.oss</groupId>
+    <artifactId>java-driver-parent</artifactId>
+    <version>4.0.0-alpha4-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>java-driver-distribution</artifactId>
+  <!-- Should be pom but Javadoc generation requires a "classpath-capable" package -->
```
tolbertam: I appreciate that you added comments to explain the rationale for specific configuration :+1:

tolbertam:
> I ended up putting all the binaries for each module in the same directory, including the module itself. That is, driver-core.jar is in lib/core mixed with all its dependencies, instead of being at the root. It's not as pretty, but frankly I've reached the limit of the effort I want to put into this, especially considering that the audience for the tarball is probably a minimal percentage of our users.

I agree that it's ok to just put the driver jars in the directories with their dependencies.  It's not a big deal to me that they aren't in the root.  The user ultimately needs the dependencies too :)

distribution/src/assembly/binary-tarball.xml
```diff
@@ -0,0 +1,133 @@
...
+      <directory>target/apidocs</directory>
+      <outputDirectory>apidocs</outputDirectory>
+    </fileSet>
+
+    <fileSet>
+      <directory>..</directory>
+      <outputDirectory>.</outputDirectory>
+      <includes>
+        <include>README*</include>
+        <include>LICENSE*</include>
```
tolbertam, line 111: I noticed that there was no `LICENSE` file in the tar, but then I realized we don't have a `LICENSE` file currently present for 4.x, we should add that.

distribution/pom.xml
```diff
@@ -0,0 +1,140 @@
...
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>java-driver-core</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>java-driver-query-builder</artifactId>
+      <version>${project.version}</version>
+    </dependency>
```
tolbertam, line 49: should we also include `java-driver-core-shaded`?  I noticed we currently don't include the shaded jar for 3.x's distributable, so I think it's fine to also not include it here.   If the user needs (which is more an exception) they can grab it off maven.

distribution/src/assembly/binary-tarball.xml
```diff
@@ -0,0 +1,133 @@
...
+      </binaries>
+    </moduleSet>
+
+  </moduleSets>
+
+  <fileSets>
+
+    <fileSet>
+      <directory>target/apidocs</directory>
+      <outputDirectory>apidocs</outputDirectory>
```
tolbertam, line 103: it doesn't look like `apidocs` doesn't end up in the tar.gz file from what I can tell.  I see it in the target directory, but not in the tar file.

tolbertam: Maybe it has something to do with how i'm running the build?  `mvn clean install -DskipTests -Prelease`

tolbertam:
Where these the javadoc warnings you were referring to?

```
6 warnings
[WARNING] Javadoc Warnings
[WARNING] /home/tolbertam/Documents/IdeaProjects/java-driver-4/core/src/main/java/com/datastax/oss/driver/internal/core/protocol/Lz4Compressor.java:22: error: package net.jpountz.lz4 does not exist
[WARNING] import net.jpountz.lz4.LZ4Compressor;
[WARNING] ^
[WARNING] /home/tolbertam/Documents/IdeaProjects/java-driver-4/core/src/main/java/com/datastax/oss/driver/internal/core/protocol/Lz4Compressor.java:23: error: package net.jpountz.lz4 does not exist
[WARNING] import net.jpountz.lz4.LZ4Factory;
[WARNING] ^
[WARNING] /home/tolbertam/Documents/IdeaProjects/java-driver-4/core/src/main/java/com/datastax/oss/driver/internal/core/protocol/Lz4Compressor.java:24: error: package net.jpountz.lz4 does not exist
[WARNING] import net.jpountz.lz4.LZ4FastDecompressor;
[WARNING] ^
[WARNING] /home/tolbertam/Documents/IdeaProjects/java-driver-4/core/src/main/java/com/datastax/oss/driver/internal/core/protocol/Lz4Compressor.java:33: error: cannot find symbol
[WARNING] private final LZ4Compressor compressor;
[WARNING] ^
[WARNING] symbol:   class LZ4Compressor
[WARNING] location: class Lz4Compressor
[WARNING] /home/tolbertam/Documents/IdeaProjects/java-driver-4/core/src/main/java/com/datastax/oss/driver/internal/core/protocol/Lz4Compressor.java:34: error: cannot find symbol
[WARNING] private final LZ4FastDecompressor decompressor;
[WARNING] ^
[WARNING] symbol:   class LZ4FastDecompressor
[WARNING] location: class Lz4Compressor
[WARNING] /home/tolbertam/Documents/IdeaProjects/java-driver-4/core/src/main/java/com/datastax/oss/driver/internal/core/protocol/SnappyCompressor.java:23: error: package org.xerial.snappy does not exist
[WARNING] import org.xerial.snappy.Snappy;
[WARNING] ^
[INFO] Building jar: /home/tolbertam/Documents/IdeaProjects/java-driver-4/distribution/target/datastax-java-driver-4.0.0-alpha4-SNAPSHOT-javadoc.jar
[INFO] 
[INFO] --- maven-gpg-plugin:1.5:sign (sign-artifacts) @ java-driver-distribution ---
```

Looks like not much of a worry since we aren't making docs for internal packages anyways, :+1:

tolbertam (commented):
Looks mostly good, only two issues:

* apidocs missing from generated tar.
* LICENSE file missing because it doesn't currently exist.

Otherwise :+1:

distribution/src/assembly/binary-tarball.xml
olim7t: That should work, I'll take a look at it.

olim7t:
> Where these the javadoc warnings you were referring to?

Yes. Indeed they are for internal packages.

distribution/pom.xml
olim7t: Since nobody asked for it on 3.x, leaving it out for now for simplicity.

distribution/pom.xml
```diff
@@ -0,0 +1,160 @@
...
+          <plugin>
+            <artifactId>maven-javadoc-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>dependencies-javadoc</id>
+                <!--
+                  For some reason this runs after the assembly plugin if bound to package, so bind
+                  to an earlier phase
+                -->
+                <phase>process-classes</phase>
```
olim7t, line 112: This takes care of `apidocs`. Another one of those things that was exactly the same as in 3.x, but didn't work 🤷‍♂️

PR #1042, opened 2018-06-21
JAVA-1883: Use custom queue implementation for LBP's query plan

olim7t:


core/src/main/java/com/datastax/oss/driver/internal/core/util/collection/QueryPlan.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  /**
+   * {@inheritDoc}
+   *
+   * <p>The returned iterator reflects the state of the queue at the time of the call, and is not
+   * affected by further modifications.
+   */
+  @Override
+  public Iterator<Node> iterator() {
```
tolbertam: From what I can tell, we never use `iterator()` on `QueryPlan` coming back from `newQueryPlan`, so this implementation would ultimately not be used and is more for completeness of implementation right?

tolbertam (approved):
:+1:

core/src/main/java/com/datastax/oss/driver/internal/core/util/collection/QueryPlan.java
olim7t: I didn't check but it might be used in tests. Also, `AbstractCollection.toString` relies on it, and you need `toString` for debugging or logs.

PR #1043, opened 2018-06-21
JAVA-1884: Add additional methods from TypeToken to GenericType

tolbertam:
For [JAVA-1884](https://datastax-oss.atlassian.net/browse/JAVA-1884).  I think this is generally useful and is simply a wrapper over `TypeToken.getRawType()`.  I also noticed we were missing `getSuperType()`, `getSubType`, and `getType`, but wasn't sure if we should add those (although would be simple to do), what do you think?

TODO:
- [ ] update commit subject and changelog entry

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/type/reflect/GenericType.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/type/reflect/GenericType.java
@@ -25,6 +25,7 @@ import com.datastax.oss.driver.shaded.guava.common.reflect.TypeParameter;
 import com.datastax.oss.driver.shaded.guava.common.reflect.TypeResolver;
 import com.datastax.oss.driver.shaded.guava.common.reflect.TypeToken;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
 import java.lang.reflect.Type;
 import java.math.BigDecimal;
 import java.math.BigInteger;
@@ -187,6 +188,14 @@ public class GenericType<T> {
     return token.isSubtypeOf(type.token);
   }
 
+  /**
+   * Returns true if this type is known to be an array type, such as {@code int[]}, {@code T[]},
+   * {@code <? extends Map<String, Integer>[]>} etc.
+   */
+  public final boolean isArray() {
+    return token.isArray();
+  }
+
   /** Returns true if this type is one of the nine primitive types (including {@code void}). */
   public final boolean isPrimitive() {
     return token.isPrimitive();
@@ -196,6 +205,7 @@ public class GenericType<T> {
    * Returns the corresponding wrapper type if this is a primitive type; otherwise returns {@code
    * this} itself. Idempotent.
    */
+  @NonNull
   public final GenericType<T> wrap() {
     if (isPrimitive()) {
       return new GenericType<>(token.wrap());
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/type/reflect/GenericType.java
```diff
@@ -226,6 +226,24 @@ public final boolean isPrimitive() {
...
+   * <ul>
+   *   <li>If {@code T} is a {@code Class} itself, {@code T} itself is returned.
+   *   <li>If {@code T} is a parameterized type, the raw type of the parameterized type is returned.
+   *   <li>If {@code T} is an array type , the returned type is the corresponding array class. For
+   *       example: {@code List<Integer>[] => List[]}.
+   *   <li>If {@code T} is a type variable or a wildcard type, the raw type of the first upper bound
+   *       is returned. For example: {@code <X extends Foo> => Foo}.
+   * </ul>
+   */
+  public Class<?> getRawType() {
```
stamhankar999: Shouldn't this be `<? super T>`?

stamhankar999 (changes requested):

core/src/main/java/com/datastax/oss/driver/api/core/type/reflect/GenericType.java
tolbertam: you are right, good catch!  I'll fix it.

tolbertam:
As discussed elsewhere, we decided we might as well add the other methods since it's low effort

tolbertam:
Added the following methods overall:


* getRawType
* getComponentType
* getSuperType
* getSubType
* getType
* isArray

olim7t (approved):
LGTM 👍 
Can you rebase on 4.x and add the `Nullable` / `NonNull` annotations?

tolbertam:
> Can you rebase on 4.x and add the Nullable / NonNull annotations?

:+1: will do!

core/src/main/java/com/datastax/oss/driver/api/core/type/reflect/GenericType.java
```diff
@@ -251,6 +252,7 @@ public final boolean isPrimitive() {
    * Returns the array component type if this type represents an array ({@code int[]}, {@code T[]},
    * {@code <? extends Map<String, Integer>[]>} etc.), or else {@code null} is returned.
    */
+  @Nullable
   @SuppressWarnings("unchecked")
   public final GenericType<?> getComponentType() {
     return new GenericType(token.getComponentType());
```
olim7t: This produces a warning because `new GenericType(...)` is never null. Which is actually the symptom of a bug: `token.getComponentType()` can be null, this must be handled.
I'm fixing it directly.

olim7t:
@stamhankar999 if all your comments have been addressed, can you approve the PR?

core/src/main/java/com/datastax/oss/driver/api/core/type/reflect/GenericType.java
tolbertam: Ah right, I didn't even consider that, good catch!

stamhankar999 (approved):
Looks great!

olim7t:
I forgot to change the commit message 🤦‍♂️ 
It's OK, it was not that far off anyway.

PR #1044, opened 2018-06-22
JAVA-1891: Allow null items in TupleValue

stamhankar999:


core/src/main/java/com/datastax/oss/driver/internal/core/data/ValuesHelper.java
```diff
@@ -47,7 +47,9 @@
     for (int i = 0; i < values.length; i++) {
       Object value = values[i];
       ByteBuffer encodedValue;
-      if (value instanceof Token) {
+      if (value == null) {
+        encodedValue = null;
+      } else if (value instanceof Token) {
```
olim7t: This works because all of our built-in codecs encode `null` as a CQL `null`.

But in theory someone could plug their own codec registry, with default codecs that handle null differently. So out of principle I would prefer letting the codec handle it. As suggested on the JIRA ticket, changing line 64 to:
```java
TypeCodec<Object> codec =
    (value == null)
        ? codecRegistry.codecFor(fieldTypes.get(i))
        : codecRegistry.codecFor(fieldTypes.get(i), value);
encodedValue = codec.encode(value, protocolVersion);
```

olim7t: I think `encodePreparedValues` has the same issue, this would be apparent in `PreparedStatement.bind(Object...)`.

olim7t (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/data/ValuesHelper.java
stamhankar999: Ahh, this totally makes sense. Will make the appropriate changes and resubmit for review.

olim7t:
Nit: we use the convention `java1891` for branch names (it's the fastest to type 😄). It's no big deal, but try to remember next time.

olim7t (approved):

PR #1046, opened 2018-06-27
JAVA-1877: Use a separate reconnection schedule for the control connection

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/connection/ReconnectionPolicy.java
```diff
@@ -15,23 +15,41 @@
...
  *
- * <p>Each time a connection to a node is lost, a {@link #newSchedule() new schedule} instance gets
- * created. Then {@link ReconnectionSchedule#nextDelay()} will be called each time the driver needs
- * to schedule the next connection attempt. When the node is back to its required number of
- * connections, the schedule will be reset (that is, the next failure will create a fresh schedule
- * instance).
+ * <p>Each time a reconnection starts, a new {@link ReconnectionSchedule} instance gets created.
+ * Then {@link ReconnectionSchedule#nextDelay()} will be called each time the driver needs to
+ * program the next connection attempt. When the reconnection succeeds, the schedule is discarded,
+ * and the next reconnection will start from a fresh instance.
```
adutra: I found myself slightly preferring the old rephrasing, would you mind adding a bit more context: "When the reconnection succeeds, the schedule is discarded; if the connection is lost again, the next reconnection attempt will query the policy again to obtain a new schedule".

core/src/main/java/com/datastax/oss/driver/internal/core/connection/ConstantReconnectionPolicy.java
```diff
@@ -19,16 +19,23 @@
...
 import com.datastax.oss.driver.api.core.connection.ReconnectionPolicy;
 import com.datastax.oss.driver.api.core.context.DriverContext;
+import com.datastax.oss.driver.api.core.metadata.Node;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.time.Duration;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** A reconnection policy that waits a constant time between each reconnection attempt. */
 public class ConstantReconnectionPolicy implements ReconnectionPolicy {
```
adutra: This implementation interrogates `DefaultDriverOption.RECONNECTION_BASE_DELAY` internally. Should we mention in the class javadocs which settings can affect this policy's behavior so that users know which settings to customize when using this policy? IOW, users trying to figure out how to use some specific policy by simply reading the javadocs will have trouble finding how to configure it properly. (This question is btw also valid for the exponential policy below, and for all policies in general).

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/Reconnection.java
```diff
@@ -50,13 +51,13 @@
 
   private final String logPrefix;
   private final EventExecutor executor;
-  private final ReconnectionPolicy reconnectionPolicy;
+  private final Supplier<ReconnectionSchedule> scheduleSupplier;
```
adutra, line 54: Love the `Supplier<ReconnectionSchedule>`, very elegant 👍 .

adutra (approved):
Nice job, had only a few trivial remarks about the javadocs.

core/src/main/java/com/datastax/oss/driver/internal/core/connection/ConstantReconnectionPolicy.java
olim7t: 👍 I'll revisit the policies to make sure that they all show a small config example.

PR #1047, opened 2018-07-05
JAVA-1897: Improve extensibility of schema metadata classes

olim7t:
Motivation:

Custom extensions of the server might add new fields to schema system
tables, calling for dedicated classes, e.g. MyCustomKeyspaceMetadata.
If Metadata methods declare exact types, converting to those custom
types will require a cast.

Modifications:

Use bounded types for all Metadata methods, e.g.

Map<CqlIdentifier, ? extends KeyspaceMetadata> getKeyspaces();

Adapt internal classes so that custom schema parsing logic can be
plugged in more easily.

Result:

Custom extensions can now extend the interfaces to specialize the types:

interface MyCustomSession extends Session {
  MyCustomMetadata getMetadata();
}

interface MyCustomMetadata extends Metadata {
  Map<CqlIdentifier, ? extends MyCustomKeyspaceMetadata> getKeyspaces();
}

The client doesn't need to cast anymore.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/CassandraSchemaParser.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/CassandraSchemaParser.java
@@ -1,0 +1,182 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.internal.core.metadata.schema.parsing;
+
+import com.datastax.oss.driver.api.core.CqlIdentifier;
+import com.datastax.oss.driver.api.core.metadata.schema.AggregateMetadata;
+import com.datastax.oss.driver.api.core.metadata.schema.FunctionMetadata;
+import com.datastax.oss.driver.api.core.metadata.schema.FunctionSignature;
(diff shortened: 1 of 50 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/CassandraSchemaParser.java
```diff
@@ -0,0 +1,182 @@
...
+    this.viewParser = new ViewParser(rows, context);
+    this.functionParser = new FunctionParser(rows.dataTypeParser(), context);
+    this.aggregateParser = new AggregateParser(rows.dataTypeParser(), context);
+  }
+
+  @Override
+  public SchemaRefresh parse() {
+    ImmutableMap.Builder<CqlIdentifier, KeyspaceMetadata> keyspacesBuilder = ImmutableMap.builder();
+    for (AdminRow row : rows.keyspaces()) {
+      KeyspaceMetadata keyspace = parseKeyspace(row);
```
tolbertam, line 74: To be consistent with other type parsing, why not also create a `KeyspaceParser`?

tolbertam (approved):
:+1: I think this approach works good.  I also prefer this approach over using generics.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/RawColumn.java
```diff
@@ -137,9 +127,9 @@ private RawColumn(
   public int compareTo(@NonNull RawColumn that) {
     // First, order by kind. Then order partition key and clustering columns by position. For
     // other kinds, order by column name.
-    if (this.kind != that.kind) {
+    if (!this.kind.equals(that.kind)) {
       return this.kind.compareTo(that.kind);
-    } else if (kind == Kind.PARTITION_KEY || kind == Kind.CLUSTERING_COLUMN) {
+    } else if (kind.equals(KIND_PARTITION_KEY) || kind.equals(KIND_CLUSTERING_COLUMN)) {
```
tolbertam, line 117: One slight difference, the previous implementation used `equalsIgnoreCase` when determining what `Kind` should be mapped to.  Here we are using `equals`.  It doesn't look like in the C* metadata that anything other then all lowercase was ever used, so not sure if that was ever needed to begin with.

PR #1048, opened 2018-07-06
JAVA-1880: Rename "config profile" to "execution profile"

olim7t:


adutra (approved):

olim7t:
Rebased on latest 4.x, this should be good to go but waiting for @tolbertam's review.
There's a test failure on `RequestLoggerIT` that I need to look into (can't reproduce in IDE), but it's not related to this changeset.

tolbertam (approved):
Looks good :+1:

PR #1049, opened 2018-07-09
JAVA-1908: TableMetadata.asCQLQuery does not add table option 'memtable_flush_period_in_ms' in the generated query

hkbharath:
Adding memtable_flush_period_in_ms option in table metadata returned from asCQLQuery from tableMetaData.


hkbharath:
How can i get this change reviewed? Please help

driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
```diff
@@ -280,6 +280,7 @@ protected StringBuilder appendOptions(StringBuilder sb, boolean formatted) {
         if (cassandraVersion.getMajor() > 3 || (cassandraVersion.getMajor() == 3 && cassandraVersion.getMinor() >= 8)) {
             and(sb, formatted).append("cdc = ").append(options.isCDC());
         }
+		and(sb, formatted).append("memtable_flush_period_in_ms = ").append(options.getMemtableFlushPeriodInMs());
```
adutra: This option has been introduced in Cassandra 2.0 by CASSANDRA-4237.

So i would suggest that you only include it when the version is 2.0 or higher:

```
        if (cassandraVersion.getMajor() > 1) {
            and(sb, formatted).append("memtable_flush_period_in_ms = ").append(options.getMemtableFlushPeriodInMs());
        }
```

adutra (changes requested):
Please address my comment and also add a test for your changes in `com.datastax.driver.core.TableMetadataTest#should_parse_table_options`. Thanks!

adutra:
Also, as per our [contribution guidelines](https://github.com/datastax/java-driver/blob/3.x/CONTRIBUTING.md) could you please add an entry to the changelog for this? Thanks!

driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
hkbharath: Thanks for the suggestion. Added suggested changes.

tolbertam:
Changed this to 3.6.0 as I think it's reasonable to include in 3.6.0 as this is a bugfix and we still have time before releasing 3.6.0.

tolbertam (approved):
Looks great to me, thanks! :+1:

adutra (approved):

adutra:
@hkbharath I fixed a few small nits manually, squashed and rebased your changes on top of 3.x as 5fa08fba41ec031713b2a7a4841245b3d689cbde. Thank you for your contribution, highly appreciated! 👍

hkbharath:
Thanks @adutra .

PR #1050, opened 2018-07-09
JAVA-1899 Support Virtual Tables

GregBestland:
Java1899 for 3.x

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
@@ -199,6 +199,15 @@ public abstract class AbstractTableMetadata {
         return options;
     }
 
+    /**
+     * Returns whether or not this keyspace is a virtual keyspace
+     * @return {@code true} if virtual keyspace
+     * default), {@code false} otherwise.
+     */
+    public boolean isVirtual(){
+        return getKeyspace().isVirtual();
+    }
+
     void add(ColumnMetadata column) {
         columns.put(column.getName(), column);
     }
--- a/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
@@ -17,7 +17,6 @@ package com.datastax.driver.core;
 
 
 import com.google.common.base.Predicate;
-import com.google.common.collect.Iterables;
 
 import java.util.*;
 
@@ -200,10 +199,11 @@ public abstract class AbstractTableMetadata {
     }
 
     /**
-     * Returns whether or not this keyspace is a virtual keyspace
+     * Returns whether or not this table is a virtual table
+     *
      * @return {@code true} if virtual keyspace, {@code false} otherwise.
      */
-    public boolean isVirtual(){
+    public boolean isVirtual() {
         return getKeyspace().isVirtual();
     }
 
@@ -253,7 +253,7 @@ public abstract class AbstractTableMetadata {
 
     protected StringBuilder appendOptions(StringBuilder sb, boolean formatted) {
         // Options
-        if(options==null){
+        if (options == null) {
             return sb;
         }
         sb.append("WITH ");
@@ -297,7 +297,7 @@ public abstract class AbstractTableMetadata {
 
     @Override
     public String toString() {
-        if(keyspace.isVirtual()){
+        if (keyspace.isVirtual()) {
             return name;
         }
 
--- a/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
@@ -286,6 +286,10 @@ public abstract class AbstractTableMetadata {
 
     @Override
     public String toString() {
+        if(keyspace.isVirtual()){
+            return name;
+        }
+
         return asCQLQuery();
     }
 
--- a/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
@@ -254,6 +254,9 @@ public abstract class AbstractTableMetadata {
 
     protected StringBuilder appendOptions(StringBuilder sb, boolean formatted) {
         // Options
+        if(options==null){
+            return sb;
+        }
         sb.append("WITH ");
         if (options.isCompactStorage())
             and(sb.append("COMPACT STORAGE"), formatted);
--- a/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
@@ -201,8 +201,7 @@ public abstract class AbstractTableMetadata {
 
     /**
      * Returns whether or not this keyspace is a virtual keyspace
-     * @return {@code true} if virtual keyspace
-     * default), {@code false} otherwise.
+     * @return {@code true} if virtual keyspace, {@code false} otherwise.
      */
     public boolean isVirtual(){
         return getKeyspace().isVirtual();
--- a/driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
@@ -371,12 +371,15 @@ public class KeyspaceMetadata {
      * @see #exportAsString
      */
     public String asCQLQuery() {
-        if(virtual){
-            return "";
-        }
         StringBuilder sb = new StringBuilder();
+        if(virtual){
+            sb.append("/* VIRTUAL ");
+        }
+        else{
+            sb.append("CREATE ");
+        }
 
-        sb.append("CREATE KEYSPACE ").append(Metadata.quoteIfNecessary(name)).append(" WITH ");
+        sb.append("KEYSPACE ").append(Metadata.quoteIfNecessary(name)).append(" WITH ");
         sb.append("REPLICATION = { 'class' : '").append(replication.get("class")).append('\'');
         for (Map.Entry<String, String> entry : replication.entrySet()) {
             if (entry.getKey().equals("class"))
@@ -385,6 +388,9 @@ public class KeyspaceMetadata {
         }
         sb.append(" } AND DURABLE_WRITES = ").append(durableWrites);
         sb.append(';');
+        if(virtual){
+            sb.append("*/");
+        }
         return sb.toString();
     }
 
--- a/driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
@@ -35,6 +35,7 @@ public class KeyspaceMetadata {
 
     private final String name;
     private final boolean durableWrites;
+    private final boolean isVirtual;
 
     private final ReplicationStrategy strategy;
     private final Map<String, String> replication;
@@ -51,6 +52,15 @@ public class KeyspaceMetadata {
         this.durableWrites = durableWrites;
         this.replication = replication;
         this.strategy = ReplicationStrategy.create(replication);
+        this.isVirtual=false;
+    }
+    @VisibleForTesting
+    KeyspaceMetadata(String name, boolean durableWrites, Map<String, String> replication, boolean isVirtual) {
+        this.name = name;
+        this.durableWrites = durableWrites;
+        this.replication = replication;
+        this.strategy = ReplicationStrategy.create(replication);
+        this.isVirtual=isVirtual;
     }
 
     static KeyspaceMetadata build(Row row, VersionNumber cassandraVersion) {
@@ -61,12 +71,16 @@ public class KeyspaceMetadata {
             replicationOptions = new HashMap<String, String>();
             replicationOptions.put("class", row.getString(STRATEGY_CLASS));
             replicationOptions.putAll(SimpleJSONParser.parseStringMap(row.getString(STRATEGY_OPTIONS)));
-            return new KeyspaceMetadata(name, durableWrites, replicationOptions);
+            return new KeyspaceMetadata(name, durableWrites, replicationOptions, false);
         } else {
             String name = row.getString(KS_NAME);
             boolean durableWrites = row.getBool(DURABLE_WRITES);
-            return new KeyspaceMetadata(name, durableWrites, row.getMap(REPLICATION, String.class, String.class));
+            return new KeyspaceMetadata(name, durableWrites, row.getMap(REPLICATION, String.class, String.class), false);
         }
+    }
+    static KeyspaceMetadata buildVirtual(Row row, VersionNumber cassandraVersion) {
+            String name = row.getString(KS_NAME);
+            return new KeyspaceMetadata(name, false, Collections.<String, String>emptyMap(), true);
     }
 
     /**
@@ -86,6 +100,15 @@ public class KeyspaceMetadata {
      */
     public boolean isDurableWrites() {
         return durableWrites;
+    }
+
+    /**
+     * Returns whether or not this keyspace is a virtual keyspace
+     * @return {@code true} if virtual keyspace
+     * default), {@code false} otherwise.
+     */
+    public boolean isVirtual() {
+        return isVirtual;
     }
 
     /**
@@ -351,6 +374,9 @@ public class KeyspaceMetadata {
      * @see #exportAsString
      */
     public String asCQLQuery() {
+        if(isVirtual){
+            return "";
+        }
         StringBuilder sb = new StringBuilder();
 
         sb.append("CREATE KEYSPACE ").append(Metadata.quoteIfNecessary(name)).append(" WITH ");
@@ -367,6 +393,9 @@ public class KeyspaceMetadata {
 
     @Override
     public String toString() {
+        if(isVirtual){
+            return name;
+        }
         return asCQLQuery();
     }
 
--- a/driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
@@ -35,7 +35,7 @@ public class KeyspaceMetadata {
 
     private final String name;
     private final boolean durableWrites;
-    private final boolean isVirtual;
+    private final boolean virtual;
 
     private final ReplicationStrategy strategy;
     private final Map<String, String> replication;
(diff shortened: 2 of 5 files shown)


driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
```diff
@@ -735,4 +761,78 @@ private boolean references(DataType dataType, String typeName) {
...
+                    viewsFuture = null,
+                    virtualKeyspacesFuture = null,
+                    virtualTableFuture = null,
+                    virtualColumnsFuture = null;
+
+            ProtocolVersion protocolVersion = cluster.getConfiguration().getProtocolOptions().getProtocolVersion();
+
+            if (isSchemaOrKeyspace)
+                ksFuture = queryAsync(SELECT_KEYSPACES + whereClause(targetType, targetKeyspace, targetName, targetSignature), connection, protocolVersion);
+            virtualKeyspacesFuture = queryAsync(SELECT_VIRTUAL_KEYSPACES + whereClause(targetType, targetKeyspace, targetName, targetSignature), connection, protocolVersion);
```
olim7t: This should be executed conditionnally, we don't need to run this query if the target type is e.g. `FUNCTION`.

I don't think we can get targeted schema notifications -- `targetType == KEYSPACE or TABLE` -- for virtual elements (but we need to double-check that, I haven't looked at the server code). If that is the case, then all virtual metadata (including tables and columns) only needs to be queried on a full schema refresh, we would include them in the `isSchemaOrKeyspace` block above.

driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
```diff
@@ -51,6 +52,15 @@
         this.durableWrites = durableWrites;
         this.replication = replication;
         this.strategy = ReplicationStrategy.create(replication);
+        this.isVirtual=false;
+    }
+    @VisibleForTesting
+    KeyspaceMetadata(String name, boolean durableWrites, Map<String, String> replication, boolean isVirtual) {
```
tolbertam: Nit: rename `isVirtual` to `virtual`

driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
```diff
@@ -51,6 +52,15 @@
...
+        this.name = name;
+        this.durableWrites = durableWrites;
+        this.replication = replication;
+        this.strategy = ReplicationStrategy.create(replication);
+        this.isVirtual=isVirtual;
```
tolbertam: Nit: space between equal sign

driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
```diff
@@ -51,6 +52,15 @@
         this.durableWrites = durableWrites;
         this.replication = replication;
         this.strategy = ReplicationStrategy.create(replication);
+        this.isVirtual=false;
```
tolbertam: Would be preferable for this constructor should reuse the new one (i.e. `this(name, durableWrites, replication, false)`).

Should also mark this constructor as deprecated as it looks like existing uses of it were updated to use the new one.

driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
```diff
@@ -61,13 +71,17 @@ static KeyspaceMetadata build(Row row, VersionNumber cassandraVersion) {
...
-            return new KeyspaceMetadata(name, durableWrites, replicationOptions);
+            return new KeyspaceMetadata(name, durableWrites, replicationOptions, false);
         } else {
             String name = row.getString(KS_NAME);
             boolean durableWrites = row.getBool(DURABLE_WRITES);
-            return new KeyspaceMetadata(name, durableWrites, row.getMap(REPLICATION, String.class, String.class));
+            return new KeyspaceMetadata(name, durableWrites, row.getMap(REPLICATION, String.class, String.class), false);
         }
     }
+    static KeyspaceMetadata buildVirtual(Row row, VersionNumber cassandraVersion) {
```
tolbertam: Nit: Add space between line 80 and 81

driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
```diff
@@ -146,6 +148,17 @@ void refresh(Cluster cluster,
             }
             keyspaces.put(keyspace.getName(), keyspace);
         }
+        if(rows.virtualKeyspaces!=null) {
```
tolbertam: nit: space between `!=`

driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
```diff
@@ -735,4 +761,78 @@ private boolean references(DataType dataType, String typeName) {
...
+
+        private static final String TABLE_NAME = "table_name";
+
+        @Override
+        SystemRows fetchSystemRows(Cluster cluster, SchemaElement targetType, String targetKeyspace, String targetName, List<String> targetSignature, Connection connection, VersionNumber cassandraVersion)
+                throws ConnectionException, BusyConnectionException, ExecutionException, InterruptedException {
+
+            boolean isSchemaOrKeyspace = (targetType == null || targetType == KEYSPACE);
+
+            ResultSetFuture ksFuture = null,
```
tolbertam, line 780: It would be nice to not duplicate a lot of this code from V3Parser, but refactoring it so it could be reused would probably be as much code, so i'm fine with this as is.

driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
```diff
@@ -491,9 +504,13 @@ private static ResultSet get(ResultSetFuture future) throws InterruptedException
         final Map<String, List<Row>> aggregates;
         final Map<String, List<Row>> views;
         final Map<String, Map<String, List<Row>>> indexes;
+        final ResultSet virtualKeyspaces;
+        final Map<String, List<Row>> virtualTables;
+        final Map<String, Map<String, Map<String, ColumnMetadata.Raw>>> virtualcolumns;
```
tolbertam: Nit: `virtualcolumns` -> `virtualColumns`

driver-core/src/main/java/com/datastax/driver/core/TableMetadata.java
```diff
@@ -73,6 +73,10 @@ private TableMetadata(KeyspaceMetadata keyspace,
     static TableMetadata build(KeyspaceMetadata ksm, Row row, Map<String, ColumnMetadata.Raw> rawCols, List<Row> indexRows, String nameColumn, VersionNumber cassandraVersion, Cluster cluster) {
 
         String name = row.getString(nameColumn);
+        if(ksm.isVirtual()){
+            return  new TableMetadata(ksm, name, new UUID(0L, 0L), Collections.<ColumnMetadata>emptyList(), Collections.<ColumnMetadata>emptyList(),
```
tolbertam: This is currently not passing in columns, so it looks like all virtual tables have no columns.

tolbertam (changes requested):
Looks good for the most part, had a few nits, and there is one issue in that columns are missing from `TableMetadata`.  This could also use a simple integration test that targets cassandra version 4.0+ that simply ensures that keyspaces/tables/columns are present for virtual tables.

driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
```diff
@@ -735,4 +761,78 @@ private boolean references(DataType dataType, String typeName) {
...
+    private static class V4SchemaParser extends V3SchemaParser {
+
+        private static final String SELECT_KEYSPACES = "SELECT * FROM system_schema.keyspaces";
+        private static final String SELECT_TABLES = "SELECT * FROM system_schema.tables";
+        private static final String SELECT_COLUMNS = "SELECT * FROM system_schema.columns";
+        private static final String SELECT_USERTYPES = "SELECT * FROM system_schema.types";
+        private static final String SELECT_FUNCTIONS = "SELECT * FROM system_schema.functions";
+        private static final String SELECT_AGGREGATES = "SELECT * FROM system_schema.aggregates";
+        private static final String SELECT_INDEXES = "SELECT * FROM system_schema.indexes";
+        private static final String SELECT_VIEWS = "SELECT * FROM system_schema.views";
```
olim7t: You could make these protected in `V3SchemaParser` instead of redefining them.

driver-core/src/main/java/com/datastax/driver/core/TableMetadata.java
```diff
@@ -416,6 +423,9 @@ public String exportAsString() {
```
olim7t, line 433: Maybe we should add a `TableMetadata.isVirtual()`, even though it will just be a trivial shortcut to `getKeyspace().isVirtual()`.

driver-core/src/main/java/com/datastax/driver/core/TableMetadata.java
```diff
@@ -73,6 +73,10 @@ private TableMetadata(KeyspaceMetadata keyspace,
     static TableMetadata build(KeyspaceMetadata ksm, Row row, Map<String, ColumnMetadata.Raw> rawCols, List<Row> indexRows, String nameColumn, VersionNumber cassandraVersion, Cluster cluster) {
 
         String name = row.getString(nameColumn);
+        if(ksm.isVirtual()){
+            return  new TableMetadata(ksm, name, new UUID(0L, 0L), Collections.<ColumnMetadata>emptyList(), Collections.<ColumnMetadata>emptyList(),
+                    Collections.<String, ColumnMetadata>emptyMap(), Collections.<String, IndexMetadata>emptyMap(), null, Collections.<ClusteringOrder>emptyList(), cassandraVersion);
+        }
```
olim7t: So a virtual table's metadata never has columns? Shouldn't we expose the contents of `system_virtual_schema.columns` here?

olim7t (changes requested):

driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
GregBestland: That was my thought as well.

driver-core/src/test/java/com/datastax/driver/core/VirtualTableMetadataTest.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.UUID;
+
+import static com.datastax.driver.core.Assertions.assertThat;
+
+@CassandraVersion("4.0.0")
+@CCMConfig()
+public class VirtualTableMetadataTest extends CCMTestsSupport {
+
+    @Test(groups = "short")
+    public void should_parse_virtual_metadata() {
```
tolbertam, line 30: Looks good, just a few suggestions:

1. Inconsistent spacing between operators in some cases (i.e. no space between `=`, `>=` in some places.
2. Would be good to add comments inline for each type validated to indicate why their values are that way, i.e.:

    ``` 
    // Keyspace name should be set, marked as virtual, and have a clients table.
    // All other values should be defaulted since they are not defined in schema tables.
    ```

driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
```diff
@@ -199,6 +199,15 @@ public TableOptionsMetadata getOptions() {
         return options;
     }
 
+    /**
+     * Returns whether or not this keyspace is a virtual keyspace
+     * @return {@code true} if virtual keyspace
```
tolbertam: Looks like `default )` was accidentally put here.

driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
```diff
@@ -47,10 +48,15 @@
 
     @VisibleForTesting
     KeyspaceMetadata(String name, boolean durableWrites, Map<String, String> replication) {
+        this(name,durableWrites,replication,false);
```
tolbertam: nit: spaces between commas.

driver-core/src/main/java/com/datastax/driver/core/KeyspaceMetadata.java
```diff
@@ -47,10 +48,15 @@
 
     @VisibleForTesting
     KeyspaceMetadata(String name, boolean durableWrites, Map<String, String> replication) {
```
tolbertam: Should mark this constructor as deprecated.

driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
```diff
@@ -491,9 +504,13 @@ private static ResultSet get(ResultSetFuture future) throws InterruptedException
...
         final Map<String, List<Row>> views;
         final Map<String, Map<String, List<Row>>> indexes;
+        final ResultSet virtualKeyspaces;
+        final Map<String, List<Row>> virtualTables;
+        final Map<String, Map<String, Map<String, ColumnMetadata.Raw>>> virtualColumns;
 
         public SystemRows(ResultSet keyspaces, Map<String, List<Row>> tables, Map<String, Map<String, Map<String, ColumnMetadata.Raw>>> columns, Map<String, List<Row>> udts, Map<String, List<Row>> functions,
-                          Map<String, List<Row>> aggregates, Map<String, List<Row>> views, Map<String, Map<String, List<Row>>> indexes) {
+                          Map<String, List<Row>> aggregates, Map<String, List<Row>> views, Map<String, Map<String, List<Row>>> indexes, ResultSet virtualKeyspaces, Map<String, List<Row>> virtualTables,
+                          Map<String, Map<String, Map<String, ColumnMetadata.Raw>>> virtualcolumns) {
```
tolbertam: nit: I see `virtualcolumns` was renamed `virtualColumns` for the class member, but the constructor parameter is still `virtualcolumns`.

driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
```diff
@@ -571,7 +591,10 @@ else if (targetType == AGGREGATE)
                     groupByKeyspace(get(aggregatesFuture)),
                     // No views nor separate indexes table in Cassandra 2:
                     Collections.<String, List<Row>>emptyMap(),
-                    Collections.<String, Map<String, List<Row>>>emptyMap());
+                    Collections.<String, Map<String, List<Row>>>emptyMap(),
+                   null,
```
tolbertam: nit: should be indented another space to match the other arguments.

tolbertam (commented):
Looking good :+1:, one small suggestion and some nits.  You mentioned out-of-band something about updating `describe` for virtual metadata, so will hold off on approving until that's sorted.

GregBestland:
Should be everything Andy let me know if I missed anything

tolbertam (approved):
Looks good! :+1:

driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
```diff
@@ -735,4 +761,70 @@ private boolean references(DataType dataType, String typeName) {
...
+                    virtualTableFuture = null,
+                    virtualColumnsFuture = null;
+
+            ProtocolVersion protocolVersion = cluster.getConfiguration().getProtocolOptions().getProtocolVersion();
+
+            if (isSchemaOrKeyspace)
+                ksFuture = queryAsync(SELECT_KEYSPACES + whereClause(targetType, targetKeyspace, targetName, targetSignature), connection, protocolVersion);
+                virtualKeyspacesFuture = queryAsync(SELECT_VIRTUAL_KEYSPACES + whereClause(targetType, targetKeyspace, targetName, targetSignature), connection, protocolVersion);
+                virtualColumnsFuture = queryAsync(SELECT_VIRTUAL_COLUMNS + whereClause(targetType, targetKeyspace, targetName, targetSignature), connection, protocolVersion);
+                virtualTableFuture = queryAsync(SELECT_VIRTUAL_TABLES + whereClause(targetType, targetKeyspace, targetName, targetSignature), connection, protocolVersion);
```
olim7t: You need curly braces around this block, currently only the first statement is executed conditionally.

driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
```diff
@@ -199,6 +199,14 @@ public TableOptionsMetadata getOptions() {
         return options;
     }
 
+    /**
+     * Returns whether or not this keyspace is a virtual keyspace
+     * @return {@code true} if virtual keyspace, {@code false} otherwise.
+     */
+    public boolean isVirtual(){
```
olim7t: Nit: the javadoc should say table,  not keyspace.

driver-core/src/test/java/com/datastax/driver/core/VirtualTableMetadataTest.java
```diff
@@ -0,0 +1,69 @@
...
+        KeyspaceMetadata km=session().getCluster().getMetadata().getKeyspace("system_views");
+        // Keyspace name should be set, marked as virtual, and have a clients table.
+        // All other values should be defaulted since they are not defined in the virtual schema tables.
+        assertThat(km.getTables().size() >= 2);
+        assertThat(km.isVirtual()).isTrue();
+        assertThat(km.getTable("clients")).isNotNull();
```
olim7t: Nit: this is duplicated at line 48, I would get rid of this line.

olim7t (changes requested):

olim7t:
LGTM, last round of minor/trivial comments and this is good to go.

driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
```diff
@@ -245,6 +253,9 @@ public String asCQLQuery() {
 
     protected StringBuilder appendOptions(StringBuilder sb, boolean formatted) {
         // Options
+        if(options==null){
```
adutra: In several places the formatting rules are not respected, please run a global format on this PR before merging, applying the rules for driver 3.x [here](https://github.com/datastax/java-driver/blob/3.x/CONTRIBUTING.md#editor-configuration).

driver-core/src/main/java/com/datastax/driver/core/TableMetadata.java
```diff
@@ -417,15 +434,20 @@ public String exportAsString() {
     @Override
     protected String asCQLQuery(boolean formatted) {
         StringBuilder sb = new StringBuilder();
-        sb.append("CREATE TABLE ").append(Metadata.quoteIfNecessary(keyspace.getName())).append('.').append(Metadata.quoteIfNecessary(name)).append(" (");
+        if(isVirtual()){
+            sb.append("/* VIRTUAL ");
```
adutra: I know this is extremely unlikely but we could theoretically produce invalid CQL here if the table contains weird column names as in ` create table t1 ("pk */" int primary key)`; this could also happen if some table options contain such forbidden sequences, for example in the `comments` field.
One simple idea to fix this would be to use single-line comments (`--` or `//`) and force `formatted` to false.
EDIT: nevermind, with virtual tables this should never happen. OK to ignore this comment.

adutra (dismissed):

driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
GregBestland: Good catch

GregBestland:
Everything should be addressed

olim7t (approved):

PR #1051, opened 2018-07-09
JAVA-1889: Upgrade dependencies to the latest minor versions

olim7t:
Pretty straightforward, just a name clash with a new AssertJ assertion, and an added null check in a Netty method where we passed `null` in our tests (passing a plain mock instead works).


tolbertam (approved):
:+1:

PR #1052, opened 2018-07-09
JAVA-1879: Duplicate basic.request options as Statement attributes

tolbertam:
For [JAVA-1879](https://datastax-oss.atlassian.net/browse/JAVA-1879).  This should not be merged until #1033, at which point some rebasing / test updates will need to be done.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/cql/Statement.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/cql/Statement.java
@@ -15,8 +15,10 @@
  */
 package com.datastax.oss.driver.api.core.cql;
 
+import com.datastax.oss.driver.api.core.ConsistencyLevel;
 import com.datastax.oss.driver.api.core.CqlIdentifier;
 import com.datastax.oss.driver.api.core.CqlSession;
+import com.datastax.oss.driver.api.core.config.DefaultDriverOption;
 import com.datastax.oss.driver.api.core.config.DriverConfigProfile;
 import com.datastax.oss.driver.api.core.context.DriverContext;
 import com.datastax.oss.driver.api.core.metadata.token.Token;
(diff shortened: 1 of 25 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/cql/Statement.java
```diff
@@ -225,6 +239,65 @@ default T setRoutingKey(@NonNull ByteBuffer... newRoutingKeyComponents) {
   @NonNull
   T setPagingState(@Nullable ByteBuffer newPagingState);
 
+  /**
+   * Returns the page size to use for the statement.
+   *
+   * @return the set page size, otherwise 0 or a negative value to use the default valued in the
+   *     configuration.
```
olim7t: Nit: should it say "the default value defined in the configuration"?

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareHandlerBase.java
```diff
@@ -284,6 +287,8 @@ private DefaultPreparedStatement cache(DefaultPreparedStatement preparedStatemen
       LOG.warn(
           "Re-preparing already prepared query. "
               + "This is generally an anti-pattern and will likely affect performance. "
+              + "The cached version of the PreparedStatement will be returned which may have a "
+              + "different base Statement configuration than the base Statement that was prepared. "
```
olim7t: That's a bit cryptic for the uninitiated, it's hard to explain in such a short message. How about:

> The cached version of the PreparedStatement will be returned, which may use different bound statement execution parameters (CL, timeout, etc.) from the current session.prepare call.

olim7t (approved):
👍 LGTM, minor javadoc/message feedback.

PR #1053, opened 2018-07-10
JAVA-1437: Enable SSL hostname validation by default

tolbertam:
For [JAVA-1437](https://datastax-oss.atlassian.net/browse/JAVA-1437).

Motivation:

When SSL is enabled, the driver should also enable validation of
certificate common names against server hostname by default. The
reasoning is that it seems logical to perform more validation than
less by default when the user enables SSL.

Modifications:

Change advanced.ssl-engine-factory.hostname-validation default to true.

Update manual and tests to reflect this change.

Result:

When DefaultSslEngineFactory is configured, hostname validation is
enabled by default.  The user may adjust the configuration to
explicitly disable hostname validation.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -4,6 +4,7 @@
 
 ### 4.0.0-beta1 (in progress)
 
+- [improvement] JAVA-1437: Enable SSL hostname validation by default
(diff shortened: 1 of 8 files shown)


olim7t (approved):

tolbertam:
Thanks for reviewing, rebased and updated changelog and merging.

PR #1054, opened 2018-07-11
JAVA-1900: Support virtual tables (aka system views)

GregBestland:
There are a couple of things here that I'm not sure about.
1. Is removing the describe for virtual keyspaces/table really what we want to do? Are there applications upstream which might utilize it in some way.
2.Is having a bunch of unpopulated fields in the table/keyspace acceptable, or do we want another type?


olim7t:
> Is removing the describe for virtual keyspaces/table really what we want to do? Are there applications upstream which might utilize it in some way.

I've given this a bit more thought. Returning an empty string is nice to automatically "ignore" the element if you're generating a script for the whole schema:
```java
StringBuilder sb = new StringBuilder();
for (KeyspaceMetadata keyspace : keyspaces) {
  sb.append(keyspace.describeWithChildren(true));
}
```

However it doesn't work that well for an individual keyspace, because the empty string is not a valid CQL query:
```java
session.execute(keyspace.describeWithChildren(true));
// SyntaxError: line 0:-1 no viable alternative at input '<EOF>'
```

So maybe we could instead have the describe methods throw `UnsupportedOperationException`, and add a `Describable.isDescribable()` method to check beforehand.

> Is having a bunch of unpopulated fields in the table/keyspace acceptable, or do we want another type?

I think it's acceptable. We should just make sure that non-collection fields that can be unpopulated for virtual elements return `Optional`. As far as I can tell, the only such field is `getId()` for tables.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/schema/TableMetadata.java
```diff
@@ -32,6 +32,10 @@
   @NonNull
   @Override
   default String describe(boolean pretty) {
+    // this is a virtual table, do not return a describe string
+    if (getPrimaryKey().size() == 0) {
+      return "";
+    }
```
olim7t: We should have an `isVirtual()` method on `TableMetadata` as well.

And, even though the PK size is probably a fairly reliable indicator, it would be cleaner to store the information explicitly, as a `virtual` field that gets passed to `DefaultTableMetadata`'s constructor. It will be easy to set since you already have a separate method to build virtual tables in `TableParser`.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/TableParser.java
```diff
@@ -226,6 +226,50 @@ TableMetadata parseTable(
         indexesBuilder.build());
   }
 
+  TableMetadata parseVirtualTable(
+      AdminRow tableRow, CqlIdentifier keyspaceId, Map<CqlIdentifier, UserDefinedType> userTypes) {
+
+    CqlIdentifier tableId =
+        CqlIdentifier.fromInternal(
+            tableRow.getString(rows.isCassandraV3 ? "table_name" : "columnfamily_name"));
```
olim7t: Nit: virtual tables are 4.0+ so the column will always be `table_name`.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/TableParser.java
```diff
@@ -226,6 +226,50 @@ TableMetadata parseTable(
...
+      ColumnMetadata column =
+          new DefaultColumnMetadata(
+              keyspaceId, tableId, raw.name, dataType, raw.kind == RawColumn.Kind.STATIC);
+      allColumnsBuilder.put(column.getName(), column);
+    }
+
+    return new DefaultTableMetadata(
+        keyspaceId,
+        tableId,
+        new UUID(0L, 0L),
```
olim7t: See my top-level comment on the PR, maybe it would be better to expose this as an `Optional`.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/TableParser.java
```diff
@@ -226,6 +226,50 @@ TableMetadata parseTable(
         indexesBuilder.build());
   }
 
+  TableMetadata parseVirtualTable(
```
olim7t, line 230: 👍 for exposing this as a separate method, that leads to a bit of duplication but it's easier to follow.

olim7t (changes requested):
Looks good, see a few minor comments inline.

olim7t:
PS - for my previous suggestion:

> So maybe we could instead have the describe methods throw `UnsupportedOperationException`, and add a `Describable.isDescribable()` method to check beforehand.

This makes things more explicit, but on the other hand it adds a bit more ceremony when dealing with script generation. So it's open for debate, any thoughts?

core/src/main/java/com/datastax/oss/driver/api/core/metadata/schema/KeyspaceMetadata.java
```diff
@@ -34,6 +34,9 @@
   /** Whether durable writes are set on this keyspace. */
   boolean isDurableWrites();
 
+  /** Whether durable writes are set on this keyspace. */
```
adutra: Wrong comment.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/SchemaParser.java
```diff
@@ -73,15 +74,19 @@ public SchemaParser(SchemaRows rows, InternalDriverContext context) {
...
+      KeyspaceMetadata keyspace = parseKeyspace(row, true);
       keyspacesBuilder.put(keyspace.getName(), keyspace);
     }
     SchemaRefresh refresh = new SchemaRefresh(keyspacesBuilder.build());
     LOG.debug("[{}] Schema parsing took {}", logPrefix, NanoTime.formatTimeSince(startTimeNs));
     return refresh;
   }
 
-  protected KeyspaceMetadata parseKeyspace(AdminRow keyspaceRow) {
+  protected KeyspaceMetadata parseKeyspace(AdminRow keyspaceRow, boolean isVirtual) {
```
adutra: Nit: the convention would be to name this boolean simply `virtual`.

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/DefaultTableMetadata.java
```diff
@@ -75,15 +75,23 @@ public CqlIdentifier getName() {
 
   @NonNull
   @Override
-  public UUID getId() {
-    return id;
+  public Optional<UUID> getId() {
+    if (id == null) {
```
adutra: This can be simplified to `Optional.ofNullable(id)`.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/DefaultViewMetadata.java
```diff
@@ -81,8 +81,8 @@ public CqlIdentifier getName() {
 
   @NonNull
   @Override
-  public UUID getId() {
-    return id;
+  public Optional<UUID> getId() {
+    return Optional.of(id);
```
adutra, line 85: If `id` can be null than this should rather be `Optional.ofNullable(id)`. (But I'm not sure if views can be virtual, can they?)

adutra (changes requested):
Just a small nit about proper usage of the Optional class, but otherwise 👍 . Fix it and feel free to merge.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/DefaultViewMetadata.java
GregBestland: I don't believe that views can be virtual at this time, just keyspaces and tables

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/DefaultTableMetadata.java
olim7t: Also, you'll want to mark the `id` field as `Nullable`, currently this generates a warning.

PR #1055, opened 2018-07-12
JAVA-1913: Expose additional counters on Node

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/metadata/Node.java
```diff
@@ -149,4 +155,21 @@
...
+   * yet.
+   *
+   * <p>This is recorded with {@link System#nanoTime()}, and should only be used to measure elapsed
+   * time.
+   *
+   * <p>Note that this feature is controlled by a configuration option ({@code
+   * advanced.metadata.nodes.last-response-time.enabled}) and <b>disabled by default</b> (i.e. will
+   * always return -1). You might want to enable it for information purposes, or for use in custom
+   * policy implementations. It incurs a bit of overhead (namely, a call to {@code
+   * System.nanoTime()} and a volatile write for every request).
```
tolbertam: It looks like in `CqlRequestHandlerBase` it was already calling `System.nanoTime()` for the purpose of metrics, but I think the documentation is fine since that detail is impl specific.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Node.java
```diff
@@ -108,6 +109,11 @@
   @NonNull
   NodeState getState();
 
+  /**
+   * The last time that this node transitioned to the UP state, or -1 if it's not up at the moment.
+   */
+  long getUpSince();
```
tolbertam: It isn't obvious that this is milliseconds instead of nanos and uses `System.currentTimeMillis()`.  Would it be worth renaming this `getUpSinceMillis()` or something to that extent?

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/NodeMetadataIT.java
```diff
@@ -58,7 +56,50 @@ public void should_expose_node_metadata() {
     assertThat(node.getDistance()).isSameAs(NodeDistance.LOCAL);
     assertThat(node.getHostId()).isNotNull();
     assertThat(node.getSchemaVersion()).isNotNull();
+    assertThat(node.getLastResponseTimeNanos()).isEqualTo(-1);
```
tolbertam: Would also be worth validating that `getUpSince` is not equal to -1.

tolbertam (approved):
Looks good, just had a few small suggestions.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Node.java
olim7t: Right, I wrote that before the implementation. `CqlRequestHandlerBase` should generally represent the majority of the traffic, so I'll just mention the volatile write.

On a side note, the `nanoTime` call is not needed at all if both this feature and metrics are disabled. We could try to handle that a bit better, I'll open a separate ticket.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Node.java
olim7t: Absolutely

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/NodeMetadataIT.java
olim7t: I also forced the node down and back up to check that the time is updated.

tolbertam (approved):
Looks great! :+1:

PR #1057, opened 2018-07-13
JAVA-1544: Check API compatibility with Revapi

olim7t:
To test:
```
mvn verify -DskipTests -DskipITs | tee revapi.log 2>&1
```
Right now it compares beta1 to alpha3, so there are still a lot of breaking changes and the build fails. Things will get more interesting once we release beta1, I think we want to start tracking API changes right then, even though they're still allowed between betas.

Note: the plugin requires Maven >= 3.2.5

TODO
- [x] test again when we have a beta1 release

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -289,6 +289,18 @@
           <artifactId>maven-bundle-plugin</artifactId>
           <version>3.5.0</version>
         </plugin>
+        <plugin>
+          <groupId>org.revapi</groupId>
+          <artifactId>revapi-maven-plugin</artifactId>
+          <version>0.10.4</version>
+          <dependencies>
+            <dependency>
+              <groupId>org.revapi</groupId>
+              <artifactId>revapi-java</artifactId>
+              <version>0.18.0</version>
+            </dependency>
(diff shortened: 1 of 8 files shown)


pom.xml
```diff
@@ -544,6 +556,23 @@ limitations under the License.]]>
...
+        <artifactId>revapi-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <goals>
+              <goal>check</goal>
+            </goals>
+            <configuration>
+              <analysisConfigurationFiles>
+                <!-- Present at the root of each module that doesn't skip this goal -->
+                <analysisConfigurationFile>revapi.json</analysisConfigurationFile>
```
olim7t, line 568: I like having a configuration file per module. Right now it has the same contents everywhere, but if we need to add exclusions in the future, it's better to have them at the module level.

adutra (approved):

olim7t:
Force-pushed to rebase on current 4.x.
Beta1 is released, so it's now the reference version. Unsurprisingly, API checks pass 😃

PR #1056, opened 2018-07-13
JAVA-1924: StatementWrapper setters should return the wrapping statement

adutra:

PR #1059, opened 2018-07-19
JAVA-1857: Add Statement.setHost

tolbertam:
For [JAVA-1857](https://datastax-oss.atlassian.net/browse/JAVA-1857).

Motivation:

With the introduction of system views in Apache C* 4.0
(CASSANDRA-7622) there is a need for the capability to
set a target node when executing queries.

This feature should be avoided in the general case and
only used in exceptional cases such as this.

Modifications:

Add Statement.setHost(Host).

Update RequestHandler to bypass the configured load balancing policy
and instead create a single host query plan targeting the configured
host if set.

Result:

Driver now offers capability of host targeting via Statement.setHost.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
@@ -21,6 +21,7 @@ import com.datastax.driver.core.policies.RetryPolicy;
 import com.datastax.driver.core.policies.RetryPolicy.RetryDecision.Type;
 import com.datastax.driver.core.policies.SpeculativeExecutionPolicy.SpeculativeExecutionPlan;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterators;
 import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
@@ -84,7 +85,14 @@ class RequestHandler {
 
         callback.register(this);
 
-        this.queryPlan = new QueryPlan(manager.loadBalancingPolicy().newQueryPlan(manager.poolsState.keyspace, statement));
+        // If host is explicitly set on statement, bypass load balancing policy.
+        if (statement.getHost() != null) {
+            this.queryPlan = new QueryPlan(Iterators.singletonIterator(statement.getHost()));
+        } else {
+            this.queryPlan = new QueryPlan(manager.loadBalancingPolicy()
+                    .newQueryPlan(manager.poolsState.keyspace, statement));
+        }
+
         this.speculativeExecutionPlan = manager.speculativeExecutionPolicy().newPlan(manager.poolsState.keyspace, statement);
         this.allowSpeculativeExecutions = statement != Statement.DEFAULT
                 && statement.isIdempotentWithDefault(manager.configuration().getQueryOptions());
--- a/driver-core/src/main/java/com/datastax/driver/core/Statement.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Statement.java
@@ -635,7 +635,7 @@ public abstract class Statement {
      * </ol>
      * <p/>
      * Configuring a specific host causes the configured {@link LoadBalancingPolicy} to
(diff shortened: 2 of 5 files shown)


driver-core/src/main/java/com/datastax/driver/core/StatementWrapper.java
```diff
@@ -210,4 +210,15 @@ public Statement setOutgoingPayload(Map<String, ByteBuffer> payload) {
...
         return wrapped.requestSizeInBytes(protocolVersion, codecRegistry);
     }
+
+    @Override
+    public Host getHost() {
+        return wrapped.getHost();
+    }
+
+    @Override
+    public Statement setHost(Host host) {
```
tolbertam, line 220: Following the strategy employed in #1056

driver-core/src/test/java/com/datastax/driver/core/HostTargetingTest.java
```diff
@@ -0,0 +1,126 @@
...
+        Host host4 = TestUtils.findHost(cluster, 4);
+        Statement statement = new SimpleStatement("select * system.local").setHost(host4);
+
+        try {
+            // when statement is executed
+            session.execute(statement);
+            fail("Query should have failed");
+        } catch (NoHostAvailableException e) {
+            // then the request should fail with a NHAE and no host was tried.
+            assertThat(e.getErrors()).isEmpty();
```
tolbertam, line 136: In this case, you get the `No host was tried` message.  I considered adding something specific (i.e. `Configured host was not available`), but I don't think we need to go out of the way for that.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -84,7 +85,14 @@ public RequestHandler(SessionManager manager, Callback callback, Statement state
 
         callback.register(this);
 
-        this.queryPlan = new QueryPlan(manager.loadBalancingPolicy().newQueryPlan(manager.poolsState.keyspace, statement));
+        // If host is explicitly set on statement, bypass load balancing policy.
+        if (statement.getHost() != null) {
+            this.queryPlan = new QueryPlan(Iterators.singletonIterator(statement.getHost()));
```
adutra, line 90: Should we also disable speculative executions? Or is it too much hassle for little gain?

driver-core/src/main/java/com/datastax/driver/core/Statement.java
```diff
@@ -610,4 +613,38 @@ protected static Boolean isBatchIdempotent(Collection<? extends Statement> state
...
+     * used in the following cases:
+     * <ol>
+     * <li>Querying node-local tables, such as tables in the {@code system} and
+     * {@code system_views} keyspaces.</li>
+     * <li>Applying a series of schema changes, where it may be advantageous to execute schema
+     * changes in sequence on the same node.</li>
+     * </ol>
+     * <p/>
+     * Configuring a specific host causes the configured {@link LoadBalancingPolicy} to
+     * be completely bypassed.  However, if the load balancing policy dictates that the host is at
```
adutra: Nit: extra space before `However`.

adutra (approved):

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
tolbertam: I think that would be a small optimization to bypass scheduling an execution, but I don't think its worth adding additional code.  It looks like an execution could get scheduled, but there is logic in `RequestHandler` to cancel any pending speculative executions if we finish a query or encounter the end of the plan (`reportNoMoreHosts()` is called).  It's possible the execution would get fired, but then encounter nothing in the plan.

olim7t: 👍 
System view queries and DDL changes &mdash; the two target use cases for this feature &mdash; are probably not very frequent queries anyway, so that short extra execution should not matter.
And if it does, it can be worked around with a custom speculative execution policy.

olim7t (approved):

tolbertam:
Thank you both for reviewing :+1:

PR #1060, opened 2018-07-20
JAVA-1762: Improve shading and OSGi manifest generation

adutra:
Summary of changes:

* In module core there is no more shading.
  * The shading of jctools is IMO unnecessary, since it is an OSGi-ready bundle.
  * The generation of the shaded manifest is now done inside core-shaded.
* In module core-shaded, the procedure to generate the shaded classes and the shaded bundle is as follows:
  1. Shade the classes, generate a first shaded jar;
  2. Unpack the shaded jar;
  3. Generate the manifest based on the unpacked shaded classes;
  4. Repack the shaded jar.

Producing the shaded manifest based on shaded classes is IMO more accurate since BND analyzes the actual content of the shaded jar to generate the manifest (whereas previously we made it generate a shaded manifest based on non-shaded classes in module core).

And a note about Guava: I'm still doubtful about the necessity of shading Guava differently (i.e. outside the driver). With the changes in this PR the shading of dependencies is clearly isolated in dedicated modules. I don't see why we couldn't do the same for Guava as we do for Netty. But as agreed, I didn't touch Guava within this PR.

### DIFF
--- a/core-shaded/pom.xml
+++ b/core-shaded/pom.xml
@@ -16,7 +16,9 @@
     limitations under the License.
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
   <parent>
@@ -26,57 +28,112 @@
   </parent>
 
   <artifactId>java-driver-core-shaded</artifactId>
-  <packaging>bundle</packaging>
 
-  <name>DataStax Java driver for Apache Cassandra(R) - core with netty shaded</name>
+  <name>DataStax Java driver for Apache Cassandra(R) - core with shaded deps</name>
 
   <dependencies>
+    <!--
+    Declare a dependency to the core driver itself so that all its classes get included;
+    this dependency will be removed from the final pom by the shade plugin.
+    -->
     <dependency>
       <groupId>com.datastax.oss</groupId>
       <artifactId>java-driver-core</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <!--
+    Repeat all dependencies of the core driver *except* the ones that are going to be shaded,
+    so that they get included in the final pom (we don't use the "promoteTransitiveDependencies"
+    option of the shade plugin because it promotes all dependencies, even nested ones, to top level).
+    -->
+    <dependency>
+      <groupId>com.datastax.oss</groupId>
+      <artifactId>native-protocol</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.datastax.oss</groupId>
+      <artifactId>java-driver-shaded-guava</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.typesafe</groupId>
+      <artifactId>config</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.github.jnr</groupId>
+      <artifactId>jnr-ffi</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.github.jnr</groupId>
+      <artifactId>jnr-posix</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.xerial.snappy</groupId>
+      <artifactId>snappy-java</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.lz4</groupId>
+      <artifactId>lz4-java</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-api</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.dropwizard.metrics</groupId>
+      <artifactId>metrics-core</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.hdrhistogram</groupId>
+      <artifactId>HdrHistogram</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.github.stephenc.jcip</groupId>
+      <artifactId>jcip-annotations</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>com.github.spotbugs</groupId>
+      <artifactId>spotbugs-annotations</artifactId>
+      <optional>true</optional>
+    </dependency>
   </dependencies>
+
+  <!--
+  Generation of the shaded driver-core bundle during package phase:
+  1) shade plugin shades the driver and creates a shaded jar + source jar
+  2) dependency plugin unpacks the shaded jar to target/classes (and removes unwanted content)
+  3) bundle plugin analyzes shaded classes and generates the bundle manifest
+  4) assembly plugin re-creates the shaded jar by packing target/classes + manifest + shaded pom
+  -->
 
   <build>
     <plugins>
-      <!-- extract shaded manifest from core jar -->
-      <plugin>
-        <artifactId>maven-dependency-plugin</artifactId>
-        <executions>
-          <execution>
-            <id>unpack-manifest</id>
-            <phase>prepare-package</phase>
-            <goals>
-              <goal>unpack</goal>
-            </goals>
-            <configuration>
-              <artifactItems>
-                <artifactItem>
-                  <groupId>com.datastax.oss</groupId>
-                  <artifactId>java-driver-core</artifactId>
-                  <version>${project.version}</version>
-                  <includes>META-INF-shaded/MANIFEST.MF</includes>
-                  <outputDirectory>${project.build.directory}</outputDirectory>
-                </artifactItem>
-              </artifactItems>
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
       <plugin>
         <artifactId>maven-shade-plugin</artifactId>
         <executions>
           <execution>
+            <id>shade-core-dependencies</id>
             <phase>package</phase>
             <goals>
               <goal>shade</goal>
             </goals>
             <configuration>
+              <createSourcesJar>true</createSourcesJar>
+              <shadeSourcesContent>true</shadeSourcesContent>
               <artifactSet>
                 <includes>
+                  <!--
+                  Include:
+                  - The core driver itself; it is not relocated but needs to be included.
+                  - All the dependencies we want to shade & relocate: currently
+                    - all the Netty artifacts;
+                    - JCTools.
+                  -->
                   <include>com.datastax.oss:java-driver-core</include>
                   <include>io.netty:*</include>
+                  <include>org.jctools:*</include>
                 </includes>
               </artifactSet>
               <relocations>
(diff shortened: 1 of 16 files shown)


core-shaded/pom.xml
```diff
@@ -28,97 +28,197 @@
   </parent>
 
   <artifactId>java-driver-core-shaded</artifactId>
-  <packaging>bundle</packaging>
 
-  <name>DataStax Java driver for Apache Cassandra(R) - core with netty shaded</name>
+  <name>DataStax Java driver for Apache Cassandra(R) - core with shaded deps</name>
```
adutra, line 32: I'm choosing a more generic name in case we add more shaded dependencies later.

core-shaded/pom.xml
```diff
@@ -28,97 +28,197 @@
...
 
   <dependencies>
     <dependency>
       <groupId>com.datastax.oss</groupId>
       <artifactId>java-driver-core</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <!--
+    Repeat optional dependencies of the driver here because the shade plugin
```
adutra: This is a bit unfortunate but the shade plugin "forgets" to promote optional dependencies.

core-shaded/pom.xml
```diff
@@ -28,97 +28,197 @@
...
+                <DynamicImport-Package>*</DynamicImport-Package>
+                <!--
+                Don't include because they aren't OSGi bundles, and the driver bundle can
+                live without them:
+                 - jnr
+                 - net.jcip.annotations
+                 - edu.umd.cs.findbugs.annotations
+                Don't include because they are imported by shaded Netty classes
+                but not used by the driver bundle:
+                 - com.google.protobuf
```
adutra: Because Netty has been shaded, BND now detects a few package imports of Netty dependencies that would otherwise appear only inside Netty bundles, not here. These are mostly compression and logging libraries. None of them is used by the driver so I just removed them.

core-shaded/pom.xml
```diff
@@ -28,97 +28,197 @@
...
+                  !org.apache.log4j.*,
+                  !org.apache.logging.log4j.*,
+                  !org.eclipse.jetty.*,
+                  !org.jboss.marshalling.*,
+                  !sun.misc.*,
+                  !sun.security.*,
+                  *
+                </Import-Package>
+                <Export-Package>
+                  com.datastax.oss.driver.*.core.*,
```
adutra: Note that we need to export not only API classes, but also internal driver classes and shaded classes (one reason being that the DSE driver needs them). I think it's a bit unfortunate to expose so many packages, but we don't have the choice.

core/pom.xml
```diff
@@ -27,6 +27,7 @@
   </parent>
 
   <artifactId>java-driver-core</artifactId>
+  <packaging>bundle</packaging>
```
adutra, line 30: We can now use packaging bundle here, since there is no more shading involved.

core/pom.xml
```diff
@@ -170,117 +166,38 @@
...
                 </Import-Package>
-                <!-- Explicitly declare shaded jctools packages since this isn't covered by shade plugin -->
                 <Export-Package>
-                  com.datastax.oss.driver.*.core.*,
-                  com.datastax.oss.driver.shaded.jctools.util;version="${project.version}";uses:="sun.misc",
-                  com.datastax.oss.driver.shaded.jctools.queues;version="${project.version}";uses:="com.datastax.oss.driver.shaded.jctools.queues.spec",
-                  com.datastax.oss.driver.shaded.jctools.queues.spec;version="${project.version}",
-                  com.datastax.oss.driver.shaded.jctools.queues.atomic;version="${project.version}";uses:="com.datastax.oss.driver.shaded.jctools.queues,com.datastax.oss.driver.shaded.jctools.queues.spec",
-                  com.datastax.oss.driver.shaded.jctools.maps;version="${project.version}"
+                  com.datastax.oss.driver.*.core.*
```
adutra, line 193: Same here, we need to export internal classes along with API classes.

integration-tests/src/test/java/com/datastax/oss/driver/osgi/BundleOptions.java
```diff
@@ -38,6 +38,7 @@ public static CompositeOption baseOptions() {
             mavenBundle("org.slf4j", "slf4j-api", getVersion("slf4j.version")),
             mavenBundle("org.hdrhistogram", "HdrHistogram", getVersion("hdrhistogram.version")),
             mavenBundle("com.typesafe", "config", getVersion("config.version")),
+            mavenBundle("org.jctools", "jctools-core", getVersion("jctools.version")),
```
adutra: JCTools being an OSGi bundle, we need to add the bundle to the provision mix.

test-infra/pom.xml
```diff
@@ -66,9 +66,15 @@
         <configuration>
           <instructions>
             <Bundle-SymbolicName>com.datastax.oss.driver.testinfra</Bundle-SymbolicName>
+            <!-- allow SessionUtils to instantiate sessions by reflection -->
+            <DynamicImport-Package>*</DynamicImport-Package>
```
adutra: That is required if we want `SessionUtils` to instantiate DSE sessions.

test-infra/pom.xml
```diff
@@ -66,9 +66,15 @@
...
             <!-- TODO: Only needed to use TestConfigLoader in integration-tests, remove this when we have
```
adutra: I confess I didn't fully understand this TODO so I left it intact.

integration-tests/src/test/java/com/datastax/oss/driver/osgi/OsgiBaseIT.java
```diff
@@ -78,7 +78,7 @@ public DriverConfigLoader configLoader() {
   @SuppressWarnings("unchecked")
   public void should_connect_and_query() {
     SessionBuilder<CqlSessionBuilder, CqlSession> builder =
-        SessionUtils.baseBuilder()
+        CqlSession.builder()
```
adutra: I prefer not to use our test infra inside this test to stay close to what users would actually do in their OSGi applications. We never know if by using the test infra we could introduce a biais (e.g. classes being loaded by the wrong classloader, etc.).

tolbertam:
> The shading of jctools is IMO unnecessary, since it is an OSGi-ready bundle.

Our decision to shade jctools was related to the fact that Netty also shades it.  We were using the shaded version in our code, which didn't work with OSGi since the netty jar doesn't export its packages.  We discussed whether or not we should do the same and decided that we should shade it as well because the jar is fairly small and it is a pretty popular package, so we didn't want to risk future dependency conflicts.

tolbertam:
Although, maybe what we should do is consider shading jctools as part of `core-shaded` only, instead of always shading it, wdyt?

integration-tests/src/test/java/com/datastax/oss/driver/osgi/OsgiBaseIT.java
tolbertam: I think this is fine, we will probably want to have a separate test set up anyways if we are using different session types as different jars will likely be required.

test-infra/pom.xml
tolbertam: It's not needed any more now that #1032 is merged and `TestConfigLoader` has been deleted, so this comment should be removed.

tolbertam (approved):
Fantastic work!  I was not completely satisfied with my previous solution of generating the core-shaded manifest in the core build, this is much better :+1: 

I'm :+1: on this, except I think we should include jctools in the core-shaded jar since it's a relatively popular library and is small enough that it doesn't add too much to the jar size.

integration-tests/src/test/java/com/datastax/oss/driver/osgi/OsgiBaseIT.java
adutra: Exactly. These tests have been failing consistently when ran against the DSE driver because, as you guessed, they rely on specific jars placed in specific folders on the filesystem. We will have to mirror these tests in the DSE driver test suite anyway.

adutra:
> Our decision to shade jctools was related to the fact that Netty also shades it. We were using the shaded version in our code, which didn't work with OSGi since the netty jar doesn't export its packages. 

First, I think it would be wiser for us to use the actual JCTools classes, not the ones shaded by Netty. Shaded classes are not really meant to be used outside the package that shaded them imo.

> We discussed whether or not we should do the same and decided that we should shade it as well because the jar is fairly small and it is a pretty popular package, so we didn't want to risk future dependency conflicts. [...] Although, maybe what we should do is consider shading jctools as part of core-shaded only, instead of always shading it, wdyt?

My only recommendation here would be to indeed never shade anything in the original core module, but do all the shading in the core-shaded module for consistency.

A part from that, I don't have strong feelings about whether or not JCTools is a good candidate for shading. On one side, as you pointed out, it is a very popular library; on the flip side though, the library evolves in a rather slow pace and its API is very stable, nothing like Guava or Netty. Wdyt?

tolbertam:
> A part from that, I don't have strong feelings about whether or not JCTools is a good candidate for shading. On one side, as you pointed out, it is a very popular library; on the flip side though, the library evolves in a rather slow pace and its API is very stable, nothing like Guava or Netty. Wdyt?

I think you've convinced me.  We should treat it as a "Cross that bridge when we come to it" kind of thing.  Until it becomes a problem, don't bother shading it.  We only use the API in one place (`MpscLinkedAtomicQueue` in `DefaultWriteCoalescer`), and I doubt we'll see breaking changes in the API.  Yeah I think I agree..lets avoid shading it for now :+1:.

tolbertam:
I've already approved this, but wanted to give an explicit :+1: again since I think we've resolved the only remaining thing I was concerned about (whether or not to shade jctools).

integration-tests/src/test/java/com/datastax/oss/driver/osgi/OsgiCustomLoadBalancingPolicyIT.java
```diff
@@ -27,9 +33,10 @@
  * DynamicImport-Package: *</code>.
  */
 public class OsgiCustomLoadBalancingPolicyIT extends OsgiBaseIT {
-  @Override
-  public Option[] additionalOptions() {
-    return new Option[0];
+
+  @Configuration
+  public Option[] config() {
```
adutra, line 38: I know that these things are a matter of taste, but I personally like it better when each class declares its own config, it makes it easier for me to understand what bundles are going to be provisioned. Hope everybody will agree here.

core-shaded/pom.xml
```diff
@@ -28,97 +28,201 @@
...
+                  com.datastax.oss.driver.shaded.*
+                </Export-Package>
+              </instructions>
+              <rebuildBundle>true</rebuildBundle>
```
adutra: I don't really understand what magic this option does but for the DSE driver it creates a much cleaner execution, so just to be on the safe side I'm also turning this on here for the OSS driver (although even without it the plugin executes without warnings here).

adutra:
I think we need to revisit our strategy for Guava. 

If I recall well, the main motivation to isolate the shaded Guava classes in a separate artifact was because we needed them in both drivers, and so isolating them allowed us to only shade once.

With the changes in the current PR we now are able to do the shading in the OSS driver and reuse the shaded classes in the DSE driver, so that motivation is not a valid one anymore.

On the other hand, having to maintain a separate artifact for shaded Guava creates some management overhead.

IMO it would be now much simpler to just consider Guava as another dependency to shade and treat it the same way we treat Netty, JCTools, etc.

core/pom.xml
```diff
@@ -170,117 +166,38 @@
...
                 <!-- Allow importing code from other packages (so reflection-based loading of policies works) -->
                 <DynamicImport-Package>*</DynamicImport-Package>
-                <!-- Don't include JNR packages since some JNR modules are not OSGi bundles,
-                     jcip because its not an OSGi bundle, jctools because its shaded.
-                     Import sun.misc as jctools does this. -->
+                <!--
+                Don't include because they aren't OSGi bundles, and the driver can live without them:
+                 - jnr
+                 - net.jcip.annotations
+                 - edu.umd.cs.findbugs.annotations
```
olim7t: Nit: listing the dependencies in the comment is redundant, this is literally a copy-paste of the 3 lines of code below.

core-shaded/pom.xml
```diff
@@ -28,97 +28,201 @@
...
+                 - org.bouncycastle
+                 - org.conscrypt
+                 - org.apache.commons.logging
+                 - org.apache.log4j
+                 - org.apache.logging.log4j
+                 - org.eclipse.jetty
+                 - org.jboss.marshalling
+                 - sun.misc
+                 - sun.security
+                -->
```
olim7t: Nit: same remark as the other POM, the comment paraphrases the code. You can simply have a "Don't include because..." comment at the beginning of each block in the code.

core-shaded/pom.xml
```diff
@@ -28,97 +28,201 @@
...
+                <relocation>
+                  <pattern>io.netty</pattern>
+                  <shadedPattern>com.datastax.oss.driver.shaded.netty</shadedPattern>
+                </relocation>
+              </relocations>
+              <!--
+              Promote all other transitive dependencies of java-driver-core other than Netty
+              to direct dependencies in the shaded pom.
+              -->
+              <promoteTransitiveDependencies>true</promoteTransitiveDependencies>
```
olim7t: Unfortunately this recurses beyond the first level of transitivity. If you look at `dependency-reduced-pom.xml`, it explicitly lists `org.ow2.asm.*` (pulled via `jnr-ffi`), `jnr-constants` (via `jnr-posix`), etc.

This is annoying if the end user wishes to exclude those dependencies manually. We need to find a way to avoid that.

olim7t (changes requested):
👍 to the new strategy for the shaded manifest. There is a small issue with transitive dependencies that we need to address (see inline).

I'd like to keep shading JCTools in order to keep the driver's dependencies minimal.

I'm addressing your questions about Guava in a separate comment.

olim7t:
> With the changes in the current PR we now are able to do the shading in the OSS driver and reuse the shaded classes

This works across projects, but Guava is also shared between core and query-builder. I think this would be hard to handle in the IDE (core references non-shaded classes, query-builder depends on shaded core so it references shaded classes, but those are not in the IDE's classpath).

> having to maintain a separate artifact for shaded Guava creates some management overhead.

Yes, but it is very minimal. There's no reason to ever bump the version except if it fixes a blocking issue, so we should be extremely conservative.

core-shaded/pom.xml
```diff
@@ -28,97 +28,243 @@
...
+                <!--
+                Export:
+                1) The driver's packages (API and internal);
+                2) All shaded packages, except Guava which resides in a separate bundle.
+                -->
+                <Export-Package>
+                  com.datastax.oss.driver.api.core.*,
+                  com.datastax.oss.driver.internal.core.*,
+                  com.datastax.oss.driver.shaded.netty.*,
+                  com.datastax.oss.driver.shaded.jctools.*
```
adutra: With today's changes I realized that the previous pattern (`com.datastax.oss.driver.shaded.*`) was too broad; as a consequence, we were _not importing_ but were instead _exporting_ shaded Guava classes. It's a miracle that it worked this way, but now it's fixed (the proper strategy is the opposite: import shaded Guava classes, but not export them since they live in a separate bundle).

core-shaded/pom.xml
```diff
@@ -28,97 +28,243 @@
...
+    <!--
+    Declare a dependency to the core driver itself so that all its classes get included;
+    this dependency will be removed from the final pom by the shade plugin.
+    -->
     <dependency>
       <groupId>com.datastax.oss</groupId>
       <artifactId>java-driver-core</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <!--
```
adutra, line 44: The only way I can think of to mimic the `promoteTransitiveDependencies` feature is to re-declare all relevant dependencies here. A bit verbose, but at least it works as expected.

olim7t (approved):

olim7t (changes requested):
Just realized that JCTools is not shaded in the default artifact...

olim7t:
At this point I'm starting to question the relevance of depending on JCTools. It's only used in the flusher and it was more of an opportunistic thing because it was already in the classpath. Let me run a few tests.

olim7t (approved):
We've decided to drop the dependency to JCTools. It's only used in `DefaultWriteCoalescer`. Our benchmarks show no significant improvement over `ConcurrentLinkedQueue` (in the overall context of request execution through the driver).

This will simplify the situation for this ticket (the default core artifact doesn't shade anything).

The coalescer component is pluggable, so there's still a way to use a different queue implementation.

adutra:
Did one final squash + rebase. I suggest that we remove JCTools within this PR to avoid having to do it twice (once for 4.x, once for this branch). I pushed a separate commit for that (02b970056); we must keep it separate when rebasing.

olim7t:
> I suggest that we remove JCTools within this PR

👍

PR #1061, opened 2018-07-24
JAVA-1921 add missing PrimaryKey, and Clustering column info to virtu…

GregBestland:
…al TableMetadata.


driver-core/src/test/java/com/datastax/driver/core/VirtualTableMetadataTest.java
```diff
@@ -50,14 +50,15 @@ public void should_parse_virtual_metadata() {
         assertThat(tm.getColumns().size()).isEqualTo(12);
         assertThat(tm.getIndexes().size()).isEqualTo(0);
         assertThat(tm.getViews().size()).isEqualTo(0);
-        assertThat(tm.getClusteringColumns().size()).isEqualTo(0);
-        assertThat(tm.getClusteringOrder().size()).isEqualTo(0);
+        assertThat(tm.getClusteringColumns().size()).isEqualTo(1);
```
tolbertam, line 57: Would also be nice to add validation for partition keys and check that the column names for clustering and partition keys are correct and that the ordering is right.

tolbertam (approved):
Looks good.  I only had one small suggestion in the test code but i ran this locally and everything looks right.

PR #1063, opened 2018-07-27
JAVA-1925: Rename context getters

olim7t:


pom.xml
```diff
@@ -567,6 +569,8 @@ limitations under the License.]]>
                 <!-- Present at the root of each module that doesn't skip this goal -->
                 <analysisConfigurationFile>revapi.json</analysisConfigurationFile>
               </analysisConfigurationFiles>
+              <!-- Prevents query-builder from re-checking core. -->
+              <checkDependencies>false</checkDependencies>
```
olim7t: I have mixed feelings about this, because it also excludes all external dependencies (Netty, etc). On the one hand, checking external API changes should not be our concern, on the other it would be good to know when upgrading a dependency breaks things.

If we remove this, core is checked again by the query-builder build. So we need to either:
- duplicate all the ignores of core's `revapi.json` in query-builder's
- use a single `revapi.json` at the parent level, but that brings the question on how to access it from the child modules: I don't like relative paths, it goes against the principles of Maven.

olim7t:
I've found a workaround for the problem I mentioned in my previous comment: use a blacklist for the packages covered by Revapi, instead of just whitelisting the current module.

Including as a separate commit, that should not be squashed because it's not directly related to this ticket.

adutra (approved):

adutra:
Rebased on top of current 4.x. As requested I'm keeping the two commits separated, but most of the changes in the second commit was already in 4.x.

PR #1064, opened 2018-07-27
JAVA-1786: Use Google code formatter

tolbertam:
For [JAVA-1786](https://datastax-oss.atlassian.net/browse/JAVA-1786)

Adds source code formatting use fmt-maven-plugin much like is done on the 4.x branch.

Since the actual format part dominates the changes, broke up into three commits:

1. 2d31d66 - maven pom changes to add fmt-maven-plugin
2. 52b615b - code change to work around import issue where google-java-format puts static imports first, but static imports come from a class that extends a class that is imported later.  (Will look into filing a bug against google-java-format for that, but not that big of a deal).
3. 046d3c0 - the result of mvn fmt:format.


tolbertam:
Looks like i'll need to do some profile magic to only activate the plugin for jdk8+, will do.

pom.xml
```diff
@@ -87,6 +87,7 @@
         <test.groups>unit</test.groups>
         <test.osgi.skip>true</test.osgi.skip>
         <javadoc.opts />
+        <format.validateOnly>true</format.validateOnly>
```
adutra: It seems that this system property is not honored anymore (or never was?). 
This is also a problem affecting the new driver (4.x).
And besides I honestly find it very annoying that by default the build only checks the files, but doesn't format them. Couldn't we switch to the `format` goal instead?

adutra (changes requested):

pom.xml
tolbertam: You are right, this option is no longer valid since the maven build uses `check`.  To format, you have to run `mvn fmt:format` explicitly.  I'll remove the property and also make sure we have that documented.

tolbertam:
This also needs to update imports to be consistent with our convention in terms of not using wildcard imports which google-java-format doesn't enforce.

pom.xml
olim7t: Yes, the property is obsolete, it was used by a previous version of the plugin that did everything with a single goal (see coveo/fmt-maven-plugin#14), I forgot to remove it when I upgraded.

> And besides I honestly find it very annoying that by default the build only checks the files, but doesn't format them. Couldn't we switch to the format goal instead?

Big 👎 on this (as already discussed).
If we switch to the format goal, the CI build will pass even if unformatted files are accidentally committed. This makes the check useless.
Failing the build if local changes are not ready to be committed is a good thing. Much like a compile error or a broken unit test.

stamhankar999: Ok, so I see that if someone committed a file before running a build (that performs the formatting), they could commit a file that requires formatting, and we'd want the CI build to fail. 

However, when building locally, I'd want the code to be formatted as part of that build.

Can we use environment variables to alter this behavior locally vs. in CI?

olim7t (approved):
❤️

tolbertam:
Will rebase, remove system property and fix imports and push that shortly.

tolbertam:
Rebased in the following away:

1. Deleted the commit that formatted the source code.
2. git rebase 3.x
3. mvn fmt:format and commited as e033702 and reorder commits so it matched previous branch state.

Fixed imports by doing a 'Reformat Code' with only the 'Organize imports' option checked and using a file mask of '*.java'

pom.xml
tolbertam: I prefer the behavior as it is, if the content of the source code is changing, I want that to be an explicit action, much like how the license formatting works.  On the other hand if we could do something that can enable that for those who don't mind it formatting we could add something.

One possibility is that we could define a profile that does formatting instead of checking, lets call it `format`.  Those who want automatic formatting instead of checking all they would to do is add this to their `/.m2/settings.xml` so the profile is implicitly activated:

```xml
  <activeProfiles>
    <activeProfile>format</activeProfile>
  </activeProfiles>
```

Alternatively, you could add `-Pformat` to your maven build.

Thoughts on that?

tolbertam: patch that does this: https://gist.github.com/tolbertam/5f6d2334bd582a166a95aba5e02da8c9

stamhankar999: 👍 I think this is a great solution. TIL some new Maven magic.

adutra (approved):

pom.xml
tolbertam: Holding off on contributing that for now as need to discuss the idea further and don't want to hold this PR up.  Created [JAVA-1930](https://datastax-oss.atlassian.net/browse/JAVA-1930) for that.

PR #1065, opened 2018-07-29
JAVA-1388: Add dynamic port discovery for system.peers_v2

tolbertam:
For [JAVA-1388](https://datastax-oss.atlassian.net/browse/JAVA-1388).

Based on #1064 for now until that is merged.

I followed the approaches in #1000 where we query the peers_v2 table first, and if that fails, downgrade to the peers table and never query v2 again.  I just realized we should probably be more selective about downgrading based on the error returned by that (both here and on 4.x), as its reasonable to expect we can get other errors (like client timeouts) that don't necessitate downgrading to the peers table.

I updated `ScassandraCluster` to support multiple nodes per IP and added a test for this, and it seems to work well.   It would be nice to have tests against real cassandra nodes, which will require CCM changes, which may take me a little bit.  Also needs tests for the broadcast address changing / query of peers table for a single host that we currently have in `ControlConnectionTest` for the existing peers table.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -1686,11 +1686,15 @@ public class Cluster implements Closeable {
       return configuration.getPolicies().getReconnectionPolicy();
     }
 
+    InetSocketAddress translateAddress(InetSocketAddress address) {
+      InetSocketAddress translated =
+          configuration.getPolicies().getAddressTranslator().translate(address);
+      return translated == null ? address : translated;
+    }
+
     InetSocketAddress translateAddress(InetAddress address) {
       InetSocketAddress sa = new InetSocketAddress(address, connectionFactory.getPort());
-      InetSocketAddress translated =
-          configuration.getPolicies().getAddressTranslator().translate(sa);
-      return translated == null ? sa : translated;
+      return translateAddress(sa);
     }
 
     private Session newSession() {
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -773,10 +773,11 @@ public class Cluster implements Closeable {
 
     /**
      * Enables host port discovery using the system.peers_v2 table added in Cassandra 4.0 (via
-     * CASSANDRA-7544). This enables running multiple Cassandra
+     * CASSANDRA-7544). This enables running multiple Cassandra nodes on the same IP address with
+     * each having its own set of ports.
      *
      * <p>Use of this option only works for clusters running version Cassandra 4.0 or newer. Using
-     * it with an older version of Cassandra will likely cause initialization to fail.
+     * it with an older version of Cassandra will cause initialization to fail.
      *
      * <p>When using this option configuration provided via {@link #withPort(int)} is unused as
      * ports are resolved from Cassandra.
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -717,6 +717,7 @@ public class Cluster implements Closeable {
     private SSLOptions sslOptions = null;
     private boolean metricsEnabled = true;
     private boolean jmxEnabled = true;
+    private boolean allowHostPortDiscovery = false;
     private boolean allowBetaProtocolVersion = false;
     private boolean noCompact = false;
 
@@ -760,11 +761,30 @@ public class Cluster implements Closeable {
      *
      * <p>If not set through this method, the default port (9042) will be used instead.
      *
+     * <p>If {@link #allowHostPortDiscovery()} is used, this value is ignored.
+     *
      * @param port the port to set.
      * @return this Builder.
      */
     public Builder withPort(int port) {
       this.port = port;
+      return this;
+    }
+
+    /**
+     * Enables host port discovery using the system.peers_v2 table added in Cassandra 4.0 (via
+     * CASSANDRA-7544). This enables running multiple Cassandra
+     *
+     * <p>Use of this option only works for clusters running version Cassandra 4.0 or newer. Using
+     * it with an older version of Cassandra will likely cause initialization to fail.
+     *
+     * <p>When using this option configuration provided via {@link #withPort(int)} is unused as
+     * ports are resolved from Cassandra.
+     *
+     * @return this builder.
+     */
+    public Builder allowHostPortDiscovery() {
+      this.allowHostPortDiscovery = true;
       return this;
     }
 
@@ -1338,7 +1358,8 @@ public class Cluster implements Closeable {
                   maxSchemaAgreementWaitSeconds,
                   sslOptions,
                   authProvider,
-                  noCompact)
+                  noCompact,
+                  allowHostPortDiscovery)
               .setCompression(compression);
 
       MetricsOptions metricsOptions = new MetricsOptions(metricsEnabled, jmxEnabled);
--- a/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Cluster.java
@@ -717,7 +717,6 @@ public class Cluster implements Closeable {
     private SSLOptions sslOptions = null;
     private boolean metricsEnabled = true;
     private boolean jmxEnabled = true;
-    private boolean allowHostPortDiscovery = false;
     private boolean allowBetaProtocolVersion = false;
     private boolean noCompact = false;
 
@@ -761,31 +760,11 @@ public class Cluster implements Closeable {
      *
      * <p>If not set through this method, the default port (9042) will be used instead.
      *
-     * <p>If {@link #allowHostPortDiscovery()} is used, this value is ignored.
-     *
      * @param port the port to set.
      * @return this Builder.
      */
     public Builder withPort(int port) {
       this.port = port;
-      return this;
-    }
-
-    /**
-     * Enables host port discovery using the system.peers_v2 table added in Cassandra 4.0 (via
-     * CASSANDRA-7544). This enables running multiple Cassandra nodes on the same IP address with
-     * each having its own set of ports.
-     *
-     * <p>Use of this option only works for clusters running version Cassandra 4.0 or newer. Using
-     * it with an older version of Cassandra will cause initialization to fail.
-     *
-     * <p>When using this option configuration provided via {@link #withPort(int)} is unused as
-     * ports are resolved from Cassandra.
-     *
-     * @return this builder.
-     */
-    public Builder allowHostPortDiscovery() {
-      this.allowHostPortDiscovery = true;
       return this;
     }
 
@@ -1359,8 +1338,7 @@ public class Cluster implements Closeable {
                   maxSchemaAgreementWaitSeconds,
                   sslOptions,
                   authProvider,
-                  noCompact,
-                  allowHostPortDiscovery)
+                  noCompact)
               .setCompression(compression);
 
       MetricsOptions metricsOptions = new MetricsOptions(metricsEnabled, jmxEnabled);
--- a/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
@@ -703,8 +703,14 @@ class ControlConnection implements Connection.Owner {
 
             @Override
             public void onFailure(Throwable t) {
-              isPeersV2 = false;
-              MoreFutures.propagateFuture(peersFuture, selectPeersFuture(connection));
+              // downgrade to system.peers if we get an invalid query as this indicates
+              // the peers_v2 table does not exist.
+              if (t instanceof InvalidQueryException) {
+                isPeersV2 = false;
+                MoreFutures.propagateFuture(peersFuture, selectPeersFuture(connection));
+              } else {
+                peersFuture.setException(t);
+              }
             }
           });
       return peersFuture;
--- a/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
@@ -85,8 +85,11 @@ class ControlConnection implements Connection.Owner {
 
   private volatile boolean isShutdown;
 
+  private final boolean isPeersV2;
+
   public ControlConnection(Cluster.Manager manager) {
     this.cluster = manager;
+    this.isPeersV2 = manager.configuration.getProtocolOptions().isAllowHostPortDiscovery();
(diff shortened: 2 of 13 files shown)


driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
```diff
@@ -621,9 +649,61 @@ private static void updateLocationInfo(
...
+   *       otherwise query <code>system.peers</code>.
+   *   <li>if <code>system.peers_v2</code> query fails, set <code>isPeersV2</code> to false and call
+   *       selectPeersFuture again.
+   * </ol>
+   *
+   * @param connection connection to send request on.
+   * @return result of peers query.
+   */
+  private ListenableFuture<ResultSet> selectPeersFuture(final Connection connection) {
+    if (isPeersV2) {
```
tolbertam, line 692: For < C* 4.x, maybe we should just not even try the peers_v2 table.  We could only allow this for protocol V5+.  I think that's an ok limitation, and we can add some kind of way to bypass the V5 requirement for our Scassandra tests.

tolbertam: Thinking about this more. I think for 3.x, this should be more of an opt-in thing via a builder option.  For a minor version making an extra query on initialization seems kinda iffy, and I think it's ok to require use of this feature to be explicit, since the 3.x driver is more driven around ip addresses and global setting of port (via `withPort`), so we should probably make a separate option so users can opt-in to this capability.

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
```diff
@@ -621,9 +641,27 @@ private static void updateLocationInfo(
...
-      Cluster.Manager cluster,
+  /**
+   * Submits query to the appropriate system.peers table based on whether or not {@link
+   * ProtocolOptions#isAllowHostPortDiscovery()} is enabled.
+   *
+   * @param connection connection to send request on.
+   * @return result of peers query.
+   */
+  private ListenableFuture<ResultSet> selectPeersFuture(final Connection connection) {
+    String query =
```
tolbertam: This is very explicit now, the `peers_v2` is only ever queried if `allowHostPortDiscovery` is used, and if the query fails, cluster initialization fails.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -724,6 +727,23 @@ public Builder withPort(int port) {
       return this;
     }
 
+    /**
+     * Enables host port discovery using the system.peers_v2 table added in Cassandra 4.0 (via
+     * CASSANDRA-7544). This enables running multiple Cassandra
```
adutra: Seems like the last sentence is truncated.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -724,6 +727,23 @@ public Builder withPort(int port) {
...
+     *
+     * <p>Use of this option only works for clusters running version Cassandra 4.0 or newer. Using
+     * it with an older version of Cassandra will likely cause initialization to fail.
+     *
+     * <p>When using this option configuration provided via {@link #withPort(int)} is unused as
+     * ports are resolved from Cassandra.
+     *
+     * @return this builder.
+     */
+    public Builder allowHostPortDiscovery() {
```
adutra: Why expose this? Can't we simply detect the presence of the new peers_v2 table? Or is it because if the cluster has some nodes with peers_v2 and others without, that might be a too complex situation to handle?

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
```diff
@@ -571,20 +585,26 @@ private static void updateInfo(
     // After CASSANDRA-9603 (2.0.17, 2.1.8, 2.2.0 rc2) local row contains one more column:
     // - listen_address
 
-    InetAddress broadcastAddress = null;
+    InetSocketAddress broadcastAddress = null;
     if (row.getColumnDefinitions().contains("peer")) { // system.peers
-      broadcastAddress = row.getInet("peer");
+      int broadcastPort =
```
adutra, line 624: Instead of zero, shouldn't we use the default port number (9042)?

driver-core/src/main/java/com/datastax/driver/core/Host.java
```diff
@@ -194,12 +194,26 @@ public InetSocketAddress getSocketAddress() {
...
   /**
-   * Returns the node broadcast address (that is, the IP by which it should be contacted by other
-   * peers in the cluster), if known.
+   * @return the node broadcast address, if known. Otherwise {@code null}.
+   * @see #getBroadcastSocketAddress()
+   * @see <a
+   *     href="https://docs.datastax.com/en/cassandra/2.1/cassandra/configuration/configCassandra_yaml_r.html">The
+   *     cassandra.yaml configuration file</a>
+   */
+  public InetAddress getBroadcastAddress() {
```
adutra, line 207: This method is now used only in tests. Are you keeping it for binary compatibility? In this case we might want to mark it `@Deprecated` just as you did for `getListenAddress()`.

driver-core/src/main/java/com/datastax/driver/core/Host.java
```diff
@@ -212,13 +226,29 @@ public InetSocketAddress getSocketAddress() {
...
-  public InetAddress getBroadcastAddress() {
+  public InetSocketAddress getBroadcastSocketAddress() {
     return broadcastAddress;
   }
 
   /**
-   * Returns the node listen address (that is, the IP the node uses to contact other peers in the
-   * cluster), if known.
+   * @return the node listen address, if known. Otherwise {@code null}.
+   * @deprecated
```
adutra: You might want to explain why it is deprecated ("Use #getListenSocketAddress() instead").

adutra (changes requested):

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
tolbertam: woops, might have accidentally triggered a vim hotkey :laughing: will fix.

driver-core/src/main/java/com/datastax/driver/core/Host.java
tolbertam: I originally marked these methods deprecated, but then I backed off when I realized for most users the `InetSocketAddress` version will return port 0.

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
tolbertam: This is for the broadcast/listen address which is typically 7000.  I felt that since we can't know for sure what it's configured to, using 0 would be better.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
tolbertam: I explain my rationale for this here: https://github.com/datastax/java-driver/pull/1065#discussion_r206263173.  I originally tried peers_v2 first, then peers if the table doesn't exist (like 4.x does).   But thinking about it more I wasn't sure if adding another query in a minor version was the right thing to do, so I thought maybe making this an Opt-in feature on 3.x would be the right way to go.

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
adutra: Oh sorry, of course. 👍

driver-core/src/main/java/com/datastax/driver/core/Host.java
adutra: Makes sense but then for consistency maybe we should _not_ mark `getListenAddress()` as deprecated?

tolbertam: D'oh!  Didn't realize that, I'll make them consistent, thanks.

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
GregBestland: One thing to think about here is whether or not we should fall back to the old system peers query if we get an invalid query request back. I think in that specific case it might be sane to fall back on old peers query.

GregBestland (approved):
+1 looks good one Just one small thing to consider. Merge at your discretion

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
tolbertam: I've wavered back and forth on this, my current line of thinking is if the user opts in using  `allowHostPortDiscovery`, they are intending to depend on the presence of the peers_v2 table to tell them what ports to use.  In absence of that table, I think failing is the right thing to do.  Whereas with java driver 4.x, we always check this table first, so in that case it makes sense to downgrade to the peers table.

tolbertam:
Tests surfaced some very minor issues (log message contents, etc.) that i need to work out. will do.

tolbertam:
rebased and fixed/enhanced tests.

adutra (approved):

tolbertam:
Rebased on 3.x

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
olim7t: Doesn't sound that bad to add that extra query, since it's done only once at the beginning, and if that fails then we always go directly to the legacy table.

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
```diff
@@ -434,81 +441,94 @@ void refreshNodeListAndTokenMap() {
...
+        // Don't include full address if port is 0.
+        String addressToUse =
+            address.getPort() != 0 ? address.toString() : address.getAddress().toString();
         logger.debug(
             "Could not find peer with broadcast address {}, "
                 + "falling back to a full system.peers scan to fetch info for {} "
                 + "(this can happen if the broadcast address changed)",
-            host.getBroadcastAddress(),
+            address,
             host);
```
olim7t: I'm a bit lost here: why do we read `broadcastSocketAddress` again?
`addressToUse` is never used.
`address` is only used in the log message.

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
```diff
@@ -594,20 +614,26 @@ private static void updateInfo(
     // After CASSANDRA-9603 (2.0.17, 2.1.8, 2.2.0 rc2) local row contains one more column:
     // - listen_address
```
olim7t, line 617: Maybe update this kind of comments to explain the new rows in `peers_v2` (or remove them / have a single comment detailing the table formats in different versions?).

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
olim7t: So are we doing the fallback to `system.peers` at all anymore? The PR description says so...

driver-core/src/main/java/com/datastax/driver/core/Host.java
```diff
@@ -44,12 +44,12 @@
   // We use that internally because
   // that's the 'peer' in the 'System.peers' table and avoids querying the full peers table in
   // ControlConnection.refreshNodeInfo.
-  private volatile InetAddress broadcastAddress;
+  private volatile InetSocketAddress broadcastAddress;
```
olim7t: Could we rename this and the corresponding setter with `Socket`?
Right now we have:
* getBroadcastAddress returning an InetAddress
* setBroadcastAddress taking an InetSocketAddress

driver-core/src/main/java/com/datastax/driver/core/Host.java
olim7t: We should at least explain that it duplicates information already present in the other method.

olim7t (changes requested):
I'm not really convinced we should make this opt-in. If you connect to a Cassandra 4 cluster with the driver, things should just work.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
tolbertam: I was erring on the side of caution by making this an 'opt-in' feature.  If you don't see that as too risky, I will revert it back to my initial implementation which mirrors the approach we took with 4.x.    With regards to not falling back to `system.peers`, my thought process was that if you explicitly opted in using `allowHostPortDiscovery()`, it didn't seem the backing off to peers was the best idea, since you are expressing that you want to utilize the capability which is not being provided by the server.  If i revert to not making this opt-in that point is moot though.

driver-core/src/main/java/com/datastax/driver/core/Host.java
tolbertam: I'll rename the setters to `*SocketAddress` for consistency.

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
tolbertam: i think this is an artifact of some refactoring, i'll fix this.

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
tolbertam: We discussed this more and decided that it's ok to try `peers_v2` table first.

tolbertam:
Rebased on 3.x and processed feedback, please let me know if I missed anything. The behavior is now 1:1 with 4.x.  The `peers_v2` table will be queried initially, and if we get an `InvalidQueryException`, we fall back to the `peers` table and logic.

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
```diff
@@ -434,81 +443,94 @@ void refreshNodeListAndTokenMap() {
...
+            isPeersV2
+                ? SELECT_PEERS_V2
+                    + " WHERE peer='"
+                    + broadcastAddress.getAddress().getHostAddress()
+                    + "' AND peer_port="
+                    + broadcastAddress.getPort()
+                : SELECT_PEERS
+                    + " WHERE peer='"
+                    + broadcastAddress.getAddress().getHostAddress()
+                    + "'";
```
olim7t, line 508: We don't handle a missing `peers_v2` here, but that's fine: we do a `refreshNodeListAndTokenMap` every time the control connection (re)connects, and it happens before the new connection is published in `connectionRef`, so we know it will never race with this method.
So `isPeersV2` will always be set correctly at this point. Even in a mixed 3.x/4.0 cluster, if we initially connected to a 4.0 node, `isPeersV2` will downgrade correctly if we reconnect to a legacy node.

olim7t (approved):

PR #1066, opened 2018-07-31
JAVA-1927 fix issue with unused native port

GregBestland:
Two things here,
1. Native port is now honored not just in the metadata but also when the the connections are actually made.
2. We only fall back to system_peers when the v2 query errors with an InvalidQueryException

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
@@ -319,7 +319,7 @@ public class DefaultTopologyMonitor implements TopologyMonitor {
       return null;
     }
     Integer row_port = row.getInteger("native_port");
-    if (row_port == null) {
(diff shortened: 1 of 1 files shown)


GregBestland:
Also we need feature/51 merged into simulacron prior to being able to author a integration test for this scenario. I manually validated it with that simulacron branch.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -317,7 +318,10 @@ private InetSocketAddress getNativeAddressFromPeers(AdminRow row) {
     if (nativeAddress == null) {
       return null;
     }
-    int row_port = port;
+    Integer row_port = row.getInteger("native_port");
```
adutra: Please use the Java conventions for variable names: `rowPort`.

adutra (dismissed):

adutra (approved):

tolbertam (approved):

PR #1067, opened 2018-08-06
JAVA-1928: Fix GuavaCompatibility for Guava 26

tolbertam:
For [JAVA-1928](https://datastax-oss.atlassian.net/browse/JAVA-1928)

Motivation:

GuavaCompatibility.isGuava_19_0_OrHigher() depends on a previously
deprecated method
Futures.transformAsync(ListenableFuture, AsyncFunction) which was
removed in Guava 26.
Futures.transformAsync(ListenableFuture, AsyncFunction, Executor)
should be used instead.

In addition there were several methods removed in Guava 26 that
need to be accounted for.

Modifications:

Use Futures.transformAsync(..., Executor) for Guava 19+ detection.

Add GuavaCompatibility.addCallback and transform methods as the
Futures methods that did not require an Executor have since been
removed.  Update methods that used Futures.addCallback and transform
to use this.

Update Version19OrHigher.withFallback to use sameTHreadExecutor().

Update
Version19OrHigher.transformAsync(ListenableFuture, AsyncFunction)
to call version that requires Executor with
MoreExecutions.directExecutor().

Remove the upper bound requirement on the guava version
requirement for OSGi so the driver doesn't need to be updated
every time there is a new guava version.

Result:

Guava26 now functions correctly with the driver.


tolbertam:
Looks like there may be more to this that we need to account for:

```
08:40:13 java.lang.NoSuchMethodError: com.google.common.util.concurrent.Futures.addCallback(Lcom/google/common/util/concurrent/ListenableFuture;Lcom/google/common/util/concurrent/FutureCallback;)V
08:40:13 	at com.datastax.driver.core.CloseFuture$Forwarding.<init>(CloseFuture.java:74)
08:40:13 	at com.datastax.driver.core.HostConnectionPool.closeAsync(HostConnectionPool.java:616)
08:40:13 	at com.datastax.driver.core.SessionManager.closeAsync(SessionManager.java:171)
08:40:13 	at com.datastax.driver.core.Cluster$Manager.close(Cluster.java:1740)
08:40:13 	at com.datastax.driver.core.Cluster$Manager.access$200(Cluster.java:1385)
08:40:13 	at com.datastax.driver.core.Cluster.closeAsync(Cluster.java:578)
08:40:13 	at com.datastax.driver.core.Cluster.close(Cluster.java:590)
08:40:13 	at com.datastax.driver.core.policies.ErrorAwarePolicyIntegrationTest.should_exclude_host_after_reaching_maximum_errors(ErrorAwarePolicyIntegrationTest.java:157)
```

tolbertam:
`MoreExecutors.directExecutor` was added in Guava 18.0, we will likely need a `GuavaCompatibiltiy` method for that.

EDIT:  We already have one, `GuavaCompatibility.sameThreadExecutor`

adutra (approved):

tolbertam:
Unfortunately there's still a few more methods that were removed in Guava 26 that we use.  I'll keep iterating on this until i get a clean run.

tolbertam:
Also this needs rebase since #1064 has been merged

tolbertam:
I rebased on 3.x and also updated things so much that I felt it was ok to do a single commit force push.  I changed code outside of `GuavaCompatibility` that used `Futures.addCallback` and `Future.transform` to now use `GuavaCompatibiliy` versions of those, that way we didn't have to add `GuavaCompatibility.INSTANCE.sameThreadExecutor` everywhere.

tolbertam:
screenshot from 2018-08-06 20-38-50

Full integration suite passed against each guava version.

driver-core/pom.xml
```diff
@@ -228,7 +228,7 @@
                             <manifestLocation>${project.build.outputDirectory}/META-INF</manifestLocation>
                             <instructions>
                                 <!-- JNR does not provide OSGi bundles, so exclude it; the driver can live without it -->
-                                <Import-Package><![CDATA[com.google.common.*;version="[16.0.1,22)",!jnr.*,io.netty.channel.epoll,*]]></Import-Package>
+                                <Import-Package><![CDATA[com.google.common.*;version="16.0.1",!jnr.*,io.netty.channel.epoll,*]]></Import-Package>
```
adutra, line 231: Why did you force the guava version to 16.0.1? Shouldn't this be `[16.0.1,27)`? (Same happens in driver-extras and driver-mapping).

adutra (changes requested):

driver-core/pom.xml
tolbertam: It's not clear, but when you set a specific version it implies "this version or later", from [section 3.2.6 of the OSGi core specification](https://osgi.org/specification/osgi.core/7.0.0/framework.module.html#i3189032):

> If a version range is specified as a single version, it must be interpreted as the range [version,∞). The default for a non-specified version range is 0, which maps to [0.0.0,∞).

adutra: Oh TIL :) Thank you for enlightening me. Agreed that it is much more flexible to specify a single version then.

adutra (approved):

mmoayyed:
Apologies for resurrecting a merged thread; having looked around some I didn't see this as an option so thought to ask: is there an opportunity for one to exercise this patch in form of a `3.6.0-SNAPSHOT` version from the likes of sonatype snapshot repositories? Or should one build the driver from source, etc?

tolbertam:
Hi @mmoayyed , no problem at all.   We don't publish snapshots of the driver currently (although could be something we could consider).   Building from source is the best option.

We'll be formally releasing 3.6.0 in the coming weeks which will include this :+1:

mmoayyed:
Thank you.

Is the publishing of snapshots something you're interested to do, and one I can help with? Perhaps through Travis CI? Create a separate issue? 

(I can definitely share instructions, notes, examples and such but the work needs to be done by someone with commit access to the repository, and one who can encrypt things via Travis CI, etc)

PR #1069, opened 2018-08-07
Make ProtocolFeature an interface

adutra:
Motivation:

ProtocolFeature is currently an enum and as such is not extensible;
if custom protocol extensions need to support proprietary features,
such features cannot be modeled as a ProtocolFeature instance.

Modification:

Make ProtocolFeature an interface, create enum DefaultProtocolFeature.

Result:

ProtocolFeature can now be implemented by other classes.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/ProtocolVersion.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/ProtocolVersion.java
@@ -43,8 +43,8 @@ public interface ProtocolVersion {
   /**
    * Whether the protocol version is in a beta status.
    *
-   * <p>Beta versions are intended for Cassandra development. They should be used in a regular
-   * application, beta features may break at any point.
+   * <p>Beta versions are intended for Cassandra development. They should not be used in a regular
+   * application, as beta features may break at any point.
    */
   boolean isBeta();
 }
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/CassandraProtocolVersionRegistry.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/CassandraProtocolVersionRegistry.java
@@ -170,13 +170,12 @@ public class CassandraProtocolVersionRegistry implements ProtocolVersionRegistry
 
   @Override
   public boolean supports(ProtocolVersion version, ProtocolFeature feature) {
-    switch (feature) {
-      case UNSET_BOUND_VALUES:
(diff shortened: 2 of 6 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/ProtocolVersion.java
```diff
@@ -43,8 +43,8 @@
   /**
    * Whether the protocol version is in a beta status.
    *
-   * <p>Beta versions are intended for Cassandra development. They should be used in a regular
-   * application, beta features may break at any point.
+   * <p>Beta versions are intended for Cassandra development. They should not be used in a regular
+   * application, as beta features may break at any point.
```
adutra, line 47: I'm proposing a fix for what I believe is a misconceived statement.

core/src/main/java/com/datastax/oss/driver/internal/core/CassandraProtocolVersionRegistry.java
```diff
@@ -170,13 +170,12 @@ public ProtocolVersion highestCommon(Collection<Node> nodes) {
 
   @Override
   public boolean supports(ProtocolVersion version, ProtocolFeature feature) {
```
adutra, line 172: Ideally, I would rather have this method directly on `ProtocolVersion`: `boolean supports(ProtocolFeature feature)` which seems more natural. However this requires promoting `ProtocolFeature` to the API packages, and I am not sure at this point if this would be a good idea.

tolbertam: Agreed, that does seem more natural, but also that promoting to API level isn't a good idea at this point.

tolbertam (approved):
:+1: makes sense.  This fits in similarly to `DefaultDriverOption`.

core/src/main/java/com/datastax/oss/driver/internal/core/CassandraProtocolVersionRegistry.java
olim7t: Also, it would make it impossible for `DefaultProtocolFeature` instances to handle custom features of downstream projects.

olim7t (approved):
`ProtocolFeature` was initially intended for the common code in the core, but indeed downstream projects can have protocol-dependent features as well, and there's no reason not to reuse the same mechanism.

PR #1071, opened 2018-08-07
JAVA-1903: Add WhiteListPolicy.ofHosts

tolbertam:
For [JAVA-1903](https://datastax-oss.atlassian.net/browse/JAVA-1903)

Motivation:

Similarly to Cluster.Builder.addContactPoints(String ...) it would be
convenient for WhiteListPolicy to support a similar mechanism of
providing hostnames to resolve IP addresses from to be considered
for host selection.

Modifications:

Add WhiteListPolicy.fromHosts to create a WhiteListPolicy that uses
DNS to resolve IP addresses to be considered.

Result:

WhiteListPolicy can now be created from hostname(s).

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/policies/WhiteListPolicy.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/policies/WhiteListPolicy.java
@@ -21,10 +21,8 @@ import com.google.common.collect.ImmutableSet;
 import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
+import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Set;
 
 /**
  * A load balancing policy wrapper that ensure that only hosts from a provided white list will ever
@@ -59,7 +57,7 @@ public class WhiteListPolicy extends HostFilterPolicy {
 
   /**
    * Private constructor solely for maintaining type from policy created by {@link
-   * #fromHosts(LoadBalancingPolicy, String...)}.
+   * #ofHosts(LoadBalancingPolicy, String...)}.
    */
   private WhiteListPolicy(LoadBalancingPolicy childPolicy, Predicate<Host> predicate) {
     super(childPolicy, predicate);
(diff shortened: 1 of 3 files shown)


driver-core/src/main/java/com/datastax/driver/core/policies/WhiteListPolicy.java
```diff
@@ -61,4 +74,51 @@ public boolean apply(Host host) {
       }
     };
   }
+
+  /**
+   * Creates a new policy with the given host names.
+   *
+   * <p>See {@link #fromHosts(LoadBalancingPolicy, String...)} for more details.
+   */
+  public static WhiteListPolicy fromHosts(
```
tolbertam: I had considered adding a `WhiteListPolicy(LoadBalancingPolicy childPolicy, Collection<String> hostnames)` but that doesn't work because of type erasure.  Having a static method seems reasonable in any case.

csplinter (approved):
+1

I ran a quick test using `fromHosts` and a single hostname that resolves to 3 different addresses and things worked as expected. Before this fix this test fails with "Cannot use HostFilterPolicy where the filter allows none of the contacts points"

driver-core/src/main/java/com/datastax/driver/core/policies/WhiteListPolicy.java
adutra: I like the static factory method but wouldn't `ofHosts` be more meaningful than `fromHosts`?

driver-core/src/main/java/com/datastax/driver/core/policies/WhiteListPolicy.java
```diff
@@ -61,4 +74,51 @@ public boolean apply(Host host) {
...
+
+  /**
+   * Creates a new policy that wraps the provided child policy but only "allows" hosts having
+   * addresses that match those from the resolved input host names.
+   *
+   * <p>Note that all host names must be resolvable; if <em>any</em> of them cannot be resolved,
+   * this method will fail.
+   *
+   * @param childPolicy the wrapped policy.
+   * @param hostnames list of host names to resolve white listed addresses from.
```
adutra: Maybe add a note here that the individual names cannot be null?

driver-core/src/main/java/com/datastax/driver/core/policies/WhiteListPolicy.java
```diff
@@ -61,4 +74,51 @@ public boolean apply(Host host) {
...
     };
   }
+
+  /**
+   * Creates a new policy with the given host names.
+   *
+   * <p>See {@link #fromHosts(LoadBalancingPolicy, String...)} for more details.
+   */
+  public static WhiteListPolicy fromHosts(
+      LoadBalancingPolicy childPolicy, Collection<String> hostnames) {
```
adutra: How about `Iterable<String>` instead of `Collection<String>`? (Then you could simply implement the logic here and make the overloaded method below call this one with `Arrays.asList(hostnames)`.)

driver-core/src/main/java/com/datastax/driver/core/policies/WhiteListPolicy.java
```diff
@@ -61,4 +74,51 @@ public boolean apply(Host host) {
...
+   *
+   * @param childPolicy the wrapped policy.
+   * @param hostnames list of host names to resolve white listed addresses from.
+   * @throws IllegalArgumentException if any of the given {@code hostnames} could not be resolved.
+   * @throws NullPointerException If null was provided for a hostname.
+   * @throws SecurityException if a security manager is present and permission to resolve the host
+   *     name is denied.
+   */
+  public static WhiteListPolicy fromHosts(LoadBalancingPolicy childPolicy, String... hostnames) {
+    final Set<InetAddress> validAddresses = new HashSet<InetAddress>();
```
adutra: Shouldn't we use a concurrent/immutable structure here, e.g. `ImmutableSet`? It will be accessed concurrently from within the predicate created below so I think it would be safer to make it thread-safe and immutable.

adutra (changes requested):

driver-core/src/main/java/com/datastax/driver/core/policies/WhiteListPolicy.java
tolbertam: Makes sense to do that, especially to be consistent with the other predicate implementation.

driver-core/src/main/java/com/datastax/driver/core/policies/WhiteListPolicy.java
tolbertam: I used `Collection` for consistency with `Cluster.Builder.addContactPoints`, but I agree that using `Iterable` is better because it is less specific.

tolbertam:
Thank you both for reviewing.  I have processed feedback and pushed a new commit, 29a58a9.

adutra (approved):

PR #1073, opened 2018-08-08
JAVA-1771: Send driver name and version in STARTUP message

tolbertam:
Motivation:

CASSANDRA-14275 adds the capability to parse driver identification
information from the STARTUP message in the form of DRIVER_NAME
and DRIVER_VERSION options.  This is useful for helping operators
understand what clients are connecting to Cassandra and what
drivers they are using.

Modifications:

DRIVER_NAME and DRIVER_VERSION configuration are now provided in
the STARTUP options map.  This is provided regardless of the
version of Cassandra we are connecting to, as this is not
a protocol version specific feature.

Result:

The driver will now send identifiying information in STARTUP
message on connection initialization.


driver-core/src/main/java/com/datastax/driver/core/Requests.java
```diff
@@ -31,6 +31,9 @@ private Requests() {}
   static class Startup extends Message.Request {
     private static final String CQL_VERSION_OPTION = "CQL_VERSION";
     private static final String CQL_VERSION = "3.0.0";
+    private static final String DRIVER_VERSION_OPTION = "DRIVER_VERSION";
+    private static final String DRIVER_NAME_OPTION = "DRIVER_NAME";
+    private static final String DRIVER_NAME = "DataStax Java Driver for Apache Cassandra";
```
adutra: Should we add `Apache Cassandra (R)` or am I being paranoid?

adutra (changes requested):

driver-core/src/main/java/com/datastax/driver/core/Requests.java
tolbertam: Follow up: We just chatted about this and decided to just use "DataStax Java Driver" for now.  Including "Apache Cassandra" is probably over descriptive, since this information will be surfaced in nodetool and virtual table output, so it doesn't really add anything.

adutra (approved):

olim7t (approved):

PR #1074, opened 2018-08-08
JAVA-1211, JAVA-1220, JAVA-1929

GregBestland:


GregBestland:
I realize it's not typical to put multiple tickets in the same pr, but these very much depend on one another, and are part of a more general fix to add sanity checkout around cluster/session state.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1448,119 +1448,125 @@ private Manager(
     // on it so synchronized is good enough.
     synchronized void init() {
       checkNotClosed(this);
-      if (isInit) return;
+      if (isInit) {
+        if (initException == null) {
+          return;
+        }
+        throw new IllegalStateException(
```
GregBestland: We should only end up here if initialization was attempted, and did not complete. This method is synchronized so we don't need to worry about a race condition.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1722,9 +1729,15 @@ private CloseFuture close() {
         logger.debug("Shutting down");
 
         // stop debouncers
-        nodeListRefreshRequestDebouncer.stop();
-        nodeRefreshRequestDebouncer.stop();
-        schemaRefreshRequestDebouncer.stop();
+        if (nodeListRefreshRequestDebouncer != null) {
```
GregBestland, line 1745: Additional null checks to allow for a partially initialized cluster to still be closed

driver-core/src/test/java/com/datastax/driver/core/ClusterInitTest.java
```diff
@@ -278,6 +278,69 @@ public void should_not_abort_init_if_host_does_not_support_protocol_version() {
       scassandraCluster.stop();
     }
   }
+  /**
```
GregBestland, line 283: I didn't add a test for 1211 because is validated implicitly by now catching all runtime execptions in cluster init, and closing the cluster in various states of failed initialization in our other tests.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1448,119 +1448,125 @@ private Manager(
     // on it so synchronized is good enough.
     synchronized void init() {
       checkNotClosed(this);
-      if (isInit) return;
+      if (isInit) {
+        if (initException == null) {
```
tolbertam: It would be better to move this logic to `checkNotClosed`, as that is used in a number of places and offers the opportunity to notify the user that there was an error during initialization (which is more specific than cluster was closed)

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1448,119 +1448,125 @@ private Manager(
...
+          return;
+        }
+        throw new IllegalStateException(
+            "Error during cluster initialization, please close and retry");
```
tolbertam: Should make sure to include the cause (i.e. `throw new IllegalStateException(..., initException)`.

Also the message should be clarified because the user shouldn't need to call `close()`, we should (and it looks like we already) do that for them.  It should instead say something like `Error encountered during cluster initialization, please create a new Cluster instance`.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -692,6 +695,9 @@ else if (fetchSize != Integer.MAX_VALUE)
    * and handle host failover.
    */
   void execute(final RequestHandler.Callback callback, final Statement statement) {
+    if (cluster.isClosed()) {
```
tolbertam: It's also possible that the Session itself is closed, so we should check that too and surface a specific error in that case.

The error message could be something like "Could not send request, <X> is closed." where X is either 'cluster' or 'session'.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -137,6 +137,9 @@ public String getLoggedKeyspace() {
 
   @Override
   public ResultSetFuture executeAsync(final Statement statement) {
+    if (cluster.isClosed()) {
```
tolbertam: I think this isn't needed, since this ultimately calls the `execute` method you also have logic for further below.  It's better to just rely on that since it will fail the future instead of the calling thread which is nicer for users since then they don't need to handle exceptions in two different places.

driver-core/src/test/java/com/datastax/driver/core/ClusterInitTest.java
```diff
@@ -278,6 +278,69 @@ public void should_not_abort_init_if_host_does_not_support_protocol_version() {
...
+            .build();
+
+    try {
+      scassandraCluster.init();
+      Session session = cluster.connect();
+      cluster.close();
+      try {
+        session.execute("SELECTS * FROM system.peers");
+        fail(
+            "This error when session.execute is called on session associated with closed cluster.");
```
tolbertam: This is sort of awkwardly phrased, should change to something like:

"Should have failed if session.execute called when cluster was closed."

driver-core/src/test/java/com/datastax/driver/core/ClusterInitTest.java
```diff
@@ -278,6 +278,69 @@ public void should_not_abort_init_if_host_does_not_support_protocol_version() {
...
   }
+  /**
+   * Ensures that if an error occurs doing initialization that subsequent attempts to use the
+   * cluster result in an appropriate error.
+   *
+   * @jira_ticket JAVA-1220
+   * @test_category host:state
+   */
+  @Test(groups = "short")
+  public void should_detect_cluster_init_failure() {
```
tolbertam: This passes even with 3.x in its current state, we should also have a more specific test that would have previously failed.

tolbertam: This test will produce the failure that is detailed in JAVA-1220:

```java
  @Test(groups = "short")
  public void should_detect_cluster_init_failure_caused_by_bad_credentials() {
    Cluster cluster =
        Cluster.builder()
            .addContactPoints("127.0.0.1")
            .withCredentials("bad", "pass")
            .withNettyOptions(nonQuietClusterCloseOptions)
            .build();

    try {
      cluster.connect();
      fail("Should not have been able to connect.");
    } catch (RuntimeException e) {
      try {
        cluster.connect();
        fail("Should throw IllegalStateException when attempting to connect again.");
      } catch (IllegalStateException e1) {
        assertThat(
            e1.getMessage().equals("Error during cluster initialization, please close and retry"));
      }
    } finally {
      cluster.close();
    }
  }
```

produces:

```
java.lang.IllegalArgumentException: initialArraySize cannot be negative but was: -2147483648
	at com.google.common.collect.CollectPreconditions.checkNonnegative(CollectPreconditions.java:39)
	at com.google.common.collect.Lists.newArrayListWithCapacity(Lists.java:186)
	at com.datastax.driver.core.HostConnectionPool.initAsync(HostConnectionPool.java:135)
	at com.datastax.driver.core.SessionManager.replacePool(SessionManager.java:362)
	at com.datastax.driver.core.SessionManager.maybeAddPool(SessionManager.java:397)
	at com.datastax.driver.core.SessionManager.createPools(SessionManager.java:129)
	at com.datastax.driver.core.SessionManager.initAsync(SessionManager.java:96)
	at com.datastax.driver.core.Cluster.connectAsync(Cluster.java:378)
	at com.datastax.driver.core.Cluster.connectAsync(Cluster.java:355)
	at com.datastax.driver.core.Cluster.connect(Cluster.java:305)
	at com.datastax.driver.core.ClusterInitTest.should_detect_cluster_init_failure(ClusterInitTest.java:303)
```

This would need to be updated to fire up a CCM cluster with authentication (see `AuthenticationTest` for reference).  To keep things simple you could just add this test to `AuthenticationTest`

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
tolbertam: It looks overly paranoid, but I think it's worth it :)

tolbertam (changes requested):
Looks good for the most part :+1: , had some suggestions.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1387,7 +1387,7 @@ private static String generateClusterName() {
     final String clusterName;
     private boolean isInit;
     private volatile boolean isFullyInit;
-
+    private Exception initException;
```
adutra: When I read this I thought that a `volatile` keyword would be required for this field, but it turns out that it is already guarded by `this` thanks to the `synchronized` block. 
_However_ : out of curiosity I checked usages of `isInit`, and IMO it should be volatile since it is accessed without holding the object's lock in `com.datastax.driver.core.Cluster.Manager.close()`. Wdyt?

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
adutra: I would even add: "_This cluster has been closed due to an_ error encountered during its initialization, please create a new Cluster instance.".

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
adutra: Hmm it might be required, `isInit` remains true even after the cluster is closed, and in this case we don't want the flow to reach the if block below.

adutra (changes requested):

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
tolbertam: :+1:, That's even better, makes it clear that it was already closed and errored out, not that the current activity caused it to error out.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
GregBestland: I added this specifically for isInit case.

driver-core/src/test/java/com/datastax/driver/core/ClusterInitTest.java
GregBestland: So this actually hits the exact same code path as my test does, just in a different way. I'm going to keep mine as it relies on simulacron rather than auth, and therefore will have a shorter run time. The test above produces the same dump now that I've fixed it.

tolbertam: I would still suggest adding my proposed test to `AuthenticationTest` since it produces the root issue of JAVA-1220.   It reuses the running ccm cluster for that test, so it adds minimal time to the overall test suite execution.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
tolbertam: If it's still required, we should change it to fail the future, not throw an Exception.

driver-core/src/test/java/com/datastax/driver/core/AuthenticationTest.java
```diff
@@ -76,12 +76,7 @@ public void should_fail_to_connect_with_wrong_credentials() throws InterruptedEx
...
                 .withCredentials("bogus", "bogus")
                 .build());
-    try {
-      cluster.connect();
-    } finally {
-      assertThat(cluster.getMetrics().getErrorMetrics().getAuthenticationErrors().getCount())
-          .isEqualTo(1);
-    }
+    cluster.connect();
   }
```
GregBestland, line 107: We now close the cluster on initialization errors. This means the cluster will be closed, when we call getMetrics, so this check is no longer applicable.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -692,6 +699,12 @@ else if (fetchSize != Integer.MAX_VALUE)
    * and handle host failover.
    */
   void execute(final RequestHandler.Callback callback, final Statement statement) {
+    if (this.isClosed()) {
+      throw new IllegalStateException("Could not send request, session is closed");
+    }
+    if (cluster.isClosed()) {
+      throw new IllegalStateException("Could not send request, cluster is closed");
+    }
```
olim7t: Closing a cluster also closes all of its sessions automatically, so it would be caught by the first test, we don't need to re-check here.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -137,6 +137,13 @@ public String getLoggedKeyspace() {
 
   @Override
   public ResultSetFuture executeAsync(final Statement statement) {
+    if (this.isClosed()) {
+      throw new IllegalStateException("Could not send request, session is closed");
+    }
+
+    if (cluster.isClosed()) {
+      throw new IllegalStateException("Could not send request, cluster is closed");
+    }
```
olim7t: If `!isInit`, we take the second branch below and call `execute()`, which does this again.

olim7t (changes requested):
I have some doubts about JAVA-1229: the proposed solution is not thread-safe (it's a check-then-act), it would be prohibitively complicated to make it so, and that's probably why it hasn't been done before.

What problem are we really trying to solve here? Is it an actual bug report, or just a nice-to-have? If a client application shuts down the driver, shouldn't it stop using it? For example, I think a properly designed webapp should stop accepting HTTP requests before it start shutting down external resources, so it shouldn't run into that issue.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
tolbertam: That was my intuition too, but there is some debate about that (see [discussion here](https://github.com/datastax/java-driver/pull/1074#discussion_r208751693)).  I need to test it out but i think you are right.

tolbertam:
> What problem are we really trying to solve here? Is it an actual bug report, or just a nice-to-have? If a client application shuts down the driver, shouldn't it stop using it? For example, I think a properly designed webapp should stop accepting HTTP requests before it start shutting down external resources, so it shouldn't run into that issue.

The gist behind [JAVA-1929](https://datastax-oss.atlassian.net/browse/JAVA-1929) is that the exception surfaced when executing a query and the cluster is closed is too vague.  `NoHostAvailableException (no host was tried)` could mean a lot of things:

* The driver thinks all hosts were down
* The lb policy couldn't identify a host
* The cluster/session was already closed.
* and probably more

There was recently an occurrence where someone ran into this.  It took us time (a couple of days of back and forth) to identify that some part of the users code caused the Cluster to be closed because of some other exception.  If we can just tell the user that the Cluster is closed, that gives them something to look into.

If we can be more explicit, it is easier for the user to know *why* they are getting the error.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
tolbertam: I see, the logic at line 151 addresses `RequestHandler(...).sendRequest()` directly, if it called `execute(future, statement)` instead, we wouldn't need this.   I think we should change that.

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -692,6 +699,12 @@ else if (fetchSize != Integer.MAX_VALUE)
    * and handle host failover.
    */
   void execute(final RequestHandler.Callback callback, final Statement statement) {
+    if (this.isClosed()) {
+      throw new IllegalStateException("Could not send request, session is closed");
```
tolbertam: Just a reminder that this should not throw the exception, but rather fail the future (the callback in this case I suppose)

olim7t:
>  If we can just tell the user that the Cluster is closed, that gives them something to look into.

Fair enough. If it is meant to help debug errors like this, then it's not too bad that it's racy.

We'll need to do it in 4.x as well, currently I think it behaves the same way. I'll create a ticket.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -1720,6 +1722,17 @@ boolean isClosed() {
       return closeFuture.get() != null;
     }
 
+    boolean errorDuringInit() {
+      if (isInit && initException != null) {
+        return true;
+      }
+      return false;
```
olim7t: This can be simplified as `return (isInit && initException != null);`

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -692,6 +692,10 @@ else if (fetchSize != Integer.MAX_VALUE)
    * and handle host failover.
    */
   void execute(final RequestHandler.Callback callback, final Statement statement) {
+    if (this.isClosed()) {
+      callback.onException(
+          null, new IllegalStateException("Could not send request, session is closed"), 0, 0);
+    }
```
olim7t: OK now the `isInit == true` branch is not covered in `executeAsync` because it runs the handler directly instead of delegating to this method.

How about we move the check to the beginning of the `RequestHandler` constructor? That way it will be done in a single place, and we don't have to worry about the different execution paths in the various session classes (which quite frankly are a mess at this point 😒 ).

olim7t (changes requested):

driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
tolbertam: Yeah it looks like the change to call execute instead of new'ing up the `RequestHandler` is missing.   Agree, lets just move it to the constructor

GregBestland: I ended up implementing Andy's initial solution. Moving the check into the constructor caused other issues, namely how we would inform the caller of the constructor that there was an issue, and abort.

Everything should be routed through the execute method now, which should provide a single point to check. I also removed the superfluous check on the parent cluster.

tolbertam (approved):
Changes look good to me :+1:

olim7t (approved):
LGTM 👍 
The history is a bit messy, so if you want to avoid a complicated rebase the simplest thing is to squash everything in a single commit. "Improve cluster init and shutdown" or something similar, and then you list the 3 tickets in the commit message's body.

tolbertam:
Spoke with Greg and he was cool with me merging.  Rebased on 3.x (only conflict was changelog) and will squash and merge once travis gives the thumbs up.

PR #1075, opened 2018-08-08
JAVA-1935: Add null check in QueryConsistencyException.getHost

tolbertam:
For [JAVA-1935](https://datastax-oss.atlassian.net/browse/JAVA-1935)


adutra:
I commented on the Jira ticket: I think it would be wiser to amend all subtypes of `CoordinatorException`, not only this one, wdyt?

tolbertam:
Good catch, I didn't make that connection.  Will update all the applicable classes.

adutra (approved):

PR #1077, opened 2018-08-10
 JAVA-1939: Exclude logback-test.xml files from test jars

adutra:
Note: the 3 commits in this PR should not be squashed together.


tolbertam (approved):
:100: seems to work great!  I was worried that excluding the file from the jar would not pull in the configuration when running tests, but it seems to still work, i.e. `mvn clean install -DdriverLevel=DEBUG` still logs everything at DEBUG.

PR #1078, opened 2018-08-10
Fix ArrayUtilsTest.should_shuffle_head()

adutra:


tolbertam (approved):
:+1:

core/src/main/java/com/datastax/oss/driver/internal/core/util/ArrayUtils.java
```diff
@@ -50,27 +52,44 @@
...
    *
+   * @param elements the array to shuffle.
+   * @param n the number of elements to shuffle; must be {@code <= elements.length}.
    * @see <a
    *     href="https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_modern_algorithm">Modern
    *     Fisher-Yates shuffle</a>
    */
-  public static <T> void shuffleHead(T[] elements, int n) {
+  public static <T> void shuffleHead(@NonNull T[] elements, int n) {
+    shuffleHead(elements, n, ThreadLocalRandom.current()::nextInt);
```
olim7t: There's probably a small overhead to translate the method reference on each call. I'm not sure if it's significant or not, but we can avoid it altogether by passing the `ThreadLocalRandom` itself instead of the method.
Then you can mock it in the test:
```java
    ThreadLocalRandom random = Mockito.mock(ThreadLocalRandom.class);
    Mockito.when(random.nextInt(Mockito.anyInt())).thenAnswer((invocation) -> {
      int i = invocation.getArgument(0);
      return i - 2;
    });
```

olim7t (changes requested):

olim7t (approved):

olim7t:
👍 
I squashed and updated your commit message.

PR #1079, opened 2018-08-10
JAVA-1940: Clean up test resources when CCM integration tests finish

emerkle826:


test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
```diff
@@ -327,6 +327,10 @@ private static File createTempStore(String storePath) {
     } catch (IOException e) {
       logger.warn("Failure to write keystore, SSL-enabled servers may fail to start.", e);
     }
+    // ensure the temp store files are deleted when the JVM exits
+    if (f != null) {
```
tolbertam: This could be in a finally block for the preceding try, although I don't expect runtime exceptions to come out of that block, so this is fine as is too.

tolbertam (approved):
:+1:

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
```diff
@@ -354,6 +358,10 @@ private Builder() {
       }
       // disable auto_snapshot by default to reduce disk usage when destroying schema.
       withCassandraConfiguration("auto_snapshot", "false");
+      // mark the ccm temp directories for deletion when the JVM exits
+      if (this.configDirectory != null) {
+        this.configDirectory.toFile().deleteOnExit();
```
adutra: Are we sure this directory will be empty? If not, I suggest doing the following instead:

```
Runtime.getRuntime()
    .addShutdownHook(
        new Thread(
            () -> {
              try (Stream<Path> walk = Files.walk(configDirectory)) {
                walk.sorted(Comparator.reverseOrder())
                    .map(Path::toFile)
                    .forEach(File::delete);
              } catch (IOException e) {
                logger.error("Could not walk " + configDirectory, e);
              }
            }));
```

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
adutra: While at it, I suggest rewriting the whole method:

```
  private static File createTempStore(String storePath) {
    File f = null;
    try (OutputStream os = new FileOutputStream(f = File.createTempFile("server", ".store"))) {
      f.deleteOnExit();
      Resources.copy(CcmBridge.class.getResource(storePath), os);
    } catch (IOException e) {
      logger.warn("Failure to write keystore, SSL-enabled servers may fail to start.", e);
    }
    return f;
  }
```
`Resources` is actually `com.datastax.oss.driver.shaded.guava.common.io.Resources`.

adutra (changes requested):

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
emerkle826: I'm pretty sure the directory will be empty. It seems that when the tests finish, they execute a `ccm remove` that cleans up everything in the directory, just not the directory itself. I did try to setup a shutdown hook to clean up the directory, but it seems to cause issues where the ccm process hangs at the completion of the test.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
```diff
@@ -314,18 +314,15 @@ public void close() {
...
-        int len;
-        while ((len = trustStoreIs.read(buffer)) != -1) {
-          trustStoreOs.write(buffer, 0, len);
-        }
-      }
+    try (OutputStream os = new FileOutputStream(f = File.createTempFile("server", ".store"))) {
+      Resources.copy(CcmBridge.class.getResource(storePath), os);
     } catch (IOException e) {
       logger.warn("Failure to write keystore, SSL-enabled servers may fail to start.", e);
+    } finally {
```
adutra: Imo this could be simplified to just `f.deleteOnExit();` placed right after the try block start, since `f` cannot be null at that point. But this way is OK too.

adutra (approved):

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
emerkle826: I think it is possible that `f` is null, but only if `createTempFile` throws an IOException or a SecurityException above. It is also possible the file is created, but an exception is thrown creating the OutputStream or doing the `Resources.copy`. Putting it in the finally and ensuring it's not null seemed to be the safest bet.

emerkle826: Actually, now that I look at it again, the file is either created or not during `createTempFile`. If that throws an Exception, the file isn't created and doesn't need `deleteOnExit()` called. If it is created, a FileOutputStream is created on top. That either throws a FileNotFoundException or a SecurityException if it fails. It would be very difficult to throw either of those if creating the temp file didn't already fail so...
TL;DR; going to do it the way you suggest. :smiley:

olim7t (approved):

PR #1080, opened 2018-08-13
JAVA-1942: Adapt type with CodecRegistry's associated AttachmentPoint

tolbertam:
For [JAVA-1942](https://datastax-oss.atlassian.net/browse/JAVA-1942).

Motivation:

To ensure tuples and udts are decoded using the codec registry and
protocol version associated with the current Session, adapt any
incoming TupleValue / UdtValue's type to the AttachmentPoint of
the session when creating their associated codecs.

Modifications:

Update CachingCodecRegistry constructor to accept AttachmentPoint
to be associated with this registry.  Update DefaultDriverContext
to pass itself as the AttachmentPoint.

Update CachingCodecRegistry.createCodec methods to adapt types
to the AttachmentPoint.

Update DataTypeIT to create an AttachmentPoint based on the expected
ProtocolVersion to be used by the Session.

Introduce SessionUtils.getMaxProtocolVersion, which replaces
CassandraResourceRule.getHighestProtocolVersion.

Result:

Session's protocol version and codec registry are ensured to be
factored in when deserializing UDTs and Tuples.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
@@ -17,6 +17,7 @@ package com.datastax.oss.driver.internal.core.type.codec.registry;
 
 import com.datastax.oss.driver.api.core.data.TupleValue;
 import com.datastax.oss.driver.api.core.data.UdtValue;
+import com.datastax.oss.driver.api.core.detach.AttachmentPoint;
 import com.datastax.oss.driver.api.core.type.CustomType;
 import com.datastax.oss.driver.api.core.type.DataType;
 import com.datastax.oss.driver.api.core.type.ListType;
@@ -29,6 +30,8 @@ import com.datastax.oss.driver.api.core.type.codec.TypeCodec;
 import com.datastax.oss.driver.api.core.type.codec.TypeCodecs;
 import com.datastax.oss.driver.api.core.type.codec.registry.CodecRegistry;
 import com.datastax.oss.driver.api.core.type.reflect.GenericType;
+import com.datastax.oss.driver.internal.core.type.DefaultTupleType;
+import com.datastax.oss.driver.internal.core.type.DefaultUserDefinedType;
 import com.datastax.oss.driver.shaded.guava.common.base.Preconditions;
 import com.datastax.oss.driver.shaded.guava.common.reflect.TypeToken;
 import com.datastax.oss.protocol.internal.util.IntMap;
@@ -65,13 +68,18 @@ public abstract class CachingCodecRegistry implements CodecRegistry {
   //   traversal is cheap).
 
   protected final String logPrefix;
+  private final AttachmentPoint attachmentPoint;
   private final TypeCodec<?>[] primitiveCodecs;
   private final TypeCodec<?>[] userCodecs;
   private final IntMap<TypeCodec> primitiveCodecsByCode;
 
   protected CachingCodecRegistry(
-      String logPrefix, TypeCodec<?>[] primitiveCodecs, TypeCodec<?>[] userCodecs) {
+      String logPrefix,
+      AttachmentPoint attachmentPoint,
+      TypeCodec<?>[] primitiveCodecs,
+      TypeCodec<?>[] userCodecs) {
     this.logPrefix = logPrefix;
+    this.attachmentPoint = attachmentPoint;
     this.primitiveCodecs = primitiveCodecs;
     this.userCodecs = userCodecs;
     this.primitiveCodecsByCode = sortByProtocolCode(primitiveCodecs);
@@ -327,16 +335,41 @@ public abstract class CachingCodecRegistry implements CodecRegistry {
         return TypeCodecs.mapOf(keyCodec, valueCodec);
       } else if (cqlType instanceof TupleType
           && TupleValue.class.isAssignableFrom(token.getRawType())) {
-        return TypeCodecs.tupleOf((TupleType) cqlType);
+        return createTupleCodec((TupleType) cqlType);
       } else if (cqlType instanceof UserDefinedType
           && UdtValue.class.isAssignableFrom(token.getRawType())) {
-        return TypeCodecs.udtOf((UserDefinedType) cqlType);
+        return createUdtCodec((UserDefinedType) cqlType);
       } else if (cqlType instanceof CustomType
           && ByteBuffer.class.isAssignableFrom(token.getRawType())) {
         return TypeCodecs.custom(cqlType);
       }
       throw new CodecNotFoundException(cqlType, javaType);
     }
+  }
+
+  protected TypeCodec<TupleValue> createTupleCodec(TupleType tupleType) {
+    // if the attachment point of the input tuple does not match the attachment point
+    // for this registry, adapt it.  This is important to ensure decoding of values uses
+    // the appropriate registry and protocol version.  It has no bearing on serialization,
+    // which will use the tuple type of the input tuple value.
+    if (tupleType.getAttachmentPoint() != attachmentPoint) {
+      tupleType = new DefaultTupleType(tupleType.getComponentTypes(), attachmentPoint);
+    }
+    return TypeCodecs.tupleOf(tupleType);
+  }
+
+  protected TypeCodec<UdtValue> createUdtCodec(UserDefinedType udtType) {
+    if (udtType.getAttachmentPoint() != attachmentPoint) {
+      udtType =
+          new DefaultUserDefinedType(
+              udtType.getKeyspace(),
+              udtType.getName(),
+              udtType.isFrozen(),
+              udtType.getFieldNames(),
+              udtType.getFieldTypes(),
+              attachmentPoint);
+    }
+    return TypeCodecs.udtOf(udtType);
   }
 
   // Try to create a codec when we haven't found it in the cache.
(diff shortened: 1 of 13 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -339,6 +347,31 @@ protected boolean matches(TypeCodec<?> codec, GenericType<?> javaType, boolean i
     }
   }
 
+  protected TypeCodec<TupleValue> createTupleCodec(TupleType tupleType) {
+    // if the attachment point of the input tuple does not match the attachment point
+    // for this registry, adapt it.  This is important to ensure decoding of values uses
+    // the appropriate registry and protocol version.  It has no bearing on serialization,
+    // which will use the tuple type of the input tuple value.
+    if (tupleType.getAttachmentPoint() != attachmentPoint) {
```
tolbertam, line 355: we want to use the same exact instance, so `!=` is used here, although we could consider using `equals` assuming the associated `AttachmentPoint` implements it.

adutra: Couldn't you use `tupleType.attach(attachmentPoint)`?

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -339,6 +347,31 @@ protected boolean matches(TypeCodec<?> codec, GenericType<?> javaType, boolean i
...
+      tupleType = new DefaultTupleType(tupleType.getComponentTypes(), attachmentPoint);
+    }
+    return TypeCodecs.tupleOf(tupleType);
+  }
+
+  protected TypeCodec<UdtValue> createUdtCodec(UserDefinedType udtType) {
+    if (udtType.getAttachmentPoint() != attachmentPoint) {
```
adutra, line 362: Same here, couldn't you use `udtType.attach(attachmentPoint)`?

adutra (commented):
I am a bit worried by the possible consequences here.

1. Passing an `AttachmentPoint` to `CachingCodecRegistry`: what if the instance is going to be shared among more than one session using different protocol versions, e.g. when using the default `CodecRegistry#DEFAULT`?
2. I wonder if the right thing to do in `DefaultTupleValue.equals()` and `DefaultUdtValue.equals()` wouldn't be to _not_ take the protocol version into account? After all, the protocol version is only helpful when serializing/deserializing, once the object is fully deserialized it shouldn't really matter with which exact version it was deserialized. Note btw that the `hashCode()` methods do not include the protocol version in the hash. (Besides, I noticed that `DefaultTupleType.equals()` and `DefaultUserDefinedType.equals()` do _not_ take the attachment point into account either, so it's a bit odd that their corresponding values do).

On a side note, I think I finally understood why the failed assertion seems to be comparing _exactly the same instances_:

```
Expecting:
 <"(0,9223372036854775807) (DefaultTupleValue@8012365a)">
to be equal to:
 <"(0,9223372036854775807) (DefaultTupleValue@8012365a)">
```

AssertJ creates these messages here:

```
  @Override
  public String unambiguousToStringOf(Object obj) {
    return obj == null ? null
        : String.format("%s (%s@%s)", toStringOf(obj), obj.getClass().getSimpleName(), toHexString(obj.hashCode()));
  }
```

So it's actually using the hash code generated by `DefaultTupleValue.hashCode()`, which indeed does not take into account the protocol version. I initially thought that these hash codes were identity hash codes (in which case it would be impossible to have a test failure here).

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
tolbertam: Oh, I didn't even realize that was an option at the time.  I just looked at the implementation though and it appears it changes the instance and its underlying component types, which is problematic, even if we create a new instance as we do here.  So we'd need to find a solution for that.

tolbertam: Actually when you initialize `DefaultTupleType` with an `AttachmentPoint` the underlying component type values *are not* updated, which can also be a problem (but it would also be a problem if they were updated in place instead of creating new copies).

tolbertam:
> Passing an AttachmentPoint to CachingCodecRegistry: what if the instance is going to be shared among more than one session using different protocol versions, e.g. when using the default CodecRegistry#DEFAULT?

`CodecRegistry#DEFAULT` does not supply an `AttachmentPoint` so it uses `AttachmentPoint.NONE`, so I think this is not a problem in that particular case.

> I wonder if the right thing to do in DefaultTupleValue.equals() and DefaultUdtValue.equals() wouldn't be to not take the protocol version into account? After all, the protocol version is only helpful when serializing/deserializing, once the object is fully deserialized it shouldn't really matter with which exact version it was deserialized. Note btw that the hashCode() methods do not include the protocol version in the hash. (Besides, I noticed that DefaultTupleType.equals() and DefaultUserDefinedType.equals() do not take the attachment point into account either, so it's a bit odd that their corresponding values do).

I think the protocol version has to be important because that may dictate how the values associated with the tuple/udt are serialized.  If you have two equivalent Tuples, but different protocol versions how they serialize the values can be different,  I suppose you could argue the same thing about the `CodecRegistry`, so it should consider that too probably.

> (Besides, I noticed that DefaultTupleType.equals() and DefaultUserDefinedType.equals() do not take the attachment point into account either, so it's a bit odd that their corresponding values do).

Yes it probably should.  I think if we adjusted this maybe things would work as we could get a different `TupleCodec` per `TupleType` which I think is really what we want.  Instead of relying `AttachmentPoint.equals` (which may not be implemented), we can factor in the `AttachmentPoint`'s `getCodecRegistry` and `getProtooclVersion`.   If we do this, we don't need to pass in `AttachmentPoint` to the `CachingCodecRegistry` constructor.  I'll explore that.

tolbertam:
Hrmm I'm not quite happy with the "factor attachment point in `DataType.equals`" solution either.

I think one thing that bothers me is that when you are encoding/decoding from a Codec, you already have the context of the `CodecRegistry` (codec belongs to registry) & `ProtocolVersion` (a parameter to encode/decode).   The `AttachmentPoint` should be irrelevant in this context.

adutra:
> I think the protocol version has to be important because that may dictate how the values associated with the tuple/udt are serialized. If you have two equivalent Tuples, but different protocol versions how they serialize the values can be different.

Agreed, but isn't the way they serialize themselves outside of the scope of Java equality? If we were talking about protocol-level equality, then yes, but I think that all we want here is to know if two Java representations of some tuple are equal or not, and for that I think comparing their components is sufficient.

olim7t:
This is indeed a hairy corner case. UDT and tuple codecs keep a reference to the data type they were created with, and will use it to create the new values when they decode. If later you retrieve the existing codec via an equal type but with a different attachement point, the codec will still use the initial type, so the decoded values will always have the initial attachement point.

> I noticed that DefaultTupleType.equals() and DefaultUserDefinedType.equals() do not take the attachment point into account

That's one solution I was considering. If they did, each differently attached type would have its own codec, which seems to solve the issue. The problem is that those fields are mutable, and the data type is part of the cache key; mutable keys are a very bad idea. Maybe we could make them immutable, but that complicates the `Detachable` interface and its other implementations (`Row`, `ColumnDefinition`...)

Taking a step back, I wonder how much of a problem that will really be in practice. There are two aspects to the attachement point: the codec registry and the protocol version.

The codec registry matters for custom mappings. My feeling is that if you have tuples or UDTs that require such mappings, and you create data types manually, it should be pretty obvious that you're going to have to attach those types to something that has the adequate registry. If you don't, things will blow up in your face pretty quickly, long before you run into arcane codec creation corner cases; basically any time you call `get(name, Foo.class)`.

As for the protocol version, the most probable use case is indeed the one mentioned by Alexandre: transferring data across two sessions connected to different clusters using different protocol versions. In that case, my advice would be: don't share anything. Use a distinct set of types and a different codec registry for each session; when you need to copy over a UDT/tuple value, create a new value with the other type and do a field-by-field copy. Again, this is required not only for this bug, but for every getter/setter use (unless you prefer reattaching the data type to each session every time you switch -- assuming that your code is not concurrent!)

All things considered, I think those use cases will be very rare, and the concerns above will have to be addressed anyway even if we fix this bug. So my suggestion is to leave things as they are, and document the gotchas in the manual.

adutra:
I still think that taking the protocol version _itself_ into account to determine object equality is conceptually wrong. I do agree that the protocol version is required to decode inner components _while computing equality_ (we can't do otherwise), but imo the version itself should be irrelevant. Put differently, if two different protocol versions produce exactly the same java representations of some CQL value, to me these representations should be considered equal to each other. And honestly I don't see any drawback in doing this change, but one nice positive outcome: the `hashCode()` method would not produce identical hashes for non-equal objects anymore.

olim7t:
> I still think that taking the protocol version itself into account to determine object equality is conceptually wrong

That's completely orthogonal to this ticket. But I agree it's worth considering.

tolbertam:
I think I agree for the most part with your comments.  It is clear to me now that my idea to adapt the AttachmentPoint of a `TupleType`/`UserDefinedType` has its own set of flaws.  I think in general case this will not be an issue today, and I don't think we can really 'fix' this without some major redesign, which I don't think is currently justified.

I also agree that we it may be nice to not consider the protocol version in `TupleValue.equals` implementation :+1: .

Closing this PR as I think we can consider it invalidated by our discussion.

olim7t:
👍 , I will create tickets for adding documentation, and the tuple/UDT value equality.

PR #1082, opened 2018-08-15
JAVA-1810: Note which setters are not propagated to PreparedStatement

tolbertam:
Just a minor doc change for [JAVA-1810](https://datastax-oss.atlassian.net/browse/JAVA-1810).

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/PreparedStatement.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/PreparedStatement.java
@@ -125,7 +125,7 @@ public interface PreparedStatement {
    * Sets a default consistency level for all bound statements created from this prepared statement.
    *
    * <p>If no consistency level is set through this method, the bound statement created from this
-   * object will use the default consistency level (ONE).
+   * object will use the default consistency level (LOCAL_ONE).
    *
    * <p>Changing the default consistency level is not retroactive, it only applies to BoundStatement
    * created after the change.
--- a/driver-core/src/main/java/com/datastax/driver/core/Session.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Session.java
@@ -249,9 +249,14 @@ public interface Session extends Closeable {
   /**
    * Prepares the provided query.
    *
-   * <p>This method behaves like {@link #prepare(String)}, but note that the resulting {@code
-   * PreparedStatement} will inherit the query properties set on {@code statement}. Concretely, this
-   * means that in the following code:
+   * <p>This method behaves like {@link #prepare(String)}, but the resulting {@code
+   * PreparedStatement} will inherit some of the properties set on {@code statement}: {@linkplain
+   * Statement#getRoutingKey(ProtocolVersion, CodecRegistry) routing key}, {@linkplain
+   * Statement#getConsistencyLevel() consistency level}, {@linkplain
+   * Statement#getSerialConsistencyLevel() serial consistency level}, {@linkplain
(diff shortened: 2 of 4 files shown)


driver-core/src/main/java/com/datastax/driver/core/Statement.java
```diff
@@ -224,6 +224,9 @@ public abstract ByteBuffer getRoutingKey(
    * method is thus only useful in case you want to punctually override the default policy for this
    * request.
    *
+   * <p>Note that unlike other configuration, when this statement is prepared {@link
+   * BoundStatement}s created off of {@link PreparedStatement} do not inherit this configuration.
+   *
```
olim7t: It looks like the retry policy is actually propagated: https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/AbstractSession.java#L131

olim7t (changes requested):

driver-core/src/main/java/com/datastax/driver/core/PreparedStatement.java
```diff
@@ -125,7 +125,7 @@
    * Sets a default consistency level for all bound statements created from this prepared statement.
    *
    * <p>If no consistency level is set through this method, the bound statement created from this
-   * object will use the default consistency level (ONE).
+   * object will use the default consistency level (LOCAL_ONE).
```
olim7t, line 128: Not related to this PR, but it was wrong. We'll keep the commit separate.

driver-core/src/main/java/com/datastax/driver/core/Session.java
```diff
@@ -249,9 +245,14 @@
   /**
    * Prepares the provided query.
    *
-   * <p>This method behaves like {@link #prepare(String)}, but note that the resulting {@code
-   * PreparedStatement} will inherit the query properties set on {@code statement}. Concretely, this
-   * means that in the following code:
+   * <p>This method behaves like {@link #prepare(String)}, but the resulting {@code
```
tolbertam, line 252: Nice :+1:

driver-core/src/main/java/com/datastax/driver/core/Session.java
```diff
@@ -15,11 +15,7 @@
  */
 package com.datastax.driver.core;
 
-import com.datastax.driver.core.exceptions.AuthenticationException;
-import com.datastax.driver.core.exceptions.NoHostAvailableException;
-import com.datastax.driver.core.exceptions.QueryExecutionException;
-import com.datastax.driver.core.exceptions.QueryValidationException;
-import com.datastax.driver.core.exceptions.UnsupportedFeatureException;
+import com.datastax.driver.core.exceptions.*;
```
tolbertam: Small nit, google formatter doesn't catch this, but we should not be using wildcard imports.

driver-core/src/main/java/com/datastax/driver/core/Statement.java
tolbertam: Ah, I missed that somehow, good catch!

driver-core/src/main/java/com/datastax/driver/core/PreparedStatement.java
tolbertam: Good catch

olim7t (approved):

tolbertam:
Thanks for reviewing and improving the docs on the prepare method, merging :+1:

PR #1083, opened 2018-08-16
JAVA-1916: Base TimestampCodec.parse on java.util.Date

adutra:


core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/TimestampCodec.java
```diff
@@ -26,44 +24,164 @@
...
+ *   <li>{@code yyyy-MM-dd}
+ *   <li>{@code yyyy-MM-ddX}
+ *   <li>{@code yyyy-MM-ddXX}
+ *   <li>{@code yyyy-MM-ddXXX}
+ *   <li>{@code yyyy-MM-dd z}
+ * </ol>
+ *
+ * When parsing, timestamp literals that do not include any time zone information will be
+ * interpreted using the system's {@linkplain TimeZone#getDefault() default time zone}. This is
+ * intended to mimic Apache Cassandra(R)'s own parsing behavior.
```
olim7t: Maybe we should mention the server-side class for easy reference, `TimestampSerializer` if I'm not mistaken.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/TimestampCodec.java
```diff
@@ -26,44 +24,164 @@
...
+ *
+ * <p>When formatting, the pattern used is always {@code yyyy-MM-dd'T'HH:mm:ss.SSSXXX} and the time
+ * zone is always UTC.
```
olim7t: Why not use the default time zone for formatting as well? In a way, that would be consistent with CQLSH.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/TimestampCodec.java
```diff
@@ -26,44 +24,164 @@
...
+        "yyyy-MM-dd z"
+      };
+
+  private static final FastThreadLocal<SimpleDateFormat> PARSER =
+      new FastThreadLocal<SimpleDateFormat>() {
+        @Override
+        protected SimpleDateFormat initialValue() {
+          SimpleDateFormat parser = new SimpleDateFormat();
+          parser.setLenient(false);
+          parser.setTimeZone(TimeZone.getDefault());
```
olim7t: How about making the time zone a constructor argument, with a default constructor that sets it to the default one? This is of no consequence to us and makes life easier if someone wants a codec that uses a different one.

olim7t (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/TimestampCodec.java
adutra: It's really a matter of taste, as long as the output is unambiguous. OK for the default time zone.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/TimestampCodec.java
adutra: Good idea, will do.

olim7t (approved):

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/TimestampCodecTest.java
```diff
@@ -15,16 +15,33 @@
...
 import org.junit.Test;
+import org.junit.runner.RunWith;
 
+@RunWith(DataProviderRunner.class)
 public class TimestampCodecTest extends CodecTestBase<Instant> {
 
   public TimestampCodecTest() {
-    this.codec = TypeCodecs.TIMESTAMP;
+    // force a given timezone for reproducible results in should_format
+    TimeZone.setDefault(TimeZone.getTimeZone(UTC));
```
olim7t: I had to add that to make the tests platform-independent.

olim7t: Mmm actually I spoke too fast, it still fails on the command line...

PR #1085, opened 2018-08-20
Include ServerError in downgrade criteria from peers_v2 to peers

tolbertam:
The logic from JAVA-1388 dictates that the driver downgrades from
system.peers_v2 to system.peers for node discovery if an
INVALID_QUERY error is received.  We should also consider SERVER_ERROR
as some versions inappropriately respond with errors of that type.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
@@ -708,9 +708,11 @@ class ControlConnection implements Connection.Owner {
 
             @Override
             public void onFailure(Throwable t) {
-              // downgrade to system.peers if we get an invalid query or server error as this
-              // indicates the peers_v2 table does not exist.
-              if (t instanceof InvalidQueryException || t instanceof ServerError) {
+              // downgrade to system.peers if we get an invalid query or server error with specific
+              // message as this indicates the peers_v2 table does not exist.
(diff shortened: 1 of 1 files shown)


driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
```diff
@@ -707,9 +708,9 @@ public void onSuccess(ResultSet result) {
 
             @Override
             public void onFailure(Throwable t) {
-              // downgrade to system.peers if we get an invalid query as this indicates
-              // the peers_v2 table does not exist.
-              if (t instanceof InvalidQueryException) {
+              // downgrade to system.peers if we get an invalid query or server error as this
+              // indicates the peers_v2 table does not exist.
+              if (t instanceof InvalidQueryException || t instanceof ServerError) {
```
olim7t: Didn't we decide to also match the error message in the case of `ServerError`?

tolbertam: My mistake, that's right.  I'll add that.

olim7t (approved):
LGTM, I think we should also mention in the comment that this is because DSE 6.0.0 to 6.0.2 does that.

tolbertam:
Sounds good, will amend the comment and also include it in the commit message, thanks for reviewing!

PR #1086, opened 2018-08-21
JAVA-1917: Add ability to set node on statement

GregBestland:
Add ability to setNode on statement, this buypasses the current query plan and creates a new one that contains only the targeted node.


changelog/README.md
```diff
@@ -9,6 +9,7 @@
 - [improvement] JAVA-1925: Rename context getters
 - [improvement] JAVA-1544: Check API compatibility with Revapi
 - [new feature] JAVA-1900: Add support for virtual tables
+- [new feature] JAVA-1917: Add ability to set node on statement
```
olim7t: Nit: could you add the ticket at the top of the changelog?

I changed this in 4.x because it feels more natural to put more recent tickets first, since versions are ordered that way. It's also how it's done in the Cassandra project.

Maybe we should adopt this convention for the next 3.x versions too for consistency.

core/src/main/java/com/datastax/oss/driver/api/core/cql/Statement.java
```diff
@@ -93,6 +94,12 @@
   @NonNull
   T setRoutingKeyspace(@Nullable CqlIdentifier newRoutingKeyspace);
 
+  @NonNull
+  T setNode(@Nullable Node node);
+
+  @Nullable
+  Node getNode();
```
olim7t: This is a pretty generic feature and it might apply to other types of requests than CQL, so I think `Request` would be a better place for this method (`setNode` can stay here, since `Request` only defines its properties, not how they are set in subtypes).
The only other implementation is `PrepareRequest`, it should never be targeted so `getNode` will return null.

Could you also add javadocs? I think [the ones from the 3.x PR](https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/Statement.java#L625-L638) can almost be copied as-is (the part about not propagating to prepared statements would be better in the javadocs of `CqlSession#prepare(SimpleStatement)`, which already detail all other fields).

core/src/main/java/com/datastax/oss/driver/api/core/cql/BatchStatementBuilder.java
```diff
@@ -146,6 +146,7 @@ public BatchStatement build() {
         pageSize,
         consistencyLevel,
         serialConsistencyLevel,
-        timeout);
+        timeout,
+        null);
```
olim7t: The last argument should be `node` (inherited from the parent builder) instead of `null`.

core/src/main/java/com/datastax/oss/driver/api/core/cql/StatementBuilder.java
```diff
@@ -199,6 +202,11 @@ public T withTimeout(@Nullable Duration timeout) {
     return self;
   }
 
+  public T withNode(@Nullable Node node) {
```
olim7t: Could you link to the setter's javadocs like other methods?

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
```diff
@@ -153,10 +154,16 @@ protected CqlRequestHandlerBase(
...
               : config.getProfile(profileName);
     }
-    this.queryPlan =
-        context
-            .getLoadBalancingPolicyWrapper()
-            .newQueryPlan(statement, executionProfile.getName(), session);
+    if (this.statement.getNode() != null) {
+      this.queryPlan = new ConcurrentLinkedQueue<Node>();
+      this.queryPlan.add(this.statement.getNode());
+
```
olim7t: Since JAVA-1883 there is a dedicated queue implementation for query plans, we can also use it here:
```java
this.queryPlan = new QueryPlan(this.statement.getNode());
```

core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultPreparedStatement.java
```diff
@@ -103,7 +103,6 @@ public DefaultPreparedStatement(
     this.consistencyLevelForBoundStatements = consistencyLevelForBoundStatements;
     this.serialConsistencyLevelForBoundStatements = serialConsistencyLevelForBoundStatements;
     this.areBoundStatementsTracing = areBoundStatementsTracing;
-
     this.codecRegistry = codecRegistry;
     this.protocolVersion = protocolVersion;
```
olim7t: Nit: could you avoid modifying unrelated lines? This adds noise when you do a git blame on the file later (admittedly not for deleted lines, but still).

integration-tests/src/test/java/com/datastax/oss/driver/api/core/loadbalancing/NodeTargetingTestIT.java
```diff
@@ -0,0 +1,115 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(ParallelizableTests.class)
+public class NodeTargetingTestIT {
```
olim7t: Nit: no need to repeat "Test" before "IT".

integration-tests/src/test/java/com/datastax/oss/driver/api/core/loadbalancing/NodeTargetingTestIT.java
```diff
@@ -0,0 +1,115 @@
...
+  @Before
+  public void clear() {
+    simulacron.cluster().clearLogs();
+    simulacron.cluster().clearPrimes(true);
+  }
+
+  @Test
+  public void should_use_host_on_statement() {
+    Collection<Node> nodeCol = sessionRule.session().getMetadata().getNodes().values();
+    List<Node> nodes = new ArrayList(nodeCol);
```
olim7t: `new ArrayList<>(nodeCol)` to avoid the warning (same in the two other methods).

integration-tests/src/test/java/com/datastax/oss/driver/api/core/loadbalancing/NodeTargetingTestIT.java
```diff
@@ -0,0 +1,115 @@
...
+      int hostIndex = i % 4 + 1;
+      Node node = nodes.get(hostIndex);
+
+      // given a statement with host explicitly set.
+      Statement statement = SimpleStatement.newInstance("select * system.local").setNode(node);
+
+      // when statement is executed
+      ResultSet result = sessionRule.session().execute(statement);
+
+      Node coordinator = result.getExecutionInfo().getCoordinator();
```
olim7t: unused variable

integration-tests/src/test/java/com/datastax/oss/driver/api/core/loadbalancing/NodeTargetingTestIT.java
```diff
@@ -0,0 +1,115 @@
...
+  @Rule public SessionRule<CqlSession> sessionRule = SessionRule.builder(simulacron).build();
+
+  @Before
+  public void clear() {
+    simulacron.cluster().clearLogs();
+    simulacron.cluster().clearPrimes(true);
+  }
+
+  @Test
+  public void should_use_host_on_statement() {
```
olim7t: Nit: 4.x uses the terminology "node" exclusively (I did that to avoid any ambiguity with `InetSocketAddress.getHostName`, with CASSANDRA-7544 there can now be multiple nodes on the same host).

integration-tests/src/test/java/com/datastax/oss/driver/api/core/loadbalancing/NodeTargetingTestIT.java
```diff
@@ -0,0 +1,115 @@
...
+    Collection<Node> nodeCol = sessionRule.session().getMetadata().getNodes().values();
+    List<Node> nodes = new ArrayList(nodeCol);
+    simulacron.cluster().node(3).prime(when(query).then(unavailable(ConsistencyLevel.ALL, 1, 0)));
+
+    // given a statement with a host configured to fail the given query.
+    Node node1 = nodes.get(3);
+    Statement statement = SimpleStatement.newInstance(query).setNode(node1);
+    // when statement is executed an error should be raised.
+    try {
+      ResultSet result = sessionRule.session().execute(statement);
```
olim7t: No need to store the result in a variable here.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/loadbalancing/NodeTargetingTestIT.java
```diff
@@ -0,0 +1,115 @@
...
+      assertThat(e.getErrors().size()).isEqualTo(1);
+      assertThat(e.getErrors().get(node1)).isInstanceOf(UnavailableException.class);
+    }
+  }
+
+  @Test
+  public void should_fail_if_host_is_not_connected() {
+    // given a statement with host explicitly set that for which we have no active pool.
+    simulacron.cluster().node(4).close();
+    ;
```
olim7t: Unnecessary semicolumn here and at line 113.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/loadbalancing/NodeTargetingTestIT.java
```diff
@@ -0,0 +1,115 @@
...
+    Collection<Node> nodeCol = sessionRule.session().getMetadata().getNodes().values();
+    List<Node> nodes = new ArrayList(nodeCol);
+    Node node4 = nodes.get(4);
+    Statement statement = SimpleStatement.newInstance("select * system.local").setNode(node4);
+    try {
+      // when statement is executed
+      sessionRule.session().execute(statement);
+      fail("Query should have failed");
+    } catch (AllNodesFailedException e) {
```
olim7t: More precisely, it's a `NoNodeAvailableException` (a subclass).

Which IMO is also more clear for the end user: "no node was available" and you targeted a node => the targeted node was not available.

olim7t (changes requested):
Looks good, see comments inline.

+ an additional one: I think `BoundStatementBuilder(BoundStatement)` should consider the template statement's node like the other builders.

olim7t (approved):

PR #1087, opened 2018-08-24
JAVA-1932: Send DRIVER_NAME and DRIVER_VERSION in Startup

emerkle826:
- Change SessionBuilder so that it can take a custom map of startup options to use when building a DriverContext.
- Change DefaultDriverContext to build a map of startup options that appends any custom options set on the SessionBuilder to the list of Internal options sent.
- Change ProtocolInitHandler to use the startup options from the DriverContext when constructing a Startup message. This requires https://github.com/datastax/native-protocol/pull/24 to be merged.

The end result here is for the DriverContext to allow for custom startup options to be added to the map of options we are already sending in a Startup message.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
@@ -178,6 +178,8 @@ public class DefaultDriverContext implements InternalDriverContext {
       new LazyReference<>("metricsFactory", this::buildMetricsFactory, cycleDetector);
   private final LazyReference<RequestThrottler> requestThrottlerRef =
       new LazyReference<>("requestThrottler", this::buildRequestThrottler, cycleDetector);
+  private final LazyReference<Map<String, String>> startupOptionsRef =
+      new LazyReference<>("startupOptions", this::buildStartupOptions, cycleDetector);
   private final LazyReference<NodeStateListener> nodeStateListenerRef;
   private final LazyReference<SchemaChangeListener> schemaChangeListenerRef;
   private final LazyReference<RequestTracker> requestTrackerRef;
@@ -228,6 +230,15 @@ public class DefaultDriverContext implements InternalDriverContext {
             "requestTracker", () -> buildRequestTracker(requestTrackerFromBuilder), cycleDetector);
     this.nodeFiltersFromBuilder = nodeFilters;
     this.classLoader = classLoader;
+  }
+
+  /**
+   * Builds a map of options to send in a Startup message.
+   *
+   * @see #getStartupOptions()
+   */
+  protected Map<String, String> buildStartupOptions() {
+    return new StartupOptionsBuilder(this).build();
   }
 
   protected Map<String, LoadBalancingPolicy> buildLoadBalancingPolicies() {
@@ -730,5 +741,11 @@ public class DefaultDriverContext implements InternalDriverContext {
   @Override
   public ProtocolVersion getProtocolVersion() {
     return getChannelFactory().getProtocolVersion();
+  }
+
+  @NonNull
+  @Override
+  public Map<String, String> getStartupOptions() {
+    return startupOptionsRef.get();
   }
 }
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/context/StartupOptionsBuilder.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/context/StartupOptionsBuilder.java
@@ -1,0 +1,80 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.internal.core.context;
+
+import com.datastax.oss.driver.api.core.session.Session;
+import com.datastax.oss.protocol.internal.request.Startup;
+import com.datastax.oss.protocol.internal.util.collection.NullAllowingImmutableMap;
+import java.util.Map;
+import net.jcip.annotations.Immutable;
+
+@Immutable
+public class StartupOptionsBuilder {
+
+  public static final String DRIVER_NAME_KEY = "DRIVER_NAME";
+  public static final String DRIVER_VERSION_KEY = "DRIVER_VERSION";
+
+  protected final InternalDriverContext context;
+
+  public StartupOptionsBuilder(InternalDriverContext context) {
+    this.context = context;
+  }
+
+  /**
+   * Builds a map of options to send in a Startup message.
+   *
+   * <p>The default set of options are built here and include {@link
+   * com.datastax.oss.protocol.internal.request.Startup#COMPRESSION_KEY} (if the context passed in
+   * has a compressor/algorithm set), and the driver's {@link #DRIVER_NAME_KEY} and {@link
+   * #DRIVER_VERSION_KEY}. The {@link com.datastax.oss.protocol.internal.request.Startup}
+   * constructor will add {@link
+   * com.datastax.oss.protocol.internal.request.Startup#CQL_VERSION_KEY}.
+   *
+   * @return Map of Startup Options.
+   */
+  public Map<String, String> build() {
+    NullAllowingImmutableMap.Builder<String, String> builder = NullAllowingImmutableMap.builder(3);
+    // add compression (if configured) and driver name and version
+    String compressionAlgorithm = context.getCompressor().algorithm();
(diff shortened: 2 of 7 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -340,10 +342,17 @@ protected DriverContext buildContext(
...
 
   private static <T> T buildIfNull(T value, Supplier<T> builder) {
     return (value == null) ? builder.get() : value;
   }
+
+  @NonNull
+  public SelfT withStartupOptions(@Nullable Map<String, String> options) {
+    this.startupOptions = options;
+    return self;
+  }
```
olim7t: Maybe we should call this `with[Additional|Extra|Custom]StartupOptions` to emphasize the fact that CQL version and compression are not required (in fact for compression it is probably even recommended to not mess with it manually, and instead let the driver inspect the configured compressor).

This method also needs javadocs to explain all this.

Nit: could we also move it after `withClassLoader`? I know that ultimately method order does not matter, but it's convenient to have all the boilerplate `withXxx` together, and then the more interesting methods.

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
```diff
@@ -228,6 +231,20 @@ public DefaultDriverContext(
...
+
+  /**
+   * Builds a map of options to use when sending a Startup message. The options argument passed in
+   * will append to, or overwrite, the Internal default options sent by the driver.
+   */
+  protected Map<String, String> buildStartupOptions(Map<String, String> options) {
+    NullAllowingImmutableMap.Builder<String, String> builder = NullAllowingImmutableMap.builder();
+    builder.putAll(new InternalStartupOptions(this).getOptions());
+    if (options != null) {
+      builder.putAll(options);
```
olim7t: `NAIM.builder()` does not support duplicate entries. Won't that be a problem if the client passes CQL_VERSION or COMPRESSION in their map?
We should either prevent that, or check the map to allow the user to override those fields (although the use case is dubious).

core/src/main/java/com/datastax/oss/driver/internal/core/context/InternalStartupOptions.java
```diff
@@ -0,0 +1,70 @@
...
+
+import com.datastax.oss.driver.api.core.MavenCoordinates;
+import com.datastax.oss.driver.internal.core.DefaultMavenCoordinates;
+import com.datastax.oss.protocol.internal.request.Startup;
+import com.datastax.oss.protocol.internal.util.collection.NullAllowingImmutableMap;
+import java.util.Map;
+import net.jcip.annotations.Immutable;
+
+@Immutable
+public class InternalStartupOptions {
```
olim7t: 👍 to use a separate utility class, this makes tests easier. But:

1. I find the name a bit misleading, because in the rest of the code "startup options" refers to a `Map<String, String>`.  This shows in `DefaultDriverContext`, where we have a `buildStartupOptions` method that constructs an `InternalStartupOptions`, but eventually returns a map.
2. currently the caller still has to merge the additional options manually, why not make this part of the utility class as well?

I'm thinking of turning it into a builder:
```java
Map<String, String> options =
    new StartupOptionsBuilder(context)
        .withAdditionalOptions(additionalOptions)
        .build();
```
`build()` would contain the logic of the current `buildOptions()`, while also avoiding duplicates if they were already provided in `additionalOptions`.

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
```diff
@@ -228,6 +231,20 @@ public DefaultDriverContext(
...
     this.nodeFiltersFromBuilder = nodeFilters;
     this.classLoader = classLoader;
+    this.startupOptions = buildStartupOptions(extraStartupOptions);
+  }
+
+  /**
+   * Builds a map of options to use when sending a Startup message. The options argument passed in
+   * will append to, or overwrite, the Internal default options sent by the driver.
+   */
+  protected Map<String, String> buildStartupOptions(Map<String, String> options) {
```
olim7t: 👍 to make this protected.

core/src/test/java/com/datastax/oss/driver/internal/core/channel/ProtocolInitHandlerTest.java
```diff
@@ -136,34 +129,6 @@ public void should_initialize() {
...
-    Startup startup = (Startup) requestFrame.message;
-
-    // STARTUP message should request compression
-    assertThat(startup.options).containsEntry("COMPRESSION", "lz4");
-
-    writeInboundFrame(buildInboundFrame(requestFrame, new Ready()));
-    writeInboundFrame(readOutboundFrame(), TestResponses.clusterNameResponse("someClusterName"));
-    assertThat(connectFuture).isSuccess();
-  }
-
```
olim7t, line 166: 👍 that makes sense, the only thing this was testing was the options map, and since it's now built outside of the handler, covering it here wouldn't be very useful.

core/src/test/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContextTest.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.function.Predicate;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+public class DefaultDriverContextTest {
```
olim7t: This should be called `InternalStartupOptionsTest` (or the corresponding test name if we rename that class).

core/src/test/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContextTest.java
```diff
@@ -0,0 +1,119 @@
...
+        .containsEntry(InternalStartupOptions.DRIVER_NAME_KEY, driverProperties.getName());
+    assertThat(startup.getOptions()).containsKey(InternalStartupOptions.DRIVER_VERSION_KEY);
+    Version version =
+        Version.parse(startup.getOptions().get(InternalStartupOptions.DRIVER_VERSION_KEY));
+    // ensure it's a 4.x version, even if a pre-release
+    assertThat(version).isEqualByComparingTo(driverProperties.getVersion());
+  }
+
+  @Test
+  public void test_default_startup_options() {
```
olim7t: Nit: could you use the `should_` convention for consistency with the rest of the codebase? For example:
`should_build_minimal_startup_options`
`should_build_startup_options_with_compression`
`should_build_startup_options_with_custom_options`

And I would also cover the cases where the custom options contain one of the default ones or the compression.

olim7t (changes requested):

core/src/test/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContextTest.java
emerkle826: Yes, I forgot about this from my last PR. I need to break my old habit of test method names.

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
```diff
@@ -731,4 +744,10 @@ public CodecRegistry getCodecRegistry() {
   public ProtocolVersion getProtocolVersion() {
     return getChannelFactory().getProtocolVersion();
   }
+
+  @NonNull
+  @Override
+  public Map<String, String> getStartupOptions() {
+    return startupOptionsBuilderRef.get().build();
```
olim7t: The options are now rebuilt on every call. Did you want to leave the door open in case we ever need an option that would change for every new connection?

core/src/main/java/com/datastax/oss/driver/internal/core/context/StartupOptionsBuilder.java
```diff
@@ -0,0 +1,93 @@
...
+   * Builds a map of options to send in a Startup message.
+   *
+   * <p>The default set of options are built here and include {@link
+   * com.datastax.oss.protocol.internal.request.Startup#COMPRESSION_KEY} (if the context passed in
+   * has a compressor/algorithm set), and the driver's {@link #DRIVER_NAME_KEY} and {@link
+   * #DRIVER_VERSION_KEY}. The {@link com.datastax.oss.protocol.internal.request.Startup}
+   * constructor will add {@link
+   * com.datastax.oss.protocol.internal.request.Startup#CQL_VERSION_KEY}.
+   *
+   * <p>Additional options can be set via {@link #withAdditionalOptions(java.util.Map)}.
```
olim7t: This is now outdated.
withAdditionalOptions is referenced in two other comments.

olim7t (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
emerkle826: No, that wasn't the intent. I imagine the options won't change over the course of normal operation.

core/src/main/java/com/datastax/oss/driver/internal/core/context/StartupOptionsBuilder.java
emerkle826: Good catch!

PR #1088, opened 2018-08-30
Add LGTM code quality badges

xcorail:
Hi there!

I thought you might be interested in adding these [code quality badges](https://lgtm.com/projects/g/datastax/java-driver/ci/#badges) to your project. They will indicate a very high code quality to potential users and contributors.
You can also check the alerts discovered by [LGTM](https://lgtm.com/projects/g/datastax/java-driver).

N.B.: I am on the team behind LGTM.com, I'd appreciate your feedback on this initiative, whether you're interested or not, if you find time to drop me a line. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # Datastax Java Driver for Apache Cassandra®
 
 [![Build Status](https://travis-ci.org/datastax/java-driver.svg?branch=3.x)](https://travis-ci.org/datastax/java-driver)
+[![Code Quality: Java](https://img.shields.io/lgtm/grade/java/g/datastax/java-driver.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/datastax/java-driver/context:java)
+[![Total Alerts](https://img.shields.io/lgtm/alerts/g/datastax/java-driver.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/datastax/java-driver/alerts)
 
 *If you're reading this on github.com, please note that this is the readme
 for the development version and that some features described here might


adutra:
Hi @xcorail, thank you for your PR.

Your project looks promising but we don't want to commit too strongly to any static analysis tool, because these oftentimes report many false positives. Besides, we already have ErrorProne as part of our build and it does a pretty good job of catching real issues.

xcorail:
Hi @adutra 

Thanks for the feedback.

Fair enough. In this case, the [LGTM alerts](https://lgtm.com/projects/g/datastax/java-driver/alerts/?mode=tree) will anyway always there for you, daily refreshed, so that you can look at them occasionally. 

**If it's not asking too much of your time**, you seem to think that the reported alerts are false positives, would you mind pointing me to those FPs, so that we can fix them? As you said code analyzers have a bad reputation with FPs, but we make a great deal about having a low FP rate, and a technology that allows us to fix them in a few days.  

Cheers

adutra:
@xcorail sure, for example all current warnings about potential int overflows, such as [this one](https://lgtm.com/projects/g/datastax/java-driver/snapshot/8c362418947e7ba47348a8aadc6c392d1aa0dd5e/files/driver-core/src/main/java/com/datastax/driver/core/HostConnectionPool.java?sort=name&dir=ASC&mode=heatmap#x7223c197987a8dda:1), are false positives given the context; such an overflow would only happen if someone is crazy enough to specify `Integer.MAX_VALUE` seconds for a connection timeout :-) .

xcorail:
I am surprised. I would tend to consider this one as a TP **in general**, because "someone crazy enough" can happen. As a matter of fact, the reason for these alerts are for these cases, of someone _malicious_ enough, who just wants to mess up. So I don't think we'll make a fix on this one. 

But I guess from your comment that **in your context**, this data is not coming from a user input or has been sanitized. In this case, I think it's fair that we classify it as a Warning, and entitle it as **Potential** overflow. And leave the decision to the users to exclude these alerts if they think they are useless in their context.

Thanks a lot!

PR #1090, opened 2018-09-11
JAVA-1962: Use the parent POM name for Driver name in properties

emerkle826:


core/src/main/resources/com/datastax/oss/driver/Driver.properties
```diff
@@ -20,4 +20,5 @@
 driver.groupId=${project.groupId}
 driver.artifactId=${project.artifactId}
 driver.version=${project.version}
-driver.name=DataStax Java driver for Apache Cassandra(R)
+# Use the parent POM for the name
+driver.name=${project.parent.name}
```
adutra, line 24: The reason why I removed the property and replaced it with a plain text is that this property does not get resolved. If you unpack the final jar and check the file contents you will see the following:

```
driver.groupId=com.datastax.oss
driver.artifactId=java-driver-core
driver.version=4.0.0-beta2-SNAPSHOT
# Use the parent POM for the name
driver.name=${project.parent.name}
```

Actually none of the properties declared in the parent project get correctly resolved here. It seems that this is caused by the special packaging type `bundle` as if I switch to `jar`, the property gets resolved just fine.

We _could_ simply switch to packaging `jar` but I think that the best thing to do would be to customize the driver name and store that name in a property in the current project. Since `project.name` is already taken and not really suitable, we could just define some other property like `driver.name`:

```
<properties>
  <driver.name>DataStax Java driver for Apache Cassandra(R)</driver.name>
</properties>
```

As a bonus, we could then customize the driver name for the shaded driver by simply changing the property to the following in the shaded pom:

```
<properties>
  <driver.name>DataStax Java driver for Apache Cassandra(R) with shaded dependencies</driver.name>
</properties>
```

But for this to work the following is also required:

1. Create another `Driver.properties` file under `core-shaded/src/main/resources/com/datastax/oss/driver/Driver.properties`, with the same contents;
2. Place the following under the `build` section of the shaded pom:
```
<resources>
  <resource>
    <directory>src/main/resources</directory>
    <filtering>true</filtering>
  </resource>
</resources>
```
3. Exclude the `Driver.properties` file from the non-shaded module by adding the following to the configuration of the dependency plugin in the shaded pom:
```
<excludes>
  META-INF/maven/com.datastax.oss/java-driver-core/**,
  META-INF/maven/io.netty/**,
  com/datastax/oss/driver/Driver.properties
</excludes>
```

adutra (changes requested):

core/src/main/resources/com/datastax/oss/driver/Driver.properties
emerkle826: Thanks for the details @adutra. I did see that you changed the property to a fixed value, but wasn't sure why. This explains it nicely.

emerkle826:
Based on the comments and discussion, closing this and leaving it as is.

PR #1091, opened 2018-09-13
JAVA-1946: Ignore protocol version in equals comparison for UdtValue/TupleValue

emerkle826:


core/src/main/java/com/datastax/oss/driver/internal/core/data/DefaultUdtValue.java
```diff
@@ -144,11 +140,11 @@ public boolean equals(Object o) {
...
       Object thisValue =
-          that.codecRegistry()
+          this.codecRegistry()
               .codecFor(innerThisType)
               .decode(this.getBytesUnsafe(i), this.protocolVersion());
       Object thatValue =
-          this.codecRegistry()
+          that.codecRegistry()
               .codecFor(innerThatType)
               .decode(that.getBytesUnsafe(i), that.protocolVersion());
```
olim7t, line 149: Yes, weird that we missed that in the last review.

olim7t (approved):

PR #1094, opened 2018-09-14
Allow for raw yaml to appended to config

GregBestland:
Ran into some test cases where It was necessary to insert nested yaml properties. 
This facilitates that.

### DIFF
--- a/test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
+++ b/test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
@@ -62,6 +62,7 @@ public class CcmBridge implements AutoCloseable {
 
   private final Map<String, Object> cassandraConfiguration;
   private final Map<String, Object> dseConfiguration;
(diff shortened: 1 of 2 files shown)


test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
```diff
@@ -259,9 +265,24 @@ synchronized void execute(String... args) {
...
+    for (String arg : args) {
+      cli.addArgument(arg, false);
+    }
+    cli.addArgument("--config-dir=" + configDirectory.toFile().getAbsolutePath());
+
+    execute(cli);
+  }
+
+  private void execute(CommandLine cli) {
+    logger.debug("Executing: " + cli);
```
olim7t, line 285: The log can no longer be copy-pasted to the command line. If that's a problem, we can revisit later.

olim7t (approved):
👍 
I extracted `execute(CommandLine)` because it was repeated between the two execute methods.

PR #1095, opened 2018-09-17
JAVA-1970: Expose a counter for in-flight speculative executions

adutra:


core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
```diff
@@ -755,13 +764,24 @@ public void cancel() {
...
       context
           .getRequestTracker()
           .onNodeError(statement, error, latencyNanos, executionProfile, node);
     }
 
+    private void maybeUpdateSpeculativeExecutionsInFlightCounter(int amount) {
+      if (this.execution > 0) {
+        ((DefaultNode) node)
+            .getMetricUpdater()
+            .incrementCounter(
```
adutra, line 778: Out of curiosity: what is the reason why we pass the profile name to all methods related to metrics? It's not used in the default implementation.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
```diff
@@ -755,13 +764,24 @@ public void cancel() {
...
-    private void trackNodeError(Node node, Throwable error) {
+    private void trackNodeError(Throwable error) {
       long latencyNanos = System.nanoTime() - this.start;
       context
           .getRequestTracker()
           .onNodeError(statement, error, latencyNanos, executionProfile, node);
     }
 
+    private void maybeUpdateSpeculativeExecutionsInFlightCounter(int amount) {
+      if (this.execution > 0) {
```
tomekl007, line 775: why we need that conditional logic here? How can be a situation when we don't have execution > 0 but the execution finished executing: `maybeUpdateSpeculativeExecutionsInFlightCounter`?

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
olim7t: I thought about splitting the metrics by execution profile at some point. I decided not to do it, but didn't want to completely close the door to the idea.

olim7t (commented):
I'm not convinced of the usefulness of this new metric.

`speculative-executions.triggered` (the old one) counts the speculative executions triggered by a slow response from this node. As an operator, if I see it going up, it's a sign that this particular node is too slow for the configured threshold, which would prompt me to further investigate performance and/or adjust the threshold.

`speculative-executions.in-flight`, on the other hand, tells me how many speculative executions _triggered by other nodes_ are running on this node. How is that useful? If I see a high number, what does that tell me about this node? Nothing really, just that the other nodes are slower, which I can already see with the other metric.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
adutra: If `execution == 0` then this is the first (i.e., non-speculative execution of the request); for `execution > 0`, we are sure that this is a speculative execution.

adutra:
Indeed there is no internal need for this metric anymore atm. Closing.

PR #1096, opened 2018-09-17
JAVA-1971: Use System.nanoTime() to track node uptimes

adutra:


olim7t:
I'm not opposed to the idea and the changes are straightforward, but I'm wondering if the change is warranted.

Our main goal with this counter is to provide a way to detect that a node is "newly UP", and warmup times seem to be in the order of minutes, so the extra granularity of `nanoTime` won't probably make much of a difference.

Also, based on this comment in the `nanoTime` javadocs:
> This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time. The value returned represents nanoseconds since some fixed but arbitrary origin time (perhaps in the future, so values may be negative)

I think it *would* be nice to have a notion of wall-clock time. Even though that's not the primary intended use, I could see someone displaying `getUpSinceMillis` in some UI: "up since last Monday 12:34". Granted it's still trivial to compute the elapsed time and apply it to the current date, but people might overlook that.

adutra:
Yes I agree with you, it is nice to expose a wall-clock time. However my requirement is for elapsed time, but there are other ways to measure the time elapsed since a node came up. I think we can safely disregard this PR.

olim7t:
OK, closing then.

PR #1097, opened 2018-09-18
JAVA-1949: Improve error message when contact points are wrong

emerkle826:


core/src/main/java/com/datastax/oss/driver/internal/core/channel/ProtocolInitHandler.java
```diff
@@ -294,9 +294,15 @@ void onResponse(Message response) {
 
     @Override
     void fail(String message, Throwable cause) {
-      Throwable finalException =
-          (message == null) ? cause : new ConnectionInitException(message, cause);
-      setConnectFailure(finalException);
+      if (this.step == Step.STARTUP) {
```
emerkle826: Overwriting the message here, specifically during only the STARTUP step.

olim7t:
So if we change things in `ProtocolInitHandler`, it changes the node-level message.
Before:
```
Exception in thread "main" com.datastax.oss.driver.api.core.AllNodesFailedException:
  All 1 node tried for the query failed (showing first 1, use getErrors() for more:
    /127.7.1.1:9042: com.datastax.oss.driver.api.core.connection.ConnectionInitException: [s0|control|connecting...] init query STARTUP: error writing )
```

After:
```
Exception in thread "main" com.datastax.oss.driver.api.core.AllNodesFailedException:
  All 1 node tried for the query failed (showing first 1, use getErrors() for more:
    /127.7.1.1:9042: com.datastax.oss.driver.api.core.connection.ConnectionInitException: Connection refused, possible wrong contact points)
```

@adutra can chime in (since he opened the ticket), but I think it would be better to change the top-level message:
```
Exception in thread "main" com.datastax.oss.driver.api.core.AllNodesFailedException:
  Could not reach any contact point, make sure you've provided valid addresses (showing first 1, use getErrors() for more:
    /127.7.1.1:9042: com.datastax.oss.driver.api.core.connection.ConnectionInitException: [s0|control|connecting...] init query STARTUP: error writing )
```
(also, changing the init handler affects all further reconnection attempts, including for nodes that were not contact points)

And the best place to do that would be when we process a failure in [ControlConnection.init](https://github.com/datastax/java-driver/blob/f394f748aa72daab9cdc0141d91c5f7ced7e1ee6/core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java#L279) (this is the deepest point where we can still distinguish the initial connection attempt from a reconnection). We can test if the error is an ANFE, and instead of just completing the init future with it, rewrap it to change the message (there's no way to do that yet but we can add a new constructor to the exception).

olim7t:
As agreed with Erik on a separate channel, I went ahead and implemented my suggestion.

emerkle826 (commented):
This looks better. :+1:

olim7t (approved):

PR #1098, opened 2018-09-19
Fix tests for DefaultLoadBalancingPolicy events

adutra:
Motivation:

Some of the tests in DefaultLoadBalancingPolicyEventsTest weren't
testing what they were supposed to test. Some mocked dependencies were
being set up *after* the creation of the policy instance and thus had
no effect on its behavior.
This anomaly was being masked by the use of the
MockitoJUnitRunner.Silent runner; the regular Mockito runner would have
raised an error.

Modifications:

- Create the policy instance only after all mocks are properly
  configured.
- Use the regular MockitoJUnitRunner runner.

Result:

All mocked dependencies are now correctly taken into account.

### DIFF
--- a/core/src/test/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicyEventsTest.java
+++ b/core/src/test/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicyEventsTest.java
@@ -29,7 +29,7 @@ import org.junit.runner.RunWith;
 import org.mockito.Mockito;
 import org.mockito.junit.MockitoJUnitRunner;
 
-@RunWith(MockitoJUnitRunner.Silent.class)
+@RunWith(MockitoJUnitRunner.class)
 public class DefaultLoadBalancingPolicyEventsTest extends DefaultLoadBalancingPolicyTestBase {
 
   private DefaultLoadBalancingPolicy policy;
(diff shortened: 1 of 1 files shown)


olim7t:
Mmm I disagree with your fix, but indeed the test is wonky:

The affected methods are `should_remove_down_node_from_live_set_when_filtered` and `should_remove_removed_node_from_live_set_when_filtered`. The goal is to check this scenario:
* LBP initializes, filter does not exclude the node yet
* filter starts to exclude the node dynamically
* node goes down or leaves the cluster => the change in the filter should not prevent the LBP from removing it from the live set.

In other words, we want to check that the filter is not called in `onDown` or `onRemove`. The mock is not called because that's specifically what we're testing.
I agree that the test is badly written and confusing. I think we could get rid of those 2 tests, and simply check that the filter has not been invoked (more than once at init time) in `should_remove_down_node_from_live_set` and `should_remove_removed_node_from_live_set`.

adutra:
Indeed, I haven't grasped the actual meaning of these tests. Now that it's clearer to me, I agree with your suggestions.

adutra:
> filter starts to exclude the node dynamically

Hmm what made me miss the intent of these tests is the fact that both have `Mockito.when(filter.test(node2)).thenReturn(true)` where it should be `....thenReturn(false)`, correct?

olim7t:
Yes, there's that too. A lot was wrong in these tests 🤦‍♂️

olim7t (approved):

PR #1099, opened 2018-09-19
Session methods returning CompletionStage should allow covariance

adutra:
Motivation:

We experienced problems by the past trying to extend classes and
override methods returning CompletionStage<SomeInterface>. It is
usually better to return CompletionStage<? extends SomeInterface> since
this can be overridden by CompletionStage<? extends SomeChildInterface>
thanks to type covariance.

We modified lots of methods but for some reason we forgot a few ones in
CqlSession and DefaultSession.

Modifications:

- Modify return types of methods in CqlSession and DefaultSession from
  CompletionStage<T> to CompletionStage<? extends T>.

Result:

Methods in CqlSession and DefaultSession can now be overriden more
easily.


olim7t (approved):
👍 makes sense.
Nit: could you change the commit message and revapi justifications to something affirmative? e.g. "Return covariant future types from session async methods"

PR #1100, opened 2018-09-22
Improvements to contribution guidelines

adutra:


CONTRIBUTING.md
```diff
@@ -198,22 +198,29 @@ Add them for all new code, with the exception of:
...
+the types from `edu.umd.cs.findbugs.annotations`, there are homonyms in the classpath.
+
+
 ## Coding style -- test code
 
 Static imports are permitted in a couple of places:
-* AssertJ's `assertThat` / `fail`.
-* Some Mockito methods, provided that you're already using a non-statically imported method at the
-  beginning of the line. For example:
+* AssertJ methods whose names start with `assertThat` or `fail`.
```
adutra: Adding this to also allow static import of `assertThatThrownBy`. There is also `catchThrowable` which is very convenient, not sure if we should include it or not.

olim7t: At this point we might as well say "all AssertJ methods". They're pretty recognizable since you know you're looking at test code.

CONTRIBUTING.md
```diff
@@ -198,22 +198,29 @@ Add them for all new code, with the exception of:
 Make sure you import the types from `net.jcip`, there are homonyms in the classpath.
 
 
+### Nullability annotations
+
+We use the [Spotbugs annotations](https://spotbugs.github.io) to document nullability of parameters,
+method return types and class members.
+
+Please annotate any new class or interface with the appropriate annotations. Make sure you import 
+the types from `edu.umd.cs.findbugs.annotations`, there are homonyms in the classpath.
```
olim7t, line 207: 👍 , it is unfortunately very easy to make that mistake.

Maybe we should also name the annotations explicitly, there is also a `@NotNull` in the classpath:
```
Please annotate any new class or interface with the appropriate annotations: `@NonNull`, `@Nullable`.
```

CONTRIBUTING.md
```diff
@@ -198,22 +198,29 @@ Add them for all new code, with the exception of:
...
 Static imports are permitted in a couple of places:
-* AssertJ's `assertThat` / `fail`.
-* Some Mockito methods, provided that you're already using a non-statically imported method at the
-  beginning of the line. For example:
+* AssertJ methods whose names start with `assertThat` or `fail`.
+* Mockito methods, e.g.:
   ```java
-  // any and eq are statically imported, it's pretty clear that they at least relate to Mockito
-  Mockito.verify(intCodec).decodePrimitive(any(ByteBuffer.class), eq(ProtocolVersion.DEFAULT));
+  verify(intCodec).decodePrimitive(any(ByteBuffer.class), eq(ProtocolVersion.DEFAULT));
```
olim7t: Sounds good to me. Again Mockito is pretty easy to recognize from the context once you're familiar with it, and we can take a bit more liberty in test code.

But then for consistency we should do a pass on existing code and static import it everywhere. I'm not sure if there's a smart way to do that in IDEA.

olim7t (changes requested):

CONTRIBUTING.md
adutra: > But then for consistency we should do a pass on existing code and static import it everywhere. I'm not sure if there's a smart way to do that in IDEA.

Yes I was planning to do this in a separate commit once we agree on the guidelines. It will probably be a mix of automatic search-replace and manual editing.

adutra: It seems like IntelliJ has a way to achieve this, see [here](https://stackoverflow.com/questions/53614944/replace-import-of-constant-by-static-import-in-intellij-idea).

olim7t (approved):

PR #1101, opened 2018-09-24
JAVA-1978: Add a config option to keep contact points unresolved

olim7t:


adutra (approved):

olim7t:
Rebased on current 4.x.

I also detected an issue: in a few places we need to find out the node of an existing `DriverChannel`; this was using the underlying Netty channel's remote address, which is resolved. `DriverChannel` now stores the unresolved address and uses that instead.

PR #1103, opened 2018-10-05
JAVA-1973: Avoid calling public methods in HandlerBase constructors

adutra:


olim7t:
I have mixed feelings: I understand the issue, but out of principle it annoys me to introduce an extra volatile write for the timeout future. The base classes are for internal reuse, not really intended to be extended from the outside, so I feel that we have enough control to ensure that bad things won't happen.

adutra:
> but out of principle it annoys me to introduce an extra volatile write for the timeout future

If that's annoying you (and I agree it's annoying now that you mention the extra volatile write cost) then we can remove it and create the timeout future as part of the object initialization.

The reason why I moved its instantiation to the `start()` method is that I thought that conceptually speaking, it would be better to only create that future once the request is "officially" started by a call to `start()`. This would be particularly true if some request handler had to perform some heavy-weight, potentially long computation _after_ the handler is instantiated but _before_ calling `start()` – but this situation seems pretty hypothetical at this point.

So I will simply go ahead and push a commit shortly to move the future instantiation back to the constructor.

adutra:
Recent changes to the RequestProcessor data model make this ticket obsolete; closing.

PR #1105, opened 2018-10-08
JAVA-1960: Ensure type parameters have meaningful names

olim7t:
Trivial change, no runtime impacts (no changelog entry for that reason), but I think it really helps.
I've kept single-letter names for cases where the type is truly abstract (e.g. `GenericType`) and/or simple one-liners.

### DIFF
--- a/core/revapi.json
+++ b/core/revapi.json
@@ -514,7 +514,7 @@
       },
       {
         "code": "java.method.addedToInterface",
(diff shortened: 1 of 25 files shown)


adutra (approved):
Code is much easier to read now 👍 
I was wondering: the naming convention for type parameters that consists of adding a final `T` to the type parameter name comes from [Google's own naming conventions](https://google.github.io/styleguide/javaguide.html#s5.2.8-type-variable-names), if I'm not mistaken. I would suggest that we add a note about naming conventions in the contribution guidelines and ask contributors to abide  by Google's conventions.

olim7t:
It's mentioned already, at the beginning of the file.

adutra:
> It's mentioned already, at the beginning of the file.

Well, that's embarrassing :) Thanks for clarifying.

PR #1109, opened 2018-10-22
JAVA-2003: Simplify CBUtil internal API to improve performance

adutra:


tolbertam (approved):
Nice :+1: avoiding going between array and list looks like a clear nice improvement.

PR #1110, opened 2018-10-22
JAVA-2002: Reimplement TypeCodec.accepts to improve performance (4.x)

adutra:


core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodec.java
```diff
@@ -82,19 +82,39 @@ default boolean accepts(@NonNull GenericType<?> javaType) {
...
-   * Whether this codec is capable of processing the given Java type.
+   * Whether this codec is capable of processing the given Java class.
    *
-   * <p>The default implementation wraps the class in a generic type and calls {@link
-   * #accepts(GenericType)}, therefore it is invariant as well.
+   * <p>This implementation simply compares the given class (or its wrapper type if it is a
+   * primitive type) against this codec's runtime (raw) class; it is <em>invariant</em> with respect
+   * to the passed argument (through the usage of {@link Class#equals(Object)} and <em>it's strongly
+   * recommended not to modify this behavior</em>. This means that a codec will only ever return
+   * {@code true} for the <em>exact</em> runtime (raw) Java class that it has been created for.
```
olim7t, line 91: That's slightly more permissive. If you pass a raw list, any list codec will now match, whereas previously the codec would have to explicitly declare a raw type:
```java
Class<?> arg = List.class;
GenericType<List<Integer>> codecsJavaClass = GenericType.listOf(Integer.class);

// Previous behavior:
assert !codecsJavaClass.equals(GenericType.of(arg));
// Codec would have to declare this to match:
GenericType<List> codecsJavaClass2 = GenericType.of(List.class);
assert codecsJavaClass2.equals(GenericType.of(arg));

// New behavior
assert codecsJavaClass.getRawType().equals(arg);
```

Admittedly, this is probably not a big deal since passing a raw list to `accepts` is wrong in the first place.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodec.java
```diff
@@ -82,19 +82,39 @@ default boolean accepts(@NonNull GenericType<?> javaType) {
...
    * <p>Implementors are encouraged to override this method if there is a more efficient way. In
-   * particular, if the codec targets a non-generic class, the check can be done with {@code
-   * Class#isAssignableFrom}. Or even better, with {@code ==} if that class also happens to be
-   * final.
+   * particular, if the codec targets a final class, the check can be done with a simple {@code ==}.
    */
   default boolean accepts(@NonNull Class<?> javaClass) {
     Preconditions.checkNotNull(javaClass);
-    return accepts(GenericType.of(javaClass));
+    if (javaClass.isPrimitive()) {
```
olim7t, line 98: That makes me think, we should probably also test for primitive classes in implementations, for example in `IntCodec`:
```java
  @Override
  public boolean accepts(@NonNull Class<?> javaClass) {
    return javaClass == Integer.class || javaClass == int.class;
  }
```

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodec.java
adutra: Indeed, it's more permissive, but I was operating under the assumption that calling this method for generic types is wrong.

Slightly related: I'm realizing that lots of overridings of this method are using `Class.isAssignableFrom`, e.g. `UdtCodec` and `TupleCodec`. While this in practice should be OK, I was wondering if for consistency we shouldn't use invariance and hence `Class.equals` instead. The general contract of this method is that if you create a codec for some interface or abstract type (say, `UdtValue`), your lookups should be done by that exact type, not by some subtype like `DefaultUdtValue`.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodec.java
adutra: Indeed, I will push the changes in a separate commit.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodec.java
adutra: Also: the codecs I mentioned do not override `accepts(GenericType)` which leads to the following inconsistency:

```
TypeCodecs.BLOB.accepts(MappedByteBuffer.class); // returns true
TypeCodecs.BLOB.accepts(GenericType.of(MappedByteBuffer.class)); // returns false
```
We must either make both methods consistently covariant where applicable, or invariant everywhere.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodec.java
adutra: To be clear: codecs always store the wrapper types for primitive types, so in practice there should be no lookup in the registry by primitive types, but it doesn't hurt to add that.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodec.java
adutra: I'm going to go with invariance everywhere simply because this method is used for both encoding and decoding.

adutra: It is probably too late but I was wondering if a better API here wouldn't be possible: if we could distinguish lookups for encoding from lookups for decoding (i.e. `codecForEncoding` vs `codecForDecoding` methods), and also distinguish two different sets of `accepts` methods (i.e., `canEncode` vs `canDecode`), then we could apply covariance consistently for all encoding operations, and contravariance for all decoding operations.

olim7t: Yeah, I remember we explored the idea of contravariance for decoding when we first added the codec registry, and concluded that it was not worth it.

olim7t: > I'm going to go with invariance everywhere

👍 both methods were meant to be invariant, and documented as such. Using `isAssignableFrom` was an oversight.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -118,7 +118,7 @@ protected CachingCodecRegistry(
       @NonNull DataType cqlType, @NonNull Class<JavaTypeT> javaType) {
     LOG.trace("[{}] Looking up codec for {} <-> {}", logPrefix, cqlType, javaType);
     TypeCodec<?> primitiveCodec = primitiveCodecsByCode.get(cqlType.getProtocolCode());
-    if (primitiveCodec != null && primitiveCodec.getJavaType().__getToken().getType() == javaType) {
+    if (primitiveCodec != null && primitiveCodec.accepts(javaType)) {
```
olim7t, line 121: I think we can also simplify this.

olim7t (approved):
👍 I agree with all the changes.

PR #1111, opened 2018-10-22
JAVA-2004: Avoid map lookups in ProtocolVersion.fromInt

adutra:


driver-core/src/main/java/com/datastax/driver/core/ProtocolVersion.java
```diff
@@ -91,9 +78,19 @@ public ProtocolVersion getLowerSupported() {
...
+        return ProtocolVersion.V2;
+      case 3:
+        return ProtocolVersion.V3;
+      case 4:
+        return ProtocolVersion.V4;
+      case 5:
+        return ProtocolVersion.V5;
+      default:
+        throw new IllegalArgumentException("No protocol version matching integer version " + i);
+    }
```
olim7t, line 94: The tradeoff is that this needs to be maintained manually.

How much of an improvement are we talking about? Was it measured in a microbenchmark, or a real-world workload?
With the immutable map, a key lookup is essentially an array traversal, so I would expect it to be pretty fast as well.

adutra: > The tradeoff is that this needs to be maintained manually.

Yes, but I don't think there is any risk of forgetting to add a switch branch here when a new enum constant is added, since the mistake would be immediately visible due to an `IllegalArgumentException` being thrown.

> How much of an improvement are we talking about? Was it measured in a microbenchmark, or a real-world workload?

A benchmark from our internal benchmarks with a mixed workload (reads + writes). The CPU profiling shows that `Frame.create` spends 100% of its time inside `ProtocolVersion.fromInt`; that method alone represents 0.01% of the total CPU time for the benchmark, which seems a lot for just a protocol version lookup.

adutra:
Closing, I tried to reproduce the benchmarks and was unable to distinguish a real performance gain here.

PR #1112, opened 2018-10-22
JAVA-2001: Handle zero timeout in admin requests

olim7t:


emerkle826 (approved):
This looks good to me. Tested on the Ironic lab with `basic.request.timeout` set to 0 and there are no PT0S timeouts.

adutra (approved):

PR #1115, opened 2018-10-26
JAVA-2011: Re-add ResultSet.getAvailableWithoutFetching() and isFullyFetched()

olim7t:


core/src/main/java/com/datastax/oss/driver/internal/core/cql/MultiPageResultSet.java
```diff
@@ -95,6 +107,10 @@ private void maybeMoveToNextPage() {
       }
     }
 
+    private boolean isFullyFetched() {
+      return currentPage.hasMorePages();
```
adutra: Should be `!currentPage.hasMorePages()`.

adutra (changes requested):

olim7t:
Merging this now, I've addressed all the comments.

PR #1118, opened 2018-11-01
JAVA-2017: Slightly optimize conversion methods on the hot path

olim7t:
Nothing spectacular, I've benchmarked locally and they should save at best a few nanoseconds per request.
I'm not sure if that's significant, but they shouldn't hurt either.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/DefaultConsistencyLevelRegistry.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/DefaultConsistencyLevelRegistry.java
@@ -18,26 +18,36 @@ package com.datastax.oss.driver.internal.core;
 import com.datastax.oss.driver.api.core.ConsistencyLevel;
 import com.datastax.oss.driver.api.core.DefaultConsistencyLevel;
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
+import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableMap;
 import net.jcip.annotations.ThreadSafe;
 
 @ThreadSafe
 public class DefaultConsistencyLevelRegistry implements ConsistencyLevelRegistry {
 
-  private static final ImmutableList<ConsistencyLevel> values =
+  private static final ImmutableList<ConsistencyLevel> VALUES =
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/DefaultConsistencyLevelRegistry.java
```diff
@@ -18,26 +18,36 @@
...
   @Override
   public ConsistencyLevel fromCode(int code) {
     return DefaultConsistencyLevel.fromCode(code);
   }
 
   @Override
-  public ConsistencyLevel fromName(String name) {
-    return DefaultConsistencyLevel.valueOf(name);
+  public int toCode(String name) {
+    return NAME_TO_CODE.get(name);
```
olim7t, line 46: This is the most significant change in this PR.
For the record I've tried immutable map, hash map and an explicit switch statement in the method, no discernable difference between the 3.

adutra: There is now a slight dissymmetry between methods `fromCode` and `toCode` and I don't see why you had to introduce it; couldn't `toCode` simply keep returning the `ConsistencyLevel` instance itself (just as `fromCode` does)? (In which case, the old name `fromName` would be a better name for it).

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/DefaultConsistencyLevelRegistry.java
olim7t: The method is only called from `Conversions`, and (String => int) is what we really need there. Previously it would convert to the enum and immediately call `getProtocolCode()` on it, which is an unnecessary extra step (String => ConsistencyLevel => int).

As for the names I see what you mean, this is confusing so maybe we should try to be more explicit:
```java
  ConsistencyLevel codeToLevel(int code);
  int nameTocode(String name);
```

adutra: I see, we are really trying to squeeze up to the last drop of performance here. OK then for the method names suggested in your last comment above.

adutra (approved):

PR #1119, opened 2018-11-02
JAVA-1914: Optimize use of System.nanoTime in CqlRequestHandlerBase

olim7t:


core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
```diff
@@ -298,22 +307,29 @@ private void setFinalResult(
...
-            .updateTimer(
-                DefaultSessionMetric.CQL_REQUESTS,
-                executionProfile.getName(),
-                totalLatencyNanos,
-                TimeUnit.NANOSECONDS);
+
+        // Only call nanoTime() if we're actually going to use it
+        long completionTimeNanos = -1, totalLatencyNanos = -1;
+        if (!(requestTracker instanceof NoopRequestTracker)) {
+          completionTimeNanos = System.nanoTime();
```
olim7t, line 316: Contrary to what I suggested on the ticket, I decided not to reuse the time of the last `NodeResponseCallback.onResponse()`, because this method does a bit more stuff before updating the metrics (notably a call to `Conversions.toResultSet`).

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
```diff
@@ -575,18 +598,17 @@ private void processErrorResponse(Error errorMessage) {
...
                             || prepareError instanceof ProtocolError) {
                           LOG.trace("[{}] Unrecoverable error on reprepare, rethrowing", logPrefix);
-                          trackNodeError(node, prepareError);
+                          trackNodeError(node, prepareError, -1);
                           setFinalError(prepareError, node, execution);
                           return null;
                         }
                       }
                     } else if (exception instanceof RequestThrottlingException) {
-                      trackNodeError(node, exception);
```
olim7t, line 584: Not directly related to this ticket, but we don't need to track the error here since throttling happens on the client.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
```diff
@@ -298,22 +307,29 @@ private void setFinalResult(
...
-            .getMetricUpdater()
-            .updateTimer(
-                DefaultSessionMetric.CQL_REQUESTS,
-                executionProfile.getName(),
-                totalLatencyNanos,
-                TimeUnit.NANOSECONDS);
+
+        // Only call nanoTime() if we're actually going to use it
+        long completionTimeNanos = -1, totalLatencyNanos = -1;
+        if (!(requestTracker instanceof NoopRequestTracker)) {
```
tomekl007, line 315: In what case we can have NoopRequestTracker as the requestTracker? Is this often the case?

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
```diff
@@ -423,7 +439,7 @@ public void operationComplete(Future<java.lang.Void> future) throws Exception {
         Throwable error = future.cause();
         if (error instanceof EncoderException
             && error.getCause() instanceof FrameTooLongException) {
-          trackNodeError(node, error.getCause());
+          trackNodeError(node, error.getCause(), -1);
```
tomekl007: I think that we can create some static variable, i.e:
```
private static final NO_MEASURMENT = -1;
```
instead of a magic number -1 and then refer to it

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
```diff
@@ -755,11 +777,20 @@ public void cancel() {
...
+    /**
+     * @param nodeResponseTimeNanos the time we received the response, if it's already been
+     *     measured. If negative, it hasn't and we need to measure it now (this is to avoid
+     *     unnecessary calls to System.nanoTime)
+     */
+    private void trackNodeError(Node node, Throwable error, long nodeResponseTimeNanos) {
+      if (requestTracker instanceof NoopRequestTracker) {
+        return;
+      }
+      if (nodeResponseTimeNanos < 0) {
```
tomekl007: then this check will be more explicit like
```
if( nodeResponseTimeNanos == NO_MEASURMENT) {
...
}
```

core/src/test/java/com/datastax/oss/driver/internal/core/cql/RequestHandlerTestHarness.java
```diff
@@ -148,6 +149,8 @@ protected RequestHandlerTestHarness(Builder builder) {
         .thenReturn(CompletableFuture.completedFuture(null));
 
     Mockito.when(session.getMetricUpdater()).thenReturn(sessionMetricUpdater);
+    Mockito.when(sessionMetricUpdater.isEnabled(any(SessionMetric.class), anyString()))
+        .thenReturn(true);
```
tomekl007, line 154: didn't we miss a test when that ```NoopRequestTracker``` is used as a requestTracker?

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
olim7t: It's the default (see `advanced.request-tracker.class` in `reference.conf`).

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
olim7t: OK, I extracted a constant.

core/src/test/java/com/datastax/oss/driver/internal/core/cql/RequestHandlerTestHarness.java
olim7t: I added `CqlRequestHandlerTrackerTest`.

tomekl007 (approved):

PR #1120, opened 2018-11-06
JAVA-1995: Use rule chains to enforce JUnit rule order

olim7t:
No changelog entry because this does not impact production code.

Summary of changes: take a look at `SessionRule` and `CassandraResourceRule`, then it's the same change in every integration test.

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/api/core/ConnectKeyspaceIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/api/core/ConnectKeyspaceIT.java
@@ -27,12 +27,16 @@ import com.datastax.oss.driver.categories.ParallelizableTests;
 import org.junit.ClassRule;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
 
 @Category(ParallelizableTests.class)
 public class ConnectKeyspaceIT {
-  @ClassRule public static CcmRule ccm = CcmRule.getInstance();
+  private static CcmRule ccm = CcmRule.getInstance();
 
-  @ClassRule public static SessionRule<CqlSession> sessionRule = SessionRule.builder(ccm).build();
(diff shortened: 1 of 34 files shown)


integration-tests/src/test/java/com/datastax/oss/driver/api/core/ConnectKeyspaceIT.java
```diff
@@ -27,12 +27,16 @@
 import org.junit.ClassRule;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
 
 @Category(ParallelizableTests.class)
 public class ConnectKeyspaceIT {
-  @ClassRule public static CcmRule ccm = CcmRule.getInstance();
+  private static CcmRule ccm = CcmRule.getInstance();
```
olim7t, line 35: JUnit only accesses the chain, so individual rules don't need to be public anymore.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/tracker/RequestLoggerIT.java
```diff
@@ -139,6 +137,13 @@
                   .build())
           .build();
 
+  @Rule
+  public TestRule chain =
+      RuleChain.outerRule(simulacronRule)
+          .around(sessionRuleRequest)
+          .around(sessionRuleNode)
+          .around(sessionRuleDefaults);
```
olim7t, line 145: Strictly speaking the 3 session rules don't require any ordering with respect to each other, but that's the only way to embed them all into the simulacron rule.

olim7t (commented):
Note that if the server rule is a `@ClassRule` but the session rule is a `@Rule`, a chain is not required because the former is automatically initialized first.

olim7t:
> Note that if the server rule is a `@ClassRule` but the session rule is a `@Rule`, a chain is not required because the former is automatically initialized first.

Forgot to mention: the only existing case of this is `DefaultRetryPolicyIT`.

emerkle826 (approved):

PR #1123, opened 2018-11-07
JAVA-1918: Document temporal types

olim7t:


manual/core/time_types/README.md
```diff
@@ -0,0 +1,113 @@
+## Time types
```
adutra: Wouldn't "Temporal types" sound more appropriate?

manual/core/time_types/README.md
```diff
@@ -0,0 +1,113 @@
...
+```
+$ TZ=UTC cqlsh
+cqlsh> SELECT * FROM test;
+
+ t
+---------------------------------
+ 2018-11-07 16:50:52.433000+0000
+```
+
+Internally, Cassandra only stores the raw number of milliseconds. You can observe that with a cast:
```
adutra: Congrats, this explanatory section is extremely helpful imo.

manual/core/time_types/README.md
```diff
@@ -0,0 +1,113 @@
...
+``` 
+
+There are various constants and methods to obtain a codec instance for a particular zone:
+
+* [TypeCodecs.ZONED_TIMESTAMP_SYSTEM]\: system default;
+* [TypeCodecs.ZONED_TIMESTAMP_UTC]\: UTC;
+* [TypeCodecs.zonedTimestampAt()]\: user-provided.
+
+Which zone you choose is application-dependent. The driver doesn't map to `ZonedDateTime` by default
+because it would have to make a arbitrary choice; we want you to think about time zones explicitly
```
adutra: Small typo: "...to make _an_ arbitrary choice".

manual/core/time_types/README.md
```diff
@@ -0,0 +1,113 @@
...
+for predictable results.
+
+#### Millisecond-only precision
+
+As already stated, `timestamp` is stored as a number of milliseconds. If you try to write an
+`Instant` or `ZonedDateTime` with higher precision through the driver, the sub-millisecond part will
+be truncated.
```
adutra: Maybe add an example of such truncation here? I'm not sure this is clear for all users.

manual/core/time_types/README.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Duration
+
+CQL type `duration` represents a period in months, days and nanoseconds. The driver maps it to a
+custom type: [CqlDuration].
+
+We deliberately avoided `java.time.Duration`, because it represents an absolute time-based amount,
+regardless of the calendar. In contrast, a `CqlDuration` of "2 months" represents a different amount
```
adutra: Suggestion: "We deliberately avoided `java.time.Period`, because it does not contain a nanoseconds part as `CqlDuration` does; and we also avoided `java.time.Duration`, because it represents an absolute time-based amount, regardless of the calendar, whereas `CqlDuration` manipulates conceptual days and months instead. Thus a `CqlDuration` of "2 months" represents a different amount of time depending on the date to which it is applied  (because months have a different number of days, and because daylight savings rules might also apply, etc).

manual/core/README.md
```diff
@@ -179,35 +179,34 @@ See [AccessibleByName] for an explanation of the conversion rules.
...
+|---------------------|----------------|----------------------|-----------------------------|
+| ascii               | getString      | java.lang.String     |                             |
+| bigint              | getLong        | long                 |                             |
+| blob                | getBytes       | java.nio.ByteBuffer  |                             |
+| boolean             | getBoolean     | boolean              |                             |
+| counter             | getLong        | long                 |                             |
+| date                | getLocalDate   | java.time.LocalDate  | [Time types](time_types/)   |
+| decimal             | getBigDecimal  | java.math.BigDecimal |                             |
+| double              | getDouble      | double               |                             |
+| duration            | getCqlDuration | CqlDuration          | [Time types](time_types/)   |
```
adutra: Should we mention the FQCN of driver-specific types? Or would it blow up the table layout?

adutra (changes requested):

manual/core/README.md
olim7t: I can link them to the javadoc.

olim7t:
Sorry I squashed the commits too soon, I thought the PR was approved already :(

manual/core/temporal_types/README.md
```diff
@@ -0,0 +1,141 @@
...
+#### Millisecond-only precision
+
+As already stated, `timestamp` is stored as a number of milliseconds. If you try to write an
+`Instant` or `ZonedDateTime` with higher precision through the driver, the sub-millisecond part will
+be truncated:
+
+```java
+CqlSession session =
+    CqlSession.builder()
+        .addTypeCodecs(TypeCodecs.ZONED_TIMESTAMP_UTC)
```
adutra, line 103: TBH the example could be simplified by using `Instant`, but since you just introduced the `ZonedDateTime` codec, I'm fine.

adutra (approved):

PR #1126, opened 2018-11-09
JAVA-2028: Use CQL form when parsing UDT types in system tables

olim7t:


integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/CaseSensitiveUdtIT.java
```diff
@@ -0,0 +1,91 @@
...
+                  .build())
+          .build();
+
+  @ClassRule public static TestRule chain = RuleChain.outerRule(ccmRule).around(sessionRule);
+
+  @BeforeClass
+  public static void createSchema() {
+    CqlSession session = sessionRule.session();
+    session.execute("CREATE TYPE \"Address\"(street text)");
+    session.execute("CREATE TABLE user(id uuid PRIMARY KEY, address \"Address\")");
```
olim7t: I'm going to expand the test to cover the other cases I mentioned on the ticket (functions, aggregates, etc.).

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/CaseSensitiveUdtIT.java
```diff
@@ -0,0 +1,129 @@
...
+ * Checks that case-sensitive UDT names are properly handled in schema metadata.
+ *
+ * <p>In Cassandra >= 2.2, whenever a UDT is referenced in a system table (e.g. {@code
+ * system_schema.columns.type}, it uses the CQL form. This is in contrast to the UDT definition
+ * itself ({@code system_schema.types.type_name}), which uses the internal form.
+ *
+ * @see <a href="https://datastax-oss.atlassian.net/browse/JAVA-2028">JAVA-2028</a>
+ */
+@Category(ParallelizableTests.class)
+@CassandraRequirement(min = "2.2", description = "Fixed bug doesn't occur with earlier versions")
```
adutra: But shouldn't we still run this test against earlier versions of C* having UDTs (2.1) ? Otherwise how can we know if the changes introduced by this PR aren't introducing any regression in C* 2.1? Granted, functions and aggregates are not concerned in 2.1, but we should at least validate that `keyspace          .getUserDefinedType(CqlIdentifier.fromInternal("Address"))` works as expected.

adutra (changes requested):

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/CaseSensitiveUdtIT.java
olim7t: True, it's not the same code but from a blackbox perspective we can still run the test against former versions with a few modifications.

adutra (approved):

PR #1127, opened 2018-11-12
JAVA-1947: Make schema parsing more lenient and allow missing system_virtual_schema

olim7t:
The issue only happens with DDAC, which reports a `release_version` > 4 but is not a true C* 4 equivalent (no virtual tables). I've simply disabled the warning if the virtual schema keyspace is missing.

I also made schema parsing more lenient: failure to query a system table does not cancel the whole refresh anymore, we keep going with the data we have.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/CassandraSchemaQueries.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/CassandraSchemaQueries.java
@@ -22,6 +22,7 @@ import com.datastax.oss.driver.internal.core.adminrequest.AdminRequestHandler;
 import com.datastax.oss.driver.internal.core.adminrequest.AdminResult;
 import com.datastax.oss.driver.internal.core.adminrequest.AdminRow;
 import com.datastax.oss.driver.internal.core.channel.DriverChannel;
+import com.datastax.oss.driver.internal.core.util.Loggers;
 import com.datastax.oss.driver.internal.core.util.NanoTime;
 import com.datastax.oss.driver.internal.core.util.concurrent.RunOrSchedule;
 import com.datastax.oss.driver.shaded.guava.common.annotations.VisibleForTesting;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/CassandraSchemaQueries.java
```diff
@@ -131,52 +132,67 @@ private void executeOnAdminExecutor() {
...
+            adminExecutor);
   }
 
   @VisibleForTesting
   protected CompletionStage<AdminResult> query(String query) {
     return AdminRequestHandler.query(channel, query, timeout, pageSize, logPrefix).start();
   }
 
+  /**
+   * @param warnIfMissing whether to log a warning if the queried table does not exist: some DDAC
```
adutra, line 177: Not a big deal, but why is this parameter called sometimes `warnIfMissing` and sometimes `warnOnError`?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/CassandraSchemaQueries.java
```diff
@@ -185,16 +201,16 @@ private void handleResult(
...
+                    handleResult(nextResult, nextError, builderUpdater, warnIfMissing),
                 adminExecutor);
-      } else {
-        pendingQueries -= 1;
-        if (pendingQueries == 0) {
-          LOG.debug(
-              "[{}] Schema queries took {}", logPrefix, NanoTime.formatTimeSince(startTimeNs));
-          schemaRowsFuture.complete(schemaRowsBuilder.build());
-        }
+        return;
```
adutra: Wasn't the if/else version a bit clearer?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/CassandraSchemaQueries.java
olim7t: Good catch, I renamed it along the way but forgot one spot.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/CassandraSchemaQueries.java
olim7t: It's not the same logic, now the end of the method (decrement `pendingQueries`) applies both to an error and an unpage/last page result. I'm going to extract that to a method and duplicate the call to make it more clear.

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/CassandraSchemaQueries.java
```diff
@@ -131,70 +132,89 @@ private void executeOnAdminExecutor() {
...
   }
 
   private void query(
       String queryString,
-      Function<Iterable<AdminRow>, CassandraSchemaRows.Builder> builderUpdater) {
+      Function<Iterable<AdminRow>, CassandraSchemaRows.Builder> builderUpdater,
+      boolean warnIfMissing) {
     assert adminExecutor.inEventLoop();
 
     pendingQueries += 1;
```
tomekl007, line 164: the pending queries don't need to be an Atomic because everything is executed by the same thread, right?

olim7t: Yes, that's correct. All queries go through the same Netty I/O thread, and all other actions (sending the queries, processing the results) are also scheduled on that channel's event loop.

PR #1129, opened 2018-11-13
JAVA-1941: Override toString for all schema metadata types

olim7t:
No changelog since this has very little user impact (mainly logs).

I don't think we should put every piece of information in the toString representation: if you need to compare individual fields between two instances, you should probably write tests / do some proper debugging.

So I went with the uniquely identifying fields for each type, along with the hash code for a quick indication of whether two instances are equal.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/metadata/schema/FunctionSignature.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/metadata/schema/FunctionSignature.java
@@ -95,4 +95,19 @@ public class FunctionSignature {
   public int hashCode() {
     return Objects.hash(name, parameterTypes);
   }
+
+  @Override
+  public String toString() {
+    StringBuilder builder = new StringBuilder(name.asInternal()).append('(');
+    boolean first = true;
+    for (DataType type : parameterTypes) {
+      if (first) {
+        first = false;
+      } else {
+        builder.append(", ");
+      }
+      builder.append(type.asCql(true, true));
+    }
+    return builder.append(')').toString();
+  }
 }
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/metadata/schema/FunctionSignature.java
```diff
@@ -95,4 +95,19 @@ public boolean equals(Object other) {
...
+  public String toString() {
+    StringBuilder builder = new StringBuilder(name.asInternal()).append('(');
+    boolean first = true;
+    for (DataType type : parameterTypes) {
+      if (first) {
+        first = false;
+      } else {
+        builder.append(", ");
+      }
+      builder.append(type.asCql(true, true));
```
tomekl007, line 109: we are sure that type is always non-null?

olim7t: Yes. The nullability annotations are not powerful enough to express nullability of collection elements, but in practice function parameters always have types.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/DefaultAggregateMetadata.java
```diff
@@ -149,4 +149,15 @@ public int hashCode() {
         stateFuncSignature,
         stateType);
   }
+
+  @Override
+  public String toString() {
+    return "DefaultAggregateMetadata@"
+        + Integer.toHexString(hashCode())
```
adutra, line 156: Do we want the actual hash code, or `java.lang.System.identityHashCode()`? With the former two semantically equivalent objects would have the same `toString` representation, with the latter they wouldn't.

olim7t: I think semantically equivalent makes the most sense, that way you can tell if two different parser executions produced different results (they always produce distinct instances).

adutra (approved):

PR #1130, opened 2018-11-13
JAVA-2014: Schedule timeouts on a separate Timer

emerkle826:
Motivation:

Scheduling request timeouts and speculative executions on the
main I/O event loop seems to lower overall I/O throughput under
load. Moving those events off of the main I/O event loop will
allow the driver to achieve higher overall throughput.

Modifications:

Add a Netty HashedWheelTimer to the NettyOptions and use that
Timer for scheduling request timeouts and speculative executions.
Also, add configuration parameters for controlling the Timer
tick duration (resolution) and wheel size.

Result:

In Cassandra stress tests, offloading the timeouts and speculative
executions onto a separate timer yields about a 10% increase in
I/O throughput.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
@@ -26,6 +26,8 @@ import io.netty.channel.DefaultEventLoopGroup;
 import io.netty.channel.EventLoopGroup;
 import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.channel.socket.nio.NioSocketChannel;
+import io.netty.util.HashedWheelTimer;
+import io.netty.util.Timer;
 import io.netty.util.concurrent.DefaultPromise;
 import io.netty.util.concurrent.EventExecutorGroup;
 import io.netty.util.concurrent.Future;
@@ -45,6 +47,7 @@ public class DefaultNettyOptions implements NettyOptions {
   private final int adminShutdownQuietPeriod;
   private final int adminShutdownTimeout;
   private final TimeUnit adminShutdownUnit;
+  private final Timer timer;
 
   public DefaultNettyOptions(InternalDriverContext context) {
     DriverExecutionProfile config = context.getConfig().getDefaultProfile();
@@ -74,6 +77,18 @@ public class DefaultNettyOptions implements NettyOptions {
             .setNameFormat(context.getSessionName() + "-admin-%d")
             .build();
     this.adminEventLoopGroup = new DefaultEventLoopGroup(adminGroupSize, adminThreadFactory);
+    // setup the Timer
+    ThreadFactory timerThreadFactory =
+        new ThreadFactoryBuilder()
+            .setThreadFactory(safeFactory)
+            .setNameFormat(context.getSessionName() + "-timer-%d")
+            .build();
+    timer =
+        new HashedWheelTimer(
+            timerThreadFactory,
+            config.getDuration(DefaultDriverOption.NETTY_TIMER_TICK_DURATION).toNanos(),
+            TimeUnit.NANOSECONDS,
+            config.getInt(DefaultDriverOption.NETTY_TIMER_TICKS_PER_WHEEL));
   }
 
   @Override
@@ -117,6 +132,13 @@ public class DefaultNettyOptions implements NettyOptions {
             ioShutdownQuietPeriod, ioShutdownTimeout, ioShutdownUnit));
     DefaultPromise<Void> closeFuture = new DefaultPromise<>(GlobalEventExecutor.INSTANCE);
     combiner.finish(closeFuture);
+    // stop the timer as well
+    timer.stop();
     return closeFuture;
+  }
+
+  @Override
+  public synchronized Timer getTimer() {
+    return timer;
   }
 }
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareHandlerBase.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareHandlerBase.java
@@ -51,10 +51,10 @@ import com.datastax.oss.protocol.internal.request.Prepare;
 import com.datastax.oss.protocol.internal.response.Error;
 import com.datastax.oss.protocol.internal.response.result.Prepared;
 import edu.umd.cs.findbugs.annotations.NonNull;
-import io.netty.util.concurrent.EventExecutor;
+import io.netty.util.Timeout;
+import io.netty.util.Timer;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
-import io.netty.util.concurrent.ScheduledFuture;
 import java.nio.ByteBuffer;
 import java.time.Duration;
 import java.util.AbstractMap;
@@ -88,9 +88,9 @@ public abstract class CqlPrepareHandlerBase implements Throttled {
   private final Queue<Node> queryPlan;
   protected final CompletableFuture<PreparedStatement> result;
   private final Message message;
-  private final EventExecutor scheduler;
+  private final Timer timer;
   private final Duration timeout;
-  private final ScheduledFuture<?> timeoutFuture;
+  private final Timeout scheduledTimeout;
   private final RetryPolicy retryPolicy;
   private final RequestThrottler throttler;
   private final Boolean prepareOnAllNodes;
@@ -144,13 +144,13 @@ public abstract class CqlPrepareHandlerBase implements Throttled {
     }
     this.message =
         new Prepare(request.getQuery(), (keyspace == null) ? null : keyspace.asInternal());
-    this.scheduler = context.getNettyOptions().ioEventLoopGroup().next();
+    this.timer = context.getNettyOptions().getTimer();
 
     this.timeout =
         request.getTimeout() != null
             ? request.getTimeout()
             : executionProfile.getDuration(DefaultDriverOption.REQUEST_TIMEOUT);
-    this.timeoutFuture = scheduleTimeout(timeout);
+    this.scheduledTimeout = scheduleTimeout(timeout);
     this.prepareOnAllNodes = executionProfile.getBoolean(DefaultDriverOption.PREPARE_ON_ALL_NODES);
 
     this.throttler = context.getRequestThrottler();
@@ -171,16 +171,16 @@ public abstract class CqlPrepareHandlerBase implements Throttled {
     sendRequest(null, 0);
   }
 
-  private ScheduledFuture<?> scheduleTimeout(Duration timeout) {
-    if (timeout.toNanos() > 0) {
-      return scheduler.schedule(
-          () -> {
-            setFinalError(new DriverTimeoutException("Query timed out after " + timeout));
+  private Timeout scheduleTimeout(Duration timeoutDuration) {
+    if (timeoutDuration.toNanos() > 0) {
+      return this.timer.newTimeout(
+          (Timeout timeout1) -> {
+            setFinalError(new DriverTimeoutException("Query timed out after " + timeoutDuration));
             if (initialCallback != null) {
               initialCallback.cancel();
             }
           },
-          timeout.toNanos(),
+          timeoutDuration.toNanos(),
           TimeUnit.NANOSECONDS);
     } else {
       return null;
@@ -188,8 +188,8 @@ public abstract class CqlPrepareHandlerBase implements Throttled {
   }
 
   private void cancelTimeout() {
-    if (this.timeoutFuture != null) {
-      this.timeoutFuture.cancel(false);
+    if (this.scheduledTimeout != null) {
+      this.scheduledTimeout.cancel();
     }
   }
 
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
@@ -94,14 +94,30 @@ public class DefaultSession implements CqlSession {
 
   private DefaultSession(InternalDriverContext context, Set<InetSocketAddress> contactPoints) {
     LOG.debug("Creating new session {}", context.getSessionName());
-    this.adminExecutor = context.getNettyOptions().adminEventExecutorGroup().next();
-    this.context = context;
-    this.singleThreaded = new SingleThreaded(context, contactPoints);
-    this.metadataManager = context.getMetadataManager();
-    this.processorRegistry = context.getRequestProcessorRegistry();
-    this.poolManager = context.getPoolManager();
     this.logPrefix = context.getSessionName();
-    this.metricUpdater = context.getMetricsFactory().getSessionUpdater();
+    this.adminExecutor = context.getNettyOptions().adminEventExecutorGroup().next();
+    try {
+      this.context = context;
+      this.singleThreaded = new SingleThreaded(context, contactPoints);
+      this.metadataManager = context.getMetadataManager();
+      this.processorRegistry = context.getRequestProcessorRegistry();
+      this.poolManager = context.getPoolManager();
+      this.metricUpdater = context.getMetricsFactory().getSessionUpdater();
+    } catch (Throwable t) {
+      // Rethrow but make sure we release any resources allocated by Netty. At this stage there are
+      // no scheduled tasks on the event loops so getNow() won't block.
+      try {
+        context.getNettyOptions().onClose().getNow();
+      } catch (Throwable suppressed) {
+        Loggers.warnWithException(
+            LOG,
+            "[{}] Error while closing NettyOptions "
+                + "(suppressed because we're already handling an init failure)",
+            logPrefix,
+            suppressed);
+      }
+      throw t;
+    }
   }
 
   private CompletionStage<CqlSession> init(CqlIdentifier keyspace) {
@@ -288,8 +304,22 @@ public class DefaultSession implements CqlSession {
         context.getSslHandlerFactory();
         context.getTimestampGenerator();
       } catch (Throwable error) {
-        initFuture.completeExceptionally(error);
         RunOrSchedule.on(adminExecutor, this::closePolicies);
+        context
+            .getNettyOptions()
+            .onClose()
+            .addListener(
+                f -> {
+                  if (!f.isSuccess()) {
+                    Loggers.warnWithException(
+                        LOG,
+                        "[{}] Error while closing NettyOptions "
+                            + "(suppressed because we're already handling an init failure)",
+                        logPrefix,
+                        f.cause());
+                  }
+                  initFuture.completeExceptionally(error);
+                });
         return;
       }
 
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -1369,6 +1369,30 @@ datastax-java-driver {
 
       shutdown {quiet-period = 2, timeout = 15, unit = SECONDS}
     }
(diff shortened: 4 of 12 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
```diff
@@ -74,6 +77,24 @@ public DefaultNettyOptions(InternalDriverContext context) {
...
+    ThreadFactory timerThreadFactory =
+        new ThreadFactoryBuilder().setThreadFactory(safeFactory).build();
+    timer =
+        new HashedWheelTimer(
+            timerThreadFactory,
+            config.getDuration(DefaultDriverOption.NETTY_TIMER_TICK_DURATION).toNanos(),
+            TimeUnit.NANOSECONDS,
+            config.getInt(DefaultDriverOption.NETTY_TIMER_TICKS_PER_WHEEL));
+    // In case things shutdown without calling close(), make sure the timer resources are released.
+    Runtime.getRuntime()
```
emerkle826: I don't like having done this. However, some of the integration tests generate a bunch of warnings like this:
```
ERROR io.netty.util.ResourceLeakDetector - LEAK: HashedWheelTimer.release() was not called before it's garbage-collected. See http://netty.io/wiki/reference-counted-objects.html for more information.
```
I am calling `timer.stop` in the `onClose()` method below, but that does not seem to get called in certain ITs. After investigating the test and finding one that caused the warning, it seems that it happens when you attempt to create a new session with a misconfigured DriverConfigLoader. Doing so will:
1) create a DriverContext that has a lazy reference to an instance of NettyOptions, then
1) dereference the DriverContext's NettyOptions, which
1) calls the constructor for DefaultNettyOptions, which
1) creates the Timer instance (line above this one)

However since the test expects the session creation to fail, there is nothing left to invoke `onClose()`, so the Timer's resources are never released, as the warning indicates. The only way I was able to eliminate the warning was to add this shutdown hook.

olim7t: Do you remember which test(s) created the warning?

If `onClose` is not called, the event loops aren't shut down either, that's also a resource leak. We need to make sure that `onClose` is always called if the options were ever initialized.

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
```diff
@@ -74,6 +77,24 @@ public DefaultNettyOptions(InternalDriverContext context) {
             .setNameFormat(context.getSessionName() + "-admin-%d")
             .build();
     this.adminEventLoopGroup = new DefaultEventLoopGroup(adminGroupSize, adminThreadFactory);
+    // setup the Timer
+    ThreadFactory timerThreadFactory =
+        new ThreadFactoryBuilder().setThreadFactory(safeFactory).build();
```
olim7t: Could you use a name format like the two other factories? This is pretty convenient when debugging stuff (thread dumps, etc.)
```suggestion
        new ThreadFactoryBuilder().setThreadFactory(safeFactory).setNameFormat(context.getSessionName() + "-timer-%d").build();
```

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareHandlerBase.java
```diff
@@ -144,13 +144,13 @@ protected CqlPrepareHandlerBase(
     }
     this.message =
         new Prepare(request.getQuery(), (keyspace == null) ? null : keyspace.asInternal());
-    this.scheduler = context.getNettyOptions().ioEventLoopGroup().next();
+    this.timer = context.getNettyOptions().getTimer();
```
olim7t, line 147: Global 👍 to timeout management in handlers.

core/src/main/resources/reference.conf
```diff
@@ -1369,6 +1369,30 @@ datastax-java-driver {
...
+      # cycles taken from I/O operations, which could lead to lower overall I/O throughput.
+      #
+      # Required: yes
+      # Modifiable at runtime: no
+      # Overridable in a profile: no
+      tick-duration = 100 milliseconds
+
+      # Number of ticks in a Timer wheel. The underlying implementation uses Netty's
+      # HashedWheelTimer, which uses a hashes to arrange the timeouts. This effectively controls the
+      # size of the timer wheel.
```
olim7t, line 1389: 👍 great docs.

Nit: I think there is a word missing in "a hashes to arrange the timeouts".

core/src/test/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerSpeculativeExecutionTest.java
```diff
@@ -250,16 +239,14 @@ public void should_fail_if_no_more_nodes_and_initial_execution_is_last(
...
 
       // Run the next scheduled task to start the speculative execution. node2 will reply with a
       // BOOTSTRAPPING error, causing a RETRY_NEXT; but the query plan is now empty so the
       // speculative execution stops.
-      ScheduledTaskCapturingEventLoop.CapturedTask<?> firstExecutionTask =
-          harness.nextScheduledTask();
-      assertThat(firstExecutionTask.getInitialDelay(TimeUnit.MILLISECONDS))
-          .isEqualTo(firstExecutionDelay);
-      firstExecutionTask.run();
+      // next sceduled timeout should be the first speculative execution. Get it and run it.
```
olim7t: Nit: sceduled => scheduled
This is repeated in other places since you copy-pasted this block.

core/src/test/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerTest.java
```diff
@@ -103,32 +100,21 @@ public void should_fail_if_no_node_available() {
...
-      CompletionStage<AsyncResultSet> resultSetFuture =
+      CqlRequestAsyncHandler requestHandler =
           new CqlRequestAsyncHandler(
-                  UNDEFINED_IDEMPOTENCE_STATEMENT,
-                  harness.getSession(),
-                  harness.getContext(),
-                  "test")
-              .handle();
+              UNDEFINED_IDEMPOTENCE_STATEMENT, harness.getSession(), harness.getContext(), "test");
+      CompletionStage<AsyncResultSet> resultSetFuture = requestHandler.handle();
```
olim7t: Any particular reason to extract the `requestHandler` variable in this test?

core/src/test/java/com/datastax/oss/driver/internal/core/util/concurrent/ScheduledTimeoutCapturingTimer.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Implementation of Netty's {@link io.netty.util.Timer Timer} interface to capture scheduled {@link
+ * io.netty.util.Timeout Timeout} instead of running them, events so they can be run manually in
+ * tests.
+ */
+public class ScheduledTimeoutCapturingTimer implements Timer {
```
olim7t: Nit: I think we don't really need "Scheduled" in the name here (in the other class it refers to the fact that it captures tasks submitted via the `schedule` method). Maybe even `CapturingTimer` would be enough...

But it's more a matter of personal taste really, feel free to use what's most clear to you.

core/src/test/java/com/datastax/oss/driver/internal/core/util/concurrent/ScheduledTimeoutCapturingTimer.java
```diff
@@ -0,0 +1,126 @@
...
+   * class that keeps track of the {@link io.netty.util.TimerTask TimerTask} and the {@link
+   * io.netty.util.Timer Timer} implementation that should only be used in tests. The intended use
+   * is to call the {@link io.netty.util.TimerTask#run(io.netty.util.Timeout) run()} method on the
+   * TimerTask when you want to execute the task (so you don't have to depend on a real timer).
+   *
+   * <p>Example:
+   *
+   * <p>// get the timeout Timeout t = retrieveTimeout(); // run the timeout task t.task().run(t);
+   */
+  private static class _Timeout implements Timeout {
```
olim7t: I know it's an internal class, but not a big fan of the underscore in the name. Could we use something more descriptive, like maybe `CapturedTimeout` or `MockTimeout`?

core/src/test/java/com/datastax/oss/driver/internal/core/util/concurrent/ScheduledTimeoutCapturingTimer.java
```diff
@@ -0,0 +1,126 @@
...
+
+  private final ArrayBlockingQueue<Timeout> timeoutQueue = new ArrayBlockingQueue<>(16);
+
+  @Override
+  public Timeout newTimeout(TimerTask task, long delay, TimeUnit unit) {
+    // delay and unit are not needed as the Timeout's TimerTask will be run manually
+    Timeout timeout = new _Timeout(task, this);
+    // add the timeout to the queue
+    timeoutQueue.add(timeout);
+    return timeout;
```
olim7t: Overall 👍 to the implementation of this class.

core/src/test/java/com/datastax/oss/driver/internal/core/util/concurrent/ScheduledTimeoutCapturingTimer.java
```diff
@@ -0,0 +1,126 @@
...
+   *
+   * <p>Example:
+   *
+   * <p>// get the timeout Timeout t = retrieveTimeout(); // run the timeout task t.task().run(t);
+   */
+  private static class _Timeout implements Timeout {
+
+    private final TimerTask task;
+    private final ScheduledTimeoutCapturingTimer timer;
+    private final AtomicBoolean canceled = new AtomicBoolean(false);
```
olim7t: Nit: cancelled

core/src/test/java/com/datastax/oss/driver/internal/core/util/concurrent/ScheduledTimeoutCapturingTimer.java
```diff
@@ -0,0 +1,126 @@
...
+  /**
+   * Implementation of Netty's {@link io.netty.util.Timeout Timeout} interface. It is just a simple
+   * class that keeps track of the {@link io.netty.util.TimerTask TimerTask} and the {@link
+   * io.netty.util.Timer Timer} implementation that should only be used in tests. The intended use
+   * is to call the {@link io.netty.util.TimerTask#run(io.netty.util.Timeout) run()} method on the
+   * TimerTask when you want to execute the task (so you don't have to depend on a real timer).
+   *
+   * <p>Example:
+   *
+   * <p>// get the timeout Timeout t = retrieveTimeout(); // run the timeout task t.task().run(t);
```
olim7t: Use this for code snippets in javadocs:
```
<pre>{@code
    your code here
}</pre>
```

core/src/test/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerSpeculativeExecutionTest.java
```diff
@@ -101,34 +101,28 @@ public void should_schedule_speculative_executions(
...
+      harness.nextScheduledTimeout(); // Discard the timeout task
 
-      ScheduledTaskCapturingEventLoop.CapturedTask<?> firstExecutionTask =
-          harness.nextScheduledTask();
-      assertThat(firstExecutionTask.getInitialDelay(TimeUnit.MILLISECONDS))
-          .isEqualTo(firstExecutionDelay);
+      Timeout speculativeExecution1 = harness.nextScheduledTimeout();
       Mockito.verifyNoMoreInteractions(nodeMetricUpdater1);
-      firstExecutionTask.run();
+      speculativeExecution1.task().run(speculativeExecution1);
```
olim7t, line 110: This looks a bit weird but that comes from the Netty API, nothing we can do about it in this PR.

olim7t (changes requested):
Looks good 👍 

I think we need to look deeper into the `onClose` issue, apart from that only very minor remarks.

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
emerkle826: `DriverConfigValidationIT` was one that caused the warning. That test tries to create a session with a config loader that has an non-existent class specified for a variety of Driver options. If I comment out most of the calls to `should_fail_to_init_with_invalid_policy`, the warning doesn't appear. Seems like a race condition of sorts.

I didn't see any warnings about the event loops, but maybe they aren't tracked the same? I'll try to investigate what's going on some more.

core/src/main/resources/reference.conf
emerkle826: Good catch, I think I originally meant it to say "a hash to arrange the timeouts" and then decided "hashes to arrange the timeouts" and forgot to remove the "a".

core/src/test/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerTest.java
emerkle826: Not any more. At one point in the implementation, I needed it for something, but it seems I don't anymore. I'll revert that bit.

core/src/test/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerSpeculativeExecutionTest.java
emerkle826: For the tasks we schedule on the timer, the Timeout argument to the `run()` call in the tests doesn't matter (it could be `null`). But that is only because the TimerTasks we schedule in the Handlers simply don't reference the Timeout in the runnable block of code. If that runnable block of code ever does, then passing `null` here could potentially throw an NPE, so I opted to pass in a known non-null reference.
But yes, it was (and still looks) weird.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareHandlerBase.java
```diff
@@ -171,25 +171,25 @@ public void onThrottleReady(boolean wasDelayed) {
...
   }
 
-  private ScheduledFuture<?> scheduleTimeout(Duration timeout) {
-    if (timeout.toNanos() > 0) {
-      return scheduler.schedule(
-          () -> {
-            setFinalError(new DriverTimeoutException("Query timed out after " + timeout));
+  private Timeout scheduleTimeout(Duration timeoutDuration) {
+    if (timeoutDuration.toNanos() > 0) {
+      return this.timer.newTimeout(
```
tomekl007, line 176: can we extract common logic for scheduling timeout for CqlRequestHandlerBase and CqlPrepareHandlerBase?
We could create a separate class that is responsible only for timeouts and reuse it for both classes

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
olim7t: This looks like a bug in `DefaultSession`: `NettyOptions.onClose` should be invoked if init fails. Not clear yet why the unclosed event loops don't prevent the VM from exiting, I'm looking into it.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareHandlerBase.java
olim7t: Code reuse in request handlers is a dilemma: they have a lot of things in common, but also a lot of differences. For example both do timeouts but CqlPrepareHandlerBase doesn't have speculative executions, CqlRequestHandlerBase doesn't have to deal with repreparing on other nodes, etc.
If we start to extract some common logic I'm worried it's going to become an ugly abstraction with conditionals everywhere. That also raises the question of exposing that logic for 3rd-party handler implementations, which could make its design even more complicated.

So granted there is a bit of code duplication with the current design, but on the other hand each handler is its own self-contained implementation, that is easy to understand and modify.

olim7t:
I've pushed a commit to address the timer issue, there were two places where `NettyOptions` wasn't closed correctly if session init failed.

This is a separate issue, so it should be kept as its own commit (and moved first when we rebase IMO).

emerkle826:
DefaultSession changes LGTM. :+1:

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
```diff
@@ -94,14 +94,30 @@
...
     LOG.debug("Creating new session {}", context.getSessionName());
-    this.adminExecutor = context.getNettyOptions().adminEventExecutorGroup().next();
-    this.context = context;
-    this.singleThreaded = new SingleThreaded(context, contactPoints);
-    this.metadataManager = context.getMetadataManager();
-    this.processorRegistry = context.getRequestProcessorRegistry();
-    this.poolManager = context.getPoolManager();
     this.logPrefix = context.getSessionName();
-    this.metricUpdater = context.getMetricsFactory().getSessionUpdater();
+    this.adminExecutor = context.getNettyOptions().adminEventExecutorGroup().next();
```
tomekl007, line 98: this method call cannot throw an exception? I see that we enclosed in a `try` everything besides that line

emerkle826: I suppose it could throw an exception, but it would only be in the DefaultNettyOptions constructor. If the constructor doesn't throw an exception, `adminEventExecutor()` simply returns the EventLoopGroup instance created in the constructor, and `next()` just returns the next executor in the group.

If an exception is thrown in the constructor, it will fall out of this `DefaultSession` constructor and bubble up the stack, just as would any Exception that would be thrown from any of the methods moved into the try block.
The reason this line is not in the try block is that if it throws an Exception, there won't be a NettyOptions instance created on the Context (it's lazily created on the first call to `getNettyOptions()` on the context). If there is no NettyOptions instance, then there isn't a need to call `onClose()` to free resources.

The idea here is that if any of these methods (including this line) throws an Exception, it should propagate up. However, if a NettyOptions is successfully created, but something else fails, we need to free the resources created in the NettyOptions before the Exception propagates.

tomekl007: Thanks for a detailed explanation

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareHandlerBase.java
tomekl007: Thanks for a detailed explanation

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
olim7t: 👍 
In theory the `DefaultNettyOptions` constructor could throw after it created the timer, but we wouldn't have a way to stop the timer from here anyway.
And given that it's the last thing done in that constructor, there's little chance that this could happen.

olim7t (approved):

olim7t:
Quick fix while I rebased: I re-added the ability to capture the delay with which a timeout was scheduled, see `CapturedTimeout.getDelay()`.

PR #1132, opened 2018-11-15
Accessor could be used to generate statements

alexott:


manual/object_mapper/using/README.md
```diff
@@ -275,6 +275,10 @@ executed:
       <td><code>ListenableFuture&lt;Result&lt;T&gt;&gt;</code></td>
       <td><code>T</code> must be a mapped class.<br/>Asynchronous execution, returns a list of mapped objects.</td>
     </tr>
+    <tr>
+      <td><code>Statement</code></td>
+      <td>Instance of <code>Statement</code> that could be executed via <code>Session</code> object.</td>
```
adutra: Maybe you could stress the fact that in this case, 1) the query is not executed when the method returns, and 2) the returned statement is a bound statement.

adutra (changes requested):

manual/object_mapper/using/README.md
alexott: Sure, I really first wrote that it returns `BoundStatement`, and then decided to go to more generic `Statement` :-)

done

adutra (approved):

PR #1134, opened 2018-11-16
Bump dropwizard-metrics

robinjhector:
The cassandra driver uses dropwizard metrics. The dropwizard-metrics modules have been updated since a long time, and is now causing conflicts where this driver is used in combination with the dropwizard framework.. 
The conflict lies in a change made in dropwizard-metrics, causing `JmxReporter` to be moved to it's own maven module. It's package name also got updated.

Hopefully this shouldn't be a breaking change for anyone.

### DIFF
--- a/driver-core/pom.xml
+++ b/driver-core/pom.xml
@@ -50,6 +50,11 @@
         </dependency>
 
         <dependency>
+            <groupId>io.dropwizard.metrics</groupId>
+            <artifactId>metrics-jmx</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
--- a/driver-core/src/main/java/com/datastax/driver/core/Metrics.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Metrics.java
@@ -17,10 +17,10 @@ package com.datastax.driver.core;
 
 import com.codahale.metrics.Counter;
 import com.codahale.metrics.Gauge;
-import com.codahale.metrics.JmxReporter;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import com.codahale.metrics.jmx.JmxReporter;
 import com.datastax.driver.core.policies.SpeculativeExecutionPolicy;
 import java.util.HashSet;
 import java.util.Set;
--- a/driver-tests/osgi/pom.xml
+++ b/driver-tests/osgi/pom.xml
@@ -67,6 +67,11 @@
         </dependency>
 
         <dependency>
+            <groupId>io.dropwizard.metrics</groupId>
+            <artifactId>metrics-jmx</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -53,7 +53,7 @@
         <guava.version>19.0</guava.version>
         <netty.version>4.0.56.Final</netty.version>
         <netty-tcnative.version>2.0.7.Final</netty-tcnative.version>
-        <metrics.version>3.2.2</metrics.version>
+        <metrics.version>4.0.2</metrics.version>
         <snappy.version>1.1.2.6</snappy.version>
         <lz4.version>1.4.1</lz4.version>
         <hdr.version>2.1.10</hdr.version>
@@ -140,6 +140,12 @@
             <dependency>
                 <groupId>io.dropwizard.metrics</groupId>
                 <artifactId>metrics-core</artifactId>
+                <version>${metrics.version}</version>
+            </dependency>
+
+            <dependency>
+                <groupId>io.dropwizard.metrics</groupId>
+                <artifactId>metrics-jmx</artifactId>
                 <version>${metrics.version}</version>
             </dependency>
 


tolbertam:
Hi @izrobin, thanks for taking a look into this.

This is something we have considered since the API is the same, but the separation of the JMX reporter into a separate module could give some problems to some of our users who have other projects which rely on metrics 3.x as it depends on metrics-core 4.  Having metrics-jmx as a declared dependency would cause users to need to explicitly exclude that dependency from the driver if they wanted to continue using metrics 3.  Between minor releases we want to keep the upgrade path as simple as we can for users, so we try not to update dependencies to major versions if we can avoid it.

We added [a section in our documentation](https://docs.datastax.com/en/developer/java-driver/3.6/manual/metrics/#metrics-4-compatibility) that covers using metrics 4.x with the driver.

For java driver 4.x, which is currently in beta,  we have upgraded to metrics 4 ([docs](https://docs.datastax.com/en/developer/java-driver/4.0-beta/manual/core/metrics/)), so we will use that there going forward, while the 3.x driver will continue to depend on metrics 3.x but will also work with metrics 4.x.

robinjhector:
Hello! Thanks for the reply @tolbertam 
Since the `metrics-jmx` module have moved its classes to its own package, there wouldn't be any need of excluding the module for 3.x users, right? You should still be able to run both 3.x and the new module side by side as far as I understand.

tolbertam:
It looks like I had a misunderstanding about how the transitive dependency from metrics-jmx to metrics-core 4 would work.  It seems that with maven at least (may behave differently for builds that use ivy like gradle/sbt) that if you explicitly call out a different version in your pom.xml that this version wins out and the newer version in the transitive versions are excluded:

With the following pom.xml:

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.mycompany.app</groupId>
  <artifactId>my-app</artifactId>
  <packaging>jar</packaging>
  <version>1.0-SNAPSHOT</version>
  <name>my-app</name>
  <url>http://maven.apache.org</url>
  <dependencies>
    <dependency>
      <groupId>com.datastax.cassandra</groupId>
      <artifactId>cassandra-driver-core</artifactId>
      <version>3.6.1-SNAPSHOT</version>
    </dependency>
    <dependency>
      <groupId>io.dropwizard.metrics</groupId>
      <artifactId>metrics-core</artifactId>
      <version>3.2.2</version>
    </dependency>
  </dependencies>
</project>
```

`mvn dependency:tree -Dverbose` generates: 

```
[INFO] com.mycompany.app:my-app:jar:1.0-SNAPSHOT
[INFO] +- com.datastax.cassandra:cassandra-driver-core:jar:3.6.1-SNAPSHOT:compile
[INFO] |  +- io.netty:netty-handler:jar:4.0.56.Final:compile
...
...
[INFO] |  +- (io.dropwizard.metrics:metrics-core:jar:4.0.2:compile - omitted for conflict with 3.2.2)
[INFO] |  +- io.dropwizard.metrics:metrics-jmx:jar:4.0.2:compile
[INFO] |  |  +- (io.dropwizard.metrics:metrics-core:jar:4.0.2:compile - omitted for duplicate) <--
[INFO] |  |  \- (org.slf4j:slf4j-api:jar:1.7.25:compile - omitted for duplicate)
[INFO] |  +- org.slf4j:slf4j-api:jar:1.7.25:compile
...
[INFO] +- io.dropwizard.metrics:metrics-core:jar:3.2.2:compile
[INFO] |  \- (org.slf4j:slf4j-api:jar:1.7.22:compile - omitted for conflict with 1.7.25)
[INFO] \- junit:junit:jar:3.8.1:test
```

However, it still might be an issue if you have a dependency in your project that depends on metrics 3.x.  In that case, it looks like it matters the order of the dependencies are declared, for example, see how `slf4j-api` is treated.  1.7.25 is favored over 1.7.22.  At first I thought this was because the newer version was being favored, but switching the order of dependencies to metrics and then the driver, 1.7.22 gets chosen over 1.7.25:

```
[INFO] com.mycompany.app:my-app:jar:1.0-SNAPSHOT
[INFO] +- io.dropwizard.metrics:metrics-core:jar:3.2.2:compile
[INFO] |  \- org.slf4j:slf4j-api:jar:1.7.22:compile
[INFO] \- com.datastax.cassandra:cassandra-driver-core:jar:3.6.1-SNAPSHOT:compile
...
[INFO]    +- (io.dropwizard.metrics:metrics-core:jar:4.0.2:compile - omitted for conflict with 3.2.2)
[INFO]    +- io.dropwizard.metrics:metrics-jmx:jar:4.0.2:compile
[INFO]    |  +- (io.dropwizard.metrics:metrics-core:jar:4.0.2:compile - omitted for conflict with 3.2.2)
[INFO]    |  \- (org.slf4j:slf4j-api:jar:1.7.25:compile - omitted for conflict with 1.7.22)
[INFO]    +- (org.slf4j:slf4j-api:jar:1.7.25:compile - omitted for conflict with 1.7.22)
...
```

As we are currently working on driver 4.x, our aim on the 3.x branch of the driver is to keep things as stable as possible while also adding capabilities to support C* 4.0 features.  Since there already is a documented path to using metrics 4.x with driver 3.x I think we will continue that path.

robinjhector:
Ah, gotcha :)
Thanks for the info! We'll wait a bit for 4.0 then 👍

PR #1136, opened 2018-11-21
JAVA-2037: Prepared statements without bound variables should not throw NPE

adutra:
Our internal CI builds are failing due to this.

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/PreparedStatementIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/PreparedStatementIT.java
@@ -19,6 +19,7 @@ import static junit.framework.TestCase.fail;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import com.datastax.oss.driver.api.core.CqlSession;
+import com.datastax.oss.driver.api.core.DefaultProtocolVersion;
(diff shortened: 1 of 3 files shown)


integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/PreparedStatementIT.java
```diff
@@ -86,18 +87,73 @@ public void setupSchema() {
...
+  }
+
+  @Test
+  public void should_have_non_empty_result_definitions_for_insert_query_with_bound_variable() {
+    try (CqlSession session = SessionUtils.newSession(ccmRule, sessionRule.keyspace())) {
+      PreparedStatement prepared =
+          session.prepare("INSERT INTO prepared_statement_test (a, b, c) VALUES (?, ?, ?)");
+      assertThat(prepared.getVariableDefinitions()).hasSize(3);
+      if (sessionRule.session().getContext().getProtocolVersion().getCode()
+          >= DefaultProtocolVersion.V4.getCode()) {
```
tomekl007, line 108: why do we need to have different logic for >= V4?

adutra: Because primary key indices were introduced in V4. For previous versions, `prepared.getPartitionKeyIndices()` is always empty.

olim7t (approved):

PR #1137, opened 2018-11-21
Make Statement.ASYNC and PrepareRequest.ASYNC of type CompletionStage<? extends X>

adutra:
Since we already agreed that all `executeAsync` methods return a `CompletionStage<? extends ResultT>`, it makes sense to also have `Statement.ASYNC` be of type `CompletionStage<? extends AsyncResultSet>` and `PrepareRequest.ASYNC` of type `CompletionStage<? extends PreparedStatement>`

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/cql/PrepareRequest.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/cql/PrepareRequest.java
@@ -61,8 +61,8 @@ public interface PrepareRequest extends Request {
    * of the driver's built-in helper methods (such as {@link
    * CqlSession#prepareAsync(SimpleStatement)}.
    */
-  GenericType<CompletionStage<PreparedStatement>> ASYNC =
-      new GenericType<CompletionStage<PreparedStatement>>() {};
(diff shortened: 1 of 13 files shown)


olim7t (commented):
I don't think this will be useful in practice:

The reason we made session methods covariant (and I agree with that) is so that someone could write:
```java
interface MyCqlSession extends CqlSession {
  @Override
  CompletionStage<? extends MyAsyncResultSet> executeAsync(Statement<?> statement);
}
```

Then you need to write a new processor; where this PR would help is if you decided to extend `CqlRequestAsyncProcessor`. But why would you do that? It's a trivial class, you'd have to override `newHandler` anyway, so the only thing you're reusing is the logic in `canProcess`, a one-liner.

Same for the handler: the only way you can extend `CqlRequestAsyncHandler` is if you just need to apply an `AsyncResultSet => MyAsyncResultSet` function to the completion stage (`CqlRequestHandlerBase` is not designed for extension and it won't let you do much else). But that's also trivially achieved by wrapping instead of extending.

In other words, the built-in CQL processors and handlers know the exact type they handle, and they are not designed for inheritance, so they don't need to be covariant.

adutra:
Agreed, we can close this.

PR #1140, opened 2018-11-29
JAVA-1909: Maven Archetype for bootstrapping Java Driver project

emerkle826:
Adds an archetype project for generating a bootstrapped project using the 4.x Java Driver.


emerkle826:
In addition to general comments, we should probably decide if this should remain a sub-module of the driver repo, or be moved into its own repo.

java-driver-archetype/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+This will install the archetype locally so you can then use it to create a bootstrap project.
+
+## Working with this Maven Archetype project
+If you are looking to create a bootstrapped project, you can skip down to
+[Creating a Bootstrap project](#creating-a-bootstrap-project)
+
+### Archetype Metadata
+The [archetype-metadata.xml][1] conforms to the [Maven Archetype Descriptor Model][2]. It defines
+properties that must be set to generate a bootstrapped project as well as fileset rules that control
```
tomekl007: I think that typo: `fileset rules`

java-driver-archetype/src/main/resources/archetype-resources/src/main/java/Main.java
```diff
@@ -0,0 +1,22 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Main {
+  private static final Logger LOG = LoggerFactory.getLogger(Main.class);
+
+  public static void main(String[] args) {
+    CqlSessionBuilder builder = CqlSession.builder();
+    // Set the host and port of the Cassandra server here
+    builder.addContactPoint(InetSocketAddress.createUnresolved("${cassandra-host}", ${cassandra-port}));
```
tomekl007: We don't want to give in this archetype a simple configuration mechanism? If we will then we could use an info when user interactively type host and port and load those settings here

java-driver-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+          <artifactId>maven-compiler-plugin</artifactId>
+          <version>${maven.compiler.plugin.version}</version>
+        </plugin>
+      </plugins>
+    </pluginManagement>
+    <plugins>
+      <plugin>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <configuration>
+          <source>1.8</source>
```
tomekl007, line 90: maybe we could also parametrize source 1.8 via generator?

java-driver-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -0,0 +1,86 @@
...
+  <fileSets>
+    <!-- 
+    setting "filtered" to "true" means the selected files will be used as Velocity templates.
+    setting "packaged" to "true" means the selected files will be generated/copied in a
+    directory structure that is prepended by the package property.
+    -->
+    <fileSet filtered="true" packaged="true" encoding="UTF-8" >
+      <directory>src/main/java</directory>
+      <includes>
+        <include>**/*.java</include>
```
tomekl007: If we support only java we should write it in the documentation of archetype. I can imagine that some person will want to use it in different lang (scale, kotlin) but if we will document that it will be explicit that user needs to add that support ourselves.

java-driver-archetype/src/main/resources/archetype-resources/src/main/java/Main.java
emerkle826: It will default to `127.0.0.1` and `9042` here, if you don't explicitly set these two properties. The defaults are in the `archetype-metadata.xml` required properties section. So, when you run `archetype:generate`, if you do not specify these properties (i.e. `-Dcassandra-host=someHost`), it uses the defaults.

You have the chance to change them as there is a confirmation prompt that shows the values to be used. If you don't like what is set, you enter the interactive mode where you can type in what you want. I tried to cover it in the README.md, in the `Interactive confirmation` section. Let me know if I should expand on that a bit more.

java-driver-archetype/src/main/resources/archetype-resources/pom.xml
emerkle826: I guess we could, however the reason this is here is that without it, it defaults to 1.5 (I think). But the code actually need to be at least 1.7 in order to compile due to language features that didn't exist before 1.7. It's also hard-coded in the compiler config for the Java driver project itself. I don't think we want to let them change it to something lower accidentally and then have to explain that it needs to be at least 1.7.

java-driver-archetype/README.md
```diff
@@ -0,0 +1,147 @@
...
+| Property | Default | Example | Description |
+| --- | --- | --- | ---| 
+| groupId | &lt;no default&gt; | com.mycompany.group | Sets the Maven `groupId` value in the pom.xml of the bootstrapped project |
+| artifactId | &lt;no default&gt; | myArtifact | Sets the Maven 'artifactId' value in the pom.xml of the bootstrapped project |
+| version | 1.0-SNAPSHOT | 1.2.3 | Sets the Maven `version` value in the pom.xml of the bootstrapped project |
+| package | &lt;groupId value&gt; | com.mycompany.poc | Controls the base package for all generated source code |
+| logging-level | DEBUG | WARN | Controls the logging level of the generated project (values must be one of &lt;TRACE&vert;DEBUG&vert;INFO&vert;WARN&vert;ERROR&gt; |
+| cassandra-host | 127.0.0.1 | 10.10.1.16 | The host or IP address of the Cassandra instance to which to connect |
+| cassandra-port | 9042 | 9042 | The port to use for the Cassandra connection |
+| datacenter | datacenter1 | datacenter1 | The name of the local datacenter |
```
olim7t: Nit: could you align the columns?
I'm generally against aligning stuff in source code, but Markdown is designed to be readable as source code as well, so it's kind of an exception.

java-driver-archetype/README.md
```diff
@@ -0,0 +1,147 @@
...
+archetype-resources/src/test/resources
+```
+
+The only difference is that the Java source and test class files are not in packaged subdirectories,
+they are flattened into the `java` directories. This is because the `archetype:generate` goal will
+copy those sources, substituting any Velocity properties with values provided, into the desired
+package structure, based on the value provided for the `package` property. See
+[Archetype Properties](#archetype-properties) for more on the `package` property.
+
+## Creating a Bootstrap project
```
olim7t, line 56: I think we should eventually move this section to the manual.

java-driver-archetype/README.md
```diff
@@ -0,0 +1,147 @@
...
+generated `pom.xml` is located in this directory, as well as a basic README.md. Any other files that
+should be copied into the root of the generated project should be placed here. Adding files here
+will require updates to [archetype-metadata.xml][1] to ensure they are explicitly listed in the
+fileset that is copied during project generation.
+
+Additionally in `archetype-resources` is the typical Maven project directory structure:
+
+```
+archetype-resources/src/main/java
+archetype-resources/src/main/rescource
```
olim7t: Small typo: rescource => resource

java-driver-archetype/src/main/resources/archetype-resources/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,14 @@
...
+log4j.appender.STDOUT=org.apache.log4j.ConsoleAppender
+
+# STDOUT uses PatternLayout.
+log4j.appender.STDOUT.layout=org.apache.log4j.PatternLayout
+log4j.appender.STDOUT.layout.ConversionPattern=%d{HH:mm:ss.SSS} [%t] %-5p %.36c - %m%n
+
+# Adjust the Cassandra Driver logging level here if you want more(DEBUG) or less(WARN) logs
+log4j.logger.com.datastax.oss.driver=INFO
+# Set the log level for all logs in the archetype project to DEBUG
+log4j.logger.${package}=${logging-level}
```
olim7t: You can remove "to DEBUG" at the end of the comment, since this is now parameterized. Same in the log4j file.

java-driver-archetype/src/main/resources/archetype-resources/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,14 @@
...
+
+# STDOUT is set to be a ConsoleAppender.
+log4j.appender.STDOUT=org.apache.log4j.ConsoleAppender
+
+# STDOUT uses PatternLayout.
+log4j.appender.STDOUT.layout=org.apache.log4j.PatternLayout
+log4j.appender.STDOUT.layout.ConversionPattern=%d{HH:mm:ss.SSS} [%t] %-5p %.36c - %m%n
+
+# Adjust the Cassandra Driver logging level here if you want more(DEBUG) or less(WARN) logs
+log4j.logger.com.datastax.oss.driver=INFO
```
olim7t: Since the app level is parameterized, should we make the driver level a parameter as well?

Or maybe neither of them, editing the level in the files is not that complicated so I'm not sure if this needs to be handled by the archetype.

java-driver-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+      <dependency>
+        <groupId>ch.qos.logback</groupId>
+        <artifactId>logback-classic</artifactId>
+        <version>${logback.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>log4j</groupId>
+        <artifactId>log4j</artifactId>
+        <version>${log4j.version}</version>
+      </dependency>
```
olim7t, line 63: Is there any way we could make the logging implementation a parameter of the archetype?
And not reference log4j at all in the POM if logback was selected, or vice-versa.

java-driver-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -0,0 +1,155 @@
...
+  <version>${version}</version>
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+    <junit.version>4.12</junit.version>
+    <slf4j.version>1.7.25</slf4j.version>
+    <logback.version>1.2.3</logback.version>
+    <log4j.version>1.2.17</log4j.version>
+    <slf.log4j.version>1.7.25</slf.log4j.version>
+    <java.driver.version>4.0.0-beta3-SNAPSHOT</java.driver.version>
```
olim7t: The driver version should definitely be configurable through the archetype.

If it's not provided, we'll pick the latest 4.x (according to semver, it will always be backward compatible with whatever code we put here). This can be expressed with a range: ```[4.0.0-beta1, 5)```

java-driver-archetype/src/main/resources/archetype-resources/pom.xml
olim7t: I'm in favor of keeping it hard-coded, since it is more of a general Maven concept rather than something directly related to driver usage.

java-driver-archetype/src/main/resources/archetype-resources/src/main/java/Main.java
olim7t: I'm not sure if we want to use `createUnresolved` here. Using unresolved addresses is an advanced feature for specific use cases, and it requires additional configuration to work properly (see JAVA-1978).

For a basic prototype we don't know if it will be needed so it is more of a distraction, I would either stick to the (String, int)  constructor, or maybe inject the address in `application.conf`.

java-driver-archetype/src/main/resources/archetype-resources/README.md
```diff
@@ -0,0 +1,60 @@
+# Cassandra Java Driver Archetype Project
+
+If you are reading this, you've successfully created the Cassandra Java Driver archetype project.
+
+## Running the project
+
+### Ensure Cassandra is running
+
+When generating this project, values for `cassandra-host` and `cassandra-port` were provided. The
+default values are `127.0.0.1` (localhost) and `9042` respectively. Ensure that Cassandra is up and
```
olim7t: You can filter this file too, and remind the exact values that were provided.

java-driver-archetype/src/main/resources/archetype-resources/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+Once you have a local Cassandra running (or have changed `Main.java` to point to some other instance
+that is running), you can build and run this project by executing the following:
+
+```
+mvn compile exec:java -Dexec.mainClass=com.mycompany.group.Main
+```
+
+where `com.mycompany.group` is the value you used when generating the project from the archetype. If
+you didn't override the default, the package should be the same as the `groupId`.
```
olim7t: Same here, you can use filtering to show the command with the package exactly as it was provided.

olim7t (changes requested):
Looks great, and good documentation 👍 

Regarding the code's location: this will evolve at a slower pace than the rest of the driver (hopefully), so I think it would be better to have it as a standalone project. That way we can release only when changes are really necessary, not whenever we release a new driver version.

Also we could put just a bit more complexity in the sample code, to illustrate how to manage the session throughout a longer program. I think a mini-cqlsh would be pretty easy to implement if we don't go too crazy: just a prompt with a few describe commands and the ability to run requests.

In the same vein, webapps are a big part of our audience, so it would be great to have another archetype that generates one using some popular framework (tbd). Again some trivial example, maybe a UI to navigate schema metadata. This can be done at a later time, but I would definitely make the standalone project multi-module ready.

java-driver-archetype/src/main/resources/archetype-resources/src/main/java/Main.java
tomekl007: Ok, so I think that this comment `Set the host and port of the Cassandra server here` is a bit misleading. If I will be a user, I would think that we should set them manually but maybe it is only my impression

java-driver-archetype/src/main/resources/archetype-resources/pom.xml
tomekl007: Ok, so maybe we can leave that info in the documentation that it uses 1.8 on default?

java-driver-archetype/src/main/resources/archetype-resources/pom.xml
emerkle826: I missed this in my fixups. I'll address it in the next one I push.

olim7t: OK, `latest` works too, at least for now.
Eventually it won't when we start releasing 5.x artifacts, but we can update the archetype to handle that.

java-driver-archetype/src/main/resources/archetype-resources/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Ensure Cassandra is running
+
+Cassandra is configured at ${cassandra-host}:${cassandra-port}. Ensure that Cassandra is up and
+running on that host and port. If the Cassandra host/port to which you want to connect has changed,
+or was not set correctly when the project was generated, you will have to edit the `Main.java`
+source file and alter this line accordingly:
+
+```
+builder.addContactPoint(InetSocketAddress.createUnresolved("${cassandra-host}", ${cassandra-port}));
```
olim7t, line 15: This needs to be updated since we 're not using `createUnresolved` anymore.

java-driver-archetype/src/main/resources/archetype-resources/README.md
```diff
@@ -0,0 +1,61 @@
...
+that is running), you can build this project by executing the following:
+
+```
+mvn clean install
+```
+
+Then run the CQL demo with:
+
+```
+mvn exec:java -Dexec.mainClass=${package}.Main -pl ${artifactId}-cql
```
olim7t, line 30: This fails if `${artifactId}-core` is not installed first.
But I think we should go back to a single module in the generated project, see my other comment.

java-driver-archetype/src/main/resources/archetype-resources/__rootArtifactId__-cql/src/main/java/Main.java
```diff
@@ -0,0 +1,138 @@
...
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.LineNumberReader;
+import java.io.PrintStream;
+import java.net.InetSocketAddress;
+import java.util.Iterator;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class Main {
```
olim7t, line 20: Nice, this strikes a good balance between doing more than a simple query without being too complicated.

java-driver-archetype/src/main/resources/archetype-resources/__rootArtifactId__-cql/src/main/java/Main.java
```diff
@@ -0,0 +1,138 @@
...
+  private void cqlshLite(CqlSession session, LineNumberReader input) throws IOException {
+    // provide a prompt
+    promptForQuery(WELCOME);
+    // get the query
+    String query = input.readLine();
+    // execute the query
+    while (!"exit".equalsIgnoreCase(query)) {
+      if (query.startsWith("describe") || query.startsWith("DESCRIBE")) {
+        handleDescribe(session, query);
+      } else {
```
olim7t, line 111: Nit: could we add a "help" command that prints a quick reminder of available commands?

java-driver-archetype/src/main/resources/archetype-resources/__rootArtifactId__-cql/src/main/java/Main.java
```diff
@@ -0,0 +1,138 @@
...
+      System.out.println("\nDescribe target not implemented: '" + describeTarget + "'");
+    }
+  }
+
+  private void executeQuery(CqlSession session, String query) {
+    try {
+      ResultSet rs = session.execute(query);
+      Iterator<Row> iterator = rs.iterator();
+      while (iterator.hasNext()) {
+        Row row = iterator.next();
```
olim7t, line 87: This would look a bit cleaner with a for-each loop:
```java
for (Row row: rs) {
  ...
}
```

java-driver-archetype/src/main/resources/archetype-resources/__rootArtifactId__-web/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>${groupId}</groupId>
+    <artifactId>${rootArtifactId}</artifactId>
+    <version>${version}</version>
+  </parent>
+  <artifactId>${artifactId}</artifactId>
```
olim7t, line 28: When I mentioned multi-module I was thinking more of the archetype project itself. So a multi-module project that deploys multiple archetypes, instead of an archetype that creates a multi-module project.

I think generated projects should remain single-module for simplicity, there will be a bit of duplication across archetypes for configuration files and stuff, but that does not really matter.

java-driver-archetype/src/main/resources/archetype-resources/pom.xml
emerkle826: I can open-end it so there is no upper bound. (i.e. `[4.0.0-beta1,]`
Or, do you want the _default_ value to be that itself, as opposed to `latest` having a special meaning?

olim7t (changes requested):
Thanks for processing the feedback, we need to figure out the multi-module stuff but then it should be good to go.

Also don't merge it directly when it's done: right now it's convenient to keep it here for reviews, but eventually we want to move it to its own github repo.

java-driver-archetype/src/main/resources/archetype-resources/__rootArtifactId__-web/pom.xml
emerkle826: Ah, I totally misunderstood the original comment. I can flip it back.

java-driver-archetype/src/main/resources/archetype-resources/README.md
emerkle826: Yes, it will. That's why I put the `mvn clean install` part right above. But it's moot if we go back to generating a single module.

java-driver-archetype/README.md
```diff
@@ -0,0 +1,146 @@
+
```
tolbertam, line 1: Should the directory be renamed from `java-driver-archetype` to simply `archetype`?  This way it matches up with the other module directories (i.e. `core` instead of `java-driver-core`

java-driver-archetype/README.md
```diff
@@ -0,0 +1,146 @@
...
+This should produce output similar to:
+
+```
+15:57:32.944 [com.mycompany.poc.Main.main()] INFO  c.d.o.d.i.c.DefaultMavenCoordinates - DataStax Java driver for Apache Cassandra(R) (com.datastax.oss:java-driver-core) version 4.0.0-beta1-SNAPSHOT
+15:57:33.200 [myArtifact-admin-0] INFO  c.d.o.d.internal.core.time.Clock - Using native clock for microsecond precision
+15:57:33.431 [com.mycompany.poc.Main.main()] INFO  com.mycompany.poc.Main - Cassandra release version: 3.11.3
+```
+
+[1]: ../blob/master/src/main/resources/META-INF/maven/archetype-metadata.xml
```
tolbertam, line 145: Doesn't appear to work, I think if you simply use `src/main/resources/META-INF/maven/archetype-metadata.xml` it should work though.

java-driver-archetype/src/main/resources/archetype-resources/__rootArtifactId__-web/pom.xml
tolbertam: :+1: I think most users who will benefit from this will probably want something really simple that demonstrates how to configure and use the driver.  Generating a single module maven project will keep things simple.

java-driver-archetype/README.md
```diff
@@ -0,0 +1,146 @@
...
+package structure, based on the value provided for the `package` property. See
+[Archetype Properties](#archetype-properties) for more on the `package` property.
+
+## Creating a Bootstrap project
+
+Generating the bootstrap project requires some properties to be set. Some properties have default
+values and do not need to be specified. All properties can be provided on the command-line used to
+generate the project, and all can be modified interactively after executing the generate command.
+
+### Archetype Properties
```
tolbertam, line 62: Very nice :+1:

tolbertam (commented):
This is really cool!  I haven't completed my review yet, but just sharing comments thus far, will continue tomorrow.

java-driver-archetype/src/main/resources/archetype-resources/README.md
olim7t: > That's why I put the mvn clean install

Oops, I completely missed that

java-driver-archetype/README.md
olim7t: Yes, but we're planning to move it to a standalone project anyway.

emerkle826:
The work in this PR has been moved to a different repository here (https://github.com/datastax/java-driver-archetypes/pull/1). I'll close this once reviewers have had a chance to review updates made in the new repository.

olim7t:
Closing this since we've moved it to its own repo. See datastax/java-driver-archetypes#1.

PR #1141, opened 2018-11-29
JAVA-1943: Fail fast in execute() when the session is closed

olim7t:
As mentioned on the JIRA ticket, this can still race if the session is stopped while another thread is still executing requests. Requests might fail with `NoNodeAvailableException` if they hit closed pools. The point is more to signal obvious errors where the user is trying to use a closed session.

I think making this completely thread-safe would be prohibitive in terms of performance, and is not that useful in practice. In a typical webapp, you would probably not shut down abruptly, but rather redirect all incoming HTTP traffic to another node or a maintenance page first, and then only close the session in a second step.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
@@ -132,8 +132,7 @@ public class DefaultNettyOptions implements NettyOptions {
             ioShutdownQuietPeriod, ioShutdownTimeout, ioShutdownUnit));
     DefaultPromise<Void> closeFuture = new DefaultPromise<>(GlobalEventExecutor.INSTANCE);
     combiner.finish(closeFuture);
-    // stop the timer as well
-    timer.stop();
+    closeFuture.addListener(f -> timer.stop());
     return closeFuture;
   }
 
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
@@ -212,16 +212,25 @@ public abstract class CqlRequestHandlerBase implements Throttled {
 
   private Timeout scheduleTimeout(Duration timeoutDuration) {
     if (timeoutDuration.toNanos() > 0) {
-      return this.timer.newTimeout(
-          (Timeout timeout1) -> {
-            setFinalError(
-                new DriverTimeoutException("Query timed out after " + timeoutDuration), null, -1);
-          },
-          timeoutDuration.toNanos(),
-          TimeUnit.NANOSECONDS);
-    } else {
-      return null;
+      try {
+        return this.timer.newTimeout(
+            (Timeout timeout1) ->
+                setFinalError(
+                    new DriverTimeoutException("Query timed out after " + timeoutDuration),
+                    null,
+                    -1),
+            timeoutDuration.toNanos(),
+            TimeUnit.NANOSECONDS);
+      } catch (IllegalStateException e) {
+        // If we raced with session shutdown the timer might be closed already, rethrow with a more
+        // explicit message
+        result.completeExceptionally(
+            ("cannot be started once stopped".equals(e.getMessage()))
+                ? new IllegalStateException("Session is closed")
+                : e);
+      }
     }
+    return null;
   }
 
   /**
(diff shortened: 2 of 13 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
```diff
@@ -132,8 +132,7 @@ public void afterChannelInitialized(Channel channel) {
             ioShutdownQuietPeriod, ioShutdownTimeout, ioShutdownUnit));
     DefaultPromise<Void> closeFuture = new DefaultPromise<>(GlobalEventExecutor.INSTANCE);
     combiner.finish(closeFuture);
-    // stop the timer as well
-    timer.stop();
+    closeFuture.addListener(f -> timer.stop());
```
olim7t, line 135: Not directly related to this ticket, but I think it's a good idea to close the timer as late as possible.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
```diff
@@ -216,13 +216,22 @@ public void onThrottleReady(boolean wasDelayed) {
...
+                    null,
+                    -1),
+            timeoutDuration.toNanos(),
+            TimeUnit.NANOSECONDS);
+      } catch (IllegalStateException e) {
+        // If we raced with session shutdown the timer might be closed already, rethrow with a more
+        // explicit message
+        throw ("cannot be started once stopped".equals(e.getMessage()))
+            ? new IllegalStateException("Session is closed")
+            : e;
```
olim7t: Unfortunately there's not better way to detect that the timer was stopped.

We didn't run into this in driver 3, because timeouts were scheduled per node, and always after the initial write.

core/src/main/java/com/datastax/oss/driver/internal/core/session/RequestProcessor.java
```diff
@@ -46,4 +46,7 @@
       DefaultSession session,
       InternalDriverContext context,
       String sessionLogPrefix);
+
+  /** Builds a failed result to directly report the given error. */
+  ResultT newFailure(RuntimeException error);
```
olim7t, line 51: All existing processors need to implement this new method. I initially thought of adding ad-hoc code in every processor, but the advantage of this approach is that we can do things in an abstract manner in `DefaultSession`, so it's a better way to ensure that it will work for every processor.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
emerkle826: I guess the `cannot be started once stopped` message is a Netty implementation thing. It stinks that we have to rely on that and hope it doesn't change.

tomekl007: Yes, and this is an exception message. I don't think that exception message is a part of a contract - so, therefore, it not needs to be backward compatible.
I think that if we want to stick with that implementation, we should write an integration test that is triggering that exception somehow. This way we will have a sanity check that if someone will increment netty, we will have immediate feedback via test failure

adutra (approved):

olim7t:
I force-pushed to rebase on the latest 4.x, addressed a small issue with async execution and added an integration test.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
olim7t: I added `ShutdownIT`, it fails if I replace this line with `throw e`.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/session/ShutdownIT.java
```diff
@@ -0,0 +1,110 @@
...
+          () -> {
+            try {
+              while (!Thread.currentThread().isInterrupted()) {
+                semaphore.acquire();
+                session
+                    .executeAsync(QUERY_STRING)
+                    .whenComplete(
+                        (ignoredResult, error) -> {
+                          semaphore.release();
+                          // Three things can happen:
```
tolbertam, line 84: Thank you for documenting this :)

integration-tests/src/test/java/com/datastax/oss/driver/api/core/session/ShutdownIT.java
```diff
@@ -0,0 +1,110 @@
...
+              }
+            } catch (InterruptedException e) {
+              // return
+            }
+          });
+    }
+    TimeUnit.MILLISECONDS.sleep(100);
+    session.forceCloseAsync();
+    assertThat(gotSessionClosedError.await(1, TimeUnit.SECONDS))
+        .as("Expected to get the 'Session was closed' error shortly after shutting down")
```
tolbertam: Small nit:  Should be 'Session is closed'

tolbertam (approved):
:+1:

tomekl007 (approved):

integration-tests/src/test/java/com/datastax/oss/driver/api/core/session/ShutdownIT.java
```diff
@@ -0,0 +1,110 @@
...
+        .cluster()
+        .prime(when(QUERY_STRING).then(noRows()).delay(20, TimeUnit.MILLISECONDS));
+    CqlSession session = SessionUtils.newSession(simulacronRule);
+
+    // When
+    // Max out the in-flight requests on the connection (from a separate thread pool to get a bit of
+    // contention), then force-close the session abruptly.
+    Set<String> unexpectedErrors = new ConcurrentSkipListSet<>();
+    ExecutorService requestExecutor = Executors.newFixedThreadPool(4);
+    Semaphore semaphore = new Semaphore(1024);
```
tomekl007: isn't that `1024` of some configuration on the server side? If it can change it will break this test - maybe we should extract it somehow?
Or parametrize SimulacronRule with `1024` to have that config explicit?

olim7t: It's from `advanced.connection.max-requests-per-connection` in the driver configuration. I can pull it from there, it will be more explicit.

PR #1143, opened 2018-11-30
Delay query plan computation until throttling ready

olim7t:
Motivation:

If a request is throttled, there will be a delay between the handler's
construction and the time it starts sending requests. So the handler
might operate on a query plan that is slightly out of date. Even though
the delay should be short, and handlers are resilient to stale plans, it
would be better to get the query plan as late as possible.

Modifications:

Build the query plan in onThrottleReady instead of the constructor. It
can't be a final field anymore, so pass it around as a method parameter
instead.

Result:

The query plan is built right before the first call to sendRequest.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
@@ -101,7 +101,6 @@ public abstract class CqlRequestHandlerBase implements Throttled {
   private final DefaultSession session;
   private final CqlIdentifier keyspace;
   private final InternalDriverContext context;
-  private final Queue<Node> queryPlan;
   @NonNull private final DriverExecutionProfile executionProfile;
   private final boolean isIdempotent;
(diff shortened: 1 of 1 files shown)


adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerBase.java
```diff
@@ -211,7 +201,13 @@ public void onThrottleReady(boolean wasDelayed) {
           System.nanoTime() - startTimeNanos,
           TimeUnit.NANOSECONDS);
     }
-    sendRequest(null, 0, 0, true);
+    Queue<Node> queryPlan =
+        this.statement.getNode() != null
```
tomekl007, line 205: it is used from one thread? If not we will need to use AtomicReference or something similar

olim7t: The reference is now passed around as a method parameter instead of being stored as a field, so we don't need additional synchronization.

The implementation itself is thread-safe.

PR #1144, opened 2018-11-30
JAVA-2038 Add jitter to delays between reconnection attempts

emerkle826:
Add random jitter to the reconnect delays for cases when there are multiple clients trying to reconnect at the same time. The implementation adds between -15% to 15% jitter to the pure exponential back-off.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -4,6 +4,7 @@
 
 ### 4.0.0-beta3 (in progress)
 
+- [improvement] JAVA-2038: Add jitter to delays between reconnection attempts
 - [improvement] JAVA-2053: Cache results of session.prepare()
 - [improvement] JAVA-2058: Make programmatic config reloading part of the public API
 - [improvement] JAVA-1943: Fail fast in execute() when the session is closed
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/connection/DecorrelatedJitterReconnectionPolicy.java
```diff
@@ -0,0 +1,103 @@
...
+ * See {@code reference.conf} (in the manual or core driver JAR) for more details.
+ */
+@ThreadSafe
+public class DecorrelatedJitterReconnectionPolicy extends ExponentialReconnectionPolicy {
+
+  private static final Logger LOG =
+      LoggerFactory.getLogger(DecorrelatedJitterReconnectionPolicy.class);
+
+  private final String logPrefix;
+  private final Random jitterRandom = new Random();
```
tomekl007: Isn't there a need to use ThreadLocalRandom?

emerkle826: I have changed the implementation and replaced `Random` with ThreadLocalRandom`. Good catch.

tolbertam (commented):
Nice work! Straightforward and easy to follow :+1:

tolbertam (approved):
Woops, I mean to approve this :)

PR #1145, opened 2018-12-05
 JAVA-2049: Add shorthand method to SessionBuilder to specify local DC

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -145,6 +146,24 @@ public SelfT addContactPoint(@NonNull InetSocketAddress contactPoint) {
...
   }
 
+  /**
+   * Specifies the datacenter that is considered "local" by the load balancing policy.
+   *
+   * <p>This is a programmatic alternative to the configuration option {@code
+   * basic.load-balancing-policy.local-datacenter}. If you pass a non-null value to this method,
+   * then any execution profile that does not declare it (or inherit it from the default profile)
+   * will use it. If the configuration option is specified, then it takes precedence and this value
+   * is ignored.
```
olim7t: Note that this is backwards compared to other options: usually the programmatic value takes precedence over the configuration (see `withKeyspace` for example). We can't do it here because the programmatic value would have to override all profiles; I think we want to keep the ability to have `withLocalDatacenter` for the default profile, but maybe change it in the configuration for a particular child profile.

adutra: I think this is going to be confusing for users. Why can't we keep the idea of "anything non null specified programmatically overrides the corresponding option in the config?" Does the fact that the option is overridable in profiles make this idea impossible?

olim7t: Consider this:
```
basic.load-balancing-policy {
  class = DefaultLoadBalancingPolicy
}
profiles {
  olap {
    basic.load-balancing-policy {
      class = DefaultLoadBalancingPolicy
      local-datacenter = olapdc
    }
  }
}
```
And suppose I want the programmatic option to override the default LBP, but not the one I set explicitly. Admittedly this is contrived.

Another approach just occurred to me: we could add the profile name to the programmatic call: `withLocalDatacenter(String profileName, String localDatacenter)`, like we did for `withNodeFilter`. I like that better actually, it lifts the ambiguity and allows us to preserve the usual order of precedence.

olim7t: I've amended the code to use that last approach.

core/src/main/java/com/datastax/oss/driver/internal/core/context/InternalDriverContext.java
```diff
@@ -114,6 +114,14 @@
   @NonNull
   MetricsFactory getMetricsFactory();
 
+  /**
+   * The value that was passed to {@link SessionBuilder#withLocalDatacenter(String,String)} for this
+   * particular profile. If it was specified through the configuration instead, this method will
+   * return {@code null}.
+   */
+  @Nullable
+  String getLocalDatacenter(String profileName);
```
adutra: Nit: you could annotate `profileName` with `@NonNull`.

adutra (approved):

tolbertam (approved):
:+1:

PR #1146, opened 2018-12-06
 JAVA-2053: Cache results of session.prepare()

olim7t:
3 commits that are meant to stay separate:
* "Remove RequestHandler interface": preliminary refactoring, would be worth considering even if we decided not to merge this ticket.
* JAVA-2053: the ticket
* "Simplify CQL handler hierarchy": not strictly related, but I thought it was nice to apply the same simplification to CQL request processors.

TODO
- [x] upgrade guide
- [x] add cache size metric
- [x] amend tests to verify query string is not sanitized


integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/PreparedStatementIT.java
```diff
@@ -357,4 +357,40 @@ private void should_not_store_metadata_for_conditional_updates(CqlSession sessio
...
+
+    // When
+    PreparedStatement preparedStatement1 = session.prepare(statement.setPageSize(1));
+    PreparedStatement preparedStatement2 = session.prepare(statement.setPageSize(4));
+
+    // Then
+    assertThat(preparedStatement1).isNotSameAs(preparedStatement2);
+    // Each bound statement uses the page size it was prepared with
+    assertThat(firstPageOf(session.executeAsync(preparedStatement1.bind()))).hasSize(1);
+    assertThat(firstPageOf(session.executeAsync(preparedStatement2.bind()))).hasSize(4);
```
olim7t, line 417: This is a very nice side-effect of this ticket: with the previous version of the code, the two calls would return the same prepared statement (using the first page size IIRC).

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
```diff
@@ -21,36 +21,40 @@
...
 public class CqlPrepareAsyncProcessor
     implements RequestProcessor<PrepareRequest, CompletionStage<PreparedStatement>> {
 
-  private final ConcurrentMap<ByteBuffer, DefaultPreparedStatement> preparedStatementsCache;
-
-  public CqlPrepareAsyncProcessor(
-      ConcurrentMap<ByteBuffer, DefaultPreparedStatement> preparedStatementsCache) {
-    this.preparedStatementsCache = preparedStatementsCache;
-  }
+  private final Cache<PrepareRequest, CompletableFuture<PreparedStatement>> cache =
```
tomekl007: Maybe we should add some metrics (gauge) that is tracking the number of elements in this cache and expose it as a driver metric?

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/PreparedStatementIT.java
```diff
@@ -357,4 +357,40 @@ private void should_not_store_metadata_for_conditional_updates(CqlSession sessio
     assertThat(ps.getResultSetDefinitions()).hasSize(0);
     assertThat(Bytes.toHexString(ps.getResultMetadataId())).isEqualTo(Bytes.toHexString(idBefore));
   }
+
+  @Test
+  public void should_return_same_instance_when_repreparing_query() {
+    // Given
+    CqlSession session = sessionRule.session();
+    String query = "SELECT * FROM prepared_statement_test WHERE a = ?";
```
tomekl007, line 371: maybe we should add parametrized test for different kinds of queries (upper case, whitespaces) to test the fact that we are treating those statements as different ones?

olim7t: 👍 will do

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
olim7t: Great idea, this is a good candidate for a metric.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardSessionMetricUpdater.java
```diff
@@ -97,4 +119,18 @@ public String buildFullName(SessionMetric metric, String profileName) {
       return () -> 0;
     }
   }
+
+  @Nullable
+  private static Cache<?, ?> getPreparedStatementCache(InternalDriverContext context) {
+    // By default, both the sync processor and the async one are registered and they share the same
+    // cache. But with a custom processor registry, there could be only one of the two present.
```
tomekl007, line 126: You mean that when the user will implement their own processor it can extend Async OR Sync?

olim7t: More like implement their own session, and only register a subset of the processors. If your application uses exclusively sync/async, you don't need the other processor.
Admittedly this is a bit contrived, having an extra unused processor registered probably has very little noticeable overhead, if any at all.

tomekl007: I wonder if there is a possiblity of some kind of race condition. I mean is the ordering of ifs matters? We are firstly checking `CqlPrepareAsyncProcessor` and then Sync version.

olim7t: I don't see any concurrency issue, this is called at initialization time only. And the two processors should share the same cache anyway, I've added a comment to the sync processor to clarify.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
```diff
@@ -21,36 +21,44 @@
...
+      CacheBuilder.newBuilder().weakValues().build();
```
adutra: What if the user wants to customize the cache? It doesn't seem easy to pass a `DriverContext` when instantiating processors.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
```diff
@@ -21,36 +21,44 @@
...
       PrepareRequest request,
       DefaultSession session,
       InternalDriverContext context,
       String sessionLogPrefix) {
-    return new CqlPrepareAsyncHandler(
-        request, preparedStatementsCache, session, context, sessionLogPrefix);
+
+    try {
+      return cache.get(
+          request,
```
adutra: So this is using `PrepareRequest` as the entry key. I  see that you added equals/hashCode implementations where appropriate, but a quick look into the implementations shows that `DefaultPrepareRequest.equals()` will call `SimpleStatement.equals()`, but `DefaultSimpleStatement.equals()` has a very complex equals method that compares much more than the query string. This could lead to a situation where two non-equal `SimpleStatement` instances sharing the same query string would actually generate two cache entries. Is this a problem?

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
olim7t: I don't think customizing the cache will be a very common need, its size/eviction strategy/etc is mainly driven by how your application uses prepared statements.

On the other hand it doesn't hurt to expose a protected constructor that takes the cache as a parameter, so I'll do that.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
olim7t: It's not a bug, it's a feature 😉 
It does generate separate cache entries, but I see that as an improvement. If you prepare with different execution parameters (e.g. same query but different CL), each prepared statement generates bound statements that honor those.

See also `PreparedStatementIT.should_create_separate_instances_for_different_statement_parameters()` and my [comment](https://github.com/datastax/java-driver/pull/1146#discussion_r239623008) above.

adutra: Ah nice, indeed 👍

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
```diff
@@ -21,22 +21,28 @@
...
     implements RequestProcessor<PrepareRequest, CompletionStage<PreparedStatement>> {
 
-  private final ConcurrentMap<ByteBuffer, DefaultPreparedStatement> preparedStatementsCache;
+  protected final Cache<PrepareRequest, CompletableFuture<PreparedStatement>> cache;
 
-  public CqlPrepareAsyncProcessor(
-      ConcurrentMap<ByteBuffer, DefaultPreparedStatement> preparedStatementsCache) {
-    this.preparedStatementsCache = preparedStatementsCache;
+  public CqlPrepareAsyncProcessor() {
+    this(CacheBuilder.newBuilder().weakValues().build());
```
tolbertam, line 40: I think there may be some value in not using weak values particularly in the case where users are dynamically creating queries.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
```diff
@@ -45,12 +51,22 @@ public boolean canProcess(Request request, GenericType<?> resultType) {
...
       DefaultSession session,
       InternalDriverContext context,
       String sessionLogPrefix) {
-    return new CqlPrepareAsyncHandler(
-        request, preparedStatementsCache, session, context, sessionLogPrefix);
+
+    try {
+      return cache.get(
+          request,
+          () -> new CqlPrepareHandler(request, session, context, sessionLogPrefix).handle());
```
tolbertam: What happens if the result of the prepare is a failed future?  Should we add code that checks for an existing failed result and reprepares?

core/src/main/resources/reference.conf
```diff
@@ -786,6 +786,12 @@ datastax-java-driver {
         # with a DriverTimeoutException (exposed as a Counter).
         // cql-client-timeouts,
 
+        # The size of the driver-side cache of CQL prepared statements.
+        #
+        # The cache uses weak values eviction, so this represents the number of PreparedStatement
```
tolbertam, line 794: It's counterintuitive, but with `weakValues()` when the value is evicted the cache entry can oddly still exist (although is not accessible) and is still factored into `cache.size()`.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/PreparedStatementIT.java
tolbertam: love it.  I guess a possible downside is we'll actually prepare the statement twice with C* nodes...but that doesn't seem so bad..

upgrade_guide/README.md
```diff
@@ -88,6 +88,17 @@ boundSelect = boundSelect.setInt("k", key);
 Note that, as indicated in the previous section, the public API exposes these types as interfaces:
 if for some reason you prefer a mutable implementation, it's possible to write your own.
 
+#### Prepared statement cache
```
tolbertam, line 91: To clarify, it would be good to add 'This is no longer considered an anti-pattern' somewhere.

core/src/main/java/com/datastax/oss/driver/api/core/CqlSession.java
```diff
@@ -123,6 +123,14 @@ default ResultSet execute(@NonNull String query) {
    *
    * If you want to customize this behavior, you can write your own implementation of {@link
    * PrepareRequest} and pass it to {@link #prepare(PrepareRequest)}.
+   *
+   * <p>The result of this method is cached: if you call it twice with the same {@link
+   * SimpleStatement}, you will get the same {@link PreparedStatement} instance. We still recommend
+   * keeping a reference to it (for example by caching it as a field in a DAO); if that's not
+   * possible (e.g. if query strings are generated dynamically), it's OK to call this method every
+   * time: there will just be a small performance overhead to check the internal cache. Note that
+   * caching is based on the query string exactly as you provided it: the driver does not perform
```
tolbertam: We should also indicate that we also cache the statement by more than the query string, but also all statement configuration values.

tolbertam (commented):
Looking good, and I think a lot of users will really appreciate this.  I had a few questions.  I'm also curious about what happens to a cache entry if the prepare future fails, I'll try this out.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
tolbertam: it looks like failed results are cached, and all subsequent prepares for that query would fail.  Test showing that here: https://gist.github.com/tolbertam/4f0581f2fbb3b90f5b79968e3a6f173d

tolbertam:
```suggestion
          () -> {
            CompletableFuture<PreparedStatement> future =
                new CqlPrepareHandler(request, session, context, sessionLogPrefix).handle();

            // invalidate this entry if it failed so subsequent requests may try preparing again.
            future.whenComplete(
                (statement, ex) -> {
                  if (ex != null) {
                    cache.invalidate(request);
                  }
                });
            return future;
          });
```

Would something like this work with regards to [this comment](https://github.com/datastax/java-driver/pull/1146#discussion_r241491133)? Cache entry is invalidated if future fails.  It makes my test pass but i'm not sure if the right fix.  

One possible concern is if somehow the future completes before the item is cached, which I suppose is possible, yet unlikely since the future completes on a different thread.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
olim7t: Why? Weak values give us exactly what you can expect from this cache: as long as the client application is using a prepared statement, it is not evicted.

If we don't use weak values, we have to introduce an eviction strategy, which complicates configuration, opens the door for weird behavior when bad configuration is used, etc.
Note that this is still possible by extending the processor, but for the default I think weak values are fine.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
olim7t: Good catch, but I share your concern about the race condition in your suggested fix: the window is small, but if it ever happens the error is cached forever (or at least until eviction), which is not acceptable. 

We need to be sure that the entry exists before we check the future for failure. How about doing it after every lookup:
```java
      CompletableFuture<PreparedStatement> result =
          cache.get(
              request,
              () -> new CqlPrepareHandler(request, session, context, sessionLogPrefix).handle());
      if (result.isCompletedExceptionally()) {
        cache.invalidate(request);
      }
      return result;
```

core/src/main/resources/reference.conf
olim7t: Yes, it's mentioned somewhere in the Guava docs, I think it's more of an implementation detail. `size()` is documented as the _approximate_ number of entries, I should also mention that in the metric's comment.

olim7t:
Squashed and rebased on latest 4.x.
I've implemented my last suggestion for the cache invalidation.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
tomekl007: If the Future will complete with the exception after the if check, our `isCompletedExceptionally()` will not work properly (and this is very probable):
```
    CompletableFuture<Boolean> result =
        new CompletableFuture<>();

    if (result.isCompletedExceptionally()) {
      System.out.println("Action on ex");
    }

    result.completeExceptionally(new Exception("a"));
```
we would to rely on such order:
```
CompletableFuture<Boolean> result =
        new CompletableFuture<>();
    result.completeExceptionally(new Exception("a"));

    if (result.isCompletedExceptionally()) {
      System.out.println("Action on ex");
    }
```
Maybe we can chain it using `completeExceptionally()` something like:
```
  CompletableFuture<Boolean> result =
        new CompletableFuture<>();

    result.exceptionally(throwable -> {
      System.out.println("on error"); //cache invalidate here
      return true;
    });

    result.completeExceptionally(new Exception("a"));
```
but seems that still, that problem occurs: `if somehow the future completes before the item is cached, `
Maybe in a guava cache loader, we can add a condition that should not cache if CompletableFuture returns false (or exception)?
Or we should not cache CompletableFuture at all - but the actual result? So in a future `onComplete`, we can do `.cache()` for actual value and later execution can wrap the value in an immediate future?

olim7t: Yes, for the initial call the future will probably not be complete yet when the `if` block is reached, it will be detected on the next call to `prepare()`. That sounded reasonable to me but on second thought, if you're doing serial calls from the same thread, it means you have to retry 2 times which is not great.
Maybe we could keep both checks: @tolbertam's `whenComplete` in the cache loader, and this one as a last resort if the future completed before the cache entry was created (which frankly will probably be very rare).

> Or we should not cache CompletableFuture at all - but the actual result?

This would make `prepareAsync()` blocking.

olim7t: Even better: put a standalone future in the cache, and only then call the handler and propagate the result.
```java
CompletableFuture<PreparedStatement> result = cache.getIfPresent(request);
if (result == null) {
  CompletableFuture<PreparedStatement> mine = new CompletableFuture<>();
  result = cache.get(request, () -> mine);
  if (result == mine) {
    new CqlPrepareHandler(request, session, context, sessionLogPrefix)
        .handle()
        .whenComplete(
            (preparedStatement, error) -> {
              if (error != null) {
                mine.completeExceptionally(error);
                cache.invalidate(request); // Make sure failure isn't cached indefinitely
              } else {
                mine.complete(preparedStatement);
              }
            });
  }
}
return result;
```

adutra: Implemented @olim7t suggestion and squashed the commits, keeping 3 separate ones as requested.

tomekl007: because the path `(result != mine)` is not covered it is prone to the same problem as the previous solution, isn't it?

tomekl007: Also, there is possiblity that two/or more threads will trigger compute because of the check `result == null`. Are we ok with that?

adutra: Yes, because only one thread will succeed in putting `mine` in the cache.

tomekl007 (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
olim7t: If `result != mine`, then another thread has beat us to putting its own future in the cache, and triggered the computation. So we discard our future, don't start the computation, and return the other thread's future.

Two or more threads can enter `result == null`, but only one can "win" the cache initialization, i.e. enter the `result == mine` block. There is a race, but the only consequence is that the losing threads have created an extra `mine` future that will be immediately discarded.

PR #1147, opened 2018-12-10
JAVA-2055: Fix intermittent test failures

tolbertam:
For [JAVA-2055](https://datastax-oss.atlassian.net/browse/JAVA-2055).

Change is pretty straightforward.  I suspect this was failing because the LBP hadn't been notified by the time cleanup happened (test only waits for host state to be UP, which will happen before LBP is engaged).   Since the test doesn't create schema, it doesn't really need a dedicated `SessionRule`.  Instead, use `SessionUtils.newSession` to create a standalone session for this specific test.

### DIFF
--- a/test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/simulacron/QueryCounter.java
+++ b/test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/simulacron/QueryCounter.java
@@ -1,0 +1,154 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.testinfra.simulacron;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.datastax.oss.driver.api.testinfra.utils.ConditionChecker;
+import com.datastax.oss.simulacron.common.cluster.QueryLog;
+import com.datastax.oss.simulacron.server.BoundNode;
+import com.datastax.oss.simulacron.server.BoundTopic;
+import com.datastax.oss.simulacron.server.listener.QueryListener;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Predicate;
+
+/**
+ * A convenience utility that keeps track of the number of queries matching a filter received by
+ * {@link BoundNode}s.
+ *
+ * <p>One tricky thing about validating query counts in context of testing is in cases where you
+ * don't require waiting for a node to respond. In this case it's possible that a user would check
+ * count criteria before the node has even processed the message. This class offers the capability
+ * to wait a specified amount of time when asserting query arrival counts on nodes.
+ */
+public class QueryCounter {
+
+  private final long beforeTimeout;
+  private final TimeUnit beforeUnit;
+  private final AtomicInteger totalCount = new AtomicInteger(0);
+  private final ConcurrentHashMap<Integer, Integer> countMap = new ConcurrentHashMap<>();
+
+  private QueryCounter(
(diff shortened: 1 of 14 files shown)


tolbertam:
There are some other intermittent test failures..i'll see if there's any obvious ones I can fix here as well.

tolbertam:
Using this ticket as an excuse to fix a number of integration tests, labeling wip while i fix things up.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/simulacron/QueryCounter.java
```diff
@@ -0,0 +1,154 @@
...
+    public QueryCounterBuilder withFilter(Predicate<QueryLog> queryLogFilter) {
+      this.queryLogFilter = queryLogFilter;
+      return this;
+    }
+
+    /**
+     * Whether or not simulacron should notify before (false) or after (true) the message is
+     * processed.
+     */
+    public QueryCounterBuilder withAfter(boolean after) {
```
olim7t: `withAfter()` and `before()` do completely different things, this is confusing. Maybe we could rename this one `withNotification(NotificationMode)`, using a new enum `NotificationMode { BEFORE_PROCESSING, AFTER_PROCESSING }`.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/simulacron/QueryCounter.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Predicate;
+
+/**
+ * A convenience utility that keeps track of the number of queries matching a filter received by
+ * {@link BoundNode}s.
+ *
+ * <p>One tricky thing about validating query counts in context of testing is in cases where you
+ * don't require waiting for a node to respond. In this case it's possible that a user would check
```
olim7t, line 37: This mostly happens in specex tests, right?

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/simulacron/SimulacronRule.java
```diff
@@ -24,14 +24,22 @@
...
 import com.datastax.oss.simulacron.server.Server;
 import java.net.InetSocketAddress;
+import java.net.SocketAddress;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
 public class SimulacronRule extends CassandraResourceRule {
   // TODO perhaps share server some other way
+  // TODO: Temporarily do not release addresses to ensure IPs are always ordered
```
olim7t, line 34: Do you suspect that some of our tests expect this?

olim7t (approved):
Thanks for jumping on this, and addressing other issues. Everything looks good to me so far, I'll take another look if you add more commits.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/simulacron/SimulacronRule.java
tolbertam: I suspect it could be a problem in tests that use `SortingLoadBalancingPolicy` and then using `simulacronCluster.node(X)` with the expectation that nodes are ordered within a simulacron cluster by IP address, which may not be the case if simulacron uses reuses IP addresses.  Something i need to look into more since I thought this was a cause of some specex test errors, but apparently isn't.

tolbertam: It looks like the source of the problem was that `buildSessionWithProfile` did not use `SortingLoadBalancingPolicy`, so order was never enforced, and somehow this didn't happen to be a problem most of the time.  I found it odd that only profile-based tests were failing, this explains it! (342b695)

tolbertam:
Have worked out a lot of little test issues here. I set up a schedule to rerun CI every 20 minutes, will see how this goes in the morning..see if there are any other tests to fix up, and will keep at it until things run reliably.

tolbertam:
Merging this today as there are a bunch of test fixes which make our builds pass much more reliably.

There are still two test issues that pop up now and then:

* `com.datastax.oss.driver.internal.core.pool.ChannelPoolResizeTest` unit tests rarely fail, looks timing oriented, but looking at the test it looks like that is accounted for...needs more research. [JAVA-2059](https://datastax-oss.atlassian.net/browse/JAVA-2059)

* `com.datastax.oss.driver.api.core.metadata.NodeStateIT.should_mark_control_node_down_when_control_connection_is_last_connection_and_dies` waiting for an up event.  Needs more investigation, but is very infrequent. [JAVA-2060](https://datastax-oss.atlassian.net/browse/JAVA-2059)

And of course there are even rarer occasions where for whatever reason CCM has troubles launching a cluster.  I'd like to see if we can rerun tests in these cases as it's not he drivers' fault.

I'll continue to investigate these when time permits.

PR #1148, opened 2018-12-12
JAVA-2070: Call onRemove instead of onDown when rack and/or DC information changes for a host

exploder86:
LoadBalancingPolicy.java defines a pair of consistent methods onAdd/onRemove and onUp/onDown. I guess a ControlConnection.java is broke this contract.


tolbertam:
Thanks for bringing this to our attention @exploder86.  That definitely does look wrong and this looks like a straightforward fix, we'll look into getting this in our next release, thanks!

tolbertam:
Created [JAVA-2070](https://datastax-oss.atlassian.net/browse/JAVA-2070) for tracking.

tolbertam (approved):
:100:!

PR #1149, opened 2018-12-12
JAVA-2057: Do not create pool when SUGGEST_UP topology event received

tolbertam:
For [JAVA-2057](https://datastax-oss.atlassian.net/browse/JAVA-2057).

This should fix an issue I noticed on our CI servers, were occasionally after forcing a node down I observed that there were still some open connections.   I have CI running repeatedly against #1147, which has this fix, so assuming that continues to run well we should consider merging this.  Lets wait overnight to see how things go.

Old description (see 41dec29 for actual change)

Motivation:

There is a small window where a NEW_NODE event may be sent over the
control connection while a pool is initializing.  This is more likely to
happen in testing, where a cluster is brought up and the driver connects
to it immediately.

In this case, the driver would erroneously create another pool, which
would create additional connections.  The first pool created between
the currently initializing one and the new one being created would be
untracked with its connections remanining open even after the Session is
closed.

Modifications:

Akin to the logic in processDistanceEvent and processStateEvent, only
invoke createOrReconnectPool in onTopologyEvent when there is not an
initializing pool future present in the pending map.

Result:

createOrReconnectPool is no longer invoked in onTopologyEvent if there
is a pending initialization for that node.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/session/PoolManager.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/session/PoolManager.java
@@ -338,9 +338,17 @@ public class PoolManager implements AsyncAutoCloseable {
       if (event.type == TopologyEvent.Type.SUGGEST_UP) {
         Node node = context.getMetadataManager().getMetadata().getNodes().get(event.address);
         if (node.getDistance() != NodeDistance.IGNORED) {
-          LOG.debug(
-              "[{}] Received a SUGGEST_UP event for {}, reconnecting pool now", logPrefix, node);
-          createOrReconnectPool(node);
+          if (pending.containsKey(node)) {
+            LOG.debug(
+                "[{}] Received a SUGGEST_UP event for {} while pool was already initializing, "
+                    + "ignoring",
+                logPrefix,
+                node);
+          } else {
+            LOG.debug(
+                "[{}] Received a SUGGEST_UP event for {}, reconnecting pool now", logPrefix, node);
+            createOrReconnectPool(node);
+          }
         }
       }
     }
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/session/PoolManager.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/session/PoolManager.java
@@ -338,16 +338,11 @@ public class PoolManager implements AsyncAutoCloseable {
       if (event.type == TopologyEvent.Type.SUGGEST_UP) {
         Node node = context.getMetadataManager().getMetadata().getNodes().get(event.address);
         if (node.getDistance() != NodeDistance.IGNORED) {
-          if (pending.containsKey(node)) {
-            LOG.debug(
-                "[{}] Received a SUGGEST_UP event for {} while pool was already initializing, "
-                    + "ignoring",
-                logPrefix,
-                node);
-          } else {
-            LOG.debug(
-                "[{}] Received a SUGGEST_UP event for {}, reconnecting pool now", logPrefix, node);
-            createOrReconnectPool(node);
+          LOG.debug(
+              "[{}] Received a SUGGEST_UP event for {}, reconnecting pool now", logPrefix, node);
+          ChannelPool pool = pools.get(node);
+          if (pool != null) {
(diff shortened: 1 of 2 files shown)


tolbertam:
Ack!  This may not completely fix the issue..looks like the issue has recurred even with this fix, so I have more investigation to do.

core/src/main/java/com/datastax/oss/driver/internal/core/session/PoolManager.java
```diff
@@ -338,9 +338,17 @@ private void onTopologyEvent(TopologyEvent event) {
...
+            LOG.debug(
+                "[{}] Received a SUGGEST_UP event for {} while pool was already initializing, "
+                    + "ignoring",
+                logPrefix,
+                node);
+          } else {
+            LOG.debug(
+                "[{}] Received a SUGGEST_UP event for {}, reconnecting pool now", logPrefix, node);
+            createOrReconnectPool(node);
+          }
```
olim7t, line 346: So the goal here is to handle this case (from `TopologyEvent.suggestUp()` javadocs):
> if the driver was currently reconnecting to the node, this causes the current ReconnectionSchedule to be reset, and the next reconnection attempt to happen immediately.

The bug is that it calls a method that _also creates the pool_ if it doesn't exist. We don't want to create, only force an immediate reconnect if it exists. So I would just replace the call to `createOrReconnectPool` with the relevant portion of its body:
```java
ChannelPool pool = pools.get(node);
if (pool != null) {
  pool.reconnectNow();
}
```
Once you do that, you don't really need to check `pending` anymore, because when a pool is still pending (i.e. initializing), it's not in `pools` yet.

PS: there's a small race condition because the pool could start reconnecting immediately after it's initialized, before we put it in `pools` (see `onPoolReady`). If a `SUGGEST_UP` event also arrives exactly at this moment, we could miss it and not trigger a reconnection. But it doesn't really matter because regular scheduled reconnections will eventually kick in, and it's the beginning of the schedule so the interval should be pretty short.

tolbertam: yeah, that seems like a better solution and more resilient to problems in the future if other methods call `createOrReconnectPool` without first checking pending.

tolbertam: Ah I misread, you are suggesting to just replace the call to `createOrReplacePool` with simply calling `pool.reconnectNow()`, got it.

tolbertam:
> Ack! This may not completely fix the issue..looks like the issue has recurred even with this fix, so I have more investigation to do.

The problem was we never put the pool in `pending` in the typical initialization case.  Doesn't that seem like a problem too?  In the case of `processDistanceEvent` / `processStateEvent` being received during pool initialization, couldn't that be a problem too?

tolbertam:
I see, I don't think that will be a problem...we probably need a pool for those distance / state events to even happen?  Although I do find it confusing that we have a `pending` map that doesn't consider pool initialization.

tolbertam:
Made the recommended adjustment and rephrased the PR.  If this is good I will also alter the CHANGELOG & commit log on merge

olim7t:
> In the case of processDistanceEvent / processStateEvent being received during pool initialization, couldn't that be a problem too?

Those events are coalesced and recorded in `pendingDistanceEvents`/`pendingStateEvents`, and replayed in `onPoolReady()` once the pool is done initializing.

olim7t:
Yeah your change is what I had in mind. Does that resolve all issues in CI?

tolbertam:
> Yeah your change is what I had in mind. Does that resolve all issues in CI?

It looks like it!   Have had enough consecutive runs without the error reproducing to feel comfortable with this.   I'm going to push a commit that will update change log and have the final full commit message to reflect what was done here.  Cool if I merge afterwords and also merge in #1147 after?

olim7t:
Works for me 👍 
I assume you plan to clean up the history in #1147, looks like you reverted a few things along the way.

tolbertam:
Definitely will clean that up, it's a bit of a mess at the moment :smile:

PR #1150, opened 2018-12-12
JAVA-2058: Make programmatic config reloading part of the public API

olim7t:
Motivation:

Until now this was only possible via the internal API (by firing an
event on the bus). However this not such an exotic use case, it sounds
reasonable to expect that it will be commonly needed.

Modifications:

- add DriverConfigLoader.reload() and supportsReloading().
- remove ForceReloadConfigEvent, now obsolete since that was the only
  reason for its existence.

Result:

DriverConfigLoader is part of the public API and accessible through
DriverContext, so this does not require any internal imports anymore.

TODO:
- [x] update manual


core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
```diff
@@ -123,10 +138,10 @@ public static DefaultDriverConfigLoaderBuilder builder() {
     private final EventExecutor adminExecutor;
     private final EventBus eventBus;
     private final DriverExecutionProfile config;
-    private final Object forceLoadListenerKey;
 
     private Duration reloadInterval;
-    private ScheduledFuture<?> reloadFuture;
+    private ScheduledFuture<?> periodicTaskHandle;
+    private CompletionStage<DriverConfig> currentReloadFuture;
```
tomekl007: unused

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
```diff
@@ -140,47 +155,59 @@ private SingleThreaded(InternalDriverContext context) {
...
       }
-      if (driverConfig.reload(configSupplier.get())) {
-        LOG.info("[{}] Detected a configuration change", logPrefix);
-        eventBus.fire(ConfigChangeEvent.INSTANCE);
-        Duration newReloadInterval = config.getDuration(DefaultDriverOption.CONFIG_RELOAD_INTERVAL);
-        if (!newReloadInterval.equals(reloadInterval)) {
-          reloadInterval = newReloadInterval;
-          scheduleReloadTask();
+      try {
+        if (driverConfig.reload(configSupplier.get())) {
```
tomekl007: I think that this method `dirverConfig.reload` could be named to smth like: `hasConfigChanged()` `checkIfConfigChange`. Currently, the name `reload()` is not telling the caller that it is checking if only change. From the name, I would guess that it triggers reload

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
```diff
@@ -140,47 +155,59 @@ private SingleThreaded(InternalDriverContext context) {
...
+          LOG.info("[{}] Detected a configuration change", logPrefix);
+          eventBus.fire(ConfigChangeEvent.INSTANCE);
```
tomekl007, line 195: could you tell me where that signal is used on the consumer side? I was not able to find that

core/src/test/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoaderTest.java
```diff
@@ -128,11 +129,11 @@ public void should_reload_if_forced() {
 
     configSource.set("int1 = 43");
 
-    eventBus.fire(ForceReloadConfigEvent.INSTANCE);
+    CompletionStage<Void> reloaded = loader.reload();
     adminExecutor.waitForNonScheduledTasks();
 
     assertThat(initialConfig).hasIntOption(MockOptions.INT1, 43);
-    Mockito.verify(eventBus).fire(ConfigChangeEvent.INSTANCE);
```
tomekl007: `reload()` is still `fire()` that event, so why we removed that check?

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
```diff
@@ -140,47 +155,59 @@ private SingleThreaded(InternalDriverContext context) {
...
       }
       if (reloadInterval.isZero()) {
         LOG.debug("[{}] Reload interval is 0, disabling periodic reloading", logPrefix);
       } else {
         LOG.debug("[{}] Scheduling periodic reloading with interval {}", logPrefix, reloadInterval);
-        reloadFuture =
+        periodicTaskHandle =
             adminExecutor.scheduleAtFixedRate(
-                this::reload,
+                () -> reload(null),
```
tomekl007: I don't understand what is an intention of passing `null` here. Is it for triggering reload without returning value to the user? If so I am suggesting to create some constant marker with `null` assigned to it. Like `TRIGGER_RELOAD_WITHOUT_RETURNIG_RESULT` and use that constant in the code. It will make the intention clean and signal that null here has a special meaning.

core/src/test/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoaderTest.java
olim7t: Right, I was thinking of the other event that I removed, this one is indeed still fired.
Will revert 👍

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
olim7t: It also reloads actually: `TypesafeDriverConfig` is mutable, and if the configuration has changed it will reflect the new values. I feel like `hasConfigChanged()` would make it sound like it only checks.
Do you think adding more explanations in the method's javadocs would be sufficient? Or maybe also rename the method to `updateWith(Config)`.

tomekl007: Docs is ok, when I looked into it the behavior was clear.
Maybe `relaoadAndCheckIfChanged()`?

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
olim7t: `MetadataManager.onConfigChanged()` and `ChannelPool.onConfigChanged()`. It's also intended for future 3rd-party components.

The trick is that `eventBus.register()` takes a `Class`, not an object, so it does not reference `INSTANCE`. I see how it can be deceptive so I'll add a comment in the enum.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
olim7t: The method is now used in two cases:
* manual reload (added by this PR): we want to let the client application know when it's done, hence the `CompletionStage`.
* periodic reload (what was already there): happens in the background so nobody's listening directly, no need to create an unnecessary object.

So yeah `null` is just a simple way to say "no stage to complete", this is internal code so we don't need to be super fancy.
On the other hand this is not on the hot path so I wouldn't mind using `Optional`, would that make the code easier to read?

tomekl007: I don't think that Optional will make it cleaner. I am suggesting just adding constant and assigning `null` or `Optional` to it will be fine

olim7t: I added a shortcut to make the method reference more explicit, and a bit of documentation.

As for extracting a constant,  I don't think it improves things here because what we want to express is the absence of a reference, and this is precisely what `null` means (whereas if it's a constant it's not clear that you should never dereference it). Besides this is internal code, and contained to this short class, so it's pretty trivial.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
olim7t: That's a bit of a mouthful... The problem if we try to express everything the method does is that this is endless, for example it also logs a warning if a profile was added at runtime.

I think `reload` will be fine. If there's any doubt as to what the result represents, the javadoc is only what shortcut/jump away.

tomekl007 (approved):

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -33,6 +39,25 @@
...
+   * @return a stage that completes once the attempt is finished. If a configuration change was
+   *     detected, it's also guaranteed that internal driver components have been notified by that
+   *     time; note however that some react to the notification asynchronously, so they may not have
+   *     completely applied all resulting changes yet. If this loader does not support programmatic
+   *     reloading &mdash; which you can check by calling {@link #supportsReloading()} before this
+   *     method &mdash; the stage will fail immediately with an {@link
+   *     UnsupportedOperationException}.
+   */
+  @NonNull
+  CompletionStage<Void> reload();
```
tolbertam: Would there be value in enclosing a `Boolean` value indicating whether or not the reload observed config changes, similar to how `TypeSafeDriverConfig#reload` returns a boolean if the config changed?  This may be a more direct way of understanding if the config has changed from a manual reload as opposed to registering for `ConfigChangeEvent`

tolbertam (approved):
Looks great.  Had a small suggestion, but not required if you don't think it is useful.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
olim7t: Yes, that makes sense, I will add it.

tolbertam: changes look good :+1:

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -49,6 +49,7 @@
    *     changes yet. If this loader does not support programmatic reloading &mdash; which you can
    *     check by calling {@link #supportsReloading()} before this method &mdash; the stage will
    *     fail immediately with an {@link UnsupportedOperationException}.
+   * @throws UnsupportedOperationException if this loader does not support programmatic reloading.
```
olim7t: My idea was to returned a failed `CompletionStage`, not throw the exception directly.

PR #1151, opened 2018-12-13
JAVA-2056: Reduce HashedWheelTimer tick duration

emerkle826:
Motivation:
Timeouts and speculative executions are scheduled on the
same HashedWheelTimer, which has a default tick duration
of 100ms. This can be too imprecise for speculative executions.
Lowering the tick duration to 1ms does not affect performance,
but does allow for more precision for scheduling speculative
executions.

Modifications:
Reduce the default timer tick duration to 1ms (from 100ms)
and increase the default wheel size to 2048 (from 512).

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -1385,7 +1385,7 @@ datastax-java-driver {
       # Required: yes
       # Modifiable at runtime: no
       # Overridable in a profile: no
-      tick-duration = 100 milliseconds
+      tick-duration = 1 millisecond
 
       # Number of ticks in a Timer wheel. The underlying implementation uses Netty's
       # HashedWheelTimer, which uses hashes to arrange the timeouts. This effectively controls the
@@ -1394,7 +1394,7 @@ datastax-java-driver {
       # Required: yes
       # Modifiable at runtime: no
       # Overridable in a profile: no
-      ticks-per-wheel = 512
+      ticks-per-wheel = 2048
     }
   }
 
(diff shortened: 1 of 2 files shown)


core/src/main/resources/reference.conf
```diff
@@ -1385,7 +1385,7 @@ datastax-java-driver {
       # Required: yes
       # Modifiable at runtime: no
       # Overridable in a profile: no
-      tick-duration = 100 milliseconds
+      tick-duration = 1 milliseconds
```
olim7t: Small nit: Config also support the singular form, so `1 millisecond` is also valid and looks prettier 😃

olim7t (approved):

core/src/main/resources/reference.conf
emerkle826: I completely overlooked that, good catch.

tolbertam (approved):
:+1:

core/src/main/resources/reference.conf
```diff
@@ -1394,7 +1394,7 @@ datastax-java-driver {
       # Required: yes
       # Modifiable at runtime: no
       # Overridable in a profile: no
-      ticks-per-wheel = 512
+      ticks-per-wheel = 2048
```
tomekl007, line 1397: This is basically a number of buckets within a hash table (that is inside of the `HashedWheelTImer`), right?

olim7t: Yes.
2048 seems like a good default to minimize collisions, considering that out of the box the driver uses a request timeout of 2000ms and no speculative executions.

PR #1152, opened 2018-12-14
JAVA-2061: Add section to upgrade guide about updated type mappings

tolbertam:
For [JAVA-2061](https://datastax-oss.atlassian.net/browse/JAVA-2061).  We received feedback that it wasn't apparent what happened to existing methods like `getTimestamp`.  #1123 adds some nice documentation about that, to make things more easily apparent, I added a section in the upgrade guide that references the new doc page and also links to the type mappings.


upgrade_guide/README.md
```diff
@@ -113,6 +113,15 @@ programming][4.x async programming] and [paging][4.x paging].
 [4.x async programming]: http://docs.datastax.com/en/developer/java-driver/4.0/manual/core/async/
 [4.x paging]: http://docs.datastax.com/en/developer/java-driver/4.0/manual/core/paging/
 
+#### CQL to Java type mappings
+
+Since the driver now has access to Java 8 types, some of the [CQL to Java type
+mappings] have changed, particularly when it comes to [temporal types] such
+as `date` and `timestamp`.
```
adutra: I think it would be even more helpful if you mentioned what changed exactly: `getDate` replaced with `getInstant` and `getLocalDate` that now returns `java.time.LocalDate`.

adutra (changes requested):

upgrade_guide/README.md
tolbertam: Sure, I can add that.  I think there are a few changes beyond that.  I'll add a listing of those similar to what was done for the 3.0.0 section of the upgrade guide.

tolbertam: Done

olim7t (approved):
You can also link to `manual/core/#cql-to-java-type-mapping`.

adutra (approved):

PR #1154, opened 2018-12-17
Test fixes for 4.0.0-beta3

tolbertam:
These changes are mostly for testing against DSE using the dse driver, as there were some assumptions that aren't necessarily true upstream.

There is also an ancillary change in`DropwizardSessionMetricUpdater` to use `gauge` instead of `register`.

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/api/core/session/ShutdownIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/api/core/session/ShutdownIT.java
@@ -105,7 +105,7 @@ public class ShutdownIT {
                             assertThat(anfe.getErrors()).hasSize(1);
                             error = anfe.getErrors().values().iterator().next();
                             if (!(error instanceof IllegalStateException)
-                                && !"Driver channel is closing".equals(error.getMessage())) {
+                                && !error.getMessage().endsWith("is closing")) {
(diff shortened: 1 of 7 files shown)


integration-tests/src/test/java/com/datastax/oss/driver/api/core/session/ShutdownIT.java
```diff
@@ -81,21 +82,32 @@ public void should_fail_requests_when_session_is_closed() throws Exception {
...
                           } else if (error != null
                               && !(error instanceof ClosedConnectionException
                                   || error instanceof NoNodeAvailableException)) {
                             unexpectedErrors.add(error.toString());
+                          } else if (error instanceof AllNodesFailedException) {
+                            AllNodesFailedException anfe = (AllNodesFailedException) error;
+                            assertThat(anfe.getErrors()).hasSize(1);
+                            error = anfe.getErrors().values().iterator().next();
+                            if (!(error instanceof IllegalStateException)
+                                && !error.getMessage().endsWith("is closing")) {
```
tolbertam: Could have changed the exception messages to be consistent, but it wasn't clear to me if "Driver channel is closing" in `DriverChannel` and "Channel is closing" in `InflightHandler` was differentiated for a reason.

adutra (approved):

PR #1155, opened 2018-12-17
JAVA-2066: Array index range error when fetching routing keys on bound statements

GregBestland:
I think this will solve the problem. We need to add some tests specifically of unordered partition keys still


adutra:
Added a small integration test and a changelog entry. The test fails without the fix and passes with it.

adutra (approved):

olim7t (commented):
👍

PR #1159, opened 2018-12-20
JAVA-2075: Document preference for LZ4 over Snappy (4.x)

tolbertam:


manual/core/compression/README.md
```diff
@@ -19,8 +19,11 @@ Compression must be set before opening a session, it cannot be changed at runtim
 
 
 Two algorithms are supported out of the box: [LZ4](https://github.com/jpountz/lz4-java) and
-[Snappy](http://google.github.io/snappy/). Both rely on third-party libraries, declared by the
-driver as *optional* dependencies; if you enable compression, you need to explicitly depend on the
+[Snappy](http://google.github.io/snappy/). The LZ4 implementation is a good
+first choice; it offers fallback implementations in case native libraries fail
+to load and benchmarks suggest it offers better performance and compression
+ratios over Snappy. Both rely on third-party libraries, declared by the driver
+as *optional* dependencies; if you enable compression, you need to explicitly depend on the
```
tomekl007: Maybe we can do similar perf tests to those: `http://java-performance.info/performance-general-compression/`
and point users to our perf tests results? If the user will see charts with a performance comparison of our driver using different compression algorithms, it will give the users actual insight why LZ4 is better

tolbertam: I think it would be adequate to just refer to that blog post for further reading, i'll add that link.   My intent here is just to make it easy for users to make a choice without having to think about it much.  At this point there isn't really a reason to adopt Snappy over LZ4 unless you have a really good reason, so LZ4 should be the direction we point people in.

changelog/README.md
```diff
@@ -2,6 +2,10 @@
 
 <!-- Note: contrary to 3.x, insert new entries *first* in their section -->
 
+### 4.0.0-rc1 (In progress)
+
+- [documentation] JAVA-2075: Document preference for LZ4 over Snappy
```
adutra, line 7: I think in the past we usually didn't include documentation tasks in the changelog, but I'm OK with including them from now on.

manual/core/compression/README.md
```diff
@@ -19,8 +19,13 @@ Compression must be set before opening a session, it cannot be changed at runtim
 
 
 Two algorithms are supported out of the box: [LZ4](https://github.com/jpountz/lz4-java) and
-[Snappy](http://google.github.io/snappy/). Both rely on third-party libraries, declared by the
-driver as *optional* dependencies; if you enable compression, you need to explicitly depend on the
+[Snappy](http://google.github.io/snappy/). The LZ4 implementation is a good
+first choice; it offers fallback implementations in case native libraries fail
+to load and
+[benchmarks](http://java-performance.info/performance-general-compression/)
```
tomekl007, line 25: I have one concern about that link. I opened it multiple times (5 times) and 2 out of it I got `error establishing a connection`. I don't know if we want to add a link to an unreliable page?

adutra (approved):

manual/core/compression/README.md
adutra: I'm getting this error too.

tolbertam:
Committed directly by 5ad20ca

olim7t:
Sorry I completely missed this. The JIRA ticket was still in "to do" and it didn't occur to me to check open pull requests.
On the bright side, my commit is almost identical 😃

PR #1160, opened 2018-12-21
JAVA-2063: Abort reconnect attempts when every error is an authentication error.

GregBestland:
We shouldn't continue trying to re-connect when all contact points are giving us auth errors.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
@@ -277,7 +277,6 @@ public class ControlConnection implements EventCallback, AsyncAutoCloseable {
             error -> {
               boolean authFailure = true;
               if (error instanceof AllNodesFailedException) {
-
                 Collection<Throwable> errors =
                     ((AllNodesFailedException) error).getErrors().values();
                 for (Throwable nodeError : errors) {
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
@@ -17,6 +17,7 @@ package com.datastax.oss.driver.internal.core.control;
 
 import com.datastax.oss.driver.api.core.AllNodesFailedException;
 import com.datastax.oss.driver.api.core.AsyncAutoCloseable;
+import com.datastax.oss.driver.api.core.auth.AuthenticationException;
 import com.datastax.oss.driver.api.core.connection.ReconnectionPolicy;
 import com.datastax.oss.driver.api.core.loadbalancing.NodeDistance;
 import com.datastax.oss.driver.api.core.metadata.Node;
@@ -45,6 +46,7 @@ import com.datastax.oss.protocol.internal.response.event.TopologyChangeEvent;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import io.netty.util.concurrent.EventExecutor;
 import java.net.InetSocketAddress;
+import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Queue;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
```diff
@@ -273,12 +275,28 @@ private void init(boolean listenToClusterEvents, boolean reconnectOnFailure) {
               firstConnectionAttemptFuture.complete(null);
             },
             error -> {
-              if (reconnectOnFailure && !closeWasCalled) {
+              boolean authFailure = true;
```
tomekl007, line 278: we are assuming upfront that error is an `authFailure`? 
I think that `if (error instanceof AllNodesFailedException) {}` is `false` it does not mean that it is an `authFailure`, or is it?

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
```diff
@@ -273,12 +275,28 @@ private void init(boolean listenToClusterEvents, boolean reconnectOnFailure) {
...
+              if (error instanceof AllNodesFailedException) {
+                Collection<Throwable> errors =
+                    ((AllNodesFailedException) error).getErrors().values();
+                for (Throwable nodeError : errors) {
+                  if (!(nodeError instanceof AuthenticationException)) {
+                    authFailure = false;
```
adutra, line 284: You could insert a `break` here. Actually, even better: you could extract this logic into a separate method.

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
```diff
@@ -273,12 +275,28 @@ private void init(boolean listenToClusterEvents, boolean reconnectOnFailure) {
               firstConnectionAttemptFuture.complete(null);
             },
             error -> {
-              if (reconnectOnFailure && !closeWasCalled) {
+              boolean authFailure = true;
+              if (error instanceof AllNodesFailedException) {
```
adutra, line 279: Shouldn't we try to generify a bit this logic? Other errors are probably as fatal as authentication errors, e.g. wrong SSL certificates. I suggest creating a method called `isFatalError(Throwable)` and checking all possible cases inside it.

adutra (changes requested):

olim7t:
As I commented in JIRA, retrying on auth errors was [explicitly requested](https://datastax-oss.atlassian.net/browse/JAVA-472) in driver 3:
> When a password was inadvertently changed on the server [the driver stopped reconnecting to that node] ... the client side never recovers until the [driver instance] is bounced. The password can easily be corrected and does not need a node restart. This has the potential to bring the production down if some one fat fingers a password

This could easily apply to the `reconnect-on-init` scenario as well: you start your client, find out that the password on the server-side is wrong, and change it without restarting the client.

That being said, it's debatable and there probably are/will be differing opinions. So I suggest a change to the `ReconnectionPolicy` API in order to let it decide whether the error is "fatal":

* pass a `Throwable` to `ReconnectionSchedule.nextDelay()`: for the control connection it's an `AllNodesFailedException`, for individual nodes we'll need a similar way to aggregate all channel failures. The method can return a special value (`null`? `Optional.empty()`? rethrow the error?) to indicate that the error should not be retried.

    "Not retrying" means different things depending on the situation:

    * control connection if `reconnect-on-init` is set and this is the initial connection attempt: rethrow to the caller.
    * control connection when reconnecting at runtime: stop reconnecting. I can't think of a good reason to do that so this will probably rarely be used. There is an internal method to restart manually: `ControlConnection.reconnectNow`.
    * individual nodes: if there are no active channels, mark the node DOWN and stop reconnecting. This is already hard-coded for some specific cases in `ChannelPool.onAllConnected()`.
 
* I'd also use the opportunity to add a parameter `ReconnectionPolicy.newControlConnectionSchedule(boolean isInitialConnection)`, since the schedule for the initial attempt may behave differently than the "runtime" ones. EDIT -- I opened #1162 to address this separately.

This is much more involved than the original scope of this ticket, and will break the public API 😐 On the flipside, I think it makes sense to give users control over this.

GregBestland:
Java-2063, PR has been migrated to the dse driver repo, and evolved significantly. I'm closing this because it's not longer valid.

PR #1162, opened 2018-12-27
JAVA-2077: Allow reconnection policy to detect first connection attempt

olim7t:
TODO
- [x] add more explanations in commit message body (see JIRA)
- [x] add revapi exception
- [x] refactor init methods
- [ ] reformat `newControlConnectionSchedule` javadoc

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/connection/ReconnectionPolicy.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/connection/ReconnectionPolicy.java
@@ -36,8 +36,10 @@ import java.time.Duration;
  *       for a node does not have its configured number of connections (see {@code
  *       advanced.connection.pool.*.size} in the configuration), a reconnection starts for that
  *       pool.
- *   <li>{@linkplain #newControlConnectionSchedule() for the control connection}: when the control
- *       node goes down, a reconnection starts to find another node to replace it.
+ *   <li>{@linkplain #newControlConnectionSchedule(boolean) for the control connection}: when the
+ *       control node goes down, a reconnection starts to find another node to replace it. This is
+ *       also used if the configuration option {@code advanced.reconnect-on-init} is set and the
+ *       driver has to retry the initial connection.
  * </ul>
  *
  * This interface defines separate methods for those two cases, but implementations are free to
@@ -49,9 +51,21 @@ public interface ReconnectionPolicy extends AutoCloseable {
   @NonNull
   ReconnectionSchedule newNodeSchedule(@NonNull Node node);
 
-  /** Creates a new schedule for the control connection. */
+  /**
+   * Creates a new schedule for the control connection.
+   *
+   * @param isInitialConnection whether this schedule is generated for the driver's initial attempt
+   *     to connect to the cluster.
+   *     <ul>
+   *       <li>{@code true} means that the configuration option {@code advanced.reconnect-on-init}
+   *           is set, the driver failed to reach any contact point, and it is now scheduling
+   *           reattempts.
+   *       <li>{@code false} means that the driver was already initialized, lost connection to the
+   *           control node, and is now scheduling attempts to connect to another node.
+   *     </ul>
+   */
   @NonNull
-  ReconnectionSchedule newControlConnectionSchedule();
+  ReconnectionSchedule newControlConnectionSchedule(boolean isInitialConnection);
 
   /** Called when the cluster that this policy is associated with closes. */
   @Override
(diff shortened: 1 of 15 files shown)


adutra (changes requested):
A revapi exception should be added:

```

{
  "code": "java.method.numberOfParametersChanged",
  "old": "method com.datastax.oss.driver.api.core.connection.ReconnectionPolicy.ReconnectionSchedule com.datastax.oss.driver.api.core.connection.ReconnectionPolicy::newControlConnectionSchedule()",
  "new": "method com.datastax.oss.driver.api.core.connection.ReconnectionPolicy.ReconnectionSchedule com.datastax.oss.driver.api.core.connection.ReconnectionPolicy::newControlConnectionSchedule(boolean)",
  "package": "com.datastax.oss.driver.api.core.connection",
  "classQualifiedName": "com.datastax.oss.driver.api.core.connection.ReconnectionPolicy",
  "classSimpleName": "ReconnectionPolicy",
  "methodName": "newControlConnectionSchedule",
  "oldArchive": "com.datastax.oss:java-driver-core:jar:4.0.0-beta3",
  "newArchive": "com.datastax.oss:java-driver-core:jar:4.0.0-rc1-SNAPSHOT",
  "elementKind": "method",
  "justification": <<<<< ADD YOUR EXPLANATION FOR THE NECESSITY OF THIS CHANGE >>>>>
},
```

core/src/main/java/com/datastax/oss/driver/api/core/connection/ReconnectionPolicy.java
```diff
@@ -49,9 +51,21 @@
   @NonNull
   ReconnectionSchedule newNodeSchedule(@NonNull Node node);
 
-  /** Creates a new schedule for the control connection. */
+  /**
+   * Creates a new schedule for the control connection.
+   *
+   * @param isInitialConnection whether this schedule is generated for the driver's initial attempt
+   *     to connect to the cluster.
+   *     <ul>
```
adutra, line 59: I think we should avoid long texts like this one inside javadoc tags. With the default CSS style, tag texts are rendered with a fixed-size font and all the formatting is lost:

image

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
```diff
@@ -294,6 +293,15 @@ private void init(boolean listenToClusterEvents, boolean reconnectOnFailure) {
       }
     }
 
+    private ReconnectionSchedule newSchedule() {
+      boolean isInitialConnection =
+          reconnectOnInit
+              // This is always set after the initial attempt, regardless of its outcome.
+              // If the first attempt failed, we create the schedule before setting it.
+              && !firstConnectionAttemptFuture.isDone();
```
adutra: This is overly complex to follow, couldn't you create a dedicated boolean variable to store that information (initial/not initial) instead?

Besides, are you sure that `firstConnectionAttemptFuture.isDone()` is always false for the initial attempt? From what I understand, in case the initial attempt fails, the thread executing the initial sequence will first call `reconnection.start()` line 276 then complete `firstConnectionAttemptFuture` exceptionally line 289; but since the reconnection process is async, the future may or may not be completed before `newSchedule` is called.

adutra (dismissed):

core/src/main/java/com/datastax/oss/driver/api/core/connection/ReconnectionPolicy.java
olim7t: I hadn't noticed that. I'll move those explanations to the main description then.

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
olim7t: I don't think concurrency is an issue, `Reconnection` uses the same admin thread and `Reconnection.start()` creates the schedule synchronously.

But you are right that the code is too convoluted. What makes it complicated is that the control connection can be initialized in two ways:

1. by default, it's the component that determines if the initial connection to the cluster succeeds.
2. But you could also have a custom `TopologyMonitor` that does that a different way, and only use the control connection for schema metadata and server events. In that case the control connection is initialized late, and `newControlConnectionSchedule()` must never be called with a `true` argument.

I think I can make the code more clear if there are separate init methods for those two cases. I'll give it a try.

adutra: > I don't think concurrency is an issue, Reconnection uses the same admin thread and Reconnection.start() creates the schedule synchronously.

Hmm my bad, `newSchedule()` is invoked synchronously while `reconnect()` is invoked asynchronously, so the future will never be done when `newSchedule()` is executed.

> I think I can make the code more clear if there are separate init methods for those two cases. I'll give it a try.

💯

olim7t:
Still OK to merge this?
We've decided to backtrack on the other reconnection policy changes in [JAVA-2086](https://datastax-oss.atlassian.net/browse/JAVA-2086), but I think this one is still nice to have.

adutra:
> Still OK to merge this?

+1 from me.

PR #1165, opened 2019-01-03
JAVA-2041: Deprecate cross-DC failover in DCAwareRoundRobinPolicy

emerkle826:


manual/load_balancing/README.md
```diff
@@ -118,6 +116,27 @@ local datacenter. In general, providing the datacenter name explicitly is a safe
 Hosts belonging to the local datacenter are at distance `LOCAL`, and appear first in query plans (in a round-robin
 fashion).
 
+#### Configuring Remote DCs as Local DCs
```
tolbertam: I think we should just remove all reference to these options from our docs.  We shouldn't bring attention to them if we are getting rid of them.  Instead, a user may encounter them in the javadocs, with the deprecating warning you added.

emerkle826: done.

tolbertam (approved):
Looks good! :+1:

PR #1166, opened 2019-01-03
JAVA-1159: Document workaround for using tuple with udt field in Mapper

tolbertam:
For [JAVA-1159](https://datastax-oss.atlassian.net/browse/JAVA-1159).  Added both documentation and a test.


olim7t (approved):
👍 
I agree with your assessment that it's not worth trying to fix this.

PR #1167, opened 2019-01-03
JAVA-1964: Complete remaining "Coming Soon" sections in docs

tolbertam:
Completed docs for:

* Authentication
* Control Connection
* Reconnection
* Statements
* UDTs

---

For [JAVA-1964](https://datastax-oss.atlassian.net/browse/JAVA-1964).  These are largely based on the 4.x docs, with the appropriate API changes and other details changed to match 3.x.

### DIFF
--- a/manual/reconnection/README.md
+++ b/manual/reconnection/README.md
@@ -4,9 +4,7 @@ If the driver loses a connection to a node, it tries to re-establish it accordin
 policy. This is used in two places:
 
 * [connection pools](../pooling/): for each node, a session has a fixed-size pool of connections to
-  execute user requests. If one or more connections drop, a reconnection gets started for the pool;
-  each attempt tries to reopen the missing number of connections. This goes on until the pool is
-  back to its expected size;
+  execute user requests. If a node is detected as down, a reconnection is started.
 * [control connection](../control_connection/): a session uses a single connection to an arbitrary
   node for administrative requests. If that connection goes down, a reconnection gets started; each
   attempt iterates through all active nodes until one of them accepts a connection. This goes on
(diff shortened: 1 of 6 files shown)


manual/reconnection/README.md
```diff
@@ -1,11 +1,58 @@
...
 
-<!--
-TODO cover:
-- reconnection policy
-- scheduled reconnections vs. gossip events
--->
+* [connection pools](../pooling/): for each node, a session has a fixed-size pool of connections to
+  execute user requests. If one or more connections drop, a reconnection gets started for the pool;
+  each attempt tries to reopen the missing number of connections. This goes on until the pool is
+  back to its expected size;
```
olim7t: In driver 3 this works differently: there is a single reconnection at the node level, that starts when it's detected as DOWN, and will try to reopen all pools (one per active session).

If the pool is under capacity, this is handled on the fly in `borrowConnection`, but I don't think we need to get into as much detail here.

manual/reconnection/README.md
```diff
@@ -1,11 +1,58 @@
...
+  attempt in 1 second;
+* [t = 1.5] the control connection to node1 goes down. A reconnection starts for the control
+  connection, with the next attempt in 1 second;
+* [t = 2.2], node3's pool tries to open its missing connection, which succeeds. The pool is back to
+  its expected size, node3's reconnection stops;
+* [t = 2.5] the control connection tries to find a new node. It invokes the
+  [load balancing policy](../load_balancing/) to get a query plan, which happens to start with
+  node4. The connection succeeds, node4 is now the control node and the reconnection stops;
+* [t = 3] node2's pool tries to open the last missing connection, which succeeds. The pool is back
+  to its expected size, node2's reconnection stops.
```
olim7t: This must also be changed to reflect the driver 3 behavior. Basically replace "connection lost" with "the node goes down", and a reconnection can only succeed by bringing the node back up and reopening all connections (at least we don't need to cover partial reconnection in this example).

olim7t (approved):
This looks great, I didn't catch any typo and the links all work.

The reconnection section needs a few adjustments for driver 3.

tolbertam:
Thanks for the review and feedback.  Will make the appropriate adjustments :+1: .

manual/statements/batch/README.md
```diff
@@ -1,5 +1,32 @@
 ## Batch statements
 
-*Coming soon... In the meantime, see the javadoc for [BatchStatement].*
+Use [BatchStatement] to execute a set of queries as an atomic operation (refer to
```
adutra: 1. Do we want to explain batch types (unlogged vs logged vs counter)?
2. I think that "atomic" is only correct for a batch if all statements share the same partition key.

manual/reconnection/README.md
tolbertam: I think i will just remove this example since as you mention the reconnection trigger logic is rather simple (triggered on node going down with schedules created).

manual/statements/batch/README.md
```diff
@@ -1,5 +1,32 @@
...
+BatchStatement batch = new BatchStatement()
+    .add(simpleInsertBalance)
+    .add(preparedInsertExpense.bind("Vera ADRIAN", 1, 7.95f, "Breakfast", false));
+
+session.execute(batch);
+```
+
+As shown in the examples above, batches can contain any combination of simple statements and bound
+statements. A given batch can contain at most 65536 statements. Past this limit, addition methods
+throw an `IllegalStateException`.
```
adutra, line 24: Is it worth mentioning that batches are also limited by `batch_size_fail_threshold_in_kb` in cassandra.yaml?

manual/udts/README.md
```diff
@@ -1,5 +1,99 @@
 ## User-defined types
 
-*Coming soon... In the meantime, see the javadoc for [UserType].*
+[CQL user-defined types][cql_doc] are ordered sets of named, typed fields. They must be defined in a
```
adutra, line 3: Should we add a note that UDTs can be frozen / non-frozen?

manual/statements/batch/README.md
tolbertam: :+1: Added a section for that

manual/statements/batch/README.md
tolbertam: I replaced 'atomic' with 'single'.  

I also added a sentence about using `UNLOGGED`, I didn't mention `COUNTER` here its advanced and covered in CQL documentation, so should be apparent how to use it if needed.

manual/udts/README.md
tolbertam: I think in the context of this documentation which is to demonstrate how to use the Driver API, I think it is ok to not cover it here.  Up until C* 3.6 UDTs were required to be frozen, so as long as our schema examples include frozen it should be ok.

manual/statements/batch/README.md
tolbertam: If you are ok with these changes I will follow on and also make them on 4.x.

adutra (approved):

PR #1168, opened 2019-01-04
Enable Travis CI builds against OpenJDK 8 and 11

adutra:
It seems that Travis statuses are not available yet due to the ongoing migration from travis-ci.org to travis-ci.com.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,13 +4,9 @@ sudo: false
 matrix:
   include:
     # 8
-    - env: JDK='Oracle JDK 8'
-      jdk: oraclejdk8
     - env: JDK='OpenJDK 8'
       jdk: openjdk8
     # 11
-    - env: JDK='Oracle JDK 11'
-      install: . ./install-jdk.sh -F 11 -L BCL
     - env: JDK='OpenJDK 11'
       install: $TRAVIS_BUILD_DIR/ci/install-jdk.sh -F 11 -L GPL
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,4 +18,4 @@ before_install:
 script: mvn test -B -Dmaven.main.skip=true  -Dmaven.test.skip=true
 cache:
   directories:
-    - $HOME/.m2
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,13 +1,26 @@
...
+      jdk: oraclejdk8
+    - env: JDK='OpenJDK 8'
+      jdk: openjdk8
+    # 11
+    - env: JDK='Oracle JDK 11'
+      install: . ./install-jdk.sh -F 11 -L BCL
+    - env: JDK='OpenJDK 11'
+      install: . ./install-jdk.sh -F 11 -L GPL
 before_install:
+  - wget https://github.com/sormuras/bach/raw/master/install-jdk.sh
```
tolbertam: What do you think about copying this script into our git repo, so it doesn't need to be pulled in every time?  Also prevents possible security issues if that script gets compromised for whatever reason (I know, I'm probably being overcautious :D).

.travis.yml
```diff
@@ -1,13 +1,26 @@
 language: java
-jdk:
-- oraclejdk8
 sudo: false
+# see https://sormuras.github.io/blog/2018-03-20-jdk-matrix.html
+matrix:
+  include:
+    # 8
+    - env: JDK='Oracle JDK 8'
```
tolbertam: Is it worth testing against both Oracle and OpenJDK?  Don't imagine we will need to worry about compatbility issues between the two.

tolbertam (approved):
Had a few optional suggestions but :+1: overall.

.travis.yml
adutra: I actually tried both and went for the wget approach, thinking that it would be a bit nicer if we could avoid hosting that file on our repo. But your security concerns just convinced me of the opposite :)

.travis.yml
adutra: OK, I will disable one of them.

tolbertam:
Recent changes look great, thanks for making the suggested changes :+1:

PR #1169, opened 2019-01-04
Experimental annotation-processing-based mapper for driver 4

olim7t:
This is a proof of concept exploring the use of annotation processing for the new mapper. Very early stage, nothing is prescriptive, the scope is very limited and the code takes many shortcuts for simplicity.

### Demo

There are example annotated sources in `integration-tests/src/test/java/c/d/o/d/mapper`.

Run `mvn clean generate-test-sources` or rebuild in your IDE. Look at the generated sources in `target/generated-test-sources`.

### Code overview

There are two new modules:
* `mapper-runtime`: what the user's code needs at runtime. Right now only the annotations, but there might also be some utility classes in the future.
* `mapper-processor`: the annotation processor. Only needed at compile time, the usual practice is to depend on it with scope `optional`. The JAR declares the processor as a service provider, the compiler will automatically pick it up and run it. The entry point is `MapperProcessor`.

Recommended reading: [Annotation processing 101](http://hannesdorfmann.com/annotation-processing/annotationprocessing101).


tomekl007:
I've did `mvn clean generate-test-sources` but `target/generated-test-sources` for both `mapper-runtime` and `mapper-processor` have nothing inside

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Table.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/** TODO is "table" the right name if we plan to support other models later (e.g. DSE graph)? */
+@Target({ElementType.TYPE})
+@Retention(RetentionPolicy.CLASS)
+public @interface Table {}
```
tomekl007, line 26: maybe `Entity`?

integration-tests/src/test/java/com/datastax/oss/driver/mapper/ProductDao.java
```diff
@@ -0,0 +1,31 @@
...
+import com.datastax.oss.driver.api.mapper.annotations.SaveQuery;
+import java.util.List;
+
+@Dao
+public interface ProductDao {
+
+  @SaveQuery(table = "product")
+  void save(Product product);
+
+  @Query("SELECT * FROM product")
```
tomekl007, line 29: How do we plan to handle named parameters here?
Passing args to `findAll(String name)` and referencing in query? like `SELECT * FROM PRODUCT where name = :name`

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/DaoImplementationGenerator.java
```diff
@@ -0,0 +1,103 @@
...
+public class DaoImplementationGenerator extends FileGenerator {
+
+  private final TypeElement interfaceElement;
+  private final ClassName implementationName;
+
+  public DaoImplementationGenerator(TypeElement interfaceElement, GenerationContext context) {
+    super(context);
+    this.interfaceElement = interfaceElement;
+    implementationName =
+        ClassName.get(interfaceElement).peerClass(interfaceElement.getSimpleName() + "_Impl");
```
tomekl007, line 45: Are we going to rely on `_Impl` suffix implicitly?  Isn't it fragile? I found that naming convention in `Spring Data` a bit hard to maintain. Refactoring is not simple if you need to stick with a proper name. Or is it only for internal use?

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Table.java
olim7t: Yes, naming is still tentative so this might change.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/ProductDao.java
olim7t: In driver 3 parameters are marked with `@Param` annotations and referenced in the query string by bind markers.
[Example](https://docs.datastax.com/en/developer/java-driver/3.6/manual/object_mapper/using/#parameters)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/DaoImplementationGenerator.java
olim7t: This is generated code so users shouldn't have to use/refactor that class directly. The only issue is if they already had a class with this name in their sources, but the unconventional name (with underscore) should minimize that. The implementation name could be customizable through the annotation, but I don't see a strong case for it.

olim7t:
The generated classes are in the `integration-tests` module. Look for `ProductDao_Impl` with your IDE.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/InventoryMapper.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.mapper;
+
+import com.datastax.oss.driver.api.mapper.annotations.Mapper;
+
+@Mapper
+public interface InventoryMapper {
```
tomekl007, line 21: `InventoryMapper` (and corresponding `InventoryMapper_Impl`) instance is created for every response?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/SaveQueryGenerator.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+    VariableElement parameter = parameters.get(0);
+    parameterName = parameter.getSimpleName().toString();
+    parameterType = parameter.asType();
+    entityDefinition = context.getEntityDefinitions().get(parameterType);
+    if (entityDefinition == null) {
+      context.getMessager().error(methodElement, "Missing definition of entity %s", parameterType);
+      throw new SkipGenerationException();
+    }
+  }
```
tomekl007, line 66: in generated `save` we are not binding and extracting parameters from Product

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/DaoImplementationGenerator.java
```diff
@@ -0,0 +1,103 @@
...
+    return implementationName.packageName() + "." + implementationName.simpleName();
+  }
+
+  public ClassName getGeneratedClassName() {
+    return implementationName;
+  }
+
+  @Override
+  protected JavaFile.Builder getContents() {
+
```
tomekl007, line 59: I wonder how hard it will be to generate pagination version of `findAll()` in this way

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/SaveQueryGenerator.java
```diff
@@ -0,0 +1,144 @@
...
+    classBuilder.addMethod(methodBuilder.build());
+  }
+
+  @Override
+  public void addConstructorInstructions(MethodSpec.Builder constructorBuilder) {
+    String queryName = methodName + "Query";
+    constructorBuilder.addCode(
+        "$[String $L = $T.insertInto($S)", queryName, QueryBuilderDsl.class, tableName);
+    for (FieldDefinition field : entityDefinition) {
+      constructorBuilder.addCode(
```
tomekl007, line 132: looking at the generated code:
`this.save_preparedStatement = session.execute(new DefaultPrepareRequest(saveQuery), PrepareRequest.SYNC);`
I have an impression that error handling (generating error handling code) will be not a trivial task. Do you know the best pattern for errror handling with this approach?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/MapperBuilderGenerator.java
```diff
@@ -0,0 +1,77 @@
...
+import com.squareup.javapoet.ClassName;
+import com.squareup.javapoet.FieldSpec;
+import com.squareup.javapoet.JavaFile;
+import com.squareup.javapoet.MethodSpec;
+import com.squareup.javapoet.TypeSpec;
+import javax.lang.model.element.Modifier;
+import javax.lang.model.element.TypeElement;
+
+public class MapperBuilderGenerator extends FileGenerator {
+
```
tomekl007, line 28: Will client use this API in the following way:
`
Session session = initSession();
InventoryMapperBuilder.build(session).getProductDao()
`?
Why we are generating also `build()` (parameterless) method? How `session` is initialized in that case?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/GenerationContext.java
```diff
@@ -0,0 +1,84 @@
...
+import javax.lang.model.util.Types;
+
+/**
+ * A custom context to share processor-level information with code generators. Basically just a way
+ * to avoid passing a gazillion parameters everywhere.
+ */
+public class GenerationContext {
+
+  private final Map<TypeMirror, EntityDefinition> entityDefinitions = new HashMap<>();
+  private final Map<TypeMirror, ClassName> generatedDaos = new HashMap<>();
```
olim7t, line 34: Mmm actually that won't work with incremental compilation: unmodified sources are not passed to the processor again, so we can't expect all DAOs and entities to have been processed in this round.

We'll have to rely on naming conventions to find DAO implementation classes, and re-parse entities on demand (with a cache within the processing round).

integration-tests/src/test/java/com/datastax/oss/driver/mapper/InventoryMapper.java
olim7t: I'm not sure I understand the question.
This is the main session wrapper, so in general there should be one per user application, with multiple DAO method factories. The implementation gets generated by the processor when you compile.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/DaoImplementationGenerator.java
olim7t: I think the best way is to expose something that mimics `ResultSet` but exposes entities instead of rows. This way the user can do the pagination however they like.

We have it in 3.x, it's called `Result` (although I'm not a big fan of the name).

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/MapperBuilderGenerator.java
olim7t: The session is mandatory so I made it a constructor argument of the builder:
```java
InventoryMapper mapper = new InventoryMapperBuilder(session)
  // .withXxx(...)
  // .withYyy(...)
  .build();
```
My instinct tells me that there will be other optional parameters that you can set on the builder. But if that turns out to not be true, we can revisit the pattern.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/SaveQueryGenerator.java
olim7t: In that case, I think errors encountered while preparing should be rethrown to the caller of the mapper builder.
If we later have methods that call `prepare()` each time instead of caching the statement as a DAO field, then errors should be rethrown to the callers of those methods.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/SaveQueryGenerator.java
olim7t: I'm working on some changes so I don't have access to the generated code right now, but that should get handled by the loop at line 113.

olim7t:
This was just a quick proof of concept, I'm going to reboot a fresh `java2078` branch with a cleaner version of the basic `Mapper` and `Dao` infrastructure, that we can build upon gradually to add entities, requests, etc.

I'm renaming this branch to [java2078proto](https://github.com/datastax/java-driver/tree/java2078proto) to keep it around for a while for reference (it should quickly become obsolete though). I can't change the branch on the pull request so I'm closing it.

PR #1170, opened 2019-01-08
JAVA-2067: Publish javadocs JAR for the shaded module

olim7t:


core-shaded/pom.xml
```diff
@@ -16,7 +16,9 @@
     limitations under the License.
 
 -->
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
adutra: This is unrelated but keeps happening frequently; we should address this by maybe creating more strict formatting rules for XML. One idea would be to use a formatting plugin, such as [this one](https://github.com/acegi/xml-format-maven-plugin).

adutra (approved):

core-shaded/pom.xml
olim7t: Good idea, I've created [JAVA-2107](https://datastax-oss.atlassian.net/browse/JAVA-2107) to keep track of it. In the meantime I'll revert this line.

PR #1171, opened 2019-01-09
JAVA-2034: Add performance recommendations in the manual

olim7t:
This should be a good start.

If you have other ideas I'm open to suggestions. We can add to the page over time as suggestions/questions come in.

### DIFF
--- a/manual/core/performance/README.md
+++ b/manual/core/performance/README.md
@@ -1,0 +1,352 @@
+## Performance
+
+This page is intended as a checklist for everything related to driver performance. Most of the
+information is already available in other sections of the manual, but it's linked here for
+easy reference if you're benchmarking your application or diagnosing performance issues.
+ 
+
+### Statements
+
+[Statements](../statements/) are some of the driver types you'll use the most. Every request needs
+one -- even `session.execute(String)` creates a `SimpleStatement` under the hood.
+
+#### Immutability and builders
+
+Statements are by default implemented with immutable classes: every call to a setter method creates
+an intermediary copy. If you have multiple attributes to set, use a builder instead:
+
+```java
+SimpleStatement statement = SimpleStatement.builder("SELECT * FROM foo")
+    .withPageSize(20)
+    .withConsistencyLevel(DefaultConsistencyLevel.QUORUM)
+    .withIdempotence(true)
+    .build();
+```
+
+Also, note that you don't need a driver `Session` to create simple statements: they can be
+initialized statically and stored as constants.
+
+#### Prepared statements
+
+[Prepared statements](../statements/prepared) allow Cassandra to cache parsed query strings
+server-side, but that's not their only benefit for performance:
+
+* the driver also caches the response metadata, which can then be skipped in subsequent responses.
+  This saves bandwidth, as well as the CPU and memory resources required to parse it.
+* in some cases, prepared statements set routing information automatically, which allows the driver
+  to target the most appropriate replicas.
+  
+You should use prepared statements for all recurring requests in your application. Simple statements
+should only be used for one-off queries, for example some initialization that will be done only once
+at startup.
+
+The driver caches prepared statements -- see [CqlSession.prepare(SimpleStatement)] for the fine
+print. However, if the query is static, it's still a good practice to cache your `PreparedStatement`
+instances to avoid calling `prepare()` every time. One common pattern is to use some sort of DAO
+component:
+
+```java
+public static class UserDao {
+
+  private final CqlSession session;
+  private final PreparedStatement preparedFindById;
+
+  public UserDao(CqlSession session) {
+    this.session = session;
+    this.preparedFindById = session.prepare("SELECT * FROM user WHERE id = ?");
+  }
+
+  public User findById(int id) {
+    Row row = session.execute(preparedFindById.bind(id)).one();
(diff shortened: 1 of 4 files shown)


manual/core/performance/README.md
```diff
@@ -0,0 +1,350 @@
...
+
+[Request trackers](../request_tracker/) are on the hot path (that is, invoked on I/O threads, each
+time a request is executed), and users can plug custom implementations.
+
+If you experience throughput issues, check if any trackers are configured, and what they are doing.
+They should avoid blocking calls, as well as any CPU-intensive computations.
+
+#### Metrics
+
+Similarly, some of the driver's [metrics](../metrics/) are updated for every request.
```
tomekl007: I think that we can state that `for every request if metric is enabled`.
Next paragraph that `By default, the driver ships with all metrics disabled` does not state that metrics are not updated - one may understand that it means that they are not exposed but updated. wdyt?

tomekl007:
Really great documentation, a lot of useful things described - Thank You @olim7t!

manual/core/performance/README.md
```diff
@@ -0,0 +1,350 @@
...
+operation so unlikely to be a bottleneck, but note that there are other options, such as a
+[thread-local](../query_timestamps/#thread-local-timestamp-generator) variant that creates slightly
+less contention, writing your own implementation or letting the server assign timestamps.
+
+#### Tracing
+
+[Tracing](../tracing/) should be used for only a small percentage of your queries. It consumes
+additional resources on the server, and fetching each trace requires background requests.
+
+Do not enable tracing for every requests; it's a sure way to bring your performance down.
```
emerkle826: Small typo. Should be:
`Do not enable tracing for every request`

manual/core/performance/README.md
```diff
@@ -0,0 +1,350 @@
...
+You may adjust those settings depending on your application's needs: higher values mean less impact
+on performance, but the driver will be slower to react to changes.  
+
+#### Schema updates
+
+You should group your schema changes as much as possible.
+
+Every change made from a client will be pushed to all other clients, causing them to refresh their
+metadata. If you have multiple client instances, it might be a good idea to
+[desactivate the metadata](../metadata/schema/#enabling-disabling) on other clients while you apply
```
emerkle826: typo
`desactivate` should be `deactivate`

manual/core/performance/README.md
olim7t: Yes, I'll add that precision.

manual/core/performance/README.md
```diff
@@ -0,0 +1,351 @@
...
+
+```java
+SimpleStatement statement = SimpleStatement.builder("SELECT * FROM foo")
+    .withPageSize(20)
+    .withConsistencyLevel(DefaultConsistencyLevel.QUORUM)
+    .withIdempotence(true)
+    .build();
+```
+
+Simple statements are also completely stateless, so you can store them as constants.
```
adutra: I would rather qualify it as "immutable" since `DefaultSimpleStatement` does have a state.

manual/core/performance/README.md
```diff
@@ -0,0 +1,351 @@
...
+ 
+
+### Statements
+
+[Statements](../statements/) are some of the driver types you'll use the most. Every request needs
+one -- even `session.execute(String)` creates a `SimpleStatement` under the hood.
+
+#### Immutability and builders
+
+Statements are immutable: every call to a setter method creates an intermediary copy. If you have
```
adutra: There is a slight confusion between interface and implementation: it's not the interfaces, but the _default implementations_ thereof that are immutable . Suggestion: "Statements are by default implemented with immutable classes: every call..."

manual/core/performance/README.md
```diff
@@ -0,0 +1,351 @@
...
+```
+
+This will save CPU and memory resources, but you lose some driver features:
+
+* if schema is disabled, `session.getMetadata().getKeyspaces()` will always be empty: your
+  application won't be able to inspect the database schema dynamically.
+* if the token map is disabled, `session.getMetadata().getTokenMap()` will always be empty, and you
+  lose the ability to use [token-aware routing](../load_balancing/#token-aware).
+
+Note that disabling the schema implicitly disables the token map (because computing the token map
```
adutra, line 231: TIL; so basically if you want token-aware routing you have to enable schema metadata entirely. Would it be worth exploring the idea of enabling token-aware routing (which is a feature that almost everyone wants) without fully enabling schema metadata (which is imo much less popular) – iow, only fetch the relevant schema bits that are required to compute the token map?

adutra (commented):
Awesome docs 👍

manual/core/performance/README.md
olim7t: Yes, the only thing the token map needs are keyspace replication options.

That being said, getting rid of partial schema updates in driver 4 was a tremendous simplification of the internal code, quite honestly I feel reluctant to reintroduce some of it.

I suggest we wait. I don't see any way this could break the API, so If users ask for it, we can address it later.

manual/core/performance/README.md
olim7t: Yes, I meant "independent from a driver session", I'll use that instead since I already mention immutability at the beginning of the paragraph.

adutra (approved):

PR #1172, opened 2019-01-10
Java2086

GregBestland:
Still need to add unit tests for onAbort case, in ReconnectionPolicySchedule

### DIFF
--- a/core/revapi.json
+++ b/core/revapi.json
@@ -700,7 +700,49 @@
         "newArchive": "com.datastax.oss:java-driver-core:jar:4.0.0-rc1-SNAPSHOT",
         "elementKind": "method",
         "justification": "JAVA-2077: Allow reconnection policy to detect first connection attempt"
-      }
+      },
+      {
+        "code": "java.method.numberOfParametersChanged",
+        "old": "method java.time.Duration com.datastax.oss.driver.api.core.connection.ReconnectionPolicy.ReconnectionSchedule::nextDelay()",
+        "new": "method java.util.Optional<java.time.Duration> com.datastax.oss.driver.api.core.connection.ReconnectionPolicy.ReconnectionSchedule::nextDelay(java.util.Optional<java.lang.Throwable>)",
+        "package": "com.datastax.oss.driver.api.core.connection",
+        "classQualifiedName": "com.datastax.oss.driver.api.core.connection.ReconnectionPolicy.ReconnectionSchedule",
+        "classSimpleName": "ReconnectionSchedule",
+        "methodName": "nextDelay",
+        "oldArchive": "com.datastax.oss:java-driver-core:jar:4.0.0-beta3",
+        "newArchive": "com.datastax.oss:java-driver-core:jar:4.0.0-rc1-SNAPSHOT",
+        "elementKind": "method",
+        "justification": "JAVA-2086 Enable abort on reconnect policies"
+      },
+      {
+        "code": "java.method.returnTypeChanged",
+        "old": "method java.time.Duration com.datastax.oss.driver.api.core.connection.ReconnectionPolicy.ReconnectionSchedule::nextDelay()",
+        "new": "method java.util.Optional<java.time.Duration> com.datastax.oss.driver.api.core.connection.ReconnectionPolicy.ReconnectionSchedule::nextDelay(java.util.Optional<java.lang.Throwable>)",
+        "oldType": "java.time.Duration",
+        "newType": "java.util.Optional<java.time.Duration>",
+        "package": "com.datastax.oss.driver.api.core.connection",
+        "classQualifiedName": "com.datastax.oss.driver.api.core.connection.ReconnectionPolicy.ReconnectionSchedule",
+        "classSimpleName": "ReconnectionSchedule",
+        "methodName": "nextDelay",
+        "oldArchive": "com.datastax.oss:java-driver-core:jar:4.0.0-beta3",
+        "newArchive": "com.datastax.oss:java-driver-core:jar:4.0.0-rc1-SNAPSHOT",
+        "elementKind": "method",
+        "justification": "JAVA-2086 Enable abort on reconnect policies"
+      },
+      {
+        "code": "java.method.numberOfParametersChanged",
+        "old": "method com.datastax.oss.driver.api.core.connection.ReconnectionPolicy.ReconnectionSchedule com.datastax.oss.driver.api.core.connection.ReconnectionPolicy::newControlConnectionSchedule()",
+        "new": "method com.datastax.oss.driver.api.core.connection.ReconnectionPolicy.ReconnectionSchedule com.datastax.oss.driver.api.core.connection.ReconnectionPolicy::newControlConnectionSchedule(boolean)",
+        "package": "com.datastax.oss.driver.api.core.connection",
+        "classQualifiedName": "com.datastax.oss.driver.api.core.connection.ReconnectionPolicy",
+        "classSimpleName": "ReconnectionPolicy",
+        "methodName": "newControlConnectionSchedule",
+        "oldArchive": "com.datastax.oss:java-driver-core:jar:4.0.0-beta3",
+        "newArchive": "com.datastax.oss:java-driver-core:jar:4.0.0-rc1-SNAPSHOT",
+        "elementKind": "method",
+        "justification": "JAVA-2086 Enable abort on reconnect policies"
+      },
+
     ]
   }
 }
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/connection/ExponentialReconnectionPolicy.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/connection/ExponentialReconnectionPolicy.java
@@ -23,6 +23,7 @@ import com.datastax.oss.driver.api.core.metadata.Node;
 import com.datastax.oss.driver.shaded.guava.common.base.Preconditions;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.time.Duration;
+import java.util.Optional;
 import java.util.concurrent.ThreadLocalRandom;
 import net.jcip.annotations.ThreadSafe;
 import org.slf4j.Logger;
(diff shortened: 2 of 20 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/pool/ChannelPool.java
```diff
@@ -259,7 +259,8 @@ private SingleThreaded(
               () -> reconnectionPolicy.newNodeSchedule(node),
               this::addMissingChannels,
               () -> eventBus.fire(ChannelEvent.reconnectionStarted(node)),
-              () -> eventBus.fire(ChannelEvent.reconnectionStopped(node)));
+              () -> eventBus.fire(ChannelEvent.reconnectionStopped(node)),
+              error -> {});
       this.configListenerKey =
```
GregBestland, line 262: This may need to be changed... the onFail here is not doing anything. I'm not 100% clear on what needs to happen here to successfuly close the channel.

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/Reconnection.java
```diff
@@ -70,21 +73,24 @@ public Reconnection(
       Supplier<ReconnectionSchedule> scheduleSupplier,
       Callable<CompletionStage<Boolean>> reconnectionTask,
       Runnable onStart,
-      Runnable onStop) {
+      Runnable onStop,
+      Consumer<Throwable> onFail) {
```
olim7t: I think `onAbort` might be a better name for this, and it would be worth adding a javadoc comment to explain that this is what happens when the policy has decided to stop reconnecting.

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/Reconnection.java
```diff
@@ -181,28 +187,37 @@ private void reallyStop() {
...
+                  LOG,
+                  "[{}] Uncaught error while starting reconnection attempt",
+                  logPrefix,
+                  f.cause());
+              scheduleNextAttempt(throwable);
+            }
+          });
+    } else {
+      Loggers.warnWithException(
+          LOG, "Reconnection attempt aborted, at discretion of reconnection policy", logPrefix);
```
olim7t: This should be a debug log. If we're aborting because the policy decided to, everything is working as expected, so we don't need to alert ops.
If people want a message at a different level, they can issue it from their policy.

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/Reconnection.java
```diff
@@ -229,7 +244,7 @@ private void onNextAttemptCompleted(Boolean success, Throwable error) {
         reallyStop();
       } else {
         assert state == State.ATTEMPT_IN_PROGRESS;
-        scheduleNextAttempt();
+        scheduleNextAttempt(error);
```
olim7t, line 246: This assumes that `reconnectionTask` will rethrow errors, which is not the case currently because the previous convention was that tasks would handle errors on their side an only return a success/failure boolean. So only unexpected errors will be passed here; see for example how `ChannelPool.onAllConnected()` swallows authentication errors, we want them to be reported to the policy now.

I think we need to change `reconnectionTask` to `Callable<CompletionStage<Void>>`: a successful stage means stop reconnecting, and failed stage means pass the error to the policy (which is already done here, we just need to remove the warning about unexpected errors).
Then `ChannelPool.onAllConnected()` can simply rethrow errors (see my other comment about throwing a wrapper), and `ControlConnection.reconnect` will do:
```java
connect (
  ...
  error -> result.completeExceptionally(error));
```

core/src/main/java/com/datastax/oss/driver/internal/core/pool/ChannelPool.java
olim7t: I think the right thing to do is to force the node down. If the policy decides to abort it means we have really given up on this node. You can see in `ChannelPool.onAllConnected()` how it already does that for invalid keyspace, instead of doing it there we should rethrow it to the policy (see my other comment).

When `onAllConnected` throws, there might be multiple errors, so I think we should use a wrapper:
```java
class PoolReconnectionException extends DriverException {
  /** The individual error for each connection that was attempted. */
  List<Throwable> getConnectionErrors();

  /** If the reconnection attempted to complete a pool that already had active channels, their count. */
  int getActiveChannels();
}
```

olim7t (changes requested):

core/src/test/java/com/datastax/oss/driver/internal/core/control/ControlConnectionTest.java
```diff
@@ -525,7 +532,7 @@ public void should_close_channel_if_closed_during_reconnection() {
     channel1.close();
     waitForPendingAdminTasks();
     Mockito.verify(eventBus).fire(ChannelEvent.channelClosed(node1));
-    Mockito.verify(reconnectionSchedule).nextDelay();
+    Mockito.verify(reconnectionSchedule).nextDelay(Optional.empty());
```
tomekl007, line 535: Maybe we can name that `Optional.empty()` to state the intent of the test? 
like `static Optional<T> NO_DELAY`

core/src/test/java/com/datastax/oss/driver/internal/core/util/concurrent/ReconnectionTest.java
```diff
@@ -295,7 +321,8 @@ public void should_stop_while_attempt_in_progress(boolean outcome) {
   @Test
   public void should_restart_after_stopped_while_attempt_in_progress() {
     // Given
-    Mockito.when(reconnectionSchedule.nextDelay()).thenReturn(Duration.ofNanos(1));
+    Mockito.when(reconnectionSchedule.nextDelay(Optional.empty()))
```
tomekl007, line 324: I think that we are missing test case when `nextDelay()` is getting not `Optional.empty` as an argument

core/revapi.json
```diff
@@ -700,7 +700,49 @@
         "newArchive": "com.datastax.oss:java-driver-core:jar:4.0.0-rc1-SNAPSHOT",
         "elementKind": "method",
         "justification": "JAVA-2077: Allow reconnection policy to detect first connection attempt"
-      }
+      },
+      {
+        "code": "java.method.numberOfParametersChanged",
```
tomekl007, line 705: do we need to make it not backward compatible change? We may introduce `newDelay()` method with parameter while leaving old parameterless version. I am curious what are our guidelines - when we can introduce non-compatible change?

olim7t: We're still pre-GA so we can afford breaking changes, and in that case I think it really makes sense so we don't need to complicate the API for the sake of backward compatibility.

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/Reconnection.java
GregBestland: I got most of the way through this... honestly I think this is the wrong way to solve this problem. Not the reconnectionTask in general, but the entire idea of delegating when to retry to the underlying policy.

Much of the validation of when to retry and when not to is getting pushed entirely down into the retry policies nextDelay()... this leaves quite a bit discretion in the retry policy, and also complicates the implementation of the retry policy as they need to know the intricacies of each scenario. I don't think it's realistic for a end user to be able to extend retry policy in this manner outside of our team, and I think I'm likely to introduce a bunch of bugs in the process. It also invites duplication of code.. since the retry policies scheduler is a private inner class, which make inheritance a little tricky.

I talked to a member of the evangelist team. I think they would be satisfied simply by having the log level increased in the cases of authentication failure. This would be far easier way to solve their problem. I could make this configurable, and only set it on the DSOD Connector

core/src/main/java/com/datastax/oss/driver/internal/core/pool/PoolReconnectionException.java
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  @NonNull
+  @Override
+  public DriverException copy() {
+    return new PoolReconnectionException(getConnectionErrors(), getActiveChannels);
+  }
+
+  public int getGetActiveChannels() {
+    return getActiveChannels;
```
tomekl007, line 44: where we plan to use that information? I don't see usage of that method anywhere

core/src/main/java/com/datastax/oss/driver/internal/core/connection/ExponentialReconnectionPolicy.java
```diff
@@ -143,6 +147,21 @@ public void close() {
     @NonNull
     @Override
     public Optional<Duration> nextDelay(Optional<Throwable> throwable) {
+      if (throwable.isPresent() && throwable.get() instanceof PoolReconnectionException) {
```
tomekl007, line 150: Logic from here and from `ConstantReconnectionPolicy.java` is the same. Can't we extract it to separate class and create `static` method for example: `checkIfThrowableIsFatal`?

core/src/test/java/com/datastax/oss/driver/internal/core/util/concurrent/ReconnectionTest.java
emerkle826: ~There is one place where it's not an empty Optional: https://github.com/datastax/java-driver/pull/1172/files#diff-513c8e00357a995493fdfebc12eed677R65
But I think that is the only one.~ My mistake, that uses empty as well

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/Reconnection.java
emerkle826: This does seem to be a bit cumbersome to have to propagate the connection Exceptions down to the ReconnectionPolicy in order to determine if we should abort the reconnection attempt(s). And I agree that end users trying to provide a custom implementation for the policy will have a difficult time if they aren't aware of all the reasons a connection attempt would fail and how they would manifest themselves in the code/Exceptions.
However, if all we do is increase the log level, the only thing that does is make it a little more obvious (if you are looking at the logs) that the client can't connect because it can't authenticate, correct? It will still keep retrying, and the log will be the only indication that it's not hung.

olim7t:
> honestly I think this is the wrong way to solve this problem. Not the reconnectionTask in general, but the entire idea of delegating when to retry to the underlying policy.

For the record: we've discussed this via other channels, and the team agrees that this idea was too ambitious. It pushes too much complexity into the policy. The only type of error that it would make sense to customize are auth errors, and even then always retrying is probably the most logical approach.
Instead we've decided to take a simpler route:
* always warn for authentication errors.
* honor the option for `advanced.connection.warn-on-init-error` for other types of errors.

Note that both behaviors are already implemented in `ChannelPool`, we just need to do the same in `ControlConnection`.

PR #1173, opened 2019-01-14
JAVA-2103: Expose partitioner name in TokenMap API

olim7t:
~Not sure if we should do 2102, see my comments on the JIRA ticket (sorry for suggesting it in the first place).~
edit - we're not doing it

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/metadata/TokenMap.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/metadata/TokenMap.java
@@ -144,4 +144,8 @@ public interface TokenMap {
   default Set<Node> getReplicas(@NonNull String keyspaceName, @NonNull TokenRange range) {
     return getReplicas(CqlIdentifier.fromCql(keyspaceName), range);
   }
+
+  /** The name of the partitioner class in use, as reported by the Cassandra nodes. */
+  @NonNull
+  String getPartitionerName();
 }
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ByteOrderedTokenFactory.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ByteOrderedTokenFactory.java
@@ -25,7 +25,14 @@ import net.jcip.annotations.ThreadSafe;
 @ThreadSafe
 public class ByteOrderedTokenFactory implements TokenFactory {
 
+  public static final String PARTITIONER_NAME = "org.apache.cassandra.dht.ByteOrderedPartitioner";
+
   public static final ByteOrderedToken MIN_TOKEN = new ByteOrderedToken(ByteBuffer.allocate(0));
+
+  @Override
+  public String getPartitionerName() {
+    return PARTITIONER_NAME;
+  }
 
   @Override
   public Token hash(ByteBuffer partitionKey) {
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/DefaultTokenFactoryRegistry.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/DefaultTokenFactoryRegistry.java
@@ -33,13 +33,13 @@ public class DefaultTokenFactoryRegistry implements TokenFactoryRegistry {
 
   @Override
   public TokenFactory tokenFactoryFor(String partitioner) {
-    if (partitioner.endsWith("Murmur3Partitioner")) {
+    if (Murmur3TokenFactory.PARTITIONER_NAME.equals(partitioner)) {
       LOG.debug("[{}] Detected Murmur3 partitioner ({})", logPrefix, partitioner);
       return new Murmur3TokenFactory();
-    } else if (partitioner.endsWith("RandomPartitioner")) {
+    } else if (RandomTokenFactory.PARTITIONER_NAME.equals(partitioner)) {
       LOG.debug("[{}] Detected random partitioner ({})", logPrefix, partitioner);
       return new RandomTokenFactory();
-    } else if (partitioner.endsWith("OrderedPartitioner")) {
+    } else if (ByteOrderedTokenFactory.PARTITIONER_NAME.equals(partitioner)) {
       LOG.debug("[{}] Detected byte ordered partitioner ({})", logPrefix, partitioner);
       return new ByteOrderedTokenFactory();
     } else {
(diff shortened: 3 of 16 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/metadata/TokenMap.java
```diff
@@ -144,4 +144,8 @@
   default Set<Node> getReplicas(@NonNull String keyspaceName, @NonNull TokenRange range) {
     return getReplicas(CqlIdentifier.fromCql(keyspaceName), range);
   }
+
+  /** The name of the partitioner class in use, as reported by the Cassandra nodes. */
+  @NonNull
+  String getPartitionerName();
```
tomekl007, line 150: Maybe instead of using `String`, we could introduce type like `PartitionerName` to make it more strong typed?

olim7t: I don't think we need strong type safety because this value is only exposed for reading, the user never passes it into the API.
Besides, the partitioner is pluggable(1), so there's no pre-defined set of values.

(1) at least in theory: you would have to write the server-side component and then plug-in your own `TokenFactoryRegistry` in the driver, that's a lot of work but feasible nevertheless.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ByteOrderedTokenFactory.java
```diff
@@ -26,6 +26,16 @@
 public class ByteOrderedTokenFactory implements TokenFactory {
 
   public static final ByteOrderedToken MIN_TOKEN = new ByteOrderedToken(ByteBuffer.allocate(0));
+  private final String partitionerName;
+
+  public ByteOrderedTokenFactory(String partitionerName) {
+    this.partitionerName = partitionerName;
+  }
+
+  @Override
```
adutra, line 32: Missing `@NonNull` annotation.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/Murmur3TokenFactory.java
```diff
@@ -26,6 +27,22 @@
...
+  }
+
+  // Only used from unit tests, where repeating the name everywhere is not really relevant
+  @VisibleForTesting
+  public Murmur3TokenFactory() {
+    this("org.apache.cassandra.dht.Murmur3Partitioner");
+  }
+
+  @Override
+  public String getPartitionerName() {
```
adutra, line 33: Missing `@NonNull` annotation.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/DefaultTokenFactoryRegistry.java
```diff
@@ -35,13 +35,13 @@ public DefaultTokenFactoryRegistry(InternalDriverContext context) {
   public TokenFactory tokenFactoryFor(String partitioner) {
     if (partitioner.endsWith("Murmur3Partitioner")) {
```
adutra: Is it really realistic to imagine that a different Murmur3 partitioner could exist other than `org.apache.cassandra.dht.Murmur3Partitioner`? And even so, shouldn't we error out in that case instead of assume that it's the same partitioner? I wonder if we shouldn't simply store the FQCN of each partitioner as a public constant field in each `TokenFactory` implementation and then reference them here using `equals` instead of `endsWith`.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/TokenITBase.java
```diff
@@ -339,4 +343,16 @@ public void should_create_token_from_partition_key() {
     assertThat(tokenMap.newToken(TypeCodecs.INT.encodePrimitive(1, protocolVersion)))
         .isEqualTo(expected);
   }
+
+  private TokenMap getTokenMap() {
+    Optional<? extends TokenMap> maybeTokenMap = session().getMetadata().getTokenMap();
+    if (maybeTokenMap.isPresent()) {
```
adutra: Could be simplified to:

```
assertThat(maybeTokenMap).isPresent().hasValueSatisfying(tokenMap -> {
  assertThat(tokenMap.getPartitionerName()).isEqualTo(expectedPartitionerName);
});
```

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ByteOrderedTokenFactory.java
olim7t: `TokenFactory` doesn't have any nullability annotations yet. Did we ever come up with a rule for internal stuff? I can add it in a separate commit.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/DefaultTokenFactoryRegistry.java
olim7t: Agreed, I copied that blindly from the 3.x codebase, but I don't think it's justified. It might be a remnant from the server side code where the package name can be omitted in the YAML file, but I don't think that ever propagates to system tables, we should always get the fully-qualified name.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/TokenITBase.java
olim7t: "simplified"? 😉 
More seriously, I had also considered:
```java
    return session()
        .getMetadata()
        .getTokenMap()
        .map(
            tokenMap -> {
              assertThat(tokenMap.getPartitionerName()).isEqualTo(expectedPartitionerName);
              return tokenMap;
            })
        .orElseThrow(() -> new AssertionError("Expected token map to be present"));
```
which has a bit more of a "functional" flavor but I still find it about as ugly as the imperative version.
I'm often disappointed by `Optional` code, without algebraic types or pattern matching you can't fully take advantage of it.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ByteOrderedTokenFactory.java
adutra: We don't, but you added one to `RandomTokenFactory`, so I figured you would like to add the same annotations to other implementations. (Tip: IntelliJ usually suggests to push down nullability annotations from the interface to all implementations).

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/TokenITBase.java
adutra: Well, 4 lines instead of 8 is a simplification to me ;-) but it doesn't matter really, the current version is fine enough.

adutra (approved):

PR #1174, opened 2019-01-18
JAVA-2082: Avoid race condition during cluster close and schema refresh

adutra:


driver-core/src/main/java/com/datastax/driver/core/EventDebouncer.java
```diff
@@ -205,14 +211,21 @@ private void scheduleDelayedDelivery() {
   }
 
   private DeliveryAttempt cancelDelayedDelivery() {
-    DeliveryAttempt previous = delayedDelivery.get();
+    return cancelDelivery(delayedDelivery.get());
+  }
+
+  private DeliveryAttempt cancelImmediateDelivery() {
```
adutra, line 217: I think it doesn't hurt to cancel the immediate delivery as well, but the real fix is to actually catch ISE inside `SchemaParser.refresh()`.

driver-core/src/main/java/com/datastax/driver/core/EventDebouncer.java
```diff
@@ -117,6 +117,12 @@ void stop() {
         break;
       }
     }
+    while (true) {
```
tomekl007, line 120: Isn't there a risk of infinite loop?

adutra: No, this is the typical idiom for CAS operations; the `break` instruction inside the loop guarantees that the thread will eventually exit the loop, even if it takes a few times to get the CAS operation to succeed.

olim7t (approved):

PR #1175, opened 2019-01-21
JAVA-2109: Bootstrap entity annotations and processing

olim7t:
New `@Entity` annotation, and processing code to look for matching getter/setter pairs. This is the bare minimum but I think it gives us something to build upon for other tickets.


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -107,6 +116,39 @@ protected String getFileName() {
...
+
+  /**
+   * If the DAO has a constant of a mapped entity, dump the definition of that entity in the DAO's
+   * javadoc.
+   *
+   * @deprecated this is a quick and dirty test to check that entity parsing works, currently the
+   *     only way because we don't use entities in generated code yet. TODO delete
+   */
+  @Deprecated
+  private void documentEntityConstants(TypeSpec.Builder classBuilder) {
```
olim7t: Example of test code:
```java
@Entity
public class Product {
  // fields, getters, setters
}
@Dao
public interface ProductDao {
  Product TEST = null;
}
```
Then look at `ProductDao_Impl.java`.

This is a bit lame but our DAOs do nothing else right now, so this can be used as a smoke test until we get a little further.

adutra:
One general remark about the OSGi manifests: both modules (runtime and processor) currently have the same Export-Package directive:

```
<Export-Package>
    com.datastax.oss.driver.*.mapper.*
</Export-Package>
```

I wonder if this cannot lead to split-package problems: e.g. some classes in the mapper-runtime module reside under `com.datastax.oss.driver.internal.mapper` but this package is a "parent package" of ` com.datastax.oss.driver.internal.mapper.processor` declared in the mapper-processor module. Just to be on the safe side, I would suggest that we adopt distinct package names for each of these modules, and change the Export-Package directives accordingly.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
adutra: I'm having a hard time trying to test/exercise this PR. 

1. I created a small ` ProductDao` interface as you suggested, the resulting ` ProductDao_Impl` seems a bit odd. Its `initAsync` method doesn't seem to do much for now, is this just a method stub?
2. I also tried to create a small mapper interface, but the generated class does not compile:

```java
public class ProductMapper_Impl implements ProductMapper {
  private final MapperContext context;

  private final ConcurrentMap<DaoCacheKey, ProductDao> productDaoCache = new ConcurrentHashMap<>();

  public ProductMapper_Impl(MapperContext context) {
    this.context = context;
  }

  @Override
  public ProductDao productDao(String table) {
    DaoCacheKey key = new DaoCacheKey((CqlIdentifier)null, table);
    return productDaoCache.computeIfAbsent(key, k -> ProductDao_Impl.init(context, k.getKeyspaceId(), k.getTableId()));
  }
}
```
Error:
```
Error:(26, 40) java: cannot find symbol
  symbol:   class CqlIdentifier
  location: class com.datastax.oss.driver.mapper.ProductMapper_Impl
```
It seems that the import of `CqlIdentifier` is missing.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -0,0 +1,122 @@
...
+            continue;
+          }
+          builder.withSetterName(methodName);
+        }
+      }
+    }
+
+    ImmutableList.Builder<PropertyDefinition> definitions = ImmutableList.builder();
+    for (DefaultPropertyDefinition.Builder builder : propertyBuilders.values()) {
+      if (builder.getGetterName() != null && builder.getSetterName() != null) {
```
adutra, line 117: Should we also provide immutable entities (i.e. only getters, and value injection through a dedicated constructor)?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -0,0 +1,122 @@
...
+        definitions.add(builder.build());
+      }
+    }
+
+    String entityName = Introspector.decapitalize(classElement.getSimpleName().toString());
```
adutra, line 122: Is the target CQL table name derived from the entity name? And if so, how can the user customize the CQL name?
Also, what I don't get is that the `@Mapper` annotation is supposed to create DAOs for different tables according to this example:

```
public interface InventoryMapper {
    ProductDao productDao(@DaoTable String table);
}
```
But in this case I would think that the entity name is irrelevant when determining the target table name?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultPropertyDefinition.java
```diff
@@ -0,0 +1,92 @@
...
+  public DefaultPropertyDefinition(
+      String cqlName, String getterName, String setterName, TypeName type) {
+    this.cqlName = cqlName;
+    this.getterName = getterName;
+    this.setterName = setterName;
+    this.type = type;
+  }
+
+  @Override
+  public String getCqlName() {
```
adutra, line 43: Same here, I suppose that we will provide means for CQL name customization in the future?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
adutra: > Its initAsync method doesn't seem to do much for now, is this just a method stub?

OK I think this is just a stub according to the design doc.

olim7t:
> One general remark about the OSGi manifests

One thing I didn't think of when I initialized the POMs: I wonder if `mapper-processor` needs to be an OSGi bundle at all, it's a compile-time thing only and there's absolutely no reason to have it as a runtime dependency of your application.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: > It seems that the import of CqlIdentifier is missing.

Have you fetched recently? I fixed that in 142621e3c yesterday.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
olim7t: > It seems that the import of CqlIdentifier is missing.

Have you fetched recently? I fixed that in 142621e3c yesterday.

> Its initAsync method doesn't seem to do much for now

Yes, it's just a stub. Eventually it will prepare all the underlying statements asynchronously, and chain the construction of the DAO to the completion of those prepare requests.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: I haven't planned it but it's something we could consider. It adds a bit of complexity though.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: Yes, the table name is directly derived from the entity for now. [JAVA-2110](https://datastax-oss.atlassian.net/browse/JAVA-2110) will introduce naming strategies. We can also add a way to override it through annotations (either as an attribute of `@Entity`, or as another annotation).

If the table name is injected from the DAO factory method, it will override anything that is defined at the DAO level. The generated code will determine that in `initAsync`, and prepare the statements to target the appropriate keyspaces: for statements generated with the query builder, it's trivial to substitute; for user-provided query strings, we'll adopt the convention that the query string must contain a placeholder (e.g. `${table}`), and the generated code will do a string substitution right before preparing.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultPropertyDefinition.java
olim7t: Yes. Quite honestly the whole `PropertyDefinition` API is tentative: it gives us something to build upon for the next tickets, but it will probably evolve over time.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
olim7t: My bad, I fixed the `CqlIdentifier` bug on `java2078`, it's not on this branch. I can rebase the PR if that's OK with you.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -0,0 +1,142 @@
...
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.VariableElement;
+import javax.lang.model.type.DeclaredType;
+import javax.lang.model.type.TypeKind;
+import javax.lang.model.type.TypeMirror;
+
+public class DefaultEntityFactory implements EntityFactory {
+
+  private final ProcessorContext context;
+  // Note: the processor code is not concurrent, we only use ConcurrentMap for computeIfAbsent
```
tomekl007: We cannot use java >= 1.8 to have `computeIfAbsent` on the `Map` interface?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -0,0 +1,142 @@
...
+      if (builder.getGetterName() != null && builder.getSetterName() != null) {
+        definitions.add(builder.build());
+      }
+    }
+
+    String entityName = Introspector.decapitalize(classElement.getSimpleName().toString());
+    return new DefaultEntityDefinition(entityName, definitions.build());
+  }
+
+  private TypeElement getEntityElement(TypeMirror typeMirror) {
```
tomekl007, line 127: Maybe we could wrap raw `TypeElement` into class `EntityElement` to have more strong types?
We could wrap all specific `TypeElement`s this way to make our code more explicit about types that is able to handle

olim7t:
FYI: I need to revisit this PR.

With the experience of my ongoing work on [JAVA-2113](https://datastax-oss.atlassian.net/browse/JAVA-2113), I think there is a better approach: process entities from the main processor loop, like we do currently for `@Mapper` and `@Dao`. For each entity, generate a "helper class" that implements the following interface:
```java
interface EntityHelper<EntityT> {
  void inject(EntityT entity, SettableByName<?> target);
  EntityT extract(GettableByName<?> source);
  OngoingSelection selectAll();
  // etc.
}
```
Most of the logic of parsing the properties will happen during that helper generation.

Higher-level generators (`@EntityInjector`, `@EntityExtractor` and query methods) will simply assume that the class exists. They probably won't need `EntityDefinition`, so we can remove the cache.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: I don't understand. We do use java >= 1.8 (see parent POM), so `computeIfAbsent` is available.

tomekl007: So we can use simple `HashMap`:
```
Map<String, String> map = new LinkedHashMap<>();
map.computeIfAbsent()
```
right?
Without an overhead of using `ConcurrentHashMap`

olim7t: Ah right, for some reason I thought it was only on `ConcurrentMap` even in 1.8. Yes, we can use `HashMap`.

olim7t:
As promised, I pushed a new commit with the refactoring explained above.

Entities are handled from the main processor loop. Given a class `@Entity User{}`, it will generate `User_Helper implements EntityHelper<User>`. Currently `EntityHelper` only defines `inject`, but we will add other methods as we go.

The current version maps child entities to UDTs, e.g. if `User` has a field of type `Address`. I need to add collections of UDTs, e.g. if it's a `List<Address>`.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: Captured as [JAVA-2125](https://datastax-oss.atlassian.net/browse/JAVA-2125).

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: Moot point now that the factory doesn't cache anymore, but thanks for the tip.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: I'm not sure if we can extract anything higher-level than `TypeMirror`/`Element`; I feel we would end up duplicating most of the API: at the end of the day we still need to handle all the weird cases, like missing types, etc.

It's still a bit early anyway. Let's wait for a few more tickets and maybe something will emerge.

olim7t:
I've squashed all the commits, rebased on the latest `java2078`, and created a follow-up ticket [JAVA-2129](https://datastax-oss.atlassian.net/browse/JAVA-2125) for UDT collections.

I think this is in pretty good shape for now, I'd like to merge it soon to unblock downstream tickets. We're early in development so it doesn't have to be perfect, we can create other follow-up tickets if needed.

adutra (approved):

PR #1176, opened 2019-01-23
Java2063 Normalize authentication error logging.

GregBestland:
This brings control connection Authentication logging inline with the channel pool authentication logging. They are both logged at warn.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
@@ -276,7 +276,11 @@ public class ControlConnection implements EventCallback, AsyncAutoCloseable {
             },
             error -> {
               if (isAuthFailure(error)) {
-                Loggers.warnWithException(LOG, "[{}] Authentication error", logPrefix, error);
+                Loggers.warnWithException(
+                    LOG,
+                    "[{}] Authentication errors encountered on all contact points. Please check our authentication configuration.",
+                    logPrefix,
+                    error);
               }
               if (reconnectOnFailure && !closeWasCalled) {
                 reconnection.start();
@@ -337,11 +341,16 @@ public class ControlConnection implements EventCallback, AsyncAutoCloseable {
                       if (closeWasCalled || initFuture.isCancelled()) {
                         onSuccess.run(); // abort, we don't really care about the result
                       } else {
-                        LOG.debug(
-                            "[{}] Error connecting to {}, trying next node",
-                            logPrefix,
-                            node,
-                            error);
+                        if (error instanceof AuthenticationException) {
+                          Loggers.warnWithException(
+                              LOG, "[{}] Authentication error", logPrefix, error);
+                        } else {
+                          LOG.debug(
+                              "[{}] Error connecting to {}, trying next node",
+                              logPrefix,
+                              node,
+                              error);
+                        }
                         Map<Node, Throwable> newErrors =
                             (errors == null) ? new LinkedHashMap<>() : errors;
                         newErrors.put(node, error);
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
```diff
@@ -501,6 +515,20 @@ private void forceClose() {
     }
   }
 
+  private boolean isAuthFailure(Throwable error) {
+    boolean authFailure = true;
+    if (error instanceof AllNodesFailedException) {
+      Collection<Throwable> errors = ((AllNodesFailedException) error).getErrors().values();
```
tomekl007, line 561: if `getErrors()` will return empty connection we will treat it as the `AuthFailure`. Is it ok for us? Is it possible?

GregBestland: That SHOULD never happen, I'll add a check regardless.

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
```diff
@@ -273,6 +275,13 @@ private void init(boolean listenToClusterEvents, boolean reconnectOnFailure) {
               firstConnectionAttemptFuture.complete(null);
             },
             error -> {
+              if (isAuthFailure(error)) {
+                Loggers.warnWithException(
+                    LOG,
+                    "[{}] Authentication errors encountered on all contact points. Please check your authentication configuration.",
+                    logPrefix,
+                    error);
+              }
```
olim7t, line 311: The errors have already been logged individually in `connect`, do we need another log here?
For example run `PlainTextAuthProviderIT.should_not_connect_without_credentials()`, there are two warning logs.

If we decide to keep this one, we should at least not include the exception, the stack trace doesn't bring any useful information.

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
```diff
@@ -332,11 +341,16 @@ private void connect(
...
-                            error);
+                        if (error instanceof AuthenticationException) {
+                          Loggers.warnWithException(
+                              LOG, "[{}] Authentication error", logPrefix, error);
+                        } else {
+                          LOG.debug(
+                              "[{}] Error connecting to {}, trying next node",
+                              logPrefix,
+                              node,
+                              error);
```
olim7t: Could you also honor `DefaultDriverOption.CONNECTION_WARN_INIT_ERROR` here?
See `ChannelPool.onAllConnected()`.

olim7t (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
GregBestland: I'd like to keep this to be explicit to the user that something about their configuration is probably wrong. I have no problem removing the exception and stack trace.

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
GregBestland: I think that should address your issues. If Feel very strongly about the cumulative logging warning I can remove it, I just want to be explicit about their configuration being incorrect.

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
olim7t: Fair enough, there could be auth errors on _some_ nodes but not all, the cumulative warning helps distinguish that case.

olim7t:
LGTM. I had a couple more trivial comments, I addressed them directly.

Last thing: tests are failing, you just need to mock the config option in `ControlConnectionTestBase`. After that this is good to go.

olim7t (approved):

olim7t:
Nit: could you use the usual conventions in the PR title next time? When you merge with "squash and merge", that's what GitHub uses for the commit message.
```
d6cb615e8 - (datastax/4.x) JAVA-2119: Add PagingIterable abstraction
9a81e4470 - Java2063 Normalize authentication error logging. (#1176)
a2af2edcc - JAVA-2084: Fix session.executeAsync examples in the docs
```
That's a small detail, but it's convenient in case you need to grep a ticket number in the log.

PR #1177, opened 2019-01-24
JAVA-2119: Add PagingIterable abstraction as a supertype of ResultSet

olim7t:
Contrary to our expectations, this doesn't seem to break backward compatibility.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/AsyncPagingIterable.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/AsyncPagingIterable.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core;
+
+import com.datastax.oss.driver.api.core.cql.ColumnDefinitions;
+import com.datastax.oss.driver.api.core.cql.ExecutionInfo;
+import com.datastax.oss.driver.internal.core.AsyncPagingIterableWrapper;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.Iterator;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+
+/**
+ * An iterable of elements which are fetched asynchronously by the driver, possibly in multiple
+ * requests.
+ */
+public interface AsyncPagingIterable<ElementT> {
+
+  /** Metadata about the columns returned by the CQL request that was used to build this result. */
+  @NonNull
+  ColumnDefinitions getColumnDefinitions();
+
+  /** Returns {@linkplain ExecutionInfo information about the execution} of this page of results. */
+  @NonNull
+  ExecutionInfo getExecutionInfo();
+
+  /** How many rows are left before the current page is exhausted. */
+  int remaining();
+
+  /**
+   * The elements in the current page. To keep iterating beyond that, use {@link #hasMorePages()}
+   * and {@link #fetchNextPage()}.
+   *
+   * <p>Note that this method always returns the same object, and that that object can only be
+   * iterated once: elements are "consumed" as they are read.
+   */
+  @NonNull
+  Iterable<ElementT> currentPage();
+
+  /**
+   * Returns the next element, or {@code null} if the results are exhausted.
+   *
+   * <p>This is convenient for queries that are known to return exactly one element, for example
+   * count queries.
+   */
+  @Nullable
+  default ElementT one() {
+    Iterator<ElementT> iterator = currentPage().iterator();
+    return iterator.hasNext() ? iterator.next() : null;
+  }
+
+  /**
+   * Whether there are more pages of results. If so, call {@link #fetchNextPage()} to fetch the next
+   * one asynchronously.
+   */
+  boolean hasMorePages();
+
+  /**
+   * Fetch the next page of results asynchronously.
+   *
+   * @throws IllegalStateException if there are no more pages. Use {@link #hasMorePages()} to check
+   *     if you can call this method.
+   */
+  @NonNull
+  CompletionStage<? extends AsyncPagingIterable<ElementT>> fetchNextPage()
+      throws IllegalStateException;
+
+  /**
+   * If the query that produced this result was a CQL conditional update, indicate whether it was
+   * successfully applied.
+   *
+   * <p>For consistency, this method always returns {@code true} for non-conditional queries
+   * (although there is no reason to call the method in that case). This is also the case for
+   * conditional DDL statements ({@code CREATE KEYSPACE... IF NOT EXISTS}, {@code CREATE TABLE... IF
+   * NOT EXISTS}), for which Cassandra doesn't return an {@code [applied]} column.
+   *
+   * <p>Note that, for versions of Cassandra strictly lower than 2.1.0-rc2, a server-side bug (<a
+   * href="https://issues.apache.org/jira/browse/CASSANDRA-7337">CASSANDRA-7337</a>) causes this
+   * method to always return {@code true} for batches containing conditional queries.
+   */
+  boolean wasApplied();
+
+  /**
+   * Creates a new instance by transforming each element of this iterable with the provided
+   * function.
+   *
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/PagingIterable.java
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * <p>Note that this object can only be iterated once: elements are "consumed" as they are read,
+ * subsequent calls to {@code iterator()} will return the same iterator instance.
+ *
+ * <p>Implementations of this type are <b>not thread-safe</b>. They can only be iterated by the
+ * thread that invoked {@code session.execute}.
+ *
+ * <p>This is a generalization of {@link ResultSet}, replacing rows by an arbitrary element type.
+ */
+public interface PagingIterable<ElementT> extends Iterable<ElementT> {
```
olim7t, line 47: The sync version never returns itself, so we don't need a self-type.

core/src/main/java/com/datastax/oss/driver/api/core/AsyncPagingIterable.java
```diff
@@ -0,0 +1,102 @@
...
+      throws IllegalStateException;
+
+  /**
+   * If the query that produced this result was a conditional update, indicate whether it was
+   * successfully applied.
+   *
+   * <p>In theory this method should not be exposed on {@link AsyncPagingIterable} since it is
+   * specific to CQL; in practice, most {@code AsyncPagingIterable} instances will be created by
+   * wrapping an {@link AsyncResultSet}, so it is convenient to make this information available
+   * without having to unwrap.
```
olim7t: This is up for debate. But I think it makes sense to map an `INSERT ... IF NOT EXISTS`, and get back `wasApplied() == false` and the existing entity if it failed.

core/src/main/java/com/datastax/oss/driver/api/core/AsyncPagingIterable.java
```diff
@@ -0,0 +1,102 @@
...
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.Iterator;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+
+/**
+ * An iterable of elements which are fetched asynchronously by the driver, possibly in multiple
+ * requests.
+ */
+public interface AsyncPagingIterable<ElementT> {
```
olim7t, line 31: There *could* be a self type here, because of `fetchNextPage()`. I tried to go down that rabbit hole, but it wasn't pretty; it propagates to the result of `map`, but it can't be materialized on the client side:
```java
PagingIterable<?, Integer> iterable = resultSet.map(row -> row.getInt(0));
```
Also, the implementation of `AsyncPagingIterableWrapper` was horrendous.

So I chose to go the simpler route and override the method with a covariant type in `AsyncResultSet`.

core/src/main/java/com/datastax/oss/driver/api/core/AsyncPagingIterable.java
```diff
@@ -0,0 +1,102 @@
...
+   * Creates a new instance by transforming each element of this iterable with the provided
+   * function.
+   *
+   * <p>Note that both instances share the same underlying data: consuming elements from the
+   * transformed iterable will also consume them from this object, and vice-versa.
+   */
+  default <TargetT> AsyncPagingIterable<TargetT> map(
+      Function<? super ElementT, ? extends TargetT> elementMapper) {
+    return new AsyncPagingIterableWrapper<>(this, elementMapper);
+  }
```
olim7t, line 107: I really like to have this here. It's the most natural thing that users might want to do with a result set, so it makes sense to have a shortcut for it. Also it will be super convenient in the object mapper.

As I said in JIRA, I think we should stay minimal and not try to implement other functional operators. They don't necessarily work well with the concept of page boundaries.

core/src/main/java/com/datastax/oss/driver/api/core/AsyncPagingIterable.java
adutra: I wonder if a self type for `fetchNextPage` wouldn't even be conceptually wrong, because the contract of this method does not mandate that implementors should return a future of themselves (iow, some implementation A might return a future that once completed yields an object of type B).

That said, I would like to I understand what was so horrible regarding the implementation of `AsyncPagingIterableWrapper`; [I gave it a try myself](https://github.com/datastax/java-driver/commit/270a2473b27dc191e1fcdcd0b23daa7e1f581b52) and found it actually pretty manageable.

core/src/main/java/com/datastax/oss/driver/api/core/AsyncPagingIterable.java
adutra: 1. I agree to have the method promoted here, but in this case why not push the logic even further and expose here everything that the native protocol provides us with, including `getColumnDefinitions()`? Granted, if you are mapping rows to entities you might want to actually forget about such metadata completely once the mapping is done; but other usages of this interface might be happy if they can inspect them. We never know.
2. This paragraph leaks a bit too many implementation details imo; besides "specific to CQL" is vague: everything we do is specific to CQL in some way :) I would suggest that we don't go into the details of explaining why this method is here and not in `AsyncResultSet` anymore and remove this paragraph completely.

core/src/main/java/com/datastax/oss/driver/api/core/AsyncPagingIterable.java
adutra: I'm not sure actually. First, it's already trivial to obtain a Java 8 stream from `currentPage()` and call `map()` on it; the same is valid for reactive streams, so on-the-fly mapping use cases are imo already well covered. 
For users wanting to transform a result set A into a result set B in one single call, other solutions exist: a) write a custom wrapper (similar to `AsyncPagingIterableWrapper `) or b) register a custom request processor. So in the end I'm not sure that this method buys us much.

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/api/core/AsyncPagingIterable.java
olim7t: 1. Makes sense, I will pull it up.
2. OK. The one comment that is specific to `ResultSet` is the paragraph about it being equivalent to `this.iterator().next().getBoolean("[applied]")`, I will keep it on that type.

core/src/main/java/com/datastax/oss/driver/api/core/AsyncPagingIterable.java
olim7t: I ended up having to parameterize it with 4 types: the source and target iterables, and the source and target element types, and there were still unchecked casts.
Indeed your implementation looks simpler. I noticed your `fetchNextPage` doesn't return a covariant future, not sure if this has an impact or not. Maybe I just over-complicated things.

But since we agree that a self-type is not desirable (the `?` in the caller code is the worst aspect IMO), I won't dig into it further.

core/src/main/java/com/datastax/oss/driver/api/core/AsyncPagingIterable.java
olim7t: > it's already trivial to obtain a Java 8 stream from `currentPage()` and call `map()` on it

If you work with async result sets, you still have to write the logic to rewrap every subsequent page, which incurs a bit of boilerplate.
If you use synchronous ones, you can indeed map the iterator, but you lose the paging methods (`isFullyFetched`, `getAvailableWithoutFetching`).

What's nice with `map()` is that you get all of that for free.

> write a custom wrapper (similar to `AsyncPagingIterableWrapper`)

Yes, that's pretty easy to write. But we're going to need it for the mapper anyway, so why not make it available?

> register a custom request processor

That's the nuclear option 😉

adutra: > but you lose the paging methods (isFullyFetched, getAvailableWithoutFetching)

Hmm OK this finally convinced me of the usefulness of the `map` method.

core/src/main/java/com/datastax/oss/driver/api/core/PagingIterable.java
```diff
@@ -0,0 +1,161 @@
...
+  /**
+   * Returns all the remaining elements as a list; <b>not recommended for queries that return a
+   * large number of elements</b>.
+   *
+   * <p>Contrary to {@link #iterator()} or successive calls to {@link #one()}, this method forces
+   * fetching the <b>full contents</b> at once; in particular, this means that a large number of
+   * background queries might have to be run, and that all the data will be held in memory locally.
+   * Therefore it is crucial to only call this method for queries that are known to return a
+   * reasonable number of results.
+   */
```
GregBestland, line 101: Would it make sense to to checks in the actual implementation to prevent blowing out the heap... or is this just an instance of user better know what they are doing?

GregBestland (approved):

core/src/main/java/com/datastax/oss/driver/api/core/PagingIterable.java
olim7t: We can't predict in advance the number of rows, or even the size each row will take in memory, so it's more the latter IMHO.
This method is already present in 3.x so users should be familiar with the pitfalls.

adutra (approved):

PR #1178, opened 2019-01-24
JAVA-1950: Log server side warnings returned from a query

emerkle826:


driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -224,6 +225,16 @@ private void setFinalResult(
                 response.getCustomPayload());
       }
       callback.onSet(connection, response, info, statement, System.nanoTime() - startTime);
+      // if the response from the server has warnings, they'll be set on the ExecutionInfo. Log them
+      // here, if enabled.
+      if (logger.isWarnEnabled()
+          && response.warnings != null
+          && !response.warnings.isEmpty()
+          && Boolean.getBoolean(RequestHandler.LOG_REQUEST_WARNINGS_PROPERTY)) {
```
tomekl007: What will happen if that property `RequestHandler.LOG_REQUEST_WARNINGS_PROPERTY))` will be not set?

emerkle826: From the Javadocs:
> If there is no property with the specified name, or if the specified name is empty or null, then false is returned.

So there shouldn't be any NullPointerException issues, and the current behavior is the default/unset behavior.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -64,6 +64,7 @@
 
   private static final boolean HOST_METRICS_ENABLED =
       Boolean.getBoolean("com.datastax.driver.HOST_METRICS_ENABLED");
+  static final String LOG_REQUEST_WARNINGS_PROPERTY = "advanced.request.log-warnings";
```
olim7t: In driver 3 we've been following a different naming convention for this kind of system property, see the previous line for example.
Also, as much as I dislike the idea of logging those warnings, I think we should do it by default, because the assumption is that users will completely miss them otherwise.

So I would invert the property and call it `com.datastax.driver.DISABLE_QUERY_WARNING_LOGS`.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -224,6 +225,16 @@ private void setFinalResult(
...
       }
       callback.onSet(connection, response, info, statement, System.nanoTime() - startTime);
+      // if the response from the server has warnings, they'll be set on the ExecutionInfo. Log them
+      // here, if enabled.
+      if (logger.isWarnEnabled()
+          && response.warnings != null
+          && !response.warnings.isEmpty()
+          && Boolean.getBoolean(RequestHandler.LOG_REQUEST_WARNINGS_PROPERTY)) {
+        for (String warning : response.warnings) {
+          logger.warn(warning);
```
olim7t: Should we provide a bit more context? Something like "The server generated a warning for a query: {}"... Maybe also include the query string, or a substring if it's too long?

driver-core/src/test/java/com/datastax/driver/core/WarningsTest.java
```diff
@@ -50,4 +54,52 @@ public void should_expose_warnings_on_execution_info() {
     List<String> warnings = rs.getExecutionInfo().getWarnings();
     assertThat(warnings).hasSize(1);
   }
+
+  @Test(groups = "short")
+  public void should_execute_query_with_server_side_warnings() {
+    // Set the system property to enable logging of server side warnings
+    final String originalLoggingFlag =
+        System.getProperty(RequestHandler.LOG_REQUEST_WARNINGS_PROPERTY, "false");
+    System.setProperty(RequestHandler.LOG_REQUEST_WARNINGS_PROPERTY, "true");
```
olim7t: Normally tests that manipulate system properties are executed in isolation, but in this case the impact on potentially concurrent tests is insignificant, so I agree with your approach.

driver-core/src/test/java/com/datastax/driver/core/WarningsTest.java
```diff
@@ -50,4 +54,52 @@ public void should_expose_warnings_on_execution_info() {
...
+    assertThat(logAppender.logs).isNotEmpty();
+    assertThat(logAppender.logs.get(0).getMessage())
+        .isEqualTo("Aggregation query used without partition key");
+    // remove the log appender
+    Logger.getRootLogger().removeAppender(logAppender);
+    // reset the logging flag
+    System.setProperty(RequestHandler.LOG_REQUEST_WARNINGS_PROPERTY, originalLoggingFlag);
+  }
+
+  private class TestAppender extends WriterAppender {
```
olim7t: We already have a `MemoryAppender` class that works this way, and handles a couple of corner cases we ran into over time. I think you could use it here.

driver-core/src/test/java/com/datastax/driver/core/WarningsTest.java
```diff
@@ -50,4 +54,52 @@ public void should_expose_warnings_on_execution_info() {
     List<String> warnings = rs.getExecutionInfo().getWarnings();
     assertThat(warnings).hasSize(1);
   }
+
+  @Test(groups = "short")
+  public void should_execute_query_with_server_side_warnings() {
```
olim7t: Good test 👍

manual/logging/README.md
```diff
@@ -192,6 +192,17 @@ that can significantly boost latencies when writing log messages.
 without stopping the application. This usually involves JMX and is available for [Logback](http://logback.qos.ch/manual/jmxConfig.html);
 Log4J provides a `configureAndWatch()` method but it is not recommended to use it inside J2EE containers (see [FAQ](https://logging.apache.org/log4j/1.2/faq.html#a3.6)).
 
+### Server Side Warnings
```
olim7t, line 195: 💯 for updating the manual

manual/logging/README.md
```diff
@@ -192,6 +192,17 @@ that can significantly boost latencies when writing log messages.
...
+
+When using the driver to execute queries, it is possible that the server will generate a warning. Normally,
+these warnings are contained in the `ResultSet`'s `ExecutionInfo` attribute, but are not logged by the driver
```
olim7t: Nit: this sounds like sometimes the warnings are not in `ExecutionInfo`. It would be better to say "These warnings are contained ... but are normally not logged..."

Though this paragraph will change anyway if you follow my other suggestion to invert the boolean.

olim7t (changes requested):

changelog/README.md
```diff
@@ -2,6 +2,7 @@
 
 ### 3.7.0 (In progress)
 
+- [improvement] JAVA-1950: Log server side warnings returned from a query.
```
adutra: Nit: for legacy drivers, changelog entries should be placed after the existing ones, not before.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -224,6 +226,11 @@ private void setFinalResult(
                 response.getCustomPayload());
       }
       callback.onSet(connection, response, info, statement, System.nanoTime() - startTime);
+      // if the response from the server has warnings, they'll be set on the ExecutionInfo. Log them
+      // here, unless they've been disabled.
+      if (logger.isWarnEnabled() && response.warnings != null && !response.warnings.isEmpty()) {
```
adutra: This is a bit unrelated, but why aren't the warnings set here on the `ExecutionInfo` object? This seems like the most logical place to do it.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -234,6 +241,23 @@ private void setFinalResult(
     }
   }
 
+  private void logServerWarnings(List<String> warnings) {
+    // log the warnings if they have NOT been disabled (double negative)
+    if (!Boolean.getBoolean(RequestHandler.DISABLE_QUERY_WARNING_LOGS)) {
```
adutra: Since you already guarded the call to this method by a few conditional checks above, I'd suggest that you add this one to the `if` block above too.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -64,6 +64,8 @@
 
   private static final boolean HOST_METRICS_ENABLED =
       Boolean.getBoolean("com.datastax.driver.HOST_METRICS_ENABLED");
+  static final String DISABLE_QUERY_WARNING_LOGS = "com.datastax.driver.DISABLE_QUERY_WARNING_LOGS";
+  private static final int MAX_QUERY_LOG_LENGTH = 50;
```
adutra: 50 seems a bit short imo. Besides, we already have `com.datastax.driver.core.QueryLogger#DEFAULT_MAX_QUERY_STRING_LENGTH`, maybe we should simply reuse that.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -234,6 +241,23 @@ private void setFinalResult(
     }
   }
 
+  private void logServerWarnings(List<String> warnings) {
+    // log the warnings if they have NOT been disabled (double negative)
+    if (!Boolean.getBoolean(RequestHandler.DISABLE_QUERY_WARNING_LOGS)) {
+      // comma separated list of warnings
+      StringBuilder warningString = new StringBuilder();
```
adutra: Use Guava's `Joiner` instead.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -234,6 +241,23 @@ private void setFinalResult(
...
+      for (String warning : warnings) {
+        warningString.append(warning).append(", ");
+      }
+      // truncate the statement query to the MAX_QUERY_LOG_LENGTH, if necessary
+      final String queryString = statement.toString();
+      logger.warn(
```
adutra: I wonder if we shouldn't print one message per warning; this way logs can be more easily grep'ed or post-processed.

driver-core/src/test/java/com/datastax/driver/core/WarningsTest.java
```diff
@@ -49,5 +69,80 @@ public void should_expose_warnings_on_execution_info() {
...
+
+  @Test(groups = "short")
+  public void should_execute_query_and_not_log_server_side_warnings() {
+    // Get the system property value for disabling logging server side warnings
+    final String disabledLogFlag =
+        System.getProperty(RequestHandler.DISABLE_QUERY_WARNING_LOGS, "false");
+    // assert that logs are NOT disabled
+    assertThat(disabledLogFlag).isEqualTo("false");
+    // Disable the logs
+    System.setProperty(RequestHandler.DISABLE_QUERY_WARNING_LOGS, "true");
```
adutra, line 125: Usually when system properties need to be modified, we use the special test group `isolated`.

driver-core/src/test/java/com/datastax/driver/core/WarningsTest.java
```diff
@@ -49,5 +69,80 @@ public void should_expose_warnings_on_execution_info() {
...
+    List<String> warnings = ei.getWarnings();
+    assertThat(warnings).isNotEmpty();
+    assertThat(warnings.size()).isEqualTo(1);
+    assertThat(warnings.get(0)).isEqualTo("Aggregation query used without partition key");
+    // And the driver di NOT log the server side warning
+    String log = logAppender.getNext();
+    assertThat(log).isNullOrEmpty();
+
+    // reset the logging flag
+    System.setProperty(RequestHandler.DISABLE_QUERY_WARNING_LOGS, disabledLogFlag);
```
adutra: For extra safety I suggest that you place this instruction in a `finally` block.

manual/logging/README.md
```diff
@@ -192,6 +192,47 @@ that can significantly boost latencies when writing log messages.
...
+to "true". This can be done at application startup (`-Dcom.datastax.driver.DISABLE_QUERY_WARNING_LOGS=true`)
+or it can be toggled programmatically in application code:
+
+```java
+// disable driver logging of server-side warnings
+System.getProperty("com.datastax.driver.DISABLE_QUERY_WARNING_LOGS", "true");
+....
+// enable driver logging of server-side warnings
+System.getProperty("com.datastax.driver.DISABLE_QUERY_WARNING_LOGS", "false");
+```
```
adutra, line 234: Very nice docs 👍

manual/logging/README.md
```diff
@@ -192,6 +192,47 @@ that can significantly boost latencies when writing log messages.
...
+
+Sometimes, it is not desirable for the driver to log server-side warnings. In such cases, logging
+these warnings can be disabled in the driver by setting the system property `com.datastax.driver.DISABLE_QUERY_WARNING_LOGS`
+to "true". This can be done at application startup (`-Dcom.datastax.driver.DISABLE_QUERY_WARNING_LOGS=true`)
+or it can be toggled programmatically in application code:
+
+```java
+// disable driver logging of server-side warnings
+System.getProperty("com.datastax.driver.DISABLE_QUERY_WARNING_LOGS", "true");
```
adutra: This should rather be `System.setProperty`.

adutra (changes requested):

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
adutra: Suggestion: maybe you could reuse some of the `QueryLogger` functionality and call `QueryLogger.statementAsString()` to format the statement.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
emerkle826: The reason I am not using the ExectionInfo instance here is that the warnings may not have been set on it at this point. It seems that ExecutionInfo instance is updated with the warnings during processing of the callback. If I want to use the ExecutionInfo directly here, I believe I would have to wait for the callback Future to complete.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
emerkle826: I think I had it that way at some point. Good call.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
emerkle826: I decided to log each warning in the list separately, so there is no need for joining anymore.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -234,6 +243,15 @@ private void setFinalResult(
     }
   }
 
+  private void logServerWarnings(List<String> warnings) {
+    // truncate the statement query to the DEFAULT_MAX_QUERY_STRING_LENGTH, if necessary
+    final String queryString = QueryLogger.builder().build().statementAsString(statement);
```
adutra: Maybe we could store `QueryLogger.builder().build()` as a constant somewhere?

adutra (approved):

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -86,6 +87,7 @@
 
   private final AtomicBoolean isDone = new AtomicBoolean();
   private final AtomicInteger executionIndex = new AtomicInteger();
+  private final QueryLogger queryLogger = QueryLogger.builder().build();
```
adutra: This change affects the hot path, so I would rather not create an instance of `QueryLogger` for every request; can you please turn this field into a static one?

olim7t (approved):

olim7t:
I realized that my comment about `logger.isWarnEnabled()` from the 4.x PR also applies here. I pushed a small commit to move it in the last position in the test.

PR #1180, opened 2019-01-30
JAVA-2113: Generate SetEntity DAO methods

olim7t:
Now that the base injection logic is handled by `EntityHelper`, this ticket is very simple: the DAO just needs to create an instance of the helper (cached in a field), and delegate to its `inject` method.

Sample code to test:
```java
@Entity
class User { ... }

@Dao
public interface UserDao {
  @EntityInjector
  void inject(User user, BoundStatement boundStatement);
}
```
(I'm aware that automatic tests are lacking so far, it's on my todo list)

### DIFF
--- a/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
+++ b/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
@@ -15,19 +15,32 @@
  */
 package com.datastax.oss.driver.internal.mapper.processor.dao;
 
+import com.datastax.oss.driver.api.mapper.annotations.SetEntity;
+import com.datastax.oss.driver.api.mapper.entity.EntityHelper;
 import com.datastax.oss.driver.internal.core.util.concurrent.BlockingOperation;
 import com.datastax.oss.driver.internal.core.util.concurrent.CompletableFutures;
 import com.datastax.oss.driver.internal.mapper.MapperContext;
 import com.datastax.oss.driver.internal.mapper.processor.GeneratedNames;
 import com.datastax.oss.driver.internal.mapper.processor.ProcessorContext;
 import com.datastax.oss.driver.internal.mapper.processor.SingleFileCodeGenerator;
+import com.datastax.oss.driver.internal.mapper.processor.SkipGenerationException;
+import com.datastax.oss.driver.internal.mapper.processor.util.NameIndex;
 import com.squareup.javapoet.ClassName;
+import com.squareup.javapoet.CodeBlock;
 import com.squareup.javapoet.FieldSpec;
 import com.squareup.javapoet.JavaFile;
 import com.squareup.javapoet.MethodSpec;
 import com.squareup.javapoet.ParameterizedTypeName;
 import com.squareup.javapoet.TypeSpec;
+import java.beans.Introspector;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 import java.util.concurrent.CompletableFuture;
+import javax.lang.model.element.Element;
+import javax.lang.model.element.ElementKind;
+import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.TypeElement;
 
@@ -35,11 +48,32 @@ public class DaoImplementationGenerator extends SingleFileCodeGenerator {
 
   private final TypeElement interfaceElement;
   private final ClassName implementationName;
+  private final NameIndex nameIndex = new NameIndex();
+  private final Map<ClassName, String> entityHelperFields = new HashMap<>();
 
   public DaoImplementationGenerator(TypeElement interfaceElement, ProcessorContext context) {
     super(context);
     this.interfaceElement = interfaceElement;
     implementationName = GeneratedNames.daoImplementation(interfaceElement);
+  }
+
+  /**
+   * Requests the generation of a field holding the {@link EntityHelper} that was generated for the
+   * given entity class, along with the initialization code in the constructor.
+   *
+   * <p>If this is called multiple times, only a single field will be created.
+   *
+   * @return the name of the field.
+   */
+  String addEntityHelperField(TypeElement entityClass) {
+    ClassName helperClass = GeneratedNames.entityHelper(entityClass);
+    return entityHelperFields.computeIfAbsent(
+        helperClass,
+        k -> {
+          String baseName =
+              Introspector.decapitalize(entityClass.getSimpleName().toString()) + "Helper";
+          return nameIndex.uniqueField(baseName);
+        });
   }
 
   @Override
@@ -50,6 +84,20 @@ public class DaoImplementationGenerator extends SingleFileCodeGenerator {
   @Override
   protected JavaFile.Builder getContents() {
 
+    List<MethodSpec.Builder> methods = new ArrayList<>();
+    for (Element child : interfaceElement.getEnclosedElements()) {
+      try {
+        if (child.getKind() == ElementKind.METHOD) {
+          ExecutableElement methodElement = (ExecutableElement) child;
+          if (methodElement.getAnnotation(SetEntity.class) != null) {
+            methods.add(new DaoSetEntityMethodGenerator(methodElement, this, context).generate());
+          }
+          // TODO handle other annotations
+        }
+      } catch (SkipGenerationException ignored) {
+      }
+    }
+
     TypeSpec.Builder classBuilder =
         TypeSpec.classBuilder(implementationName)
             .addJavadoc(JAVADOC_GENERATED_WARNING)
@@ -59,26 +107,9 @@ public class DaoImplementationGenerator extends SingleFileCodeGenerator {
                 FieldSpec.builder(MapperContext.class, "context", Modifier.PRIVATE, Modifier.FINAL)
                     .build());
 
-    MethodSpec.Builder initAsyncBuilder =
-        MethodSpec.methodBuilder("initAsync")
-            .returns(
-                ParameterizedTypeName.get(
-                    ClassName.get(CompletableFuture.class), ClassName.get(interfaceElement)))
-            .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
-            .addParameter(MapperContext.class, "context")
-            .addStatement(
-                "return $T.completedFuture(new $T(context))",
-                CompletableFuture.class,
-                implementationName);
+    MethodSpec.Builder initAsyncBuilder = getInitAsyncContents();
 
-    MethodSpec.Builder initBuilder =
-        MethodSpec.methodBuilder("init")
-            .returns(ClassName.get(interfaceElement))
-            .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
-            .addParameter(MapperContext.class, "context")
-            .addStatement("$T.checkNotDriverThread()", BlockingOperation.class)
-            .addStatement(
-                "return $T.getUninterruptibly(initAsync(context))", CompletableFutures.class);
+    MethodSpec.Builder initBuilder = getInitContents();
 
     MethodSpec.Builder constructorBuilder =
         MethodSpec.constructorBuilder()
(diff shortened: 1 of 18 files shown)


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -50,6 +84,20 @@ protected String getFileName() {
   @Override
   protected JavaFile.Builder getContents() {
 
+    List<PartialClassGenerator> methodGenerators = new ArrayList<>();
```
olim7t: I will probably remove `PartialClassGenerator` soon: it's not very flexible, I prefer the pattern where the class generator passes itself to child generators and exposes higher-level hooks, such as `addEntityHelperField()` above.

It's an easy refactoring so it can be done later.

tomekl007:
> (I'm aware that automatic tests are lacking so far, it's on my todo list)

How do we plan to test mapper code? 
1. Do we want to examine the generated code line by line?
2. Or maybe we could use generated code and assert that it works as expected?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -15,33 +15,67 @@
...
     super(context);
     this.interfaceElement = interfaceElement;
     implementationName = GeneratedNames.daoImplementation(interfaceElement);
   }
 
+  /**
+   * Requests the generation of a field holding the {@link EntityHelper} that was generated for the
+   * given entity class, along with the initialization code in the constructor.
+   *
+   * <p>If this is called multiple times, only a single field will be created.
```
tomekl007, line 64: Maybe better to state that only the first call will trigger computation?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -15,33 +15,67 @@
...
+   *
+   * @return the name of the field.
+   */
+  String addEntityHelperField(TypeElement entityClass) {
+    ClassName helperClass = GeneratedNames.entityHelper(entityClass);
+    return entityHelperFields.computeIfAbsent(
+        helperClass,
+        k -> {
+          String baseName =
+              Introspector.decapitalize(entityClass.getSimpleName().toString()) + "Helper";
```
tomekl007, line 74: The name will be generated in a camel-case pattern?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -50,6 +84,20 @@ protected String getFileName() {
...
 
+    List<PartialClassGenerator> methodGenerators = new ArrayList<>();
+    for (Element child : interfaceElement.getEnclosedElements()) {
+      try {
+        if (child.getKind() == ElementKind.METHOD) {
+          ExecutableElement methodElement = (ExecutableElement) child;
+          if (methodElement.getAnnotation(EntityInjector.class) != null) {
+            methodGenerators.add(new EntityInjectorGenerator(methodElement, this, context));
+          }
+          // TODO handle other annotations
```
tomekl007, line 95: Is there a need to handle other annotations here? I see that this class is called `DaoImplementationGenerator` so it should it handle only `@Dao` annotation, right? If we will want to handle other annotations here it will no be longer `DaoGenerator`.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -86,6 +134,20 @@ protected String getFileName() {
...
+      methodGenerator.addConstructorInstructions(constructorBuilder);
+    }
+
+    for (Map.Entry<ClassName, String> entry : entityHelperFields.entrySet()) {
+      ClassName fieldTypeName = entry.getKey();
+      String fieldName = entry.getValue();
+
+      classBuilder.addField(
+          FieldSpec.builder(fieldTypeName, fieldName, Modifier.PRIVATE, Modifier.FINAL).build());
+      constructorBuilder.addStatement("this.$L = new $T(context)", fieldName, fieldTypeName);
```
tomekl007: What will be a type of a `context` argument of the generated constructor?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/EntityInjectorGenerator.java
```diff
@@ -0,0 +1,115 @@
...
+  public EntityInjectorGenerator(
+      ExecutableElement methodElement,
+      DaoImplementationGenerator daoImplementationGenerator,
+      ProcessorContext context) {
+    this.methodElement = methodElement;
+    this.daoImplementationGenerator = daoImplementationGenerator;
+    // We're expecting a void method, where one parameter is an annotated entity, and the other a
+    // SettableByName or subtype
+    TypeMirror returnType = methodElement.getReturnType();
+    if (returnType.getKind() != TypeKind.VOID || methodElement.getParameters().size() != 2) {
```
tomekl007: maybe we can name that condition like `isVoidMethodAndTwoHasParameters` and extract such method?
Also, I don't see a check for:
>one parameter is an annotated entity, and the other a
>    // SettableByName or subtype

Only validation on a number of arguments. I see that it is checked after, maybe we can do a pre-validation or all pre-conditions needed at the entry of this processing?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/EntityInjectorGenerator.java
```diff
@@ -0,0 +1,115 @@
...
+      context
+          .getMessager()
+          .error(
+              methodElement,
+              "%s methods must have two parameters and a void return type",
+              EntityInjector.class.getSimpleName());
+      throw new SkipGenerationException();
```
tomekl007: Don't we want to include a message in the `SkipGenerationException`?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/EntityInjectorGenerator.java
```diff
@@ -0,0 +1,115 @@
...
+        MethodSpec.methodBuilder(methodElement.getSimpleName().toString())
+            .addAnnotation(Override.class)
+            .addModifiers(Modifier.PUBLIC);
+    for (VariableElement parameterElement : methodElement.getParameters()) {
+      overridingMethodBuilder.addParameter(
+          ClassName.get(parameterElement.asType()), parameterElement.getSimpleName().toString());
+    }
+    // Forward to the base injector in the helper:
+    overridingMethodBuilder.addStatement(
+        "$L.inject($L, $L)", helperFieldName, entityParameterName, targetParameterName);
```
tomekl007: what the `inject` call means? We are using some already existing code?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
GregBestland: I see a decapatlize here am I missing something?.. why would this be camel case?

olim7t: Yes, it lower-cases the first letter.
```
Product => productHelper
SomeClass => someClassHelper
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
olim7t: The generated class will implement all the methods declared on the interface, so we'll eventually handle all "query" annotations here.

```java
@Dao
public interface UserDao {
  @EntityInject // <--
  void inject(User user, BoundStatement statement);

  @Insert // <--
  void save(User user);

  @Select // <--
  User findById(UUID id);
}
```
The loop iterates on the elements _inside_ `UserDao`.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
olim7t: `MapperContext`. It's in the `mapper-runtime` module.

olim7t:
> How do we plan to test mapper code?

I created a ticket to explore a few ideas: [JAVA-2131](https://datastax-oss.atlassian.net/browse/JAVA-2131)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/EntityInjectorGenerator.java
olim7t: As we implement more query methods we can extract common patterns if we see them emerge. Right now I'm not sure if `isVoidMethodAndTwoHasParameters` is going to be useful anywhere else.

The check for `SettableByName` is at line 65:
```java
      if (context.getClassUtils().implementsSettableByName(parameterType)) {
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/EntityInjectorGenerator.java
olim7t: The message is passed to `context.getMessager()` on the previous line. It will materialize as a compiler error. It's a bit more convenient to do it here because we have all the context needed, in particular the `Element`, which determines the line of the error.

`SkipGenerationException` is just a simple way to abort quickly when we know that the generation won't succeed. The convention is that the code that throws it should take care of any message, the catching code will just ignore it and proceed (in that case it's `DaoImplementationGenerator.java:97`). It's a bit hacky, but convenient.

Note that we can never throw exceptions from processor code, errors have to be reported via the messager.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/EntityInjectorGenerator.java
olim7t: Yeah that's a bit hard to follow, it's code from another generated class.

For each `@Entity`, the processor now generates an implementation of `EntityHelper` (in mapper-runtime), that defines `inject()`.
Upon encountering the entity, this class requests that the generated DAO code create an instance of this helper, and store it in a field (line 94):
```java
String helperFieldName = daoImplementationGenerator.addEntityHelperField(entityElement);
```
Then it generates this line of code that delegates to this helper field.

Here's what the generated code looks like:
```java
public class UserDao_Impl implements UserDao {
  private final MapperContext context;

  private final User_Helper userHelper; // The helper field

  private UserDao_Impl(MapperContext context) {
    this.context = context;
    this.userHelper = new User_Helper(context);
  }

  // This is the part that's generated by EntityInjectorGenerator:
  @Override
  public void inject(User user, BoundStatement boundStatement) {
    userHelper.inject(user, boundStatement); // delegate to the helper
  }

  ... // (init methods, not relevant here)
}
```

olim7t:
I pushed a few changes. The DAO implementation class now creates entity helper instances in its `initAsync` method:

```java
  private final User_Helper userHelper;

  private final Address_Helper addressHelper;

  private UserDao_Impl(MapperContext context, User_Helper userHelper,
      Address_Helper addressHelper) {
    this.context = context;
    this.userHelper = userHelper;
    this.addressHelper = addressHelper;
  }

  public static CompletableFuture<UserDao> initAsync(MapperContext context) {
    User_Helper userHelper = new User_Helper(context);
    Address_Helper addressHelper = new Address_Helper(context);
    UserDao_Impl dao = new UserDao_Impl(context,
        userHelper,
        addressHelper);
    return CompletableFuture.completedFuture(dao);
  }
```
This will make sense when we generate methods that need prepared statements.

olim7t:
I just remembered that our bound statements are immutable 🤦‍♂️ `inject` needs to return the `SettableByName`...
 I'll push the required changes shortly.

GregBestland (approved):

PR #1181, opened 2019-01-31
Allow ControlConnection to initialize if reconnectOnFailure is false

emerkle826:
Since JAVA-2077, initialization of the ControllConnection in `DefaultTopologyMonitor` will fail if `advanced.reconnect-on-init` is set to `false`, which is the default in reference.conf/application.conf. This is because the connection is always initialized with `useInitialReconnectionSchedule` set to true.

There is a precondition check that fails the ControlConnection initialization only when `reconnectOnFailure` (set from `advanced.reconnect-on-init` in the configuration) is `false` and `useInitialReconnectionSchedule` is true. All other combinations pass the precondition. However, the documentation in `ControlConnection.init()` suggests that `useInitialReconnectionSchedule` will simply be ignored if `reconnectOnFailure` is `false`.

In this PR, I have 2 proposed solutions:
1) Change the precondition check to a simple warning log (so the initialization does not fail)
2) Change DefaultTopologyMonitor to only set `useInitialReconnectionSchedule` to `true` if  `reconnectOnFailure` is `true`

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
@@ -39,7 +39,6 @@ import com.datastax.oss.driver.internal.core.util.concurrent.CompletableFutures;
 import com.datastax.oss.driver.internal.core.util.concurrent.Reconnection;
 import com.datastax.oss.driver.internal.core.util.concurrent.RunOrSchedule;
 import com.datastax.oss.driver.internal.core.util.concurrent.UncaughtExceptions;
-import com.datastax.oss.driver.shaded.guava.common.base.Preconditions;
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
 import com.datastax.oss.protocol.internal.Message;
 import com.datastax.oss.protocol.internal.ProtocolConstants;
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -89,7 +89,9 @@ public DefaultTopologyMonitor(InternalDriverContext context) {
     if (closeFuture.isDone()) {
       return CompletableFutures.failedFuture(new IllegalStateException("closed"));
     }
-    return controlConnection.init(true, reconnectOnInit, true);
+    // we only want to us the initial reconnection schedule here if reconnectOnInit is true
+    final boolean useInitialReconnectionSchedule = reconnectOnInit;
+    return controlConnection.init(true, reconnectOnInit, useInitialReconnectionSchedule);
```
emerkle826: We do not need this change if we should always set `useInitialReconnectionSchedule` to true and we simply log a warning when `reconnectOnInit` is `false`. However, I'm not certain on if/when we should set `useInitialReconnectionSchedule` to `true` here. Perhaps there is more logic that needs to be evaluated?

emerkle826: After discussing this more, we're simply going to remove the Precondition, as it's not an abnormal situation (as the javadocs suggest).

olim7t (approved):

PR #1182, opened 2019-01-31
Reformat test files to comply with updated contribution guidelines

adutra:
Structural Search & Replace didn't work unfortunately (weird behavior on IntelliJ 2018). However a regex search and replace followed by a global file reformat did the trick.

There is nothing else in this PR than the result of these search and replace + reformat operations.


olim7t (approved):
I didn't verify every change (obviously), but LGTM

PR #1183, opened 2019-02-04
JAVA-2115: Generate Insert DAO methods

olim7t:


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -0,0 +1,186 @@
...
+   * @param cqlName the CQL name to set ({@code "id"})
+   * @param type the Java type of the value ({@code UUID})
+   * @param entityElement if the value is a mapped entity, the corresponding element
+   * @param valueExtractor the code snippet to extract the value ({@code entity.getId()}
+   * @param targetName the name of the target {@link SettableByName} instance ({@code target})
+   * @param methodBuilder where to add the code
+   * @param enclosingClass a reference to the parent generator (in case type constants or entity
+   *     helpers are needed)
+   */
+  public static void setValue(
```
olim7t, line 98: This is now shared between `EntityHelperSetMethodGenerator` and `DaoInsertMethodGenerator`.

This makes the code a bit harder to read because a lot of aspects are parameterized. But it's worth it because it's a central feature that will get reused by other query methods (`@Delete`, `@Query`, etc). [JAVA-2129](https://datastax-oss.atlassian.net/browse/JAVA-2129) will add more complexity, we'll only need to do it in one place.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/model/inventory/ProductDao.java
```diff
@@ -32,4 +33,10 @@
 
   @SetEntity
   void set(Dimensions dimensions, UdtValue udtValue);
+
+  @Insert
+  void save(Product product);
+
+  @Insert(customClause = " USING TIMESTAMP :timestamp")
+  Product saveWithBoundTimestamp(Product product, long timestamp);
```
olim7t, line 62: Custom clause example. The second argument is injected as a bind variable with the same name as the Java parameter (case-insensitive). The only catch is that it must not collide with any of the entity's properties.
We could introduce an annotation to customize the name, although I don't feel a strong need for that.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -156,34 +211,59 @@ protected String getFileName() {
...
+
+    initAsyncBuilder
+        .addComment("Build the DAO when all statements are prepared")
+        .addCode("$[return $T.allSuccessful(prepareStages)", CompletableFutures.class)
+        .addCode("\n.thenApply(v -> ($T) ", interfaceElement)
+        .addCode(newDaoStatement.build())
+        .addCode(")\n$<$<.toCompletableFuture();$]\n")
+        .nextControlFlow("catch ($T t)", Throwable.class)
+        .addStatement("return $T.failedFuture(t)", CompletableFutures.class)
+        .endControlFlow();
```
olim7t, line 270: See `ProductDao_Impl` in test generated sources for an example of this.

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/CompletableFutures.java
```diff
@@ -72,6 +73,40 @@
       List<CompletionStage<T>> inputs, Runnable callback, Executor executor) {
     allDone(inputs).thenRunAsync(callback, executor).exceptionally(UncaughtExceptions::log);
   }
+  /**
+   * @return a completion stage that completes when all inputs are successful, or fails if any of
+   *     them failed.
+   */
+  public static <T> CompletionStage<Void> allSuccessful(List<CompletionStage<T>> inputs) {
```
tomekl007, line 80: Do we use some different implementation of `CompletionStage` than `CompletableFuture`?
If no why don't just use `CompletableFuture.allOf()`?

integration-tests/src/test/java/com/datastax/oss/driver/mapper/model/inventory/ProductDao.java
```diff
@@ -32,4 +33,10 @@
 
   @SetEntity
   void set(Dimensions dimensions, UdtValue udtValue);
+
+  @Insert
+  void save(Product product);
+
+  @Insert(customClause = " USING TIMESTAMP :timestamp")
```
tomekl007: is this white space before `USING` needed? If yes I think that it may be error-prone and we should take 
`"USING ...` as an argument and take care of inserting whitespace.

olim7t: No it's not needed, the processor code does indeed add a whitespace. That's a typo, I'll remove the whitespace here.

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/CompletableFutures.java
olim7t: Good catch, it didn't occur to me to check if the JDK had a built-in method for this version.

However, `CompletableFuture.allOf`:
* deals with `CompletableFuture` instances, not `CompletionStage`, so we would have to convert each one with `toCompletableFuture()`
* only takes a vararg parameter. In our generated code it's a list, so we'd have to add a `toArray` call.
* only reports the first failure that it encounters. I think it's nice to have them all at once, so that if there are multiple errors you have an opportunity to fix them all at once.

PR #1184, opened 2019-02-06
JAVA-2106: Log server side warnings returned from a query

emerkle826:


core/src/main/resources/reference.conf
```diff
@@ -759,6 +759,16 @@ datastax-java-driver {
...
       consistency = ONE
     }
+
+    # Whether logging of server warnings generated during query execution should be disabled by the
+    # driver. All server generated warnings will be available programmatically via the ExecutionInfo
+    # object on the executed statement's ResultSet. If set to "true", this will prevent the driver
+    # from logging these warnings.
+    #
+    # Required: yes
+    # Modifiable at runtime: yes, the new value will be used for traces fetched after the change.
```
adutra: Nit: "for traces fetched" -> "for query warnings received"

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
```diff
@@ -0,0 +1,224 @@
...
+                      DefaultDriverConfigLoaderBuilder.profileBuilder()
+                          .withString(DefaultDriverOption.REQUEST_DISABLE_LOG_WARNINGS, "true")
+                          .build())
+                  .build())
+          .build();
+
+  private static final SessionRule<CqlSession> SIMULACRON_SESSION_RULE =
+      SessionRule.builder(SIMULACRON).build();
+
+  @ClassRule public static TestRule CCM_CHAIN = RuleChain.outerRule(CCM).around(SESSION_RULE);
```
adutra: Nit: this isn't final so `ccmChain` would be more appropriate.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
```diff
@@ -0,0 +1,224 @@
...
+
+  @ClassRule
+  public static TestRule simulacronChain =
```
adutra: Do we really need Simulacron in this test class? It doesn't seem used in the tests below.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
```diff
@@ -0,0 +1,224 @@
...
+            Strings.repeat("1", 2 * 1024), Strings.repeat("1", 3 * 1024));
+    Statement<?> st = SimpleStatement.builder(String.format(query)).build();
+    ResultSet result = SESSION_RULE.session().execute(st);
+    ExecutionInfo executionInfo = result.getExecutionInfo();
+    assertThat(executionInfo).isNotNull();
+    List<String> warnings = executionInfo.getWarnings();
+    assertThat(warnings).isNotNull();
+    assertThat(warnings).isNotEmpty();
+    // verify the log was generated
+    Mockito.verify(appender, after(500).times(1)).doAppend(loggingEventCaptor.capture());
```
adutra: The contribution guidelines changed, you should now static-import all Mockito methods.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
```diff
@@ -0,0 +1,224 @@
...
+    assertThat(warning).isEqualTo("Aggregation query used without partition key");
+    // verify the log was NOT generated
+    Mockito.verify(appender, after(500).times(0)).doAppend(loggingEventCaptor.capture());
+  }
+
+  @Test
+  public void should_expose_warnings_on_execution_info() {
+    // the default batch size warn threshold is 5 * 1024 bytes, but after CASSANDRA-10876 there must
+    // be
+    // multiple mutations in a batch to trigger this warning so the batch includes 2 different
```
adutra: Nit: line wrappings could be improved here.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
```diff
@@ -0,0 +1,224 @@
...
+                + "INSERT INTO test (k, v) VALUES (1, '%s')\n"
+                + "INSERT INTO test (k, v) VALUES (2, '%s')\n"
+                + "APPLY BATCH",
+            Strings.repeat("1", 2 * 1024), Strings.repeat("1", 3 * 1024));
+    Statement<?> st = SimpleStatement.builder(String.format(query)).build();
+    ResultSet result = SESSION_RULE.session().execute(st);
+    ExecutionInfo executionInfo = result.getExecutionInfo();
+    assertThat(executionInfo).isNotNull();
+    List<String> warnings = executionInfo.getWarnings();
+    assertThat(warnings).isNotNull();
```
adutra: `isNotNull` is not required if `isNotEmpty` is present.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
```diff
@@ -0,0 +1,224 @@
...
+    assertThat(loggingEventCaptor.getValue().getMessage()).isNotNull();
+    String logMessage = loggingEventCaptor.getValue().getFormattedMessage();
+    assertThat(logMessage)
+        .startsWith("Query '")
+        // query will only be logged up to DefaultDriverOption.REQUEST_LOGGER_MAX_QUERY_LENGTH
+        // characters
+        .contains(
+            query.substring(
+                0,
+                SESSION_RULE
```
adutra: Maybe we car replace this with a constant `500` since we now that that's the value we set above?

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
```diff
@@ -188,6 +190,7 @@ protected CqlRequestHandler(
 
     this.throttler = context.getRequestThrottler();
     this.throttler.register(this);
+    this.requestLogFormatter = new RequestLogFormatter(context);
```
emerkle826: @adutra You commented in #1178 that the QueryLogger should be made static to avoid object instantiation on the hot path. I believe this RequestLogFormatter is on a similar hot path here. However, I'm not sure if there is an easy way to avoid it as the RequestLogFormatter needs an instance of DriverContext, which is only set when this request handler is instantiated. Ideas?

core/src/main/resources/reference.conf
emerkle826: whoops, copy-paste error on my part.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
emerkle826: We don't. This is from copying too much from another test and not pruning back. Thanks for the catch!

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
adutra: You are absolutely right. I see that `RequestLogFormatter` only needs the context to access the codec registry and the protocol version. So I see two options: 

1. Do nothing, maybe it's not that bad to instantiate that component at every request; or
2. Create another constructor in `RequestLogFormatter` that takes a codec registry and a protocol version, and use default values for both (this could break if the user is using custom types).

In the end, I think I'd go with option 1.

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
tomekl007: Maybe `RequestLogFormatter` should belong to the `DriverContext`?
This way when the `DriverContext` is created we will create `RequestLogFormatter` only once and we will not need to instantiate it in the `CqlRequestHandler` constructor. Then we can call `context.getRequestLogFormatter` and assign it to this `CqlRequestHandler` - similarly to `this.requestTracker = context.getRequestTracker();`

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
```diff
@@ -356,11 +359,34 @@ private void setFinalResult(
...
       setFinalError(error, callback.node, -1);
     }
   }
 
+  private void logServerWarnings(List<String> warnings) {
+    // use the RequestLogFormatter to format the query
+    StringBuilder statementString = new StringBuilder();
+    requestLogFormatter.appendRequest(
+        statement,
+        executionProfile.getInt(DefaultDriverOption.REQUEST_LOGGER_MAX_QUERY_LENGTH, 500),
```
tomekl007: those default values `500, false, 0, 0` are used also in other places, right?
(for example in `RequestLogger` `executionProfile.getInt(DefaultDriverOption.REQUEST_LOGGER_MAX_QUERY_LENGTH, 500);`)
So maybe we should refactor those Defaults as constants somewhere and use those constants in all places where that default is needed (to avoid duplication)?

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
```diff
@@ -356,11 +359,34 @@ private void setFinalResult(
               TimeUnit.NANOSECONDS);
         }
       }
+      // log the warnings if they have NOT been disabled
+      if (LOG.isWarnEnabled()
```
tomekl007: is there an option for this `LOG.isWarnEnabled()` to be changed dynamically?
If not then I think that we can evaluate those conditions only once and assign to a field (to avoid re-evaluation for every request)

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
```diff
@@ -0,0 +1,191 @@
...
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+import org.slf4j.LoggerFactory;
+
+@Category(ParallelizableTests.class)
+public class ExecutionInfoWarningsIT {
+
+  private static final int MAX_QUERY_LENGTH = 500;
```
tomekl007: another usage of default that could be moved and accessed here

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
```diff
@@ -0,0 +1,191 @@
...
+  private static final CcmRule CCM = CcmRule.getInstance();
+  private static final SessionRule<CqlSession> SESSION_RULE =
```
tomekl007, line 55: which version of DSE is picked for that test if we are not specifying it explicitly?

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
```diff
@@ -0,0 +1,191 @@
...
+    // setup the log appender
+    logger = (Logger) LoggerFactory.getLogger(CqlRequestHandler.class);
+    originalLoggerLevel = logger.getLevel();
+    logger.setLevel(Level.WARN);
+    logger.addAppender(appender);
+  }
+
+  @After
+  public void cleanupLogger() {
+    logger.setLevel(originalLoggerLevel);
```
tomekl007, line 109: Will this `@Before` and `@After` work properly together with `ParallelizableTests`? Isn't there a possiblity that this will work in a non-deterministic way?

core/src/main/resources/reference.conf
```diff
@@ -759,6 +759,16 @@ datastax-java-driver {
...
+
+    # Whether logging of server warnings generated during query execution should be disabled by the
+    # driver. All server generated warnings will be available programmatically via the ExecutionInfo
+    # object on the executed statement's ResultSet. If set to "true", this will prevent the driver
+    # from logging these warnings.
+    #
+    # Required: yes
+    # Modifiable at runtime: yes, the new value will be used for query warnings received after the change.
+    # Overridable in a profile: yes
+    disable-log-warnings = false
```
tomekl007: Do we want to expose that setting in a double negation way? (we are setting that it should NOT be disabled)
Maybe it would be better to have `enable-log-warning = true`?

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
emerkle826: It is possible to change the log level of a given logger in client code (I don't think any of the driver code currently does this). I'm not sure how often, if ever, it is done in production though.

core/src/main/resources/reference.conf
emerkle826: Good point. I'll change this in an update.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
olim7t: Side note: it should probably be final, and therefore named like a constant. But that's not hugely important, and existing tests don't follow a consistent convention. I've created [JAVA-2153](https://datastax-oss.atlassian.net/browse/JAVA-2153) to address it later.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
olim7t: +1 for exposing an instance of the log formatter in `InternalDriverContext`.

This is a simple class so the context can build it directly, no need to have the class name in the configuration. See how `getNettyOptions()` is implemented in `DefaultDriverContext`, it should be exactly the same.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
olim7t: We should assume that log levels can change at runtime. I don't know how common it is in practice, but it definitely makes sense to have a way to adjust them to debug integration/production issues (I would build that into a client application).

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
olim7t: Tests don't pick a version, they only define version requirements (`@CassandraRequirement`, `@DseRequirement`). We pick a version when we run the test suite, with the `ccm.version` system property; if a test has requirements that don't match this version, it doesn't run.

If there is a `CassandraRequirement` and we run with DSE, we pick compare using the closest matching Cassandra version. See `CcmBridge.getCassandraVersion()`. You can also take a look at `BaseCcmRule.apply()` for more details.

In this case the test has no requirements, so it will run with whichever DSE version you start it with.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
olim7t: Right, if the default level is `ERROR` and other tests also manipulate this particular level dynamically, there could be race conditions. The test should be serial (no `@Category` annotation, build the `CcmRule` with `CustomCcmRule.builder().build()`).

Now that you mention it, this could be an issue in the 1.x PR as well.

core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOption.java
```diff
@@ -171,6 +171,8 @@
 
   NETTY_TIMER_TICK_DURATION("advanced.netty.timer.tick-duration"),
   NETTY_TIMER_TICKS_PER_WHEEL("advanced.netty.timer.ticks-per-wheel"),
+
+  REQUEST_DISABLE_LOG_WARNINGS("advanced.request.disable-log-warnings"),
```
olim7t: 👍 for adding it last, otherwise it would break backward compatibility.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
```diff
@@ -356,11 +359,34 @@ private void setFinalResult(
...
+          && !executionInfo.getWarnings().isEmpty()
```
olim7t: Nit -- this might be premature optimization but it won't hurt: `!executionInfo.getWarnings().isEmpty()` is a simpler operation than `LOG.isWarnEnabled()` (if you look at how logback implements the latter). It's also probably the case for `executionProfile.getBoolean`. Could you flip the evaluation order?
```java
if (!executionInfo.getWarnings().isEmpty()
    && !executionProfile.getBoolean(DefaultDriverOption.REQUEST_DISABLE_LOG_WARNINGS)
    && LOG.isWarnEnabled()) {
```

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
```diff
@@ -356,11 +359,34 @@ private void setFinalResult(
...
+          && !executionProfile.getBoolean(DefaultDriverOption.REQUEST_DISABLE_LOG_WARNINGS)) {
+        logServerWarnings(executionInfo.getWarnings());
+      }
```
olim7t, line 365: Note: the `if (result.complete(resultSet)) { ... }` block above is in case another speculative execution beat us to completing the result.
We log the warnings outside of that block, which means we'll do so even if we raced. I think I'm OK with it but I thought I'd just mention it in passing.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
```diff
@@ -356,11 +359,34 @@ private void setFinalResult(
...
+        executionProfile.getInt(DefaultDriverOption.REQUEST_LOGGER_MAX_QUERY_LENGTH, 500),
+        executionProfile.getBoolean(DefaultDriverOption.REQUEST_LOGGER_VALUES, false),
+        executionProfile.getInt(DefaultDriverOption.REQUEST_LOGGER_MAX_VALUES, 0),
+        executionProfile.getInt(DefaultDriverOption.REQUEST_LOGGER_MAX_VALUE_LENGTH, 0),
+        statementString);
+    // log each warning separately
+    warnings.forEach(
+        (warning) -> {
+          LOG.warn("Query '{}' generated server side warning(s): {}", statementString, warning);
+        });
```
olim7t: Nit: unnecessary curly braces for this lambda.

core/src/main/resources/reference.conf
```diff
@@ -759,6 +759,16 @@ datastax-java-driver {
       # Overridable in a profile: yes
       consistency = ONE
     }
+
+    # Whether logging of server warnings generated during query execution should be disabled by the
+    # driver. All server generated warnings will be available programmatically via the ExecutionInfo
+    # object on the executed statement's ResultSet. If set to "true", this will prevent the driver
+    # from logging these warnings.
```
olim7t, line 766: We should also mention that it will reuse the formatting options in `advanced.request-tracker`.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
```diff
@@ -356,11 +359,34 @@ private void setFinalResult(
...
+  private void logServerWarnings(List<String> warnings) {
+    // use the RequestLogFormatter to format the query
+    StringBuilder statementString = new StringBuilder();
+    requestLogFormatter.appendRequest(
+        statement,
+        executionProfile.getInt(DefaultDriverOption.REQUEST_LOGGER_MAX_QUERY_LENGTH, 500),
+        executionProfile.getBoolean(DefaultDriverOption.REQUEST_LOGGER_VALUES, false),
+        executionProfile.getInt(DefaultDriverOption.REQUEST_LOGGER_MAX_VALUES, 0),
+        executionProfile.getInt(DefaultDriverOption.REQUEST_LOGGER_MAX_VALUE_LENGTH, 0),
+        statementString);
```
olim7t: 👍 for re-reading all options every time, we want them to be modifiable at runtime.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
```diff
@@ -0,0 +1,191 @@
...
+                          .withString(DefaultDriverOption.REQUEST_DISABLE_LOG_WARNINGS, "true")
+                          .build())
+                  .build())
+          .build();
+  private static final String KEY = "test";
+
+  @ClassRule public static TestRule ccmChain = RuleChain.outerRule(CCM).around(SESSION_RULE);
+
+  private final Appender<ILoggingEvent> appender =
+      (Appender<ILoggingEvent>) Mockito.mock(Appender.class);
```
olim7t: Nit: you can use Mockito annotations to avoid the unchecked cast warning on this line.

Annotate the class with `@RunWith(MockitoJUnitRunner.class)`, then make this field non-final, annotate it with `@Mock` and remove the initializer. Mockito will initialize it for you.
You can do the same for the captor on the next line:
```java
  @Mock private Appender<ILoggingEvent> appender;
  @Captor private ArgumentCaptor<ILoggingEvent> loggingEventCaptor;
```

olim7t (approved):
👍 , the warning logs look nice!

LGTM provided that you address two earlier comments: inverting the boolean option and exposing the formatter through the context.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
emerkle826: Will do.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
emerkle826: For the 1.x PR, I have the only test that changes the System property is marked with `isolated` here: https://github.com/datastax/java-driver/pull/1178/files#diff-7a9fae4b8210f01e97727cc6bf26e1ecR117
Though now that I look at it again, I'm not changing the log level in the 1.x PR.

tomekl007 (approved):

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
adutra: > The test should be serial (no `@Category` annotation, build the `CcmRule` with `CustomCcmRule.builder().build()`).

Hmm why use `CustomCmmRule` instead of `@Category`? This won't prevent the race condition. Imo we should add `@Category(IsolatedTests.class)` (but the CCM rule could be the non-custom one).

adutra (changes requested):
Had a doubt about whether or not to add `@Category(IsolatedTests.class)` to the test class.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
adutra: > Though now that I look at it again, I'm not changing the log level in the 1.x PR.

But you do set a system property, so it's OK.

adutra (approved):

PR #1185, opened 2019-02-07
Expose getMaxAttempts for ExponentialReconnectionPolicy and treat Void as a success response

tomekl007:


core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRequestHandler.java
```diff
@@ -60,7 +65,17 @@ public static AdminRequestHandler query(
         new Query(
             query,
             buildQueryOptions(pageSize, serialize(parameters, channel.protocolVersion()), null));
-    String debugString = "query '" + query + "'";
+    return createQuery(channel, message, parameters, timeout, logPrefix);
+  }
+
+  private static AdminRequestHandler createQuery(
```
adutra: Nit: could be renamed to `createAdminRequestHandler` – `createQuery` is a bit misleading.

core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRequestHandler.java
```diff
@@ -148,7 +163,9 @@ public void onResponse(Frame responseFrame) {
       ByteBuffer pagingState = rows.getMetadata().pagingState;
       AdminRequestHandler nextHandler = (pagingState == null) ? null : this.copy(pagingState);
       setFinalResult(new AdminResult(rows, nextHandler, channel.protocolVersion()));
-    } else if (message instanceof Prepared) {
+    } else if (message instanceof Prepared
```
adutra: I think we can broaden the scope of this check to `message instanceof Result` – `Rows` being just a special case of `Result` that was already handled in the previous `if` block.

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRequestHandler.java
tomekl007: Good idea - taking into account the fact `Void` and `Prepared` is extending `Result`.
But the question is: do we want to treat `SchemaChange` and `SetKeyspace` as successes here as well? (both are extending `Result`). I think that we can, wdyt?

adutra: The case has never presented itself until now (only `Rows` and `Prepared` were handled so far), so I assume that we can safely consider these as successes as well.

tomekl007: ok, done

adutra (approved):

PR #1186, opened 2019-02-07
JAVA-2144: Expose internal API to hook into the session lifecycle

olim7t:


core/src/main/java/com/datastax/oss/driver/internal/core/context/LifecycleListener.java
```diff
@@ -0,0 +1,51 @@
...
+   *
+   * <p>Note that this method gets called even if the shutdown results from a failed initialization.
+   * In that case, implementations should be ready to handle a call to this method even though
+   * {@link #onSessionReady()} hasn't been invoked.
+   *
+   * <p>This method is invoked on a driver thread, it should complete relatively quickly and not
+   * block.
+   */
+  @Override
+  void close() throws Exception;
```
adutra, line 53: Is this meant to be where implementors shut down their own listeners, or as a notification hook that gets invoked when the session closes? For clarity I think we should dissociate both and add a method `onSessionClosed` to capture the second use case. Also, we might want to pass the `Throwable` that caused the session initialization to fail.

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
```diff
@@ -393,6 +394,7 @@ private void afterInitialSchemaRefresh(CqlIdentifier keyspace) {
                   if (error != null) {
                     initFuture.completeExceptionally(error);
                   } else {
+                    notifyLifecycleListeners();
```
adutra: Shouldn't the call come after `initFuture.complete(DefaultSession.this);`?

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/context/LifecycleListener.java
```diff
@@ -0,0 +1,51 @@
...
+   * Invoked when the session is ready to process user requests.
+   *
+   * <p>This corresponds to the moment when the {@link SessionBuilder#build()} returns, or the
+   * future returned by {@link SessionBuilder#buildAsync()} completes. If the session initialization
+   * fails, this method will not get called.
+   *
+   * <p>This method is invoked on a driver thread, it should complete relatively quickly and not
+   * block.
+   */
+  void onSessionReady();
```
tomekl007, line 37: We will need to get the unique session identifier as a parameter here (and for `close()` as well)
If the listener will get the notification for every new session it may want to create a resource for that specific session. Then to be able to identify that resource in the `close()` callback we would need to have session id.

tomekl007 (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/context/LifecycleListener.java
olim7t: We've discussed this elsewhere already, but just to document the answer: a given context is associated to exactly one session, so that won't be an issue.

core/src/main/java/com/datastax/oss/driver/internal/core/context/LifecycleListener.java
olim7t: It gets called when the session closes.
Implementors should perform any necessary cleanup. A typical example would be to cancel a scheduled task that they had scheduled on a driver admin event loop.

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
olim7t: It's really up to us to decide and I don't think it makes such a difference either way.

But given how I worded it in the javadocs, you're right, I should probably do it after.

tomekl007 (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/context/LifecycleListener.java
olim7t: I don't understand how `onSessionClosed` would be a different use case, both methods would be called when the session closes.
As for passing the `Throwable` that caused an init failure, I don't see a strong case for it. This method mostly exists so that we don't leave anything running on the admin event loops, I think it's unlikely that a listener would do something differently depending on the error.
And if we find the need to add it later, it will be relatively easy to have a default implementation of `close(Throwable)` that calls `close()`.

adutra: It's a matter of semantics: `onSessionClosed` means: "I am being notified that a given session has been closed" whereas the contract of `close` (as defined in `AutoCloseable`) is "I am being told to close myself". Here, it sounds to me as if we are distorting the original purpose of `close`.

More generally speaking, a typical listener API would expose methods such as `onEventXSuccessful(Foo receiver)`, `onEventYFailed(Bar receiver, Throwable error)`. It would also expose methods to register/unregister the listener against a given Session. None of this is present in this interface, which is why I was a bit thrown off initially.

But OK to keep things simple for now; it's an internal API after all so we are free to change it later.

adutra (approved):

PR #1187, opened 2019-02-08
Improve manual on prepared statements

olim7t:
I was pointing someone to the "prepared statements" page of the manual the other day, and I noticed that some explanations lacked clarity.

### DIFF
--- a/manual/core/statements/prepared/README.md
+++ b/manual/core/statements/prepared/README.md
@@ -55,34 +55,68 @@ client                            driver                Cassandra
   |<--------------------------------|                     |
 ```
 
+Beyond saving a bit of parsing overhead on the server, prepared statements have other advantages;
+the `PREPARED` response also contains useful metadata about the CQL query:
+
+* information about the result set that will be produced when the statement gets executed. The
+  driver caches this, so that the server doesn't need to include it with every response. This saves
+  a bit of bandwidth, and the resources it would take to decode it every time.
+* the CQL types of the bound variables. This allows bound statements' `set` methods to perform
+  better checks, and fail fast (without a server round-trip) if the types are wrong.
+* which bound variables are part of the partition key. This allows bound statements to automatically
+  compute their [routing key](../../load_balancing/#token-aware).
+* more optimizations might get added in the future. For example, [CASSANDRA-10813] suggests adding
+  an "[idempotent](../../idempotence)" flag to the response.
+
+If you have a unique query that is executed only once, a [simple statement](../simple/) will be more
+efficient. But note that this should be pretty rare: most client applications typically repeat the
+same queries over and over, and a parameterized version can be extracted and prepared.  
+
+### Preparing
+
+`Session.prepare()` accepts either a plain query string, or a `SimpleStatement` object. If you use a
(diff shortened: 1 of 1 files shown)


manual/core/statements/prepared/README.md
```diff
@@ -55,35 +55,69 @@ client                            driver                Cassandra
...
+* the CQL types of the bound variables. This allows bound statements' `set` methods to perform
+  better checks, and fail fast (without a server round-trip) if the types are wrong.
+* which bound variables are part of the partition key. This allows bound statements to automatically
+  compute their [routing key](../../load_balancing/#token-aware).
+* more optimizations might get added in the future. For example, [CASSANDRA-10813] suggests adding
+  an "[idempotent](../../idempotence)" flag to the response.
+
+If you have a unique query that is executed only once, a [simple statement](../simple/) will be more
+efficient. But note that this should be pretty rare: most client applications typically repeat the
+same queries over and over, and a parameterized version can be extracted and prepared.
```
tomekl007, line 73: maybe add "for example in the typical REST application your `BoundStatement`  can have a bound variable `user-id`" or something similar?

manual/core/statements/prepared/README.md
```diff
@@ -55,35 +55,69 @@ client                            driver                Cassandra
...
   sanitizing.
 * all other execution parameters: for example, preparing two statements with identical query strings
-  but different consistency levels will yield distinct prepared statements.
+  but different consistency levels will yield two distinct prepared statements (that each produce
+  bound statements with their respective consistency level).
 
 The size of the cache is exposed as a session-level [metric](../../metrics/)
 `cql-prepared-cache-size`. The cache uses [weak values]([guava eviction]) eviction, so this
 represents the number of `PreparedStatement` instances that your application has created, and is
 still holding a reference to.
```
tomekl007, line 119: Maybe we can add info that by default the size of the cache in unbounded?

adutra (approved):

manual/core/statements/prepared/README.md
olim7t: There are examples of parameters and binding further down the page, so I don't think we need to go into too much detail here.

manual/core/statements/prepared/README.md
olim7t: The thing is, if we mention that, it plants the idea that the cache could be bounded. And I'm not sure how that would work:

Let's say the cache has a capacity of 10 with an LRU eviction strategy. You prepare statement number 11, which means another statement gets evicted, let's stay number 1. But all that means is that the next `session.prepare` call for statement 1's query won't hit the cache. The application might still hold a reference to `preparedStatement1`, and executing it might still work if it's not been evicted server-side. The only thing we're achieving is that if the query is prepared again, it will hit the server (unnecessarily) and the driver will build a duplicate `PreparedStatement` instance. So we might be wasting heap space instead of saving it.

Unbounded with weak values works well because it adapts to what the application is preparing and holding onto. Of course if you have a rogue application that prepares in a loop and keeps a reference to every statement, it won't work well, but neither will a bounded cache from the explanations above.

PR #1189, opened 2019-02-13
JAVA-2123: Allow to use QueryBuilder for building queries against Materialized Views

adutra:


driver-core/src/main/java/com/datastax/driver/core/querybuilder/Select.java
```diff
@@ -857,5 +870,14 @@ public Select from(TableMetadata table) {
       previousSelection = null;
       return super.from(table);
     }
+
+    @Override
+    public Select from(MaterializedViewMetadata view) {
+      if (previousSelection != null) {
+        addName(previousSelection);
+      }
+      previousSelection = null;
```
tomekl007, line 879: Why we are setting it back explicitly to `null`? To prevent memory leak?

driver-core/src/test/java/com/datastax/driver/core/querybuilder/QueryBuilderExecutionTest.java
```diff
@@ -924,4 +927,45 @@ public void should_support_group_by() throws Exception {
...
+              @Override
+              public Boolean call() {
+                return session().execute("SELECT * FROM " + mv).all().size() == 3;
+              }
+            })
+        .before(1, MINUTES)
+        .becomesTrue();
+
+    assertThat(session().execute(select().column("cc").as("mycc").from(mv)))
+        .containsExactly(row(0), row(1), row(2));
```
tomekl007, line 973: this `row(n)` syntax is validating only the fact that row exists? or also the content of a row?

driver-core/src/test/java/com/datastax/driver/core/querybuilder/QueryBuilderExecutionTest.java
```diff
@@ -924,4 +927,45 @@ public void should_support_group_by() throws Exception {
...
+            new Callable<Boolean>() {
+              @Override
+              public Boolean call() {
+                return session().execute("SELECT * FROM " + mv).all().size() == 3;
+              }
+            })
+        .before(1, MINUTES)
+        .becomesTrue();
+
+    assertThat(session().execute(select().column("cc").as("mycc").from(mv)))
```
tomekl007: I see that we added a method `from(MaterializedViewMetadata view)` that takes `MaterializedViewMetadata` type as an argument. Here we are passing `mv` that is of a `String` type. Where the conversion from `String` to `MaterializedViewMetadata` is happening?

driver-core/src/main/java/com/datastax/driver/core/querybuilder/Select.java
olim7t: That's a trick to handle aliased selections:
```java
select().column("column1").as("alias1")
```
Internally all selectors are added to the `columnNames` list. But if a column is aliased, it must be wrapped into an `Alias` before (see implementation of `as()`).
`column()` doesn't know yet if its argument will be aliased or not, because that only comes in the `as()` called that is chained after. So instead it stores its argument temporarily in `previousSelection`.
If the next method is `as()`, it will wrap `previousSelection` and add it.
If it's any other method that can be chained after `column`, it adds it directly. `from()` is one of these methods.

`previousSelection` must be reset so that another method doesn't add it again later.

driver-core/src/test/java/com/datastax/driver/core/querybuilder/QueryBuilderExecutionTest.java
olim7t: There's no conversion, the implementation can work either with a plain string or a metadata object.

That's probably an oversight. If we want to cover the functionality that's been added, we should indeed extract a `MaterializedViewMetadata` from the cluster metadata and pass that to the query builder.

olim7t (changes requested):
Needs a small change in the test per Tomasz's comment. Apart from that, LGTM.

driver-core/src/test/java/com/datastax/driver/core/querybuilder/QueryBuilderExecutionTest.java
adutra: Good catch thanks.

driver-core/src/test/java/com/datastax/driver/core/querybuilder/QueryBuilderExecutionTest.java
adutra: It also asserts the contents, see `ResultSetAssert`.

tomekl007 (approved):

olim7t (approved):

PR #1190, opened 2019-02-19
Various manual fixes

olim7t:
This is a batch of trivial documentation tickets for 4.0.0.

I don't think we need changelog entries.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -237,25 +237,102 @@ factor some common code in a parent abstract class named with "XxxTestBase", and
 different families of tests into separate child classes. For example, `CqlRequestHandlerTestBase`,
 `CqlRequestHandlerRetryTest`, `CqlRequestHandlerSpeculativeExecutionTest`...
 
-Unit tests live in their respective production code's module. They should be fast and not start any
+### Unit tests
+
+They live in the same module as the code they are testing. They should be fast and not start any
 external process. They usually target one specific component and mock the rest of the driver
 context.
 
-Integration tests live in the aptly-named `integration-tests` module. They exercise the whole driver
-stack against an external process -- either simulated with
-[Simulacron](https://github.com/datastax/simulacron), or a live Cassandra cluster run with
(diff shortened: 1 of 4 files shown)


faq/README.md
```diff
@@ -8,12 +8,23 @@ sure you don't accidentally ignore their result:
...
+boundSelect = boundSelect.setInt("k", key).setPageSize(1000);
+```
+
+The driver also provides builders:
+
+```java
+BoundStatement boundSelect =
+    preparedSelect.boundStatementBuilder()
+        .setInt("k", key)
+        .withPageSize(1000)
```
olim7t, line 26: On a side note, this bugs me: the value setters have to be called `set` because they are inherited, but the other builder methods start with `with`.
It's very late for this, but maybe we should rename builder methods to `setPageSize`.

CONTRIBUTING.md
```diff
@@ -382,7 +438,6 @@ where the changes should be squashed:
 (Note that the message refers to the other commit's subject line, not the SHA-1. This way it's still
 relevant if there are intermediary rebases.)
 
-If you *really* need a newer commit from the base branch, or if the history is getting too
-complicated, it can be OK to force-push occasionally, provided that **all current reviewers agree**.
-
-Don't push a merge commit to a pull request under any circumstance.
+If you need new stuff from the base branch, it's fine to rebase and force-push, as long as you don't
+rewrite the history. Just give a heads up to the reviewers beforehand. Don't push a merge commit to
```
adutra, line 442: If you use Github web UI to update the branch or resolve conflicts, it will create a merge commit. I personally don't think we absolutely must avoid merge commits in a PR.

adutra (approved):

CONTRIBUTING.md
olim7t: I find them really annoying, it's an extra set of changes that have nothing to do with the PR. Also they tend to get in the way when you do interactive rebases locally.

If we allow forced pushes, it's so much cleaner to rebase.

PR #1192, opened 2019-02-20
JAVA-2112 Support "EntityGet" methods

GregBestland:


integration-tests/src/test/java/com/datastax/oss/driver/mapper/GetEntityIt.java
```diff
@@ -0,0 +1,18 @@
+/*
```
GregBestland: IT is not where I want it yet. It will be flushed out when I return

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+  public DaoGetEntityMethodGenerator(
+      ExecutableElement methodElement,
+      DaoImplementationGenerator daoImplementationGenerator,
+      ProcessorContext context) {
+    this.methodElement = methodElement;
+    this.daoImplementationGenerator = daoImplementationGenerator;
+    // Methods should only have one parameter
+    if (methodElement.getParameters().size() != 1) {
+      context
+          .getMessager()
```
tomekl007, line 73: typo here: `getMessanger()`

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+  private boolean pagingReturnType;
+
+  public DaoGetEntityMethodGenerator(
+      ExecutableElement methodElement,
+      DaoImplementationGenerator daoImplementationGenerator,
+      ProcessorContext context) {
+    this.methodElement = methodElement;
+    this.daoImplementationGenerator = daoImplementationGenerator;
+    // Methods should only have one parameter
+    if (methodElement.getParameters().size() != 1) {
```
tomekl007, line 71: If I understand correctly, all code in a constructor is a validation, right?
Isn't there a risk that we will finish with the partially constructed object if the constructor invocation fails in let say half-of the-process. Maybe MethodGenerator should have `pre-validate()` method that is called from an external component before calling the constructor of the implementation?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+      tmpParameterString = parameterElement.getSimpleName().toString();
+    }
+    // Check return type. Make sure it matches the expected type
+    TypeMirror returnType = methodElement.getReturnType();
+    if (returnType.getKind() == TypeKind.DECLARED) {
+      Element element = ((DeclaredType) returnType).asElement();
+      // Simple case return type is an entity type
+      if (element.getKind() == ElementKind.CLASS) {
+        if (element.getAnnotation(Entity.class) != null) {
+          returnTypeElement = ((TypeElement) ((DeclaredType) returnType).asElement());
```
tomekl007: don't we need checks for instance of before that?
Is this `(returnType.getKind() == TypeKind.DECLARED) ` imply that we can cast to DeclaredType, right?
What about cast to TypeElement? This check takes care of it`(if (element.getKind() == ElementKind.CLASS))`?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+        }
+        // Return type is a an iterable type, check nested generic as well
+      } else if (element.getKind() == ElementKind.INTERFACE) {
```
tomekl007: can't we be more specific in that check?
Is there a way to check at this level that `element.getKind()` is also implementing `AsyncPagingIterable` or `PagingIterable`?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+        List<? extends TypeMirror> generics = ((DeclaredType) returnType).getTypeArguments();
+        if (generics.size() == 1) {
+          TypeMirror generic = generics.get(0);
+          Element genericElement = ((DeclaredType) generic).asElement();
+          if (genericElement.getAnnotation(Entity.class) != null) {
+            if (element.getSimpleName().toString().equals(PagingIterable.class.getSimpleName())
```
tomekl007: isn't it better to check equals on classes instead of strings?
Like `element.getClass.equals(PagingIterable.class)`?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+          }
+        }
+      }
+    }
+    if (returnTypeElement == null) {
+      context
+          .getMessager()
+          .error(
+              methodElement,
+              "Invalid return type specified. Expected annotated entity, PagingIterable, or AsyncPagingIterable ");
```
tomekl007: maybe in a log error, we should specify fully qualified class names?
Instead of `com.datastax.driver.core.PagingIterable` to give more explicit info about expected type.
Also, we could assign it to constants

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+      throw new SkipGenerationException();
+    }
+    // Parameter name as a string
+    String tmpParameterString = null;
+    // The return type as an element
+    TypeElement returnTypeElement = null;
+    VariableElement parameterElement = methodElement.getParameters().get(0);
+    TypeMirror parameterType = parameterElement.asType();
+    // Check the parameter type make sure it matches an expected type
+    if (context.getClassUtils().implementsGettableByName(parameterType)
```
tomekl007: I think that maybe we can create 3 different paths of execution instead of handling everything in one.
So we could have 3 ifs:
```
if(context.getClassUtils().implementsGettableByName(parameterType)){
   handleGettableByName()
}
else if(context.getClassUtils().isSame(parameterType, ResultSet.class)){
handleResultSet()
}
else if(context.getClassUtils().isSame(parameterType, AsyncResultSet.class)){
handleAsyncResultSet()
}
```
Then in every of the `handle*` method we can focus on validaing only one possbile path instead of doing everything in multiple ifs that needs to take care of every of those 3 combinations, wdyt?

mapper-processor/pom.xml
```diff
@@ -94,6 +94,13 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.revapi</groupId>
+        <artifactId>revapi-maven-plugin</artifactId>
+        <configuration>
+          <skip>true</skip>
```
tomekl007: I have problems with revapi configuration that doesn't allow me to compile the project (I've added two skip rules for `mapper-runtime` and `mapper-processor`)

integration-tests/src/test/java/com/datastax/oss/driver/mapper/GetEntityIt.java
tomekl007: There is a problem with the compilation of this branch. 
The `SetEntityIT` fails because there is no class `com.datastax.oss.driver.mapper.model.inventory.InventoryMapperBuilder;`
https://travis-ci.com/datastax/java-driver/jobs/180213823

I cannot compile project in intelliJ because I got got:
`Error:java: Bad service configuration file, or exception thrown while constructing Processor object: javax.annotation.processing.Processor: Provider com.datastax.oss.driver.internal.mapper.processor.MapperProcessor could not be instantiated: java.lang.NoClassDefFoundError: com/datastax/oss/driver/internal/mapper/processor/ProcessorContext`

Other question: how to start the annotation processor from `mvn`?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
```diff
@@ -18,37 +18,98 @@
...
+    String returnName = "returnValue";
+    getBuilder.addStatement("$T $L = new $T()", returnType, returnName, returnType);
+
+    int udtIndex = 0;
+    for (PropertyDefinition property : entityDefinition.getProperties()) {
+      TypeName type = property.getType();
+      String cqlName = property.getCqlName();
+      String setterName = property.getSetterName();
+      TypeElement childEntityElement = property.getEntityElement();
+      getBuilder.addCode("\n");
```
tomekl007, line 77: It looks a bit suspicious. 
Isn't there a method in the getBuilder that encapsulate that new line invocation (something like `beginControlFlow` similar to `endControlFlow`)? If not maybe we want to use `System.lineSeparator()` to make that system agnostic

tomekl007 (changes requested):

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
GregBestland: This naming is inherited from javax.annotation.processing's Messager interface. The naming is wonky, but it's consistent.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
GregBestland: This was my initial approach however what you will find is the validation that needs to be done is for the most part divorced from the parameter type, and that the code we create in each handel method would be just as convulted and complex as what we have now.

mapper-processor/pom.xml
olim7t: The plugin fails because I forgot to add `revapi.json` files.

> I've added two skip rules

👍  for `mapper-processor` (it's purely internal). For `mapper-runtime` there are public types, such as the annotations, so we need to keep the plugin enabled and add a configuration file. I'll do this on `java2078`.

olim7t:
I've pushed 2 commits to `java2078` that should take care of the Revapi and test class generation issues. @GregBestland is it ok if I rebase and force-push this branch?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
olim7t: `element.getClass()` will give you the actual class of the model object (a subtype of `javax.lang.model.element.Element`), not the class that it represents.

However another problem here is that this would match any `PagingIterable` class, regardless of the package.

There is a method `context.getClassUtils().isSame(Element, Class<?>)` for this kind of check.

olim7t:
Rebased on `java2078` and force-pushed, as explained in my previous comment.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
olim7t: Yes, the constructor code is all validation. If validation fails, the constructor throws `SkipGenerationException` and the parent class will discard this instance.

We could add a `validate()` method, make computed fields like `entityElement` non-final, and move the code around, but that wouldn't make a big difference.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
olim7t: If a type mirror is of kind `DECLARED`, you can cast it to `DeclaredType` and get to the corresponding element (the body of the type that is declared).
If an element is of kind `CLASS`, you can cast it to `TypeElement`.

What's a bit confusing here is that there is some duplication, line 81 can be simplified as `returnTypeElement = (TypeElement) element;`

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
olim7t: That's done below, but I agree that it would be a bit more clear to do all the checks on the main type first, and only then inspect the type parameters. I'll change it.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
olim7t: I don't see a specific method in JavaPoet (`CodeBlock` is the class that all those methods delegate to).
Note that JavaPoet uses `\n` for all line breaks internally.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
olim7t: It's nice to be precise, but on the other hand we want to keep messages reasonably short and it should be pretty obvious that those are driver classes. I'll try with the full path to see what the compiler messages look like.

As for constants, one already exists, the class itself:
```java
error(methodElement, "blah blah %s", PagingIterable.class.getSimpleName())
```
The nice thing is that it survives refactorings.

I did that for annotations already, but not for other driver classes, I'll revisit my code.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
```diff
@@ -18,37 +18,98 @@
...
 
+  private static final Map<TypeName, String> PRIMITIVE_ACCESSORS =
+      ImmutableMap.<TypeName, String>builder()
+          .put(TypeName.BOOLEAN, "Boolean")
+          .put(TypeName.BYTE, "Byte")
+          .put(TypeName.DOUBLE, "Double")
+          .put(TypeName.FLOAT, "Float")
+          .put(TypeName.INT, "Int")
+          .put(TypeName.LONG, "Long")
+          .build();
```
olim7t, line 44: This is duplicated from `EntityHelperSetMethodGenerator` but I'm leaving it here for now, it will all get extracted to a common class in JAVA-2115.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
```diff
@@ -18,37 +18,98 @@
...
             .addParameter(
                 ParameterSpec.builder(ClassName.get(GettableByName.class), "source").build())
-            .returns(entityDefinition.getClassName())
-            // TODO generate proper implementation (JAVA-2112)
-            .addStatement("throw new $T($S)", UnsupportedOperationException.class, "TODO");
+            .returns(entityDefinition.getClassName());
+
+    TypeName returnType = entityDefinition.getClassName();
+    String returnName = "returnValue";
+    getBuilder.addStatement("$T $L = new $T()", returnType, returnName, returnType);
```
olim7t: Nit: you can number placeholders to avoid repeating the values:
```suggestion
    getBuilder.addStatement("$1T $2L = new $1T()", returnType, returnName);
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
```diff
@@ -18,37 +18,98 @@
...
+      String setterName = property.getSetterName();
+      TypeElement childEntityElement = property.getEntityElement();
+      getBuilder.addCode("\n");
+      if (childEntityElement != null) {
+        udtIndex += 1;
+        // Populate udtInformation
+        String udtValueName = "udtValue" + udtIndex;
+        String valueName = "value" + udtIndex;
+        // Extract UdtValue to pass it on to underlying helper method
+        getBuilder.addStatement("UdtValue $L = source.getUdtValue($S)", udtValueName, cqlName);
```
olim7t: Types should be injected with `$T` even if they are known statically. Otherwise the generated code might not have the import (in this case I think it works because the set method already imports it).
```suggestion
        getBuilder.addStatement("$T $L = source.getUdtValue($S)", UdtValue.class, udtValueName, cqlName);
```
Also this more refactoring-friendly (although `UdtValue` is unlikely to get renamed at this stage).

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+        || context.getClassUtils().isSame(parameterType, ResultSet.class)
+        || context.getClassUtils().isSame(parameterType, AsyncResultSet.class)) {
+      tmpParameterString = parameterElement.getSimpleName().toString();
+    }
```
olim7t, line 96: That's the only place where we set this variable, so we can handle the error case directly here, and we don't need a temporary variable.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+          ClassName.get(parameterElement.asType()), parameterElement.getSimpleName().toString());
+    }
+    if (!pagingReturnType) {
+
+      overridingMethodBuilder.addStatement(
+          "return $L.get($L)", helperFieldName, targetParameterName);
+    } else {
+      overridingMethodBuilder.addStatement(
+          "return $L.map($L::get)", targetParameterName, helperFieldName);
+    }
```
olim7t, line 194: I think it would be nice to handle one additional case: converting a single-row result set to an entity instance.
```java
Product get(ResultSet|AsyncResultSet rs);
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+    TypeElement returnTypeElement = null;
+    VariableElement parameterElement = methodElement.getParameters().get(0);
+    TypeMirror parameterType = parameterElement.asType();
+    // Check the parameter type make sure it matches an expected type
+    if (context.getClassUtils().implementsGettableByName(parameterType)
+        || context.getClassUtils().isSame(parameterType, ResultSet.class)
+        || context.getClassUtils().isSame(parameterType, AsyncResultSet.class)) {
+      tmpParameterString = parameterElement.getSimpleName().toString();
+    }
+    // Check return type. Make sure it matches the expected type
```
olim7t: Additionally, we should check that the return type matches the argument, because not all combinations are valid:
```java
GettableByName => Entity
ResultSet => PagingIterable or Entity
AsyncResultSet => AsyncPagingIterable or Entity
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
olim7t: Nit: this error can be targeted more precisely to `parameterElement`.

olim7t (changes requested):
Greg is busy on another ticket so we agreed I would help finish this PR.
Posting my feedback anyway to explain my changes.

tomekl007 (approved):

olim7t (approved):

PR #1193, opened 2019-02-21
JAVA-2158: Allow BuildableQuery to build statement with values

olim7t:
TODO
- [x] add variant for named values
- [x] reword commit message (to match changelog and PR title)

### DIFF
--- a/query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/BuildableQuery.java
+++ b/query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/BuildableQuery.java
@@ -20,6 +20,7 @@ import com.datastax.oss.driver.api.core.cql.SimpleStatement;
 import com.datastax.oss.driver.api.core.cql.SimpleStatementBuilder;
 import com.datastax.oss.driver.api.core.cql.Statement;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Map;
 
 /**
  * End state for the query builder DSL, which allows the generation of a CQL query.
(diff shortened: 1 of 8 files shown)


query-builder/src/test/java/com/datastax/oss/driver/api/querybuilder/BuildableQueryTest.java
```diff
@@ -0,0 +1,130 @@
...
+        deleteFrom("foo").whereColumn("k").isEqualTo(bindMarker()),
+        new Object[] {1},
+        "DELETE FROM foo WHERE k=?",
+        true
+      },
+      {
+        deleteFrom("foo").whereColumn("k").isEqualTo(bindMarker()).ifExists(),
+        new Object[] {1},
+        "DELETE FROM foo WHERE k=? IF EXISTS",
+        false
```
tomekl007, line 60: why this query is not idempotent? I think that this DELETE should be

query-builder/src/test/java/com/datastax/oss/driver/api/querybuilder/BuildableQueryTest.java
```diff
@@ -0,0 +1,130 @@
...
+      },
+      {
+        update("foo").setColumn("v", bindMarker()).whereColumn("k").isEqualTo(bindMarker()),
+        new Object[] {3, 1},
+        "UPDATE foo SET v=? WHERE k=?",
+        true
+      },
+      {
+        update("foo")
+            .setColumn("v", function("non_idempotent_func"))
```
tomekl007, line 82: is there a way to create a function that is `idempotent`?

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/BuildableQuery.java
```diff
@@ -47,14 +47,32 @@
...
+   *
+   * <pre>{@code
+   * SimpleStatement.newInstance(asCql(), values)
+   * }</pre>
+   *
+   * In addition, some implementations might try to infer additional statement properties (such as
+   * {@link Statement#isIdempotent()}).
+   */
+  @NonNull
+  default SimpleStatement build(Object... values) {
```
adutra: While at it, why not also add variants for named values?

query-builder/src/test/java/com/datastax/oss/driver/api/querybuilder/BuildableQueryTest.java
olim7t: Conditional queries are always considered non idempotent. This is counter-intuitive, but there are subtle interactions with retries. See explanation [here](http://coppi.datastax.lan/en/developer/java-driver/4.0-rc/manual/query_builder/idempotence/#conditional-statements).

query-builder/src/test/java/com/datastax/oss/driver/api/querybuilder/BuildableQueryTest.java
olim7t: No. The query builder is pessimistic by default and will always consider the query as idempotent as soon as there is a function call.
If you know that all your functions are safe, the workaround is to override the flag on the statement.

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/BuildableQuery.java
olim7t: Indeed. `SimpleStatement.newInstance` has that variant, so it makes sense to add it here too. Will do.

PR #1194, opened 2019-02-21
JAVA-2150: Improve query builder error message on unsupported literal type

olim7t:


query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/QueryBuilder.java
```diff
@@ -400,7 +400,18 @@ public static Literal literal(@Nullable Object value) {
    */
   @NonNull
   public static Literal literal(@Nullable Object value, @NonNull CodecRegistry codecRegistry) {
-    return literal(value, (value == null) ? null : codecRegistry.codecFor(value));
+    try {
+      return literal(value, (value == null) ? null : codecRegistry.codecFor(value));
```
tomekl007, line 404: is `literal` throwing CodecNotFoundException if the codec passed to it as an `argument == null`?
Maybe we can handle that at this higher level without the need to throw an exception?

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/QueryBuilder.java
```diff
@@ -400,7 +400,18 @@ public static Literal literal(@Nullable Object value) {
...
+    } catch (CodecNotFoundException e) {
+      assert value != null;
+      throw new IllegalArgumentException(
+          String.format(
+              "Could not inline literal of type %s. "
+                  + "This happens because the driver doesn't know how to map it to a CQL type. "
+                  + "Try passing a TypeCodec or CodecRegistry to literal().",
```
adutra, line 411: Maybe add: 

```
Try passing a TypeCodec or CodecRegistry to literal(), or register a TypeCodec that accepts "+ value.getClass().getName() + "."
```

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/QueryBuilder.java
olim7t: It throws if `codecFor` doesn't find a codec to handle `value`.

If `value` is null, `codecFor` is not called so CNFE is never thrown.

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/QueryBuilder.java
olim7t: This query builder method is completely detached from any session though (and the default registry is immutable), so even if you register your codec somewhere, you'll still have to pass the `CodecRegistry`.

adutra (approved):

PR #1197, opened 2019-02-26
JAVA-2131: Figure out our testing strategy for Mapper 4.1

tomekl007:


integration-tests/src/test/java/com/datastax/oss/driver/mapper/DaoAnnotationTest.java
```diff
@@ -0,0 +1,58 @@
...
+                JavaFileObjects.forSourceLines(
+                    PACKAGE_NAME + ".ProductDao", loadLinesFromSourceFile(sourceFileLocation)));
+
+    // then
+    assertThat(compilation).succeededWithoutWarnings();
+    assertThat(compilation)
+        .generatedFile(StandardLocation.SOURCE_OUTPUT, PACKAGE_NAME, "ProductDao_Impl.java")
+        .hasSourceEquivalentTo(
+            JavaFileObjects.forSourceLines(
+                PACKAGE_NAME + "ProductDao_Impl_Expected",
```
GregBestland: My only concern here is that this will create some maintenance overhead, as we add/remove or change our code generation.... I can't however come up with a better way to do this, so for now I think it will work

GregBestland (approved):

integration-tests/src/test/java/com/datastax/oss/driver/mapper/DaoAnnotationTest.java
tomekl007: I agree that it has some maintenance cost. 
But also, testing normal code (not annotation processor generation) in a unit test way and writing test cases for new functionalities has that cost. 
The generated code should be tested in some way and every change to generated class is something that should be reflected in the test. So I think that this maintenance cost is worth it.

emerkle826 (approved):
This looks good. I agree about the maintenance cost, but it should be minimal once the mapper work is settled out.

olim7t:
Please don't merge this yet.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/DaoAnnotationTest.java
olim7t: The problem is doing character-to-character comparison of the generated code. This is very sensitive to changes; if we rely on this too much, my fear is that we'll soon have to update a gazillion reference files any time we touch something. You've already run into that (last commit "update generated files"), and I'm seeing it again as I've pushed new changes to `java2078` and am rebasing this PR.

I think we should reserve this approach for:
* processor errors and warnings
* checking that files have been generated, without necessarily testing their contents (a great use case for that would be JAVA-2136: Support processing of nested types)

Also, the test code should be simple enough that it can be inlined directly in the test. That's more readable than having to open separate resource files. If something is too complicated to be inlined, it probably warrants a full-fledged integration test.

I'll push a few examples in that spirit once I'm done rebasing.

olim7t:
Force-pushed to rebase on latest `java2078`.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/DaoAnnotationTest.java
```diff
@@ -0,0 +1,63 @@
...
+        compileWithMapperProcessor(
+            "test",
+            TypeSpec.classBuilder(ClassName.get("test", "Foo"))
+                .addModifiers(Modifier.PUBLIC)
+                .addType(
+                    TypeSpec.interfaceBuilder("Bar")
+                        .addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+                        .addAnnotation(Dao.class)
+                        .build())
+                .build());
```
olim7t, line 42: JavaPoet works pretty well to embed source code directly in the test. It is a bit more refactoring-friendly than raw strings (for example if we rename annotations).

I like that the test is self-contained (no external source files): it's easier to read, also it lowers the bar to add more tests (no need to think of names for the new files, etc).

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/DaoAnnotationTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+    assertThat(compilation).succeededWithoutWarnings();
+    assertThat(compilation)
+        .generatedFile(StandardLocation.SOURCE_OUTPUT, "test", "Foo_BarImpl__MapperGenerated.java")
+        .contentsAsUtf8String()
+        .contains("class Foo_BarImpl__MapperGenerated extends DaoBase implements Foo.Bar");
```
olim7t, line 48: If we check the generated source at all, it doesn't have to be the whole file, we can focus on the most important parts.
In this test it's not even required, we could rely on the file name only.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/DaoAnnotationTest.java
```diff
@@ -0,0 +1,63 @@
...
+  public void should_fail_on_class() {
+    Compilation compilation =
+        compileWithMapperProcessor(
+            "test",
+            TypeSpec.classBuilder(ClassName.get("test", "Foo"))
+                .addModifiers(Modifier.PUBLIC)
+                .addAnnotation(Dao.class)
+                .build());
+
+    assertThat(compilation).hadErrorContaining("Only INTERFACE elements can be annotated with Dao");
```
olim7t, line 61: Since the source snippets are very small and focused on a single test, I think we don't need to check the line and column of the error.

olim7t (commented):
As promised, a few examples of tests with inlined source codes.

Btw these are unit tests, so they should leave in the mapper-processor module, not integration-tests.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/DaoAnnotationTest.java
tomekl007: > You've already run into that (last commit "update generated files")
My last commit was because on the `java2078` branch there was a new feature and I was rebasing this branch so I've added tests cases to cover those new functionalities. Maybe if we will add them when implementing this maybe it won't be painful?

>processor errors and warnings 

+1 on that. It is a good tool to validate if we are failing with the proper message

>checking that files have been generated, without necessarily testing their contents (a great use case for that would be JAVA-2136: Support processing of nested types)

It sounds good but I am a bit skeptical about that. Validating only the fact that file was generated properly sometimes may not be enough. But I am also +1 on doing this inline in code and maybe validate that specific line of code appeared in the output instead of validating whole content.

On the other hand, I agree with @emerkle826 comment:
>I agree about the maintenance cost, but it should be minimal once the mapper work is settled out.
Right now maybe it is a higher maintenance cost because we are developing mapper. Once it will be stable, tests that have full coverage of generated code will give us a lot more confidence that everything works as expected. Also, when we will be evolving fairly stable code base it will be a lot easier to validate that nothing was broken when having expected generated file compared to the actual one. Wdyt?

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/DaoAnnotationTest.java
tomekl007: I agree that embedding source code in a test makes it better for maintenance. 
But on the other hand, if our users that have no previous experience with annotation processors will want to understand how to use mapper they may want to go to tests to see how it work.
If they don't know javaPoet that test may be really hard to understand for them. If they will have raw `Foo.java` file used by the test that will be easier for them to reason about the behaviour

olim7t: I wouldn't consider those unit tests as examples. The integration tests are a better fit for that IMO (and we'll definitely also cover the mapper in examples / manual).

PR #1198, opened 2019-02-27
Fix flaky tests for JAVA-1950

emerkle826:


emerkle826:
Adjusting when the MemoryApender is enabled/disabled seems to have made the log verification more stable.

PR #1199, opened 2019-02-27
guard call to jnr Platform.getCPU via Native wrapper

tomekl007:


core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -75,6 +77,32 @@ public static int getProcessId() {
...
+  /**
+   * Returns the current processor architecture the JVM is running on, as reported by {@link
+   * Platform#getCPU()}.
+   *
+   * @return the current processor architecture.
+   * @throws UnsupportedOperationException if JNR Platform library is not loaded.
+   */
+  public static String getCPU() {
+    if (!isPlatformAvailable())
+      throw new UnsupportedOperationException("JNR Platform class not loaded");
```
adutra: I see that similar methods in this class use `IllegalStateException`, shouldn't we do the same here?

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -150,4 +178,27 @@ private Timeval(Runtime runtime) {
...
+
+    static {
+      Platform platform;
+      try {
+        Class<?> platformClass = Class.forName("jnr.ffi.Platform");
+        Method getNativePlatform = platformClass.getMethod("getNativePlatform");
+        platform = (Platform) getNativePlatform.invoke(null);
+      } catch (Throwable t) {
+        platform = null;
+        if (LOG.isDebugEnabled())
```
adutra: I see that contrary to 3.x, in 4.x we only log at DEBUG level, again I think we should harmonize the code.

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
tomekl007: done

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
tomekl007: done

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -75,6 +77,34 @@ public static int getProcessId() {
...
+   * Platform#getCPU()}.
+   *
+   * @return the current processor architecture.
+   * @throws UnsupportedOperationException if JNR Platform library is not loaded.
+   */
+  public static String getCPU() {
+    if (!isPlatformAvailable())
+      throw new IllegalStateException(
+          "JNR Platform class not loaded. "
+              + "Check isGetProcessIdAvailable() before calling this method.");
```
adutra: Should be: "Check isPlatformAvailable()".

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
tomekl007: done, sorry for that copy/pasting

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -75,6 +77,34 @@ public static int getProcessId() {
...
+      return false;
+    }
+  }
+
+  /**
+   * Returns the current processor architecture the JVM is running on, as reported by {@link
+   * Platform#getCPU()}.
+   *
+   * @return the current processor architecture.
+   * @throws UnsupportedOperationException if JNR Platform library is not loaded.
```
adutra: Should be `IllegalStateException`

adutra (changes requested):

adutra (approved):

PR #1201, opened 2019-02-27
Fix flaky test for JAVA-2106

emerkle826:
Fixes a few issues with the tests.
- for the aggregate query without a partition key, the server only returns a warning in Cassandra 3.0+
- for the query that exceeds the query threshold, the number of warnings and the specific wording varies with different Cassandra versions
The tests should now pass against Cassandra 2.1, 2.2 3.0, and 3.11+

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
@@ -17,7 +17,7 @@
 package com.datastax.oss.driver.api.core.cql;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.Mockito.after;
+import static org.mockito.Mockito.timeout;
 import static org.mockito.Mockito.verify;
 
 import ch.qos.logback.classic.Level;
@@ -26,6 +26,7 @@ import ch.qos.logback.classic.spi.ILoggingEvent;
 import ch.qos.logback.core.Appender;
 import com.datastax.oss.driver.api.core.CqlSession;
 import com.datastax.oss.driver.api.core.config.DefaultDriverOption;
+import com.datastax.oss.driver.api.testinfra.CassandraRequirement;
 import com.datastax.oss.driver.api.testinfra.ccm.CustomCcmRule;
 import com.datastax.oss.driver.api.testinfra.session.SessionRule;
 import com.datastax.oss.driver.api.testinfra.session.SessionUtils;
@@ -34,10 +35,10 @@ import com.datastax.oss.driver.internal.core.cql.CqlRequestHandler;
 import com.datastax.oss.driver.internal.core.tracker.RequestLogger;
 import com.google.common.base.Strings;
 import java.util.List;
+import java.util.stream.Collectors;
 import org.junit.After;
 import org.junit.Before;
-import org.junit.BeforeClass;
-import org.junit.ClassRule;
+import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.RuleChain;
 import org.junit.rules.TestRule;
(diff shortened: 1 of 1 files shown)


integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
```diff
@@ -169,24 +165,36 @@ public void should_expose_warnings_on_execution_info() {
...
     Statement<?> st = SimpleStatement.builder(String.format(query)).build();
-    ResultSet result = SESSION_RULE.session().execute(st);
+    ResultSet result = sessionRule.session().execute(st);
     ExecutionInfo executionInfo = result.getExecutionInfo();
     assertThat(executionInfo).isNotNull();
     List<String> warnings = executionInfo.getWarnings();
     assertThat(warnings).isNotEmpty();
     // verify the log was generated
-    verify(appender, after(500).times(1)).doAppend(loggingEventCaptor.capture());
+    verify(appender, after(500).atLeast(1)).doAppend(loggingEventCaptor.capture());
```
tomekl007: what is a reason why it works in at least once way?

adutra: I think it is explained in the comment above: "the number of warnings and the specific wording varies with different Cassandra versions".

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
```diff
@@ -169,24 +165,36 @@ public void should_expose_warnings_on_execution_info() {
...
+      String logMessage = log.getFormattedMessage();
+      // all logs should start with the driver log prefix
+      assertThat(logMessage)
+          .startsWith("Query '")
+          // query will only be logged up to MAX_QUERY_LENGTH characters
+          .contains(query.substring(0, RequestLogger.DEFAULT_REQUEST_LOGGER_MAX_QUERY_LENGTH))
+          .contains("' generated server side warning(s): ");
+      // we're now only looking for:
+      // "Batch ... for [KS.test] is of size .... exceeding specified threshold"
+      if (logMessage.contains("Batch")) {
```
adutra: Why not create an assertion for that? E.g.:

```
assertThat(logMessage)
    .contains("Batch")
    .contains(String.format("for [%s.test] is of size", sessionRule.keyspace().asCql(true)))
    .contains("exceeding specified threshold");
```
Besides, you are affecting true to `testPassed` inside a loop, which means that as long as the variable is set to true once, the whole test will pass, which imo is not what we want. Iow I think it would be better to get rid of the `testPassed`  variable.

adutra (changes requested):

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
emerkle826: Yes, that is correct. For Cassandra 2.2 and 3.0, this query generates 2 warnings, both of which are logged. In 3.11, only one warning is generated.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/ExecutionInfoWarningsIT.java
emerkle826: The reason for the loop and the flag is related to the multiple logs issue for certain versions of Cassandra. For versions 2.2 and 3.0, when 2 logs are generated, one of the logs does not contain any of the above assertions. And I'm not sure I am guaranteed the same order in the log capture. So I loop through the logs here, and when I come across one that has "Batch", I assert the rest of the message, set a flag and break out of the loop on line 195 below.

In the case where it's an older version of Cassandra, and I get two logs, this skips over the other log that doesn't appear in 3.11, either because it it doesn't contain "Batch", or the loop was broken out of on the log that did.

adutra: I see. Would the following work?

```
List<String> logMessages =
    loggingEventCaptor.getAllValues().stream()
        .map(ILoggingEvent::getFormattedMessage)
assertThat(logMessages)
    .hasOnlyOneElementSatisfying(
        logMessage -> {
          assertThat(logMessage)
              .contains("Batch")
              .contains(
                  String.format("for [%s.test] is of size", sessionRule.keyspace().asCql(true)))
              .contains("exceeding specified threshold");
        });
```

emerkle826: I can change it up to do this.

emerkle826: So I tried implementing this, but it still fails as `hasOnlyOneElementSatisfying` will fail the assertion if the List of logMessages has more than one element. I thought that method would verify that the list only contained one element that passes the assertion, but it in fact asserts that there is only one element AND that it passes the provided assertion.
I'm not seeing an easy way to assert that, no matter how many logs are in the Captor, there is only one that passes this specific set of assertions, without looping through the logs.

PR #1203, opened 2019-03-01
JAVA-2164: Rename statement builder methods to setXxx

tomekl007:


tomekl007:
Should we change `with*` to `set*` for `SessionBuilder` as well?

olim7t (changes requested):
Looks good on the code side 👍 
Please add a changelog entry.
The manual also needs to be updated. I did a text search for the regex `with\w` in all `*.md` files and found relevant matches in the following directories:
* faq/
* manual/core/:
  * idempotence/
  * paging/
  * performance/
  * query_timestamps/
  * statements/prepared/
  * statements/simple/
  * tracing/

olim7t:
I went ahead and did the manual, it was almost as fast as pointing out where the changes were.
The changelog is still to do.

tomekl007:
> I went ahead and did the manual, it was almost as fast as pointing out where the changes were.
> The changelog is still to do.

Thx! :) I've added the changelog entry

olim7t (approved):

PR #1204, opened 2019-03-01
java2174 quote empty identifier

tomekl007:


driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -241,11 +241,15 @@ public static String quoteIfNecessary(String id) {
   /**
    * We don't need to escape an identifier if it matches non-quoted CQL3 ids ([a-z][a-z0-9_]*), and
    * if it's not a CQL reserved keyword.
+   *
+   * <p>When 'Migrating from compact storage' after DROP COMPACT STORAGE on the table, it can have a
```
adutra, line 245: I would mention the jira here for future reference.

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -241,11 +241,15 @@ public static String quoteIfNecessary(String id) {
...
+   * column with an empty name. For that case, we need to escape empty column name.
    */
   private static boolean needsQuote(String s) {
     // this method should only be called for C*-provided identifiers,
     // so we expect it to be non-null and non-empty.
```
adutra: The comment needs to be changed, it's not true anymore.

adutra (changes requested):

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
emerkle826: Nit: we can remove the comment here about it being expected to be "non-empty", as you have it covered in the javadoc comment above (and it's no longer true)

emerkle826 (approved):

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
emerkle826: adutra beat me to it!

tolbertam (approved):

tolbertam:
Looks good to me :+1:, please add a changelog entry when merging.

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
```diff
@@ -241,11 +241,16 @@ public static String quoteIfNecessary(String id) {
   /**
    * We don't need to escape an identifier if it matches non-quoted CQL3 ids ([a-z][a-z0-9_]*), and
    * if it's not a CQL reserved keyword.
+   *
+   * <p>When 'Migrating from compact storage' after DROP COMPACT STORAGE on the table, it can have a
+   * column with an empty name. (See @jira_ticket JAVA-2174 for the reference) For that case, we
```
adutra: Sorry to be nit-picky but please remove the `@jira_ticket` tag, this is an internal tag that is only recognized in tests by some of our CI tools (I don't even remember which one now). In any case this isn't a valid javadoc tag so it shouldn't appear in production code.

adutra (changes requested):

changelog/README.md
```diff
@@ -18,6 +18,7 @@
 - [improvement] JAVA-1950: Log server side warnings returned from a query.
 - [improvement] JAVA-2123: Allow to use QueryBuilder for building queries against Materialized Views.
 - [bug] JAVA-2082: Avoid race condition during cluster close and schema refresh.
+- [bug] JAVA-2174: quote empty identifier
```
adutra: Also the correct changelog entry is `Metadata.needsQuote should accept empty strings.` (with a final period).

adutra (changes requested):

driver-core/src/main/java/com/datastax/driver/core/Metadata.java
tomekl007: thx, I was also not sure that It can be in production code. Changed

adutra (approved):

PR #1205, opened 2019-03-05
JAVA-2165: Abstract node connection information

olim7t:
Motivation:

Until now, the only information used to connect to a node was its socket
address (host+port). Internally, the driver also used this address as a
unique node identifier (it was the key in the Metadata.nodes map).
However, there are more complex deployment scenarios where the address
might not be sufficient: for example, if nodes are accessed through a
proxy that uses SNI routing, the connection information becomes
host+port+sni_server_name, and the host+port part is no longer unique.
We need to support those alternative connection methods, and use a
different unique node identifier.

Modifications:

- Introduce EndPoint as an abstract wrapper around connection
  information.  EndPoint.resolve() returns the socket address that will
  be used to open connections, but an endpoint may also contain
  additional pieces of information that will be used in other places
  (for example a custom SslHandlerFactory).
- Add Node.getEndPoint(), and modify DefaultTopologyMonitor to fill it.
  Allow contact points to be specified as EndPoints (preserving legacy
  methods).
- Replace InetSocketAddress by EndPoint everywhere that it was used to
  identify a node.
- Index Metadata.nodes by the true unique node identifier: host_id from
  system tables. This has two consequences:
  - we can't put contact points in Metadata.nodes directly, because we
    don't know their host_id yet. Store them separately
    (MetadataManager.contactPoints) and transfer them during the first
    node list refresh.
  - when we receive a status event, we can't use a simple map lookup
    anymore, so do a linear traversal to find the node with a matching
    broadcast_rpc_address (side effect: we don't need to translate it).

Result:

We no longer depend on nodes having a unique socket address.
We can plug in custom EndPoint implementations.


olim7t:
Note that all Simulacron multi-node tests are currently failing because of datastax/simulacron#72.

EDIT -- `ProtocolVersionMixedClusterIT` is one of those tests.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -142,6 +146,32 @@ public SelfT addContactPoints(@NonNull Collection<InetSocketAddress> contactPoin
...
+  }
+
+  /**
+   * Adds contact points to use for the initial connection to the cluster.
+   *
+   * <p>You only need this method if you use a custom {@link EndPoint} implementation. Otherwise,
+   * use {@link #addContactPoints(Collection)}.
+   */
+  @NonNull
+  public SelfT addContactEndPoints(@NonNull Collection<EndPoint> contactPoints) {
```
olim7t, line 160: Not a huge fan of the name, but I had to change it because it has the same erasure as the `InetSocketAddress` variant.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -272,10 +272,23 @@ private NodeInfo asNodeInfo(AdminRow row) {
...
+   */
   protected DefaultNodeInfo.Builder nodeInfoBuilder(
-      AdminRow row, InetSocketAddress connectAddress) {
-    DefaultNodeInfo.Builder builder = DefaultNodeInfo.builder().withConnectAddress(connectAddress);
+      AdminRow row, InetSocketAddress broadcastRpcAddress) {
+
+    // Deployments that use a custom EndPoint implementation will need their own TopologyMonitor.
+    // One simple approach is to extend this class and override this method.
+    EndPoint endPoint =
+        new DefaultEndPoint(context.getAddressTranslator().translate(broadcastRpcAddress));
```
olim7t, line 286: My initial plan was to completely remove `AddressTranslator`, but on second thought I decided to keep it. It fills a sweet spot for simple proxying scenarios where overriding the whole topology monitor would be overkill.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/EndPoint.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Encapsulates the information needed to open connections to a node.
+ *
+ * <p>By default, the driver assumes plain TCP connections, and this is just a wrapper around an
+ * {@link InetSocketAddress}. However, more complex deployment scenarios might use a custom
+ * implementation that contains additional information; for example, if the nodes are accessed
+ * through a proxy with SNI routing, an SNI server name is needed in addition to the proxy address.
+ */
+public interface EndPoint {
```
tomekl007, line 29: `EndPoint` -> `Endpoint`?

core/src/main/java/com/datastax/oss/driver/api/core/metadata/EndPoint.java
```diff
@@ -0,0 +1,47 @@
...
+  SocketAddress resolve();
+
+  /**
+   * Returns an alternate string representation for use in node-level metric names.
+   *
+   * <p>Because metrics names are path-like, dot-separated strings, raw IP addresses don't make very
+   * good identifiers. So this method will typically replace the dots by another character, for
+   * example {@code 127_0_0_1_9042}.
+   */
+  String asMetricPrefix();
```
tomekl007, line 46: are you sure that this method should be in the `EndPoint` class? I think that logic for creating `metricPrefix` will do the same escaping for every implementation. So maybe it should be moved to some other class and method will look like this: `String asMetricPrefix(InetSocketAddress addressAndPort)`, wdyt?

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Metadata.java
```diff
@@ -36,11 +37,45 @@
...
+  /**
+   * Finds the node with the given <em>untranslated</em> {@linkplain Node#getBroadcastRpcAddress()
+   * broadcast RPC address}, if it exists.
+   *
+   * <p>Note that this method performs a linear search of {@link #getNodes()}.
+   */
+  @NonNull
+  default Optional<Node> findNode(@NonNull InetSocketAddress broadcastRpcAddress) {
+    for (Node node : getNodes().values()) {
+      Optional<InetSocketAddress> o = node.getBroadcastRpcAddress();
```
tomekl007, line 72:
```
Optional<InetSocketAddress> o = node.getBroadcastRpcAddress();
      if (o.isPresent() && o.get().equals(broadcastRpcAddress)) {
        return Optional.of(node);
      }
    }
    return Optional.empty();
```
can be simplified to one line:
`node.getBroadcastRpcAddress().filter(o -> o.equals(broadcastRpcAddress));`

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultEndPoint.java
```diff
@@ -0,0 +1,84 @@
...
+  public String toString() {
+    return address.toString();
+  }
+
+  @Override
+  public String asMetricPrefix() {
+    return metricPrefix;
+  }
+
+  private static String buildMetricPrefix(InetSocketAddress addressAndPort) {
```
tomekl007, line 68: see note concerning `buildMetricPrefix` at https://github.com/datastax/java-driver/pull/1205/files#diff-d3237ad2c0149c6e12fa4a97dd528d78R46

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -301,40 +314,59 @@ private NodeInfo asNodeInfo(AdminRow row) {
...
+    if (port < 0) {
+      SocketAddress address = channel.getEndPoint().resolve();
+      if (address instanceof InetSocketAddress) {
+        port = ((InetSocketAddress) address).getPort();
+      }
     }
   }
 
-  private InetSocketAddress getNativeAddressFromPeers(AdminRow row) {
+  private InetSocketAddress getBroadcastRpcAddress(AdminRow row) {
```
tomekl007, line 354: Shouldn't this method be on the `AdminRow` class?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/InitialNodeListRefresh.java
```diff
@@ -0,0 +1,108 @@
...
+
+    return new Result(
+        oldMetadata.withNodes(newNodes, tokenMapEnabled, tokensChanged, tokenFactory, context),
+        eventsBuilder.build());
+  }
+
+  private DefaultNode findIn(Iterable<Node> nodes, EndPoint endPoint) {
+    for (Node node : nodes) {
+      if (node.getEndPoint().equals(endPoint)) {
+        return (DefaultNode) node;
```
tomekl007, line 104: Is this cast safe?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/InitialNodeListRefresh.java
```diff
@@ -0,0 +1,108 @@
...
+ * The first node list refresh: contact points are not in the metadata yet, we need to copy them
+ * over.
+ */
+@ThreadSafe
+class InitialNodeListRefresh extends NodesRefresh {
+
+  private static final Logger LOG = LoggerFactory.getLogger(InitialNodeListRefresh.class);
+
+  @VisibleForTesting final Iterable<NodeInfo> nodeInfos;
+  @VisibleForTesting final Set<Node> contactPoints;
```
tomekl007: shouldn't it be `Set<DefaultNode>`? I see that all additions and retrievals with that `Set` are casting to that `DefaultNode` implementation

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/LoadBalancingPolicyWrapper.java
```diff
@@ -198,11 +195,13 @@ private void processNodeStateEvent(NodeStateEvent event) {
...
 
-  private static Map<InetSocketAddress, Node> excludeDownHosts(Metadata metadata) {
-    ImmutableMap.Builder<InetSocketAddress, Node> nodes = ImmutableMap.builder();
-    for (Node node : metadata.getNodes().values()) {
+  private static Map<UUID, Node> excludeDownHosts(Metadata metadata) {
+    ImmutableMap.Builder<UUID, Node> nodes = ImmutableMap.builder();
+    for (Map.Entry<UUID, Node> entry : metadata.getNodes().entrySet()) {
+      UUID id = entry.getKey();
+      Node node = entry.getValue();
       if (node.getState() == NodeState.UP || node.getState() == NodeState.UNKNOWN) {
```
tomekl007, line 203: why are we treating `UKNOWN` as not down?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
```diff
@@ -274,28 +290,26 @@ private SingleThreaded(InternalDriverContext context, DriverExecutionProfile con
...
-        Set<InetSocketAddress> addresses, CompletableFuture<Void> initNodesFuture) {
-      apply(new InitContactPointsRefresh(addresses));
-      initNodesFuture.complete(null);
-    }
-
     private Void refreshNodes(Iterable<NodeInfo> nodeInfos) {
+      MetadataRefresh refresh =
+          didFirstNodeListRefresh
+              ? new FullNodeListRefresh(nodeInfos)
+              : new InitialNodeListRefresh(nodeInfos, contactPoints);
```
tomekl007, line 297: looking at the naming I don't see a difference between `full` and `initial`. Maybe it should be called `PartialNodeListRefresh`?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeStateManager.java
```diff
@@ -170,75 +172,87 @@ private void onDebouncedTopologyEvent(TopologyEvent event) {
...
+            }
+          } else {
             LOG.debug(
-                "[{}] Received UP event for unknown node {}, adding it", logPrefix, event.address);
-            metadataManager.addNode(event.address);
-          } else if (node.state == NodeState.FORCED_DOWN) {
-            LOG.debug("[{}] Not setting {} UP because it is FORCED_DOWN", logPrefix, node);
-          } else if (node.distance == NodeDistance.IGNORED) {
-            setState(node, NodeState.UP, "it is IGNORED and an UP topology event was received");
+                "[{}] Received UP event for unknown node {}, adding it",
```
tomekl007, line 187: In what situation it may happen that we are receiving event but we cannot identify a specific node?

core/src/main/java/com/datastax/oss/driver/api/core/metadata/EndPoint.java
olim7t: We can't predict what endpoints will be in the future, and address and port might not be enough to uniquely identify a node (as in the SNI proxy example). So it sounds like a good idea to abstract it here.
Maybe mentioning metrics is too specific, it could be called `toDotlessString` but that's a bit ugly.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
olim7t: `AdminRow` is a simplified row implementation used by internal admin tasks, but it's still a generic utility that can be used with any table.
This method is used specifically for `system.local` and `system.peers`, so I feel that it doesn't belong to the generic type.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/InitialNodeListRefresh.java
olim7t: Yes, the driver doesn't allow you to plug your own `Node` implementation. In this particular case (and a few others), the interface/implementation separation is only meant to hide irrelevant details to the end user.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/InitialNodeListRefresh.java
olim7t: Yes, in this particular case the set comes from `MetadataManager.getContactPoints`, which is purely internal, so I can make that change to avoid the casts.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/LoadBalancingPolicyWrapper.java
olim7t: `UNKNOWN` only indicates that the driver has not tried to connect yet, the LBP will still want to assign a distance to those nodes.

Actually while reading this I'm now questioning the relevance of even excluding `DOWN` nodes. I'll create another ticket.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
olim7t: Yeah I struggled to find a good name. Both are full refreshes of the node list. `initial` is the first one, where we have to copy over contact points; `full` is any one after that, where we don't.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeStateManager.java
olim7t: Normally never, the only way is if the driver metadata somehow got out of sync.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Metadata.java
olim7t: Actually the last line is outside of the loop. If `!o.isPresent()`, we want to keep iterating.

tomekl007: Ok, so it could be re-written to:
```
  for (Node node : getNodes().values()) {
      Optional<Node> node1 = node
          .getBroadcastRpcAddress()
          .filter(o -> o.equals(broadcastRpcAddress))
          .map(t -> node);
      if(node1.isPresent()){
        return node1;
      }
    }
    return Optional.empty();
```
but doubtful it is more readable

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/InitialNodeListRefresh.java
tomekl007: Since `findIn` is private we can also change the signature to `Iterable<DefaultNode>` to avoid the cast, right?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultEndPoint.java
```diff
@@ -0,0 +1,84 @@
...
+    StringBuilder prefix = new StringBuilder();
+    InetAddress address = addressAndPort.getAddress();
+    int port = addressAndPort.getPort();
+    if (address instanceof Inet4Address) {
+      // Metrics use '.' as a delimiter, replace so that the IP is a single path component
+      // (127.0.0.1 => 127_0_0_1)
+      prefix.append(address.getHostAddress().replace('.', '_'));
+    } else {
+      assert address instanceof Inet6Address;
```
scottcarey, line 77: This is a bug, in 4.0.0 I can not start my app that worked fine with 4.0.0-rc1.  

This will fail with an assertion error now:
`CqlSession.builder().addContactPoints(Arrays.asList(InetSocketAddress.createUnresolved("localhost", 4444)).withLocalDatacenter("datacenter1").build();`

You can not assume that the provided address inside the ISA is resolved.

scottcarey: perhaps it should just use the name from the ISA if the InetAddress is null?

PR #1206, opened 2019-03-06
Java 2178: QueryBuilder: Alias after function column is not included in a query

tomekl007:


adutra (approved):
Looks good, although I was really surprised that `FunctionSelector` extends `CollectionSelector`; I understand this was done to avoid a bit of code duplication, but it does look like a hack (the function name is passed to the superclass as part of the collection's opening sequence). I wonder if we shouldn't dissociate these two.

changelog/README.md
```diff
@@ -7,6 +7,7 @@
 - [improvement] JAVA-2158: Allow BuildableQuery to build statement with values
 - [improvement] JAVA-2150: Improve query builder error message on unsupported literal type
 - [documentation] JAVA-2149: Improve Term javadocs in the query builder
+- [bug] Java 2178: QueryBuilder: Alias after function column is not included in a query
```
adutra: In driver 4 you should add new entries at the top of the list, not the bottom.

adutra (changes requested):

changelog/README.md
tomekl007: ok, thx

tomekl007:
> Looks good, although I was really surprised that `FunctionSelector` extends `CollectionSelector`; I understand this was done to avoid a bit of code duplication, but it does look like a hack (the function name is passed to the superclass as part of the collection's opening sequence). I wonder if we shouldn't dissociate these two.

I was also surprised by that. +1 for looking into that but I think we should do it on a separate Jira Task.  I can create one

adutra (approved):

olim7t (commented):
Nit: the changelog entry and commit message don't follow the existing convention (JAVA-2178). It's a small detail but uniformity will make our life easier in the long run (think text searches, grepping a ticket in the git log, etc.).

Otherwise everything good 👍

tomekl007:
> Nit: the changelog entry and commit message don't follow the existing convention (JAVA-2178). It's a small detail but uniformity will make our life easier in the long run (think text searches, grepping a ticket in the git log, etc.).
> 
> Otherwise everything good 👍

Concerning changelog it is in exactly the same format as other changelogs:
```
- [bug] Java 2178: QueryBuilder: Alias after function column is not included in a query
- [improvement] JAVA-2150: Improve query builder error message on unsupported literal type
- [documentation] JAVA-2149: Improve Term javadocs in the query builder
```
So what is that existing convention?

adutra:
The jira ticket number should be in the form: JAVA-1234, to facilitate text searches.

PR #1207, opened 2019-03-07
JAVA-2182: Add insertInto().json() variant that takes an object in QueryBuilder

tomekl007:


query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/InsertInto.java
```diff
@@ -31,4 +37,29 @@
   /** Makes this statement an INSERT JSON with a bind marker, as in {@code INSERT JSON ?}. */
   @NonNull
   JsonInsert json(@NonNull BindMarker bindMarker);
+
+  /**
+   * Makes this statement an INSERT JSON with a custom type mapping, as in {@code INSERT JSON ?}.
```
adutra: The description is not accurate: `as in {@code INSERT JSON ?}` does not apply here.

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/InsertInto.java
```diff
@@ -31,4 +37,29 @@
...
+   * @throws CodecNotFoundException if {@code codecRegistry} does not contain any codec that can
+   *     handle {@code value}.
+   * @see DriverContext#getCodecRegistry()
+   * @see DefaultInsert#json(Object, CodecRegistry)
+   */
+  @NonNull
+  JsonInsert json(@Nullable Object value, @NonNull CodecRegistry codecRegistry);
+
+  /**
+   * Makes this statement an INSERT JSON with a custom type mapping, as in {@code INSERT JSON ?}.
```
adutra: Same thing here.

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/InsertInto.java
```diff
@@ -31,4 +37,29 @@
...
+   * The value will be turned into a string with {@link TypeCodec#format(Object)}, and inlined in
+   * the query.
+   *
+   * @see DefaultInsert#json(T, TypeCodec)
+   */
+  @NonNull
+  <T> JsonInsert json(@Nullable T value, @Nullable TypeCodec<T> codec);
```
adutra: I don't think `codec` can be null.

adutra (changes requested):

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/InsertInto.java
tomekl007: It is true. The json insert when using `json(String)` is:
```
INSERT INTO json_jackson_row JSON '{ "id": 2, "name": "Alice", "age": 3 }'
```
When using variants with codec:
```
INSERT INTO json_jackson_row JSON '{"id":2,"name":"Alice","age":3}'
```
So in both cases the `INSERT JSON` is used

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/InsertInto.java
tomekl007: yes, you right. Changed

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/InsertInto.java
```diff
@@ -15,8 +15,15 @@
  */
 package com.datastax.oss.driver.api.querybuilder.insert;
 
+import com.datastax.oss.driver.api.core.context.DriverContext;
+import com.datastax.oss.driver.api.core.type.codec.CodecNotFoundException;
+import com.datastax.oss.driver.api.core.type.codec.TypeCodec;
+import com.datastax.oss.driver.api.core.type.codec.registry.CodecRegistry;
 import com.datastax.oss.driver.api.querybuilder.BindMarker;
+import com.datastax.oss.driver.internal.querybuilder.insert.DefaultInsert;
+import com.sun.istack.internal.NotNull;
```
olim7t: Wrong annotation, the correct one is `NonNull` (next line).
I make that mistake all the time too 😢

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/InsertInto.java
```diff
@@ -31,4 +38,29 @@
...
+   * registry should contain a codec that can format the value. Typically, this will be your
+   * session's registry, which is accessible via {@code session.getContext().getCodecRegistry()}.
+   *
+   * @throws CodecNotFoundException if {@code codecRegistry} does not contain any codec that can
+   *     handle {@code value}.
+   * @see DriverContext#getCodecRegistry()
+   * @see DefaultInsert#json(Object, CodecRegistry)
+   */
+  @NonNull
+  JsonInsert json(@Nullable Object value, @NonNull CodecRegistry codecRegistry);
```
olim7t: I think this method could have a `default` implementation that calls the `TypeCodec` variant.

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/insert/DefaultInsert.java
```diff
@@ -95,6 +98,32 @@ public JsonInsert json(@NonNull BindMarker json) {
         keyspace, table, json, missingJsonBehavior, ImmutableMap.of(), timestamp, ifNotExists);
   }
 
+  @NonNull
+  @Override
+  public JsonInsert json(@Nullable Object value, @NonNull CodecRegistry codecRegistry) {
+    return new DefaultInsert(
+        keyspace,
+        table,
+        QueryBuilder.literal(value, (value == null) ? null : codecRegistry.codecFor(value)),
```
olim7t: We should catch `CodecNotFoundException` and rethrow with a more context-specific message, like we did in [JAVA-2150](https://datastax-oss.atlassian.net/browse/JAVA-2150). See `QueryBuilder.literal(Object, CodecRegistry)`.

query-builder/revapi.json
```diff
@@ -17,6 +17,28 @@
...
+      {
+        "code": "java.method.addedToInterface",
+        "new": "method <T> com.datastax.oss.driver.api.querybuilder.insert.JsonInsert com.datastax.oss.driver.api.querybuilder.insert.InsertInto::json(T, com.datastax.oss.driver.api.core.type.codec.TypeCodec<T>)",
+        "package": "com.datastax.oss.driver.api.querybuilder.insert",
+        "classQualifiedName": "com.datastax.oss.driver.api.querybuilder.insert.InsertInto",
+        "classSimpleName": "InsertInto",
+        "methodName": "json",
+        "newArchive": "com.datastax.oss:java-driver-query-builder:jar:4.0.0-rc2-SNAPSHOT",
+        "elementKind": "method",
+        "justification": "add missing Inert json() method"
```
olim7t: Typo: Inert=>Insert
Also, could you just copy over the changelog/commit message?

changelog/README.md
```diff
@@ -4,6 +4,7 @@
 
 ### 4.0.0 (in progress)
 
+- [improvement] JAVA-2182: QueryBuilder is missing json() method that takes object
```
olim7t: Could you describe what the commit _does_, instead of what's missing/not working before the commit? For example:
> Add insertInto().json() variant that takes an object in QueryBuilder

I know I'm being picky, and here it makes almost no difference, but for more complex changes I find that this approach produces better messages.

integration-tests/src/test/java/com/datastax/oss/driver/api/querybuilder/JsonInsertIT.java
```diff
@@ -0,0 +1,253 @@
...
+public class JsonInsertIT {
+  private static final CcmRule ccmRule = CcmRule.getInstance();
+  private static final JacksonJsonCodec<User> JACKSON_JSON_CODEC =
+      new JacksonJsonCodec<>(User.class);
+
+  private static SessionRule<CqlSession> sessionRule =
+      SessionRule.builder(ccmRule)
+          .withConfigLoader(
+              SessionUtils.configLoaderBuilder()
+                  .withDuration(DefaultDriverOption.REQUEST_TIMEOUT, Duration.ofSeconds(30))
```
olim7t, line 58: Note that you can also use `sessionRule.slowProfile()` on a per-statement basis:
```java
    sessionRule
        .session()
        .execute(
            SimpleStatement.builder(
                    "CREATE TABLE json_jackson_row(id int PRIMARY KEY, name text, age int)")
                .withExecutionProfile(sessionRule.slowProfile())
                .build());
```
So you don't necessarily need to override it for the whole session.

But this approach works too, both are valid.

integration-tests/src/test/java/com/datastax/oss/driver/api/querybuilder/JsonInsertIT.java
```diff
@@ -0,0 +1,253 @@
...
+      User user = new User(2, "bob", 35);
+      PreparedStatement pst =
+          session.prepare(insertInto("json_jackson_row").json(bindMarker("user")).build());
+
+      // when
+      session.execute(pst.bind().set("user", user, User.class));
+
+      // then
+      List<Row> rows = session.execute(selectFrom("json_jackson_row").json().all().build()).all();
+      Assertions.assertThat(rows.get(0).get(0, User.class)).isEqualTo(user);
```
olim7t: Nit: remove `Assertions.`, the method is statically imported.
(this happens in a few other places)

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/insert/DefaultInsert.java
```diff
@@ -29,6 +31,7 @@
 import com.datastax.oss.driver.internal.querybuilder.ImmutableCollections;
 import com.datastax.oss.driver.shaded.guava.common.base.Preconditions;
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableMap;
+import com.sun.istack.internal.NotNull;
```
olim7t: Wrong annotation here too.

integration-tests/src/test/java/com/datastax/oss/driver/api/querybuilder/JsonInsertIT.java
```diff
@@ -0,0 +1,253 @@
...
+import org.junit.After;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+
+@Category(ParallelizableTests.class)
+public class JsonInsertIT {
```
olim7t, line 49: Great coverage 👍 

I wonder if we need an integration test for this: the query builder only does text generation, so any part that requires a session here is not something that was impacted by this ticket. You could do all those checks in a unit test (building a `DefaultCodecRegistry` with your codec).

Integration tests take longer to run, so I think we should favor unit tests when we can.

query-builder/revapi.json
```diff
@@ -17,6 +17,28 @@
...
+        "code": "java.method.addedToInterface",
+        "new": "method com.datastax.oss.driver.api.querybuilder.insert.JsonInsert com.datastax.oss.driver.api.querybuilder.insert.InsertInto::json(java.lang.Object, com.datastax.oss.driver.api.core.type.codec.registry.CodecRegistry)",
+        "package": "com.datastax.oss.driver.api.querybuilder.insert",
+        "classQualifiedName": "com.datastax.oss.driver.api.querybuilder.insert.InsertInto",
+        "classSimpleName": "InsertInto",
+        "methodName": "json",
+        "newArchive": "com.datastax.oss:java-driver-query-builder:jar:4.0.0-rc2-SNAPSHOT",
+        "elementKind": "method",
+        "justification": "add missing Inert json() method"
+      }
```
olim7t: Revapi includes a lot of information in its suggestions, but in practice not every field is necessary. For this error code, the minimal set is `code, new, newArchive, justification`, all the others can be deduced.
For the sake of keeping this file short and easy to read, I think we should not include them.

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/InsertInto.java
olim7t: API Plumber could check that. It doesn't support annotations yet, but I've created an issue to add it: datastax/api-plumber-doclet#1

olim7t (changes requested):

query-builder/revapi.json
olim7t: It turns out that the Revapi plugin has an option to include the minimal necessary fields in suggestions. I've enabled it in 6194b1d9b.

query-builder/revapi.json
tomekl007: Good idea, done

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/InsertInto.java
tomekl007: After checking again the `codec` can be null because it proxy that into the `literal`. And `literal` is taking `codec` that is `@Nullable`:
```
literal(@Nullable T value, @Nullable TypeCodec<T> codec)
```
This is a case when the `value` that is passed to the `json` is `null`. Then the logic is:
```
value, (value == null) ? null : codecRegistry.codecFor(value)
```

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/insert/DefaultInsert.java
tomekl007: Good idea. But I see one problem here:
https://github.com/datastax/java-driver/pull/1194/files#diff-ec647efb880ba814afddcf79eaa95570R402
`@Nullable Object value` can be null but next, we are doing assert `value != null`.
Then there is a possiblity in this line:
https://github.com/datastax/java-driver/pull/1194/files#diff-ec647efb880ba814afddcf79eaa95570R412
that it will cause NPE, isn't it?
Shouldn't the logic be:
```
          String.format(
              "Could not inline literal of type %s. "
                  + "This happens because the driver doesn't know how to map it to a CQL type. "
                  + "Try passing a TypeCodec or CodecRegistry to literal().",
              (value == null) ? null : value.getClass().getName()),
          e);`
```
I will make it llike this in the `json()` `DefaultInsert`. Let me know wdyt. I can update logic in a `QueryBuilder` accordingly

integration-tests/src/test/java/com/datastax/oss/driver/api/querybuilder/JsonInsertIT.java
tomekl007: Thx for the suggestion, I've seen that approach in a code base. Here when we are reusing the same session multiple times I think that solution with doing it once is better because needs less boilerplate code in every test case

integration-tests/src/test/java/com/datastax/oss/driver/api/querybuilder/JsonInsertIT.java
tomekl007: I've was examining that approach to test it similarly to `JsonInsertTest` but I think that unit test for `json()` will not give us as much assurance as of the `IT`.
A unit test is only testing the generated SQL. For json with codec variant, we still need to create some class (like `User`) and codec that is able to process it. 
If we are doing this the next step is to just write an integration test. When I was implementing that functionality it was very useful to have the `e-2-e` test for that. Also, we have no integration test for `JSON` so I think it will be beneficial to have one. In the `JsonInsertIT` I am using `ParallelizableTests` so it shares the same ccm with other tests - it is reasonably fast (on my mac ~1s for all tests. With already started ccm by other tests it will be faster). 
So I would like to leave this test as the `IT`, wdyt?

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/InsertInto.java
adutra: But there is no bind marker here, so why are the javadocs mentioning `{@code INSERT JSON ?}`? (Note the question mark.)

tomekl007: Oh, right. Sorry for missing that. I've updated the doc.

integration-tests/src/test/java/com/datastax/oss/driver/api/querybuilder/JacksonJsonCodec.java
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  @Nullable
+  @Override
+  public ByteBuffer encode(@Nullable T value, @NonNull ProtocolVersion protocolVersion) {
+    if (value == null) return null;
+    try {
+      return ByteBuffer.wrap(objectMapper.writeValueAsBytes(value));
+    } catch (JsonProcessingException e) {
+      throw new InvalidTypeException(e.getMessage(), e);
```
adutra: As I said in #1208 : In driver 4.0 we don't have any specific exception for codecs except `CodecNotFoundException`. When a codec cannot serialize or deserialize something, we use `IllegalArgumentException`. I don't think you need to introduce a special exception in these examples.

adutra (changes requested):

integration-tests/src/test/java/com/datastax/oss/driver/api/querybuilder/JsonInsertIT.java
olim7t: OK, works for me.

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/insert/DefaultInsert.java
olim7t: `null` is a valid literal in some CQL queries, but `INSERT ... JSON null` is invalid. So we can mark the value as `@NonNull`.
(the only case against it was if a codec encoded `null` into something else than the CQL literal `null`, but that's too contrived to be realistic)

olim7t (approved):

olim7t:
I pushed another commit to make everything `@NonNull` and slightly alter the message (we can be a bit more specific than `literal()` because there is no object-only variant).
If you agree, OK to merge for me 👍 (please squash everything into one commit and rebase to solve the changelog conflict).

olim7t:
wrong button 😓

olim7t:
I forgot one thing: please remove the `newArchive` field in `revapi.json`, as it won't work well when we do the release.

integration-tests/src/test/java/com/datastax/oss/driver/api/querybuilder/JacksonJsonCodec.java
```diff
@@ -98,27 +98,16 @@ public String format(T value) {
   public T parse(String value) {
     if (value == null || value.isEmpty() || value.equalsIgnoreCase("NULL")) return null;
```
adutra: In driver 4 the contribution guidelines mandate that if/for/while blocks must be enclosed with braces. Same thing for the line below.

adutra (changes requested):

adutra (approved):

PR #1209, opened 2019-03-09
JAVA-2177: Don't exclude down nodes when initializing LBPs

olim7t:
Motivation:

In driver 4, the LBP "pushes" the distances to the driver, not the other
way around. So if we don't set the distance of a node in init(), it will
stay at IGNORED and the driver won't try to connect to it.
Therefore all nodes should be passed to init().

Modifications:

Don't filter the nodes in LoadBalancingPolicyWrapper.
Update the javadocs of LoadBalancingPolicy to explain the contract.
Update DefaultLoadBalancingPolicy to only add UP nodes to its live set.

Result:

Down nodes don't stay ignored anymore.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/loadbalancing/LoadBalancingPolicy.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/loadbalancing/LoadBalancingPolicy.java
@@ -32,13 +32,27 @@ public interface LoadBalancingPolicy extends AutoCloseable {
    * Initializes this policy with the nodes discovered during driver initialization.
    *
    * <p>This method is guaranteed to be called exactly once per instance, and before any other
-   * method in this class.
+   * method in this class. At this point, the driver has successfully connected to one of the
+   * contact points, and performed a first refresh of topology information (by default, the contents
+   * of {@code system.peers}), to discover other nodes in the cluster.
    *
-   * @param nodes the nodes discovered by the driver when it connected to the cluster. When this
-   *     method is invoked, their state is guaranteed to be either {@link NodeState#UP} or {@link
-   *     NodeState#UNKNOWN}. Node states may be updated concurrently while this method executes, but
-   *     if so you will receive a notification.
+   * <p>This method must call {@link DistanceReporter#setDistance(Node, NodeDistance)
+   * distanceReporter.setDistance} for each provided node (otherwise that node will stay at distance
+   * {@link NodeDistance#IGNORED IGNORED}, and the driver won't open connections to it). Note that
+   * the node's {@link Node#getState() state} can be either {@link NodeState#UP UP} (for the
+   * successful contact point), {@link NodeState#DOWN DOWN} (for contact points that were tried
+   * unsuccessfully), or {@link NodeState#UNKNOWN UNKNOWN} (for contact points that weren't tried,
+   * or any other node discovered from the topology refresh). Node states may be updated
+   * concurrently while this method executes, but if so this policy will get notified after this
+   * method has returned, through other methods such as {@link #onUp(Node)} or {@link
+   * #onDown(Node)}.
+   *
+   * @param nodes all the nodes that are known to exist in the cluster (regardless of their state)
+   *     at the time of invocation.
    * @param distanceReporter an object that will be used by the policy to signal distance changes.
+   *     Implementations will typically store a this in a field, since new nodes may get {@link
+   *     #onAdd(Node) added} later and will need to have their distance set (or the policy might
+   *     change distances dynamically over time).
    */
   void init(@NonNull Map<UUID, Node> nodes, @NonNull DistanceReporter distanceReporter);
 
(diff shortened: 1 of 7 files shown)


integration-tests/src/test/java/com/datastax/oss/driver/api/core/ConnectIT.java
```diff
@@ -127,6 +133,36 @@ public void should_cleanup_on_lbp_init_failure() {
...
+    BoundCluster boundCluster = simulacronRule.cluster();
+    boundCluster.node(0).rejectConnections(0, RejectScope.STOP);
+
+    try (CqlSession session = SessionUtils.newSession(simulacronRule)) {
+      Map<UUID, Node> nodes = session.getMetadata().getNodes();
+
+      // Before JAVA-2177, this would fail every other time because if the node was tried first for
+      // the initial connection, it was marked down and not passed to LBP.init(), and therefore
+      // stayed at distance IGNORED.
+      Node node0 = nodes.get(boundCluster.node(0).getHostId());
```
olim7t: Nice side-effect of JAVA-2165: finding a Simulacron node in the driver metadata is now straightforward.

core/src/test/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicyInitTest.java
```diff
@@ -129,6 +130,9 @@ public void should_warn_if_contact_points_not_in_local_dc() {
   public void should_include_nodes_from_local_dc() {
     // Given
     when(metadataManager.getContactPoints()).thenReturn(ImmutableSet.of(node1, node2));
+    when(node1.getState()).thenReturn(NodeState.UP);
```
tomekl007, line 133: maybe we can name `node1` - `upNode`, `node2` -> `downNode`, `node3` -> `uknownNode` for better readability

integration-tests/src/test/java/com/datastax/oss/driver/api/core/ConnectIT.java
```diff
@@ -127,6 +133,36 @@ public void should_cleanup_on_lbp_init_failure() {
...
+      assertThat(node0.getState()).isEqualTo(NodeState.DOWN);
+      assertThat(node0.getDistance()).isEqualTo(NodeDistance.LOCAL);
+      assertThat(node0.getOpenConnections()).isEqualTo(0);
+      assertThat(node0.isReconnecting()).isTrue();
```
tomekl007: should it be wrapped with the `ConditionChecker` and wait for it for some time? (I think it works in an async way)

core/src/main/java/com/datastax/oss/driver/api/core/loadbalancing/LoadBalancingPolicy.java
```diff
@@ -32,12 +32,20 @@
...
-   * method in this class.
+   * method in this class. At this point, the driver has successfully connected to one of the
+   * contact points, and performed a first refresh of topology information (by default, the contents
+   * of {@code system.peers}) to find out which other nodes exist in the cluster.
    *
-   * @param nodes the nodes discovered by the driver when it connected to the cluster. When this
-   *     method is invoked, their state is guaranteed to be either {@link NodeState#UP} or {@link
-   *     NodeState#UNKNOWN}. Node states may be updated concurrently while this method executes, but
-   *     if so you will receive a notification.
+   * @param nodes the nodes in the cluster. This method must call {@link
```
adutra: The parameter description is a bit too long imo, specially considering that it will be rendered in fixed-size font.

adutra (approved):

core/src/test/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicyInitTest.java
adutra: These are instance fields and they are used in other tests as well, so I guess it would be difficult to rename them since they might play different roles in different tests.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/ConnectIT.java
olim7t: Good catch 👍 👍 👍 
If the "good" contact point is tried first, the "bad" one will remain UNKNOWN and we only detect it DOWN when we try to open the pool to it.
Opening the pool is not racy: the driver waits for all channels to fail, and triggers a reconnection. But the way the node state is updated is that that reconnection fires `ChannelEvent.reconnectionStarted(node)` when it starts. Channel events are processed by `NodeStateManager` on a separate thread (in this particular case we'll enter the branch `it has no connections and started reconnecting`).

So indeed we need to wait with a condition checker.

core/src/main/java/com/datastax/oss/driver/api/core/loadbalancing/LoadBalancingPolicy.java
olim7t: OK I moved it to the main description.

PR #1211, opened 2019-03-11
JAVA-2161: Annotate mutating methods with @CheckReturnValue

adutra:


core/src/test/java/com/datastax/oss/driver/internal/core/data/AccessibleByIdTestBase.java
```diff
@@ -468,14 +468,15 @@ public void should_get_with_explicit_codec_by_name() {
     assertThat(s).isEqualTo("1");
   }
 
+  @SuppressWarnings("UnusedAssignment")
```
adutra, line 471: If you don't re-assign you get a compile error, so the only alternative here was to suppress the unused assignment warning.

olim7t: It's a good thing that the error is issued by the ErrorProne compiler, client applications are free to use it or not depending on the level of check that they desire.

changelog/README.md
```diff
@@ -4,6 +4,7 @@
 
 ### 4.0.0 (in progress)
 
+- [improvement] JAVA-2161: Annotate mutating methods with @CheckReturnValue
```
olim7t: Nit: could you enclose `@CheckReturnValue` in backticks, here and in the commit message? Sometimes GitHub interprets annotation names as @-mentions and sends notifications. I think the reason it's not doing it here is that no user with this name exists, but I've had this kind of issue in the past.

core/src/test/java/com/datastax/oss/driver/internal/core/data/DefaultUdtValueTest.java
```diff
@@ -70,8 +70,8 @@ public void should_serialize_and_deserialize() {
             .withField(CqlIdentifier.fromInternal("field2"), DataTypes.TEXT)
             .build();
     UdtValue in = type.newValue();
-    in.setBytesUnsafe(0, Bytes.fromHexString("0x00000001"));
-    in.setBytesUnsafe(1, Bytes.fromHexString("0x61"));
+    in = in.setBytesUnsafe(0, Bytes.fromHexString("0x00000001"));
+    in = in.setBytesUnsafe(1, Bytes.fromHexString("0x61"));
```
olim7t, line 74: Sounds perfectly reasonable to me to require that this is inherited by mutable children, after all this is just an implementation detail.

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/condition/ConditionalStatement.java
```diff
@@ -38,6 +39,7 @@
    * <p>If any column conditions were added before, they will be cleared.
    */
   @NonNull
+  @CheckReturnValue
   SelfT ifExists();
```
olim7t, line 43: Kudos for thinking of the query builder.

olim7t (approved):
LGTM 👍 I did a quick check of the rest of the public API, and I don't see any other place where we should use this.
I added a small blurb in the manual, as users might be interested to use ErrorProne in their build.

Feel free to merge at your convenience.

PR #1212, opened 2019-03-11
JAVA-2183: Enable materialized views when testing against Cassandra 4

olim7t:
To test, run `DescribeIT` against a local checkout of the Cassandra trunk:
```
-Dccm.version=4.0.0 -Dccm.directory=/path/to/cassandra
```
Note that the test still fails after views are enabled ([JAVA-2184](https://datastax-oss.atlassian.net/browse/JAVA-2184)), but you can see that it's a different error.

### DIFF
--- a/test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/DefaultCcmBridgeBuilderCustomizer.java
+++ b/test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/DefaultCcmBridgeBuilderCustomizer.java
@@ -15,9 +15,14 @@
  */
 package com.datastax.oss.driver.api.testinfra.ccm;
 
+import com.datastax.oss.driver.api.core.Version;
+
 public class DefaultCcmBridgeBuilderCustomizer {
 
   public static CcmBridge.Builder configureBuilder(CcmBridge.Builder builder) {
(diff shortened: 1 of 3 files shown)


test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/DefaultCcmBridgeBuilderCustomizer.java
```diff
@@ -15,9 +15,14 @@
  */
 package com.datastax.oss.driver.api.testinfra.ccm;
 
+import com.datastax.oss.driver.api.core.Version;
+
 public class DefaultCcmBridgeBuilderCustomizer {
 
   public static CcmBridge.Builder configureBuilder(CcmBridge.Builder builder) {
+    if (!CcmBridge.DSE_ENABLEMENT && CcmBridge.VERSION.compareTo(Version.V4_0_0) >= 0) {
```
tomekl007, line 23: Shouldn't we check here that DSE_Version is > `6.7.0`?

olim7t: We only need to enable views in Cassandra.

tomekl007 (approved):

PR #1213, opened 2019-03-11
JAVA-2189: Exclude virtual keyspaces from token map computation

olim7t:


tomekl007 (approved):

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SchemaIT.java
```diff
@@ -241,4 +245,29 @@ public void should_get_virtual_metadata() {
     assertThat(cm.getType()).isEqualTo(DataTypes.BIGINT);
     assertThat(cm.getName().toString()).isEqualTo("progress");
   }
+
+  @CassandraRequirement(min = "4.0", description = "virtual tables introduced in 4.0")
+  @Test
+  public void should_exclude_virtual_keyspaces_from_token_map() {
+    // Special case: DSE 6.0 reports C* 4.0 but does not support virtual tables
+    if (ccmRule.getDseVersion().isPresent()) {
```
adutra: Nit: this logic is repeated above, maybe you could factorize it?

integration-tests/src/test/resources/logback-test.xml
```diff
@@ -24,7 +24,7 @@
   <root level="ERROR">
     <appender-ref ref="STDOUT"/>
   </root>
-  <logger name="com.datastax.oss.driver" level="${driverLevel:-ERROR}"/>
+  <logger name="com.datastax.oss.driver" level="${driverLevel:-WARN}"/>
```
adutra, line 27: What is the motivation for this change?

adutra (approved):

integration-tests/src/test/resources/logback-test.xml
tomekl007: I guess because the problem manifests in the `WARN`:
```
15:49:53.465 [s0-admin-0] WARN  c.d.o.d.i.c.metadata.DefaultMetadata - [s0] Unexpected error while refreshing token map, keeping previous version (NullPointerException: Missing replication strategy class in {})
```

olim7t: See the commit message, I think WARN is too aggressive, it would indeed have hidden this issue.

PR #1215, opened 2019-03-13
JAVA-2182: Don't reference prod code from documentation in internal.* package

tomekl007:


query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/InsertInto.java
```diff
@@ -40,15 +38,6 @@
...
    * value} will be mapped to a JSON string.
-   *
-   * <p>This is an alternative to {@link #json(String)} for custom type mappings. The provided
-   * registry should contain a codec that can format the value. Typically, this will be your
-   * session's registry, which is accessible via {@code session.getContext().getCodecRegistry()}.
-   *
-   * @throws CodecNotFoundException if {@code codecRegistry} does not contain any codec that can
-   *     handle {@code value}.
-   * @see DriverContext#getCodecRegistry()
-   * @see DefaultInsert#json(Object, CodecRegistry)
```
adutra, line 51: Why remove so much? I think you could remove only the offending line: `@see DefaultInsert#json(Object, CodecRegistry)`

adutra (changes requested):

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/InsertInto.java
tomekl007: Ok, I thought that all `@see` and `code` are problematic. Reverted

adutra (approved):

PR #1217, opened 2019-03-18
JAVA-2192: Don't return generic types with wildcards

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/MappedAsyncPagingIterable.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core;
+
+import java.util.function.Function;
+
+/** The result of calling {@link #map(Function)} on another async iterable. */
+public interface MappedAsyncPagingIterable<ElementT>
```
tomekl007, line 21: I wonder why we need that interface. Is only used as a marker interface?

core/src/main/java/com/datastax/oss/driver/api/core/AsyncPagingIterable.java
```diff
@@ -101,7 +100,7 @@ default ElementT one() {
    * <p>Note that both instances share the same underlying data: consuming elements from the
    * transformed iterable will also consume them from this object, and vice-versa.
    */
-  default <TargetT> AsyncPagingIterable<TargetT> map(
+  default <TargetT> MappedAsyncPagingIterable<TargetT> map(
```
tomekl007, line 103: We can change it to:
```
 default <TargetT> AsyncPagingIterableWrapper<ElementT, TargetT> map(
      Function<? super ElementT, ? extends TargetT> elementMapper) {
    return new AsyncPagingIterableWrapper<>
        (this, elementMapper);
  }
```
Then we will eliminate the need for having `MappedAsyncPagingIterable` interface

core/src/main/java/com/datastax/oss/driver/api/core/MappedAsyncPagingIterable.java
olim7t: It's used to materialize `AsyncPagingIterable`'s self-type in calling code. Consider this for example:

```java
session
    .executeAsync("SELECT * FROM foo")
    .thenAccept(
        rs -> {
          MappedAsyncPagingIterable<String> mapped = rs.map(Object::toString);
          CompletionStage<MappedAsyncPagingIterable<String>> mapped2Future =
              mapped.fetchNextPage();
        });
```

If we want `map` to return `AsyncPagingIterable`, I think its signature would have to be:
```java
<TargetT, ResultT extends AsyncPagingIterable<TargetT, ResultT>> ResultT map(
      Function<? super ElementT, ? extends TargetT> elementMapper)
```

But then try to rewrite the first example.

core/src/main/java/com/datastax/oss/driver/api/core/AsyncPagingIterable.java
olim7t: `AsyncPagingIterableWrapper<SourceT, TargetT>` is an internal class.
`MappedAsyncPagingIterable<TargetT>` is how it's exposed in the public API.

It has a different name because the fact that it's a wrapper is an implementation detail, the only thing that matters to clients is that the elements have been transformed to something that is not a `Row`. That's also why it doesn't have the `SourceT` type parameter, calling code can't do anything with it so it's not relevant.

core/src/main/java/com/datastax/oss/driver/api/core/MappedAsyncPagingIterable.java
tomekl007: Ok, I see that it would be hard to use by the client because he would need to use the whole self-type when assigning `mapped` to a variable. Thx for clarification

tomekl007 (approved):

PR #1218, opened 2019-03-20
JAVA-2193: Fix flaky tests in WarningsTest

emerkle826:


driver-core/src/test/java/com/datastax/driver/core/WarningsTest.java
```diff
@@ -59,16 +60,16 @@ public void should_expose_warnings_on_execution_info() throws Exception {
           // query will only be logged up to QueryLogger.DEFAULT_MAX_QUERY_STRING_LENGTH characters
           .contains(query.substring(0, QueryLogger.DEFAULT_MAX_QUERY_STRING_LENGTH))
           .contains("' generated server side warning(s): ")
-          .contains(
-              String.format(
-                  "Batch for [%s.foo] is of size 5152, exceeding specified threshold of 5120 by 32.",
```
tomekl007, line 64: don't we want to validate that log contains the threshold and how much it was exceeded?

emerkle826: I would like to, but depending on the version of Cassandra/DSE being used, the value is a little different.
For Cassandra 2.2, it looks like this:
`Batch of prepared statements for [ks_1.foo] is of size 5154, exceeding specified threshold of 5120 by 34.`
For Cassandra 3.0, it looks like this:
`Batch of prepared statements for [ks_1.foo] is of size 5152, exceeding specified threshold of 5120 by 32.`
For Cassandra 3.11, it looks like this:
`Batch for [ks_1.foo] is of size 5152, exceeding specified threshold of 5120 by 32.`
So I've trimmed the log matching down to be just the common parts. Otherwise I'd have to keep the test up-to-date with conditions to check Cassandra versions and alter the expected text in the log.

tomekl007 (approved):

PR #1220, opened 2019-03-27
Configuration improvements (JAVA-2201 and JAVA-2205)

olim7t:
TODO:
- [x] document OngoingConfigOptions
- [x] annotate with `@CheckReturnValue`
- [x] remove URL tests

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
@@ -18,6 +18,7 @@ package com.datastax.oss.driver.api.core.config;
 import com.datastax.oss.driver.api.core.context.DriverContext;
 import com.datastax.oss.driver.api.core.session.SessionBuilder;
 import com.datastax.oss.driver.internal.core.config.typesafe.DefaultDriverConfigLoader;
+import com.datastax.oss.driver.internal.core.config.typesafe.DefaultProgrammaticDriverConfigLoaderBuilder;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
 import edu.umd.cs.findbugs.annotations.NonNull;
@@ -134,6 +135,64 @@ public interface DriverConfigLoader extends AutoCloseable {
                   .resolve();
           return config.getConfig(DefaultDriverConfigLoader.DEFAULT_ROOT_PATH);
         });
+  }
+
+  /**
+   * Starts a builder that allows configuration options to be overridden programmatically.
+   *
+   * <p>For example:
+   *
+   * <pre>{@code
+   * DriverConfigLoader loader =
+   *     DriverConfigLoader.programmaticBuilder()
+   *         .withDuration(DefaultDriverOption.REQUEST_TIMEOUT, Duration.ofSeconds(5))
+   *         .startProfile("slow")
+   *         .withDuration(DefaultDriverOption.REQUEST_TIMEOUT, Duration.ofSeconds(30))
+   *         .endProfile()
+   *         .build();
+   * }</pre>
+   *
+   * produces the same overrides as:
+   *
+   * <pre>
+   * datastax-java-driver {
+   *   basic.request.timeout = 5 seconds
+   *   profiles {
+   *     slow {
+   *       basic.request.timeout = 30 seconds
+   *     }
+   *   }
+   * }
+   * </pre>
+   *
+   * The resulting loader still uses the driver's default implementation (based on Typesafe config),
+   * except that the programmatic configuration takes precedence. More precisely, configuration
+   * properties are loaded and merged from the following (first-listed are higher priority):
+   *
+   * <ul>
+   *   <li>system properties
+   *   <li>properties that were provided programmatically
+   *   <li>{@code application.conf} (all resources on classpath with this name)
+   *   <li>{@code application.json} (all resources on classpath with this name)
+   *   <li>{@code application.properties} (all resources on classpath with this name)
+   *   <li>{@code reference.conf} (all resources on classpath with this name). In particular, this
+   *       will load the {@code reference.conf} included in the core driver JAR, that defines
+   *       default options for all mandatory options.
+   * </ul>
+   *
+   * Note that {@code application.*} is entirely optional, you may choose to only rely on the
+   * driver's built-in {@code reference.conf} and programmatic overrides.
+   *
+   * <p>The resulting configuration is expected to contain a {@code datastax-java-driver} section.
+   *
+   * <p>The loader will honor the reload interval defined by the option {@code
+   * basic.config-reload-interval}.
+   *
+   * <p>Note that the returned builder is <b>not thread-safe</b>.
+   */
+  @NonNull
+  static ProgrammaticDriverConfigLoaderBuilder programmaticBuilder() {
+    return new DefaultProgrammaticDriverConfigLoaderBuilder();
   }
 
   /**
--- a/core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
@@ -16,13 +16,125 @@
 package com.datastax.oss.driver.api.core.config;
 
 import com.datastax.oss.driver.api.core.context.DriverContext;
+import com.datastax.oss.driver.api.core.session.SessionBuilder;
+import com.datastax.oss.driver.internal.core.config.typesafe.DefaultDriverConfigLoader;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.io.File;
+import java.net.URL;
 import java.util.concurrent.CompletionStage;
 
 /**
  * Manages the initialization, and optionally the periodic reloading, of the driver configuration.
+ *
+ * @see SessionBuilder#withConfigLoader(DriverConfigLoader)
  */
 public interface DriverConfigLoader extends AutoCloseable {
+
+  /**
+   * Builds an instance using the driver's default implementation (based on Typesafe config), except
+   * that application-specific options are loaded from a classpath resource with a custom name.
+   *
+   * <p>More precisely, configuration properties are loaded and merged from the following
+   * (first-listed are higher priority):
+   *
+   * <ul>
+   *   <li>system properties
+   *   <li>{@code <resourceBaseName>.conf} (all resources on classpath with this name)
+   *   <li>{@code <resourceBaseName>.json} (all resources on classpath with this name)
+   *   <li>{@code <resourceBaseName>.properties} (all resources on classpath with this name)
+   *   <li>{@code reference.conf} (all resources on classpath with this name). In particular, this
+   *       will load the {@code reference.conf} included in the core driver JAR, that defines
+   *       default options for all mandatory options.
+   * </ul>
+   *
+   * The resulting configuration is expected to contain a {@code datastax-java-driver} section.
+   *
+   * <p>The returned loader will honor the reload interval defined by the option {@code
+   * basic.config-reload-interval}.
+   */
+  @NonNull
+  static DriverConfigLoader fromClasspath(@NonNull String resourceBaseName) {
+    return new DefaultDriverConfigLoader(
+        () -> {
+          ConfigFactory.invalidateCaches();
+          Config config =
+              ConfigFactory.defaultOverrides()
+                  .withFallback(ConfigFactory.parseResourcesAnySyntax(resourceBaseName))
+                  .withFallback(ConfigFactory.defaultReference())
+                  .resolve();
+          return config.getConfig(DefaultDriverConfigLoader.DEFAULT_ROOT_PATH);
+        });
+  }
+
+  /**
+   * Builds an instance using the driver's default implementation (based on Typesafe config), except
+   * that application-specific options are loaded from the given file.
+   *
+   * <p>More precisely, configuration properties are loaded and merged from the following
+   * (first-listed are higher priority):
+   *
+   * <ul>
+   *   <li>system properties
+   *   <li>the contents of {@code file}
+   *   <li>{@code reference.conf} (all resources on classpath with this name). In particular, this
+   *       will load the {@code reference.conf} included in the core driver JAR, that defines
+   *       default options for all mandatory options.
+   * </ul>
+   *
+   * The resulting configuration is expected to contain a {@code datastax-java-driver} section.
+   *
+   * <p>The returned loader will honor the reload interval defined by the option {@code
+   * basic.config-reload-interval}.
+   */
+  @NonNull
+  static DriverConfigLoader fromFile(@NonNull File file) {
+    return new DefaultDriverConfigLoader(
+        () -> {
+          ConfigFactory.invalidateCaches();
+          Config config =
+              ConfigFactory.defaultOverrides()
+                  .withFallback(ConfigFactory.parseFileAnySyntax(file))
+                  .withFallback(ConfigFactory.defaultReference())
+                  .resolve();
+          return config.getConfig(DefaultDriverConfigLoader.DEFAULT_ROOT_PATH);
+        });
+  }
+
+  /**
+   * Builds an instance using the driver's default implementation (based on Typesafe config), except
+   * that application-specific options are loaded from the given URL.
+   *
+   * <p>More precisely, configuration properties are loaded and merged from the following
+   * (first-listed are higher priority):
+   *
+   * <ul>
+   *   <li>system properties
+   *   <li>the contents of {@code url}
+   *   <li>{@code reference.conf} (all resources on classpath with this name). In particular, this
+   *       will load the {@code reference.conf} included in the core driver JAR, that defines
+   *       default options for all mandatory options.
+   * </ul>
+   *
+   * The resulting configuration is expected to contain a {@code datastax-java-driver} section.
+   *
+   * <p>The returned loader will honor the reload interval defined by the option {@code
+   * basic.config-reload-interval}.
+   */
+  @NonNull
+  static DriverConfigLoader fromUrl(@NonNull URL url) {
+    return new DefaultDriverConfigLoader(
+        () -> {
+          ConfigFactory.invalidateCaches();
+          Config config =
+              ConfigFactory.defaultOverrides()
+                  .withFallback(ConfigFactory.parseURL(url))
+                  .withFallback(ConfigFactory.defaultReference())
+                  .resolve();
+          return config.getConfig(DefaultDriverConfigLoader.DEFAULT_ROOT_PATH);
+        });
+  }
 
   /**
    * Loads the first configuration that will be used to initialize the driver.
--- a/core/src/main/java/com/datastax/oss/driver/api/core/config/DriverExecutionProfile.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/config/DriverExecutionProfile.java
@@ -37,7 +37,7 @@ import java.util.SortedSet;
  *
  * @see DriverConfig
  */
-public interface DriverExecutionProfile {
+public interface DriverExecutionProfile extends OngoingConfigOptions<DriverExecutionProfile> {
 
   /**
    * The name of the default profile (the string {@value}).
@@ -64,9 +64,6 @@ public interface DriverExecutionProfile {
   }
 
   @NonNull
-  DriverExecutionProfile withBoolean(@NonNull DriverOption option, boolean value);
-
-  @NonNull
   List<Boolean> getBooleanList(@NonNull DriverOption option);
 
   @Nullable
@@ -75,18 +72,11 @@ public interface DriverExecutionProfile {
     return isDefined(option) ? getBooleanList(option) : defaultValue;
   }
 
-  @NonNull
-  DriverExecutionProfile withBooleanList(
-      @NonNull DriverOption option, @NonNull List<Boolean> value);
-
   int getInt(@NonNull DriverOption option);
 
   default int getInt(@NonNull DriverOption option, int defaultValue) {
     return isDefined(option) ? getInt(option) : defaultValue;
   }
-
-  @NonNull
-  DriverExecutionProfile withInt(@NonNull DriverOption option, int value);
 
   @NonNull
   List<Integer> getIntList(@NonNull DriverOption option);
@@ -97,16 +87,11 @@ public interface DriverExecutionProfile {
     return isDefined(option) ? getIntList(option) : defaultValue;
   }
 
-  @NonNull
-  DriverExecutionProfile withIntList(@NonNull DriverOption option, @NonNull List<Integer> value);
-
   long getLong(@NonNull DriverOption option);
 
   default long getLong(@NonNull DriverOption option, long defaultValue) {
     return isDefined(option) ? getLong(option) : defaultValue;
   }
-
-  DriverExecutionProfile withLong(@NonNull DriverOption option, long value);
 
   @NonNull
   List<Long> getLongList(@NonNull DriverOption option);
@@ -116,17 +101,11 @@ public interface DriverExecutionProfile {
     return isDefined(option) ? getLongList(option) : defaultValue;
   }
 
-  @NonNull
-  DriverExecutionProfile withLongList(@NonNull DriverOption option, @NonNull List<Long> value);
-
   double getDouble(@NonNull DriverOption option);
 
   default double getDouble(@NonNull DriverOption option, double defaultValue) {
     return isDefined(option) ? getDouble(option) : defaultValue;
   }
-
-  @NonNull
-  DriverExecutionProfile withDouble(@NonNull DriverOption option, double value);
 
   @NonNull
   List<Double> getDoubleList(@NonNull DriverOption option);
@@ -138,18 +117,12 @@ public interface DriverExecutionProfile {
   }
 
   @NonNull
-  DriverExecutionProfile withDoubleList(@NonNull DriverOption option, @NonNull List<Double> value);
-
-  @NonNull
   String getString(@NonNull DriverOption option);
 
   @Nullable
   default String getString(@NonNull DriverOption option, @Nullable String defaultValue) {
     return isDefined(option) ? getString(option) : defaultValue;
   }
-
-  @NonNull
-  DriverExecutionProfile withString(@NonNull DriverOption option, @NonNull String value);
 
   @NonNull
   List<String> getStringList(@NonNull DriverOption option);
@@ -161,9 +134,6 @@ public interface DriverExecutionProfile {
   }
 
   @NonNull
-  DriverExecutionProfile withStringList(@NonNull DriverOption option, @NonNull List<String> value);
-
-  @NonNull
   Map<String, String> getStringMap(@NonNull DriverOption option);
 
   @Nullable
@@ -171,10 +141,6 @@ public interface DriverExecutionProfile {
       @NonNull DriverOption option, @Nullable Map<String, String> defaultValue) {
     return isDefined(option) ? getStringMap(option) : defaultValue;
   }
-
-  @NonNull
-  DriverExecutionProfile withStringMap(
-      @NonNull DriverOption option, @NonNull Map<String, String> value);
 
   /**
    * @return a size in bytes. This is separate from {@link #getLong(DriverOption)}, in case
@@ -189,20 +155,12 @@ public interface DriverExecutionProfile {
 
   /** @see #getBytes(DriverOption) */
   @NonNull
-  DriverExecutionProfile withBytes(@NonNull DriverOption option, long value);
-
-  /** @see #getBytes(DriverOption) */
-  @NonNull
   List<Long> getBytesList(DriverOption option);
 
   @Nullable
   default List<Long> getBytesList(DriverOption option, @Nullable List<Long> defaultValue) {
     return isDefined(option) ? getBytesList(option) : defaultValue;
   }
-
-  /** @see #getBytes(DriverOption) */
-  @NonNull
-  DriverExecutionProfile withBytesList(@NonNull DriverOption option, @NonNull List<Long> value);
 
   @NonNull
   Duration getDuration(@NonNull DriverOption option);
@@ -213,9 +171,6 @@ public interface DriverExecutionProfile {
   }
 
   @NonNull
-  DriverExecutionProfile withDuration(@NonNull DriverOption option, @NonNull Duration value);
-
-  @NonNull
   List<Duration> getDurationList(@NonNull DriverOption option);
 
   @Nullable
@@ -223,14 +178,6 @@ public interface DriverExecutionProfile {
       @NonNull DriverOption option, @Nullable List<Duration> defaultValue) {
     return isDefined(option) ? getDurationList(option) : defaultValue;
   }
-
-  @NonNull
-  DriverExecutionProfile withDurationList(
-      @NonNull DriverOption option, @NonNull List<Duration> value);
-
-  /** Unsets an option. */
-  @NonNull
-  DriverExecutionProfile without(@NonNull DriverOption option);
 
   /**
    * Returns a representation of all the child options under a given option.
--- a/core/src/main/java/com/datastax/oss/driver/api/core/config/OngoingConfigOptions.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/config/OngoingConfigOptions.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core.config;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+
+/** An object where config options can be set programmatically. */
+public interface OngoingConfigOptions<SelfT extends OngoingConfigOptions<SelfT>> {
+
+  @NonNull
+  SelfT withBoolean(@NonNull DriverOption option, boolean value);
+
+  @NonNull
+  SelfT withBooleanList(@NonNull DriverOption option, @NonNull List<Boolean> value);
+
+  @NonNull
+  SelfT withInt(@NonNull DriverOption option, int value);
+
+  @NonNull
+  SelfT withIntList(@NonNull DriverOption option, @NonNull List<Integer> value);
+
+  @NonNull
+  SelfT withLong(@NonNull DriverOption option, long value);
+
+  @NonNull
+  SelfT withLongList(@NonNull DriverOption option, @NonNull List<Long> value);
+
+  @NonNull
+  SelfT withDouble(@NonNull DriverOption option, double value);
+
+  @NonNull
+  SelfT withDoubleList(@NonNull DriverOption option, @NonNull List<Double> value);
+
+  @NonNull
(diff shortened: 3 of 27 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -16,14 +16,123 @@
...
+   *
+   * <p>More precisely, configuration properties are loaded and merged from the following
+   * (first-listed are higher priority):
+   *
+   * <ul>
+   *   <li>system properties
+   *   <li>{@code <resourceBaseName>.conf} (all resources on classpath with this name)
+   *   <li>{@code <resourceBaseName>.json} (all resources on classpath with this name)
+   *   <li>{@code <resourceBaseName>.properties} (all resources on classpath with this name)
+   *   <li>{@code reference.conf} (all resources on classpath with this name). In particular, this
```
tomekl007, line 48: What will happen if the user will create a file `reference.conf` and put it on it's classpath?

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -16,14 +16,123 @@
...
+   *
+   * The resulting configuration is expected to contain a {@code datastax-java-driver} section.
+   *
+   * <p>The returned loader will honor the reload interval defined by the option {@code
+   * basic.config-reload-interval}.
+   */
+  static DriverConfigLoader fromClasspath(String resourceBaseName) {
+    return new DefaultDriverConfigLoader(
+        () -> {
+          ConfigFactory.invalidateCaches();
```
tomekl007, line 62: Why we need that step?

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -16,14 +16,123 @@
...
+   *       will load the {@code reference.conf} included in the core driver JAR, that defines
+   *       default options for all mandatory options.
+   * </ul>
+   *
+   * The resulting configuration is expected to contain a {@code datastax-java-driver} section.
+   *
+   * <p>The returned loader will honor the reload interval defined by the option {@code
+   * basic.config-reload-interval}.
+   */
+  static DriverConfigLoader fromUrl(URL url) {
```
tomekl007: Shouldn't we have one generic method that is taking, for example, a `File`? Then methods like `fromUrl` or `fromClasspath` could only load a file using url or string and call the generic method that operates on file. That's why we will not need to have 3 almost identical methods.

Or another solution: we could create a generic method that takes a supplier function:
```
 static DriverConfigLoader from(Supplier<Config> fallbackSupplier ) {
    return new DefaultDriverConfigLoader(
        () -> {
          Config config =
                  .withFallback(fallabackSupplier)
                  .withFallback(ConfigFactory.defaultReference())
                  .resolve();
          return config.getConfig("datastax-java-driver");
        });
  }
```
And then call that method for every varian:
`from(() -> ConfigFactory.parseURL(url))`, `from(() -> ConfigFactory.parseFileAnySyntax(file)`, ...

core/src/test/resources/config/customApplication.conf
```diff
@@ -0,0 +1,3 @@
+datastax-java-driver {
+  basic.request.timeout = 5 seconds
+}
```
tomekl007, line 4: I think that we should add a property that has exactly the same key in each of those files(but different value) and write a test that validates which value is the final one (and which was overridden)

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
olim7t: The two files are merged, the order of precedence is undefined.

There is no reason for a client application to do that. One valid case where it might happen is if another dependency is using Typesafe config, but in that case the options won't collide thanks to the `datastax-java-driver` prefix.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
olim7t: For periodic config reloading. Otherwise Typesafe Config caches the values and does not detect changes to the file.

core/src/test/resources/config/customApplication.conf
olim7t: I'm not sure if Typesafe Config formally defines an order of precedence. It's very unlikely that someone would use multiple formats anyway. On that note, maybe I should even split the test into 3 ones using different names, but it's just convenient to do it all in one test.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
olim7t: > a generic method that takes a supplier function

That would work if it's a private helper (we don't want to leak `Config` through the public API). The only thing is that we can't have private methods in an interface, so it would have to live somewhere else. It's just a few lines of code duplicated so I thought it was not worth it.

tomekl007 (approved):

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -79,25 +79,23 @@
...
+   *   <li>{@code reference.conf} (all resources on classpath with this name). In particular, this
+   *       will load the {@code reference.conf} included in the core driver JAR, that defines
+   *       default options for all mandatory options.
    * </ul>
    *
-   * The core driver JAR includes a {@code reference.conf} file with sensible defaults for all
-   * mandatory options, except the contact points.
+   * The resulting configuration is expected to contain a {@code datastax-java-driver} section.
+   *
+   * <p>The returned loader will honor the reload interval defined by the option {@code
```
adutra: I think this paragraph doesn't apply here.

core/src/test/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoaderTest.java
```diff
@@ -174,4 +178,48 @@ public void should_not_notify_from_manual_reload_if_config_has_not_changed() {
...
+    assertThat(config.getDefaultProfile().getDuration(DefaultDriverOption.REQUEST_TIMEOUT))
+        .isEqualTo(Duration.ofSeconds(5));
+    // From reference.conf:
+    assertThat(config.getDefaultProfile().getString(DefaultDriverOption.REQUEST_SERIAL_CONSISTENCY))
+        .isEqualTo(DefaultConsistencyLevel.SERIAL.name());
+  }
+
+  @Test
+  public void should_load_from_url() throws Exception {
+    URL url = new File("src/test/resources/config/customApplication.conf").toURI().toURL();
```
adutra: From what I saw in the TypeSafe code this is not really exercising URLs because file URLs are special-cased, but not sure if we really want to got as far as creating other types of URLs for tests. Maybe we could create a special `URLStreamHandler` for a `foo` scheme? Or use WireMock?

core/src/test/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoaderTest.java
```diff
@@ -174,4 +178,48 @@ public void should_not_notify_from_manual_reload_if_config_has_not_changed() {
...
+    assertThat(config.getDefaultProfile().getString(DefaultDriverOption.REQUEST_SERIAL_CONSISTENCY))
+        .isEqualTo(DefaultConsistencyLevel.SERIAL.name());
+  }
+
+  @Test
+  public void should_load_from_file() {
+    File file = new File("src/test/resources/config/customApplication.conf");
+    DriverConfigLoader loader = DriverConfigLoader.fromFile(file);
+    DriverConfig config = loader.getInitialConfig();
+    // From customApplication.json:
```
adutra: You probably meant `customApplication.conf`.

core/src/test/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoaderTest.java
```diff
@@ -174,4 +178,48 @@ public void should_not_notify_from_manual_reload_if_config_has_not_changed() {
...
+    assertThat(config.getDefaultProfile().getString(DefaultDriverOption.REQUEST_SERIAL_CONSISTENCY))
+        .isEqualTo(DefaultConsistencyLevel.SERIAL.name());
+  }
+
+  @Test
+  public void should_load_from_url() throws Exception {
+    URL url = new File("src/test/resources/config/customApplication.conf").toURI().toURL();
+    DriverConfigLoader loader = DriverConfigLoader.fromUrl(url);
+    DriverConfig config = loader.getInitialConfig();
+    // From customApplication.json:
```
adutra: Same here, You probably meant `customApplication.conf`.

adutra (changes requested):
First round of review for JAVA-2205.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -133,6 +134,51 @@ static DriverConfigLoader fromUrl(URL url) {
...
+   *
+   * <p>The resulting configuration is expected to contain a {@code datastax-java-driver} section.
+   *
+   * <p>The loader will honor the reload interval defined by the option {@code
+   * basic.config-reload-interval}.
+   *
+   * <p>Note that the returned builder is <b>not thread-safe</b>.
+   */
+  @NonNull
+  static ProgrammaticConfigBuilder programmaticBuilder() {
```
adutra: 1. How can users combine JAVA-2205 and JAVA-2201, e.g. by using the programmatic API _and_ loading from a different classpath resource name? Is the idea to use `new DefaultProgrammaticConfigBuilder(Supplier<Config>, String)`?
2. Why not follow similar patterns in the code base and call this  interface `DriverConfigLoaderBuilder` (or maybe `ProgrammaticDriverConfigLoaderBuilder`)? Similarly, shouldn't this method be called just `builder()`?

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverExecutionProfile.java
```diff
@@ -37,7 +37,7 @@
  *
  * @see DriverConfig
  */
-public interface DriverExecutionProfile {
+public interface DriverExecutionProfile extends OngoingConfigOptions<DriverExecutionProfile> {
```
adutra, line 40: This bothers me a bit. The methods that were pulled up have different behaviors depending on the implementation class:

* `ProgrammaticConfigBuilder`: builder pattern applies, each invocation of these methods return the same object.
* `DriverExecutionProfile`: each invocation returns a different object (dynamic profile creation).

I am actually not sure that it is a good idea to group these methods in a common super-interface.

core/src/main/java/com/datastax/oss/driver/api/core/config/OngoingConfigOptions.java
```diff
@@ -0,0 +1,71 @@
...
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core.config;
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+
+public interface OngoingConfigOptions<SelfT extends OngoingConfigOptions<SelfT>> {
```
adutra, line 24: 1. Missing javadocs.
2. The name is a bit strange, but I guess you already explored all other options.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
```diff
@@ -127,7 +127,11 @@ public void close() {
    * <p>In the general case, use of this is not recommended, but it may be useful in situations
    * where configuration must be defined at runtime or is derived from some other configuration
    * source.
+   *
+   * @deprecated this feature is now available in the public API. Use {@link
+   *     DriverConfigLoader#programmaticBuilder()} instead.
    */
+  @Deprecated
```
adutra, line 136: Can't we simply remove it, since it's an internal package?

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticConfigBuilder.java
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  @NonNull
+  @Override
+  public ProgrammaticConfigBuilder withStringList(
+      @NonNull DriverOption option, @NonNull List<String> value) {
+    return with(option, value);
+  }
+
+  @SuppressWarnings("unchecked")
```
adutra: Unnecessary annotation.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticConfigBuilder.java
```diff
@@ -0,0 +1,225 @@
...
+
+  private ProgrammaticConfigBuilder with(@NonNull String path, @Nullable Object value) {
+    values.put(path, value);
+    return this;
+  }
+
+  @NonNull
+  @Override
+  public ProgrammaticConfigBuilder startProfile(@NonNull String profileName) {
+    Preconditions.checkNotNull(profileName);
```
adutra: 1. Nit: couldn't we switch from Guava to regular Java and use `Objects.requireNonNull` instead?
2. My IDE complains about the return value of these two methods not being assigned, so I'd suggest the following idiom instead:

```
currentProfileName = Objects.requireNonNull(profileName);
```

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticConfigBuilder.java
```diff
@@ -0,0 +1,225 @@
...
+import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Supplier;
+import net.jcip.annotations.NotThreadSafe;
+
+@NotThreadSafe
+public class DefaultProgrammaticConfigBuilder implements ProgrammaticConfigBuilder {
+
+  public static final Supplier<Config> DEFAULT_FALLBACK_SUPPLIER = ConfigFactory::defaultReference;
```
adutra: Could be private.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/config/DriverExecutionProfileIT.java
```diff
@@ -79,11 +78,8 @@ public void should_use_profile_request_timeout() {
     DriverConfigLoader loader =
         SessionUtils.configLoaderBuilder()
             .withDuration(DefaultDriverOption.REQUEST_TIMEOUT, Duration.ofSeconds(2))
-            .withProfile(
-                "olap",
-                DefaultDriverConfigLoaderBuilder.profileBuilder()
-                    .withDuration(DefaultDriverOption.REQUEST_TIMEOUT, Duration.ofSeconds(10))
-                    .build())
+            .startProfile("olap")
```
adutra, line 81: Nice 👍

integration-tests/src/test/java/com/datastax/oss/driver/api/core/ConnectIT.java
```diff
@@ -87,8 +87,9 @@ public void should_wait_for_contact_points_if_reconnection_enabled() throws Exce
     DriverConfigLoader loader =
         SessionUtils.configLoaderBuilder()
             .withBoolean(DefaultDriverOption.RECONNECT_ON_INIT, true)
-            .withClass(
```
adutra: Why remove `withClass`?

integration-tests/src/test/java/com/datastax/oss/driver/api/core/tracker/RequestLoggerIT.java
```diff
@@ -132,9 +120,14 @@
               DefaultDriverOption.REQUEST_LOGGER_MAX_VALUES,
               RequestLogger.DEFAULT_REQUEST_LOGGER_MAX_VALUES)
           .withBoolean(DefaultDriverOption.REQUEST_LOGGER_STACK_TRACES, true)
-          .withProfile("low-threshold", lowThresholdProfile)
```
adutra, line 142: I see that we cannot compose configs from other configs anymore. Not sure if this is worrisome?

adutra (changes requested):
And another round of feedback for JAVA-2201.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
olim7t: Right, it should say "the default loader", it's not returned.

core/src/test/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoaderTest.java
olim7t: 👍 , I fixed it in a later commit.

core/src/test/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoaderTest.java
olim7t: > From what I saw in the TypeSafe code this is not really exercising URLs

Good catch.
Yeah I'm not sure if it's worth starting a web container from the test just for this. It's not even our code. I think I'll just do a quick manual test to validate that it works, and then we can trust [Config's own tests](https://github.com/lightbend/config/blob/master/config/src/test/scala/com/typesafe/config/impl/HttpTest.scala).

integration-tests/src/test/java/com/datastax/oss/driver/api/core/tracker/RequestLoggerIT.java
olim7t: Yeah I lazily inlined everything, but I guess I could have instead created static methods that add the profiles to a given builder.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverExecutionProfile.java
olim7t: What I like is that the list of all possible setters lives in a single interface. If we add a method on execution profiles, we won't have to remember to add it on the programmatic builder every time.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
olim7t: 1. There's only so much we can do without leaking Typesafe config in the public API. I've tried to pick what seemed to be the most generic cases, but yes if you want programmatic _and_ URL-based it's not possible to compose, you'll have to switch to the internal API (and indeed use that constructor).
2. When we did the internal version there was a debate and we compromised on a crappy name. I want "programmatic" in the name (type and method) because that's what it is: the thing where you set options from your code instead of loading them from a file. Formally it should be `ProgrammaticDriverConfigLoaderBuilder` indeed, but that's a mouthful. So I shortened it to keep the bare minimum: it's programmatic, it relates to the config, and it's a builder. The fact that it produces a loader is implied by the return type of `build()`. I'm not hellbent on that, if you think it's not clear we can go back to the longer name.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
olim7t: Even though we allow ourselves to break the internal API, I think we should avoid it as much as possible.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverExecutionProfile.java
adutra: Which means that we are back to the same situation as in `BoundStatementBuilder` vs `BoundStatement`: you will have to annotate all the methods in this interface with `@CheckReturnValue` to warn people but it won't be necessary sometimes to re-assign the returned value. OK, I guess I can live with that.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
adutra: 1. OK
2. Agreed to have "programmatic" in the name. What bothers me is that the current name reads to me as "a programmatic builder for configs" while it actually is a "programmatic builder for loaders of configs". I personally prefer the long name.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/ConnectIT.java
olim7t: It has very little value, I thought it wasn't worth spreading the API surface (if we add it, why not also handle enum values, etc).

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
adutra: It bothers me that we already have deprecated stuff in a 4.0.1 release. I know we recommended some people out there to start using this internal API, but honestly atm I would say that no more than 2-3 users are actively using it.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/ConnectIT.java
adutra: I wonder if such methods, although of little value for developers, aren't appreciated by users. I would advocate for us supporting classes and enum values here (specially given that TypeSafe has built-in support for enum values). This is the typical type of nice-to-haves that makes the user experience more streamlined.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -16,14 +16,172 @@
...
+  static DriverConfigLoader fromUrl(@NonNull URL url) {
+    return new DefaultDriverConfigLoader(
+        () -> {
+          ConfigFactory.invalidateCaches();
+          Config config =
+              ConfigFactory.defaultOverrides()
+                  .withFallback(ConfigFactory.parseURL(url))
+                  .withFallback(ConfigFactory.defaultReference())
+                  .resolve();
+          return config.getConfig("datastax-java-driver");
```
tomekl007: can we create a constant `datastax-java-driver`?

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticConfigBuilder.java
```diff
@@ -0,0 +1,223 @@
...
+
+  private String currentProfileName = DriverExecutionProfile.DEFAULT_NAME;
+
+  /**
+   * @param fallbackSupplier the supplier that will provide fallback configuration for options that
+   *     haven't been specified programmatically.
+   * @param rootPath the root path used in non-programmatic sources (fallback reference.conf and
+   *     system properties).
+   */
+  public DefaultProgrammaticConfigBuilder(Supplier<Config> fallbackSupplier, String rootPath) {
```
tomekl007: Are we sure that we want to expose a method that takes different `rootPath` as the argument?
That was a community request to be able to put driver properties in a different path than `datastax-java-driver`?

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticConfigBuilder.java
```diff
@@ -0,0 +1,223 @@
...
+  @NonNull
+  @Override
+  public ProgrammaticConfigBuilder startProfile(@NonNull String profileName) {
+    currentProfileName = Objects.requireNonNull(profileName);
+    return this;
+  }
+
+  @NonNull
+  @Override
+  public ProgrammaticConfigBuilder endProfile() {
```
tomekl007: We don't have any test for `endProfile()`

integration-tests/src/test/java/com/datastax/oss/driver/api/core/config/DriverExecutionProfileIT.java
```diff
@@ -107,11 +103,8 @@ public void should_use_profile_request_timeout() {
   public void should_use_profile_default_idempotence() {
     DriverConfigLoader loader =
         SessionUtils.configLoaderBuilder()
-            .withProfile(
-                "idem",
-                DefaultDriverConfigLoaderBuilder.profileBuilder()
-                    .withBoolean(DefaultDriverOption.REQUEST_DEFAULT_IDEMPOTENCE, true)
-                    .build())
+            .startProfile("idem")
```
tomekl007, line 106: We don't have unit test for two profiles set programmatically

core/src/main/java/com/datastax/oss/driver/api/core/config/ProgrammaticConfigBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+   */
+  @NonNull
+  ProgrammaticConfigBuilder startProfile(@NonNull String profileName);
+
+  /**
+   * Ends the definition of a profile. All options set after this call, and before the next call to
+   * {@link #startProfile(String)}, will apply to the default profile.
+   */
+  @NonNull
+  ProgrammaticConfigBuilder endProfile();
```
tomekl007: Why do we need `endProfile` name? When the user creates a new profile it can use `startProfile(p1)`. Then when the user doesn't want to create another profile maybe it is enough to call `build`?
Otherwise, if he wants to define another profile he can call `startProfile(p2)` - in that way we are simplifying the API because there is no need to call `endProfile`

tomekl007: From the current implementation I see that endProfile() does:
`currentProfileName = DriverExecutionProfile.DEFAULT_NAME;`
So shouldn't it be named: `backToDefaultProfile` or smth similar?

core/src/main/java/com/datastax/oss/driver/api/core/config/ProgrammaticConfigBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+/**
+ * A builder that allows the creation of a config loader where options are overridden
+ * programmatically.
+ *
+ * @see DriverConfigLoader#programmaticBuilder()
+ */
+public interface ProgrammaticConfigBuilder extends OngoingConfigOptions<ProgrammaticConfigBuilder> {
```
tomekl007: I think that both options of loading config: 
`Programmatic` and `from-file` should extend an interface that has a name stating that it is a builder.
Right now if the user will want to use `ProgrammaticConfigBuilder` he will need to use this class a variable type.
When he will want to switch to `from file loader` he will need to use `DriverExecutionProfile` as a variable type.
He can use: `OngoingConfigOptions<T>` but the name OngoingConfigOptions doesn't say that this a builder.

core/src/test/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticConfigBuilderTest.java
```diff
@@ -0,0 +1,66 @@
...
+    DriverConfigLoader loader =
+        new DefaultProgrammaticConfigBuilder(() -> ConfigFactory.parseString(FALLBACK_CONFIG), "")
+            .startProfile("profile2")
+            .withInt(MockOptions.INT1, 3)
+            .build();
+    DriverConfig config = loader.getInitialConfig();
+    assertThat(config.getDefaultProfile().getInt(MockOptions.INT1)).isEqualTo(1);
+    assertThat(config.getProfile("profile1").getInt(MockOptions.INT1)).isEqualTo(11);
+    assertThat(config.getProfile("profile2").getInt(MockOptions.INT1)).isEqualTo(3);
+    assertThat(config.getProfile("profile2").getInt(MockOptions.INT2)).isEqualTo(2);
```
tomekl007: why the `INT2` for `profile2` == `2`? In fallback config it is not set per profile but globally:
`"int1 = 1\nint2 = 2\nprofiles.profile1 { int1 = 11 }";` so I think that this value should end up in a default profile.

The fact that the `.startProfile("profile2")` is setting global `currentProfileName` in the `DefaultProgrammaticConfigBuilder` is counter-intuitive and I think that users could be confused about it

core/src/test/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticConfigBuilderTest.java
```diff
@@ -0,0 +1,66 @@
...
+    DriverConfig config = loader.getInitialConfig();
+    assertThat(config.getDefaultProfile().getInt(MockOptions.INT1)).isEqualTo(1);
+    assertThat(config.getProfile("profile1").getInt(MockOptions.INT1)).isEqualTo(3);
+  }
+
+  @Test
+  public void should_override_option_in_new_profile() {
+    DriverConfigLoader loader =
+        new DefaultProgrammaticConfigBuilder(() -> ConfigFactory.parseString(FALLBACK_CONFIG), "")
+            .startProfile("profile2")
```
tomekl007: no test for > 1 `startProfiles`

integration-tests/src/test/java/com/datastax/oss/driver/api/core/tracker/RequestLoggerIT.java
tomekl007: I think it is useful to paste a profile. In users code, it may be often a case where profiles are created separately and finally, the list of profiles is set on the builder using `withProfile` for every entry.

tomekl007 (dismissed):

integration-tests/src/test/java/com/datastax/oss/driver/api/core/config/DriverExecutionProfileIT.java
olim7t: It's indirectly covered in integration tests. But yes, I can add unit tests for that too.

core/src/test/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticConfigBuilderTest.java
olim7t: Profiles inherit unset options from the default profile. The test builds the equivalent of this:
```
int1 = 1
int2 = 2
profiles {
  profile1 {
    int1 = 11
  }
  // Part that was set programmatically:
  profile2 {
    int1 = 3
  }
}
```
Both `profile1` and `profile2` inherit `int2`.
If we didn't inherit options, you would have to redefine everything in each profile.

core/src/main/java/com/datastax/oss/driver/api/core/config/ProgrammaticConfigBuilder.java
olim7t: > So shouldn't it be named: backToDefaultProfile or smth similar?

That's pretty much the same isn't it?
If you look at the structure of the file, you have the default profile at the root level, and then the `profiles` section. So it should be pretty clear that ending a profile goes back to the default profile. And if not, that's documented in the javadoc.

core/src/main/java/com/datastax/oss/driver/api/core/config/ProgrammaticConfigBuilder.java
olim7t: I'm not following. The setters on `DriverExecutionProfile` are used to create new profiles on an already loaded configuration (see [derived profiles](https://docs.datastax.com/en/developer/java-driver/4.0/manual/core/configuration/#derived-profiles) in the manual), that's not specific to loading from a file or not. And they don't follow a builder pattern.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
olim7t: Fair enough, I'll extract it.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticConfigBuilder.java
olim7t: No, but it's something that could be useful if you have two driver instances with different configurations co-existing in the same VM.

Note that this is the internal API, it's not exposed in the external one.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
olim7t: OK I'll change the name.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/tracker/RequestLoggerIT.java
olim7t: The same can be achieved with methods:
```java
ProgrammaticConfigBuilder addFooProfile(ProgrammaticConfigBuilder target) {
  return target
      .startProfile("foo")
      .withBoolean(DefaultDriverOption.REQUEST_LOGGER_STACK_TRACES, true)
      .endProfile();
}
```

integration-tests/src/test/java/com/datastax/oss/driver/api/core/ConnectIT.java
olim7t: I've readded withClass.

core/src/main/java/com/datastax/oss/driver/api/core/config/OngoingConfigOptions.java
```diff
@@ -15,57 +15,84 @@
...
 
   @NonNull
+  @CheckReturnValue
   SelfT withString(@NonNull DriverOption option, @NonNull String value);
 
+  /**
+   * Note that this is just a shortcut to call {@link #withString(DriverOption, String)} with {@code
+   * value.getName()}.
+   */
+  default SelfT withClass(@NonNull DriverOption option, @NonNull Class<?> value) {
```
adutra, line 58: Missing `@NonNull @CheckReturnValue`

adutra (dismissed):

core/src/main/java/com/datastax/oss/driver/api/core/config/ProgrammaticConfigBuilder.java
tomekl007: To be honest, that behavior of automatically getting back to default profile was not clear to me. 
I wonder if our users will not have a similar experience. 
I was using similar builders from Spring framework for configuration for example security and it was more explicit so maybe that is why it seems not intuitive to me.

olim7t:
Force-pushed to clean up the history (sorry I shouldn't have put 2 tickets in the same PR).

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverExecutionProfile.java
olim7t: Actually the methods were not annotated with `@CheckReturnValue` in `DriverExecutionProfile`. If I add the annotations now it will cause a Revapi violation.
Given that `withXxx` method are less ambiguous, and I don't expect derived profiles to be widely used, I prefer not adding another set of exclusions now.

core/src/main/java/com/datastax/oss/driver/api/core/config/OngoingConfigOptions.java
```diff
@@ -0,0 +1,82 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.time.Duration;
+import java.util.List;
+import java.util.Map;
+
+/** An object where config options can be set programmatically. */
+public interface OngoingConfigOptions<SelfT extends OngoingConfigOptions<SelfT>> {
+
+  @NonNull
+  SelfT withBoolean(@NonNull DriverOption option, boolean value);
```
adutra, line 27: Why don't you annotate these methods with `@CheckReturnValue`? This would be useful for `DriverExecutionProfile` implementations.

olim7t: See [this comment](https://github.com/datastax/java-driver/pull/1220#discussion_r271875069).

PR #1221, opened 2019-04-01
JAVA-2210: Add ability to set TTL for modification queries

emerkle826:
* Adds the `usingTtl` capability back to QueryBuilder
* fixes some broken documentation links


query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/Insert.java
```diff
@@ -44,4 +44,10 @@
    */
   @NonNull
   Insert usingTimestamp(@Nullable BindMarker bindMarker);
+
+  @NonNull
```
emerkle826, line 59: Need to add missing javadocs

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/update/UpdateStart.java
```diff
@@ -41,4 +41,10 @@
    */
   @NonNull
   UpdateStart usingTimestamp(@NonNull BindMarker bindMarker);
+
+  @NonNull
```
emerkle826, line 56: Again, add missing javadocs

query-builder/revapi.json
```diff
@@ -2752,6 +2752,26 @@
...
+      },
+      {
+        "code": "java.method.addedToInterface",
+        "new": "method com.datastax.oss.driver.api.querybuilder.update.UpdateStart com.datastax.oss.driver.api.querybuilder.update.UpdateStart::usingTtl(com.datastax.oss.driver.api.querybuilder.BindMarker)",
+        "justification": "JAVA-2210: Add ability to set TTL for modification queries"
+      },
+      {
+        "code": "java.method.addedToInterface",
+        "new": "method com.datastax.oss.driver.api.querybuilder.update.UpdateStart com.datastax.oss.driver.api.querybuilder.update.UpdateStart::usingTtl(int)",
+        "justification": "JAVA-2210: Add ability to set TTL for modification queries"
```
olim7t, line 2774: 👍

manual/query_builder/README.md
```diff
@@ -183,6 +183,6 @@ For a complete tour of the API, browse the child pages in this manual:
   * [Terms](term/)
   * [Idempotence](idempotence/)
   
-[QueryBuilder]:  http://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/query-builder/QueryBuilder.html
-[SchemaBuilder]: http://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/query-builder/SchemaBuilder.html
+[QueryBuilder]:  http://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/querybuilder/QueryBuilder.html
+[SchemaBuilder]: http://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/querybuilder/SchemaBuilder.html
```
olim7t, line 188: Thanks for keeping this in a separate commit.

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/insert/DefaultInsert.java
```diff
@@ -192,6 +220,14 @@ public String asCql() {
         builder.append(timestamp);
       }
     }
+    if (ttl != null) {
+      builder.append(" USING TTL ");
```
olim7t: If the timestamp was also specified, this must be `"  AND TTL "`.

Since there are only two possible clauses I don't think we need to get too fancy, we can just re-check `timestamp != null` to determine which keyword to use.

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/Insert.java
```diff
@@ -44,4 +44,10 @@
...
+  Insert usingTtl(int ttl);
```
olim7t: Good use of the naming conventions, we don't want to upper-case TTL here 👍

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/Insert.java
```diff
@@ -45,9 +45,23 @@
   @NonNull
   Insert usingTimestamp(@Nullable BindMarker bindMarker);
 
+  /**
+   * Adds a USING TTL clause to this statement with a literal value. Setting a value of {@code 0}
+   * will clear a previously set TTL.
```
olim7t: I find this a bit ambiguous because it's not clear whether this applies to the query builder or Cassandra.
* passing `null` removes the `USING TTL` clause from the generated query
* passing 0 will adds `USING TTL 0` clause to the generated query, which when executed will delete the TTL for the column(s)

olim7t (changes requested):

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/Insert.java
emerkle826: Good catch, I'll update the docs to reflect it.

manual/query_builder/insert/README.md
```diff
@@ -87,4 +87,20 @@ insertInto("user").json(bindMarker()).usingTimestamp(bindMarker())
 
 If you call the method multiple times, the last value will be used.
 
-[QueryBuilder]: http://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/query-builder/QueryBuilder.html
+### Time To Live (TTL)
+
+Similar to Timestamps, you can specify data expiration TTLs with a literal value:
```
adutra: I would harmonize the description between inserts and updates around stg along the lines of "You can generate a `USING TTL` clause that will cause column values to be deleted (marked with a tombstone) after the
specified time (in seconds) has expired. This can be done with a literal: ...".

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/Insert.java
adutra: 1. Please add a `@param` tag to the javadocs and specify that the value must be understood in seconds, I think it's important to note that.
2. I would also rename the parameter to `ttlInSeconds`.

That brings the question: should we provide a method overload that takes a `Duration` instead? That could be nice, but then we would have to cope with sub-second durations, etc. Not sure if it's really helpful.

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/Insert.java
```diff
@@ -44,4 +44,24 @@
...
+
+  /**
+   * Adds a USING TTL clause to this statement with a bind marker. Setting a value of {@code 0} will
+   * clear a previously set TTL.
+   *
+   * <p>If this method or {@link #usingTtl(int) } is called multiple times, the value from the last
+   * invocation is used.
+   */
+  @NonNull
+  Insert usingTtl(@Nullable BindMarker bindMarker);
```
adutra, line 74: Again I would include a `@param` tag and specify that the bind marker should be understood in seconds.

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/update/DefaultUpdate.java
```diff
@@ -154,6 +180,15 @@ public String asCql() {
       }
     }
 
+    if (ttl != null) {
+      builder.append(" USING TTL ");
```
adutra: Same here, we need to check if both clauses are present and append an `AND` if required.

query-builder/src/test/java/com/datastax/oss/driver/api/querybuilder/insert/RegularInsertTest.java
```diff
@@ -74,4 +74,10 @@ public void should_generate_if_not_exists_and_timestamp_clauses() {
     assertThat(insertInto("foo").value("a", bindMarker()).ifNotExists().usingTimestamp(1))
         .hasCql("INSERT INTO foo (a) VALUES (?) IF NOT EXISTS USING TIMESTAMP 1");
   }
+
+  @Test
+  public void should_generate_ttl_clause() {
+    assertThat(insertInto("foo").value("a", bindMarker()).usingTtl(10))
```
adutra, line 86: I suggest that we add tests for when both `USING TIMESTAMP` and `USING TTL` are present.

adutra (changes requested):

manual/query_builder/insert/README.md
```diff
@@ -87,4 +87,20 @@ insertInto("user").json(bindMarker()).usingTimestamp(bindMarker())
...
+
+```java
+insertInto("user").json(bindMarker()).usingTtl(500)
+// INSERT INTO user JSON ? USING TTL 500
```
tomekl007: maybe our example should not use JSON to not complicate that? Only simple bindMarker for field and TTL to focus the attention of the reader on the TTL part of the example?

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/Insert.java
tomekl007: For builders, we are extensively using `Duration` so I am +1 on that. I am not sure if we should still keep `int ttlInSeconds` method if we would provide the method that takes `Duration`

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/update/UpdateStart.java
```diff
@@ -41,4 +41,24 @@
...
+
+  /**
+   * Adds a USING TTL clause to this statement with a bind marker. Setting a value of {@code 0} will
+   * clear a previously set TTL.
+   *
+   * <p>If this method or {@link #usingTtl(int) } is called multiple times, the value from the last
+   * invocation is used.
+   */
+  @NonNull
+  UpdateStart usingTtl(@NonNull BindMarker bindMarker);
```
tomekl007, line 71: I see that usingTtl methods are repeated in multiple interfaces. Maybe we should extract them to interface `ttlSupport` or smth similar and just implements that in all interfaces that need ttl support? (to not duplicate the code)

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/insert/DefaultInsert.java
```diff
@@ -62,6 +63,7 @@ public DefaultInsert(
       @Nullable MissingJsonBehavior missingJsonBehavior,
       @NonNull ImmutableMap<CqlIdentifier, Term> assignments,
       @Nullable Object timestamp,
+      @Nullable Object ttl,
```
tomekl007: Shouldn't it be specific type instead of Object?

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/update/DefaultUpdate.java
tomekl007: Can't we extract that logic somewhere? (superclass)?

query-builder/src/test/java/com/datastax/oss/driver/api/querybuilder/insert/RegularInsertTest.java
tomekl007: And also for `AND TTL`

tomekl007 (changes requested):

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/update/UpdateStart.java
emerkle826: I would, but both `Insert` and `UpdateStart` inherit from different super interfaces already, and neither of those should have `usingTtl` added to them.

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/insert/DefaultInsert.java
emerkle826: In this case, it is possible to set the TTL as a literal, or a BindMarker. They don't share a common Object hierarchy outside of `Object`. Later in the code, where the cql statement is built, the logic checks to see if the TTL value is a bind marker or a literal and builds accordingly.

olim7t: We could introduce yet another wrapper class. But this is private code, local to an internal class. An `instanceof` check is good enough, we don't need to add more clutter.

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/update/UpdateStart.java
olim7t: Just for two methods that are repeated two times I think it's not worth it.

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/update/DefaultUpdate.java
olim7t: Again, not really worth it, this would "save" like 4 lines of code.

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/Insert.java
olim7t: I don't know. So far the query builder has stayed pretty close to the raw CQL syntax. If we go down that route, how about time types for the timestamp? I'd rather keep things simple rather than opening another can of worms.

emerkle826: I'll revert to just an `int` and name it `ttlInSeconds`

manual/query_builder/update/README.md
```diff
@@ -32,6 +32,35 @@ update("user").usingTimestamp(bindMarker());
...
+```
+
+You can clear a previously set TTL by setting the value to 0:
+
+```java
+update("user").usingTtl(0).setColumn("v", bindMarker()).whereColumn("k").isEqualTo(bindMarker());
+// UPDATE user USING TTL 0 SET v=? WHERE k=?
+```
+
+Setting the value to 0 will result in removing the TTL from the column in Cassandra when the query
```
adutra, line 59: I see that you updated this paragraph to distinguish the two situations, but now I think you should update the corresponding paragraph in the `Insert` documentation above.

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/update/DefaultUpdate.java
tomekl007: My worry is not duplication but the problem of remembering to add that specific part of the code to every implementation that is doing insert/modification. (to prevent bugs in future similar to what we had in https://datastax-oss.atlassian.net/browse/JAVA-2178)
Maybe `usingTtl` method should be created in the `OngoingAssignment` interface?

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/insert/DefaultInsert.java
```diff
@@ -192,6 +255,14 @@ public String asCql() {
         builder.append(timestamp);
       }
     }
+    if (ttlInSeconds != null) {
+      builder.append((timestamp != null) ? " AND " : " USING ").append("TTL ");
+      if (ttlInSeconds instanceof BindMarker) {
+        ((BindMarker) ttlInSeconds).appendTo(builder);
+      } else {
```
tomekl007, line 270: That `else` is worrying me. What if someone will pass some custom object as the `Object ttlInSeconds`?
We will set it on the builder and the user will get the error while executing the query, right?
Maybe we should do 
```
else if(ttlInSeconds instanceof Integer){
...
}else{
throw IllegalArgumentException()
}
```
To fail fast on not-supported type?

emerkle826: I see your point. I modeled this after the code for `timestamp`, as it supports a literal value and a BindMarker as well. For `ttlInSeconds`, I believe the literal can be an `Integer` or a `String` in this case, so long as it is a whole number. It could be a `Long` as well, as long as the value isn't larger than what is supported in Cassandra. In fact, it could be any Object that has a `toString()` implementation that returns a whole number between 0 and whatever the max is in Cassandra. 

We could be a little permissive and only throw the Exception if we can't parse an Integer value from the Object:
```
try {
  Integer ttlValue = Integer.parseInt(ttlInSeconds.toString());
  if (ttlValue.compareTo(0) < 0) {
    throw new IllegalArgumentException("Provided ttlInSeconds value can not be negative");
  }
  // add  USING TTL clause
} catch (NumberFormatException nfe) {
  throw new IllegalArgumentException("Provided ttlInSeconds is not a valid value");
}
```
I guess there is value in failing to build the Builder just prior to trying to have the Builder build the CQL. That would be the only benefit to failing at this point.

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/update/UpdateStart.java
```diff
@@ -41,4 +41,32 @@
...
+   * will update the data and remove any TTL on the column when the statement is executed,
+   * overriding any TTL (table or column) that may exist in Cassandra.
+   *
+   * <p>If this method or {@link #usingTtlInSeconds(int)} is called multiple times, the value from
+   * the last invocation is used.
+   *
+   * @param bindMarker A bind marker that is understood to be a value in seconds.
+   */
+  @NonNull
+  UpdateStart usingTtlInSeconds(@NonNull BindMarker bindMarker);
```
olim7t: I think the feedback was to rename the parameter, but I'm not sure about the method itself. I think it's good to keep it close to the CQL clause.
cc @adutra

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/update/DefaultUpdate.java
```diff
@@ -154,6 +180,19 @@ public String asCql() {
...
+    if (ttlInSeconds != null) {
+      // choose the correct keyword based on whether or not we have a timestamp
+      builder.append((timestamp != null) ? " AND " : " USING ").append("TTL ");
+      if (ttlInSeconds instanceof BindMarker) {
+        ((BindMarker) ttlInSeconds).appendTo(builder);
+      } else if (ttlInSeconds instanceof Integer) {
+        builder.append(ttlInSeconds);
+      } else {
+        throw new IllegalArgumentException(
+            "TTL value must be an Integer greater than or equal to 0");
```
olim7t: I would move the check to the constructor, there's no point in letting the user build an object that will fail later. While we're at it we can do the same for the timestamp.
Nit: the message should say Integer or BindMarker.

olim7t (changes requested):

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/update/UpdateStart.java
adutra: Yes, all I wanted was to see a mention somewhere that the bind marker is meant to be in seconds. The old method name was better imo.

emerkle826: I'll revert the method name change

olim7t (approved):

olim7t:
I moved the argument checks to the constructors of `DefaultInsert` and `DefaultUpdate`. LGTM.

tomekl007 (approved):

adutra (approved):

PR #1223, opened 2019-04-02
JAVA-2203: Handle unresolved addresses in DefaultEndPoint

olim7t:


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultEndPoint.java
```diff
@@ -65,20 +62,8 @@ public String asMetricPrefix() {
...
-      prefix.append(address.getHostAddress().replace('.', '_'));
-    } else {
-      assert address instanceof Inet6Address;
-      // IPv6 only uses '%' and ':' as separators, so no replacement needed
-      prefix.append(address.getHostAddress());
-    }
+  private static String buildMetricPrefix(InetSocketAddress address) {
     // Append the port since Cassandra 4 supports nodes with different ports
-    return prefix.append(':').append(port).toString();
+    return address.getHostString().replace('.', '_') + ':' + address.getPort();
```
olim7t: Note that, unlike the previous implementation, this will use the host name if that's what was provided (even for resolved addresses).
If users provide their contact points by name, I think it makes sense to use that for metrics too.

adutra: I see that `address.getHostString()` can return null, at least in theory (although it seems to be a pathological case). Do we want to defensively prevent against NPEs here?

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/DefaultEndPointTest.java
```diff
@@ -0,0 +1,50 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.net.InetSocketAddress;
+import org.junit.Test;
+
+public class DefaultEndPointTest {
+
+  @Test
+  public void should_create_from_host_name() {
+    DefaultEndPoint endPoint = new DefaultEndPoint(new InetSocketAddress("localhost", 9042));
```
adutra, line 31: You might want to add a few tests exercising null parameters.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultEndPoint.java
tomekl007: Maybe you can reuse class that I've created to format addresses in Insights related code?
```

class AddressFormatter {

  static String nullSafeToString(Object address) {
    if (address instanceof InetAddress) {
      return nullSafeToString((InetAddress) address);
    } else if (address instanceof InetSocketAddress) {
      return nullSafeToString((InetSocketAddress) address);
    } else if (address instanceof String) {
      return address.toString();
    } else {
      return "";
    }
  }

  static String nullSafeToString(InetAddress inetAddress) {
    return inetAddress != null ? inetAddress.getHostAddress() : null;
  }

  static String nullSafeToString(InetSocketAddress inetSocketAddress) {
    if (inetSocketAddress != null) {
      if (inetSocketAddress.isUnresolved()) {
        return String.format(
            "%s:%s",
            nullSafeToString(inetSocketAddress.getHostName()), inetSocketAddress.getPort());
      } else {
        return String.format(
            "%s:%s", nullSafeToString(inetSocketAddress.getAddress()), inetSocketAddress.getPort());
      }
    }
    return null;
  }
}
```

olim7t: That class does not exist in the OSS driver. Also one additional requirement is to replace dots in the generated string.

olim7t: @adutra yeah, I don't see how it can happen. The method is called directly from the constructor, so I'll just throw an IAE just in case.

tomekl007: > That class does not exist in the OSS driver. Also one additional requirement is to replace dots in the generated string.

I mean that we could cherry-pick that class.

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultEndPoint.java
olim7t: Should be good for now, it's just 2 lines of code and not exactly the same (replacing the dots + use getHostString instead of getHostName).

PR #1224, opened 2019-04-03
Fix wrong type signature in custom codecs documentation

aki-ks:
The documentation states the following:
```
For example, once your `int <-> String` codec is registered, you can also read a CQL
`list<int>` as a Java `List<Integer>`:
```

Since we convert from `int <-> String` you've probably meant we can now convert from `list<int> <-> List<String>`.


adutra (approved):

adutra:
@aki-ks did you sign our CLA? I thought you did but can't find the record now.

adutra:
Nvm, I found the record. All good here. Thanks for your contribution.

PR #1225, opened 2019-04-04
JAVA-2212 Add truncate to QueryBuilder

tomekl007:


emerkle826 (approved):
This looks straight forward. The only thing I would suggest is maybe a small README.md in the manual, similar to the other QueryBuilder docs there (https://github.com/datastax/java-driver/tree/4.x/manual/query_builder). Maybe a manual section for TRUNCATE is a bit overkill though.

tomekl007:
One question that I am not 100% sure. Should the `TRUNCATE` operation return `true` for `isIdempotent`? It seems that for `3.x` it returned true because from that logic:
```
 @Override
  public Boolean isIdempotent() {
    // If a value was forced with setIdempotent, it takes priority
    if (idempotent != null) return idempotent;

    // Otherwise return the computed value
    return !hasNonIdempotentOps();
  }
```
The `hasNonIdempotentOps` returned `false`

manual/query_builder/truncate/README.md
```diff
@@ -0,0 +1,16 @@
...
+case-sensitive identifiers or case-insensitive strings:
+
+```java
+import static com.datastax.oss.driver.api.querybuilder.QueryBuilder.*;
+
+InsertInto insertCaseInsensitve = truncate("ks", "table");
+InsertInto insertCaseSensitive = truncate(CqlIdentifier.fromCql("ks"), CqlIdentifier.fromCql("table"));
+```
+
+Note that, at this stage, the query is ready to built. After creating a TRUNCATE query it does not take any values.
```
emerkle826: typo:
`the query is ready to built` should be
`the query is ready to build`

emerkle826 (changes requested):
Just a typo in the docs, and you may want to add a link at the bottom of the query_builder README.md to the doc you created: https://github.com/datastax/java-driver/blob/4.x/manual/query_builder/README.md#building-queries

changelog/README.md
```diff
@@ -7,6 +7,7 @@
 - [improvement] JAVA-2211: Upgrade Jersey examples to fix security issue sid-3606
 - [bug] JAVA-2193: Fix flaky tests in ExecutionInfoWarningsIT
 - [improvement] JAVA-2197: Skip deployment of examples and integration tests to Maven central
+- [improvement] JAVA-2212: Add truncate to QueryBuilder
```
olim7t: Nit: please add the entry at the beginning

manual/query_builder/truncate/README.md
```diff
@@ -0,0 +1,16 @@
+## INSERT
```
olim7t: Should be TRUNCATE.

manual/query_builder/truncate/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+To create a TRUNCATE query, use one of the `truncate` methods in [QueryBuilder]. There are
+several variants depending on whether your table name is qualified, and whether you use
+case-sensitive identifiers or case-insensitive strings:
+
+```java
+import static com.datastax.oss.driver.api.querybuilder.QueryBuilder.*;
+
+InsertInto insertCaseInsensitve = truncate("ks", "table");
+InsertInto insertCaseSensitive = truncate(CqlIdentifier.fromCql("ks"), CqlIdentifier.fromCql("table"));
```
olim7t: InsertInto=>Truncate
Also the variables need to be renamed truncateXxx (and notice that there's an "i" missing in the first one CaseInsensit**i**ve).

Could you also add a comment with the generated CQL under each one? See the pages for the other statements for examples of how we've done it so far.

query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/QueryBuilder.java
```diff
@@ -470,4 +472,47 @@ public static UserDefinedType udt(@NonNull CqlIdentifier name) {
...
+   * Creates a new {@code TRUNCATE} query.
+   *
+   * @param keyspace the name of the keyspace to use.
+   * @param table the name of the table to truncate.
+   * @return the truncation query.
+   */
+  public static Truncate truncate(@Nullable String keyspace, @NonNull String table) {
+    return truncate(
+        keyspace == null ? null : CqlIdentifier.fromCql(keyspace), CqlIdentifier.fromCql(table));
+  }
```
olim7t, line 523: 👍 
Can you add a comment for the string variants to indicate that they use `CqlIdentifier.fromCql` for the conversion? See selectFrom at line 68 for example.

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/truncate/DefaultTruncate.java
```diff
@@ -0,0 +1,73 @@
...
+  @Override
+  public String asCql() {
+    StringBuilder builder = new StringBuilder();
+    builder.append("TRUNCATE ");
+    if (keyspace != null) {
+      builder.append(keyspace.asCql(true)).append(".");
+    }
+    builder.append(table.asCql(true));
+    return builder.toString();
+  }
```
olim7t, line 43: We have an internal utility to generate qualified names:
```java
    StringBuilder builder = new StringBuilder("TRUNCATE ");
    CqlHelper.qualify(keyspace, table, builder);
    return builder.toString();
```

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/truncate/DefaultTruncate.java
```diff
@@ -0,0 +1,73 @@
...
+
+  @NonNull
+  @Override
+  public SimpleStatementBuilder builder() {
+    return SimpleStatement.builder(asCql()).setIdempotence(IS_IDEMPOTENT);
```
olim7t: Please don't use a constant here. `setIdempotence(true)` is perfectly clear, and it's not like we might decide to use "false" to mean "idempotent" at any point in the future.

query-builder/src/main/java/com/datastax/oss/driver/internal/querybuilder/truncate/DefaultTruncate.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  @NonNull
+  @Override
+  public SimpleStatement build(@NonNull Object... values) {
+    throw new UnsupportedOperationException(
+        "TRUNCATE doesn't take values as parameters. Use build() method instead.");
```
olim7t, line 55: 👍 
We don't do any kind of validation on the values for other types of statements, but here it's trivial so why not.

olim7t (changes requested):

emerkle826 (approved):

olim7t (approved):

PR #1226, opened 2019-04-04
JAVA-2214: Fix flaky RequestLoggerIT test

emerkle826:
Also fixes a couple other test failures:
* Ensure JSON Insert tests are run against Cassandra 2.2+
* Ensure LifecycleListenerIT tests allow listener to close before asserting it is closed.
* Allow adminEventLoopGroup more time to complete admin tasks for DefaultSessionPoolsTest

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/api/core/context/LifecycleListenerIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/api/core/context/LifecycleListenerIT.java
@@ -75,7 +75,7 @@ public class LifecycleListenerIT {
       simulacronRule.cluster().acceptConnections();
     }
     assertThat(listener.ready).isFalse();
-    assertThat(listener.closed).isTrue();
+    ConditionChecker.checkThat(() -> listener.closed).before(1, SECONDS).becomesTrue();
   }
 
   private CqlSession newSession(TestLifecycleListener listener) {
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -4,6 +4,7 @@
 
 ### 4.0.1 (in progress)
 
+- [bug] JAVA-2214: Fix flaky RequestLoggerIT test
 - [bug] JAVA-2203: Handle unresolved addresses in DefaultEndPoint
 - [bug] JAVA-2210: Add ability to set TTL for modification queries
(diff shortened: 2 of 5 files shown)


integration-tests/src/test/java/com/datastax/oss/driver/api/core/context/LifecycleListenerIT.java
```diff
@@ -75,7 +75,7 @@ public void should_not_notify_listener_when_init_fails() {
       simulacronRule.cluster().acceptConnections();
     }
     assertThat(listener.ready).isFalse();
-    assertThat(listener.closed).isTrue();
+    ConditionChecker.checkThat(() -> listener.closed).before(1, SECONDS).becomesTrue();
   }
```
tomekl007, line 79: maybe if you already changing this test you can change it:
`try (CqlSession session = newSession(listener)) {`
to 
`try (CqlSession ignored = newSession(listener)) {`?
My IDE complains that `session` is not used and proper `ignored` name for that.
It can be changed for both tests:
`should_notify_listener_of_init_and_shutdown` and `should_not_notify_listener_when_init_fails`

integration-tests/src/test/java/com/datastax/oss/driver/api/core/context/LifecycleListenerIT.java
```diff
@@ -75,7 +75,7 @@ public void should_not_notify_listener_when_init_fails() {
       simulacronRule.cluster().acceptConnections();
```
tomekl007, line 75: in this test we have a logic:
```
  try (CqlSession session = newSession(listener)) {
      ConditionChecker.checkThat(() -> listener.ready).before(1, SECONDS).becomesTrue();
      assertThat(listener.closed).isFalse();
    }
    assertThat(listener.ready).isTrue();
    ConditionChecker.checkThat(() -> listener.closed).before(1, SECONDS).becomesTrue();
```
I am not sure why we need that check: ` assertThat(listener.ready).isTrue();`
after the try block. It is already validatied withing try block:
`ConditionChecker.checkThat(() -> listener.ready).before(1, SECONDS).becomesTrue();`
Maybe we can remove that second check?

tomekl007 (approved):

PR #1227, opened 2019-04-04
JAVA-2206: Emphasize statement immutability in the docs

olim7t:


emerkle826 (approved):
This looks good to me.
Maybe, at the risk of overkill, we should update the code example comments (like here https://github.com/datastax/java-driver/pull/1227/files#diff-d29e502ab89473f56c2c7279ad0e8b7cR39) to explicitly state that the object should be reassigned. For example, change:

```java
// Do this instead
```
to
```java
// Do this instead: reassign the statement object to the NEW instance returned
```

Again, maybe that's overkill. If it's not clear from all the bold-face `immutable`s, a few extra words in a comment of a code example may not either.

tomekl007 (approved):

PR #1229, opened 2019-04-10
JAVA-2129: Handle mapped collections of UDTs

olim7t:
This one is pretty convoluted, I recommend looking at the generated code in `Container_Helper` to get a sense of what's going on.

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/mapper/model/udts/Container.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/mapper/model/udts/Container.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.mapper.model.udts;
+
+import com.datastax.oss.driver.api.mapper.annotations.Entity;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.UUID;
+
+@Entity
+public class Container {
+
+  private UUID id;
+  private List<Type1> list;
+  private Map<String, List<Type1>> map1;
+  private Map<Type1, Set<List<Type2>>> map2;
+  private Map<Type1, Map<String, Set<Type2>>> map3;
+
+  public Container() {}
+
+  public Container(
+      UUID id,
+      List<Type1> list,
+      Map<String, List<Type1>> map1,
+      Map<Type1, Set<List<Type2>>> map2,
+      Map<Type1, Map<String, Set<Type2>>> map3) {
+    this.id = id;
+    this.list = list;
+    this.map1 = map1;
+    this.map2 = map2;
+    this.map3 = map3;
+  }
+
+  public UUID getId() {
+    return id;
+  }
+
+  public void setId(UUID id) {
+    this.id = id;
+  }
+
+  public List<Type1> getList() {
+    return list;
+  }
+
+  public void setList(List<Type1> list) {
+    this.list = list;
+  }
+
+  public Map<String, List<Type1>> getMap1() {
+    return map1;
+  }
+
+  public void setMap1(Map<String, List<Type1>> map1) {
+    this.map1 = map1;
+  }
+
+  public Map<Type1, Set<List<Type2>>> getMap2() {
+    return map2;
+  }
+
+  public void setMap2(Map<Type1, Set<List<Type2>>> map2) {
+    this.map2 = map2;
+  }
+
+  public Map<Type1, Map<String, Set<Type2>>> getMap3() {
+    return map3;
+  }
+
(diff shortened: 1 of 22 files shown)


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
```diff
@@ -68,52 +63,180 @@ public void addMembers(TypeSpec.Builder classBuilder) {
...
+        mappedValueName = rawValueName;
+      } else {
+        mappedValueName = enclosingClass.getNameIndex().uniqueField("mappedValue");
+        convertUdtsIntoEntities(rawValueName, mappedValueName, mappedValueType, getBuilder);
+      }
+      getBuilder
+          .addStatement("$L.put($L, $L)", mappedObjectName, mappedKeyName, mappedValueName)
+          .endControlFlow();
+    } else {
+      throw new AssertionError("Unsupported type " + type.asTypeName());
```
tomekl007, line 236: It means that we are not supporting `Map<Map<Map` (3 level of enclosing), right? 
What is a reason about it? Is it because Cassandra is not supporting it?

integration-tests/src/test/java/com/datastax/oss/driver/mapper/model/udts/Container.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.Set;
+import java.util.UUID;
+
+@Entity
+public class Container {
+
+  private UUID id;
+  private List<Type1> list;
+  private Map<String, List<Type1>> map1;
+  private Map<Type1, Set<List<Type2>>> map2;
```
tomekl007, line 31: shouldn't we have one additional type to test:
`private Map<Type1, Map<String, List<Type2>>> map2;`
?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/Classes.java
```diff
@@ -46,6 +52,9 @@ public Classes(Types typeUtils, Elements elementUtils) {
         typeUtils.erasure(elementUtils.getTypeElement(SettableByName.class.getName()).asType());
     this.gettableByNameType =
         typeUtils.erasure(elementUtils.getTypeElement(GettableByName.class.getName()).asType());
+    this.listElement = elementUtils.getTypeElement(List.class.getCanonicalName());
```
tomekl007, line 55: Is it possible that someone will use a different type of collection to read values from Cassandra? For example `Queue`? Maybe we should have fallback hander for a `Collection.class` type? We can only throw that it is not supported but maybe that case is already handled.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/PropertyType.java
```diff
@@ -0,0 +1,183 @@
...
+   * replacing every entity class by {@code UdtValue}.
+   */
+  public abstract TypeName asRawTypeName();
+
+  /**
+   * A type that does not contain any mapped entity.
+   *
+   * <p>Note that it can still be a collection, for example {@code Map<String, List<Integer>>}.
+   */
+  public static class Simple extends PropertyType {
```
tomekl007, line 86: Maybe the `PropertyType` should be enum instead of the class? It could be simpler for a caller because the compiler will check that all enum cases were covered and give warning otherwise (just a suggestion)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
olim7t: No, we support arbitrary levels of nesting, `PropertyType.EntityMap` can use itself as a key or value.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/model/udts/Container.java
olim7t: We could test an infinite number of types, it's just a matter of getting reasonable coverage... Any reason why we should add this one in particular?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/Classes.java
olim7t: That would add a significant amount of complexity because we'd need to make everything pluggable: how to create instances of this collection, how to add elements to it...

And that custom logic would need to be plugged into _generated_ code, which would make it even more complex.

We've never had that kind of request for mapper 3, so I don't think it's that common. And you always have the workaround of registering a custom codec to do the conversion.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/PropertyType.java
olim7t: `PropertyType` does not have a predefined set of values, because it wraps an underlying type mirror.
For example, a `PropertyType.Simple` can correspond to an `int`, a `String`, a `Map<String, List<Integer>>`, etc.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/model/udts/Container.java
tomekl007: We are covering `Map<List>` and `Map<Set>` as enclosed collections test so I thought that it will be a good idea to add `Map<Map>`

olim7t: OK done.

PR #1230, opened 2019-04-12
Upgrade Jacoco dependency for JDK 11 support

emerkle826:


tolbertam (approved):
:+1:

olim7t (approved):

olim7t:
We also need to do it in the DSE driver.

emerkle826:
> We also need to do it in the DSE driver.

already done.

PR #1231, opened 2019-04-12
JAVA-2226: Support IPv6 contact points in the configuration

olim7t:


core/src/test/java/com/datastax/oss/driver/internal/core/ContactPointsTest.java
```diff
@@ -0,0 +1,177 @@
...
+        .containsExactly(
+            new DefaultEndPoint(InetSocketAddress.createUnresolved("localhost", 9042)));
+  }
+
+  @Test
+  public void should_parse_host_and_port_and_resolve_all_a_records() throws UnknownHostException {
+    int localhostARecordsCount = InetAddress.getAllByName("localhost").length;
+    assumeTrue(
+        "This test assumes that localhost resolves to multiple A-records",
+        localhostARecordsCount >= 2);
```
olim7t, line 105: On MacOS I get 3 records, but I'm not sure if that's true for all systems.

adutra (approved):

core/src/test/java/com/datastax/oss/driver/internal/core/ContactPointsTest.java
```diff
@@ -0,0 +1,177 @@
...
+
+    assertThat(endPoints)
+        .containsExactly(new DefaultEndPoint(new InetSocketAddress("127.0.0.1", 9042)));
+  }
+
+  @Test
+  public void should_parse_ipv6_address_and_port_in_configuration() {
+    Set<EndPoint> endPoints =
+        ContactPoints.merge(Collections.emptySet(), ImmutableList.of("0:0:0:0:0:0:0:1:9042"), true);
+
```
tomekl007, line 83: Shouldn't we add a couple more tests for ipv6? 
Maybe all tests could be parameterized with ipv4 and ipv6 to run variants for both?

core/src/test/java/com/datastax/oss/driver/internal/core/ContactPointsTest.java
```diff
@@ -0,0 +1,177 @@
...
+    assertThat(endPoints).hasSize(localhostARecordsCount);
+    assertLog(
+        Level.INFO,
+        "Contact point localhost:9042 resolves to multiple addresses, will use them all");
+  }
+
+  @Test
+  public void should_ignore_malformed_host_and_port_and_warn() {
+    Set<EndPoint> endPoints =
+        ContactPoints.merge(Collections.emptySet(), ImmutableList.of("foobar"), true);
```
tomekl007, line 119: I think that we can create a parameterized test for `merge` that covers all cases

olim7t: I gave it a quick try, but I think it's not worth it. The parameters would be programmatic contact points, configuration contact points, whether to resolve, expected resulting endpoints, expected warnings:

```java
  @DataProvider
  public static Object[][] mergeScenarios() {
    final boolean resolve = true, doNotResolve = false; // for readability
    return new Object[][] {
      {
        Collections.emptySet(),
        ImmutableList.of("127.0.0.1:9042"),
        resolve,
        ImmutableSet.of(new DefaultEndPoint(new InetSocketAddress("127.0.0.1", 9042))),
        Collections.emptyList()
      },
    };
  }
```

That's harder to read than plain code, and not shorter. So considering that there are only a handful of tests in this class, it's better to favor clarity I think.

core/src/test/java/com/datastax/oss/driver/internal/core/ContactPointsTest.java
olim7t: The only thing we need to cover is that we use the last colon to separate the port. The IPv6 parsing itself is delegated to the JDK, so we don't need to test different formats.

I've added the short format `::2` in the test, but that's the same really.

PR #1233, opened 2019-05-08
Mapper 4 documentation

olim7t:
Early preview of the annotation javadocs. I've temporarily published the HTML version [here](http://olim7t.github.io/mapper_annotations/index.html) to facilitate reviews.

I will be adding the manual over the next few days.


olim7t:
Added manual as promised.

PR #1234, opened 2019-05-13
JAVA-2249: Remove stripTrailingZeroBytes from OPPToken

amandachow:
Remove stripTrailingZeroBytes from OPPToken
 
See https://datastax-oss.atlassian.net/projects/JAVA/issues/JAVA-2249
 
Modifications:

- Remove code to strip trailing zero bytes upon creation of OPPToken
- Remove corresponding unit test
- Changed the condition for padding bytes in the split logic. In the [existing behavior](https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/Token.java#L421), it looks like 4 bytes will always be added because the existing condition doesn't detect when `start.equals(end)`. Previously this went unnoticed because the 4 zero bytes would be stripped off. This came up in `OPPTokenFactoryTest.should_split_range_producing_empty_splits_near_ring_end()`

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Token.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Token.java
@@ -418,7 +418,8 @@ public abstract class Token implements Comparable<Token> {
             start = toBigInteger(oppStartToken.value, significantBytes);
             end = toBigInteger(oppEndToken.value, significantBytes);
             range = end.subtract(start);
-            if (addedBytes == 4 || range.compareTo(bigNumberOfSplits) >= 0) break;
+            if (addedBytes == 4 || start.equals(end) || range.compareTo(bigNumberOfSplits) >= 0)
+              break;
             significantBytes += 1;
             addedBytes += 1;
           }
@@ -486,21 +487,7 @@ public abstract class Token implements Comparable<Token> {
 
     @VisibleForTesting
     OPPToken(ByteBuffer value) {
-      this.value = stripTrailingZeroBytes(value);
-    }
-
-    /** @return A new ByteBuffer from the input Buffer with any trailing 0-bytes stripped off. */
-    private static ByteBuffer stripTrailingZeroBytes(ByteBuffer b) {
-      byte result[] = Bytes.getArray(b);
-      int zeroIndex = result.length;
-      for (int i = result.length - 1; i > 0; i--) {
-        if (result[i] == 0) {
-          zeroIndex = i;
-        } else {
-          break;
-        }
-      }
-      return ByteBuffer.wrap(result, 0, zeroIndex);
+      this.value = value;
     }
 
     @Override
--- a/driver-core/src/test/java/com/datastax/driver/core/OPPTokenFactoryTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/OPPTokenFactoryTest.java
@@ -46,20 +46,6 @@ public class OPPTokenFactoryTest {
   }
 
   @Test(groups = "unit")
-  public void should_strip_trailing_0_bytes() {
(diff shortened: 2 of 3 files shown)


adutra (approved):

driver-core/src/main/java/com/datastax/driver/core/Token.java
```diff
@@ -418,7 +418,8 @@ OPPToken hash(ByteBuffer partitionKey) {
             start = toBigInteger(oppStartToken.value, significantBytes);
             end = toBigInteger(oppEndToken.value, significantBytes);
             range = end.subtract(start);
-            if (addedBytes == 4 || range.compareTo(bigNumberOfSplits) >= 0) break;
+            if (addedBytes == 4 || start.equals(end) || range.compareTo(bigNumberOfSplits) >= 0)
```
adutra, line 421: @amandachow shouldn't we add the same extra check `start.equals(end)` to the other branch of the `if` block below? I don't think so (I couldn't think of any case where this could happen), but just wanted to confirm with you.

amandachow: I think it's correct to _not_ add this condition to the other case. For example if your start is 0x10 and your end is 0x1, and you don't pad any bytes, your significant bytes is 2 even though your split param may require more than 2 bytes in the computation. Since in this wraparound condition the ring length is added to it, it's correct to increment the significant bytes until the corresponding ring length accommodates the number of requested splits

adutra: We are on the same page then, thanks 👍

driver-core/src/test/java/com/datastax/driver/core/OPPTokenFactoryTest.java
```diff
@@ -45,20 +45,6 @@ public void should_split_range_producing_empty_splits_near_ring_end() {
...
-    Token fromStringWith0Bytes = factory.fromString("040000");
-
-    assertThat(with0Bytes).isEqualTo(without0Bytes).isEqualTo(fromStringWith0Bytes);
-
-    Token withMixed0Bytes = factory.fromString("0004000400");
-    Token withoutMixed0Bytes = factory.fromString("00040004");
-
-    assertThat(withMixed0Bytes).isEqualTo(withoutMixed0Bytes);
-  }
-
```
olim7t: Can we add a new test method to cover the additional `start.equals(end)` check?
For example:
```java
factory.split(
    token(Bytes.fromHexString("0x11")),
    token(Bytes.fromHexString("0x1100")),
    whatever)
```
should yield `0x1100` as the split boundaries.

olim7t (approved):
👍 Looks good to me

driver-core/src/test/java/com/datastax/driver/core/OPPTokenFactoryTest.java
amandachow: done 👍

driver-core/src/test/java/com/datastax/driver/core/OPPTokenFactoryTest.java
```diff
@@ -45,6 +46,15 @@ public void should_split_range_producing_empty_splits_near_ring_end() {
...
 
+  @Test(groups = "unit")
+  public void should_split_range_where_start_as_int_equals_end_as_int() {
+    Token start = token(Bytes.fromHexString("0x11"));
+    Token end = token(Bytes.fromHexString("0x1100"));
+
+    List<Token> splits = factory.split(start, end, 3);
+    assertThat(splits).containsExactly(end, end);
+  }
+
```
olim7t, line 57: 👍

adutra:
Thank you @amandachow!

PR #1235, opened 2019-05-16
JAVA-2042: Add mapper support for UPDATE queries

tomekl007:


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperUpdateMethodGenerator.java
```diff
@@ -0,0 +1,69 @@
...
+            .beginControlFlow("if (tableId == null)")
+            .addStatement("tableId = DEFAULT_TABLE_ID")
+            .endControlFlow()
+            .addStatement(
+                "$1T update = (keyspaceId == null)\n"
+                    + "? $2T.update(tableId)\n"
+                    + ": $2T.update(keyspaceId, tableId)",
+                UpdateStart.class,
+                QueryBuilder.class)
+            .addCode("$[return update");
```
tomekl007: One of the things that are different comparing` @Insert` generator is the moment of adding `USING TIMESTAMP | USING TTL"`. According to documentation https://docs.datastax.com/en/cql/3.3/cql/cql_reference/cqlUpdate.html and in our `Update` builder the `usingTimestamp` and `usingTtl` are available on the `UpdateStart` class.
Because of that, we should do it on this line. 
Unfortunately, we don't have annotation values (`customUsingClause`) in this class. Other solution will be to move this block (https://github.com/datastax/java-driver/pull/1235/files#diff-75706beabe9b72dbf270c073d07d201fR59 to line 64)
to 
`generatePrepareRequest` method
https://github.com/datastax/java-driver/pull/1235/files#diff-25ba2cdaf8c214d0d4a6f05f1c7922acR147.
In that case, this `generate()` method will return `UpdateStart`, not the `UpdateWithAssignments` (so we will be able to append usingTtl or using TImestamp in the `generatePrepareRequest(`.
But the problem is that we will need to pass `EntityDefinition entityDefinition` to `generatePrepareRequest` metho as well. It could be done in the `EntityHelperGenerator` method: `methodGenerator.generate(entityDefinition)`
but it is not present in `DaoImplementationGenerator` nor `MapperImplementationGenerator` classes - so it will need bigger refactor. @olim7t let me know wdyt

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGeneratorTest.java
```diff
@@ -23,6 +23,7 @@
 
 public class DaoImplementationGeneratorTest extends DaoMethodGeneratorTest {
 
+  // todo handle Update where there is no set
```
tomekl007: fails in this test are related to this line:
https://github.com/datastax/java-driver/pull/1235/files#diff-75706beabe9b72dbf270c073d07d201fR61
To set all columns for Update we need to get all non PKs columns: `entityDefinition.getRegularColumns`.
If there is a case (like in this test) that the Entity for which we are generating update prepared statement has only one column and that column is PK the generated Update statement is incorrect.
(The WHERE clause in Update is mandatory and it cannot contains `PK`). Because of that, the generated update statement will be incorrect. I am not sure how should we handle such a situation: maybe we should not generate update code path if the entity has only PK? But if the user is annotating the method with `@Update` for such entity it will be confusing for end-user: so other solution will be to throw an exception that we cannot generate update for the entity with only one column. Wdyt?

To make this test pass I can just add one additional non-PK key to `protected static final TypeSpec ENTITY_SPEC =` in the DaoMethodGeneratorTest but I am not 100% sure why in this test the Mapper tries to auto-generate  
```
@Override
  public UpdateWithAssignments update() {
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperUpdateMethodGenerator.java
olim7t: I see. We have three options:
1. make `EntityHelper.update` not return the statement directly, but take an existing statement and modify it. Then the DAO method would be responsible for starting the statement.
2. modify the query builder to allow `usingTimestamp/usingTtl` at later steps.
3. cast to `DefaultUpdate`, which gives us access to the methods again. (this is safe here because we control the creation of the statement, we know it's the default implementation)

I would have liked 2, but it breaks the API. So I think we should do 3.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGeneratorTest.java
olim7t: Yes, I think we should throw in corner cases like this. I handled a similar case in `EntityHelperDeleteByPrimaryKeyMethodGenerator`, you can see the generated `throw` statement at the beginning of the `generate()` method.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Update.java
```diff
@@ -0,0 +1,34 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.CLASS)
+public @interface Update {
+
+  boolean ifExists() default false;
+
+  String ifCondition() default "";
```
olim7t: Could you name this `customIfClause()` to be consistent with `@Delete`?

Also, `@Delete` requires the IF keyword in the string, this one doesn't, it should be the same everywhere.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/UpdateEntityIT.java
```diff
@@ -0,0 +1,270 @@
...
+  public interface InventoryMapper {
+    @DaoFactory
+    ProductDao productDao(@DaoKeyspace CqlIdentifier keyspace);
+  }
+
+  @Dao
+  public interface ProductDao {
+
+    @Update(whereClause = "id = :id")
+    void update(Product product);
```
olim7t, line 427: We shouldn't have to specify the where clause here, we know it's the PK since the method receives an entity instance, it can be generated automatically. For example, `@Delete` does it that way.

In fact  I think we could get rid of `whereClause` completely, the argument will always be an entity (you can do update queries that don't target an entity, but the mapper can really generate anything automatically in that case, so it's as simple to use `@Query` and provide the whole query string).

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -130,7 +130,7 @@ protected String getFileName() {
                 .getMessager()
                 .error(
                     methodElement,
-                    "Unrecognized method signature: no implementation will be generated");
+                    "[DaoImplementationGenerator] Unrecognized method signature: no implementation will be generated");
```
olim7t: I disagree: this will be surfaced to the user as a compiler error. The user has no idea what `DaoImplementationGenerator` is, the mapper processor is a black box to them.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateMethodGenerator.java
```diff
@@ -0,0 +1,189 @@
...
+import java.util.List;
+import java.util.Optional;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.VariableElement;
+
+public class DaoUpdateMethodGenerator extends DaoMethodGenerator {
+
+  private static final EnumSet<ReturnTypeKind> SUPPORTED_RETURN_TYPES =
+      EnumSet.of(VOID, FUTURE_OF_VOID, RESULT_SET, FUTURE_OF_ASYNC_RESULT_SET);
```
olim7t: I would also add `BOOLEAN` and ` FUTURE_OF_BOOLEAN` for `ifExists`.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateMethodGenerator.java
```diff
@@ -0,0 +1,189 @@
...
+      context
+          .getMessager()
+          .error(
+              methodElement,
+              "Invalid return type: %s methods must return either void or the entity class "
+                  + "(possibly wrapped in a CompletionStage/CompletableFuture)",
+              Update.class.getSimpleName());
+      return Optional.empty();
+    }
+    if (returnType.entityElement != null && !returnType.entityElement.equals(entityElement)) {
```
olim7t: This is not needed here: the types in `SUPPORTED_RETURN_TYPES` never contain an entity.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateMethodGenerator.java
```diff
@@ -0,0 +1,189 @@
...
+
+    // Generate the method:
+    String helperFieldName = enclosingClass.addEntityHelperField(ClassName.get(entityElement));
+    String statementName =
+        enclosingClass.addPreparedStatement(
+            methodElement,
+            (methodBuilder, requestName) ->
+                generatePrepareRequest(methodBuilder, requestName, helperFieldName));
+
+    MethodSpec.Builder insertBuilder = GeneratedCodePatterns.override(methodElement);
```
olim7t: Nit: this should be called `updateBuilder` for consistency. Although I've been wondering if it really needs to be named after the type of method, `methodBuilder` would be good too.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/mapper/MapperImplementationGenerator.java
```diff
@@ -105,7 +105,7 @@ protected String getFileName() {
                 .getMessager()
                 .error(
                     methodElement,
-                    "Unrecognized method signature: no implementation will be generated");
+                    "[MapperImplementationGenerator] Unrecognized method signature: no implementation will be generated");
```
olim7t: Same remark as in `DaoImplementationGenerator`.

olim7t (changes requested):
A few remarks for consistency with the existing annotations, but otherwise this looks good.

Please also update the documentation (following the style of what's already done):

* javadocs for `@Update` annotation
* javadocs for `EntityHelper.update`
* new manual page under `mapper/daos/update`

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperUpdateMethodGenerator.java
tomekl007: I was thinking about the 2nd approach but you are right - it breaks the API.
3 seems to be a clever trick to alleviate that problem. I've implemented it in this commit:
https://github.com/datastax/java-driver/pull/1235/commits/eb1c29f2338b4341a9759ca9fa9a641ea9936d62

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateMethodGenerator.java
```diff
@@ -0,0 +1,241 @@
...
+  private static final EnumSet<ReturnTypeKind> SUPPORTED_RETURN_TYPES =
+      EnumSet.of(VOID, FUTURE_OF_VOID, RESULT_SET, FUTURE_OF_ASYNC_RESULT_SET);
+  private static final Pattern USING_TIMESTAMP_VALUE_PATTER =
+      Pattern.compile("(?i)USING TIMESTAMP (\\d+$)");
+  private static final Pattern USING_TIMESTAMP_BIND_MARKER_PATTERN =
+      Pattern.compile("(?i)USING TIMESTAMP :(\\w+$)");
+  private static final Pattern USING_TTL_VALUE_PATTER = Pattern.compile("(?i)USING TTL (\\d+)$");
+  private static final Pattern USING_TTL_BIND_MARKER_PATTERN =
+      Pattern.compile("(?i)USING TTL :(\\w+)$");
+  // todo should we support ? bindMarker
```
tomekl007: I am not sure if we should support `USING TTL ?` syntax in the `customUsingClause`. 
I think we don't need it since we are binding parameters from the method arguments using named parameters

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGeneratorTest.java
tomekl007: Error handling implemented in https://github.com/datastax/java-driver/pull/1235/commits/07644279cf79e5e86905da796da7d75c14badb1b.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Update.java
tomekl007: Implemented in https://github.com/datastax/java-driver/pull/1235/commits/87c8ef779b814b68e760348a1d4f13c74ed34d48

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
tomekl007: I was trying to find a way to make it easier to debug and was planning to remove it before PR.
Thx for reminder. I agree that it should be black-box

integration-tests/src/test/java/com/datastax/oss/driver/mapper/UpdateEntityIT.java
tomekl007: Implemented in: https://github.com/datastax/java-driver/pull/1235/commits/6f2fa9167d05a42d62230fac5f3e16b30f96f75c

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateMethodGenerator.java
tomekl007: Yes, I was also considering that. Implemented in this commit:
https://github.com/datastax/java-driver/pull/1235/commits/95f96621ed26dab04a2bd7f7ce904be9ff9758a3
I've added tests and `BOOLEAN` as a supporter return type. 
What is not clear to me how it was handled automatically - I thought that I will need to add mapping logic (from results set to BOOLEAN) but it worked out of the box.
Could you tell me where is that mapping logic?

integration-tests/src/test/java/com/datastax/oss/driver/mapper/UpdateEntityIT.java
tomekl007: Don't we want to give our users the ability to specify WHERE IN: 
 ```UPDATE users
  SET state = 'TX'
  WHERE user_uuid
  IN (88b8fd18-b1ed-4e96-bf79-4280797cba80,
    06a8913c-c0d6-477c-937d-6c1b69a95d43,
    bc108776-7cb5-477f-917d-869c12dfffa8);
```
?

**Comment by tomekl007** [comment]
> A few remarks for consistency with the existing annotations, but otherwise this looks good.
> 
> Please also update the documentation (following the style of what's already done):
> 
> * javadocs for `@Update` annotation
> * javadocs for `EntityHelper.update`
> * new manual page under `mapper/daos/update`

Added documentation: https://github.com/datastax/java-driver/pull/1235/commits/27d2c4bba00b3725e40c22ac31b5c18da7e0d9e8
I think the PR is ready for review

File: mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateMethodGenerator.java (hunk 6 above)
Comment (olim7t) [reply mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateMethodGenerator.java]: In `ReturnTypeKind.addExecuteStatement`. The generated code relies on `executeXxx` helper methods defined in `DaoBase`.

Comment (tomekl007) [reply mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateMethodGenerator.java]: OK, I see:
```protected boolean executeAndMapWasAppliedToBoolean(Statement<?> statement) {
    ResultSet rs = execute(statement);
    return rs.wasApplied();
  }
```
thx

integration-tests/src/test/java/com/datastax/oss/driver/mapper/UpdateEntityIT.java
olim7t: Yeah actually you're right, we should preserve that ability.

So if the custom where clause is empty, it's an update by primary key and the mapper generates the where clause.
If it's not empty, then we concatenate it as-is at the end of our request; the primary key components of the entity are ignored.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateMethodGenerator.java
olim7t: As we discussed offline, let's support named parameters only.

However this makes me realize something else: unlike `Insert`, we can't simply append the provided string at the end of the query (because USING is not at the end for UPDATE, which is quite an unfortunate choice by the CQL language designers tbh).

Hence the string parsing you have to do here; it's completely justified but I find it a bit awkward. So let's revisit the annotation syntax to have the user provide the values directly instead:
```
@Update(ttl = "...", timestamp = "...")
```
Both attributes are strings that default to empty.
* if the string is empty we don't generate the corresponding clause
* if it starts with `:` it's a bind marker name
* otherwise we try to parse it as an int (for ttl) or long (for timestamp). If parsing fails we issue a compiler error.

I'm creating a ticket to implement a similar change for `@Insert` ([JAVA-2257](https://datastax-oss.atlassian.net/browse/JAVA-2257)).

olim7t: One more thing 😃 
With 2257, concatenating won't work anymore for `@Insert.customWhereClause` (because USING comes after). We'll have to use `whereRaw()` (like you initially did on this PR), and therefore it's simpler if we don't require the `WHERE` keyword in the provided string.
Same for `customIfClause` (no `IF` keyword in the string, use `ifRaw` to concatenate).

Again I'm creating a ticket to revisit existing annotations ([JAVA-2258](https://datastax-oss.atlassian.net/browse/JAVA-2258)), do it the new way directly in this PR.

olim7t: I fixed 2257 on the base branch. You should be able to reuse the two protected methods in `DaoMethodGenerator`.

tomekl007: I see one potential problem with `ifRaw`  -> I think that there could be a problem with creating the enclosed ifs:
So the following clause: "IF a = a AND b =b" - previously we would just append code. With `ifRaw()` you will need to create two `ifRaw()` statements (I didn't test that yet)

Concerning  `@Update(ttl = "...", timestamp = "...")` - I agree, I was thinking about that as well. It is more explicit and safe that just string value

olim7t: `ifRaw` appends whatever you pass as-is, so if you put the `AND` keyword in your string it should work fine.
```java
@Update(customIfClause = "a = :a AND b = :b"
```

olim7t:
Nit: when you use the @ sign in commit messages / PR comments, escape it so that GitHub doesn't interpret it as a mention:
```
adapt error message to `@update` annotation
```
I'm not sure about commit messages, but I know that for PR comments it notifies the user if it exists.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateMethodGeneratorTest.java
```diff
@@ -0,0 +1,92 @@
...
+    assertThat(builder.build().code).isEqualTo(CodeBlock.of(expected));
+  }
+
+  @DataProvider
+  public static Object[][] usingTimestampProvider() {
+    return new Object[][] {
+      {"1", ".usingTimestamp(1)"},
+      {
+        ":ts",
+        ".usingTimestamp(com.datastax.oss.driver.api.querybuilder.QueryBuilder.bindMarker(\"ts\"))"
```
olim7t: Nit: you could create code blocks directly here instead of doing the conversion in the test method.
```suggestion
        CodeBlock.of(".usingTimestamp($T.bindMarker($S))", QueryBuilder.class, "ts")
```
This is a bit easier to read.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateGeneratorTest.java
```diff
@@ -0,0 +1,85 @@
...
+import com.squareup.javapoet.TypeName;
+import com.tngtech.java.junit.dataprovider.DataProvider;
+import com.tngtech.java.junit.dataprovider.DataProviderRunner;
+import com.tngtech.java.junit.dataprovider.UseDataProvider;
+import javax.lang.model.element.Modifier;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+@RunWith(DataProviderRunner.class)
+public class DaoUpdateGeneratorTest extends DaoMethodGeneratorTest {
```
olim7t, line 35: Could you merge this with `DaoUpdateMethodGeneratorTest`? I find it confusing to have two classes with almost the same name that test the same target class (also all the other classes that test compiler outputs are called `XxxMethodGeneratorTest`, except this one).

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateMethodGenerator.java
```diff
@@ -0,0 +1,196 @@
...
+    }
+    methodBuilder.addStatement(
+        "$T boundStatementBuilder = $L.boundStatementBuilder()",
+        BoundStatementBuilder.class,
+        statementName);
+
+    List<? extends VariableElement> parameters = methodElement.getParameters();
+    String entityParameterName = parameters.get(0).getSimpleName().toString();
+    methodBuilder.addStatement(
+        "$L.set($L, boundStatementBuilder)", helperFieldName, entityParameterName);
```
olim7t: There is a problem here: if the annotation has a custom WHERE clause, we can't set the whole entity at once because there might be no placeholders for PK components.

You can observe it by changing the DAO in the integration test like this:
```java
    @Update(customWhereClause = "id IN (:id1, :id2)")
    void updateWhereIdIn(Product product, UUID id1, UUID id2);
```
(currently it uses `id IN (:id, :id2)`, which masks the issue)

So you need to add a conditional here:
* if `customWhereClause` is empty: OK, do what you are doing right now
* else: you need to bind only the *non-PK* properties of the entity. You can copy what the Delete code does [here](https://github.com/datastax/java-driver/blob/java2078/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java#L171), it's almost the same except that you need to traverse `getRegularColumns` instead of `getPrimaryKey`.

olim7t (changes requested):
I detected a bug that I had missed on the previous review. After that it should be good to go.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateMethodGenerator.java
tomekl007: I think that this situation is not possible. Please take a look at a test in this commit:
https://github.com/datastax/java-driver/pull/1235/commits/74251c9ea3fc7d02fe45be6eccfa984b9c36c34a
1. I created the table with PK and Clustering: https://github.com/datastax/java-driver/pull/1235/commits/74251c9ea3fc7d02fe45be6eccfa984b9c36c34a#diff-8337f2ae926b239e7dcfbcb474e854ffR36
2. I've created a DAO that has no PK nor clustering: 
https://github.com/datastax/java-driver/pull/1235/commits/74251c9ea3fc7d02fe45be6eccfa984b9c36c34a#diff-8337f2ae926b239e7dcfbcb474e854ffR124
3. Then there is an update with partialDao: 
https://github.com/datastax/java-driver/pull/1235/commits/74251c9ea3fc7d02fe45be6eccfa984b9c36c34a#diff-7ab34dc135c21c80962589033fdf5217R484
Please note the `customWhereClause`:  `@Update(customWhereClause = "id IN (:id, :id2) AND clustering = 1")`

If you will remove the `AND clustering = 1` part you will get the server side fail:
`com.datastax.oss.driver.api.core.servererrors.InvalidQueryException: Some clustering keys are missing: clustering`
So it means that the user cannot create a where clause that doesn't have all PKs(including Clustering Keys) in it. So it means that custom where clause must provide a clause for all PKs.

The update is setting regular columns only: https://github.com/datastax/java-driver/pull/1235/files#diff-b2adc444056441bbb59c5f7f513ce380R70 already so there should be no problems (we are not updating PKs(and clustering) in any case).

Please let me know if I am missing something.

olim7t: > id IN (:id, :id2) 

This only works because you named your placeholder the same as the column. Even if all PK fields must be present, you can't be sure that the user will use the correct names (as opposed to generated clauses where the processor controls the names).

I pushed the example from my previous comment. The query is valid, but if you run the test you will see that it fails because the processor is looking for a placeholder that isn't here:
```
java.lang.IllegalArgumentException: id is not a variable in this bound statement
```

tomekl007: Ok, thx I didn't catch that from the previous comment (I thought that this is a matter of DTO).
It is fixed now

olim7t (approved):

olim7t:
Good job 👍

PR #1236, opened 2019-05-16
JAVA-2110: Add annotation to control the naming strategy

olim7t:


manual/mapper/entities/README.md
```diff
@@ -30,6 +30,88 @@ for:
...
+convention used above, the mapper expects the following schema:
+
+```
+CREATE TABLE "PRODUCT"("PRODUCT_ID" int primary key ...)
+```
+
+For the list of all available conventions, look at the enum constants in [NamingConvention].
+
+If you don't annotate your class with [@NamingStrategy], the mapper defaults to the
+`SNAKE_CASE_INSENSITIVE` convention.
```
olim7t, line 70: This feels like the most sane default to me.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/SingleFileCodeGenerator.java
```diff
@@ -34,14 +35,16 @@ protected SingleFileCodeGenerator(ProcessorContext context) {
 
   @Override
   public void generate() {
-    context.getFiler().write(getFileName(), getContents());
+    ClassName typeName = getPrincipalTypeName();
+    String fileName = typeName.simpleName();
+    if (!typeName.packageName().isEmpty()) {
+      fileName = typeName.packageName() + '.' + fileName;
+    }
+    context.getFiler().write(fileName, getContents());
```
olim7t, line 43: Those changes are unrelated to this ticket.
I just noticed that the mapper crashed for types in the root package.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/entity/naming/NamingConvention.java
```diff
@@ -0,0 +1,82 @@
...
+import com.datastax.oss.driver.api.mapper.annotations.Entity;
+import com.datastax.oss.driver.api.mapper.annotations.NamingStrategy;
+
+/**
+ * A built-in convention to infer CQL column names from the names used in an {@link
+ * Entity}-annotated class.
+ *
+ * @see NamingStrategy
+ */
+public enum NamingConvention {
```
olim7t, line 27: Unlike mapper 3 there is no notion of a "Java convention", it expects camel-case, period. I don't think that complexity was worth it, it's very unlikely that someone would name their methods `getProduct_Id`.

Also I'm open to suggestions for other conventions, but I think we should be conservative (even some of those are already a bit exotic IMO).

tomekl007:
Travis is failing with exception:
```Failed tests: 
  EntityAnnotationTest.should_work_on_nested_class:59 Compilation produced the following errors:
test/Foo.java:7: error: Unexpected error while writing generated code: java.lang.NullPointerException
  public static class Bar {```

**Comment by olim7t** [comment]
Fixed the failing test, thanks.

File: mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/BuiltInNameConversions.java [hunk 4]
```diff
@@ -0,0 +1,55 @@
...
+    switch (convention) {
+      case CASE_INSENSITIVE:
+        return javaName;
+      case EXACT_CASE:
+        return Strings.doubleQuote(javaName);
+      case LOWER_CAMEL_CASE:
+        // Piggy-back on Guava's CaseFormat. Note that we indicate that the input is upper-camel
+        // when in reality it can be lower-camel for a property name, but CaseFormat is lenient and
+        // handles that correctly.
+        return Strings.doubleQuote(CaseFormat.UPPER_CAMEL.to(CaseFormat.LOWER_CAMEL, javaName));
```
tomekl007, line 41: How do we know the format from which we are doing the conversion?
I understand that this part: `to(CaseFormat.LOWER_CAMEL,` can be assumed from this `case LOWER_CAMEL_CASE:` but from the `CaseFormat.UPPER_CAMEL` I am not sure how can we assume that (same for other similar cases)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/CqlNameGenerator.java
```diff
@@ -0,0 +1,73 @@
...
+  public CqlNameGenerator(NamingConvention namingConvention) {
+    this(namingConvention, null);
+  }
+
+  public CqlNameGenerator(TypeMirror converterClassMirror) {
+    this(null, converterClassMirror);
+  }
+
+  private CqlNameGenerator(NamingConvention namingConvention, TypeMirror converterClassMirror) {
+    Preconditions.checkArgument(namingConvention == null ^ converterClassMirror == null);
```
tomekl007: If this check fails will it be converted to easy to read the message?
I am assuming that not every client will know the meaning of using `^` in this part of the code and if this will propagate to exception message he maybe confused

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -230,4 +264,87 @@ private int getClusteringColumnIndex(
...
+        break;
+      }
+    }
+    assert annotationMirror != null; // We've checked that in the caller already
+
+    for (Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> entry :
+        annotationMirror.getElementValues().entrySet()) {
+      if (entry.getKey().getSimpleName().contentEquals("customConverterClass")) {
+        @SuppressWarnings("unchecked")
+        List<? extends AnnotationValue> values = (List) entry.getValue().getValue();
```
tomekl007, line 340: Isn't enough to cast to Collection?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -230,4 +264,87 @@ private int getClusteringColumnIndex(
...
+      }
+      return new CqlNameGenerator(conventions[0]);
+    } else {
+      if (customConverterClasses.length > 1) {
+        context
+            .getMessager()
+            .warn(
+                classElement,
+                "Too many custom converters: %s must have at most one "
+                    + "'customConverterClass' argument (will use the first one: %s)",
```
tomekl007, line 316: using the first one can be a bit problematic - is it repeatable?
In other words are we sure that list of customConverterClasses is sorted? If not the client will get non-deterministic ordering.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -117,15 +117,24 @@ public static void bindParameters(
...
+        setValue(
+            parameterName,
+            type,
+            CodeBlock.of("$L", parameterName),
+            "boundStatementBuilder",
+            methodBuilder,
+            enclosingClass);
+      } else {
+        setValue(
+            bindMarkerNames.get(i),
```
tomekl007, line 131: why does that change? I think we were supporting bind marker names previously, aren't we?

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityNamingStrategyTest.java
```diff
@@ -0,0 +1,131 @@
...
+            .addAnnotation(
+                AnnotationSpec.builder(NamingStrategy.class)
+                    .addMember("convention", "$T.CASE_INSENSITIVE", NamingConvention.class)
+                    .addMember("customConverterClass", "$T.class", CUSTOM_CONVERTER_CLASS_NAME)
+                    .build())
+            .build());
+  }
+
+  @Test
+  public void should_fail_if_neither_convention_nor_converter_specified() {
```
tomekl007, line 85: I don't understand why in this case we should expect an error. Shouldn't we fallback to default naming strategy if nothing was specified?

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/MapperContext.java
```diff
@@ -65,4 +75,30 @@ public CqlIdentifier getKeyspaceId() {
...
+   * Returns an instance of the given converter class.
+   *
+   * <p>The results of this method are cached at the mapper level. If no instance of this class
+   * exists yet for this mapper, a new instance is built by looking for a public no-arg constructor.
+   */
+  public NameConverter getNameConverter(Class<? extends NameConverter> converterClass) {
+    return nameConverterCache.computeIfAbsent(converterClass, MapperContext::buildNameConverter);
+  }
+
+  private static NameConverter buildNameConverter(Class<? extends NameConverter> converterClass) {
```
tomekl007, line 89: I think that code could be refactored to be more generic - `buildInstanceUsingNoArgsConstructor` and the reused in multiple places (I think that we are constructing newInstances() in this way in couple places in the code and we could replace that code we invocation of this method)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/BuiltInNameConversions.java
olim7t: Guava's `CaseFormat` is pretty lenient, `LOWER_CAMEL` accepts both lower-camel and upper-camel input.
However `to()` does nothing if its argument is the same as the target object, that's why I switched it here to make sure it's always different.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/CqlNameGenerator.java
olim7t: This is not meant to catch user errors, the annotation parameters are already validated before calling this.

This is more of a safeguard against programming errors /  documentation of an invariant of the class. Notice that with the current public constructors there is no way this check will fail.
In fact this should probably be an assertion instead, I'll change it.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: We use `get(int)` in the loop below.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: Specifying multiple classes is already a user error, hopefully the user will notice the warning and fix the issue.

Arguably we could issue an error instead of a warning for this kind of mistake.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityNamingStrategyTest.java
olim7t: We fall back to the default when there is no annotation. If the annotation is present, it must have one or the other.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
olim7t: Nothing changes, it's just that now in some cases the bind marker names can be code blocks (when we retrieve them from the entity definition), and in others they're plain strings (when we use the Java parameter names as-is).

olim7t: For example, in this case:
```java
@Delete(entityClass = Product.class)
void deleteById(UUID productId);
```
`productId` could be called anything really, we don't care about the name because by looking at the entity definition, we know the bind marker will be called `id`. So we just say "bind the first parameter (whatever its name is), to `id`".
This corresponds to the branch `bindMarkerNames != null` in `bindParameters`, and now these names are code blocks.

In this case:
```java
@Select(customWhereClause = "WHERE description LIKE :searchString")
PagingIterable<Product> findByDescription(String searchString);
```
We don't know at compile time what the name of the bind marker will be, because it's in the user-provided string. So we require that it matches exactly: "bind the parameter called `searchString` to a placeholder called `searchString`".

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/MapperContext.java
olim7t: I think that's the first place in the mapper where we use reflection. But it will be used again for [JAVA-2118](https://datastax-oss.atlassian.net/browse/JAVA-2118), I'll see if some reuse is possible.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
tomekl007: - Ok, isn't that first case of deleting a bit dangerous?
I mean that user can add a different name for that argument purposely (for example non-PK column) because he is new to Cassandra and doesn't know that deleting by non-pk is not possible. Then when issuing delete with that different argument he may get surprised that entity was not deleted - it may lead to weird problems (assuming that this non-PK key has the same type as inferred PK)

- We are doing type check, right? - so if the name of the argument is different/or the same as PK but the type is not matching PK what will happen?

olim7t: That's a bit far-fetched though, nowhere else in the mapper do we generate conditions on specific columns based on the name of a parameter.
Beyond that, I would say reading the docs and writing unit tests should prevent this kind of error.

If the types don't match, we generate an [error](https://github.com/datastax/java-driver/blob/java2078/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java#L121-L134).

integration-tests/src/test/java/com/datastax/oss/driver/mapper/NamingStrategyIT.java
```diff
@@ -0,0 +1,269 @@
...
+    CustomNamesEntityDao dao = mapper.customNamesEntityDao(sessionRule.keyspace());
+    CustomNamesEntity entity = new CustomNamesEntity(1);
+
+    dao.save(entity);
+    CustomNamesEntity retrievedEntity = dao.findById(1);
+    assertThat(retrievedEntity.getEntityId()).isEqualTo(1);
+  }
+
+  @Entity
+  public static class DefaultStrategyEntity {
```
tomekl007, line 127: nit: I think all public keywords can be removed

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityNamingStrategyTest.java
```diff
@@ -0,0 +1,131 @@
...
+        entityTemplate()
+            .addAnnotation(
+                AnnotationSpec.builder(NamingStrategy.class)
+                    .addMember(
+                        "customConverterClass",
+                        "{ $1T.class, $1T.class }",
+                        CUSTOM_CONVERTER_CLASS_NAME)
+                    .build())
+            .build());
+  }
```
tomekl007, line 130: I think we should add a test when `@CqlName` is specified and `@NamingStrategy` and assert that warning was generated

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/NamingStrategy.java
```diff
@@ -0,0 +1,76 @@
...
+   * hard-codes the converted strings. In other words, this is slightly more efficient than a custom
+   * converter because no conversion happens at runtime.
+   *
+   * <p>This is mutually exclusive with {@link #customConverterClass()}.
+   *
+   * <p>Note that, for technical reasons, this is an array, but only one element is expected. If you
+   * specify more than one element, the mapper processor will generate a compile-time warning, and
+   * proceed with the first one.
+   */
+  NamingConvention[] convention() default {};
```
tomekl007, line 60: Maybe instead of using an array here (and therefore need for handling that array case), we could create `NamingConvetion.NULL` - as a maker for the default? (or `NamingConvetion.DEFAULT`)
Then this method will be 
`NamingConvention convention() defualt NamingConvetion.NULL`
We will simplify error handling in this way 
Or set it to:
`NamingConvention[] convention() default NamingConvention.SNAKE_CASE_INSENSITIVE;` if `SNAKE_CASE_INSENSITIVE` is a default

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/SingleFileCodeGenerator.java
tomekl007: Maybe we can add an integration test for a class that is in a root package? Using `MapperProcessorTest`

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/BuiltInNameConversions.java
```diff
@@ -0,0 +1,55 @@
...
+      case UPPER_CAMEL_CASE:
+        return Strings.doubleQuote(CaseFormat.LOWER_CAMEL.to(CaseFormat.UPPER_CAMEL, javaName));
+      case SNAKE_CASE_INSENSITIVE:
+        return CaseFormat.LOWER_CAMEL.to(CaseFormat.LOWER_UNDERSCORE, javaName);
+      case UPPER_SNAKE_CASE:
+        return Strings.doubleQuote(
+            CaseFormat.LOWER_CAMEL.to(CaseFormat.UPPER_UNDERSCORE, javaName));
+      case UPPER_CASE:
```
tomekl007, line 49: If we have UPPER_CASE, shouldn't we have `LOWER_CASE` as well?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -230,4 +264,87 @@ private int getClusteringColumnIndex(
...
+    if (conventions.length > 0 && customConverterClasses.length > 0) {
+      context
+          .getMessager()
+          .error(
+              classElement,
+              "Invalid annotation configuration: %s must have either a 'convention' "
+                  + "or 'customConverterClass' argument, but not both",
+              NamingStrategy.class.getSimpleName());
+      // Return a generator anyway, so that the processor doesn't crash downstream
+      return new CqlNameGenerator(conventions[0]);
```
tomekl007, line 287: Did we document that preference? (choosing convention over customConverterClass?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -230,4 +264,87 @@ private int getClusteringColumnIndex(
...
+      }
+      return new CqlNameGenerator(customConverterClasses[0]);
+    }
+  }
+
+  private TypeMirror[] readCustomConverterClasses(TypeElement classElement) {
+    // customConverterClass references a class that might not be compiled yet, so we can't read it
+    // directly, we need to go through mirrors.
+    AnnotationMirror annotationMirror = null;
+    for (AnnotationMirror candidate : classElement.getAnnotationMirrors()) {
```
tomekl007, line 328: This is executed in a context of the specific class, right?
So we are getting only custom converters for a specific entity - not all converters for all classes?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGenerator.java
```diff
@@ -119,6 +113,10 @@ public String addEntityHelperField(ClassName childEntityName) {
     GeneratedCodePatterns.addFinalFieldAndConstructorArgument(
         ClassName.get(MapperContext.class), "context", classContents, constructorContents);
 
+    // TODO create name converter
```
tomekl007: Is that TODO still relevant? 
I think that we have name converter for entity name implemented, right?

tolbertam:
Just a general comment, the 3.x mapper had a way to set the default naming strategy (among other options) through [`MappingConfiguration`](https://docs.datastax.com/en/developer/java-driver/3.7/manual/object_mapper/using/#mapping-configuration).  

I see on the jira ticket you mention:

> Alternatively, we could allow the annotation on entities only. If you want to inherit a common configuration, make all your entities inherit from a common class/interface, and put the annotation there.

I agree that having a top-level interface that entities can extend from to inherit common configuration via Annotations is a good way to go.  I always like when there's one way to configure things as it keeps things simple.  Looking at `DefaultEntityFactory.getDefinition` i see we have a TODO to support inheriting annotations which will be done as part of [JAVA-2111](https://datastax-oss.atlassian.net/browse/JAVA-2111) :+1:

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: We only return something here so that the rest of the processor code doesn't crash (it's simpler than trying to abort the whole generation).

The user won't be able to observe this behavior, their build is already failing with an error (from the previous line).

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: Yes, we are reading the converter class declared by the annotation for this entity (note that after [JAVA-2111](https://datastax-oss.atlassian.net/browse/JAVA-2111), the annotation might be inherited).

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/BuiltInNameConversions.java
olim7t: Lower-case is the same as case-insensitive.

In other words, `SELECT * FROM "foo"` matches exactly the same thing as `SELECT * FROM Foo`.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/NamingStrategy.java
olim7t: This would work pretty well with an enum, but when the element type is a `Class` you have to create a dummy `NullClass` which I find really ugly.

So I prefer the array approach, and for consistency it's better to do it the same way everywhere.

For the user it's almost transparent. We handle a multi-valued array for completeness' sake, but really no one is going to do that, it makes no sense.
```java
@NamingStrategy(convention = { LOWER_CAMEL_CASE, UPPER_CAMEL_CASE })
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGenerator.java
olim7t: Good catch, it is indeed outdated.

I ended up inlining the retrieval of the converter everywhere instead of storing it in a field. For example in `NamingStrategyIT_NameConverterEntityHelper__MapperGenerated`:
```java
target = target.setInt(context.getNameConverter(NamingStrategyIT.TestNameConverter.class).toCassandraName("entityId"), entity.getEntityId());
```
This might seem a bit extreme, but `getNameConverter` is cached so there shouldn't be much of a performance hit.

Also TBH I don't want to spend to much time optimizing that case, because it should be the exception rather than the norm. If your naming strategy *for a single entity* can't be covered by a default convention (with maybe a few `@CqlName` here and there), it's probably a bit whacky.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/NamingStrategyIT.java
olim7t: Yes, it would work for these  tests because everything lives in the same package.

But in real life mapper interfaces/entities will most likely be public, so it doesn't hurt to be a bit more realistic here.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityNamingStrategyTest.java
olim7t: Mmm there is no warning when a column name is overridden.

Are you referring to [this comment](https://github.com/datastax/java-driver/pull/1236#discussion_r286427436)? If so, it generates an error that is covered at line 69.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: But to answer your original question, yes, [the order is deterministic](https://docs.oracle.com/javase/7/docs/api/javax/lang/model/element/AnnotationValue.html):
> representing the elements, in declared order, if the value is an array

PR #1238, opened 2019-05-17
JAVA-2260: Handle empty collections in PreparedStatement.bind(...)

olim7t:
Motivation:

`preparedStatement.bind(new HashSet<>())` fails with a
CodecNotFoundException.

This is because the codec registry defaults to Set<Boolean> when it
tries to infer the type of the value (this works fine for lookups by
Java type only, because all empty collections are encoded in the same
way, so any collection codec will do -- but here we happen to also know
the CQL type, and it doesn't match).

Modifications:

Special-case empty collections at the beginning of codecFor(CqlType,
value): simply look up a collection codec for an arbitrary element type.

Result:

We return a codec that can encode the provided value.
Note that it doesn't accept the CQL type, nor the declared Java type of
the value, but the contract of the method doesn't say that it should.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
@@ -31,8 +31,10 @@ import com.datastax.oss.driver.api.core.type.codec.registry.CodecRegistry;
 import com.datastax.oss.driver.api.core.type.reflect.GenericType;
 import com.datastax.oss.driver.shaded.guava.common.base.Preconditions;
 import com.datastax.oss.driver.shaded.guava.common.reflect.TypeToken;
+import com.datastax.oss.protocol.internal.ProtocolConstants;
 import com.datastax.oss.protocol.internal.util.IntMap;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.nio.ByteBuffer;
@@ -70,7 +72,9 @@ public abstract class CachingCodecRegistry implements CodecRegistry {
   private final IntMap<TypeCodec> primitiveCodecsByCode;
 
   protected CachingCodecRegistry(
-      String logPrefix, TypeCodec<?>[] primitiveCodecs, TypeCodec<?>[] userCodecs) {
+      @NonNull String logPrefix,
+      @NonNull TypeCodec<?>[] primitiveCodecs,
+      @NonNull TypeCodec<?>[] userCodecs) {
     this.logPrefix = logPrefix;
     this.primitiveCodecs = primitiveCodecs;
     this.userCodecs = userCodecs;
@@ -85,7 +89,7 @@ public abstract class CachingCodecRegistry implements CodecRegistry {
    * future calls).
    */
   protected abstract TypeCodec<?> getCachedCodec(
-      DataType cqlType, GenericType<?> javaType, boolean isJavaCovariant);
+      @Nullable DataType cqlType, @Nullable GenericType<?> javaType, boolean isJavaCovariant);
 
   @NonNull
   @Override
@@ -95,8 +99,11 @@ public abstract class CachingCodecRegistry implements CodecRegistry {
   }
 
   // Not exposed publicly, (isJavaCovariant=true) is only used for internal recursion
+  @NonNull
   protected <JavaTypeT> TypeCodec<JavaTypeT> codecFor(
-      DataType cqlType, GenericType<JavaTypeT> javaType, boolean isJavaCovariant) {
+      @NonNull DataType cqlType,
+      @NonNull GenericType<JavaTypeT> javaType,
+      boolean isJavaCovariant) {
     LOG.trace("[{}] Looking up codec for {} <-> {}", logPrefix, cqlType, javaType);
     TypeCodec<?> primitiveCodec = primitiveCodecsByCode.get(cqlType.getProtocolCode());
     if (primitiveCodec != null && matches(primitiveCodec, javaType, isJavaCovariant)) {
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -413,4 +423,14 @@ protected boolean matches(TypeCodec<?> codec, GenericType<?> javaType, boolean i
...
   }
+
+  // These are mock types that are used as placeholders when we try to find a codec for an empty
+  // Java collection instance. All empty collections are serialized in the same way, so any element
+  // type will do:
+  private static final GenericType<List<Boolean>> JAVA_TYPE_FOR_EMPTY_LISTS =
+      GenericType.listOf(Boolean.class);
+  private static final GenericType<Set<Boolean>> JAVA_TYPE_FOR_EMPTY_SETS =
+      GenericType.setOf(Boolean.class);
+  private static final GenericType<Map<Boolean, Boolean>> JAVA_TYPE_FOR_EMPTY_MAPS =
```
tomekl007, line 509: If the type is not important why we don't pass Object?
Or maybe even better we could create marker `class Any` and use it here for a readability

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -157,6 +157,16 @@ protected CachingCodecRegistry(
     Preconditions.checkNotNull(value);
     LOG.trace("[{}] Looking up codec for CQL type {} and object {}", logPrefix, cqlType, value);
 
+    // Special case for empty collections: inspectType won't work because we have nothing to infer
```
adutra: Two remarks:

1) I don't think this is going to solve the problem for all cases. For example, if you change your unit test to this:

```
    ListType cqlType = DataTypes.listOf(DataTypes.listOf(DataTypes.INT));
    List<List<Integer>> value = Collections.singletonList(Collections.emptyList());
```
The test will fail because the enclosing collection is not empty.
IMO the correct way to handle this is inside `inspectType` and `getCachedCodec`. You might need to give up on generic codecs for empty collections and actually create a codec that truly accepts the target CQL type.

2) I also think that this added code snippet is appearing too soon, and in particular it is bypassing all user-registered codecs, but users might have registered special codecs for collections, in which case we should use them if possible.

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistryTest.java
```diff
@@ -265,6 +266,26 @@ public void should_create_list_codec_for_cql_type_and_java_value() {
         .lookup(DataTypes.listOf(DataTypes.INT), GenericType.listOf(GenericType.INTEGER), true);
   }
 
+  @Test
+  public void should_create_list_codec_for_cql_type_and_empty_java_value() {
+    ListType cqlType = DataTypes.listOf(DataTypes.listOf(DataTypes.INT));
+    List<List<Integer>> value = Collections.emptyList();
```
adutra: Like I said above the tests would be more meaningful if the enclosing collection wasn't empty but the enclosed one was.

adutra (changes requested):
I went ahead and tried to fix the issues I pointed out here in a branch named `java2260-adu`. Waiting for CI now. Let me know what you think.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
olim7t: Because the registry will still look for a codec for the element type, and there is no `Object` codec. We pick one of the primitive CQL types.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
olim7t: Good catch.
I like with your solution of inferring the Java type from the CQL type if available, I'm cherry-picking your commits on this PR.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -290,6 +291,71 @@ protected boolean matches(TypeCodec<?> codec, GenericType<?> javaType, boolean i
...
+      DataType elementType = ((SetType) cqlType).getElementType();
+      return GenericType.setOf(inferJavaTypeFromCqlType(elementType));
+    } else if (cqlType instanceof MapType) {
+      DataType keyType = ((MapType) cqlType).getKeyType();
+      DataType valueType = ((MapType) cqlType).getValueType();
+      return GenericType.mapOf(
+          inferJavaTypeFromCqlType(keyType), inferJavaTypeFromCqlType(valueType));
+    }
+    switch (cqlType.getProtocolCode()) {
+      case ProtocolConstants.DataType.CUSTOM:
```
olim7t, line 317: IDEA reports a few duplicate branches here, we can factor the cases that return the same result.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -279,9 +291,74 @@ protected boolean matches(TypeCodec<?> codec, GenericType<?> javaType, boolean i
     }
   }
 
+  protected GenericType<?> inferJavaTypeFromCqlType(@NonNull DataType cqlType) {
+    if (cqlType instanceof ListType) {
+      DataType elementType = ((ListType) cqlType).getElementType();
+      return GenericType.listOf(inferJavaTypeFromCqlType(elementType));
```
tomekl007, line 306: don't we want to provide some upper bound for that recursion?
Is it possible that the user will create so many enclosing collections that our code will throw `StackOverflow`?

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
adutra: Yeah I noticed them too but couldn't decide what is best: to factor together or to leave the case instructions in the order in which they are declared in `ProtocolConstants.DataType`. Your call.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
adutra: I think it would blow up server-side first, because you would need to create a collection type with thousands of nested types for that to happen, as in `list<list<list<list<...`.

adutra (approved):

PR #1239, opened 2019-05-22
JAVA-2255: Add annotation to indicate a property is transient

tolbertam:
For [JAVA-2255](https://datastax-oss.atlassian.net/browse/JAVA-2255)

---

Motivation:

It is sometimes desirable for properties of an entity to not be
considered by an object mapper.

Modifications:

Introduced a `@Transient` annotation that when applied to a field or
getter, causes the mapper to not include this field for generated
queries.

In addition, if a field includes the 'transient' keyword, it is also not
considered.

Result:

properties including a `@Transient` annotation or having a field with
transient keyword will not be considered by the mapper.

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/mapper/TransientIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/mapper/TransientIT.java
@@ -1,0 +1,288 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.mapper;
+
+import static com.datastax.oss.driver.assertions.Assertions.assertThat;
+
+import com.datastax.oss.driver.api.core.CqlIdentifier;
+import com.datastax.oss.driver.api.core.CqlSession;
+import com.datastax.oss.driver.api.core.cql.SimpleStatement;
+import com.datastax.oss.driver.api.mapper.annotations.Dao;
+import com.datastax.oss.driver.api.mapper.annotations.DaoFactory;
+import com.datastax.oss.driver.api.mapper.annotations.DaoKeyspace;
+import com.datastax.oss.driver.api.mapper.annotations.DaoTable;
+import com.datastax.oss.driver.api.mapper.annotations.Entity;
+import com.datastax.oss.driver.api.mapper.annotations.Insert;
+import com.datastax.oss.driver.api.mapper.annotations.Mapper;
+import com.datastax.oss.driver.api.mapper.annotations.PartitionKey;
+import com.datastax.oss.driver.api.mapper.annotations.Select;
+import com.datastax.oss.driver.api.mapper.annotations.Transient;
+import com.datastax.oss.driver.api.testinfra.ccm.CcmRule;
+import com.datastax.oss.driver.api.testinfra.session.SessionRule;
+import com.datastax.oss.driver.categories.ParallelizableTests;
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+
+@Category(ParallelizableTests.class)
+public class TransientIT {
+
+  private static CcmRule ccm = CcmRule.getInstance();
+
+  private static SessionRule<CqlSession> sessionRule = SessionRule.builder(ccm).build();
+
+  @ClassRule public static TestRule chain = RuleChain.outerRule(ccm).around(sessionRule);
+
+  private static TestMapper mapper;
+
+  private static final AtomicInteger keyProvider = new AtomicInteger(0);
+
+  @BeforeClass
+  public static void setup() {
+    CqlSession session = sessionRule.session();
+
+    session.execute(
+        SimpleStatement.builder("CREATE TABLE entity(id int primary key, v int)")
+            .setExecutionProfile(sessionRule.slowProfile())
+            .build());
+
+    mapper = new TransientIT_TestMapperBuilder(session).build();
+  }
+
+  @Test
+  public void should_ignore_field_with_transient_annotated_field() {
+    EntityWithTransientAnnotatedFieldDao dao =
+        mapper.entityWithTransientAnnotatedFieldDao(
+            sessionRule.keyspace(), CqlIdentifier.fromCql("entity"));
+
+    int key = keyProvider.incrementAndGet();
+    EntityWithTransientAnnotatedField entity = new EntityWithTransientAnnotatedField(key, 1, 7);
+    dao.save(entity);
+
+    EntityWithTransientAnnotatedField retrievedEntity = dao.findById(key);
+    assertThat(retrievedEntity.getId()).isEqualTo(key);
+    assertThat(retrievedEntity.getV()).isEqualTo(1);
+    // column should not have been set since field was @Transient-annotated
+    assertThat(retrievedEntity.getNotAColumn()).isNull();
+  }
+
+  @Test
+  public void should_ignore_field_with_transient_annotated_getter() {
+    EntityWithTransientAnnotatedGetterDao dao =
+        mapper.entityWithTransientAnnotatedGetterDao(
+            sessionRule.keyspace(), CqlIdentifier.fromCql("entity"));
+
+    int key = keyProvider.incrementAndGet();
+    EntityWithTransientAnnotatedGetter entity = new EntityWithTransientAnnotatedGetter(key, 1, 7);
+    dao.save(entity);
+
+    EntityWithTransientAnnotatedGetter retrievedEntity = dao.findById(key);
+    assertThat(retrievedEntity.getId()).isEqualTo(key);
+    assertThat(retrievedEntity.getV()).isEqualTo(1);
+    // column should not have been set since field was @Transient-annotated
+    assertThat(retrievedEntity.getNotAColumn()).isNull();
+  }
+
+  @Test
+  public void should_ignore_field_with_transient_keyword() {
+    EntityWithTransientKeywordDao dao =
+        mapper.entityWithTransientKeywordDao(
+            sessionRule.keyspace(), CqlIdentifier.fromCql("entity"));
+
+    int key = keyProvider.incrementAndGet();
+    EntityWithTransientKeyword entity = new EntityWithTransientKeyword(key, 1, 7);
+    dao.save(entity);
+
+    EntityWithTransientKeyword retrievedEntity = dao.findById(key);
+    assertThat(retrievedEntity.getId()).isEqualTo(key);
+    assertThat(retrievedEntity.getV()).isEqualTo(1);
+    // column should not have been set since field was @Transient-annotated
+    assertThat(retrievedEntity.getNotAColumn()).isNull();
+  }
+
+  @Entity
+  public static class EntityWithTransientAnnotatedField {
+
+    @PartitionKey private int id;
+
+    private int v;
+
+    @Transient private Integer notAColumn;
+
+    public EntityWithTransientAnnotatedField() {}
+
+    public EntityWithTransientAnnotatedField(int id, int v, Integer notAColumn) {
+      this.id = id;
+      this.v = v;
+      this.notAColumn = notAColumn;
+    }
+
+    public int getId() {
+      return id;
+    }
+
+    public void setId(int id) {
+      this.id = id;
+    }
+
+    public int getV() {
+      return v;
+    }
+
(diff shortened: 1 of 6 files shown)


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -347,4 +352,17 @@ private CqlNameGenerator buildCqlNameGenerator(TypeElement classElement) {
     }
     return new TypeMirror[0];
   }
+
+  private boolean isTransient(ExecutableElement getMethod, VariableElement field) {
+    Transient getterAnnotation = getMethod.getAnnotation(Transient.class);
```
tolbertam: Of note: we don't check that other annotations, such as `PartitionKey` or `ClusteringKey` are present and emit a warning, is that worth doing?

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Transient.java
```diff
@@ -0,0 +1,40 @@
...
+ * </pre>
+ *
+ * This information is used by the mapper processor to exclude the property from generated queries.
+ *
+ * <p>Please note that {@code Transient} takes precedence over {@link PartitionKey} and {@link
+ * ClusteringColumn} annotations.
+ */
+@Target({ElementType.FIELD, ElementType.METHOD})
+@Retention(RetentionPolicy.CLASS)
+public @interface Transient {}
```
tolbertam, line 40: Some things that the 3.x mapper has that this (currently) doesn't:

1. There was a way of specifying a [`PropertyTransienceStrategy`](https://github.com/datastax/java-driver/blob/3.x/driver-mapping/src/main/java/com/datastax/driver/mapping/PropertyTransienceStrategy.java), which allowed one to choose whether properties need to be explicitly annotated or not to be included.  Since we don't have a `@Column` annotation, this didn't seem appropriate here.

2. You could provide a global configuration for specifying transient property names, i.e.: [`addTransientPropertyNames`](https://github.com/datastax/java-driver/blob/3.x/driver-mapping/src/main/java/com/datastax/driver/mapping/DefaultPropertyMapper.java#L179-L242).  I think a good solution for this might be to add an entity-scoped annotation named `@TransientProperties` that is akin to jackson-annotations' [`JsonIgnoreProperties`](https://fasterxml.github.io/jackson-annotations/javadoc/2.6/com/fasterxml/jackson/annotation/JsonIgnoreProperties.html), i.e.:

```java
@TransientProperties({"a", "b", "c", "d"})
@Entity
class MyEntity {}
```

what do you think?

integration-tests/src/test/java/com/datastax/oss/driver/mapper/TransientIT.java
```diff
@@ -0,0 +1,288 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+
+@Category(ParallelizableTests.class)
+public class TransientIT {
```
tolbertam, line 46: I'm only using `Transient` here because that was what we used in 3.x.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: Yes, if it's not too much work I think it would be worth it.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Transient.java
olim7t: 1. I guess the question is: do we need an `@Column` annotation? I'm not really convinced, mapping all by default sounds pretty good as long as there's a proper exclusion mechanism.
Also this can always be added later as long as it defaults to "map all", so no need to rush it.

2. That sound like a good idea, for those cases where you inherit from a class that you don't own, and which has properties that you don't want to map.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/TransientIT.java
```diff
@@ -0,0 +1,288 @@
...
+            .build());
+
+    mapper = new TransientIT_TestMapperBuilder(session).build();
+  }
+
+  @Test
+  public void should_ignore_field_with_transient_annotated_field() {
+    EntityWithTransientAnnotatedFieldDao dao =
+        mapper.entityWithTransientAnnotatedFieldDao(
+            sessionRule.keyspace(), CqlIdentifier.fromCql("entity"));
```
olim7t, line 74: Nice use of table parameterization 👍

olim7t (approved):
LGTM so far, will review again after you do the changes we discussed (improved warnings and entity-level exclusions).

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
tolbertam: :+1: will add that

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Transient.java
tolbertam: 1. Agreed that this isn't compelling enough to add a `@Column`.
2. :+1:, I'll add that.

tolbertam:
Pushed c580333 which adds an entity-level `@TransientProperties` annotation and additional validation around using other annotations on properties that are considered transient.

manual/mapper/entities/README.md
```diff
@@ -159,10 +159,21 @@ private int day;
 This information is used by some of the DAO method annotations; for example,
 [@Select](../daos/select/)'s default behavior is to generate a selection by primary key.
 
+#### Transient properties
```
tolbertam, line 162: Just realized I need to add doc for `@TransientProperties`, will do

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -347,4 +352,80 @@ private CqlNameGenerator buildCqlNameGenerator(TypeElement classElement) {
...
+      ExecutableElement getMethod,
+      VariableElement field,
+      int partitionKeyIndex,
+      int clusteringColumnIndex) {
+
+    // check if property is present in @TransientProperties
+    Class<?> otherAnnotationClass =
+        partitionKeyIndex >= 0
+            ? PartitionKey.class
+            : clusteringColumnIndex >= 0 ? ClusteringColumn.class : null;
```
olim7t: We should also check other mapper annotations: `CqlName`, `Computed` and maybe others in the future. I'm not sure if passing a value for each annotation will scale well, so maybe we could just re-read them on the field and getter.
```java
for (Class<?> annotationClass : ImmutableList.of(PartitionKey.class, ClusteringColumn.class, CqlName.class) {
  if (getMethod.getAnnotation(annotationClass) != null || field != null && field.getAnnotation(annotationClass) != null) {
    ...
  }
}
```
We could extract that into another method so that it's easier to call multiple times in this method (see my other comment).
```java
void checkTransientNotAnnotated(
      ExecutableElement getMethod,
      VariableElement field,
      Element elementToWarnOn)
```
Also possibly report all offending annotations if there is more than one.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -347,4 +352,80 @@ private CqlNameGenerator buildCqlNameGenerator(TypeElement classElement) {
...
+                otherAnnotationClass.getSimpleName());
+      }
+      return true;
+    }
+
+    // check if property is annotated with @Transient
+    Transient getterAnnotation = getMethod.getAnnotation(Transient.class);
+    boolean isTransient;
+    Element transientElement = getMethod;
+    isTransient = getterAnnotation != null;
```
olim7t: This is really hard to follow. How about treating the cases separately:
```java
if (getMethod.getAnnotation(Transient.class) != null) {
  checkTransientNotAnnotated(field, getMethod, getMethod);
  return true;
} else if (field != null && (field.getAnnotation(Transient.class) != null || field.getModifiers().contains(Modifier.TRANSIENT))) {
  checkTransientNotAnnotated(field, getMethod, field);
  return true;
}
return false;
```

And `checkTransientNotAnnotated` would have a more generic message "Transient properties can't be annotated with %s" to handle both field annotation and modifier.

olim7t (changes requested):

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
tomekl007: Maybe we can go one step further and extract `field != null && (field.getAnnotation(Transient.class) != null || field.getModifiers().contains(Modifier.TRANSIENT)` condition to separete method to name that check (to make it more readable)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
tolbertam: I like this idea, I think we could take it one step further and use it for more than just transient as I will also face the same issue with `@Computed`.  I realized that the 3.x mapper must have similar logic to prevent multiple conflicting annotations from being used on a property and that is the case: 

https://github.com/datastax/java-driver/blob/3.x/driver-mapping/src/main/java/com/datastax/driver/mapping/AnnotationChecks.java#L53-L93

I think the solution there works but is a bit more complicated than it needs to be.  I"ll try to come up with something.

tolbertam: Actually, there are parts of it that I like a lot, such as resolving an annotation in a general way (i.e. pass both the field and the method to resolve the annotation), which will be nice for custom inheritence strategies when we add it.   I'll take a stab at refactoring things in a way to meet my needs, without trying to complicate things too much.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -347,4 +304,126 @@ private CqlNameGenerator buildCqlNameGenerator(TypeElement classElement) {
...
+      Map<Class<? extends Annotation>, Annotation> annotations) {
+    for (Class<? extends Annotation> annotationClass : annotations.keySet()) {
+      if (EXCLUSIVE_PROPERTY_ANNOTATIONS.contains(annotationClass)) {
+        return annotationClass;
+      }
+    }
+    return null;
+  }
+
+  private void scanFieldAnnotations(
```
tolbertam, line 392: code for `scanFieldAnnotations` and `scanMethodAnnotations` is largely the same, but anticipate that changing with [JAVA-2111](https://datastax-oss.atlassian.net/browse/JAVA-2111)

tolbertam:
looks like i have an errorprone issue to fix, will do

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityPropertyAnnotationsTest.java
```diff
@@ -105,7 +105,7 @@ public void should_fail_with_expected_error(String expectedError, TypeSpec entit
   public static Object[][] entitiesWithErrors() {
     return new Object[][] {
       {
-        "Properties can't be annotated with both @PartitionKey and @ClusteringColumn.",
+        "Properties can't be annotated with both @ClusteringColumn and @PartitionKey.",
```
tolbertam, line 109: Was wondering if we had a way to test for compiler errors being surfaced, I should add some more tests as I had been testing manually up to this point, but if it expedites things more I can add tests when i work on [JAVA-2111](https://datastax-oss.atlassian.net/browse/JAVA-2111)

olim7t (approved):
OK, only nit is that it allows `Transient` + `CqlName`, which strictly speaking is not a valid combination, but we can live with that.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityPropertyAnnotationsTest.java
olim7t: See `EntityPropertyAnnotationsTest`.

Those tests are pretty slow, so try not to add too many (I'm thinking of adding something in the POM to skip them).

olim7t: Mmm forget that, looks like I misunderstood your question.

olim7t: No need to expedite this PR, we might as well add all necessary tests now.

tolbertam:
> only nit is that it allows Transient + CqlName

I considered that case, but didn't seem too important to validate :+1:

PR #1240, opened 2019-05-23
JAVA-2118: Generate QueryProvider DAO methods

olim7t:
TODO:
- [x] manual


integration-tests/src/test/java/com/datastax/oss/driver/mapper/QueryProviderIT.java
```diff
@@ -0,0 +1,231 @@
...
+
+      Select select = this.selectStart;
+      if (month != null) {
+        select = select.whereColumn("month").isEqualTo(bindMarker());
+        if (day != null) {
+          select = select.whereColumn("day").isEqualTo(bindMarker());
+        }
+      }
+      PreparedStatement preparedStatement =
+          session.execute(new DefaultPrepareRequest(select.build()), PrepareRequest.SYNC);
```
olim7t: So far the mapper wraps the top-level `Session` interface, which will force `QueryProvider` implementations to use the generic execute method.

I'm a bit on the fence about that: on the one hand it's nice to be as generic as possible, on the other that generic signature is a pain to use, and the probability that someone would initialize the mapper with something that is not a `CqlSession` is really thin.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/QueryProviderIT.java
```diff
@@ -0,0 +1,231 @@
...
+
+    public FindSliceProvider(
+        MapperContext context, EntityHelper<SensorReading> sensorReadingHelper) {
+      this.session = context.getSession();
+      this.sensorReadingHelper = sensorReadingHelper;
+      this.selectStart =
+          sensorReadingHelper.selectStart().whereColumn("sensor_id").isEqualTo(bindMarker());
+    }
+
+    public PagingIterable<SensorReading> findSlice(int id, Integer month, Integer day) {
```
tomekl007, line 121: Did we consider the approach of implementing `SensorDao` with custom query provider logic instead of using annotation with a provider? 
`PagingIterable<SensorReading> findSlice(int id, Integer month, Integer day);` 
needs to have the same signature in FindSliceProvider and SensorDao but it is not verified in a client code at compile time. (I am just wondering about the pros and cons of both solutions)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoQueryProviderMethodGenerator.java
```diff
@@ -0,0 +1,191 @@
...
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.element.VariableElement;
+import javax.lang.model.type.TypeKind;
+import javax.lang.model.type.TypeMirror;
+
+public class DaoQueryProviderMethodGenerator extends DaoMethodGenerator {
+
+  private static final EnumSet<ReturnTypeKind> SUPPORTED_RETURN_TYPES =
+      EnumSet.of(
```
tomekl007: We are supporting all types, right?
If so maybe we should move that list to `ReturnTypeKind` and call it `ALL` 
Then here only reference `ReturnTypeKind.ALL`

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoQueryProviderMethodGenerator.java
```diff
@@ -0,0 +1,191 @@
...
+    for (Map.Entry<? extends ExecutableElement, ? extends AnnotationValue> entry :
+        annotationMirror.getElementValues().entrySet()) {
+      if (entry.getKey().getSimpleName().contentEquals("providerClass")) {
+        return ((TypeMirror) entry.getValue().getValue());
+      }
+    }
+    // providerClass is mandatory on the annotation, so if we get here the user will already have to
+    // deal with a compile error.
+    // But return something so that the processor doesn't crash downstream.
+    return context.getTypeUtils().getPrimitiveType(TypeKind.INT);
```
tomekl007, line 95: Why we are choosing `INT`? Wouldn't it confuse users?

olim7t: See the comment. What we return is not important, we have already issued a compile-time error.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/QueryProviderIT.java
olim7t: You mean copy over the code in our generated implementation? I'm not sure if that's possible.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoQueryProviderMethodGenerator.java
olim7t: True, we can even get rid of the set altogether and only test `returnType.kind == UNSUPPORTED` in the code.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/QueryProviderIT.java
olim7t: All things considered I'm going to do the change to take a `CqlSession`.

Heads up that I'll force-push it on this PR.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/QueryProviderIT.java
olim7t: > it is not verified in a client code at compile time

Actually it is. Here's how the DAO method is implemented:
```java
  @Override
  public PagingIterable<QueryProviderIT.SensorReading> findSlice(int id, Integer month,
      Integer day) {
    return findSliceInvoker.findSlice(id, month, day);
  }
```
This won't compile if the parameters don't match.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoQueryProviderMethodGenerator.java
olim7t: And in fact we don't even need to validate the return type at all here. It can be anything as long as the provider method matches.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/QueryProviderIT.java
```diff
@@ -0,0 +1,224 @@
...
+  @Mapper
+  public interface SensorMapper {
+    @DaoFactory
+    SensorDao sensorDao(@DaoKeyspace CqlIdentifier keyspace);
+  }
+
+  @Dao
+  public interface SensorDao {
+    @QueryProvider(providerClass = FindSliceProvider.class, entityHelpers = SensorReading.class)
+    PagingIterable<SensorReading> findSlice(int id, Integer month, Integer day);
```
tolbertam, line 102: It took me a little bit of time to understand the value of this, but now i'm warming up to it :).  This feels like a 'power user' thing, where I expect most to be comfortable with existing capabilities and not need this level of flexibility.

I think the main things are:

1. Provides a level of customization at Dao level that didn't previously exist without injecting statement/statementbuilder or ResultSet/Row into the dao interface.  If you wanted to do something fancy previously, you'd either have to use `@Query` to provide CQL, pass in a `BoundStatementBuilder` to inject entity values or on the result mapping side, or pass in a ResultSet/Row from an executed query to map the result back.

2. Having access to `EntityHelper`s gives you direct access to methods for mapping between entities and results, getting statement basis and so forth.

Am I understanding that right?

integration-tests/src/test/java/com/datastax/oss/driver/mapper/QueryProviderIT.java
```diff
@@ -0,0 +1,224 @@
...
+    SensorReading readingFeb3 = new SensorReading(1, 2, 3, 9.3);
+    SensorReading readingFeb2 = new SensorReading(1, 2, 2, 8.6);
+    SensorReading readingFeb1 = new SensorReading(1, 2, 1, 8.7);
+    SensorReading readingJan31 = new SensorReading(1, 1, 31, 8.2);
+    dao.save(readingFeb3);
+    dao.save(readingFeb2);
+    dao.save(readingFeb1);
+    dao.save(readingJan31);
+
+    assertThat(dao.findSlice(1, null, null).all())
```
tolbertam, line 86: Wish java had optional arguments :D.  I guess technically you could overload this in the Dao interface and in `FindSliceProvider`

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -103,6 +105,22 @@ public String addPreparedStatement(
     return fieldName;
   }
 
+  @Override
+  public String addQueryProvider(
+      ExecutableElement methodElement,
+      TypeMirror providerClass,
+      List<ClassName> entityHelperTypes) {
+    // Invokers are not shared between methods, so always generate a new name
+    String fieldName = nameIndex.uniqueField(methodElement.getSimpleName().toString() + "Invoker");
```
tolbertam, line 114: Nice, i was wondering if this would work for overloaded methods and it appears it does (i.e. `findSliceInvoker1` is defined if a previous `findSliceInvoker` exists.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/QueryProvider.java
```diff
@@ -0,0 +1,99 @@
...
+   *
+   * <p>This is optional; if not provided, it defaults to the name of the annotated DAO method.
+   */
+  String providerMethod() default "";
+
+  /**
+   * A list of entities for which {@link EntityHelper} instances should be injected into the
+   * provider class's constructor (in addition to the mapper context).
+   *
+   * <p>For example, if {@code entityHelpers = {Product.class, Dimensions.class}}, your provider
```
tolbertam, line 91: What are the use cases for specifying multiple `EntityHelper` classes?  The only thing that comes to mind is maybe batch statements?

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
```diff
@@ -113,14 +113,45 @@ protected static SimpleStatement replaceKeyspaceAndTablePlaceholders(
...
+    parameterTypes[0] = MapperContext.class;
+    for (int i = 1; i < parameterTypes.length; i++) {
+      parameterTypes[i] = EntityHelper.class;
+    }
+
+    Object[] arguments = new Object[1 + entityHelpers.length];
+    arguments[0] = context;
+    System.arraycopy(entityHelpers, 0, arguments, 1, entityHelpers.length);
+    try {
+      return providerClass.getDeclaredConstructor(parameterTypes).newInstance(arguments);
```
tolbertam: This means that if the constructor does not match what is expected, things will fail at runtime and not at compile time right?  Is there any reason we couldn't create the instance in generated code?

tolbertam (commented):
This is pretty neat!  Really like the idea, it feels like something that most won't need, but might be needed in some cases to push away some complexity from the dao interface methods.

Had some feedback, and still need to review the doc changes in the manual which I saw was just pushed.  :+1:

tolbertam:
docs look good :+1:

integration-tests/src/test/java/com/datastax/oss/driver/mapper/QueryProviderIT.java
olim7t: Yes, definitely power user.

1. The main point for me is the ability to build the query dynamically depending on the parameters. With all the other annotations the query string is completely defined at compile time, so yeah the only alternative would be to execute the query outside of the mapper, and only use `GetEntity` / `SetEntity` to inject your parameters and extract the results. This is essentially that, but more cleanly encapsulated, from the caller's POV it's just another DAO method.

2. Correct, entity helpers give you access to the lower-level building blocks used internally by the mapper.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/QueryProviderIT.java
olim7t: Yes.

Actually for this particular case it would probably be more simple to declare 3 `@Query` methods, but that's just a silly example.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
olim7t: Yes. Same for prepared statements.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/QueryProvider.java
olim7t: It could be useful for nested UDTs too.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
olim7t: 🤦‍♂  you're right, we don't need reflection here

integration-tests/src/test/java/com/datastax/oss/driver/mapper/QueryProviderIT.java
tomekl007: I was thinking about this approach:
https://docs.spring.io/spring-data/data-commons/docs/1.6.1.RELEASE/reference/html/repositories.html
`1.3.1 Adding custom behavior to single repositories`
But if the current solution will generate compile error that's good.

tomekl007 (approved):

tolbertam (approved):
Change to use constructor directly looks great!

PR #1241, opened 2019-05-23
JAVA-2141: Support null saving strategies

tomekl007:


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -259,6 +292,113 @@ public static void setValue(
...
+
+  private static void generatePrimitiveSet(
+      CodeBlock cqlName,
+      CodeBlock valueExtractor,
+      String targetName,
+      MethodSpec.Builder methodBuilder,
+      String primitiveAccessor,
+      CodeBlock nullSavingStrategy) {
+
+    // todo how to handle primitive type getters? (int id == null is not possible)
```
tomekl007: @olim7t should we handle primitive types?
Primitive types cannot have null value so I think null handling strategy does not apply

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -162,7 +168,25 @@ public static void setValue(
...
       BindableHandlingSharedCode enclosingClass) {
+    // invoke with SET_TO_NULL to preserve default behaviour when retrieving an entity
+    setValue(
+        cqlName,
+        type,
+        valueExtractor,
+        targetName,
+        methodBuilder,
+        enclosingClass,
+        CodeBlock.of("$T.$L", NullSavingStrategy.class, NullSavingStrategy.SET_TO_NULL));
```
tomekl007: This method is using for retrieval of the entity (`@Select`) and `@Delete` - it does not apply to save so I fall back to default - `NullSavingStrategy.SET_TO_NULL` - it means that it just set nulls on the retrieved entity.
Other solution will be to create `CodeBlock.of("noOp")` and add `else if` branch here:
https://github.com/datastax/java-driver/pull/1241/files#diff-8992e8a0bae64eda9aaf4ff5012fa95fR378
and handle that case exactly like previously but I don't think that we need to add that complexity. Let me know wdyt

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Query.java
```diff
@@ -141,4 +141,6 @@
    * was built with. See the top-level javadocs of this class for more explanations.
    */
   String value();
+
+  // todo do we need NullSavingStrategy to Query? (I think that it only retrieves data)
```
tomekl007: I am not sure why in JIRA ticket there was info that we need to handle `NullSavingStrategy` for `@Query`.
In `QueryIT` I see that all queries are `SELECT` or `DELETE` so `nullSavingStrategy` has no influence over them.
Do we plan to support `INSERT ..` in `@Query` annotation?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
tolbertam: that sounds right to me, no special logic should be needed for primitives :+1:

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -259,6 +292,113 @@ public static void setValue(
...
+
+  private static void generateSetWithNullSavingStrategy(
+      CodeBlock cqlName,
+      CodeBlock valueExtractor,
+      String targetName,
+      MethodSpec.Builder methodBuilder,
+      CodeBlock nullSavingStrategy,
+      CodeBlock nonNullStatement) {
+    if (nullSavingStrategy.equals(DO_NOT_SET_CODE_BLOCK)) {
+      methodBuilder.beginControlFlow("if($1L == null)", valueExtractor);
```
tolbertam: This essentially generates:

```java
if (___ == null) {
  // do not set this field
} else {
  __nonNullStatement__
}
```

Couldn't it just  be rewritten as 

```java
if (___ != null) {
  __nonNullStatement__
}
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -259,6 +292,113 @@ public static void setValue(
...
+      // the nullSavingStrategy is dynamic - we cannot infer it at compile time
+      methodBuilder.beginControlFlow("if($1L == null)", valueExtractor);
+      methodBuilder.beginControlFlow(
+          "if($1L.equals($2T.$3L))",
+          nullSavingStrategy,
+          NullSavingStrategy.class,
+          NullSavingStrategy.SET_TO_NULL);
+      methodBuilder.addStatement("$1L = $1L.setToNull($2L)", targetName, cqlName);
+      methodBuilder.endControlFlow();
+      methodBuilder.beginControlFlow(
```
tolbertam: Similarly, this else if branch isn't needed if it doesn't doing anything right?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
tolbertam: Its an interesting decision point because it's possible that fields may be initialized with some non-null value in their default constructors.  However, conceptually, I think what you are doing is the right thing here.  If the value is unset in C*, it should be effectively null in the entity itself.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Insert.java
```diff
@@ -125,4 +126,11 @@
    * the rules of {@link Long#parseLong(String)}), the mapper will issue a compile-time warning.
    */
   String timestamp() default "";
+
+  /**
+   * The nullSavingStrategy to use in the generated INSERT query.
+   *
+   * <p>The default value for @Insert is {@link NullSavingStrategy#SET_TO_NULL}
+   */
+  NullSavingStrategy nullSavingStrategy() default NullSavingStrategy.SET_TO_NULL;
```
tolbertam: I understand that this is effectively needed because protocol V3 does not support leaving values unset, but this being the default was a truly sharp edge for the mapper in 3.x.

I know that C* 2.1 is still very popular, but I wonder if there's anything we should do to help mitigate our users dumping a ton of tombstones into their tables.

One possibility is that we could make `NullSavingStrategy.DO_NOT_SET` the default, and if is protocol V3 is selected, throw a preemptive runtime error earlier with a descriptive error.  I think this is better than adding tombstones unknowingly, and as more users embrace C 3.0+ this becomes a non-issue.   This wasn't done in the 3.x mapper because we added this capability after the mapper was originally released and we didn't want to change default behavior, but now that we're doing a major revision, changing the default is worth considering.

Thoughts?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -259,6 +292,113 @@ public static void setValue(
...
+      CodeBlock valueExtractor,
+      String targetName,
+      MethodSpec.Builder methodBuilder,
+      CodeBlock nullSavingStrategy,
+      CodeBlock nonNullStatement) {
+    if (nullSavingStrategy.equals(DO_NOT_SET_CODE_BLOCK)) {
+      methodBuilder.beginControlFlow("if($1L == null)", valueExtractor);
+      methodBuilder.addComment("do not set this field");
+      methodBuilder.endControlFlow();
+      methodBuilder.beginControlFlow("else");
```
tolbertam: Another thought: if we leave values unset, this will raise an error:

```
java.lang.IllegalStateException: Unset value at index 1. If you want this value to be null, please set it to null explicitly.
```

Fortunately, we never send the query to the server, so I think that's how we want this to behave.

But I think we don't have any exposure to this in our testing because all mapper tests that use this option have a `@CassandraRequirement(min="3.4", description = "Creates a SASI index")` annotation.  We should probably reorg our tests so we can run as much of it on older versions as possible.  I think this would touch a lot of classes, so we should probably do this towards the end of development in this phase.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/InsertIT.java
```diff
@@ -238,6 +267,12 @@ public void should_insert_entity_if_not_exists_returning_optional_asynchronously
     @Insert
     void save(Product product);
 
+    @Insert(nullSavingStrategy = NullSavingStrategy.DO_NOT_SET)
```
tolbertam, line 270: In mapper 3.x, this could be configured at the mapper level ([mapper options](https://docs.datastax.com/en/developer/java-driver/3.7/manual/object_mapper/using/#mapper-options)), so all save operations could behave in a consistent way.   I think we'd also want a similar concept in the new mapper as well right?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
olim7t: Agreed. If you want the ability to do partial updates on a property, it should be mapped to the boxed type.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Query.java
olim7t: `@Query` is completely free form, it can be whatever you put in the query string.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/InsertIT.java
olim7t: We can't really propagate annotations from the mapper to DAOs. A DAO is processed independently, it doesn't know which mapper it "belongs" to. In fact, more than one mapper could return a given DAO interface.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Insert.java
olim7t: I agree with having 3.0-friendly defaults, and throwing early for lower versions (probably from the DAO's `initAsync`).

I think the default should depend on the annotation: for `@Update`, `DO_NOT_SET` feels intuitive for the partial update behavior, but with `@Insert`/`@SetEntity` I think I would expect everything to be saved/bound so `SET_TO_NULL`. `@Query` is less obvious as the query can be anything, I would go with `SET_TO_NULL` too.

(PS - just to be clear, that's how @tomekl007 did it already)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSetMethodGenerator.java
```diff
@@ -63,6 +65,7 @@ public EntityHelperSetMethodGenerator(
             .addTypeVariable(settableT)
             .addParameter(ParameterSpec.builder(entityDefinition.getClassName(), "entity").build())
             .addParameter(ParameterSpec.builder(settableT, "target").build())
+            .addParameter(ParameterSpec.builder(nullSavingStrategyT, "nullSavingStrategy").build())
```
olim7t: You don't need to explicitly build a type name, the builder accepts a simple class:
```suggestion
            .addParameter(ParameterSpec.builder(NullSavingStrategy.class, "nullSavingStrategy").build())
```
(I only did it for `SettableT` because it's more complex)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -259,6 +292,113 @@ public static void setValue(
...
+    } else if (nullSavingStrategy.equals(SET_TO_NULL_CODE_BLOCK)) {
+      methodBuilder.beginControlFlow("if($1L == null)", valueExtractor);
+      methodBuilder.addStatement("$1L = $1L.setToNull($2L)", targetName, cqlName);
+      methodBuilder.endControlFlow();
+      methodBuilder.beginControlFlow("else");
+      methodBuilder.addStatement(nonNullStatement);
+      methodBuilder.endControlFlow();
+    } else {
+      // the nullSavingStrategy is dynamic - we cannot infer it at compile time
+      methodBuilder.beginControlFlow("if($1L == null)", valueExtractor);
```
olim7t: Could you add a space after `if` in the generated code? That's how it's formatted everywhere else.
```suggestion
      methodBuilder.beginControlFlow("if ($1L == null)", valueExtractor);
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -259,6 +292,113 @@ public static void setValue(
...
+    if (nullSavingStrategy.equals(DO_NOT_SET_CODE_BLOCK)) {
+      methodBuilder.beginControlFlow("if($1L == null)", valueExtractor);
+      methodBuilder.addComment("do not set this field");
+      methodBuilder.endControlFlow();
+      methodBuilder.beginControlFlow("else");
+      methodBuilder.addStatement(nonNullStatement);
+      methodBuilder.endControlFlow();
+    } else if (nullSavingStrategy.equals(SET_TO_NULL_CODE_BLOCK)) {
+      methodBuilder.beginControlFlow("if($1L == null)", valueExtractor);
+      methodBuilder.addStatement("$1L = $1L.setToNull($2L)", targetName, cqlName);
```
olim7t: You don't need to call `setToNull` explicitly, all object setters handle null properly. This generates the following code:
```java
    if (searchString == null) {
      boundStatementBuilder = boundStatementBuilder.setToNull("searchString");
    }
    else {
      boundStatementBuilder = boundStatementBuilder.set("searchString", searchString, String.class);
    }
```
But it could just be the same as before:
```
boundStatementBuilder = boundStatementBuilder.set("searchString", searchString, String.class);
```

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Insert.java
tolbertam: > for `@Update`, DO_NOT_SET feels intuitive for the partial update behavior

Ah, I did not realize `@Update` was implemented in this way.

> with @Insert/@SetEntity I  think I would expect everything to be saved/bound so SET_TO_NULL

I don't think we should treat the null setting behavior based on the kind of mutation operation differently as I think in either case we are being opinionated about whether the user intends to apply a tombstone or not when really we can't be sure.

One could even argue that setting to null on update makes more sense than insert as creating a tombstone is meaningful in that its a marker that can mean something is  deleted which is important when merging cells and determining if replicas are consistent.  Not creating a tombstone does not have that impact, which is why you could argue that inserts shouldn't create them.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
olim7t: Agreed (referring to Andy's comment). Example of generated code (in `InventoryITBase_ProductHelper__MapperGenerated.set`):
```java
    if(entity.getId() == null) {
      if(nullSavingStrategy.equals(NullSavingStrategy.SET_TO_NULL)) {
        target = target.setToNull("id");
      }
      else if(nullSavingStrategy.equals(NullSavingStrategy.DO_NOT_SET)) {
        // do not set this field
      }
    }
    else {
      target = target.set("id", entity.getId(), UUID.class);
    }
```
Could be simplified as:
```java
if (entity.getId() != null || nullSavingStrategy == NullSavingStrategy.SET_TO_NULL) {
  target = target.set("id", entity.getId(), UUID.class);
}
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -162,7 +168,25 @@ public static void setValue(
...
+  }
 
+  public static void setValue(
+      CodeBlock cqlName,
+      PropertyType type,
+      CodeBlock valueExtractor,
+      String targetName,
+      MethodSpec.Builder methodBuilder,
+      BindableHandlingSharedCode enclosingClass,
+      CodeBlock nullSavingStrategy) {
```
olim7t: Passing the strategy as a code block makes things hard to follow. Also I don't like how we sometimes artificially build a block that contains a known strategy, just for the sake of comparing it to `DO_NOT_SET_CODE_BLOCK` or `SET_TO_NULL_CODE_BLOCK`.

There are 3 different cases in our generated code:
1. There is no strategy. E.g. when we bind parameters of custom clauses for `@Select` or `@Delete`. In this case we want to keep generating the exact same code as before this PR.
2. The strategy comes from a local variable (or parameter) in scope. This is the case when we generate helper methods. As I mentioned in another comment, we want to generate the following code:
    ```java
    if (entity.getId() != null || nullSavingStrategy == NullSavingStrategy.SET_TO_NULL) {
      target = target.set("id", entity.getId(), UUID.class);
    }
    ```
3. The strategy is known at compile time. This will be the case when we bind the parameters in `@Query`, because we call `setValue` directly from the main method generation, without going through a helper.

To simplify things, we can actually merge 3 into 2 by generating a local variable at the beginning, before we bind the arguments:
```java
    // In the main method generator, initialize the variable to what's declared on the annotation:
    NullSavingStrategy nullSavingStrategy = NullSavingStrategy.DO_NOT_SET;
    // Then call setValue to generate this:
    if (entity.getId() != null || nullSavingStrategy == NullSavingStrategy.SET_TO_NULL) {
      target = target.set("id", entity.getId(), UUID.class);
    }
```
This looks a bit weird but whatever. If there's a performance impact it will be minimal, the VM will probably optimize it at runtime anyway.

So with 2 and 3 merged, we are left with two cases to handle:
1. Don't use a strategy, generate the same code as before.
2. Use a strategy that comes from a local variable. I would simplify even further and assume that this variable is always named `nullSavingStrategy`.

With this, the last parameter of `setValue` can be a simple boolean: `useNullSavingStrategy`; and we only have two cases to handle in the generated code.

WDYT?

olim7t (changes requested):

integration-tests/src/test/java/com/datastax/oss/driver/mapper/InsertIT.java
tomekl007: Yes, I was also thinking about it but as Olivier mentioned - it is processed on a different level.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
tomekl007: I think it is a good idea. 
I was a bit worried about the performance impact of generating: 
```
if (entity.getId() != null || nullSavingStrategy == NullSavingStrategy.SET_TO_NULL) {
    target = target.set("id", entity.getId(), UUID.class);
}
```
if we know up front that `nullSavingStrategy == NullSavingStrategy.DO_NOT_SET`.
But you are right that JIT should optimize that so the impact should be negligible.
I've implemented it in this way and it simplified the generation of the code a lot.

One con of it is that we need to rely on the fact that the `NullSavingStrategy  nullSavingStrategy` variable needs to be present on the method builder. But if you are ok with it I am +1 on that. I've added `nullSavingStrategy` to `@Query` annotation as well.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
tomekl007: Yes, I think that `java.lang.IllegalStateException: Unset value at index 1. If you want this value to be null, please set it to null explicitly.` is expected in this case.
I am +1 on extending our tests all annotations for Cassandra 2, 3 and 4. 
I think we should create a separate ticket to reorg our tests and cover 2 and 4 as well.
Do you want to create such a ticket or should I do it?
I've added an integration test for `nullSavingStraetgy` that is covering Cassandra 2.1 and
capturing that exception: `java.lang.IllegalStateException: Unset value at index 1. If you want this value to be null, please set it to null explicitly.`
Please see this commit:
https://github.com/datastax/java-driver/pull/1241/commits/d1042844209bbad29945e2dbcbd7aa7c0d135e11

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Insert.java
tomekl007: - Concerning catching exception add adding custom: 
 don't you think that this:
`Unset value at index 0. If you want this value to be null, please set it to null explicitly.`
is self-explanatory? (I've added test to capture that for 2.1:
https://github.com/datastax/java-driver/pull/1241/commits/d1042844209bbad29945e2dbcbd7aa7c0d135e11#diff-f21dbbd50b1f380dcadfd0e026c9e219R69)
Also as @tolbertam pointed out it is generated at the client side so there is no roundtrip involved.

- Concerning defaults - I agree with @tolbertam that it is better to have the same default for all cases.

>I don't think we should treat the null setting behavior based on the kind of mutation operation differently as I think in either case we are being opinionated about whether the user intends to apply a tombstone or not when really we can't be sure.

Setting it to `DO_NOT_SET` for `@Update` is our guess and may surprise users when for example they will choose to use firstly `@Query("UPDATE a ...)` and then they will decide to migrate `@Update`. In that case, the null saving strategy will change to `SET_TO_NULL` and they may be surprised by that fact. 
So maybe we should pick one default and use for all operations? (`@Query`, `@SetEntity`, `@Update` and `@Insert`)?
Another aspect is that I think we should instruct our users to set that nullSavingStrategy parameter explicitly - it will enforce them to make a conscious decision and read about those strategies.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -222,7 +245,8 @@ public static void setValue(
           .addStatement("$T $L = $L.newValue()", UdtValue.class, udtValueName, udtTypeName);
       String childHelper = enclosingClass.addEntityHelperField(entityClass);
       methodBuilder
-          .addStatement("$L.set($L, $L)", childHelper, valueName, udtValueName)
+          .addStatement(
+              "$L.set($L, $L, $L)", childHelper, valueName, udtValueName, NULL_SAVING_STRATEGY)
```
olim7t: The null saving strategy we pass to `EntityHelper.set` doesn't really matter here: the driver doesn't have the ability to send partial UDTs, unset values will be serialized to null anyway (see `UdtCodec`). So for clarity's sake I think we should hard-code this to `DO_NOT_SET` in the generated code.

Same for collections of entities, so you don't need to propagate a strategy to `convertEntitiesIntoUdts`.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -222,7 +245,8 @@ public static void setValue(
...
           .addStatement("$1L = $1L.setUdtValue($2L, $3L)", targetName, cqlName, udtValueName)
           .endControlFlow();
```
olim7t, line 264: Just realized that we should also handle null for entities / collections of entities (not specific to this PR, I forgot to do it in the original code).

So we should generate an else block here:
```java
    if (value != null) {
      UserDefinedType udtType = (UserDefinedType) target.getType("dimensions");
      UdtValue udtValue = udtType.newValue();
      dimensionsHelper.set(value, udtValue, nullSavingStrategy);
      target = target.setUdtValue("dimensions", udtValue);
    // new code:
    } else if (nullSavingStrategy == NullSavingStrategy.SET_TO_NULL) {
      target = target.setUdtValue("dimensions", null);
    }
```

Same at line 283 for collections of entities.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoQueryMethodGenerator.java
```diff
@@ -157,8 +158,14 @@ public DaoQueryMethodGenerator(
         BoundStatementBuilder.class,
         statementName);
 
+    Query annotation = methodElement.getAnnotation(Query.class);
+    NullSavingStrategy nullSavingStrategy = annotation.nullSavingStrategy();
+
+    queryBuilder.addStatement(
+        "$1T nullSavingStrategy = $1T.$2L", NullSavingStrategy.class, nullSavingStrategy);
```
olim7t, line 130: 👍

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoSetEntityMethodGenerator.java
```diff
@@ -117,15 +118,20 @@ public DaoSetEntityMethodGenerator(
...
 
+    NullSavingStrategy nullSavingStrategy =
+        methodElement.getAnnotation(SetEntity.class).nullSavingStrategy();
+
     // Forward to the base injector in the helper:
     return Optional.of(
         GeneratedCodePatterns.override(methodElement)
             .addStatement(
-                "$L$L.set($L, $L)",
+                "$1L$2L.set($3L, $4L, $5T.$6L)",
```
olim7t, line 128: Nit: you don't strictly need to number placeholders if none of them is used more than once (but if you find it easier to read that's fine with me as well).

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateMethodGenerator.java
```diff
@@ -119,13 +120,22 @@ public DaoUpdateMethodGenerator(
...
+    NullSavingStrategy nullSavingStrategy = annotation.nullSavingStrategy();
+
+    methodBuilder.addStatement(
+        "$1T nullSavingStrategy = $1T.$2L", NullSavingStrategy.class, nullSavingStrategy);
     if (customWhereClause.isEmpty()) {
       // We generated an update by primary key (see maybeAddWhereClause), all entity properties are
       // present as placeholders.
       methodBuilder.addStatement(
-          "$L.set($L, boundStatementBuilder)", helperFieldName, entityParameterName);
+          "$1L.set($2L, boundStatementBuilder, $3L)",
```
olim7t: Nit: you can hardcode `nullSavingStrategy` here since it's constant
```suggestion
          "$1L.set($2L, boundStatementBuilder, nullSavingStrategy)",
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateMethodGenerator.java
```diff
@@ -119,13 +120,22 @@ public DaoUpdateMethodGenerator(
     List<? extends VariableElement> parameters = methodElement.getParameters();
     String entityParameterName = parameters.get(0).getSimpleName().toString();
 
-    String customWhereClause = methodElement.getAnnotation(Update.class).customWhereClause();
+    Update annotation = methodElement.getAnnotation(Update.class);
+    String customWhereClause = annotation.customWhereClause();
+    NullSavingStrategy nullSavingStrategy = annotation.nullSavingStrategy();
+
+    methodBuilder.addStatement(
+        "$1T nullSavingStrategy = $1T.$2L", NullSavingStrategy.class, nullSavingStrategy);
```
olim7t: Nit: I would only declare the variable in the `else` below, and hard-code it (like for `Insert`) in the first `if` branch.

olim7t (changes requested):
Thanks for addressing my feedback; indeed this simplified the code a lot, I like how it looks now.

Here are a few more details I missed on the last review. I will address defaults separately in reply to Andy's comment.

I was wondering if we need the ability to override the strategy for specific parameters, for example:
```java
@Query("UPDATE foo SET ... /* use DO_NOT_SET here */ WHERE ... /* use SET_TO_NULL here */")
```
But I think this is unlikely because AFAIK placeholders in WHERE clauses cannot be null. If someone comes up with a case, we can address it in a follow-up ticket.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Insert.java
olim7t: > catching exception

As discussed privately, it would be nice to generate a one-time check of the protocol version in `DAO.initAsync` if any method in the DAO interface uses `DO_NOT_SET`. You can do it with a callback `DaoImplementationSharedCode#addUnsetCapabilityCheck`.

> defaults

Yeah Andy convinced me. I was thinking of `Insert` in the case where the row does not exist yet, but then the strategy doesn't matter because the column will be null whether it was set or not.
Forcing the user to declare the strategy every time seems a bit harsh, so let's default to `DO_NOT_SET` everywhere.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/InsertIT.java
tolbertam: I still think we should consider some concept of Session-level configuration for the mapper for things such as this.  For example, if we make `DO_NOT_SET` the default and you are on 2.1, having to make sure this is set on every operation seems like it could be overbearing and possibly error prone.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/NullSavingStrategyIT.java
```diff
@@ -0,0 +1,177 @@
...
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+
+@Category(ParallelizableTests.class)
+@CassandraRequirement(
+  min = "2.1",
+  max = "2.2",
+  description = "required for testing behaviour when V3 protocol has no support for unset values"
+)
+public class NullSavingStrategyIT {
```
tolbertam, line 47: :+1:

integration-tests/src/test/java/com/datastax/oss/driver/mapper/InsertIT.java
olim7t: We could propagate a default value through `MapperContext`. Then our generated DAO methods would initialize the value like this:
```java
NullSavingStrategy nullSavingStrategy =
    (context.getNullSavingStrategy() != null)
        ? context.getNullSavingStrategy()
        : /* the value from the annotation */;
```

I have a few improvements coming in [JAVA-2262](https://datastax-oss.atlassian.net/browse/JAVA-2262) that would facilitate this (I'll open the PR shortly).

EDIT -- this is a bit more tricky because we can't distinguish when the annotation value is the default because it wasn't set, vs when it was set explicitly to that value.

EDIT 2 🤣  -- actually there is a way, when using the `AnnotationMirror` it doesn't fill in the defaults

olim7t: I suggest we create a follow-up ticket, since this depends on 2262 and there is already a lot going on in this PR.

tolbertam: agreed :+1:

tomekl007: I've created https://datastax-oss.atlassian.net/browse/JAVA-2265

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
tomekl007: I've created a JIRA ticket: https://datastax-oss.atlassian.net/browse/JAVA-2266

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Insert.java
tomekl007: 1. I've implemented custom one-time check where `V3` and `DO_NOT_SET`: https://github.com/datastax/java-driver/pull/1241/commits/ca4a9a3c9b1d08f1e642ca62f3da10f7dd731f50
I think that we don't need to add this method: `DaoImplementationSharedCode#addUnsetCapabilityCheck` to `DaoImplementationSharedCode` interface. It is enough to keep it private in `DaoImplementationGenerator`

2. I've changed defaults to DO_NOT_SET.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoSetEntityMethodGenerator.java
tomekl007: Thx, I noticed that but I think that In this case having indices on them makes the code more readable.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
tomekl007: This suggestion contradicts this one: 
https://github.com/datastax/java-driver/pull/1241#discussion_r287412551

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
tomekl007: - Implemented for UDT:
https://github.com/datastax/java-driver/pull/1241/commits/54fc4854a9d950394f25422289acd2d7883453f1
- for Collections:
https://github.com/datastax/java-driver/pull/1241/commits/106085e812ece59626b5322d4afb5a51ef83af78

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
tomekl007: Done https://github.com/datastax/java-driver/pull/1241/commits/c08b89a64a74124525aee9ae90bdf13e13275824

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Insert.java
olim7t: > It is enough to keep it private

👍
You implemented differently than what I had in mind, indeed no need to make it an interface method.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -222,8 +245,23 @@ public static void setValue(
...
+          .addStatement(
+              "$L.set($L, $L,  $T.$L)",
+              childHelper,
+              valueName,
+              udtValueName,
+              NullSavingStrategy.class,
+              NullSavingStrategy.DO_NOT_SET)
           .addStatement("$1L = $1L.setUdtValue($2L, $3L)", targetName, cqlName, udtValueName)
+          .endControlFlow()
+          .beginControlFlow(
```
olim7t: Note that
```java
          .endControlFlow()
          .beginControlFlow(
```
Can be shortened as:
```java
          .nextControlFlow(
```
I'm addressing this directly during the merge.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/NullSavingStrategyIT.java
```diff
@@ -0,0 +1,143 @@
...
+import org.junit.experimental.categories.Category;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+
+@Category(ParallelizableTests.class)
+@CassandraRequirement(
+  min = "2.1",
+  max = "2.2",
+  description = "required for testing behaviour when V3 protocol has no support for unset values"
+)
```
olim7t: It is possible to force the protocol version in the configuration. This way we don't need to restrict the Cassandra version for this test.
I'm also fixing that up during the merge.

olim7t (approved):

olim7t:
Merged 👍 
I sneaked in an update to the manual, and the two minor things I mentioned in comments.

PR #1242, opened 2019-05-23
JAVA-2254: Add support for computed properties

tolbertam:
For [JAVA-2254](https://datastax-oss.atlassian.net/browse/JAVA-2254)

Marking as wip as this needs to be rebased after #1239 is merged, and there's something I'm not completely satisfied with in `EntityHelperGetMethodGenerator` which I could use some feedback on.

---

Motivation:

It is useful to map values that are the result of some Cassandra-side
computation, typically function calls, to properties.

Modifications:

Introduce a `@Computed` annotation that can be applied to a field or
getter that allows specifying a formula to instruct Cassandra to
compute and return a value back that can be mapped back to this
property.

Introduce AliasedPropertyDefinition for enabling use of aliases in query
specification.  This is needed because the resulting value name for a
function call is not deterministic.

Add EntityDefinition.getComputedValues() for retrieving computed values
and added getAllValues(), which is similar to getAllColumns() but also
returns computed values.

Add PropertyDefinition.getCqlResultName() to differentiate between
request name and result name for alias definitions.

Update EntityHelperGetMethodGenerator to allow selectively retrieving
values when their cql name does not match their result name.

Update EntityHelperSelectStartMethodGenerator to include computed values
in generated select query.

Result:

`@Computed` annotation may be used to specify properties that are based
on result of some function call.  This is only usable by
`@Select`-annotated dao methods.


manual/mapper/entities/README.md
```diff
@@ -159,10 +159,43 @@ private int day;
...
+
+[@Computed] does not support case-sensitivity. If the expression contains case-sensitive column 
+or function names, you'll have to escape them:
+
+```java
+@Computed("\"myFunction\"(\"myColumn\")")
+private int f;
+```
+
+Finally, note that [@Computed] fields are only used for [@Select]-annotated dao methods. Both
```
tolbertam: Is this ok? Mapper 3.x was similar with `@Computed` not working with `@Accessor`, for which we opened [JAVA-832](https://datastax-oss.atlassian.net/browse/JAVA-832).   I'm not sure there is an easy way to resolve this without changing query contents.

One possible way to make that work is to alias the result to the name of the property.  We can then instruct others that when they make their queries to do the same, i.e.:

```
@Computed("ttl(x)")
private int xttl;
```

Would cause our select query to query this as `ttl(x) as xttl` instead of what i'm currently doing which is `ttl(x) as "[alias0]"`.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
```diff
@@ -62,24 +62,34 @@ public EntityHelperGetMethodGenerator(
...
     getBuilder.addStatement("$1T $2L = new $1T()", returnType, returnName);
 
-    for (PropertyDefinition property : entityDefinition.getAllColumns()) {
+    for (PropertyDefinition property : entityDefinition.getAllValues()) {
       PropertyType type = property.getType();
       CodeBlock cqlName = property.getCqlName();
+      CodeBlock cqlResultName = property.getCqlResultName();
       String setterName = property.getSetterName();
       getBuilder.addCode("\n");
+      // if the result name is not the same as the name, it's likely this is a computed
```
tolbertam: Not really happy with this, namely because throwing exceptions can be costly, but with `GettableByName` there is no way to determine if a definition is present without throwing an exception (`firstIndexOf` will throw an `IllegalArgumentException`).   If this was a `Row`, i could use `getDefinitions`.

Alternatively if we based alias names on property names (as described [here](https://github.com/tolbertam)) we could just remove this logic and assume the value is present, and if not fail like we would do for any other column.

manual/mapper/entities/README.md
olim7t: Yeah, I think it would make sense to use `getCqlName()` as the alias, and require that people do the  same in their custom queries.

tolbertam: Awesome, definitely like this solution, its much more pleasant and also solves my other problem in `EntityHelperGetMethodGenerator` and also opens up computed to be used with the other query-based annotations :+1:

tolbertam:
Rebased on java2078, will proceed with changing how alias works

manual/mapper/entities/README.md
tolbertam: Took a stab at that in 3767374 and I think it's a much more elegant solution overall.

I also made it work such that naming strategy / and @CqlName are regarded.

Updated tests to verify `@Query` and `@GetEntity` work assuming you make the right query, and updated the documentation to give instruction on how to do that.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
tolbertam: this is no longer an issue as this special logic is not needed.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/ComputedPropertyDefinition.java
```diff
@@ -0,0 +1,74 @@
...
+      String setterName,
+      PropertyType type,
+      CqlNameGenerator cqlNameGenerator) {
+    this.cqlResultName =
+        customCqlName
+            .map(n -> CodeBlock.of("$S", n))
+            .orElse(cqlNameGenerator.buildCqlName(javaName));
+    // the queried name is "formula as alias"
+    this.cqlName =
+        CodeBlock.of("$S + $S +", formula, " as ").toBuilder().add(this.cqlResultName).build();
```
tolbertam: Is there a better way to do this?  The generated code looks like:

```java
    return selectFrom
        .column("id")
        .column("c_id")
        .column("v")
        .raw("writetime(v)" + " as " +"writetime")
        .raw("ttl(v)" + " as " +"myttl");
```

Which works, but would have been nice to do it as one clean string constant.

olim7t: You can alias raw expressions:
```java
.raw("writetime(v)").as("writetime")
.raw("ttl(v)").as("myttl");
```

tolbertam: Nice, that's definitely better.  It will require some reorienting of `ComputedPropertyDefinition` and how it is used in `EntityHelperSelectStartMethodGenerator`, but I should probably do that anyways.

tolbertam: Was a little easier to do that than i originally anticipated (6b24969)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -81,6 +81,8 @@ public EntityDefinition getDefinition(TypeElement classElement) {
     SortedMap<Integer, PropertyDefinition> partitionKey = new TreeMap<>();
     SortedMap<Integer, PropertyDefinition> clusteringColumns = new TreeMap<>();
     ImmutableList.Builder<PropertyDefinition> regularColumns = ImmutableList.builder();
+    ImmutableList.Builder<PropertyDefinition> computedValues = ImmutableList.builder();
+    int aliasCounter = 0;
```
olim7t: Nit: unused

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityDefinition.java
```diff
@@ -54,4 +56,17 @@
...
+   *     {@link #getRegularColumns()}, and {@link #getComputedValues()} in that order.
+   */
+  default List<PropertyDefinition> getAllValues() {
+    return ImmutableList.<PropertyDefinition>builder()
+        .addAll(getPartitionKey())
+        .addAll(getClusteringColumns())
+        .addAll(getRegularColumns())
+        .addAll(getComputedValues())
+        .build();
+  }
```
olim7t, line 71: 👍 the terminology makes sense.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/ComputedPropertyDefinition.java
```diff
@@ -0,0 +1,72 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.internal.mapper.processor.entity;
+
+import com.datastax.oss.driver.internal.mapper.processor.util.generation.PropertyType;
+import com.squareup.javapoet.CodeBlock;
+import java.util.Optional;
+
+public class ComputedPropertyDefinition implements PropertyDefinition {
```
olim7t: Do we really need to duplicate the whole class? It feels like adding an `Optional<String> formula` parameter to the original class would be pretty easy.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/PropertyDefinition.java
```diff
@@ -22,6 +22,10 @@
 
   CodeBlock getCqlName();
 
+  default CodeBlock getCqlResultName() {
+    return getCqlName();
+  }
+
```
olim7t, line 46: I get which is which when I read this class, but in the rest of the code my head is exploding 😃 

First I think we need javadocs:
* `getCqlName`: a Java snippet that produces the corresponding expression in a SELECT statement, for example:
  * `"id"` in `selectFrom.column("id")` for a regular column
  * `"writetime(v)"` in `selectFrom.raw("writetime(v)")` for a computed value
* `getCqlResultName`: a Java snippet that produces the name of the property in a GettableByName or SettableByName, for example `"id"` in `row.get("id")`.

Then I would also rename them. Proposal:
* `getCqlName` => `getSelector`
* `getCqlResultName` => `getAlias`

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
```diff
@@ -62,9 +62,10 @@ public EntityHelperGetMethodGenerator(
     String returnName = "returnValue";
     getBuilder.addStatement("$1T $2L = new $1T()", returnType, returnName);
 
-    for (PropertyDefinition property : entityDefinition.getAllColumns()) {
+    for (PropertyDefinition property : entityDefinition.getAllValues()) {
       PropertyType type = property.getType();
       CodeBlock cqlName = property.getCqlName();
```
olim7t, line 67: `cqlName` shouldn't be referenced at all in this method. There are a couple of uses for UDTs and collections of UDTs below, but I think those should be changed to use `cqlResultName` too.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
```diff
@@ -74,12 +75,12 @@ public EntityHelperGetMethodGenerator(
           // Primitive type: use dedicated getter, since it is optimized to avoid boxing
           //     returnValue.setLength(source.getInt("length"));
           getBuilder.addStatement(
-              "returnValue.$L(source.get$L($L))", setterName, primitiveAccessor, cqlName);
+              "returnValue.$L(source.get$L($L))", setterName, primitiveAccessor, cqlResultName);
```
olim7t: There are other method generators where we should do that change, for example `EntityHelperInsertMethodGenerator`. In practice things work because they happen to deal exclusively with regular columns where `cqlName == cqlResultName`, but strictly speaking they should use `cqlResultName`.

If we do the `cqlName => selector` change that I'm suggesting in another comment, it will be more obvious and we can do a quick pass on the code.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/ComputedIT.java
```diff
@@ -0,0 +1,306 @@
...
+
+    ComputedEntity retrievedValue = computedDao.findByIdQuery(key, 1);
+    assertThat(retrievedValue.getId()).isEqualTo(key);
+    assertThat(retrievedValue.getcId()).isEqualTo(1);
+    assertThat(retrievedValue.getV()).isEqualTo(2);
+
+    // these should be set
+    assertThat(retrievedValue.getTtl()).isEqualTo(3600);
+    assertThat(retrievedValue.getWritetime()).isEqualTo(time);
+  }
```
tomekl007, line 248: I think we are missing test case that does `@Computed("ttl(v)")` and `@CqlName("myttl")` but cql name is not present in results from the query

integration-tests/src/test/java/com/datastax/oss/driver/mapper/ComputedIT.java
```diff
@@ -0,0 +1,306 @@
...
+    computedDao.saveWithTime(entity, 3600, time);
+
+    CqlSession session = sessionRule.session();
+    // query with the computed values included.
+    // since we the mapper expects the result name to match the property name, we used aliasing
+    // here.
+    ResultSet result =
+        session.execute(
+            SimpleStatement.newInstance(
+                "select id, c_id, v, writetime(v) as writetime, ttl(v) as myttl from "
```
tomekl007, line 180: Looking at this definition:
``` @CqlName("myttl")
    @Computed("ttl(v)")
    private int ttl;
```
Don't you think it is a little bit confusing? 
If the select query is returning `mytll` alias maybe it will be enough to do: `@Compued(ttl(myttl))`?

tolbertam: i'm intentionally doing this to test that the `@CqlName` annotation is used over the property name:

```java

    @CqlName("myttl")
    @Computed("ttl(v)")
    private int ttl;
```

I'll update the comment to reflect that.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/ComputedIT.java
tolbertam: You mean just to validate that an exception is thrown if the user provided query does not match the expectation of what the result alias should be?

olim7t (changes requested):
Looking really good, a few comments about naming.

manual/mapper/entities/README.md
```diff
@@ -159,6 +159,67 @@ private int day;
...
+
+[@Computed] does not support case-sensitivity. If the expression contains case-sensitive column
+or function names, you'll have to escape them:
+
+```java
+@Computed("\"myFunction\"(\"myColumn\")")
+private int f;
+```
+
+[@Computed] fields are only used for queries, so they will not be considered for [@Update] or
```
adutra: To clarify I would add: `[@Computed] fields are only used for [@Select] queries, so they will not be considered for [@Update]...`

manual/mapper/entities/README.md
```diff
@@ -159,6 +159,67 @@ private int day;
...
+@Entity
+class MyEntity {
+  @PartitionKey private int k;
+
+  private int v;
+
+  @Computed("ttl(v)")
+  private int myTtl;
+
+  @Computed("writetime(v)")
```
adutra, line 207: Wouldn't `@Computed(expression = "writetime(v)", alias = "ts")` be more explicit?
If not: wouldn't `@Alias` be a better name than `@CqlName`?

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Computed.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Annotates the field or getter of an {@link Entity} property, to indicate that when retrieving
+ * data that the property should be set to the result of computation on the Cassandra side,
+ * typically a function call.
+ *
+ * <p>Example:
+ *
+ * <pre>
+ * private int v;
```
adutra: This line seems unnecessary.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/ComputedPropertyDefinition.java
```diff
@@ -0,0 +1,72 @@
...
+
+  private final CodeBlock cqlName;
+  private final CodeBlock cqlResultName;
+  private final String getterName;
+  private final String setterName;
+  private final PropertyType type;
+
+  public ComputedPropertyDefinition(
+      String javaName,
+      Optional<String> customCqlName,
```
adutra: It is not a good practice to pass `Optional` as a method parameter, to avoid unnecessary wrapping/unwrapping (and also because `Optional` is not serializable). I know it's already done in a few other places, but wanted to pointed that out. A better approach is to pass the string annotated with `@Nullable`.

adutra (changes requested):

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/ComputedPropertyDefinition.java
tolbertam: I'll be factoring this code out into DefaultPropertyDefinition, will update the optional parameters in that constructor when I get to it.

olim7t: > It is not a good practice to pass Optional as a method parameter

I don't 100% agree with that IDE warning. Performance is not paramount in the processor (at least not to the point where wrapping something in an optional will make a difference), and we don't store the optional in a field so serialization is not an issue (we don't serialize this object anyway for that matter).

Passing an optional here works nicely because the annotation parsing code that is used to fill that parameter already returns one.

manual/mapper/entities/README.md
olim7t: > `@Computed(expression = "writetime(v)", alias = "ts")`

This introduces a special case though, isn't it better if computed values behave like any other property?

> wouldn't `@Alias` be a better name than `@CqlName`?

👍 to that. This is in line with the renaming I suggested [here](https://github.com/datastax/java-driver/pull/1242#discussion_r287579567).

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/PropertyDefinition.java
tolbertam: Sounds good, i was taking a path of least resistance initially.  Agree that renaming things makes things easier to understand.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/ComputedPropertyDefinition.java
adutra: I agree with you, the only real bad thing is to store optionals as fields, but that is not the case here, so I'm fine if we keep them.

manual/mapper/entities/README.md
tolbertam: It also works for `@GetEntity` and `@Query` as well (see sections below), so I think it makes sense as written, but can adjust it if you prefer.

adutra: It's just that, at least to my ears, "queries" can refer to any form of CQL query, including inserts and updates. But OK to leave it as is if you think it's clear enough.

olim7t: Maybe we could list them explicitly.

adutra:
`ComputedIT` seems to be failing against C* 2.1:

https://jenkins-drivers.build.dsinternal.org/view/java-driver/job/datastax.java-driver.PR.1242.default/8/testReport/

manual/mapper/entities/README.md
tolbertam: > Isn't it better if computed values behave like any other property?

That's what I was aiming for by using `@CqlName`.  `CqlName` is used to indicate the name of the cql column, but also now the aliased result name for computed values.  I think it fits really well.

If we change it to `@Alias`, the intent could be interpreted differently, effectively as a rename, but that will not generally be the case (only for `@Computed` properties).  I don't think there would be too much of a use case for adding something like `@Alias` to rename how to get at the resulting value beyond computed values.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
tolbertam: Correct me if I am mistaken, but I do not think this is relevant here now that we are renaming `getCqlResultName()` -> `getAlias()`.   Aliasing is only relevant in the context of select-based methods ([doc](https://github.com/apache/cassandra/blob/trunk/doc/cql3/CQL.textile#select-clause)), so it does not apply in the context insert/update/delete operations.

manual/mapper/entities/README.md
olim7t: > If we change it to `@Alias`, the intent could be interpreted differently

Mmm that's a good point, "alias" would imply that we use a different name than the column's, but for non-computed columns that is not the case.
So I change my vote to keep `CqlName`.

adutra: I'm sorry but I find `CqlName` way too vague. If I follow you correctly this annotation is being used to:

1. Customize regular column names: why not introduce a special `@Column` annotation?
2. To define aliases for computed properties: why don't we revert then to my initial idea of `@Computed(expression = "writetime(v)", alias = "ts")`? It seems that only computed properties will ever need aliases, so that option sounds fair to me.

olim7t: > why not introduce a special `@Column` annotation?

That sounds even more vague to me. Marking something as a "column" doesn't say anything about naming, and you might have other properties that are not annotated but will still map to a column.

> To define aliases for computed properties

The advantage of the current solution is that it works the same way whether the property is computed or not:

"What name should I use to get the property on my `Row`?"
* if `CqlName` is present, that
* otherwise, the name of the Java property, translated with the naming strategy

Why make it different for computed properties?

manual/mapper/entities/README.md
tolbertam: updated to `select-based` queries so it's more clear.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
tolbertam: After discussion we opted for `getSelector()` and `getCqlName()`, which makes it clear that `getCqlName` should be used here and most other places except when generating select query.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/PropertyDefinition.java
tolbertam: [More discussion here](https://github.com/datastax/java-driver/pull/1242#discussion_r288206815)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/ComputedPropertyDefinition.java
tolbertam: :+1:, I went ahead and merged logic into `DefaultPropertyDefinition`

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
tolbertam: After reorienting `PropertyDefinition`, `cqlName` is now appropriate again :+1:

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSelectStartMethodGenerator.java
```diff
@@ -56,8 +56,13 @@ public EntityHelperSelectStartMethodGenerator(
                 QueryBuilder.class)
             .addCode("$[return selectFrom");
 
-    for (PropertyDefinition property : entityDefinition.getAllColumns()) {
-      selectStartBuilder.addCode("\n.column($L)", property.getCqlName());
+    for (PropertyDefinition property : entityDefinition.getAllValues()) {
+      if (property.getSelector().equals(property.getCqlName())) {
```
olim7t: Nit: this does not clearly indicate that we distinguish computed values.

We could either introduce `PropertyDefinition.isComputed()`, or iterate `getAllColumns` and `getComputedValues` separately.

olim7t (approved):
LGTM

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSelectStartMethodGenerator.java
tolbertam: That's actually how I previously had it, my reasoning for changing it was in possible anticipation of aliasing not necessarily being bound to computed properties, although that seems rather unlikely.

it probably is better to make it more explicit that this is why things are being aliased here, I'll change it back.

manual/mapper/entities/README.md
adutra: You know what, I'm retiring myself from this endless debate and let you guys decide what is best. Good luck ;-)

adutra (approved):

PR #1243, opened 2019-05-24
JAVA-2262: Allow user to store custom state in MapperContext

olim7t:
Note - I don't think this will be needed very often, it's more of an escape hatch. So no need to be super fancy (the map stores Object, you need to cast, etc.)

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/mapper/QueryProviderIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/mapper/QueryProviderIT.java
@@ -41,6 +41,7 @@ import com.datastax.oss.driver.api.testinfra.ccm.CcmRule;
 import com.datastax.oss.driver.api.testinfra.session.SessionRule;
 import com.datastax.oss.driver.categories.ParallelizableTests;
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
@@ -56,6 +57,10 @@ public class QueryProviderIT {
   private static SessionRule<CqlSession> sessionRule = SessionRule.builder(ccm).build();
   @ClassRule public static TestRule chain = RuleChain.outerRule(ccm).around(sessionRule);
 
+  // Dummy counter to exercize the "custom state" feature: it gets incremented each time the query
+  // provider is called.
+  private static AtomicInteger executionCount = new AtomicInteger();
+
   private static SensorDao dao;
 
   @BeforeClass
@@ -70,7 +75,10 @@ public class QueryProviderIT {
             .setExecutionProfile(sessionRule.slowProfile())
             .build());
 
-    SensorMapper mapper = new QueryProviderIT_SensorMapperBuilder(session).build();
+    SensorMapper mapper =
+        new QueryProviderIT_SensorMapperBuilder(session)
+            .withCustomState("executionCount", executionCount)
+            .build();
     dao = mapper.sensorDao(sessionRule.keyspace());
   }
 
@@ -85,11 +93,15 @@ public class QueryProviderIT {
     dao.save(readingFeb1);
     dao.save(readingJan31);
 
+    assertThat(executionCount.get()).isEqualTo(0);
+
     assertThat(dao.findSlice(1, null, null).all())
         .containsExactly(readingFeb3, readingFeb2, readingFeb1, readingJan31);
     assertThat(dao.findSlice(1, 2, null).all())
         .containsExactly(readingFeb3, readingFeb2, readingFeb1);
     assertThat(dao.findSlice(1, 2, 3).all()).containsExactly(readingFeb3);
+
+    assertThat(executionCount.get()).isEqualTo(3);
   }
 
   @Mapper
@@ -109,12 +121,14 @@ public class QueryProviderIT {
 
   public static class FindSliceProvider {
     private final CqlSession session;
+    private final AtomicInteger executionCount;
     private final EntityHelper<SensorReading> sensorReadingHelper;
     private final Select selectStart;
 
     public FindSliceProvider(
         MapperContext context, EntityHelper<SensorReading> sensorReadingHelper) {
       this.session = context.getSession();
+      this.executionCount = ((AtomicInteger) context.getCustomState().get("executionCount"));
       this.sensorReadingHelper = sensorReadingHelper;
       this.selectStart =
           sensorReadingHelper.selectStart().whereColumn("id").isEqualTo(bindMarker());
@@ -124,6 +138,8 @@ public class QueryProviderIT {
       if (month == null && day != null) {
         throw new IllegalArgumentException("Can't specify day if month is null");
       }
+
+      executionCount.incrementAndGet();
 
       Select select = this.selectStart;
       if (month != null) {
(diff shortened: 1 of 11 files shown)


mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MapperBuilder.java
```diff
@@ -0,0 +1,59 @@
...
+   *
+   * <p>This is intended mainly for {@link QueryProvider} methods: since provider classes are
+   * instantiated directly by the generated mapper code, they have no way to access non-static state
+   * from the rest of your application. This method allows you to pass that state while building the
+   * mapper, and access it later at runtime.
+   *
+   * <p>Note that this state will be accessed concurrently, it should be thread-safe.
+   */
+  @NonNull
+  public MapperBuilder<MapperT> withCustomState(@Nullable Object key, @Nullable Object value) {
```
olim7t, line 53: No self type here. The subtypes are all generated and they never add their own methods.

manual/mapper/mapper/README.md
```diff
@@ -36,6 +36,22 @@ CqlSession session = CqlSession.builder().build();
...
+  
+  static MapperBuilder<InventoryMapper> builder(CqlSession session) {
+    return new InventoryMapperBuilder(session);
+  }
+  ...
+}
+
+InventoryMapper inventoryMapper = InventoryMapper.builder(session).build();
+```
+
```
olim7t, line 54: Not strictly related to this PR, but I just realized you can do this and I think it's pretty cool.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/mapper/MapperImplementationGenerator.java
```diff
@@ -117,7 +117,7 @@ protected ClassName getPrincipalTypeName() {
         MethodSpec.constructorBuilder().addModifiers(Modifier.PUBLIC);
 
     GeneratedCodePatterns.addFinalFieldAndConstructorArgument(
-        ClassName.get(MapperContext.class), "context", classContents, constructorContents);
+        ClassName.get(DefaultMapperContext.class), "context", classContents, constructorContents);
```
tomekl007, line 120: Wouldn't it limit extensibility of `MapperContext`? What if some pro user decides to provide own implementation of `MapperContext`? With this change, it seems that implementation is hard-coded now

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DefaultMapperContext.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+public class DefaultMapperContext implements MapperContext {
+
+  private final CqlSession session;
+  private final CqlIdentifier keyspaceId;
+  private final CqlIdentifier tableId;
+  private final ConcurrentMap<Class<? extends NameConverter>, NameConverter> nameConverterCache;
+  private final Map<Object, Object> customState;
```
tomekl007, line 37: Shouldn't the `customState` be `ConcurrentMap`? (same as `nameConverterCache`)?

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DefaultMapperContext.java
```diff
@@ -0,0 +1,117 @@
...
+    this.session = session;
+    this.keyspaceId = keyspaceId;
+    this.tableId = tableId;
+    this.nameConverterCache = nameConverterCache;
+    this.customState = customState;
+  }
+
+  public DefaultMapperContext withKeyspaceAndTable(
+      @Nullable CqlIdentifier newKeyspaceId, @Nullable CqlIdentifier newTableId) {
+    return (Objects.equals(newKeyspaceId, this.keyspaceId)
```
tomekl007, line 64: What is the reason for that logic? (returning `this`)
Do we just want to use `withKeyspaceAndTable` to achieve `computeIfAbsent` semantic?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/mapper/MapperImplementationGenerator.java
olim7t: `MapperContext` instances are only created from generated code. In order to plug your own context implementation, you'd have to write your own processor extension, I think this is beyond where even very advanced users are willing to go.

Also the "custom state" feature introduced in this PR provides some level of customization of the context.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DefaultMapperContext.java
olim7t: It's immutable, you can only add entries from the builder.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DefaultMapperContext.java
olim7t: `DefaultMapperContext` is immutable, calling `withKeyspaceAndTable` returns a new instance.

It's more convenient that way, because every DAO factory method uses its own keyspace/table combination, and those are called concurrently.

The only part that is not mutable is the name converter cache, but name converters are global to a mapper instance.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/QueryProviderIT.java
```diff
@@ -107,12 +119,14 @@ public void should_invoke_query_provider() {
...
   public static class FindSliceProvider {
     private final CqlSession session;
+    private final AtomicInteger executionCount;
     private final EntityHelper<SensorReading> sensorReadingHelper;
     private final Select selectStart;
 
     public FindSliceProvider(
         MapperContext context, EntityHelper<SensorReading> sensorReadingHelper) {
       this.session = context.getSession();
+      this.executionCount = ((AtomicInteger) context.getCustomState().get("executionCount"));
```
tomekl007, line 131: maybe we should add a method on the context that is retrieving state by a key a doing casting to hide that from a client?
`context.getCustomStateByKey("executionCount", AtomicInteger.class)`
it may simplify tests because you do not need to mock the whole custom state but the only value returned by the key

tomekl007 (approved):

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MapperContext.java
```diff
@@ -18,87 +18,48 @@
...
-    }
-  }
+  /**
+   * Retrieves any custom state that was set while building the mapper with {@link
+   * MapperBuilder#withCustomState(Object, Object)}.
+   *
+   * <p>The returned map is immutable. If no state was set on the builder, it will be empty.
+   */
+  @NonNull
+  Map<Object, Object> getCustomState();
```
adutra, line 64: Not sure we need to expose the entire map. The typical API for context objects is `context.getAttribute(String name)` and `context.setAttribute(String name, Object value)`, and the underlying map implementation is completely hidden. E.g. Jackson uses this convention in its `DeserializationContext`.

tolbertam (approved):
:+1:

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MapperContext.java
olim7t: But then you lose the ability to iterate keys. As I said this feature should be rarely needed, let's not overthink it.

PR #1244, opened 2019-05-27
 Log query warnings before setting the final result

adutra:
The commits in this PR should not be squashed together.

Motivation:

Currently the callback result is set _before_ the query warnings are
logged. As a consequence, client code dependent on the callback's
completion may get executed before the warnings are logged.

Modification:

Set the callback result after the query warnings are logged.

Result:

Client code is guaranteed to be executed after the query warnings are
logged.

### DIFF
--- a/driver-core/src/test/java/com/datastax/driver/core/WarningsTest.java
+++ b/driver-core/src/test/java/com/datastax/driver/core/WarningsTest.java
@@ -34,8 +34,7 @@ public class WarningsTest extends CCMTestsSupport {
   @CassandraVersion("2.2.0")
   public void should_expose_warnings_on_execution_info() throws Exception {
     // the default batch size warn threshold is 5 * 1024 bytes, but after CASSANDRA-10876 there must
-    // be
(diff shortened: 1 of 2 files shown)


driver-core/src/test/java/com/datastax/driver/core/WarningsTest.java
```diff
@@ -34,8 +34,7 @@ public void onTestContextInitialized() {
   @CassandraVersion("2.2.0")
   public void should_expose_warnings_on_execution_info() throws Exception {
     // the default batch size warn threshold is 5 * 1024 bytes, but after CASSANDRA-10876 there must
-    // be
-    // multiple mutations in a batch to trigger this warning so the batch includes 2 different
+    // be multiple mutations in a batch to trigger this warning so the batch includes 2 different
```
tolbertam, line 37: :+1: this is probably a relic of running google-java-format initially.  Since multi-line comment format isn't used here, it doesn't know to correlate the following lines.

tolbertam (approved):
:+1: lgtm!

PR #1245, opened 2019-05-28

tomekl007:


integration-tests/src/test/java/com/datastax/oss/driver/mapper/SchemaValidationIT.java
```diff
@@ -0,0 +1,416 @@
...
+        .hasStackTraceContaining("Undefined column name description_with_incorrect_name");
+  }
+
+  @Test
+  public void should_not_throw_on_table_with_properly_mapped_udt_field() {
+    assertThatCode(() -> mapper.productDao(sessionRule.keyspace())).doesNotThrowAnyException();
+  }
+
+  @Test
+  public void should_throw_when_use_not_properly_mapped_entity_with_udt() {
```
tomekl007, line 131: @olim7t 
Concerning: 
> how do we know which entities map to tables vs UDTs?

This is a test that validates that UDT schema validation is working. The logic for UDT handling is here:
https://github.com/datastax/java-driver/pull/1245/files#diff-0a9bd90f05027e85a6a6a029c5994393R153

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
```diff
@@ -0,0 +1,185 @@
...
+            .addAnnotation(Override.class)
+            .addModifiers(Modifier.PUBLIC)
+            .returns(TypeName.VOID);
+
+    // Handle case where keyspace = null. In such case we cannot infer and validate schema for table
+    // or udt
+    methodBuilder.beginControlFlow(
+        "if (!context.getSession().getMetadata().getKeyspace(context.getKeyspaceId()).isPresent())");
+    methodBuilder.addStatement(
+        "return"); // todo maybe we should do log.warn("we cannot validate schema because keyspace
```
tomekl007, line 59: in `QueryIT` we have tests that don't pass keyspace to `@Dao`. In such example `context.getKespaceId()` returns `null` and we are not able to infer tables or udt schema.
This if guards against that, but maybe we should generate some `log.warning` that because keyspace was not known as the schema validation for that `@Dao` cannot work wdyt?

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Dao.java
```diff
@@ -56,4 +56,13 @@
...
-public @interface Dao {}
+public @interface Dao {
+
+  /**
+   * When the new instance of a class annotated with {@code @Dao} is created an automatic check for
+   * schema validation is performed. It verifies if all {@code @Dao} entity fields are present in
+   * CQL table. If not the exception is thrown. This check has startup overhead so once your app is
+   * stable you may want to disable it. The schema Validation check is enabled by default.
+   */
+  boolean enableEntitySchemaValidation() default true;
```
tomekl007, line 67: >The mapper builder could have an option to turn this off (for faster startup once the app is stable).

To see that property working please take a look at the test that using this `@Dao`:
https://github.com/datastax/java-driver/pull/1245/files#diff-af276dd4cc66aa56a5f9c0dcf2d8d611R180

The implementation is here:
https://github.com/datastax/java-driver/pull/1245/files#diff-1d89fc55d9834b473028aaba93c44dc9R255

tomekl007 (commented):
Last pending thing from the ticket description:

>if we do this for every entity helper in every DAO, there could be duplicates. Maybe we could have a mapper-level cache to remember which keyspace/table/entity combinations we've already checked.

Could you please elaborate on that more? What duplicates do you have on mind?

olim7t (commented):
I've rescheduled this ticket to 4.2.0. I think it's not a must-have for the initial mapper release, and we could benefit from a few more rounds of improvements:

* for the case where `keyspace == null`, try `session.getKeyspace()`
* enabling schema validation through an annotation is a bit too static. It would probably be more flexible with a runtime parameter:
    ```java
    new InventoryMapperBuilder(session).setSchemaValidation(true).build()
    ```
* trying to match a table and then falling back to a UDT is a good default. But maybe we could also have an optional annotation to provide hints to the mapper (keeping the name generic because other hints might be useful in the future):
    ```java
    @Entity
    @SchemaHint(targetElement = TABLE)
    public class Product { ... }
    ```
* we could also validate the partition key and clustering columns
* similarly, I think we have enough information to validate the types: we know the Java type from `EntityDefinition`, and the CQL type at runtime from the schema metadata. So we could call `codecRegistry.codecFor` at runtime, and fail if it returns no codec.
* we recently introduced an `EntityHelperBase` parent class. There's probably opportunities to move some code there instead of generating it, for example methods such as:
    ```java
    void validateColumnExists(TableMetadata table, String columnName, GenericType<?> javaType)
    ```
* 👍 for reporting all missing columns instead of failing on the first one. We could extend this to all the entity helpers of a given DAO, this way we validate the DAO's complete schema in one go.

olim7t:
This got closed because I deleted the `java2078` branch, reopening as #1279.

PR #1246, opened 2019-05-29
JAVA-2265: Configure NullSavingStrategy at the DAO level

tomekl007:
…or Update annotation

### DIFF
--- a/manual/mapper/daos/null_saving/README.md
+++ b/manual/mapper/daos/null_saving/README.md
@@ -1,18 +1,11 @@
-## Null saving strategies
+## Null saving strategy
 
-The [@Insert](../insert/), [@Query](../query/), [@SetEntity](../setentity/) and
-[@Update](../update/) annotations allow you to choose how null entity properties should be handled:
-
-```java
-import static com.datastax.oss.driver.api.mapper.entity.saving.NullSavingStrategy.SET_TO_NULL;
-
-@Update(nullSavingStrategy = SET_TO_NULL)
-void update(Product product);
-```
+The null saving strategy controls how null entity properties are handled when writing to the
+database. It can be configured either for each method, or globally at the DAO level.
 
 Two strategies are available:
 
-* `DO_NOT_SET`: the mapper won't call the corresponding setter on the [BoundStatement]. The
+* [DO_NOT_SET]: the mapper won't call the corresponding setter on the [BoundStatement]. The
   generated code looks approximately like this:
 
     ```java
@@ -29,16 +22,80 @@ Two strategies are available:
     strategy with a lower Cassandra version, the mapper will throw an `IllegalArgumentException`
     when you try to build the corresponding DAO.
 
-* `SET_TO_NULL`: the mapper will always call the setter, even with a null value. The generated code
+* [SET_TO_NULL]: the mapper will always call the setter, even with a null value. The generated code
   looks approximately like this:
   
     ```java
     // Called even if entity.getDescription() == null
     boundStatement = boundStatement.setString("description", entity.getDescription());
     ```
-    
-If you don't specify an explicit strategy for a method, the mapper defaults to `SET_TO_NULL`.
 
-[BoundStatement]: https://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/core/cql/BoundStatement.html
+### Method level
+
+Specify `nullSavingStrategy` on the method annotation:
+
+```java
+import static com.datastax.oss.driver.api.mapper.entity.saving.NullSavingStrategy.SET_TO_NULL;
+
+@Update(nullSavingStrategy = SET_TO_NULL)
+void update(Product product);
+```
+
+This applies to [@Insert](../insert/), [@Query](../query/), [@SetEntity](../setentity/) and
+[@Update](../update/) (other method types don't need it since they don't write data).
+
+### DAO level
+
+Annotate your [DAO](../../daos/) interface with [@DefaultNullSavingStrategy]. Any method that does
+not explicitly define its strategy inherits the DAO-level one:
+
+```java
+@Dao
+@DefaultNullSavingStrategy(SET_TO_NULL)
+public interface ProductDao {
+
+  @Insert
+  void insert(Product product); // inherits SET_TO_NULL
+
+  @Update(nullSavingStrategy = DO_NOT_SET)
+  void update(Product product); // uses DO_NOT_SET
+}
+```
+
+If you don't define a DAO-level default, any method that does not declare its own value defaults to
+[DO_NOT_SET]:
+
+```java
+@Dao
+public interface ProductDao {
+
+  @Insert
+  void insert(Product product); // defaults to DO_NOT_SET
+}
+```
+
+Note that you can use inheritance to set a common default for all your DAOs:
+
+```java
+@DefaultNullSavingStrategy(SET_TO_NULL)
+public interface InventoryDao {}
+
+@Dao
+public interface ProductDao extends InventoryDao {
+  @Insert
+  void insert(Product product); // inherits SET_TO_NULL
+}
+
+@Dao
+public interface UserDao extends InventoryDao {
+  @Insert
+  void insert(User user); // inherits SET_TO_NULL
+}
+```
+
+[@DefaultNullSavingStrategy]: https://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/mapper/annotations/DefaultNullSavingStrategy.html
+[BoundStatement]:             https://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/core/cql/BoundStatement.html
+[DO_NOT_SET]:                 https://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/mapper/entity/saving/NullSavingStrategy.html#DO_NOT_SET
+[SET_TO_NULL]:                https://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/mapper/entity/saving/NullSavingStrategy.html#SET_TO_NULL
       
 [CASSANDRA-7304]: https://issues.apache.org/jira/browse/CASSANDRA-7304
--- a/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
+++ b/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
@@ -17,10 +17,7 @@ package com.datastax.oss.driver.internal.mapper.processor.dao;
 
 import com.datastax.oss.driver.api.core.cql.PreparedStatement;
 import com.datastax.oss.driver.api.mapper.MapperContext;
-import com.datastax.oss.driver.api.mapper.annotations.Insert;
-import com.datastax.oss.driver.api.mapper.annotations.Query;
-import com.datastax.oss.driver.api.mapper.annotations.SetEntity;
-import com.datastax.oss.driver.api.mapper.annotations.Update;
+import com.datastax.oss.driver.api.mapper.annotations.DefaultNullSavingStrategy;
 import com.datastax.oss.driver.api.mapper.entity.saving.NullSavingStrategy;
 import com.datastax.oss.driver.internal.core.util.concurrent.BlockingOperation;
 import com.datastax.oss.driver.internal.core.util.concurrent.CompletableFutures;
@@ -31,7 +28,6 @@ import com.datastax.oss.driver.internal.mapper.processor.ProcessorContext;
 import com.datastax.oss.driver.internal.mapper.processor.SingleFileCodeGenerator;
 import com.datastax.oss.driver.internal.mapper.processor.util.NameIndex;
 import com.datastax.oss.driver.internal.mapper.processor.util.generation.GenericTypeConstantGenerator;
-import com.datastax.oss.protocol.internal.ProtocolConstants;
 import com.squareup.javapoet.ClassName;
 import com.squareup.javapoet.CodeBlock;
 import com.squareup.javapoet.FieldSpec;
@@ -50,6 +46,7 @@ import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.function.BiConsumer;
+import java.util.stream.Collectors;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
 import javax.lang.model.element.ExecutableElement;
@@ -71,11 +68,13 @@ public class DaoImplementationGenerator extends SingleFileCodeGenerator
   private final Map<ClassName, String> entityHelperFields = new LinkedHashMap<>();
   private final List<GeneratedPreparedStatement> preparedStatements = new ArrayList<>();
   private final List<GeneratedQueryProvider> queryProviders = new ArrayList<>();
+  private final NullSavingStrategyValidation nullSavingStrategyValidation;
 
   public DaoImplementationGenerator(TypeElement interfaceElement, ProcessorContext context) {
     super(context);
     this.interfaceElement = interfaceElement;
     implementationName = GeneratedNames.daoImplementation(interfaceElement);
+    nullSavingStrategyValidation = new NullSavingStrategyValidation(context);
   }
 
   @Override
@@ -124,6 +123,12 @@ public class DaoImplementationGenerator extends SingleFileCodeGenerator
     }
     queryProviders.add(new GeneratedQueryProvider(fieldName, providerClass, entityHelperNames));
     return fieldName;
+  }
+
+  @Override
+  public Optional<NullSavingStrategy> getNullSavingStrategy() {
+    return Optional.ofNullable(interfaceElement.getAnnotation(DefaultNullSavingStrategy.class))
+        .map(DefaultNullSavingStrategy::value);
   }
 
   @Override
@@ -307,61 +312,10 @@ public class DaoImplementationGenerator extends SingleFileCodeGenerator
   }
 
   private void generateProtocolVersionCheck(MethodSpec.Builder builder) {
-    boolean hasDoNotSetStatement =
-        preparedStatements
-            .stream()
-            .map(v -> v.methodElement)
-            .anyMatch(
-                v ->
-                    updateHasDoNotSet(v)
-                        || insertHasDoNotSet(v)
-                        || setEntityHasDoNotSet(v)
-                        || queryHasDoNotSet(v));
-
-    if (hasDoNotSetStatement) {
-      builder
-          .beginControlFlow(
-              "if (context.getSession().getContext().getProtocolVersion().getCode() <= $T.Version.V3)",
-              ProtocolConstants.class)
-          .addStatement(
-              "throw new IllegalArgumentException($S)",
-              "You cannot use NullSavingStrategy.DO_NOT_SET for protocol version V3.")
-          .endControlFlow()
-          .build();
-    }
-  }
-
-  private boolean queryHasDoNotSet(ExecutableElement v) {
-    Query annotation = v.getAnnotation(Query.class);
-    if (annotation != null) {
-      return annotation.nullSavingStrategy() == NullSavingStrategy.DO_NOT_SET;
-    }
-    return false;
-  }
-
-  private boolean setEntityHasDoNotSet(ExecutableElement v) {
-    SetEntity annotation = v.getAnnotation(SetEntity.class);
-    if (annotation != null) {
-      return annotation.nullSavingStrategy() == NullSavingStrategy.DO_NOT_SET;
-    }
-    return false;
-  }
-
-  private boolean insertHasDoNotSet(ExecutableElement v) {
-    Insert annotation = v.getAnnotation(Insert.class);
-    if (annotation != null) {
-      return annotation.nullSavingStrategy() == NullSavingStrategy.DO_NOT_SET;
-    } else {
-      return false;
-    }
-  }
-
-  private boolean updateHasDoNotSet(ExecutableElement v) {
-    Update annotation = v.getAnnotation(Update.class);
-    if (annotation != null) {
-      return annotation.nullSavingStrategy() == NullSavingStrategy.DO_NOT_SET;
-    } else {
-      return false;
+    List<ExecutableElement> methodElements =
+        preparedStatements.stream().map(v -> v.methodElement).collect(Collectors.toList());
+    if (nullSavingStrategyValidation.hasDoNotSetOnAnyLevel(methodElements, interfaceElement)) {
+      builder.addStatement("throwIfProtocolVersionV3(context)");
     }
   }
 
--- a/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoInsertMethodGenerator.java
+++ b/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoInsertMethodGenerator.java
@@ -49,12 +49,14 @@ public class DaoInsertMethodGenerator extends DaoMethodGenerator {
           FUTURE_OF_ENTITY,
           OPTIONAL_ENTITY,
           FUTURE_OF_OPTIONAL_ENTITY);
+  private final NullSavingStrategyValidation nullSavingStrategyValidation;
 
   public DaoInsertMethodGenerator(
       ExecutableElement methodElement,
       DaoImplementationSharedCode enclosingClass,
       ProcessorContext context) {
     super(methodElement, enclosingClass, context);
+    nullSavingStrategyValidation = new NullSavingStrategyValidation(context);
   }
 
   @Override
@@ -126,8 +128,11 @@ public class DaoInsertMethodGenerator extends DaoMethodGenerator {
           statementAttributesParam.getSimpleName().toString());
     }
     String entityParameterName = parameters.get(0).getSimpleName().toString();
+
     NullSavingStrategy nullSavingStrategy =
-        methodElement.getAnnotation(Insert.class).nullSavingStrategy();
+        nullSavingStrategyValidation.getNullSavingStrategy(
+            Insert.class, Insert::nullSavingStrategy, methodElement, enclosingClass);
+
     insertBuilder.addStatement(
         "$1L.set($2L, boundStatementBuilder, $3T.$4L)",
         helperFieldName,
--- a/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/NullSavingStrategyValidation.java
+++ b/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/NullSavingStrategyValidation.java
@@ -1,0 +1,235 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.internal.mapper.processor.dao;
+
+import com.datastax.oss.driver.api.mapper.annotations.DefaultNullSavingStrategy;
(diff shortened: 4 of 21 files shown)


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
```diff
@@ -214,6 +218,36 @@ protected void maybeAddTimestamp(String timestamp, MethodSpec.Builder methodBuil
     }
   }
 
+  protected boolean nullSavingStrategyExplicitlySet(
+      ExecutableElement methodElement, String annotationCanonicalName) {
```
tomekl007: This is code responsible for finding out if the `nullSavingStrategy` value was explicitly set on the `@Update` annotation using `AnnotationMirror`. It looks a bit complex, @olim7t do you know if there is a way to create `ExecutableElement` using java-poet for the purpose of this test?

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/DefaultNullSavingStrategy.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.datastax.oss.driver.api.mapper.entity.saving.NullSavingStrategy;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.CLASS)
+public @interface DefaultNullSavingStrategy {
```
tomekl007, line 57: I think If we choose to add new annotation `DefaultNullSavingStrategy` this annotation should have `nullSavingStrategy` without default. If someone is adding that at the `@Dao` level it means that he must specify `nullSavingStrategy` that will apply to inherited strategy for methods: `@Update`, `@Insert`, `@SetEntity`, `@Query`.

tomekl007 (commented):
Implemented only for `@Update` and waiting for feedback.
`@SetEntity`, `@Insert` and `Query` level handing todo (it is fairly simple: code from line https://github.com/datastax/java-driver/pull/1246/files#diff-25ba2cdaf8c214d0d4a6f05f1c7922acR129 to https://github.com/datastax/java-driver/pull/1246/files#diff-25ba2cdaf8c214d0d4a6f05f1c7922acR135)

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/DefaultNullSavingStrategy.java
olim7t: I agree with no default.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
olim7t: JavaPoet only generates strings. The only way to create an `ExecutableElement` for a test is either through mocking or with a full compiler execution I think.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
```diff
@@ -214,6 +218,36 @@ protected void maybeAddTimestamp(String timestamp, MethodSpec.Builder methodBuil
...
+        .filter(
+            v -> {
+              // Find only for Update annotation
+              DeclaredType annotationType = v.getAnnotationType();
+              if (annotationType instanceof Type.ClassType) {
+                return ((Type.ClassType) annotationType)
+                    .tsym
+                    .getQualifiedName()
+                    .toString()
+                    .equals(annotationCanonicalName);
```
olim7t: You can compare the type with a `Class` directly:
```java
context.getClassUtils().isSame(getAnnotationType(), Update.class)
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
```diff
@@ -214,6 +218,36 @@ protected void maybeAddTimestamp(String timestamp, MethodSpec.Builder methodBuil
...
+              } else {
+                return false;
+              }
+            })
+        .flatMap(v -> v.getElementValues().values().stream())
+        .filter(v -> v instanceof Attribute.Enum)
```
olim7t: From `getElementValues()` javadocs:
> Only those elements with values explicitly present in the annotation are included, not those that are implicitly assuming their default values.

So we just need to iterate `getElementValues().keySet()` and find out if there is an element such that `key.getSimpleName().contentEquals("nullSavingStrategy")`.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
```diff
@@ -214,6 +218,36 @@ protected void maybeAddTimestamp(String timestamp, MethodSpec.Builder methodBuil
     }
   }
 
+  protected boolean nullSavingStrategyExplicitlySet(
+      ExecutableElement methodElement, String annotationCanonicalName) {
+    return !methodElement
+        .getAnnotationMirrors()
+        .stream()
```
olim7t: I think the Stream API really makes things worse here (and to be honest that's true most of the time -- that's why it's discouraged in the contribution guidelines).

There's an imperative version for a similar case [here](https://github.com/datastax/java-driver/blob/java2078/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoQueryProviderMethodGenerator.java#L84). It's not pretty either, but I find it way more readable.

tomekl007: I've changed (https://github.com/datastax/java-driver/pull/1246/commits/75a978baf7118914f2907e47e89a46ead8e7561f) it to imperative way but I am not sure if this is more readable that stream versions after simplification:
```
  return methodElement
        .getAnnotationMirrors()
        .stream()
        .filter(v -> classUtils.isSame(v.getAnnotationType(), javaClass))
        .flatMap(v -> v.getElementValues().keySet().stream())
        .anyMatch(v -> v.getSimpleName().contentEquals("nullSavingStrategy"));
```

tolbertam:
Sorry for delay in review, this will be my first priority tomorrow :+1:

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -320,56 +328,59 @@ private void generateProtocolVersionCheck(MethodSpec.Builder builder) {
...
+  private void addVersionCheckToBuilder(MethodSpec.Builder builder) {
+    builder
+        .beginControlFlow(
+            "if (context.getSession().getContext().getProtocolVersion().getCode() <= $T.Version.V3)",
+            ProtocolConstants.class)
+        .addStatement(
+            "throw new IllegalArgumentException($S)",
+            "You cannot use NullSavingStrategy.DO_NOT_SET for protocol version V3.")
+        .endControlFlow()
+        .build();
```
olim7t: You could move this to a static method on `DaoBase`, so that you only have to generate the method call.

On the other hand the code is still relatively simple so it's not a big deal to generate it. Your call.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/DefaultNullSavingStrategy.java
```diff
@@ -0,0 +1,46 @@
...
+   * of: {@link Update#nullSavingStrategy()}, {@link Insert#nullSavingStrategy()}, {@link
+   * SetEntity#nullSavingStrategy()} {@link Query#nullSavingStrategy()} explicitly the {@code
+   * DefaultNullSavingStrategy#nullSavingStrategy()} will be taken.
+   *
+   * <p>If you do not specify {@code DefaultNullSavingStrategy} at the {@link Dao} level the default
+   * or explicitly user set values from one of: {@link Update#nullSavingStrategy()}, {@link
+   * Insert#nullSavingStrategy()}, {@link SetEntity#nullSavingStrategy()} {@link
+   * Query#nullSavingStrategy()} will be taken
+   */
+  NullSavingStrategy nullSavingStrategy();
```
olim7t: I think we don't need a dedicated parameter name here, since it's the only one and it's pretty obvious from the name of the annotation. So I would call this method `value()` so that it can be written directly like this:

```java
@DefaultNullSavingStrategy(SET_TO_NULL)
```

manual/mapper/daos/defaultnullsavingstrategy/README.md
```diff
@@ -0,0 +1,33 @@
+## DefaultNullSavingStrategy
+
+If you don't specify `nullSavingStrategy` parameter on the method level you can suppress 
+defaults by using `@DefaultNullSavingStrategy` annotation.
```
olim7t: Can you make `@DefaultNullSavingStrategy` a link to the Javadocs?

manual/mapper/daos/null_saving/README.md
```diff
@@ -39,6 +39,11 @@ Two strategies are available:
     
 If you don't specify an explicit strategy for a method, the mapper defaults to `SET_TO_NULL`.
 
+## Default Dao level null saving strategy
+
+If you don't specify `nullSavingStrategy` parameter on the method level you can suppress defaults by using
+`@DefaultNullSavingStrategy` annotation. See [@DefaultNullSavingStrategy](../defaultnullsavingstrategy/) for more details.
```
olim7t: I don't think a separate page is necessary here, I would inline it in this section.
It fits in nicely since it relates to the main topic.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -71,11 +69,13 @@
   private final Map<ClassName, String> entityHelperFields = new LinkedHashMap<>();
   private final List<GeneratedPreparedStatement> preparedStatements = new ArrayList<>();
   private final List<GeneratedQueryProvider> queryProviders = new ArrayList<>();
+  private final NullSavingStrategyValidation nullSavingStrategyValidation;
 
   public DaoImplementationGenerator(TypeElement interfaceElement, ProcessorContext context) {
     super(context);
     this.interfaceElement = interfaceElement;
     implementationName = GeneratedNames.daoImplementation(interfaceElement);
+    nullSavingStrategyValidation = new NullSavingStrategyValidation(context);
```
olim7t, line 77: 👍 Good idea to extract the boilerplate into a separate component

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoInsertMethodGenerator.java
```diff
@@ -127,8 +129,21 @@ public DaoInsertMethodGenerator(
 
     List<? extends VariableElement> parameters = methodElement.getParameters();
     String entityParameterName = parameters.get(0).getSimpleName().toString();
-    NullSavingStrategy nullSavingStrategy =
-        methodElement.getAnnotation(Insert.class).nullSavingStrategy();
+
+    Insert annotation = methodElement.getAnnotation(Insert.class);
+    NullSavingStrategy nullSavingStrategy;
+
+    Optional<NullSavingStrategy> defaultNullSavingStrategy = enclosingClass.getNullSavingStrategy();
```
olim7t: Maybe calling this `daoNullSavingStrategy` would make the code below a bit easier to read?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/NullSavingStrategyValidation.java
```diff
@@ -0,0 +1,246 @@
...
+import com.datastax.oss.driver.internal.mapper.processor.ProcessorContext;
+import com.datastax.oss.driver.internal.mapper.processor.util.Classes;
+import java.util.List;
+import java.util.Optional;
+import javax.lang.model.element.AnnotationMirror;
+import javax.lang.model.element.Element;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.TypeElement;
+
+class NullSavingStrategyValidation {
```
olim7t: We might want to make this class (and its two package-private methods) public.
We're not committing to the API because it's in an internal package, and it could come in handy if we extend the generators or create new ones from the DSE mapper.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/NullSavingStrategyValidation.java
```diff
@@ -0,0 +1,246 @@
...
+      return true;
+      // if DAO level DO_NOT_SET check if all underlying override it explicitly to SET_TO_NULL
+    } else if (hasDaoLevelDoNotSetStatement(daoLevelInterface)
+        && !hasSetToNullStatementSetExplicitlyForAll) {
+      return true;
+      // if DAO level annotation do not present, check method level strategy
+      // (including the default one)
+    } else {
+      return hasDaoLevelUnset(daoLevelInterface) && hasDoNotSetStatement;
+    }
```
olim7t, line 98: This is a bit hard to follow, but I think it's more because of the inherent complexity, I'm not sure I could do better :)

Maybe a few naming suggestions (as you see fit):
```
hasDoNotSetStatement => anyMethodHasOrDefaultsToDoNotSet
hasDoNotSetStatementSetExplicitly => anyMethodHasDoNotSetExplicitly
hasSetToNullStatementSetExplicitlyForAll => allMethodsHaveSetToNull (no need to say "explicitly" here, SET_TO_NULL is always explicit)

hasDaoLevelSetToNullStatement => daoHasSetToNull
hasDaoLevelDoNotSetStatement => daoHasDoNotSet
hasDaoLevelUnset => daoIsNotAnnotated
```

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/DefaultNullSavingStrategy.java
```diff
@@ -0,0 +1,46 @@
...
+   * <p>If you specify {@link DefaultNullSavingStrategy} at the {@link Dao} level and do not set one
+   * of: {@link Update#nullSavingStrategy()}, {@link Insert#nullSavingStrategy()}, {@link
+   * SetEntity#nullSavingStrategy()} {@link Query#nullSavingStrategy()} explicitly the {@code
+   * DefaultNullSavingStrategy#nullSavingStrategy()} will be taken.
+   *
+   * <p>If you do not specify {@code DefaultNullSavingStrategy} at the {@link Dao} level the default
+   * or explicitly user set values from one of: {@link Update#nullSavingStrategy()}, {@link
+   * Insert#nullSavingStrategy()}, {@link SetEntity#nullSavingStrategy()} {@link
+   * Query#nullSavingStrategy()} will be taken
+   */
```
olim7t: I would move this as the main javadoc for the annotation. Also a few suggestions to reorganize the explanations:
```java
/**
 * Annotates a {@link Dao} interface to define a default {@link NullSavingStrategy}, that will apply
 * to all methods that don't explicitly declare one.
 *
 * <p>For example, given this interface:
 *
 * <pre>
 * &#64;Dao
 * &#64;DefaultNullSavingStrategy(SET_TO_NULL)
 * public interface ProductDao {
 *
 *   &#64;Insert
 *   void insert(Product product);
 *
 *   &#64;Update(nullSavingStrategy = DO_NOT_SET)
 *   void update(Product product);
 * }
 * </pre>
 *
 * <ul>
 *   <li>{@code insert(Product)} will use {@link NullSavingStrategy#SET_TO_NULL SET_TO_NULL}
 *       (inherited from the DAO's default).
 *   <li>{@code update(Product)} will use {@link NullSavingStrategy#DO_NOT_SET DO_NOT_SET}.
 * </ul>
 *
 * If the DAO interface isn't annotated with {@link DefaultNullSavingStrategy}, any method that does
 * not declare its own value defaults to {@link NullSavingStrategy#DO_NOT_SET DO_NOT_SET}.
 *
 * <p>Note that null saving strategies are only relevant for {@link Update}, {@link Insert}, {@link
 * Query} and {@link SetEntity} methods.
 */
```

olim7t (changes requested):

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/DefaultNullSavingStrategy.java
tolbertam, line 57: What do you think about naming this simply `@NullSavingStrategy`?  Default seems kind of implicit, as to me it is intuitive to expect that if you were to also define `nullSavingStrategy` at the method-annotation level that it would override the interface level annotation.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/NullSavingStrategyValidation.java
```diff
@@ -0,0 +1,246 @@
...
+      return annotation.nullSavingStrategy() == NullSavingStrategy.SET_TO_NULL;
+    }
+    return false;
+  }
+
+  private boolean hasDaoLevelUnset(TypeElement interfaceElement) {
+    return interfaceElement.getAnnotation(DefaultNullSavingStrategy.class) == null;
+  }
+
+  private boolean queryHasDoNotSet(ExecutableElement v, boolean explicitSet) {
```
tolbertam, line 123: There is a lot of boilerplate here between `queryHasDoNotSet`, `setEntityHasDoNotSet`, `insertHasDoNotSet` and `updateHasDoNotSet` that could probably be factored into a common method, for example you could do something like this:

```java
  private <A extends Annotation> boolean hasDoNotSet(Class<A> clazz,
                              Function<A, NullSavingStrategy> extractor,
                              ExecutableElement v,
                              boolean explicitSet) {
    A annotation = v.getAnnotation(clazz);
    if (annotation != null) {
      NullSavingStrategy strategy = extractor.apply(annotation);
      if (explicitSet) {
        return strategy == NullSavingStrategy.DO_NOT_SET
          && nullSavingStrategyExplicitlySet(v, clazz);
      } else {
        return strategy == NullSavingStrategy.DO_NOT_SET;
      }
    }
    return false;
  }

  private boolean queryHasDoNotSet(ExecutableElement v, boolean explicitSet) {
    return hasDoNotSet(Query.class, Query::nullSavingStrategy, v, explicitSet);
  }
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/NullSavingStrategyValidation.java
```diff
@@ -0,0 +1,246 @@
...
+            && nullSavingStrategyExplicitlySet(v, Update.class);
+      } else {
+        return annotation.nullSavingStrategy() == NullSavingStrategy.DO_NOT_SET;
+      }
+    } else {
+      return false;
+    }
+  }
+
+  private boolean queryHasSetToNullExplicitly(ExecutableElement v) {
```
tolbertam, line 157: same here, these `HasSetToNullExplicitly` methods could be factored into a common method.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/NullSavingStrategyValidation.java
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  private boolean isOperationWithNullSavingStrategy(ExecutableElement v) {
+    return v.getAnnotation(Update.class) != null
+        || v.getAnnotation(Insert.class) != null
+        || v.getAnnotation(SetEntity.class) != null
+        || v.getAnnotation(Query.class) != null;
+  }
+
+  boolean nullSavingStrategyExplicitlySet(Element methodElement, Class<?> javaClass) {
```
tolbertam: If we were to remove `default NullSavingStrategy.DO_NOT_SET` from all annotations so they are null instead couldn't this be more effective as all you would have to do is null check?

manual/mapper/daos/defaultnullsavingstrategy/README.md
```diff
@@ -0,0 +1,33 @@
+## DefaultNullSavingStrategy
+
+If you don't specify `nullSavingStrategy` parameter on the method level you can suppress
```
tolbertam: instead of `you can suppress defaults`, I would recommend `you can adjust the default behavior`

manual/mapper/daos/defaultnullsavingstrategy/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Strategies when using `@DefaultNullSavingStrategy` with method level `nullSavingStrategy` on annotations: 
+[@Insert](../insert/), [@Query](../query/), [@SetEntity](../setentity/) or
+[@Update](../update/):
+
+* If you specify `@DefaultNullSavingStrategy` at the [@Dao] level and set `nullSavingStrategy` parameter on one of the 
+  [@Update](../update/), [@Insert](../insert/), [@Query](../query/) or [@SetEntity](../setentity./)
+  explicitly, the method level annotation value will be taken.
+
+* If you specify `@DefaultNullSavingStrategy` at the [@Dao] level and do not set `nullSavingStrategy` parameter on one of the
```
tolbertam: These next two paragraphs don't add much as I think it should be apparent that if you only use one of the ways to configure null saving, the one that is used is used.

manual/mapper/daos/defaultnullsavingstrategy/README.md
```diff
@@ -0,0 +1,33 @@
...
+```java
+@Dao
+@DefaultNullSavingStrategy(nullSavingStrategy = NullSavingStrategy.SET_TO_NULL)
+public interface DefaultNullStrategyDao {
+  // dao methods
+}
+
+```
+
+Strategies when using `@DefaultNullSavingStrategy` with method level `nullSavingStrategy` on annotations:
```
tolbertam: If you agree and remove the two paragraphs below, you don't need this first line either.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoInsertMethodGenerator.java
```diff
@@ -127,8 +129,21 @@ public DaoInsertMethodGenerator(
 
     List<? extends VariableElement> parameters = methodElement.getParameters();
     String entityParameterName = parameters.get(0).getSimpleName().toString();
-    NullSavingStrategy nullSavingStrategy =
-        methodElement.getAnnotation(Insert.class).nullSavingStrategy();
+
+    Insert annotation = methodElement.getAnnotation(Insert.class);
```
tolbertam: This code is repeated among the next four classes, could probably be refactored similarly to how I describe in `NullSavingStrategyValidation`

tolbertam (changes requested):
Looks really great, lots of good test coverage.  I had some suggestions, but :+1: to the approach.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/NullSavingStrategyValidation.java
olim7t: Removing the default makes the attribute mandatory on the annotation.

tolbertam: ah, right.  That didn't occur to me at the time, in any case, could we just set default to null?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
tomekl007: Yes, I think that this is a good idea to have a "normal" java code where possible (instead of generated). Changed (https://github.com/datastax/java-driver/pull/1246/commits/37aada0af2153961c75d4fb7714b7b42ffc56ee4)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/NullSavingStrategyValidation.java
tomekl007: I was also thinking about it - I had an impression that sometimes we are favoring code readability over DRY but If that is your impression I am +1 on this as well.
Implemented in https://github.com/datastax/java-driver/pull/1246/commits/0e99daa6a413387ae8478a01b5fd88c463edbc4f

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/NullSavingStrategyValidation.java
tomekl007: If we will set it to `null` we will still need to interpret is somehow for a case where 
`@DefaultNullSavingStrategy` on the `@Dao` level is not a set AND the method level `nullSavingStrategy` property for example on `@Update` will not be set. 
In such case that null needs to become `DO_NOT_SET` or `SET_TO_NULL`. I think that it is better to have that default logic embedded within the annotation default:
`NullSavingStrategy nullSavingStrategy() default NullSavingStrategy.DO_NOT_SET;`
instead of adding another case to handle when the value is null.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoInsertMethodGenerator.java
tomekl007: https://github.com/datastax/java-driver/pull/1246/commits/6d1e5e67a568b52db19fbc6a587288a9a593894e

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/NullSavingStrategyValidation.java
tomekl007: I like those names. Changed in https://github.com/datastax/java-driver/pull/1246/commits/7a40e48bbd042518390c5541b72ab67f4d7ae392

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/DefaultNullSavingStrategy.java
tomekl007: I agree with you that `@NullSavingStrategy` in this context should be enough.
The `@DefaultNullSavingStrategy` name was an initial idea of @olim7t - wdyt about changing that name to `@NullSavingStrategy`

manual/mapper/daos/defaultnullsavingstrategy/README.md
tomekl007: The method level annotations have `NullSavingStrategy nullSavingStrategy() default NullSavingStrategy.DO_NOT_SET;` value. When the user is using it will have a default set.
Don't you think that some of our users may think that this unset default can take precedence?
I think that there is a small chance that someone may think like this so maybe it is better to have it explicitly documented?

tolbertam (approved):
:+1: my feedback has been addressed by recent changes.  Only question left is on the naming, but that is just a suggestion and i'm ok with either choice.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/DefaultNullSavingStrategy.java
olim7t: In theory I would agree, but then the annotation would have the same name as the enum. I find this really unfortunate given that one references the other.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/NullSavingStrategyValidation.java
olim7t: Plus `null` is not a valid default for an annotation parameter.

olim7t (approved):

olim7t:
Merged with `@DefaultNullSavingStrategy`: I felt there was a pretty strong case to not clash with the enum name.
I also took the liberty of rearranging the manual page.

PR #1247, opened 2019-05-29
JAVA-2207: Add bulk value assignment to QueryBuilder Insert

olim7t:


query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/insert/OngoingValues.java
```diff
@@ -38,4 +40,33 @@
...
+   *
+   * <p>If the map contains columns that had already been added to this statement, the previous
+   * entries are discarded and the new entries will be added at the end of the list.
+   *
+   * <p>Implementation note: this is a default method only for backward compatibility. The default
+   * implementation calls {@link #value(CqlIdentifier, Term)} in a loop; it should be overridden if
+   * a more efficient alternative exists.
+   */
+  @NonNull
+  default RegularInsert valuesByIds(@NonNull Map<CqlIdentifier, Term> newAssignments) {
```
tolbertam, line 56: Seems like a better alternative when compared to `throw new UnsupportedOperationException` :)

tolbertam (approved):
:+1:

PR #1248, opened 2019-05-30
JAVA-2234: Handle terminated executor when the session is closed twice

olim7t:


core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
```diff
@@ -246,14 +247,31 @@ public SessionMetricUpdater getMetricUpdater() {
...
+      try {
+        RunOrSchedule.on(adminExecutor, action);
+      } catch (RejectedExecutionException e) {
+        // Checking the future is racy, there is still a tiny window that could get us here.
+        LOG.warn(
+            "[{}] Ignoring terminated executor. "
+                + "This generally happens if you close the session multiple times concurrently, "
+                + "and can be safely ignored if the close() call returns normally.",
+            logPrefix,
+            e);
```
olim7t, line 272: I don't think there's any way to cover this case in tests, except maybe brute force but it's a very tiny window.

The second call has to land between `onClose` and `closeFuture.complete` in `onChildrenClosed` (line 500):
```java
      context
          .getNettyOptions()
          .onClose()
/* HERE */
          .addListener(
              f -> {
                if (!f.isSuccess()) {
                  closeFuture.completeExceptionally(f.cause());
                } else {
                  LOG.debug("[{}] Shutdown complete", logPrefix);
                  closeFuture.complete(null);
                }
              });
```

tolbertam: Its been a while since I looked at it, but this could be a good use case for [byteman](https://byteman.jboss.org/).  But this is a pretty straightforward change, so not sure we need to go that far.

I think logging at warn is fine.  I'm not sure how likely it is to manifest, but you could argue using a lower logging level because if this happens, there really isn't anything suspicious you need to look into right?

tolbertam (approved):
:+1:

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
olim7t: You'd need very (un)lucky timing for that second call to try to schedule its task right before the future is completed, so I think it will be very rare (and you have to close twice in the first place).

One reason that I picked WARN level is that it could also happen if someone closes their event loop group manually. Granted, that's an obvious mistake and things probably start to go south way before shutdown if you do that.

PR #1249, opened 2019-05-30
Small change in mapper manual snippet

newkek:
The last code snippet seems to be missing the `inventoryMapper` on which to call `productDao()`


olim7t (approved):

olim7t:
Thanks!

PR #1250, opened 2019-05-30
JAVA-2111: Add HierarchyScanStrategy to configure inheritance behavior

tolbertam:
For [JAVA-2111](https://datastax-oss.atlassian.net/browse/JAVA-2111)

TODO:

- [x] validate that this actually works beyond unit level :)
- [x] Check the `scanAnnotations` field
- [x] Implement in Dao code once #1246 is merged
- [x] Documentation
- [x] Test all remaining DAO query methods

---

Motivation:

The ability to define properties and annotations of Entity-annotated
classes in parent classes or interfaces will make it easier for
implementors to reuse code and enforce behavior across multiple
entities.

In addition, the same applies for Dao-annotated interfaces.  For
example, being able to express a NullSavingStrategy on a shared parent
interface would be desirable for those who wish to share that
configuration among multiple Daos.

Modifications:

Added `@HierarchyScanStrategy` that can be defined anywhere in a Dao or
Entity type hierarchy.  This strategy dictates how to treat annotations
and properties defined in parent classes and interfaces.

By default, all parent interfaces and classes are considered.

Result:

The mapper can now make use of properties and annotations defined in
parent classes and interfaces of Dao and Entity-annotated types.


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java
```diff
@@ -0,0 +1,191 @@
...
+    ImmutableSet.Builder<TypeElement> hierarchy = ImmutableSet.builder();
+    traverseFullHierarchy(hierarchyScanStrategy, typeElement, types, hierarchy::add);
+    return hierarchy.build();
+  }
+
+  private static HierarchyScanStrategy resolveHierarchyScanStrategy(
+      TypeElement classElement, Types types) {
+    // Use the default HierarchyScanStrategy to find the configured HierarchyScanStrategy.
+    // This is done because the default strategy is the most permissive.
+    HierarchyScanStrategy strategy =
```
tolbertam: possible optimization could be done here such that if the strategy is not defined, return the fully traversed hierarchy so we don't have to traverse it again...but it's probably such a cheap operation that is only done for each dao/entity that it might not be worth it.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -72,97 +77,107 @@ public DefaultEntityFactory(ProcessorContext context) {
...
-              .error(
-                  getMethod,
-                  "Duplicate clustering column index: if multiple properties are annotated "
-                      + "with @%s, the annotation must be parameterized with an integer "
-                      + "indicating the position. Found duplicate index %d for %s and %s.",
-                  ClusteringColumn.class.getSimpleName(),
-                  clusteringColumnIndex,
-                  previous.getGetterName(),
-                  property.getGetterName());
+        VariableElement field = findField(typeHierarchy, propertyName, typeMirror);
```
tolbertam, line 124: Might need some logic here to bound the highest type parsed for the field to where the getMethod was defined...not sure if that's worth it or not though.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -72,97 +77,107 @@ public DefaultEntityFactory(ProcessorContext context) {
...
+        HierarchyScanner.resolveTypeHierarchy(classElement, context.getTypeUtils());
+    CqlNameGenerator cqlNameGenerator = buildCqlNameGenerator(typeHierarchy);
+    Set<String> transientProperties = getTransientPropertyNames(typeHierarchy);
 
-    // TODO inherit annotations and properties from superclass / parent interface
-
-    CqlNameGenerator cqlNameGenerator = buildCqlNameGenerator(classElement);
-    Set<String> transientProperties = getTransientPropertyNames(classElement);
-
+    Set<String> encounteredPropertyNames = Sets.newTreeSet();
```
tomekl007: Why do we need `TreeSet`? Is ordering of property names important here?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -179,15 +194,21 @@ public EntityDefinition getDefinition(TypeElement classElement) {
...
-      VariableElement field = (VariableElement) child;
-      if (field.getSimpleName().toString().equals(propertyName)
-          && context.getTypeUtils().isSameType(field.asType(), fieldType)) {
-        return field;
+      for (Element child : classElement.getEnclosedElements()) {
+        if (child.getKind() != ElementKind.FIELD) {
+          continue;
+        }
+        VariableElement field = (VariableElement) child;
+        if (field.getSimpleName().toString().equals(propertyName)
```
tomekl007, line 217: is `getSimpleName` check enough? Isn't there a risk that the same name of the property can be defined for types in different packages? Shouldn't we use ful qualified name?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -179,15 +194,21 @@ public EntityDefinition getDefinition(TypeElement classElement) {
...
+            && context.getTypeUtils().isSameType(field.asType(), fieldType)) {
+          return field;
+        }
       }
     }
     return null;
```
tomekl007, line 223: Maybe we should follow a pattern of returning Optional instead of null? (similar to other annotation processor methods)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -364,11 +391,12 @@ private boolean isTransient(
     return isTransient;
   }
 
-  private Set<String> getTransientPropertyNames(TypeElement classElement) {
-    TransientProperties transientProperties = classElement.getAnnotation(TransientProperties.class);
+  private Set<String> getTransientPropertyNames(Set<TypeElement> typeHierarchy) {
```
tomekl007, line 406: We will need to implement similar logic with a call to `getClassAnnotation` for every annotation that should be inherited, right?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/AnnotationScanner.java
```diff
@@ -0,0 +1,74 @@
...
+      return Optional.of(new ResolvedAnnotation<>(annotation, getMethod));
+    }
+
+    // otherwise navigate the hierarchy until an annotation is found.
+    for (TypeElement typeElement : typeHierarchy) {
+      for (Element child : typeElement.getEnclosedElements()) {
+        Set<Modifier> modifiers = child.getModifiers();
+        if (child.getKind() != ElementKind.METHOD
+            || modifiers.contains(Modifier.STATIC)
+            || modifiers.contains(Modifier.PRIVATE)) {
```
tomekl007, line 55: So if someone by mistake annotates private method we will just ignore that? Maybe some kind of log stating that we are not taking private methods into accounts?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -364,11 +391,12 @@ private boolean isTransient(
...
+    Optional<ResolvedAnnotation<TransientProperties>> annotation =
+        getClassAnnotation(TransientProperties.class, typeHierarchy);
 
-    return transientProperties != null
-        ? Sets.newHashSet(transientProperties.value())
+    return annotation.isPresent()
+        ? Sets.newHashSet(annotation.get().getAnnotation().value())
```
tomekl007, line 411: Isn't that Set has always one element? Why do we need to return set here?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -436,20 +463,23 @@ private void scanFieldAnnotations(
...
-      Class<? extends Annotation> exclusiveAnnotation) {
+      Map<Class<? extends Annotation>, Annotation> annotations) {
+    Class<? extends Annotation> exclusiveAnnotation = getExclusiveAnnotation(annotations);
     for (Class<? extends Annotation> annotationClass : PROPERTY_ANNOTATIONS) {
-      Annotation annotation = getMethod.getAnnotation(annotationClass);
-      if (annotation != null) {
+      Optional<? extends ResolvedAnnotation<? extends Annotation>> annotation =
+          getMethodAnnotation(annotationClass, getMethod, typeHierarchy);
+      if (annotation.isPresent()) {
         if (EXCLUSIVE_PROPERTY_ANNOTATIONS.contains(annotationClass)) {
```
tomekl007, line 491: what if someone will annotate a method with 3 exclusive annotations?
We will report in the `reportMultipleAnnotationError` method only the first two, right?
Maybe we should report all of them?

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScannerTest.java
```diff
@@ -0,0 +1,208 @@
...
+    MockInterface y = i("y", z);
+    MockInterface x = i("x", y);
+    MockInterface w = i("w");
+    MockClass b = c("b", null, a);
+    MockClass c = c("c", b, x, w);
+
+    // when no HierarchyScanStrategy is defined, then the default behavior should be used
+    // which is to scan the entire tree and return it in the correct (bottom up) order.
+    assertThat(HierarchyScanner.resolveTypeHierarchy(c.classElement, types).toString())
+        .isEqualTo("[c, b, x, w, a, y, z]");
```
tomekl007, line 90: is the ordering of `a, y, z` is not important? Shouldn't it be `y,z,a`?

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScannerTest.java
```diff
@@ -0,0 +1,208 @@
...
+    Mockito.when(bHighest.highestAncestor()).thenReturn((Class) this.getClass());
+    Mockito.when(bHighest.includeHighestAncestor()).thenReturn(false);
+    MockInterface a = i("a");
+    MockInterface r = i("r");
+    MockInterface z = i("z", a);
+    MockInterface y = i("y", z);
+    MockInterface x = i("x", y);
+    MockInterface w = i("w");
+
+    MockClass b = c(this.getClass().getName(), null, a, r);
```
tomekl007: is there a particular reason why in this mock we choose to use `this.getClass().getName()` instead of the following pattern from different c mocks? (like `c("b", null, a,r)`
Same question to line https://github.com/datastax/java-driver/pull/1250/files#diff-26867dfb0fa8ecefa97f1251414342dcR85

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java
```diff
@@ -0,0 +1,191 @@
...
+import javax.lang.model.element.TypeElement;
+import javax.lang.model.type.TypeKind;
+import javax.lang.model.type.TypeMirror;
+import javax.lang.model.util.Types;
+
+/** Provides mechanisms for building and traversing a class/interfaces hierarchy. */
+public class HierarchyScanner {
+
+  // placeholder class for resolving the default HierarchyScanStrategy
+  @HierarchyScanStrategy
```
tomekl007, line 42: maybe we should specify values that we expect here explicitly instead of relying on defaults from `@HierarchyScanStrategy`?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java
```diff
@@ -0,0 +1,191 @@
...
+            .toString()
+            .equals(hierarchyScanStrategy.highestAncestor().getName());
+    while (!atHighestClass) {
+      // add super class
+      TypeMirror superClass = classElement.getSuperclass();
+      TypeElement superClassElement = null;
+      if (superClass.getKind() == TypeKind.DECLARED) {
+        superClassElement = (TypeElement) types.asElement(superClass);
+        atHighestClass =
+            superClassElement
```
tomekl007: I think that you can simplify that check with the usage of `ProcessorContext.Classes.isSame()`

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
tolbertam: definitely not necessary, HashSet should suffice.  Will update.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
tolbertam: I think `getSimpleName` is what we want, and not sure there is a difference between the qualified and simple name for fields and if there was, simple name is probably preferred since we can scan multiple class for fields.   This is not something I changed (just shows up as a change from nesting things further in code).

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
tolbertam: That's right, for now i've only done this for Entity as there's no Dao interface-level annotations yet (that I know of).

There may be an exception or two, such as the class-level `@CqlName` on an entity, which I think we would not want to inherit.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
tolbertam: the annotation value is an array, so there can be multiple elements.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScannerTest.java
tolbertam: a occurs before y and z in this context because b implements a.

If you look at the comment above with the graph the order of elements traversed should match the graph reading from bottom up, and left to right within the same depth of the graph, so for this case the order is encountered as

c
b x w
a y
z

since a has already been encountered, it's not added again.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScannerTest.java
tolbertam: We need to do it this way because `HierarchyScanStrategy.highestAncestor` takes a `Class` and you cannot mock classes since they are final.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java
tolbertam: nice, that's way better, and more correct. will use that.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java
tolbertam: I think it's safe to assume the annotation defaults are the most permissive.  In addition this is also used a placeholder if no `@HierarchyScanStrategy` is present, so should match the default anyways.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/AnnotationScanner.java
tolbertam: I think it's fine to quietly ignore these as we did the same with the previous mapper, we do the same for identifying property methods.

In addition, I don't think it can reasonably happen, you can't override a parent method that is private.  We could possibly even remove the check for static & private, but I think it's fine to be paranoid.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
tolbertam: this would mean passing optionals between methods, which is kinda ugly, but I see your point.  I agree that it's not good to be consistent though.  I'll change to standardize on not using Optional and using `@Nullable` where applicable.

tolbertam: actually looking at this more, the two places we use Optional, we pass in as constructors to `DefaultEntityDefinition`.  I'm going to leave that be since it isn't relevant for this PR.   I will annotate arguments as nullable though where applicable.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
tolbertam: I agree that it would be a nice to have, but i see it as low value for the effort it would take.  If have some time I'll look into it.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScannerTest.java
tolbertam: with your suggestion to use `Classes.isSame()` I no longer need to do this :+1:

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java
```diff
@@ -0,0 +1,191 @@
...
+        ClassForDefaultScanStrategy.class.getAnnotation(HierarchyScanStrategy.class);
+
+    // traverse hierarchy until a strategy is found.
+    final AtomicReference<HierarchyScanStrategy> ref = new AtomicReference<>(strategy);
+    traverseHierarchy(
+        strategy,
+        classElement,
+        types,
+        (TypeElement t) -> {
+          HierarchyScanStrategy discoveredStrategy = t.getAnnotation(HierarchyScanStrategy.class);
```
olim7t: I'm surprised that this works. Generally you can't inspect an annotation's `Class<?>` attribute during a processing round, because it might not be possible to load the referenced class yet (see `DefaultEntityFactory#readCustomConverterClasses` for example).

Have you tried with `HierarchyScanStrategy#highestAncestor()` being an application class?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java
```diff
@@ -0,0 +1,191 @@
...
+   *   <li>Visit <code>e</code>'s parent class (<code>p</code>), if <code>p</code> is null, stop
+   *       traversing after evaluating remaining interfaces
+   *   <li>Visit <code>e</code>'s interfaces, and record those interfaces' parents for later use (
+   *       <code>ni</code>)
+   *   <li>Visit <code>i</code>, and append those interface's parents to <code>ni</code> for later
+   *       use)
+   *   <li>If <code>p != null</code> </code>Set <code>e := p, i := ni</code> and repeat starting at
+   *       step 3
+   *   <li>Visit <code>ni</code> interfaces and their parents until we've reached root
+   * </ol>
```
olim7t, line 68: These explanations would be a bit easier to follow if you named your variables the same as in the code: `e` => `typeElement`, `i` => `interfacesToScan`, etc.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -72,97 +77,107 @@ public DefaultEntityFactory(ProcessorContext context) {
...
+        }
+        TypeMirror typeMirror = getMethod.getReturnType();
+        if (typeMirror.getKind() == TypeKind.VOID) {
+          continue;
+        }
+        String propertyName = Introspector.decapitalize(getMethodName.substring(3));
+        // skip properties we've already encountered.
+        if (encounteredPropertyNames.contains(propertyName)) {
+          continue;
+        }
```
olim7t, line 117: If I read `HierarchyScanner` correctly, `typeHierarchy` is in ascending order, so we'll run into child classes first. This could be a problem if we want our annotations to be inherited, for example:
```java
interface Identifiable {
  @PrimaryKey UUID getId();
  void setId(UUID id);
}
@Entity
class Product implements Identifiable {
  @Override public UUID getId() { ... }
  @Override public void setId(UUID id) { ... }

}
```
Maybe making our annotations `@Inherited` would be enough to address this, but it's probably not resolved when parsing annotation mirrors.

Another approach could be to traverse the set in reverse order. But that would prevent a class from overriding a method with additional annotations.

In any case we need to define rules for annotation inheritance, write them down in the manual and test them.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -72,97 +77,107 @@ public DefaultEntityFactory(ProcessorContext context) {
...
-                      + "with @%s, the annotation must be parameterized with an integer "
-                      + "indicating the position. Found duplicate index %d for %s and %s.",
-                  PartitionKey.class.getSimpleName(),
-                  partitionKeyIndex,
-                  previous.getGetterName(),
-                  property.getGetterName());
+        String setMethodName = getMethodName.replaceFirst("get", "set");
+        ExecutableElement setMethod = findSetMethod(typeElement, setMethodName, typeMirror);
+        if (setMethod == null) {
+          continue; // must have both
```
olim7t, line 122: This won't detect a property if the getter and setter come from different types in the hierarchy. That might be a fair requirement, but on the other hand it's pretty straightforward to scan the whole hierarchy for the setter, like you did for the field.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: Actually I just realized you're handling inheritance in `AnnotationScanner#getMethodAnnotation`, correct?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: It seems unlikely that a field with the same name would exist at a higher level _and_ be annotated,  I think we can safely ignore that case.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
tolbertam: Sure, I can add that.   I figured that `get` and `set` would be defined on the same type, but it's fair to define a super interface that only has getters.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
tolbertam: That's right, it works similarly to the 3.x mapper, parsing out the properties is done separately from parsing the annotations.  Once you've identified the properties, you find the annotations, which can be pulled from different levels of the hierarchy.

tolbertam: Although that is only the case for method annotations.  We only pull annotations from the identified field (also similar to 3.x mapper)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java
tolbertam: I'm not confident it will work as I haven't tested it yet at the compile phase, will see how it goes today ;)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: Sounds good, fields can't be overridden anyway so it's unlikely that the same one could be present at different levels.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java
tolbertam: You're right, it doesn't work, will follow the strategy you referenced :+1:

```
Error:(676, 3) java: Unexpected error while writing generated code: javax.lang.model.type.MirroredTypeException: Attempt to access Class object for TypeMirror com.datastax.oss.driver.mapper.EntityPolymorphismIT.Asset
...
com.datastax.oss.driver.internal.mapper.processor.util.HierarchyScanner.traverseHierarchy(HierarchyScanner.java:136)
  	at com.datastax.oss.driver.internal.mapper.processor.util.HierarchyScanner.traverseFullHierarchy(HierarchyScanner.java:111)
  	at com.datastax.oss.driver.internal.mapper.processor.util.HierarchyScanner.resolveTypeHierarchy(HierarchyScanner.java:75)
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java
tolbertam: Decided it's not really worth it, would require changing the API a bit and at this point things work great...

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java
tolbertam: Fixed this in 8fdcc880a1f6050cf7b915127e29f5c4c1a56c25

tolbertam:
Added documentation, I think there still might be issues with generic type variables in some Dao query methods..working on testing and fixing that.

manual/mapper/daos/README.md
```diff
@@ -48,4 +48,68 @@ ProductDao dao = inventoryMapper.productDao("someKeyspace");
 
 The returned object is thread-safe, and can securely be shared throughout your application.
 
+### Inheritance
```
tolbertam, line 51: I considered making a dedicated page for entity/dao inheritance, but wasn't sure if it'd be worth it or not, instead I added a section for each entity & dao.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScannerTest.java
```diff
@@ -0,0 +1,342 @@
...
+  class MockElement {
+    final String name;
+    final TypeMirror mirror;
+    final TypeElement classElement;
+    final List<MockElement> interfaces;
+    final MockClass parent;
+    final Name qfName;
+
+    @SuppressWarnings("unchecked")
+    MockElement(
```
tolbertam, line 254: The amount of mocking needed ended up getting out of control, perhaps we could consider removing this test now given we have decent integration test coverage.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -63,18 +70,48 @@
...
     for (VariableElement parameterElement : interfaceMethod.getParameters()) {
-      result.addParameter(
-          ClassName.get(parameterElement.asType()), parameterElement.getSimpleName().toString());
+      TypeName type = getTypeName(parameterElement.asType(), typeParameters);
+      result.addParameter(type, parameterElement.getSimpleName().toString());
     }
     return result;
   }
 
+  private static TypeName getTypeName(TypeMirror mirror, Map<Name, TypeElement> typeParameters) {
```
tolbertam, line 90: Feels like there should be an easier way to do this, but was the best I could come up with.

tolbertam:
Think I have this all sorted out.  There were a ton of fun little edge cases that I had to work out, hopefully I caught them all!

integration-tests/src/test/java/com/datastax/oss/driver/mapper/NullSavingStrategyIT.java
```diff
@@ -171,6 +184,10 @@ ProductSimpleDaoGlobalLevelDoNotSetOverrideSetToNull productDaoLocalOverride(
     ProductSimple findById(UUID productId);
   }
 
+  @Dao
+  public interface ProductSimpleDaoSetToNullFromParentInterface
+      extends ProductSimpleDaoImplicit, SetToNull {}
```
tomekl007, line 189: Why do we need to extend `ProductSimpleDaoImplicit` (instead of inlining it in the `ProductSimpleDaoSetToNullFromParentInterface`)?
It seems that we are testing two aspects - maybe it will be better to split it into two tests?
Or maybe we should also test a case where ProductSimpleDaoImplicit should extend `SetToNull`?

integration-tests/src/test/java/com/datastax/oss/driver/mapper/NullSavingStrategyIT.java
```diff
@@ -125,11 +131,18 @@ public void should_do_not_throw_when_construct_dao_with_global_level_SET_TO_NULL
...
+    @DaoFactory
+    ProductSimpleDaoSetToNullFromParentInterface productDaoSetToNullFromParentInterface(
+        @DaoKeyspace CqlIdentifier keyspace);
+
     @DaoFactory
     ProductSimpleDaoGlobalLevelDoNotSetOverrideSetToNull productDaoLocalOverride(
         @DaoKeyspace CqlIdentifier keyspace);
   }
 
+  @DefaultNullSavingStrategy(NullSavingStrategy.SET_TO_NULL)
```
tomekl007, line 143: Will this work if the @`Mapper` will be annotated with `@DefaultNullSavingStrategy`?
If so maybe we should add a test for it.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/EntityPolymorphismIT.java
```diff
@@ -85,25 +89,40 @@ public static void setup() {
   }
 
   // another parent interface with dao methods
-  interface WriteTimeDao<Y extends WriteTimeProvider> extends ShapeDao<Y> {
+  interface WriteTimeDao<Y extends WriteTimeProvider> extends BaseDao<Y> {
```
tomekl007, line 125: awesome

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java
```diff
@@ -0,0 +1,241 @@
...
+          }
+          return true;
+        });
+
+    return ref.get();
+  }
+
+  private static class HierarchyScanStrategyOptions {
+    boolean scanAncestors = defaultStrategy.scanAncestors();
+    TypeMirror highestAncestor = null;
```
tomekl007, line 120: why we don't take `defaultStrategy.highestAncestor`?

tolbertam: Good catch.  I think it ultimately doesn't have an effect, but we can get a TypeMirror for that (Object.class) :+1:

integration-tests/src/test/java/com/datastax/oss/driver/mapper/NullSavingStrategyIT.java
tolbertam: I don't believe so, at least I didn't intend for it to work that way.  We'd have to make `Mapper` annotations accessible for the Dao code, which is probably doable, but I think `Dao`-level and Daos supporting inheriting annotations may be powerful enough.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/NullSavingStrategyIT.java
tolbertam: I chose to use `ProductSimpleDaoImplicit` because it provides methods we can use and isn't already annotated with `DefaultNullSavingStrategy`.  So this is to verify that a parent annotation is inherited.  We could have redefined the `ProductSimpleDaoImplicit`s methods here, but this seemed easier.

manual/mapper/daos/README.md
```diff
@@ -48,4 +48,68 @@ ProductDao dao = inventoryMapper.productDao("someKeyspace");
...
+  @DaoFactory
+  SphereDao sphereDao(@DaoKeyspace CqlIdentifier keyspace);
+}
+```
+
+Note that interfaces that declare generic type variables should not be annotated with
+[@Dao].
+
+In addition to inheriting methods from parent interfaces, interface-level annotations such as
+[DefaultNullSavingStrategy] are also inherited:
```
olim7t:
```suggestion
[@DefaultNullSavingStrategy] are also inherited:
```

manual/mapper/daos/README.md
```diff
@@ -48,4 +48,68 @@ ProductDao dao = inventoryMapper.productDao("someKeyspace");
...
+
+```java
+import static com.datastax.oss.driver.api.mapper.entity.saving.NullSavingStrategy.SET_TO_NULL;
+
+@DefaultNullSavingStrategy(SET_TO_NULL)
+interface BaseDao {
+}
+
+interface RectangleDao extends BaseDao {};
```
olim7t: Nit: should probably be annotated with `@Dao` (not relevant here but more realistic)

manual/mapper/daos/README.md
```diff
@@ -48,4 +48,68 @@ ProductDao dao = inventoryMapper.productDao("someKeyspace");
...
+```
+
+Scanning priority is driven by proximity to the [@Dao] interface and within the same level 
+interfaces are given priority by the order they are in the type signature.  To control how the
```
olim7t: I find the wording a bit hard to understand.

manual/mapper/entities/README.md
```diff
@@ -303,6 +303,224 @@ If you want the name to be case-sensitive, it must be enclosed in double-quotes,
...
+Each concrete class must be annotated with [@Entity] and abstract classes and interfaces must not
+use this annotation.
+
+Scanning priority is driven by proximity to the [@Entity] class and within the same level classes
+are given priority over interfaces.  To control how the class hierarchy is scanned, annotate
+classes with [@HierarchyScanStrategy].
+
+Here is an example of a polymorphic mapping:
+
+```java
```
olim7t, line 317: That's a long example. For the sake of conciseness, I think you could:
* omit all constructors
* put trivial methods on one line:

    ```java
    public UUID getId() { return id; }
    ```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/CodeGeneratorFactory.java
```diff
@@ -75,5 +77,7 @@
    * @see #newDaoImplementation(TypeElement)
    */
   Optional<MethodGenerator> newDaoImplementationMethod(
-      ExecutableElement methodElement, DaoImplementationSharedCode enclosingClass);
+      ExecutableElement methodElement,
+      Map<Name, TypeElement> typeParameters,
+      DaoImplementationSharedCode enclosingClass);
```
olim7t, line 82: Nit: the javadoc of `newDaoImplementation` references this method, the link is now broken.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
```diff
@@ -73,11 +79,30 @@ protected TypeElement typeArgumentAsEntityElement(TypeMirror mirror) {
     return asEntityElement(declaredType.getTypeArguments().get(0));
   }
 
+  /** Evaluates whether or not this type argument could by an entity (declared or a typevar). */
```
olim7t:
```suggestion
  /** Evaluates whether or not this type argument could be an entity (declared or a typevar). */
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
```diff
@@ -93,8 +96,12 @@ public DaoGetEntityMethodGenerator(
     TypeElement entityElement = null;
     Transformation transformation = null;
     TypeMirror returnType = methodElement.getReturnType();
-    if (returnType.getKind() == TypeKind.DECLARED) {
-      Element element = ((DeclaredType) returnType).asElement();
+    if (couldBeEntity(returnType)) {
```
olim7t: `couldBeEntity` is a bit of a deceptive name, because it also matches iterables of entities. What it really is is `isDeclaredOrTypeVar`. Or we could just inline it, it's only used in two places.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java
```diff
@@ -93,8 +96,12 @@ public DaoGetEntityMethodGenerator(
...
+      Element element = asEntityElement(returnType);
+      if (element == null && returnType.getKind() == TypeKind.DECLARED) {
+        // if it's not an entity, treat it as a normal declared type.
+        element = ((DeclaredType) returnType).asElement();
+      }
```
olim7t: This really confuses me. `asEntityElement` already checks the element kind, but we check it again at line 106. Also, line 106 produces a null warning on `entity`.

How about doing it this way instead:
```java
Transformation transformation = null;
TypeMirror returnType = methodElement.getReturnType();
TypeElement entityElement = asEntityElement(returnType);
if (entityElement != null) {
  entityElement = (TypeElement) element;
  transformation = parameterIsGettable ? Transformation.NONE : Transformation.ONE;  
} else if (returnType.getKind() == TypeKind.DECLARED) {
  Element element = ((DeclaredType) returnType).asElement();
  if (context.getClassUtils().isSame(element, PagingIterable.class)) {
    ... // no change
  } else if (context.getClassUtils().isSame(element, MappedAsyncPagingIterable.class)) {
    ... // no change
  }
}
```
```

File: integration-tests/src/test/java/com/datastax/oss/driver/mapper/EntityPolymorphismIT.java [hunk 33]
```diff
@@ -0,0 +1,972 @@
...
+  }
+
+  @DataProvider
+  public static Object[][] setAndGetProvider() {
+    Function<CqlIdentifier, CircleDao> circleDao = keyspace -> mapper.circleDao(keyspace);
+    Function<CqlIdentifier, RectangleDao> rectangleDao = keyspace -> mapper.rectangleDao(keyspace);
+    Function<CqlIdentifier, SquareDao> squareDao = keyspace -> mapper.squareDao(keyspace);
+    Function<CqlIdentifier, SphereDao> sphereDao = keyspace -> mapper.sphereDao(keyspace);
+    return new Object[][] {
+      new Object[] {
```
olim7t:
```suggestion
      {
```
You don't need to repeat that for each inner array.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -70,13 +79,45 @@
...
 
+  // tracks interface type variable mappings as child interfaces discover them.
+  private final Map<TypeMirror, Map<Name, TypeElement>> typeMappingsForInterface =
+      Maps.newHashMap();
+
+  private final Set<TypeMirror> interfaces;
+  private final Map<Class<? extends Annotation>, Annotation> annotations;
+
+  private static final Set<Class<? extends Annotation>> ANNOTATIONS =
+      ImmutableSet.of(DefaultNullSavingStrategy.class);
```
olim7t, line 91: That doesn't really bother me, but this has the same name as the field. So maybe `ANNOTATIONS_TO_SCAN` instead?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -70,13 +79,45 @@
...
+          // don't replace annotations from lower levels.
+          annotations.putIfAbsent(annotationClass, annotation);
+        }
+      }
+    }
+    return ImmutableMap.copyOf(annotations);
+  }
+
+  @SuppressWarnings("unchecked")
+  private <A> Optional<A> getAnnotation(Class<A> annotationClass) {
```
olim7t:
```suggestion
  private <A extends Annotation> Optional<A> getAnnotation(Class<A> annotationClass) {
```
Avoids the "suspicious call" warning on `annotations.get`.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -70,13 +79,45 @@
...
+    return Optional.ofNullable((A) annotations.get(annotationClass));
```
olim7t:
```suggestion
    return Optional.ofNullable(annotationClass.cast(annotations.get(annotationClass)));
```
Essentially the same, but it eliminates the need for the `@SuppressWarnings` on the method.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -127,15 +168,80 @@ public String addQueryProvider(
...
-        .map(DefaultNullSavingStrategy::value);
+    return getAnnotation(DefaultNullSavingStrategy.class).map(DefaultNullSavingStrategy::value);
   }
 
   @Override
   protected ClassName getPrincipalTypeName() {
     return implementationName;
   }
 
+  private Map<Name, TypeElement> parseTypeParameters(TypeMirror mirror) {
```
olim7t, line 202: Since this is a pretty hairy method, maybe you could add comments that explain each part based on a common example.

E.g. `CircleDao extends ShapeDao<Circle>`, we are going through the method for `ShapeDao`.
* `mirror` is `ShapeDao<Circle>`
* `element` is `interface ShapeDao<T>`
* etc.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -127,15 +168,80 @@ public String addQueryProvider(
...
+    // Map interface type arguments to their concrete class.
+    Map<Name, TypeElement> typeParameters = Maps.newHashMap();
+    Map<Name, TypeElement> childTypeParameters =
+        typeMappingsForInterface.getOrDefault(mirror, Collections.emptyMap());
+
+    if (mirror instanceof DeclaredType) {
+      TypeElement element = (TypeElement) context.getTypeUtils().asElement(mirror);
+      DeclaredType declaredType = (DeclaredType) mirror;
+      for (int i = 0;
+          i < declaredType.getTypeArguments().size() && i < element.getTypeParameters().size();
```
olim7t: Can they ever be different? (or is it to handle code that doesn't even compile)

integration-tests/src/test/java/com/datastax/oss/driver/mapper/EntityPolymorphismIT.java
```diff
@@ -0,0 +1,972 @@
...
+
+    @Update
+    void update(T t);
+  }
+
+  // another parent interface with dao methods
+  interface WriteTimeDao<Y extends WriteTimeProvider> extends BaseDao<Y> {
+    @Insert(timestamp = ":writeTime")
+    void saveWithTime(Y y, long writeTime);
+  };
```
olim7t:
```suggestion
  }
```

integration-tests/src/test/java/com/datastax/oss/driver/mapper/EntityPolymorphismIT.java
```diff
@@ -0,0 +1,972 @@
...
+  // Define an intermediate interface with same type variable name to ensure
+  // this doesn't cause any issue in code generation.
+  interface ArbitraryInterface<Y extends Number> {
+    default long increment(Y input) {
+      return input.longValue() + 1;
+    }
+  }
+
+  @Dao
+  interface SquareDao extends WriteTimeDao<Square>, ArbitraryInterface<Float> {};
```
olim7t:
```suggestion
  interface SquareDao extends WriteTimeDao<Square>, ArbitraryInterface<Float> {}
```

integration-tests/src/test/java/com/datastax/oss/driver/mapper/EntityPolymorphismIT.java
```diff
@@ -0,0 +1,972 @@
...
+
+  @Dao
+  interface CircleDao extends WriteTimeDao<Circle> {}
+
+  @Dao
+  interface SphereDao extends WriteTimeDao<Sphere> {};
```
olim7t:
```suggestion
  interface SphereDao extends WriteTimeDao<Sphere> {}
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -127,15 +168,80 @@ public String addQueryProvider(
...
+          if (childTypeParameters.containsKey(previousName)) {
+            typeParameters.put(name, childTypeParameters.get(previousName));
+          } else {
+            context
+                .getMessager()
+                .error(
+                    element,
+                    "Could not resolve type parameter %s "
+                        + "on %s from child interfaces. This error usually means an interface "
+                        + "was inappropriately annotated with @Dao. Interfaces should only be annotated "
```
olim7t: Nit: could you inject `Dao.class.getSimpleName()` instead of hardcoding it in the string?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -127,15 +168,80 @@ public String addQueryProvider(
...
+        if (typeArgument instanceof DeclaredType) {
+          // if its a DeclaredType, we have the concrete type
+          Element concreteType = ((DeclaredType) typeArgument).asElement();
+          typeParameters.put(name, (TypeElement) concreteType);
+        } else if (typeArgument instanceof TypeVariable) {
+          // if its a TypeVariable, we resolve the concrete type
+          // from the previously encountered interface and alias it to the
+          // type on this class.
+          TypeVariable genericType = (TypeVariable) typeArgument;
+          Name previousName = genericType.asElement().getSimpleName();
```
olim7t: Sorry for being picky, but the names are throwing me off a bit here:
* why `genericType` for a type argument? (maybe inline it)
* `previousName`: if we're processing `BaseDao<T>` this would be `T` right? So why "previous"?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -73,97 +79,112 @@ public DefaultEntityFactory(ProcessorContext context) {
...
+                .getMessager()
+                .error(
+                    getMethod,
+                    "Duplicate partition key index: if multiple properties are annotated "
+                        + "with @%s, the annotation must be parameterized with an integer "
+                        + "indicating the position. Found duplicate index %d for %s and %s.",
+                    PartitionKey.class.getSimpleName(),
+                    partitionKeyIndex,
+                    previous.getGetterName(),
+                    property.getGetterName());
```
olim7t, line 163: I was wondering if this would generate the method multiple times when more than one concrete entity class is parsed, but it looks like `Messager` handles the de-duplication.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -434,26 +474,33 @@ private void scanFieldAnnotations(
...
       ExecutableElement getMethod,
-      Map<Class<? extends Annotation>, Annotation> annotations,
-      Class<? extends Annotation> exclusiveAnnotation) {
+      Map<Class<? extends Annotation>, Annotation> annotations) {
+    Class<? extends Annotation> exclusiveAnnotation = getExclusiveAnnotation(annotations);
     for (Class<? extends Annotation> annotationClass : PROPERTY_ANNOTATIONS) {
-      Annotation annotation = getMethod.getAnnotation(annotationClass);
-      if (annotation != null) {
+      Optional<? extends ResolvedAnnotation<? extends Annotation>> annotation =
+          getMethodAnnotation(annotationClass, getMethod, typeHierarchy);
```
olim7t: Nit: could you avoid static imports for non-obvious cases like this?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/AnnotationScanner.java
```diff
@@ -0,0 +1,74 @@
...
+          continue;
+        }
+        ExecutableElement candidateMethod = (ExecutableElement) child;
+        TypeMirror typeMirror = candidateMethod.getReturnType();
+        if (typeMirror.getKind() == TypeKind.VOID) {
+          continue;
+        }
+
+        if (candidateMethod.getSimpleName().equals(getMethod.getSimpleName())) {
```
olim7t, line 64: Comparing by name is enough because there's no way getters/setters with the same name but different types could coexist in the type hierarchy.
It also has a very interesting side-effect: the property type can be generic on a parent class/interface.
```java
interface Identifiable<IdentifierT> {
  @PartitionKey
  IdentifierT getId();
}
@Entity
public class Product implements Identifiable<UUID> { ... }
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java
```diff
@@ -0,0 +1,241 @@
...
+        // skip if at highest ancestor.
+        boolean atHighest = hierarchyScanStrategy.atHighest(interfaceType);
+        if (!atHighest || hierarchyScanStrategy.includeHighestAncestor) {
+          if (!typeConsumer.apply(interfaceType)) {
+            return;
+          }
+        }
+        if (!atHighest) {
+          newInterfacesToScan.addAll(interfaceElement.getInterfaces());
+        }
```
olim7t, line 239: 1. If `atHighest` happens for an interface, we don't communicate that to the caller. Wouldn't it then continue to navigate superclasses?
2. If it happens for anything but the first interface, we've already added the parents of the previous interfaces at the same level. Example: assuming highest=I2, we will traverse I3 (and all its parents I think).
    ```
    I3
    |
    I1 I2
    |  /
     C
```

**Review by olim7t — CHANGES_REQUESTED** [review]
Amazing how far you've taken this, I haven't yet found a case that isn't covered 👍 

See a bunch of remarks inline, most of them minor except for interface scanning, where I think we need to tighten a couple of edge cases.

**Comment by tolbertam** [comment]
Thanks for the feedback @olim7t , working on addressing that all now :+1:

File: mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java (hunk 47 above)
Comment (tolbertam) [reply mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java:239]: > If atHighest happens for an interface, we don't communicate that to the caller. Wouldn't it then continue to navigate superclasses?

That's right, I had considered changing `highestAncestor` to `highestAncestors`, but after thinking about it some more realized that'd maybe be overthinking it.  Not sure we should expect very complex type hierarchies where you need to prevent scanning at both the class and interface level.  I could add that though if you think it is worth it.

> If it happens for anything but the first interface, we've already added the parents of the previous interfaces at the same level

Correct, that's the way I intended it.  If `I2` is highest, that effectively means don't scan anything above `I2`, since `I1` is a separate hierarchy, `I3` would also be scanned.

File: mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java (hunk 43 above)
Comment (tolbertam) [reply mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java]: Not a problem, agree that the naming is a bit confusing there, i'll clean it up.

File: mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java (hunk 31 above)
Comment (tolbertam) [reply mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java]: I initially factored it into a separate method as was worried about missing a case and thought this may make it easier to communicate intent, but since it ended up only being needed in two places i'll change it back.

File: manual/mapper/entities/README.md (hunk 28 above)
Comment (tolbertam) [reply manual/mapper/entities/README.md:317]: sounds good.   I was aiming to provide an example that would compile (so i could test that I didn't miss anything), I'll clean it up.

File: manual/mapper/daos/README.md (hunk 27 above)
Comment (tolbertam) [reply manual/mapper/daos/README.md]: will clean up.

File: manual/mapper/daos/README.md (hunk 26 above)
Comment (tolbertam) [reply manual/mapper/daos/README.md]: thats an oversight, will fix.

File: mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java (hunk 47 above)
Comment (olim7t) [reply mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java:239]: > Not sure we should expect very complex type hierarchies

Agreed, in practice it will probably be a couple of levels of either parent classes or interfaces.

If the current behavior is as intended, that's fine with me as long as we document it somewhere in the manual.

File: mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java (hunk 32 above)
Comment (tolbertam) [reply mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoGetEntityMethodGenerator.java]: thats definitely more comprehensible :+1:

File: mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java (hunk 43 above)
Comment (tolbertam) [reply mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java]: `previousName` definitely not right.   I'll change it to `typeVariableName` as thats more descriptive of what it is.

File: mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java (hunk 38 above)
Comment (tolbertam) [reply mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java]: Ah nope, they can't. I previously had a bug (never committed) where I was pulling from the wrong element so I added this protection, but its no longer needed.

File: mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java (hunk 47 above)
Comment (tolbertam) [reply mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/HierarchyScanner.java:239]: I added details about this to `highestAncestor` javadoc in `HierarchyScanStrategy`.  I don't cover `HierarchyScanStrategy` in the manual, other than to mention and link to the javadoc for it in both inheritance sections as I expect a large percentage of the time the default behavior should be sufficient, and the javadoc does a good job explaining how to use it, but let me know if you'd like me to add more to the manual.

File: mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/CodeGeneratorFactory.java (hunk 29 above)
Comment (tolbertam) [reply mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/CodeGeneratorFactory.java:82]: fixed

File: manual/mapper/entities/README.md (hunk 28 above)
Comment (tolbertam) [reply manual/mapper/entities/README.md:317]: Went ahead and updated the manual section for this. :+1:

File: manual/mapper/daos/README.md (hunk 27 above)
Comment (tolbertam) [reply manual/mapper/daos/README.md]: cleaned this up a bit and referenced parts of the example where this applies.

File: mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java (hunk 37 above)
Comment (tolbertam) [reply mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java:202]: I took a stab at explaining this, first in the javadoc for the method to explain an example of an interface and how it and its parents are processed when parsing type arguments, and inline in the code, hope that makes it more clear.

**Comment by tolbertam** [comment]
I think I've addressed everything and this should be ready for another review pass :+1:

File: mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java (hunk 37 above)
Comment (olim7t) [reply mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java:202]: LGTM. To improve readability I would use a plain implementation comment for the method as well, we don't really need a javadoc here.

File: mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/HierarchyScanStrategy.java [hunk 48]
```diff
@@ -0,0 +1,80 @@
...
+   *
+   * <p>Defaults to <code>true</code>.
+   */
+  boolean scanAncestors() default true;
+
+  /**
+   * The {@link Class} to consider the highest ancestor, meaning the classes that this class extends
+   * or implements will not be scanned for annotations.
+   *
+   * <p>Note that a limitation of this configuration is the that the chosen highest ancestor only
```
olim7t:
```suggestion
   * <p>Note that a limitation of this configuration is that the chosen highest ancestor only
```

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/HierarchyScanStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+   * applies to the part of the type hierarchy that contains this class.
```
olim7t: I don't understand this what this paragraph means.

olim7t (approved):
LGTM

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
tolbertam: :+1: will do

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/HierarchyScanStrategy.java
tolbertam: yeah, this looks awkward and doesn't make sense at all.  It wasn't what I was trying to express at all, the following paragraph alone should be sufficient, so i'll just remove this.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
```diff
@@ -89,7 +89,7 @@ public EntityHelperGetMethodGenerator(
           // category. Their getter creates a GenericType under the hood, so there's no performance
           // advantage in calling them instead of the generic get().
           getBuilder.addStatement(
-              "returnValue.$L(source.get($L, $T))",
+              "returnValue.$L(source.get($L, $L))",
```
tolbertam, line 92: Noticed tests failing on 2.1 and it appears because I missed `setTags` so `tags` was never considered a property in my tests.  On fixing this I noticed helpers failed to compile because this expected a type, but really it should be a literal since `enclosingClass.addGenericTypeConstant(typeName)` adds and returns a type constant name.

tolbertam:
think i've sorted out remaining issues, will merge if jenkins runs cleanly.

PR #1251, opened 2019-06-03
JAVA-2264: HashedWheelTimer tick duration should be >= 10 ms on Windows

adutra:
Not sure how to test this other than manually verifying CPU usage.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -4,6 +4,7 @@
 
 ### 4.1.0 (in progress)
 
+- [bug] JAVA-2264: Adjust HashedWheelTimer tick duration from 1 to 100 ms
 - [bug] JAVA-2260: Handle empty collections in PreparedStatement.bind(...)
 - [improvement] JAVA-2278: Pass the request's log prefix to RequestTracker
 - [bug] JAVA-2253: Don't strip trailing zeros in ByteOrderedToken
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
@@ -33,12 +33,19 @@ import io.netty.util.concurrent.EventExecutorGroup;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GlobalEventExecutor;
 import io.netty.util.concurrent.PromiseCombiner;
+import io.netty.util.internal.PlatformDependent;
+import java.time.Duration;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
 import net.jcip.annotations.Immutable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 @Immutable
 public class DefaultNettyOptions implements NettyOptions {
+
+  private static final Logger LOG = LoggerFactory.getLogger(DefaultNettyOptions.class);
+
   private final EventLoopGroup ioEventLoopGroup;
   private final EventLoopGroup adminEventLoopGroup;
   private final int ioShutdownQuietPeriod;
@@ -83,10 +90,21 @@ public class DefaultNettyOptions implements NettyOptions {
             .setThreadFactory(safeFactory)
             .setNameFormat(context.getSessionName() + "-timer-%d")
             .build();
+
+    Duration tickDuration = config.getDuration(DefaultDriverOption.NETTY_TIMER_TICK_DURATION);
+    // JAVA-2264: tick durations on Windows cannot be less than 100 milliseconds,
+    // see https://github.com/netty/netty/issues/356.
+    if (PlatformDependent.isWindows() && tickDuration.toMillis() < 100) {
+      LOG.warn(
+          "Timer tick duration was set to a value too aggressive for Windows: {} ms; "
+              + "doing so is known to cause extreme CPU usage. "
+              + "Please set advanced.netty.timer.tick-duration to 100 ms or higher.",
+          tickDuration.toMillis());
+    }
     timer =
         new HashedWheelTimer(
             timerThreadFactory,
-            config.getDuration(DefaultDriverOption.NETTY_TIMER_TICK_DURATION).toNanos(),
+            tickDuration.toNanos(),
             TimeUnit.NANOSECONDS,
             config.getInt(DefaultDriverOption.NETTY_TIMER_TICKS_PER_WHEEL));
   }
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -88,6 +88,11 @@ datastax-java-driver {
     # For any serious deployment, we recommend that you use separate configuration profiles for DDL
     # and DML; you can then set the DML timeout much lower (down to a few milliseconds if needed).
     #
+    # Note that, because timeouts are scheduled on the driver's timer thread, the duration specified
+    # here must be greater than the timer tick duration defined by the
+    # advanced.netty.timer.tick-duration setting (see below). If that is not the case, timeouts will
+    # not be triggered as timely as desired.
+    #
     # Required: yes
     # Modifiable at runtime: yes, the new value will be used for requests issued after the change.
     # Overridable in a profile: yes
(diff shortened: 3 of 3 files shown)


core/src/main/resources/reference.conf
```diff
@@ -1400,13 +1400,18 @@ datastax-java-driver {
...
+      # The timer tick duration.
       # This is the how frequent the timer should wake-up to check for timed out tasks.
       # Lower resolution (i.e. longer durations) will leave more CPU cycles for running I/O
       # operations at the cost of precision of exactly when a request timeout will expire or a
       # speculative execution will run. Higher resolution (i.e. shorter durations) will result in
       # more precise request timeouts and speculative execution scheduling, but at the cost of CPU
       # cycles taken from I/O operations, which could lead to lower overall I/O throughput.
+      # Note for Windows users: the tick duration cannot be less than 10 milliseconds. If that is
+      # the case, the driver will use 10 milliseconds as the tick duration instead. Also, the tick
+      # duration must be a multiple of 10; if that is not the case, the driver will automatically
```
tomekl007: where is that round down logic?

tolbertam: https://github.com/netty/netty/blob/netty-4.1.28.Final/common/src/main/java/io/netty/util/HashedWheelTimer.java#L557-L564

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
```diff
@@ -83,10 +85,16 @@ public DefaultNettyOptions(InternalDriverContext context) {
             .setThreadFactory(safeFactory)
             .setNameFormat(context.getSessionName() + "-timer-%d")
             .build();
+    Duration tickDuration = config.getDuration(DefaultDriverOption.NETTY_TIMER_TICK_DURATION);
+    // JAVA-2264: tick durations on Windows cannot be less than 10 milliseconds,
+    // see https://github.com/netty/netty/issues/356.
+    if (PlatformDependent.isWindows() && tickDuration.toMillis() < 10) {
+      tickDuration = Duration.ofMillis(10);
```
emerkle826: On Windows, do we want to automatically raise the tick duration to the minimum supported on Windows, even if the client configures a value that's below the minimum? My concern is silently altering a mis-configured value. Would we rather log here that the driver is automatically changing the configured value, or throw an Exception/error out indicating the value is not valid?

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
```diff
@@ -83,10 +85,16 @@ public DefaultNettyOptions(InternalDriverContext context) {
             .setThreadFactory(safeFactory)
             .setNameFormat(context.getSessionName() + "-timer-%d")
             .build();
+    Duration tickDuration = config.getDuration(DefaultDriverOption.NETTY_TIMER_TICK_DURATION);
+    // JAVA-2264: tick durations on Windows cannot be less than 10 milliseconds,
+    // see https://github.com/netty/netty/issues/356.
+    if (PlatformDependent.isWindows() && tickDuration.toMillis() < 10) {
```
tolbertam: I think we may need to make the floor even higher, at least in my limited testing, setting to 10ms seems to prevent the thread from being in a tight loop, but still has high cpu utilization.

I wonder if we should maybe add another option, like `windows-tick-duration-minimum` and set it to 100ms by default.   If a windows user wants to reduce the tick interval, they will have to reduce this value as well.

Having a high resolution timer is really only needed for speculative execution.  I expect most users won't use it on windows.  I might be assuming too much, but I expect Windows to be more likely used in development settings, and deployed servers would run something like linux.

tolbertam: ah, this gets messy though, in [JAVA-2056](https://datastax-oss.atlassian.net/browse/JAVA-2056) we adjusted ticks per duration respective to the tick interval.

Not sure what the best course of action here is.   Maybe we could consider switching back to 100ms in general as we may not understand all the implications of using such a high resolution as 1ms, but there are downsides to doing that too.  If we do that, maybe we also add some logic to our spec exec code to warn if the constant threshold is < the tick duration/2 or something like that.

adutra:
PR updated as discussed with the team.

core/src/main/resources/reference.conf
```diff
@@ -1400,18 +1412,29 @@ datastax-java-driver {
...
+      # The timer tick duration.
+      # This is how frequent the timer should wake up to check for timed-out tasks or speculative
+      # executions. Lower resolution (i.e. longer durations) will leave more CPU cycles for running
+      # I/O operations at the cost of precision of exactly when a request timeout will expire or a
       # speculative execution will run. Higher resolution (i.e. shorter durations) will result in
       # more precise request timeouts and speculative execution scheduling, but at the cost of CPU
       # cycles taken from I/O operations, which could lead to lower overall I/O throughput.
       #
+      # The default value is 100 milliseconds, which is a comfortable value for most use cases.
+      # However if your are using more agressive timeouts or speculative execution delays, then you
```
tolbertam: `your are` -> `you are`

core/src/main/resources/reference.conf
```diff
@@ -1400,18 +1412,29 @@ datastax-java-driver {
...
+      # should lower the timer tick duration as well, so that its value is always equal to or lesser
+      # than the timeout duration and/or speculative execution delay you intend to use.
+      #
+      # Note for Windows users: avoid setting this to aggressive values, that is, anything under 100
+      # milliseconds; doing so is known to cause extreme CPU usage. Also, the tick duration must be
+      # a multiple of 10 under Windows; if that is not the case, the driver will automatically round
```
tolbertam: technically netty is what is rounding things down, right?

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
```diff
@@ -83,10 +90,21 @@ public DefaultNettyOptions(InternalDriverContext context) {
             .setThreadFactory(safeFactory)
             .setNameFormat(context.getSessionName() + "-timer-%d")
             .build();
+
+    Duration tickDuration = config.getDuration(DefaultDriverOption.NETTY_TIMER_TICK_DURATION);
+    // JAVA-2264: tick durations on Windows cannot be less than 100 milliseconds,
+    // see https://github.com/netty/netty/issues/356.
+    if (PlatformDependent.isWindows() && tickDuration.toMillis() < 100) {
+      LOG.warn(
+          "Timer tick duration was set to a value too aggressive for Windows: {} ms; "
```
tolbertam, line 99: Well said :+1:

changelog/README.md
```diff
@@ -2,8 +2,9 @@
 
 <!-- Note: contrary to 3.x, insert new entries *first* in their section -->
 
-### 4.0.2 (in progress)
+### 4.1.0 (in progress)
 
+- [bug] JAVA-2264: HashedWheelTimer tick duration should be >= 100 ms on Windows
```
tolbertam: Just a suggestion, but would make it more clear that we're changing the default overall here, something like:

```suggestion
- [bug] JAVA-2264: Adjust HashedWheelTimer tick duration from 1 to 100 ms
```

tolbertam (approved):
Looks great, thanks!

core/src/main/resources/reference.conf
emerkle826: Is it Netty or is it the JVM implementation on Windows that can't handle a granularity smaller than 10ms?

emerkle826 (approved):
LGTM, outside of the merge conflict.

core/src/main/resources/reference.conf
tolbertam: it's a windows thing, explained here: https://www.javamex.com/tutorials/threads/sleep_issues.shtml 

netty rounds down to nearest 10 here: https://github.com/netty/netty/blob/netty-4.1.28.Final/common/src/main/java/io/netty/util/HashedWheelTimer.java#L557-L564

olim7t (approved):

core/src/main/resources/reference.conf
adutra: It's Netty that rounds down – do you think we should mention Netty here?

tolbertam: I like your change, which is to mention that it is being rounded down (but not what is rounding it down).

PR #1252, opened 2019-06-03
JAVA-2253: Don't strip trailing zeros in ByteOrderedToken

adutra:
This commit also creates a defensive copy of the token's ByteBuffer
in order to guarantee its immutability.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ByteOrderedToken.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ByteOrderedToken.java
@@ -29,10 +29,11 @@ public class ByteOrderedToken implements Token {
 
   private final ByteBuffer value;
 
-  public ByteOrderedToken(ByteBuffer value) {
-    this.value = stripTrailingZeroBytes(value);
+  public ByteOrderedToken(@NonNull ByteBuffer value) {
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ByteOrderedToken.java
```diff
@@ -29,10 +29,11 @@
 
   private final ByteBuffer value;
 
-  public ByteOrderedToken(ByteBuffer value) {
-    this.value = stripTrailingZeroBytes(value);
+  public ByteOrderedToken(@NonNull ByteBuffer value) {
+    this.value = ByteBuffer.wrap(Bytes.getArray(value)).asReadOnlyBuffer();
```
adutra, line 33: Since the class is marked `@Immutable`, I'm creating a defensive, read-only copy of the input buffer.

olim7t: :+1: that's justified, the byte buffer can be provided directly by the end user (via `TokenMap.newToken`).

olim7t (approved):
👍 
Same remark as the 3.x PR: we can cover the new start=end check in a test. I'm adding it in the merge.

PR #1253, opened 2019-06-03
JAVA-2280: Driver should be resilient to invalid peer info

adutra:


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -77,7 +75,6 @@ public DefaultTopologyMonitor(InternalDriverContext context) {
     this.logPrefix = context.getSessionName();
     this.context = context;
     this.controlConnection = context.getControlConnection();
-    this.addressTranslator = context.getAddressTranslator();
```
adutra, line 80: Unused.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -125,7 +122,7 @@ public DefaultTopologyMonitor(InternalDriverContext context) {
                 ImmutableMap.of(
                     "address",
                     node.getBroadcastAddress().get().getAddress(),
-                    "peer",
```
adutra, line 128: I believe there is a bug here, which went unnoticed so far because the unit tests do not exercise this code and the integration tests only hit this line for peers v2.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -277,8 +273,7 @@ private NodeInfo asNodeInfo(AdminRow row) {
    *     from {@link #findInPeers(AdminResult, InetSocketAddress)}. Callers that don't already have
    *     it can use {@link #getBroadcastRpcAddress}.
    */
-  protected DefaultNodeInfo.Builder nodeInfoBuilder(
```
adutra: I prefer to return the object fully built since this method was always called with `nodeInfoBuilder(...).build()`. This way we can avoid returning a concrete type. (Note: we will have to modify the DSE driver accordingly.)

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitorTest.java
```diff
@@ -390,10 +430,6 @@ private StubbedQuery(String queryString, Map<String, Object> parameters, AdminRe
     private StubbedQuery(String queryString, AdminResult result) {
       this(queryString, Collections.emptyMap(), result);
     }
-
-    private CompletionStage<AdminResult> throwException() throws Exception {
```
adutra, line 394: Unused.

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitorTest.java
```diff
@@ -461,8 +497,4 @@ private AdminResult mockResult(AdminRow... rows) {
     when(result.iterator()).thenReturn(Iterators.forArray(rows));
     return result;
   }
-
-  private AdminResult errorResult() throws Exception {
```
adutra, line 465: Unused.

tomekl007:
I think that after this fix there is still a problem when one from:
data_center, host_id, rack, tokens is `null` - according to 3.x implementation of it
https://datastax-oss.atlassian.net/browse/JAVA-852 those also should be excluded (here we are only filtering out if rpc_address is `null`)

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -62,10 +61,13 @@
   // Assume topology queries never need paging
   private static final int INFINITE_PAGE_SIZE = -1;
 
+  private static final boolean EXTENDED_PEER_CHECK =
```
adutra: I'm note sure we need this system property in driver 4. Since this component is pluggable, I think we could just enforce the extended check and if users want a more relaxed one, they can override the `isPeerValid` method.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -259,69 +251,121 @@ private String retrievePeerTableName() {
...
-   *     from {@link #findInPeers(AdminResult, InetSocketAddress)}. Callers that don't already have
-   *     it can use {@link #getBroadcastRpcAddress}.
+   * Creates a {@link NodeInfo} instance from the given row; returns empty if the row does not
+   * contain enough information, in which case the node will be discarded and won't show up in the
+   * nodes list.
    */
-  protected DefaultNodeInfo.Builder nodeInfoBuilder(
-      AdminRow row, InetSocketAddress broadcastRpcAddress) {
+  @NonNull
+  protected Optional<NodeInfo> buildNodeInfo(@NonNull AdminRow row) {
```
adutra: I had to remove the second parameter because for a peer row, we need first to validate the row and only then obtain the broadcast RPC address; if we allow callers to pass that address, the peer validation might be broken.
On the flip side, the method signature becomes more straightforward, which is probably a good thing given that this method is meant to be overridden (see DSE driver).

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -259,69 +251,135 @@ private String retrievePeerTableName() {
...
+    if (broadcastInetAddress == null) {
+      // in system.peers or system.peers_v2
+      broadcastInetAddress = row.getInetAddress("peer");
+    }
+
     int broadcastPort = 0;
-    if (row.contains("peer_port")) {
+    if (row.contains("broadcast_port")) {
+      // system.local for Cassandra >= 4.0
+      broadcastPort = row.getInteger("broadcast_port");
```
adutra, line 321: I'm adding `broadcast_port`, seems like an oversight.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -259,69 +251,135 @@ private String retrievePeerTableName() {
...
+        return Optional.empty();
+      }
+      // Deployments that use a custom EndPoint implementation will need their own TopologyMonitor.
+      // One simple approach is to extend this class and override this method.
+      endPoint = new DefaultEndPoint(context.getAddressTranslator().translate(broadcastRpcAddress));
+    } else {
+      // Don't rely on system.local.rpc_address for the control node, because it mistakenly
+      // reports the normal RPC address instead of the broadcast one (CASSANDRA-11181). We
+      // already know the endpoint anyway since we've just used it to query.
+      endPoint = controlConnection.channel().getEndPoint();
```
adutra: This hopefully eliminates the need of that awkward downcast in the previous version.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -351,22 +409,28 @@ private void savePort(DriverChannel channel) {
...
-        return null;
+        // This could only happen if system tables are corrupted, but handle gracefully
+        return Optional.empty();
       }
-      return new InetSocketAddress(rpcAddress, port);
+      return Optional.of(new InetSocketAddress(rpcAddress, port));
     } else {
-      Integer rowPort = row.getInteger("native_port");
+      // system.local for Cassandra >= 4.0
+      Integer rowPort = row.getInteger("rpc_port");
```
adutra: Adding `rpc_port`, seems like an oversight.

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitorTest.java
```diff
@@ -150,8 +154,8 @@ public void should_refresh_node_from_peers_if_broadcast_address_is_present_v2()
     topologyMonitor.stubQueries(
         new StubbedQuery(
             "SELECT * FROM system.peers_v2 WHERE peer = :address and peer_port = :port",
-            ImmutableMap.of("address", ADDRESS2.getAddress(), "peer", 9042),
-            mockResult(mockPeersV2Row(2, node2.getHostId()))));
+            ImmutableMap.of("address", ADDRESS2.getAddress(), "port", 9042),
```
adutra, line 162: Same here, seems to be a bug in the tests.

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitorTest.java
```diff
@@ -256,16 +260,17 @@ public void should_get_new_node_from_peers() {
     verify(peer2).getInetAddress("rpc_address");
     verify(peer2, never()).getString(anyString());
 
-    verify(peer1).getInetAddress("rpc_address");
+    // the added node will have its rpc_address tried twice
+    verify(peer1, times(2)).getInetAddress("rpc_address");
```
adutra, line 268: This is new, we cannot reuse the retrieved address anymore inside `buildNodeInfo`, so the address gets computed twice.

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitorTest.java
```diff
@@ -363,7 +477,13 @@ private void stubQueries(StubbedQuery... queries) {
       assertThat(nextQuery.queryString).isEqualTo(queryString);
       assertThat(nextQuery.parameters).isEqualTo(parameters);
       if (nextQuery.error) {
-        new CompletableFuture<AdminResult>().completeExceptionally(new Exception("PlaceHolder"));
+        CompletableFuture<AdminResult> future = new CompletableFuture<>();
```
adutra: I think this was a bug: the future wasn't being returned to the caller.

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitorTest.java
```diff
@@ -329,7 +330,121 @@ public void should_refresh_node_list_from_local_and_peers() {
   }
 
   @Test
-  public void should_stop_executing_queries_once_closed() throws Exception {
+  public void should_refresh_node_list_from_local_and_peers_v2() {
```
adutra: This test is not directly related to peer checks, but It was missing so I added it.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Node.java
```diff
@@ -71,10 +89,13 @@
   Optional<InetSocketAddress> getBroadcastAddress();
```
adutra: It's too late obviously but I wonder why we went for optionals for some getters and nullable values for others.

core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRow.java
```diff
@@ -94,6 +94,15 @@ public boolean contains(String columnName) {
     return columnSpecs.containsKey(columnName);
   }
 
+  public boolean isNotEmpty(String columnName) {
+    if (contains(columnName)) {
+      int index = columnSpecs.get(columnName).index;
+      ByteBuffer bb = data.get(index);
+      return bb != null && bb.remaining() > 0;
```
adutra: This should cover empty collections too. It will also catch empty strings, but I think it's fine.

core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRow.java
```diff
@@ -94,6 +94,15 @@ public boolean contains(String columnName) {
     return columnSpecs.containsKey(columnName);
   }
 
+  public boolean isNotEmpty(String columnName) {
```
tomekl007: Isn't it better to make this method `isEmpty`?
There are some cases that we need to do negation of `isNotEmpty` and it is a little bit confusing

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
tomekl007: If any field that we are checking is `null` is causing the driver to get those exceptions (logged in the ticket) I think that we should have that check always enabled. I wonder what was a reason to add that null check to `3.x` - maybe backward compatibility?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -351,22 +392,59 @@ private void savePort(DriverChannel channel) {
...
+  /**
+   * Returns {@code true} if the given peer row is valid, and {@code false otherwise}.
+   *
+   * <p>This method must ensure that at least the row contains enough information to extract the
+   * node's broadcast RPC address; otherwise {@link #nodeInfoBuilder(AdminRow, InetSocketAddress)}
+   * may throw an exception.
+   */
+  protected boolean isPeerValid(AdminRow peerRow) {
+    boolean valid =
+        isSchemaV2
```
tomekl007: if we are setting `isSchemaV2 = true` initially (see line 89 and comment: 
```
// Set this to true initially, after the first refreshNodes is called this will either stay true
    // or be set to false;
```
and isValid check is executed before 
`
this.isSchemaV2 = false; // We should not attempt this query in the future.
`
Isn't there a risk that `isPeerValid` will return false(although the peer is valid for v1) because we tried to check `peerRow.isNotEmpty("native_address") && peerRow.isNotEmpty("native_port")` 
instead of `peerRow.isNotEmpty("rpc_address")`? Isn't there a race condition on those those checks?

Maybe we should check both v1 and v2 similarly to:
```
   InetAddress nativeAddress = row.getInetAddress("native_address");
    if (nativeAddress == null) {
      // in system.peers or system.local
      InetAddress rpcAddress = row.getInetAddress("rpc_address");
```
instad of one OR another?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -351,22 +392,59 @@ private void savePort(DriverChannel channel) {
...
+            ? peerRow.isNotEmpty("native_address") && peerRow.isNotEmpty("native_port")
+            : peerRow.isNotEmpty("rpc_address");
```
tomekl007: Shouldn't we check also `native_port` for v1?
SImilar to this:
```
else {
      // system.local for Cassandra >= 4.0
      Integer rowPort = row.getInteger("rpc_port");
      if (rowPort == null || rowPort == 0) {
        // system.peers_v2
        rowPort = row.getInteger("native_port");
        if (rowPort == null || rowPort == 0) {
          rowPort = port;
        }
      }
```

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -259,80 +256,124 @@ private String retrievePeerTableName() {
...
-    if (broadcastAddress == null) {
-      broadcastAddress = row.getInetAddress("peer"); // in system.peers
+      @NonNull AdminRow row, @Nullable InetSocketAddress broadcastRpcAddress) {
+
+    boolean peer = row.contains("peer");
+
+    EndPoint endPoint;
+    if (peer) {
+      // If this node is a peer, its broadcast RPC address must be present
+      assert broadcastRpcAddress != null;
```
tomekl007: Are you sure that this is always the case?
It seems that for example here:
https://github.com/datastax/java-driver/pull/1253/files#diff-6dcbde897818abed394fdf5233fb74b7R265
The `InetSocketAddress broadcastRpcAddress` argument can be null

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitorTest.java
```diff
@@ -329,7 +328,121 @@ public void should_refresh_node_list_from_local_and_peers() {
...
+    assertThat(node2.broadcastAddress).isNotNull().isEqualTo(ADDRESS2);
+  }
+
+  @Test
+  public void should_skip_invalid_peers_row_with_extended_check() {
+    // Given
+    topologyMonitor.isSchemaV2 = false;
+    node2.broadcastAddress = ADDRESS2;
+    AdminRow peer2 = mockPeersRow(2, node2.getHostId());
+    when(peer2.isNotEmpty("tokens")).thenReturn(false);
```
tomekl007: shouldn't we create a parameterized test for all columns that we checking with `isNotEmpty`?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
adutra: At the line you mention `broadcastRpcAddress` cannot be null since it was tested by `isValidPeer`. All calls to `nodeInfoBuilder` are protected by `isValidPeer` except [this one](https://github.com/datastax/java-driver/blob/cd295a34979e122a94da6d04aa32c1067d545e79/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java#L201) because this one is the local (control) host.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
adutra: `isValidPeer` cannot be executed before `isSchemaV2` is set to false, it is always called after the first complete node refresh.
But I agree that this is not obvious when you read the code, so I will change the implementation like you suggested.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
adutra: I think you meant `rpc_port`? The problem is that it is not present before C* 4.

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitorTest.java
adutra: Good idea, will do.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
tomekl007: yes, `rpc_port`. So what will happen if `rpc_port` in C* > 4 will be `null`? Isn't there a risk that we will get null pointer somewhere if `rpc_port` will be null (but we didn't check that)?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
adutra: TBH the ticket was logged because the broadcast RPC address was null. This value is the only one that really cannot be null, otherwise we cannot connect to that peer. But the driver could survive without the checks.

core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRow.java
adutra: I think `isEmpty` is a bit ambiguous because it can be true in two situations: when the column doesn't exist, or when it exists but is null/empty.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
adutra: If `rpc_port` is null the port will be the default port, see [here](https://github.com/datastax/java-driver/blob/f1fe61726296dba41e5106bbb90f3839167ae352/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java#L414). This won't make the driver error out (but it could end up being unable to connect to that node).
The problem if we start checking specific Cassandra versions is that we would be opening a can of worms: e.g. we would have to check if the node is running Cassandra 4+ but _not_ DSE 6.7 (which also reports Cassandra 4 but does not have that column).

core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRow.java
```diff
@@ -50,10 +51,12 @@ public AdminRow(
     this.protocolVersion = protocolVersion;
   }
 
+  @Nullable
```
adutra, line 54: Added nullability annotations here to better understand what's happening in `DefaultTopologyMonitor`. This is creating warnings in many other places though, I didn't fix them as we already have JAVA-2284.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultNode.java
```diff
@@ -97,19 +98,19 @@ public EndPoint getEndPoint() {
     return Optional.ofNullable(listenAddress);
   }
 
-  @NonNull
+  @Nullable
```
adutra, line 101: This was my own mistake, I guess, when I added annotations to these classes. The interface methods were marked nullable, but the implementations here were mistakenly marked non-null.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeInfo.java
```diff
@@ -107,20 +118,24 @@
...
    * will be copied as-is into {@link Node#getExtras()}.
    */
+  @Nullable
   Map<String, Object> getExtras();
 
   /**
    * The host ID that is assigned to this host by cassandra. The driver uses this to uniquely
    * identify a node.
    */
+  @NonNull
```
adutra, line 138: So I double-checked this and this needs to be non-null, otherwise the driver will throw NPEs in many places. I changed the peer check logic in `DefaultTopologyMonitor` to include the broadcast RPC address and the host ID.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodesRefresh.java
```diff
@@ -39,7 +40,7 @@ protected static boolean copyInfos(
     node.listenAddress = nodeInfo.getListenAddress().orElse(null);
     node.datacenter = nodeInfo.getDatacenter();
     node.rack = nodeInfo.getRack();
-    node.hostId = nodeInfo.getHostId();
+    node.hostId = Objects.requireNonNull(nodeInfo.getHostId());
```
adutra, line 43: Added this check to ensure that no null host ID will ever come in.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -218,7 +212,7 @@ public DefaultTopologyMonitor(InternalDriverContext context) {
       return CompletableFuture.completedFuture(true);
     }
     DriverChannel channel = controlConnection.channel();
-    return new SchemaAgreementChecker(channel, context, port, logPrefix).run();
+    return new SchemaAgreementChecker(channel, context, logPrefix).run();
```
adutra, line 220: Port was unused, I took the opportunity to remove it.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Node.java
```diff
@@ -157,10 +178,10 @@
...
    * identify a node even when the underling IP address changes.
    *
-   * <p>This should be non-null in a healthy deployment, but the driver will still function, and
-   * report {@code null} here, if the server metadata was corrupted.
+   * <p>This information is required; the driver will not function properly if this method returns
+   * {@code null}.
    */
-  @Nullable
+  @NonNull
   UUID getHostId();
```
adutra, line 184: So I double-checked this and this needs to be non-null, otherwise the driver will throw NPEs in many places. I changed the peer check logic in `DefaultTopologyMonitor` to include the broadcast RPC address and the host ID.

tomekl007 (approved):
LGTM but I think that someone with more experience with the driver should take a look at this as well.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Node.java
olim7t: My bad, it only really became non-null after [JAVA-2165](https://datastax-oss.atlassian.net/browse/JAVA-2165), I forgot to change it then.

core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRow.java
olim7t: Why can't we just use `row.getInetAddress("rpc_address") != null`?

It produces the same result. The only difference is that it decodes the value unnecessarily, but in the context of parsing rows I don't think it matters.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
olim7t: 👍 another leftover from 2165

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
olim7t: One thing I'll note is that `broadcastRpcAddress` is not derived from the endpoint anymore, so it might be wrong (CASSANDRA-11181). But we don't use it to connect anymore so it won't hurt the driver. One might view it as an accurate reflection of the system tables, even if they are wrong.

olim7t (commented):
There's a lot going on in this PR 😓 

Apart from the peer check, I see at least the following:
* Add nullability annotations in AdminRow
* Remove unused port in SchemaAgreementChecker
* Fix nullability annotation on Node.getHostId()
* Add more details to Node javadocs
* Fix various bugs in C*4 system tables parsing (peers->port in query, read broadcast_port and rpc_port from system.local...)
* Add integration test for peers_v2 parsing

Is there any way you could separate that better, at least in different commits? It's really hard to review one big chunk where everything changes at once. I'm also thinking about future maintainability (e.g. if we ever need to revert something).

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
adutra: That was precisely my intent.

core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRow.java
adutra: My intent was to avoid decoding the value indeed. I will change that if you think it's OK to decode.

adutra:
> Is there any way you could separate that better, at least in different commits?

Done, although I frankly don't think that this PR was that hard to review. We've seen much worse :-)

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
olim7t: We're in the branch `nativeAddress != null`, so it's a peers_v2 row. There's no rpc_port column in that table.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Node.java
```diff
@@ -178,10 +178,10 @@
    * The host ID that is assigned to this node by Cassandra. This value can be used to uniquely
    * identify a node even when the underling IP address changes.
    *
-   * <p>This should be non-null in a healthy deployment, but the driver will still function, and
-   * report {@code null} here, if the server metadata was corrupted.
+   * <p>This information is required; the driver will not function properly if this method returns
+   * {@code null}.
```
olim7t: The user doesn't provide `Node` instances, the driver does, so I think we should rather formulate this as "this information is always present" or something similar.

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitorTest.java
olim7t: nit: this could be shortened with `CompletableFutures.failedFuture`.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
olim7t: OK, works for me.

olim7t: Actually, I just realized that re-reading the control connection's channel is racy: it could have reconnected since we executed the requests. You can handle that by reading the control endpoint in `refreshNodeList` and passing it into this method.

olim7t (changes requested):
Thanks for taking the time to split the commits, it really made things easier to follow for me.

See my feedback in reply to previous comments. I think "Take into account rpc_port column (system.local for C* 4.0+)" is unnecessary, and we should address the race in the topology monitor.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
adutra: Right, the method implementation is not correct, I will amend that and simplify.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
adutra: Good catch, will do.

olim7t:
LGTM.

Last detail, could you reword those two commit messages when you rebase:
* `Take into account rpc_port column (system.local for C* 4.0+)`: not true anymore after you squash the follow-up commit.
* `JAVA-2280: Driver should be resilient to invalid peer info`: does not accurately describe the fix. Suggestion: "Ignore peer rows with missing host id or RPC address" (same for changelog entry).

olim7t (approved):

olim7t:
> `Take into account rpc_port column (system.local for C* 4.0+)`: not true anymore after you squash the follow-up commit.

My bad, the message is still applicable.

PR #1254, opened 2019-06-03
JAVA-2272: Make DAO method return types pluggable

olim7t:
Motivation:

We want to allow some degree of extensibility in the mapper. In
particular, it would be nice if a custom mapper extension could handle
new result types for DAO methods (e.g. `@Select`), without having to
rewrite the whole method generator.

Modifications:

Extract a superinterface DaoReturnTypeKind from the current enum.
Introduce a pluggable DaoReturnTypeParser in CodeGeneratorFactory.
Modify DAO method generators, so that the set of supported return type
kinds can be customized by overriding a method.

Result:

To support new return types, downstream projects will:
- write their own implementation of DaoReturnTypeKind (most likely as
  another enum)
- write their own DaoReturnTypeParser implementation
- extend the default method generators to override
  getSupportedReturnTypes()
- write their own CodeGeneratorFactory to return all those custom
  elements.

TODO:
- [x] try to make parser more extensible
- [x] try to factor result checking code in `DaoMethodGenerator`

### DIFF
--- a/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
+++ b/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
@@ -15,30 +15,30 @@
  */
 package com.datastax.oss.driver.internal.mapper.processor.dao;
 
-import static com.datastax.oss.driver.internal.mapper.processor.dao.ReturnTypeKind.BOOLEAN;
-import static com.datastax.oss.driver.internal.mapper.processor.dao.ReturnTypeKind.FUTURE_OF_ASYNC_RESULT_SET;
-import static com.datastax.oss.driver.internal.mapper.processor.dao.ReturnTypeKind.FUTURE_OF_BOOLEAN;
-import static com.datastax.oss.driver.internal.mapper.processor.dao.ReturnTypeKind.FUTURE_OF_VOID;
-import static com.datastax.oss.driver.internal.mapper.processor.dao.ReturnTypeKind.RESULT_SET;
-import static com.datastax.oss.driver.internal.mapper.processor.dao.ReturnTypeKind.VOID;
+import static com.datastax.oss.driver.internal.mapper.processor.dao.DefaultDaoReturnTypeKind.BOOLEAN;
+import static com.datastax.oss.driver.internal.mapper.processor.dao.DefaultDaoReturnTypeKind.FUTURE_OF_ASYNC_RESULT_SET;
+import static com.datastax.oss.driver.internal.mapper.processor.dao.DefaultDaoReturnTypeKind.FUTURE_OF_BOOLEAN;
+import static com.datastax.oss.driver.internal.mapper.processor.dao.DefaultDaoReturnTypeKind.FUTURE_OF_VOID;
+import static com.datastax.oss.driver.internal.mapper.processor.dao.DefaultDaoReturnTypeKind.RESULT_SET;
+import static com.datastax.oss.driver.internal.mapper.processor.dao.DefaultDaoReturnTypeKind.VOID;
 
 import com.datastax.oss.driver.api.core.cql.BoundStatement;
 import com.datastax.oss.driver.api.core.cql.BoundStatementBuilder;
 import com.datastax.oss.driver.api.core.cql.SimpleStatement;
 import com.datastax.oss.driver.api.mapper.annotations.Delete;
-import com.datastax.oss.driver.internal.core.util.concurrent.CompletableFutures;
 import com.datastax.oss.driver.internal.mapper.processor.ProcessorContext;
 import com.datastax.oss.driver.internal.mapper.processor.entity.EntityDefinition;
 import com.datastax.oss.driver.internal.mapper.processor.entity.PropertyDefinition;
 import com.datastax.oss.driver.internal.mapper.processor.util.generation.GeneratedCodePatterns;
+import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableSet;
 import com.squareup.javapoet.ClassName;
 import com.squareup.javapoet.CodeBlock;
 import com.squareup.javapoet.MethodSpec;
 import com.squareup.javapoet.TypeName;
-import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 import javax.lang.model.element.AnnotationMirror;
 import javax.lang.model.element.AnnotationValue;
@@ -50,16 +50,17 @@ import javax.lang.model.type.TypeMirror;
 
 public class DaoDeleteMethodGenerator extends DaoMethodGenerator {
 
-  private static final EnumSet<ReturnTypeKind> SUPPORTED_RETURN_TYPES =
-      EnumSet.of(
-          VOID, FUTURE_OF_VOID, BOOLEAN, FUTURE_OF_BOOLEAN, RESULT_SET, FUTURE_OF_ASYNC_RESULT_SET);
-
   public DaoDeleteMethodGenerator(
       ExecutableElement methodElement,
       Map<Name, TypeElement> typeParameters,
       DaoImplementationSharedCode enclosingClass,
       ProcessorContext context) {
     super(methodElement, typeParameters, enclosingClass, context);
+  }
+
+  protected Set<DaoReturnTypeKind> getSupportedReturnTypes() {
+    return ImmutableSet.of(
+        VOID, FUTURE_OF_VOID, BOOLEAN, FUTURE_OF_BOOLEAN, RESULT_SET, FUTURE_OF_ASYNC_RESULT_SET);
   }
 
   @Override
@@ -99,7 +100,7 @@ public class DaoDeleteMethodGenerator extends DaoMethodGenerator {
       return Optional.empty();
     }
     VariableElement firstParameter = parameters.get(0);
-    entityElement = asEntityElement(firstParameter);
+    entityElement = EntityUtils.asEntityElement(firstParameter, typeParameters);
     hasEntityParameter = (entityElement != null);
     if (hasEntityParameter) {
       entityDefinition = context.getEntityFactory().getDefinition(entityElement);
@@ -142,15 +143,9 @@ public class DaoDeleteMethodGenerator extends DaoMethodGenerator {
     }
 
     // Validate the return type:
-    ReturnType returnType = parseReturnType(methodElement.getReturnType());
-    if (!SUPPORTED_RETURN_TYPES.contains(returnType.kind)) {
-      context
-          .getMessager()
-          .error(
-              methodElement,
-              "Invalid return type: %s methods must return void, boolean or a result set, "
-                  + "or a CompletableFuture/CompletionStage of Void, Boolean or AsyncResultSet",
-              Delete.class.getSimpleName());
+    DaoReturnType returnType =
+        parseAndValidateReturnType(getSupportedReturnTypes(), Delete.class.getSimpleName());
+    if (returnType == null) {
       return Optional.empty();
     }
 
@@ -162,18 +157,14 @@ public class DaoDeleteMethodGenerator extends DaoMethodGenerator {
             (methodBuilder, requestName) ->
                 generatePrepareRequest(methodBuilder, requestName, helperFieldName));
 
-    MethodSpec.Builder deleteBuilder =
-        GeneratedCodePatterns.override(methodElement, typeParameters);
-    if (returnType.kind.isAsync) {
-      deleteBuilder.beginControlFlow("try");
-    }
+    CodeBlock.Builder methodBodyBuilder = CodeBlock.builder();
 
-    deleteBuilder.addStatement(
+    methodBodyBuilder.addStatement(
         "$T boundStatementBuilder = $L.boundStatementBuilder()",
         BoundStatementBuilder.class,
         statementName);
     if (statementAttributeParam != null) {
-      deleteBuilder.addStatement(
+      methodBodyBuilder.addStatement(
           "boundStatementBuilder = populateBoundStatementWithAttributes(boundStatementBuilder, $L)",
           statementAttributeParam.getSimpleName().toString());
     }
@@ -186,7 +177,7 @@ public class DaoDeleteMethodGenerator extends DaoMethodGenerator {
             property.getType(),
             CodeBlock.of("$L.$L()", firstParameter.getSimpleName(), property.getGetterName()),
             "boundStatementBuilder",
-            deleteBuilder,
+            methodBodyBuilder,
             enclosingClass);
       }
       nextParameterIndex = 1;
@@ -202,7 +193,7 @@ public class DaoDeleteMethodGenerator extends DaoMethodGenerator {
       GeneratedCodePatterns.bindParameters(
           parameters.subList(0, primaryKeyNames.size()),
           primaryKeyNames,
-          deleteBuilder,
+          methodBodyBuilder,
           enclosingClass,
           context,
           false);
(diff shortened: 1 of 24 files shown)


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DefaultDaoReturnTypeParser.java
```diff
@@ -0,0 +1,99 @@
...
+public class DefaultDaoReturnTypeParser implements DaoReturnTypeParser {
+
+  private final ProcessorContext context;
+
+  public DefaultDaoReturnTypeParser(ProcessorContext context) {
+    this.context = context;
+  }
+
+  @Override
+  public DaoReturnType parse(TypeMirror returnTypeMirror) {
```
olim7t: I didn't try to make any part of this method reusable, but it might be possible without being too messy. I'll give it a try.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoSelectMethodGeneratorTest.java
```diff
@@ -41,19 +41,19 @@ public void should_fail_with_expected_error(String expectedError, MethodSpec met
   public static Object[][] invalidSignatures() {
     return new Object[][] {
       {
-        "Invalid return type: Select methods must return an Entity-annotated class, or a "
-            + "CompletionStage, CompletableFuture, PagingIterable or "
-            + "CompletionStage/CompletableFuture<MappedAsyncPagingIterable> thereof",
+        "Invalid return type: Select methods must return one of [ENTITY, OPTIONAL_ENTITY, "
+            + "FUTURE_OF_ENTITY, FUTURE_OF_OPTIONAL_ENTITY, PAGING_ITERABLE, "
+            + "FUTURE_OF_ASYNC_PAGING_ITERABLE]",
```
olim7t, line 46: This is less user-friendly than the previous message, and also gets ugly for `@Query`.

But it has the merit of being exhaustive, and adapting automatically to extensions. Return types are already well covered in the documentation, so I think this could be fine.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoInsertMethodGenerator.java
```diff
@@ -83,18 +82,20 @@ public DaoInsertMethodGenerator(
...
-    ReturnType returnType = parseReturnType(methodElement.getReturnType());
-    if (!SUPPORTED_RETURN_TYPES.contains(returnType.kind)) {
+    DaoReturnType returnType = parseReturnType(methodElement.getReturnType());
+    Set<DaoReturnTypeKind> supportedReturnTypes = getSupportedReturnTypes();
+    if (!supportedReturnTypes.contains(returnType.getKind())) {
       context
           .getMessager()
           .error(
               methodElement,
-              "Invalid return type: %s methods must return either void or the entity class "
```
tomekl007, line 95: If we are touching that logic maybe we could move it as a method to `DaoBase` and call with `annotation` and `supportedReturnTypes` parameters from every specific method generator?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoQueryMethodGenerator.java
```diff
@@ -48,25 +51,35 @@ public DaoQueryMethodGenerator(
     nullSavingStrategyValidation = new NullSavingStrategyValidation(context);
   }
 
+  protected Set<DaoReturnTypeKind> getSupportedReturnTypes() {
```
tomekl007, line 56: Is there a chance that it will be called multiple times?
If so shouldn't we construct it once (possibly in static block) and assign to constant? (only returning it here)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DefaultDaoReturnTypeParser.java
tomekl007: Yes, I think ti will be useful to make it easily extensible for clients.
If they will decide to plug custom logic overriding this:
```
@Override
  public DaoReturnTypeParser getDaoReturnTypeParser() {
    return daoReturnTypeParser;
  }
```
This is possible that the client will want to override handling only one specific `getKind()` so I think it will be useful to handle rest of the kinds by simply calling `super` instead of copy-pasting this method

olim7t:
Force-pushed to rebase on latest java2078.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoInsertMethodGenerator.java
olim7t: Yes I thought about it. It's not used by every subclass but it could be worth it, I'll give it a try.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoQueryMethodGenerator.java
olim7t: We'd still have to expose the protected method in order to allow subclasses to override it, so that wouldn't prevent badly written code to call the method instead of using the constant.

In the end it's all our own internal code so the risk is low. And instantiating the set twice is not catastrophic, just slightly inefficient.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
```diff
@@ -141,16 +141,9 @@ public DaoDeleteMethodGenerator(
...
-          .error(
-              methodElement,
-              "Invalid return type: %s methods must return one of %s",
-              Delete.class.getSimpleName(),
-              supportedReturnTypes);
+    DaoReturnType returnType =
+        parseAndValidateReturnType(getSupportedReturnTypes(), Delete.class.getSimpleName());
+    if (returnType == null) {
       return Optional.empty();
     }
```
olim7t, line 150: Not super happy about this. We could make `parseAndValidate...` return an optional, and map the rest of the method onto it. But then why not do it for everything else... I don't want to rewrite the whole method in a functional style.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DefaultDaoReturnTypeParser.java
```diff
@@ -0,0 +1,196 @@
...
+        DEFAULT_FUTURE_OF_ENTITY_CONTAINER_MATCHES);
+  }
+
+  protected DefaultDaoReturnTypeParser(
+      ProcessorContext context,
+      Map<TypeKind, DaoReturnType> typeKindMatches,
+      Map<Class<?>, DaoReturnType> classMatches,
+      Map<Class<?>, DaoReturnTypeKind> entityContainerMatches,
+      Map<Class<?>, DaoReturnType> futureOfClassMatches,
+      Map<Class<?>, DaoReturnTypeKind> futureOfEntityContainerMatches) {
```
olim7t, line 118: OK, this looks much better now. Subclasses can use this constructor to pass their own maps instead of the defaults, so it would be trivial to support other container types (e.g. `Publisher<Product>`) without copy-pasting everything.

This also makes `parse()` easier to read I think.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
```diff
@@ -40,8 +40,24 @@ public DaoMethodGenerator(
     this.context = context;
   }
 
-  protected DaoReturnType parseReturnType(TypeMirror returnTypeMirror) {
-    return context.getCodeGeneratorFactory().getDaoReturnTypeParser().parse(returnTypeMirror);
+  protected DaoReturnType parseAndValidateReturnType(
```
tomekl007, line 52: Missing  `nullable` annotation on return

tomekl007 (approved):

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
```diff
@@ -49,17 +49,18 @@
...
-          VOID, FUTURE_OF_VOID, BOOLEAN, FUTURE_OF_BOOLEAN, RESULT_SET, FUTURE_OF_ASYNC_RESULT_SET);
-
   public DaoDeleteMethodGenerator(
       ExecutableElement methodElement,
       DaoImplementationSharedCode enclosingClass,
       ProcessorContext context) {
     super(methodElement, enclosingClass, context);
   }
 
+  protected Set<DaoReturnTypeKind> getSupportedReturnTypes() {
```
tolbertam, line 61: Noticed that 5/8 implementations of `DaoMethodGenerator` have a `getSupportedReturnTypes()` definition.  Is it worth making it part of `DaoMethodGenerator` API and having `parseAndValidateReturnTypes()` implicitly call it?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoInsertMethodGenerator.java
```diff
@@ -83,18 +82,13 @@ public DaoInsertMethodGenerator(
     }
 
     // Validate the return type:
-    ReturnType returnType = parseReturnType(methodElement.getReturnType());
-    if (!SUPPORTED_RETURN_TYPES.contains(returnType.kind)) {
```
tolbertam, line 90: Nice that this removes repetitive code (logging invalid return type)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
tolbertam: Ahh I see @tomekl007 had a [similar comment](https://github.com/datastax/java-driver/pull/1254/files#r290603859)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoReturnType.java
```diff
@@ -0,0 +1,66 @@
...
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.internal.mapper.processor.dao;
+
+import javax.lang.model.element.TypeElement;
+
+/** Holds information about the return type of a DAO method. */
+public class DaoReturnType {
+
+  public static final DaoReturnType VOID = new DaoReturnType(DefaultDaoReturnTypeKind.VOID);
```
tolbertam, line 23: Nice, like this better than `DaoMethodGenerator#parseReturnType`

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoSelectMethodGeneratorTest.java
tolbertam: Agreed that it is not as nice, but it still is pretty apparent what is valid :+1:

tolbertam (approved):
This looks good to me :+1:.  I imagine this may be difficult to rebase with changes from #1250, but I think test suite should cover things, so :+1: on merging assuming rebase goes well.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DefaultDaoReturnTypeParser.java
```diff
@@ -0,0 +1,229 @@
...
+      if (element != null) {
+        for (Map.Entry<Class<?>, DaoReturnType> entry : classMatches.entrySet()) {
+          Class<?> simpleClass = entry.getKey();
+          if (context.getClassUtils().isSame(element, simpleClass)) {
+            return entry.getValue();
+          }
+        }
+      }
+
+      // DAO parameterization by more complex types (futures, containers...) is not supported
```
olim7t, line 224: I think we've discussed that already in #1250. I don't think this will be a common need, but we can always revisit later if needed.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/EntityUtils.java
```diff
@@ -0,0 +1,84 @@
...
+  public static TypeElement asEntityElement(
+      TypeMirror mirror, Map<Name, TypeElement> typeParameters) {
+    Element element;
+    if (mirror.getKind() == TypeKind.TYPEVAR) {
+      // extract concrete implementation for type variable.
+      TypeVariable typeVariable = ((TypeVariable) mirror);
+      Name name = typeVariable.asElement().getSimpleName();
+      element = typeParameters.get(name);
+      if (element == null) {
+        return null;
```
olim7t, line 68: The original code issued an error message here, but I don't have enough context to keep it:
```java
        // this should not happen, but error out just in case.
        context
            .getMessager()
            .error(methodElement, "Could not resolve the concrete implementation for %s", name);
```

I think it's fine, because the only way to get here is if a generic interface is improperly annotated as a DAO:
```java
  @Dao
  interface TestDao<T> {
    @Select
    T findById(int id);
  }
```
And this already produces an error message

> Could not resolve type parameter T on TestDao<T> from child interfaces. This error usually means an interface was inappropriately annotated with @Dao. Interfaces should only be annotated with @Dao if all generic type variables are declared."

olim7t (commented):
Squashed and rebased on latest java2078.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/EntityUtils.java
tolbertam: That's right.  During my implementation of #1250, things lined up in such a way where I didn't initially correctly implement my logic for resolving type parameters, so sometimes this case would happen (`Could not resolve the concrete implementation`).  After I added some additional logic in `DaoImplementationGenerator` and added the error where if we can't resolve the concrete type from previously encountered interfaces this is no longer possible with how this code is currently used, so should be ok to remove this.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DefaultDaoReturnTypeParser.java
tolbertam: :+1: although just to make sure I understand the comment correctly, these are for parameterization of types that we don't already handle, such as `CompletableFuture`, `Optional`, etc, right?  I think this is fine, especially with this PR adding the capability to add your own return type parsing logic.

olim7t: This wouldn't work:
```java
interface BaseDao<SelectResultT> {
  @Select
  SelectResultT findAll();
}
@Dao
interface ProductDao extends BaseDao<PagingIterable<Product>> {
}
```
Because the `typeParameters` map has `TypeElement` values. We'd need to store type mirrors in order to resolve `SelectResultT` correctly.
But that's a silly example, I don't think it will come up in practice.

tolbertam: Ahh I see, not a scenario I considered and seems unlikely :+1:

PR #1255, opened 2019-06-04
JAVA-2275: Allow Entity annotation to specify a default keyspace

tomekl007:


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityDefinition.java
```diff
@@ -44,6 +47,7 @@ public DefaultEntityDefinition(
         customCqlName
             .map(n -> CodeBlock.of("$S", n))
             .orElse(cqlNameGenerator.buildCqlName(javaName));
+    this.defaultKeyspace = defaultKeyspace;
```
tomekl007, line 50: I am not sure if we want to apply cqlNameGenerator on the `defaultKeyspace`.
If someone is adding explicit `@Entity(defaultKeyspace ="a")` I am hesitating about transforming it to other ks name using `cqlNameGenerator` in this case. @tolbertam @olim7t wdyt?

olim7t: You're right, it shouldn't be transformed. The user provides a CQL name directly here, like in `@CqlName`. If it's case-sensitive, they'll have to double quote manually.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/DefaultKeyspaceIT.java
```diff
@@ -0,0 +1,336 @@
...
+
+  @BeforeClass
+  public static void setup() {
+    CqlSession session = sessionRule.session();
+    session.execute(
+        SimpleStatement.builder(
+                "CREATE KEYSPACE IF NOT EXISTS "
+                    + DEFAULT_KEYSPACE
+                    + " "
+                    + "WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1}")
```
olim7t: Nit: you can use `String.format` to inject the keyspace name, it will be a bit more readable.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/DefaultKeyspaceIT.java
```diff
@@ -0,0 +1,336 @@
...
+import java.util.UUID;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+
+@Category(ParallelizableTests.class)
+public class DefaultKeyspaceIT {
```
olim7t, line 46: 👍

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityDefinition.java
```diff
@@ -26,6 +27,9 @@
 
   CodeBlock getCqlName();
 
+  @Nullable
+  String getDefaultKeyspace();
```
olim7t, line 31: We should probably generalize the use of nullability annotations in the code. I tend to leave them out for internal code, but really it doesn't hurt to have them either.

Created [JAVA-2284](https://datastax-oss.atlassian.net/browse/JAVA-2284).

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityAnnotationTest.java
```diff
@@ -74,4 +75,29 @@ public void should_fail_on_interface() {
             .addAnnotation(Entity.class)
             .build());
   }
+
+  @Test
+  public void should_fail_when_default_keyspace_attribute_has_more_than_one_element() {
```
olim7t: Nit: compilation doesn't really fail, this should be named `should_warn...`

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Entity.java
```diff
@@ -74,4 +74,24 @@
  */
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.CLASS)
-public @interface Entity {}
+public @interface Entity {
+  /**
+   * Specifies a default keyspace to use when doing operations on a given {@code @Entity}
```
olim7t:
```suggestion
   * Specifies a default keyspace to use when doing operations on this entity.
```
Javadoc doesn't need to link to itself.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Entity.java
```diff
@@ -74,4 +74,24 @@
...
+   *
+   * <p>The defaultKeyspace is ignored if you will create your {@link Dao} using {@link DaoFactory}
+   * with explicitly setting {@link DaoKeyspace}:
```
olim7t: I would just say "if you pass a keyspace when building your DAO" (and modify the example as suggested in my other comment).

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Entity.java
```diff
@@ -74,4 +74,24 @@
...
+   *
+   * <pre>
+   *  &#64;Mapper
+   *  public interface InventoryMapper {
+   *    &#64;DaoFactory
+   *    ProductDao productDao(@DaoKeyspace String keyspace);
+   *  }
+   * </pre>
```
olim7t, line 102: This might be obvious, but just to drive the message home I would make the example more descriptive:

Add an outline of the entity definition before the mapper:
```java
@Product(defaultKeyspace = "inventory")
public class Product { ... }
```

Add a parameterless factory method to the mapper:
```java
@DaoFactory
ProductDao productDao();
```

Add sample client code:
```java
ProductDao productDao = mapper.productDao();
productDao.insert(product); // inserts into inventory.product

ProductDao productDaoTest = mapper.productDao("test");
productDaoTest.insert(product); // inserts into test.product
```

manual/mapper/entities/README.md
```diff
@@ -251,8 +251,32 @@ i.e.:
...
+```
+
+The `defaultKeyspace` is ignored if you will create your [@Dao] using [@DaoFactory] with explicitly setting [@DaoKeyspace]:
+```java
+@Mapper
+public interface InventoryMapper { 
+  @DaoFactory
+  ProductDao productDao(@DaoKeyspace String keyspace);
+}
+```
```
olim7t, line 282: See my suggestions on the annotation javadocs.

olim7t (changes requested):
LGTM, just a few minor comments about docs.

olim7t:
Force-pushed to rebase on latest java2078.

olim7t (approved):

olim7t:
Merged 👍 
One thing I missed on my previous review is that we can simply use an empty string as the default value for `@Entity.defaultKeyspace()`, as we already do for other string parameters. I did the change before merging.

PR #1256, opened 2019-06-04
Developer Guide Async Examples

tomekl007:
Driver concurrent examples needed for @jorgebay dev guide
http://docsreview.datastax.lan/en/dse/6.7-dev-DSP-18407/dse-dev/datastax_enterprise/appDevGuide/driverManagingConcurrency.html
There were written as part of JAVA-2148 but we decided to merge the only example that was using `advanced.throttler`. We postponed merge of those two until they will be actually needed.
The existing (merged) example has one caveat: 
>
Using a request throttler in this case will create `TOTAL_NUMBER_OF_INSERTS` queued statement executions, if you increase the number (i.e. to millions), it will cause excessive memory pressure.

Both examples from this PR are limiting the number of submitted (in-flight) requests alleviating the problem with excessive memory pressure in this described case.


examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustom.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Semaphore;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Creates a keyspace and tables, and loads data using Multi-Threaded approach into them.
```
tolbertam:
```suggestion
 * Creates a keyspace and table, and loads data using a multi-threaded approach.
```

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustom.java
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * <p>This example makes usage of a {@link CqlSession#execute(String)} method, which is responsible
+ * for executing requests in a blocking way. It uses {@link ExecutorService} to limit number of
+ * concurrent request to {@code CONCURRENCY_LEVEL}. It leverages {@link CompletableFuture} to
+ * achieve concurrency. It maintains at most {@code IN_FLIGHT_REQUESTS} using {@link Semaphore}.
+ *
+ * <p>Preconditions:
+ *
+ * <ul>
+ *   <li>An Apache Cassandra(R) cluster is running and accessible through the contacts points
```
tolbertam:
```suggestion
 *   <li>An Apache Cassandra(R) cluster is running and accessible through the contact points
```

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustom.java
```diff
@@ -0,0 +1,148 @@
...
+ *       identified by basic.contact-points (see application.conf).
+ * </ul>
+ *
+ * <p>Side effects:
+ *
+ * <ul>
+ *   <li>creates a new keyspace "examples" in the session. If a keyspace with this name already
+ *       exists, it will be reused;
+ *   <li>creates a table "examples.tbl_sample_kv". If it exist already, it will be reused;
```
tolbertam:
```suggestion
 *   <li>creates a table "examples.tbl_sample_kv". If it exists already, it will be reused;
```

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustomAsync.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.BiConsumer;
+
+/**
+ * Creates a keyspace and tables, and loads data using Async API into them.
```
tolbertam:
```suggestion
 * Creates a keyspace and table, and loads data using an async API.
```

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustomAsync.java
```diff
@@ -0,0 +1,191 @@
...
+ *
+ * <p>This example makes usage of a {@link CqlSession#executeAsync(String)} method, which is
+ * responsible for executing requests in a non-blocking way. It uses {@link CompletableFuture} to
+ * limit number of concurrent request to {@code CONCURRENCY_LEVEL}.
+ *
+ * <p>Preconditions:
+ *
+ * <ul>
+ *   <li>An Apache Cassandra(R) cluster is running and accessible through the contacts points
```
tolbertam:
```suggestion
 *   <li>An Apache Cassandra(R) cluster is running and accessible through the contact points
```

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustomAsync.java
```diff
@@ -0,0 +1,191 @@
...
+
+    // Construct CONCURRENCY_LEVEL number of ranges.
+    // Each range will be executed independently.
+    List<Range> ranges = createRanges(CONCURRENCY_LEVEL, TOTAL_NUMBER_OF_INSERTS);
+
+    // List of pending CONCURRENCY_LEVEL features that we will wait for at the end of the program.
+    List<CompletableFuture<?>> pending = new ArrayList<>();
+
+    // Every range will have dedicated CompletableFuture handling the execution.
+    for (Range range : ranges) {
```
tolbertam, line 89: This is an interesting way of controlling concurrency but it suffers a weakness in that if you have a slow response it may delay submitting requests in that range while other ranges may have otherwise completed, so you aren't effectively optimizing your concurrency.  For the intent of the example it's fine though, especially since it mirrors the nodejs implementation.

An alternative way to accomplish this would be to submit requests from the same thread, and on completion notify/wake up the thread in some way (could be as simple as enqueuing the next task into a queue) to evaluate the current number of inflight requests and then submit more.  I think we don't need to go that far though.

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustom.java
```diff
@@ -0,0 +1,148 @@
...
+    System.out.println(
+        String.format(
+            "Finished executing %s queries with a concurrency level of %s.",
+            insertsCounter.get(), threads.size()));
+    // Shutdown executor to free resources
+    executor.shutdown();
+  }
+
+  private static void createSchema(CqlSession session) {
+    session.execute(
```
tolbertam, line 132: If you have a multi-node cluster, its possible that the schema queries will timeout after 2s.  Maybe we should add a `schema` profile to application.conf and have all our examples use it.

olim7t:
[Initial discussion](https://github.com/datastax/java-driver/pull/1208#discussion_r264722982) on the JAVA-2148 PR.

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustom.java
```diff
@@ -0,0 +1,148 @@
...
+    // For every i we will insert a record to db
+    for (int i = 0; i < TOTAL_NUMBER_OF_INSERTS; i++) {
+      // Before submitting a request, we need to acquire 1 permit.
+      // If there is no permits available it blocks caller thread.
+      SEMAPHORE.acquire();
+      // Copy to final variable for usage in a separate thread
+      final int counter = i;
+
+      // We are running CqlSession.execute in a separate thread pool (executor)
+      CompletableFuture.supplyAsync(
```
tolbertam: I think it may be easier to follow if you just use `executor.submit`.  Also, there would be no need to return `executeRequest`, since we aren't using the resulting future anyways.

tolbertam (approved):
looks good to me, had a few suggestions, but in general these look good.

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustom.java
```diff
@@ -0,0 +1,148 @@
...
+          // Here the separate thread pool is passed as the argument
+          executor);
+    }
+    // Await for execution of TOTAL_NUMBER_OF_INSERTS
+    REQUEST_LATCH.await();
+
+    System.out.println(
+        String.format(
+            "Finished executing %s queries with a concurrency level of %s.",
+            insertsCounter.get(), threads.size()));
```
adutra: Do we really need to store the thread names just to count how many were actually used? Sound a bit anecdotic to me; the size of this set is likely to be equal to `CONCURRENCY_LEVEL` anyway.

adutra: Same for `insertsCounter`, it can only be equal to `TOTAL_NUMBER_OF_INSERTS`.

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustom.java
```diff
@@ -0,0 +1,148 @@
...
+    // Shutdown executor to free resources
+    executor.shutdown();
```
adutra, line 127: You should call `executor.awaitTermination()` after `shutdown()`.

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustom.java
adutra: Another reason why `supplyAsync` is not the best option is that suppliers are not really meant to throw exceptions, but `session.execute` may throw an exception.

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustom.java
```diff
@@ -0,0 +1,148 @@
...
+          () -> {
+            insertsCounter.incrementAndGet();
+            threads.add(Thread.currentThread().getName());
+
+            ResultSet executeRequest;
+            try {
+              executeRequest =
+                  session.execute(
+                      pst.bind().setUuid("id", UUID.randomUUID()).setInt("value", counter));
+            } finally {
```
adutra, line 110: I think you should handle errors here for the sake of example, to show users how to deal with failed queries (a simple `printStackTrace()` is enough).

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustomAsync.java
```diff
@@ -0,0 +1,191 @@
...
+ *       exists, it will be reused;
+ *   <li>creates a table "examples.tbl_sample_kv". If it exist already, it will be reused;
+ *   <li>inserts a TOTAL_NUMBER_OF_INSERTS of rows into the table.
+ * </ul>
+ *
+ * @see <a href="http://datastax.github.io/java-driver/manual/">Java driver online manual</a>
+ */
+public class LimitConcurrencyCustomAsync {
+  private static final int CONCURRENCY_LEVEL = 32;
+  private static final int TOTAL_NUMBER_OF_INSERTS = 10_000;
```
adutra, line 60: I find the usage of `TOTAL_NUMBER_OF_INSERTS` a bit artificial. What people will typically use is a list, or queue, or stream of statements to execute. I think that the examples would be more helpful if we demonstrated how to execute statements stored in such structures concurrently.

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustomAsync.java
adutra: I think we should provide users with something simpler to grasp. How about this (using my own suggestion of creating a queue of statements first):

```
private static void insertConcurrent(CqlSession session) throws InterruptedException {
  PreparedStatement pst =
      session.prepare(
          insertInto("examples", "tbl_sample_kv")
              .value("id", bindMarker("id"))
              .value("value", bindMarker("value"))
              .build());

  // Executor service with CONCURRENCY_LEVEL number of threads that states an upper limit
  // on number of request in progress.
  ExecutorService executor = Executors.newFixedThreadPool(CONCURRENCY_LEVEL);

  // create statements
  Queue<BoundStatement> statements = new ConcurrentLinkedQueue<>();
  for (int i = 0; i < TOTAL_NUMBER_OF_INSERTS; i++) {
    statements.add(pst.bind().setUuid("id", UUID.randomUUID()).setInt("value", i));
  }

  // execute statements
  for (int i = 0; i < CONCURRENCY_LEVEL; i++) {
    executor.submit(() -> acquireQueryAndRelease(executor, session, statements));
  }

  REQUEST_LATCH.await();

  // Shutdown executor to free resources
  executor.shutdown();
}

private static void acquireQueryAndRelease(
    ExecutorService executor, CqlSession session, Queue<BoundStatement> statements) {
  BoundStatement stmt = statements.poll();
  if (stmt == null) {
    return;
  }
  SEMAPHORE.acquireUninterruptibly();
  session
      .executeAsync(stmt)
      .whenCompleteAsync(
          (result, error) -> {
            REQUEST_LATCH.countDown();
            SEMAPHORE.release();
            if (error != null) {
              error.printStackTrace();
            }
            acquireQueryAndRelease(executor, session, statements);
          },
          executor);
}
```

adutra (changes requested):

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustomAsync.java
tolbertam: That's a good solution :+1: and I like that better than my suggestion.  One suggestion, since the generated statements are based on an incrementing value, is there a need to pregenerate them?  You could just use an atomic long and increment and generate the statements as you execute them, and then return in `acquireQueryAndRelease` when the long is above `TOTAL_NUMBER_OF_INSERTS`.

jorgebay: I think for the purpose of this example, having multiple sequential async executions is good enough. It's true that it may suffer from a single slow response in each `executeOneAtATime()` call, but with a large enough sample it should average out.

I would like to avoid generating the statements upfront in a queue as we will be allocating all the memory upfront, making the example unsuitable for unbounded `TOTAL_NUMBER_OF_INSERTS` values.

adutra: Unbounded inserts are rare. I still believe that 99% of users will have if not the statements, at least the data ready in memory before the insert operation begins. Large datasets will typically be broken into smaller batches of a few 1000 rows if need be. But in any case, nobody generates data to be inserted on-the-fly, unless you are populating the database with random data.
Truly unbounded use cases would imo be typically handled by using the java Stream API, a Producer-Consumer setup with shared queue, or by reactive programming.

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustom.java
tomekl007: Good idea, I think it will be good to do it in a separate PR.
I've created JIRA ticket to capture that:
https://datastax-oss.atlassian.net/browse/JAVA-2318

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustomAsync.java
tomekl007: I think that the main question is: do we want to have parity between examples in nodejs and java. Currently, the nodejs example is doing it using ranges and existing java example does the same. I like @adutra example because is more java idiomatic but still there is a problem of throttling requests - if we will have unbounded `Queue` there could be an out of memory error.
This line: 
```
  for (int i = 0; i < TOTAL_NUMBER_OF_INSERTS; i++) {
    statements.add(pst.bind().setUuid("id", UUID.randomUUID()).setInt("value", i));
  }
```
still is not solving that problem of out of memory on statements.

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustom.java
tomekl007: Concerning `threads` I agree (changed).
Concerning `insertCounter` I think it is beneficial to have that information in a log to see an actual number of inserts that succeeded.
For `LimitConcurrencyRequestThrottler` we have `List<CompletableFuture<?>> pending ` list to get that info. Here to assert that all records were properly inserter we need some additional counter. (I moved the incrementing logic after the execute call)
https://github.com/datastax/java-driver/pull/1256/commits/704ac8e626e00df0387d79fee52a70254c72c843
wdyt?

examples/src/main/java/com/datastax/oss/driver/examples/concurrent/LimitConcurrencyCustomAsync.java
adutra: 1. I think it's better if each driver strives to present idiomatic examples for its language.
2. About the potential out of memory problem: this is just an example, so why not simply add a comment above the for-loop that creates the statements:

> The code below creates statements in a loop for the sake of simplicity. In real life, you would probably obtain the data to insert from an external resource or system (e.g. a file, or a REST service); care should be taken not to run out of memory if too much data is held in memory.

tomekl007: @adutra Looking at your example I think that it is almost the same as existing: `LimitConcurrencyCustom`. Both are using `SEMAPHORE` to limit `IN_FLIGHT` and both are using     
`ExecutorService executor = Executors.newFixedThreadPool(CONCURRENCY_LEVEL);` to limit number of concurrent requests. There is only one difference - in your example, you are using intermediate `Queue`.
My suggestion is like this:
- I can add a `Queue` to the `LimitConcurrencyCustom` example to make a clearer distinction between consumer and producer - this will be java idiomatic way OR add the 3rd example with a Queue.
- keep `LimitConcurrencyCustomAsync` according to @jorgebay suggestion to also have an example that is showing the throttling of produced queries (and is one to one with `NodeJs` example).
Wdyt?

jorgebay: I agree with @tomekl007 about keeping `LimitConcurrencyCustomAsync` in the current state, as it shows a technique that chains futures without relying on any external component and concepts (latch / queue / semaphore). It covers important concepts that are hard to grasp for most users in few lines of code.

adutra: Ok to keep the code as is. I don't mind, really.

adutra (approved):

PR #1257, opened 2019-06-04
JAVA-2138: Allow query methods to set statement attributes with annotations

GregBestland:


mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Attributes.java
```diff
@@ -0,0 +1,48 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.CLASS)
+public @interface Attributes {
+
+  String ExecutionProfileName() default "";
```
olim7t:
```suggestion
  String executionProfileName() default "";
```

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Attributes.java
```diff
@@ -0,0 +1,48 @@
...
+  int timeout() default Integer.MIN_VALUE;
+
+  String routingKeyspace() default "";
+
+  /**
+   * How to handle null query parameters.
+   *
+   * <p>This defaults to {@link NullSavingStrategy#DO_NOT_SET}.
+   */
+  NullSavingStrategy nullSavingStrategy() default NullSavingStrategy.DO_NOT_SET;
```
olim7t: Looks unused, and probably not needed here.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/StatementAttributesIT.java
```diff
@@ -198,11 +243,23 @@ private void validateQueryOptions(QueryLog log) {
     @Insert
     void save(Simple simple, StatementAttributes attributes);
 
+    @Insert
+    @Attributes(consistencyLevel = "ANY", serialConsistencyLevel = "QUORUM", pageSize = 13)
```
olim7t: This is probably my OCD speaking, but it bothers me that the annotation does not use the same terminology as the parameter type. Of course it would be ludicrous to name them the same, but it would be nice to have some consistency.
My suggestion is `@WithStatementAttributes`, but I'm open to other ideas.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
```diff
@@ -169,6 +171,17 @@ public DaoDeleteMethodGenerator(
...
+          "boundStatementBuilder = populateBoundStatementWithAttributes(boundStatementBuilder, $S, $S, $S, $L, $L, $L, $S)",
+          attributes.ExecutionProfileName(),
+          attributes.consistencyLevel(),
+          attributes.serialConsistencyLevel(),
+          attributes.idempotence(),
+          attributes.pageSize(),
+          attributes.timeout(),
+          attributes.routingKeyspace());
+    }
     if (statementAttributeParam != null) {
```
olim7t: This allows both the annotation and the parameter.
1. :+1: I think this is the right thing to do. I can see cases where you might want one attribute static and another dynamic.
2. We should cover that in tests.
3. There is going to be a problem with the way default values are handled. Currently the parameter version of `populateBoundStatementWithAttributes` always sets the values, it might overwrite an annotation value with a default that the user didn't set explicitly. Actually I think we might in fact need a way to tell which values were set on `StatementAttributes` (sorry I made you change that in 2139): either `xxxWasSet` methods, or we make every getter return an optional.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
```diff
@@ -169,6 +171,17 @@ public DaoDeleteMethodGenerator(
...
+      deleteBuilder.addStatement(
+          "boundStatementBuilder = populateBoundStatementWithAttributes(boundStatementBuilder, $S, $S, $S, $L, $L, $L, $S)",
+          attributes.ExecutionProfileName(),
+          attributes.consistencyLevel(),
+          attributes.serialConsistencyLevel(),
+          attributes.idempotence(),
+          attributes.pageSize(),
+          attributes.timeout(),
+          attributes.routingKeyspace());
+    }
```
olim7t: This can be extracted as a shared method in `DaoMethodGenerator`.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
```diff
@@ -143,6 +147,50 @@ public static BoundStatementBuilder populateBoundStatementWithAttributes(
...
+    if (value.equals("")) {
+      return null;
+    }
+    return value;
+  }
+
+  private ConsistencyLevel getConsistencyLevelFromName(String name) {
+    DefaultDriverContext ddContext = (DefaultDriverContext) (context.getSession().getContext());
+    ConsistencyLevelRegistry registry = ddContext.getConsistencyLevelRegistry();
+    return registry.codeToLevel(registry.nameToCode(name));
```
olim7t, line 168: 👍 , but cast to `InternalDriverContext` instead.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Attributes.java
```diff
@@ -0,0 +1,48 @@
...
+
+  int pageSize() default Integer.MIN_VALUE;
+
+  boolean idempotence() default false;
+
+  String consistencyLevel() default "";
+
+  String serialConsistencyLevel() default "";
+
+  int timeout() default Integer.MIN_VALUE;
```
olim7t: Maybe we should use a string for this, with the format expected by [Duration.parse()](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-) (I hesitate calling that "more user-friendly", that format is horrific).

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Attributes.java
```diff
@@ -0,0 +1,48 @@
...
+
+import com.datastax.oss.driver.api.mapper.entity.saving.NullSavingStrategy;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.CLASS)
+public @interface Attributes {
```
olim7t: 👍 I think this is the right set of attributes to expose here, the others don't really make sense statically.

olim7t (changes requested):
This is looking great, we just need to figure out defaults handling when both the annotation and the parameter are present.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
```diff
@@ -169,6 +171,17 @@ public DaoDeleteMethodGenerator(
         "$T boundStatementBuilder = $L.boundStatementBuilder()",
         BoundStatementBuilder.class,
         statementName);
+    if (attributes != null) {
```
tomekl007: Don't we want to support `Update` as well?
I think that we should create tests in mapper tests per operation (InsertIT, DeleteIt)

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
```diff
@@ -143,6 +147,50 @@ public static BoundStatementBuilder populateBoundStatementWithAttributes(
...
+      String serialConsistencyLevel,
+      boolean idempotent,
+      int pageSize,
+      long timeout,
+      String keyspace) {
+
+    Duration duration = null;
+    if (timeout != Long.MIN_VALUE) {
+      duration = Duration.ofMillis(timeout);
+    }
```
tomekl007, line 142: don't we need a similar check and set default for pageSize and idempotent?
Or those two defaults:
```

  int pageSize() default Integer.MIN_VALUE;

  boolean idempotence() default false;
``` 
are safe to pass to the `builder` without conversion to builder neutral values? (like for example for `setExecutionProfileName` the `null` is set if annotation didn't set it explicitly)

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Attributes.java
tomekl007: On which level the `Attributes` annotation can be specified? Only on the method level?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoInsertMethodGenerator.java
```diff
@@ -122,6 +124,17 @@ public DaoInsertMethodGenerator(
...
+          "boundStatementBuilder = populateBoundStatementWithAttributes(boundStatementBuilder, $S, $S, $S, $L, $L, $L, $S)",
+          attributes.ExecutionProfileName(),
+          attributes.consistencyLevel(),
+          attributes.serialConsistencyLevel(),
+          attributes.idempotence(),
+          attributes.pageSize(),
+          attributes.timeout(),
+          attributes.routingKeyspace());
+    }
     if (statementAttributesParam != null) {
```
tomekl007: should the `statementAttributesParam` programmatic values override those specified by the annotation?
Maybe we should log a warning if both mechanisms are used on the same method?

GregBestland: The statementAttributesParam should definitely override the Attributes set at compile time, provided they are being explictly set. As Olivier mentioned above right now we have  aproblem where the defaults of statementAttributesParam will override ones that are explictly set by Attributes, which is the real problem here.  


Here is how I think it should work.
1. Attributes are applied, if they aren't specific defaults are used.
2. statementAttributes are applied... their defaults should not override Specific Attributes that are set.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Attributes.java
GregBestland: Yes only on the method level.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
GregBestland: I don't think that's the right way to distribute the tests. The attribute, attribute statement tests all leverage a specific configuration and setup in simulacron. Replicating this setup in the other various unit tests would be costly IMHO.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
GregBestland: This has been reworked.

But to answer your original question, if not set these are the default values the builder would use anyway.

tomekl007:
@GregBestland I think that you need to do rebase on java2078 - right now it shows too much changes

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
```diff
@@ -215,6 +216,21 @@ protected void maybeAddTimestamp(String timestamp, MethodSpec.Builder methodBuil
     }
   }
 
+  protected void populateBuilderWithStatementAttribute(
+      MethodSpec.Builder builder, WithStatementAttributes withStatementAttributes) {
```
olim7t: 1. `methodElement` is accessible from this class, so maybe you can extract the annotation here instead of passing it as a parameter.
2. nit: should be populateBuilderWithStatementAttribute**s** to be consistent

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
```diff
@@ -114,35 +118,105 @@ protected static SimpleStatement replaceKeyspaceAndTablePlaceholders(
...
-            .setQueryTimestamp(attributes.getQueryTimestamp())
-            .setNode(attributes.getNode())
-            .setRoutingToken(attributes.getRoutingToken());
-    for (String customPayloadKey : attributes.getCustomPayload().keySet()) {
-      builder =
-          builder.addCustomPayload(
-              customPayloadKey, attributes.getCustomPayload().get(customPayloadKey));
+    if (attributes.getExecutionProfileName().isPresent()) {
+      builder = builder.setExecutionProfileName(attributes.getExecutionProfileName().get());
+    }
```
olim7t: You can simplify this as:
```java
attributes.getExecutionProfileName().ifPresent(builder::setExecutionProfileName);
```
You just need to make `builder` final (for the custom payload loop you don't need to reassign builder, we know it's mutable).

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
```diff
@@ -114,35 +118,105 @@ protected static SimpleStatement replaceKeyspaceAndTablePlaceholders(
...
+      String profileName,
+      String consistencyLevel,
+      String serialConsistencyLevel,
+      boolean idempotent,
+      int pageSize,
+      String timeout,
+      String keyspace) {
+
+    Duration duration = null;
+    if (!timeout.equals("")) {
```
olim7t: nit: `timeout.isEmpty()`

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DefaultStatementAttributes.java
```diff
@@ -80,88 +81,88 @@ public DefaultStatementAttributes(
 
   @Nullable
   @Override
-  public DriverExecutionProfile getExecutionProfile() {
-    return profile;
+  public Optional<DriverExecutionProfile> getExecutionProfile() {
+    return Optional.ofNullable(profile);
```
olim7t: I know IDEA complains about it, but I think we should have `Optional<...>` fields instead of rewrapping on every call. I can easily see someone storing a `StatementAttributes` in a constant to reduce GC pressure, we should not reallocate every time it's used.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/StatementAttributes.java
```diff
@@ -49,8 +48,7 @@
    *
    * @see DriverExecutionProfile
    */
-  @Nullable
-  DriverExecutionProfile getExecutionProfile();
+  Optional<DriverExecutionProfile> getExecutionProfile();
```
olim7t: Can you annotate all those methods as `@NonNull`?

olim7t (changes requested):
Last few remarks.

Please also update the docs:
* javadocs of `withStatementAttributes`
* `manual/mapper/daos/statement_attributes/README.md`: add a new section "As an annotation"

integration-tests/src/test/java/com/datastax/oss/driver/mapper/StatementAttributesIT.java
```diff
@@ -78,6 +75,88 @@ public void setup() {
...
+    SimpleDao simpleDao = inventoryMapper.simpleDao(sessionRule.keyspace());
+    StatementAttributes attributes = buildRunTimeAttributes();
+    simulacronRule.cluster().clearLogs();
+    simpleDao.save(simple, attributes);
+    ClusterQueryLogReport report = simulacronRule.cluster().getLogs();
+    validateQueryOptions(report.getQueryLogs().get(0), true);
+  }
+
+  @Test
+  public void should_honor_attributes_insert() {
```
tomekl007: maybe you could change the name of this test to: `should_honor_annotation_attributes_insert` (ant other tests that use: `@WithStatementAttributes`) for better readability?

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/WithStatementAttributes.java
```diff
@@ -0,0 +1,40 @@
...
+@Retention(RetentionPolicy.CLASS)
+public @interface WithStatementAttributes {
+
+  String executionProfileName() default "";
+
+  int pageSize() default Integer.MIN_VALUE;
+
+  boolean idempotence() default false;
+
+  String consistencyLevel() default "";
```
tomekl007: Why we are not using `com.datastax.oss.driver.api.core.ConsistencyLevel` here instead of a `String`?

integration-tests/src/test/java/com/datastax/oss/driver/mapper/StatementAttributesIT.java
```diff
@@ -198,11 +243,35 @@ private void validateQueryOptions(QueryLog log) {
...
+      serialConsistencyLevel = "QUORUM",
+      pageSize = 13
+    )
+    void delete2(Simple simple);
+
     @Select
     Simple findByPk(UUID pk, StatementAttributes attributes);
+
+    @Select
+    @WithStatementAttributes(
```
tomekl007: Are there any edge cases that exclude each other in the `@WithStatementAttributes` configuration?
For example, do `serialConsistency` can be higher than `consistencyLevel`?
I am asking if we should add some kind of validation or we want to pass those properties to actual statements as they are

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/WithStatementAttributes.java
GregBestland: Annotations don't support complex objects as data types.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DefaultStatementAttributes.java
GregBestland: I just do what my IDE tells me. I'll make the change

integration-tests/src/test/java/com/datastax/oss/driver/mapper/StatementAttributesIT.java
olim7t: No, the attributes are independent from each other.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
```diff
@@ -42,6 +42,8 @@
   protected final ExecutableElement methodElement;
   protected final DaoImplementationSharedCode enclosingClass;
   protected final ProcessorContext context;
+  protected final String FUNCTION_STATEMENT_BUILDER_TYPE =
+      "java.util.function.Function<com.datastax.oss.driver.api.core.cql.BoundStatementBuilder,com.datastax.oss.driver.api.core.cql.BoundStatementBuilder>";
```
GregBestland: I do no like this. I tried for a few hours to come up with a better method for type matching than String comparison and came up empty handed. It's relatively trivial to match A TypeMirror object to a Class provided generics are not involved. I was able to figure out how introspect and check the function generics  parameters against a class. The problem was that the TypeMirror object representing the Function parameter would not match against Function.class, and I was not able to construct a TypeMirror object that included the generics. If anyone has a better way to do this I'm all ears.  The current solution is a kludge but it should work.

olim7t: You need to extract the element from the mirror to match the raw type, and then match the type parameters separately. I think this will do it:
```java
VariableElement parameter = ...
TypeMirror mirror = parameter.asType();
if (mirror.getKind() == TypeKind.DECLARED) {
  DeclaredType declaredType = (DeclaredType) mirror;
  if (context.getClassUtils().isSame(declaredType.asElement(), Function.class)
      && context.getClassUtils().isSame(declaredType.getTypeArguments().get(0), BoundStatementBuilder.class)
      && context.getClassUtils().isSame(declaredType.getTypeArguments().get(1), BoundStatementBuilder.class)) {
    // match
  }
}
```

GregBestland: This worked, it occurred to me to use The DeclaredType to match the TypeArguments of the function, but I didn't understand that isSame would would with the element but not with TypeMirror.

olim7t (approved):

PR #1258, opened 2019-06-04
JAVA-2276: Document how to configure mapper with Lombok

olim7t:


manual/mapper/config/README.md
```diff
@@ -31,6 +31,8 @@ configuration (make sure you use version 3.5 or higher):
       <artifactId>maven-compiler-plugin</artifactId>
       <version>3.8.1</version>
       <configuration>
+        <source>1.8</source> <!-- (or higher) -->
+        <target>1.8</target> <!-- (or higher) -->
```
olim7t, line 37: Not directly related, but I noticed the build fails without this, better be copy-paste safe.

manual/mapper/config/README.md
```diff
@@ -76,3 +78,95 @@ dependencies {
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.Setter;
+import lombok.ToString;
+
+@Entity
+@EqualsAndHashCode
+@ToString
+public class Product {
+  @PartitionKey @Getter @Setter private int id;
```
tomekl007, line 103: If we are saying that ordering of annotation processors then maybe in this example we should follow the same ordering? so firstly Lombok annotations, and them ours:
```
@EqualsAndHashCode
@ToString
@Entity
public class Product {
 @Getter @Setter @PartitionKey private int id;
```

manual/mapper/config/README.md
```diff
@@ -76,3 +78,95 @@ dependencies {
...
+
+The mapper can process Lombok-annotated classes just like regular code. The only requirement is that
+Lombok's annotation processor must run *before* the mapper's.
+
+With Maven, declaring Lombok as a provided dependency is not enough; you must also redeclare it in
+the `<annotationProcessorPaths>` section, before the mapper:
+
+```xml
+<properties>
+  <java-driver.version>4.0.1-SNAPSHOT</java-driver.version>
```
tomekl007: should we use SNAPSHOT in docs?

manual/mapper/config/README.md
olim7t: The declaration order of the annotations does not matter, so I see no reason to imply that in our examples.

manual/mapper/config/README.md
olim7t: Good catch, I will fix that.

PR #1261, opened 2019-06-05
JAVA-2261: Fix reconnection jitter test validation

emerkle826:


core/src/test/java/com/datastax/oss/driver/internal/core/connection/ExponentialReconnectionPolicyTest.java
```diff
@@ -53,14 +53,19 @@ public void setup() {
...
     ReconnectionPolicy.ReconnectionSchedule schedule = policy.newControlConnectionSchedule(false);
+    // keep track of the previously calculated exponential delay to avoid long overflows
+    long previousDelay = 0L;
     // generate a number of delays and make sure they are all within the base/max values range
     for (int i = 0; i < 128; ++i) {
-      // compute the min and max delays based on attempt count (i)
-      long exponentialDelay = Math.min(baseDelay * (1L << i), maxDelay);
+      // compute the min and max delays based on attempt count (i) and prevent long overflows
+      long exponentialDelay =
+          (previousDelay >= maxDelay) ? maxDelay : Math.min(baseDelay * (1L << i), maxDelay);
```
emerkle826: Not sure if this is the best way to fix the test. It's essentially short-circuiting the attempt loop. Once the loop count gets up to 54, the bit shift (`1L << i`) creates a long value that overflows Java long when multiplied by 115 below (to calculate the maximum possible jitter delay). The logic here just caps the exponential delay to the `maxDelay` so that the jitter calculation doesn't overflow the `minJitterDelay` and `maxJitterDelay` values used to ensure the scheduled delay is within the range it should be.
Now that I think about it, maybe the test should just make 50 attempts and be done.

olim7t (approved):

PR #1262, opened 2019-06-06
JAVA-2289: Revisit keyspace and table handling in entity helpers

olim7t:
- resolve the final values (context or default) in the constructor and
  cache the results in fields.
- expose this information on the public interface (might be useful for
  custom query providers).
- use the default keyspace in `@Query` methods.
- simplify error handling in `@Query` methods.


olim7t:
Don't review yet, I'm making changes to this.

olim7t:
OK for review now. I solved JAVA-2290 as well because that makes the checks much easier to implement.
Also I split `QueryIT` into 3 different classes, it was getting too hard to maintain.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/entity/EntityHelper.java
```diff
@@ -214,4 +220,37 @@
...
+  /**
+   * The table used in the queries generated by this helper.
+   *
+   * <p>This is determined by the following rules:
+   *
+   * <ul>
+   *   <li>If the DAO that this helper belongs to was created with a table (see {@link DaoTable}),
+   *       use that;
+   *   <li>Otherwise, if the entity class is annotated with {@link CqlName}, use that;
+   *   <li>Otherwise, use the name of the entity class, transformed by its {@link NamingStrategy}.
```
tolbertam, line 248: javadoc plugin doesn't like that ul isn't closed:

```
10:55:42 [ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.0.1:jar (attach-javadocs) on project java-driver-mapper-runtime: MavenReportException: Error while generating Javadoc:
10:55:42 [ERROR] Exit code: 1 - /home/jenkins/workspace/datastax.java-driver.java2289.default/736569c2/mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/entity/EntityHelper.java:244: error: element not closed: ul
10:55:42 [ERROR] * <ul>
```

tolbertam (approved):
lgtm :+1:

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/entity/EntityHelper.java
olim7t: That's an oversight, will fix.

GregBestland (approved):
Checked out the branch, Ran all the tests locally, and compared generated code before and after changes. LGTM. Much cleaner now.  I can't find anything wrong with this :+1:

PR #1263, opened 2019-06-06
JAVA-2279: Add logs to the generated mapper code

tomekl007:


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/MapperProcessor.java
```diff
@@ -40,16 +40,20 @@
 
 @AutoService(Processor.class)
 public class MapperProcessor extends AbstractProcessor {
+  private static final boolean DEFAULT_MAPPER_LOGS_ENABLED = false;
```
tomekl007: I think that sane default is `false`. If someone will have a problem with debugging then he may override this param passing `true` as the value

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -269,13 +274,15 @@ protected ClassName getPrincipalTypeName() {
...
-              PREPARED_STATEMENT_STAGE,
-              preparedStatement.fieldName,
-              simpleStatementName)
-          .addStatement("prepareStages.add($L)", preparedStatement.fieldName);
+      initAsyncBuilder.addCode(
+          "$T $L = prepare($L, context)",
+          PREPARED_STATEMENT_STAGE,
+          preparedStatement.fieldName,
+          simpleStatementName);
+      loggingGenerator.generateWhenCompletePrepare(initAsyncBuilder, preparedStatement.fieldName);
```
tomekl007: log `whenComplete` on every prepared statement

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -299,12 +306,18 @@ protected ClassName getPrincipalTypeName() {
...
-        .addCode(")\n$<$<.toCompletableFuture();$]\n")
+    MethodSpec.Builder builder =
+        initAsyncBuilder
+            .addComment("Build the DAO when all statements are prepared")
+            .addCode("return $T.allSuccessful(prepareStages)", CompletableFutures.class)
+            .addCode("\n.thenApply(v -> ($T) ", interfaceElement)
+            .addCode(newDaoStatement.build())
+            .addCode(")");
+
+    loggingGenerator.generateWhenCompleteDao(builder, interfaceElement);
```
tomekl007: log `whenComplete` dao creation

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGeneratorTest.java
```diff
@@ -49,4 +63,77 @@ public void should_ignore_default_methods() {
...
+  }
+
+  @Test
+  @UseDataProvider("disabledLoggingOptions")
+  public void should_compile_with_logging_disabled(String loggingOption) {
+    Compilation compilation =
+        should_succeed_without_warnings(
+            "test",
+            Collections.singletonList(
+                "-Acom.datastax.oss.driver.mapper.logs.enabled=" + loggingOption),
```
tomekl007: Test that we are not generating logging code if logging `enabled = false`

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGeneratorTest.java
```diff
@@ -49,4 +63,77 @@ public void should_ignore_default_methods() {
             .addModifiers(Modifier.PUBLIC, Modifier.DEFAULT)
             .build());
   }
+
+  @Test
+  public void should_compile_with_logging_enabled() {
+    Compilation compilation =
+        should_succeed_without_warnings(
+            "test",
+            Collections.singletonList("-Acom.datastax.oss.driver.mapper.logs.enabled=true"),
```
tomekl007: Test that we are generating logging code if logging `enabled = true`

tomekl007:
The main problem (and question) is about testing logging behavior when passing `-Acom.datastax.oss.driver.mapper.logs.enabled=true or false`
I've created tests using compile testing library: https://github.com/datastax/java-driver/pull/1263/files#diff-a081909d4b0aa84a097fa16722727aeeR68
Another option is to pass that `logs.enabled` the parameter to the maven compiler arguments but this way we won't be able to test both combinations easily (`true` AND `false`)
But maybe test that uses:
```
Compiler.javac()
        .withProcessors(new MapperProcessor())
        .withOptions(options)
        .compile(files);
```
is enough. Please let me know wdyt @olim7t

olim7t:
Rebased on latest java2078. Tomasz is out of office so I'm going to take over from here.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/MapperProcessor.java
olim7t: I'm torn on this.
I suggested false on the ticket, but now that I see it, processor arguments are super user-unfriendly. I have the feeling that most first-time mapper users will benefit from having the logs. And the only user-specific stuff we are going to log are the prepared queries, which are not the most sensitive information (I would say the actual data is).
So maybe true would be a better default after all.

adutra:
Two general remarks:

1. Couldn't we consider that logging is always enabled and let users decide whether or not to activate log messages using standard logging library mechanisms? (IOW, activating/deactivating specific loggers in logback.xml or log4j.properties).
2. If we are concerned with security (logging CQL queries or sensitive information, etc.), maybe we could use SLF4J markers to tag log messages coming from the mapper. This is precisely [one of the intended uses of markers](https://stackoverflow.com/questions/16813032/what-are-markers-in-java-logging-frameworks-and-what-is-a-reason-to-use-them) in SLF4J / Logback.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/DefaultProcessorContext.java
```diff
@@ -98,4 +101,9 @@ public CodeGeneratorFactory getCodeGeneratorFactory() {
   public EntityFactory getEntityFactory() {
     return entityFactoryRef.get();
   }
+
+  @Override
+  public boolean getLogsEnabled() {
```
adutra: Standard JavaBeans conventions mandate that this should be named `isLogsEnabled`.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/ProcessorContext.java
```diff
@@ -36,4 +36,6 @@
   CodeGeneratorFactory getCodeGeneratorFactory();
 
   EntityFactory getEntityFactory();
+
+  boolean getLogsEnabled();
```
adutra: Same here -> `isLogsEnabled`.

olim7t:
I think we already considered markers in the past, and our conclusion was that it was not necessarily well supported by all SLF4J backend (IIRC, log4j ignores them).

You are right that controlling categories is probably good enough, so that's another argument for defaulting to true. I think we can leave the processor property since it's pretty simple to have it.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/MapperProcessor.java
tolbertam: Agreed, true seems like the right default :+1:

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
tolbertam: Would it be worth also adding a log statement before executing the prepare?  The might be helpful in tracking something like long prepare delays or something like that.

(Same comment for when we start initializing the DAO)

tolbertam (approved):
:+1: on the approach, had a small suggestion about logging before beginning dao initialization and prepare, but fine with things either way.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
olim7t: Yeah I agree, I'm doing a few changes around that. I think we also shouldn't log errors since they are rethrown.

olim7t:
Force-pushed:
- changed the default to logging enabled
- log before preparing, include the query string
- added log prefix

manual/mapper/README.md
```diff
@@ -152,3 +152,22 @@ From the mapper, you can then obtain a DAO instance and execute queries:
...
+the same way as the [core driver logs](../core/logging/).
+
+They can help you figure out which queries the mapper is generating under the hood, for example:
+
+```
+DEBUG ProductDaoImpl__MapperGenerated - [s0] Initializing new instance for keyspace = ks_0 and table = null
+DEBUG ProductHelper__MapperGenerated - [s0] Entity Product will be mapped to ks_0.product
+DEBUG ProductDaoImpl__MapperGenerated - [s0] Preparing query
+      `SELECT id,description,dimensions FROM ks_0.product WHERE id=:id`
+      for method findById(java.util.UUID)
```
olim7t, line 168: That's all we log for now, I can't think of anything else useful.

olim7t (approved):

PR #1265, opened 2019-06-11
Document mapper usage with Kotlin

olim7t:


manual/mapper/config/README.md
```diff
@@ -175,3 +175,82 @@ dependencies {
...
+```kotlin
+@Entity
+data class Product(@PartitionKey var id: Int? = null, var description: String? = null)
+```
+
+All of the [property annotations](../entities/#property-annotations) can be declared directly on the
+constructor properties.
+
+If you want to take advantage of [null saving strategies](../daos/null_saving/), your properties
+should be nullable. If you don't mind inserting everything all the time, you can make them non-null,
```
tomekl007: >you can make them non-null,

Isn't that contradicts 

>you can make them non-null,

?
So to make data classes work the client needs to declare them as var (nullable)

olim7t: I need to reformulate that. What I meant is that if you want to use null saving strategies you should make properties nullable, but if you don't care about null saving strategies then you can make them non-null.
I think nullable is a better fit with the mapper's concepts though.

manual/mapper/config/README.md
```diff
@@ -175,3 +175,82 @@ dependencies {
...
+    apply plugin: 'kotlin'
+    apply plugin: 'kotlin-kapt'
+    
+    dependencies {
+        kapt group: 'com.datastax.oss', name: 'java-driver-mapper-processor', version: javaDriverVersion
+        ...
+    }
+    ```
+
+You can use Kotlin [data classes] for your entities. Just keep in mind that the mapper expects a
```
tolbertam, line 236: Neat!  Tried this out (with maven) and it worked!  Thought kotlin may be a problem but cool that it works.  Would be nice if we could use `val`, but hopefully whatever we come up with for [JAVA-2125](https://datastax-oss.atlassian.net/browse/JAVA-2125) provides a path for that to work.

manual/mapper/config/README.md
```diff
@@ -175,3 +175,82 @@ dependencies {
...
+#### Kotlin
+
+[Kotlin](https://kotlinlang.org/) is an alternative language for the JVM. Its compact syntax and
+native support for annotation processing make it a good fit for the mapper.
+
+To set up your project, refer to the Kotlin website:
+
+* Maven: configure [dual compilation][maven_kotlin_java] of Kotlin and Java sources. In addition,
+  you'll need an additional execution of the [kotlin-maven-plugin:kapt][maven_kapt] goal with the
+  mapper processor before compilation:
```
tolbertam, line 188: Would be nice to clarify that this execution belongs under `kotlin-maven-plugin`.

tolbertam (approved):
:+1:

PR #1267, opened 2019-06-13
JAVA-2288: Improve error reporting when DAO is missing keyspace

tolbertam:
For [JAVA-2288](https://datastax-oss.atlassian.net/browse/JAVA-2288)

---

Motivation:

If no keyspace can be discerned from the session's keyspace, the
entity's default keyspace, or the dao's parameters, the mapper should
preemptively throw a RuntimeException with a helpful message to guide
the user to provide a keyspace.

Modifications:

Added throwIfKeyspaceMissing call in Dao initAsync method to check if
keyspace is set.

Result:

User is now presented with a helpful exception message when keyspace is
not configured.

### DIFF
--- a/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
+++ b/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
@@ -417,7 +417,6 @@ public class DaoImplementationGenerator extends SingleFileCodeGenerator
       String fieldName = entry.getValue();
       // - create an instance
       initAsyncBuilder.addStatement("$1T $2L = new $1T(context)", fieldTypeName, fieldName);
-      initAsyncBuilder.addStatement("throwIfKeyspaceMissing($L, context.getSession())", fieldName);
       // - add it as a parameter to the constructor call
       newDaoStatement.add(",\n$L", fieldName);
     }
--- a/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
+++ b/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
@@ -417,6 +417,7 @@ public class DaoImplementationGenerator extends SingleFileCodeGenerator
       String fieldName = entry.getValue();
       // - create an instance
       initAsyncBuilder.addStatement("$1T $2L = new $1T(context)", fieldTypeName, fieldName);
+      initAsyncBuilder.addStatement("throwIfKeyspaceMissing($L, context.getSession())", fieldName);
       // - add it as a parameter to the constructor call
       newDaoStatement.add(",\n$L", fieldName);
     }
--- a/mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
+++ b/mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
@@ -26,14 +26,10 @@ import com.datastax.oss.driver.api.core.cql.ResultSet;
 import com.datastax.oss.driver.api.core.cql.Row;
 import com.datastax.oss.driver.api.core.cql.SimpleStatement;
 import com.datastax.oss.driver.api.core.cql.Statement;
-import com.datastax.oss.driver.api.core.session.Session;
 import com.datastax.oss.driver.api.core.type.DataTypes;
 import com.datastax.oss.driver.api.mapper.MapperContext;
 import com.datastax.oss.driver.api.mapper.MapperException;
 import com.datastax.oss.driver.api.mapper.annotations.Dao;
-import com.datastax.oss.driver.api.mapper.annotations.DaoFactory;
-import com.datastax.oss.driver.api.mapper.annotations.DaoKeyspace;
-import com.datastax.oss.driver.api.mapper.annotations.Entity;
 import com.datastax.oss.driver.api.mapper.annotations.Query;
 import com.datastax.oss.driver.api.mapper.entity.EntityHelper;
 import com.datastax.oss.driver.api.mapper.entity.saving.NullSavingStrategy;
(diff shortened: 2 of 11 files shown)


mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
```diff
@@ -270,4 +274,19 @@ protected static void throwIfProtocolVersionV3(MapperContext context) {
               NullSavingStrategy.class.getSimpleName(), NullSavingStrategy.DO_NOT_SET.name()));
     }
   }
+
+  protected static <EntityT> void throwIfKeyspaceMissing(
```
tolbertam: The jira ticket called for an error message like:

```
Missing keyspace for userDao(). Suggestions: use SessionBuilder.withKeyspace() when creating your session, specify a default keyspace on User with @Entity(defaultKeyspace), or use a DAO factory method with an @DaoKeyspace parameter
```

I was almost able to match this 1:1 except for providing the dao method name.  Fortunately that shows up in the stack trace.  I think we could probably add some context to resolve it, but wasn't sure how important that was.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -417,6 +417,7 @@ protected ClassName getPrincipalTypeName() {
       String fieldName = entry.getValue();
       // - create an instance
       initAsyncBuilder.addStatement("$1T $2L = new $1T(context)", fieldTypeName, fieldName);
+      initAsyncBuilder.addStatement("throwIfKeyspaceMissing($L, context.getSession())", fieldName);
```
tolbertam: One thought is that in theory, a user could only specify Dao methods that don't require a keyspace, like `@GetEntity` and `@SetEntity`, we could choose not to fail in this case, but not sure if it's worth worrying about?

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
olim7t: 👍 if it appears in the stack trace that should be clear enough.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
olim7t: We could move the check to each entity helper method, we have the context there so I think we have everything we need. Another advantage of that is that the check would also kick in when you call a helper method from a query provider.

We can introduce an `EntityHelperBase` parent class to factor the method.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
```diff
@@ -285,4 +289,19 @@ protected static void throwIfProtocolVersionV3(MapperContext context) {
...
+      EntityHelper<EntityT> helper, Session session) {
+    if (helper.getKeyspaceId() == null && !session.getKeyspace().isPresent()) {
+      throw new MapperException(
+          String.format(
+              "Missing keyspace. Suggestions: use SessionBuilder.withKeyspace() "
+                  + "when creating your session, specify a default keyspace on %s with @%s"
+                  + "(defaultKeyspace), or use a @%s method with a @%s parameter",
+              helper.getEntityClass(),
```
olim7t: In similar messages (see `replaceKeyspaceAndTablePlaceholders`), I've used the entity's simple name. My reasoning was that it would be pretty obvious from the context which class it is.
I don't mind either way but we should be consistent, if you think it's worth having the package we'll modify the other message.

olim7t (changes requested):
LGTM, but I think doing the check at the entity helper level (per my other comment) is not too much extra work and would be more precise.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
tolbertam: yeah this seems easy enough to do.  I had initially went about it this way, but seemed like validating in each of these methods was less simple than validating it on entity helper creation. I'll change it.

tolbertam: Adding `EntityHelperBase` also eliminates some boilerplate in the generated code, so that looks like a good change overall.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
tolbertam: Fixed that to use `getSimpleName()` :+1:

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/entity/EntityHelperBase.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }
+
+  @Override
+  public CqlIdentifier getKeyspaceId() {
+    return keyspaceId;
+  }
+
+  @Override
+  public CqlIdentifier getTableId() {
```
olim7t, line 61: Missing `@NonNull` annotation.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/entity/EntityHelperBase.java
```diff
@@ -0,0 +1,75 @@
...
+
+  protected final CqlIdentifier tableId;
+
+  protected final MapperContext context;
+
+  protected EntityHelperBase(MapperContext context, String tableName) {
+    this(context, null, tableName);
+  }
+
+  protected EntityHelperBase(MapperContext context, String keyspaceName, String tableName) {
```
olim7t: Maybe name those parameters `defaultKeyspaceName` and `defaultTableName`.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/entity/EntityHelperBase.java
```diff
@@ -0,0 +1,75 @@
...
+    this.tableId =
+        context.getTableId() != null ? context.getTableId() : CqlIdentifier.fromCql(tableName);
+    if (keyspaceName == null) {
+      this.keyspaceId = context.getKeyspaceId();
+    } else {
+      this.keyspaceId =
+          context.getKeyspaceId() != null
+              ? context.getKeyspaceId()
+              : CqlIdentifier.fromCql(keyspaceName);
+    }
```
olim7t: I'm generally not a big fan of chained ternary operators, but in this case this might be a bit easier to follow:
```java
    this.keyspaceId =
        context.getKeyspaceId() != null
            ? context.getKeyspaceId()
            : (keyspaceName == null ? null : CqlIdentifier.fromCql(keyspaceName));
```

olim7t (approved):
👍

tolbertam:
CI looks happy after recent changes, mergin' away

PR #1268, opened 2019-06-17
JAVA-2266 Run mapper tests for C* 2.1 when possible

tolbertam:
Opening a little early just to see how CI behaves... still more work to do here, as we have a lot of tests that set min version 2.2 to avoid the default null saving behavior.


integration-tests/src/test/java/com/datastax/oss/driver/mapper/NullSavingStrategyBehaviorIT.java
```diff
@@ -0,0 +1,413 @@
...
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+
+@Category(ParallelizableTests.class)
+@CassandraRequirement(min = "2.2", description = "support for unset values")
+public class NullSavingStrategyBehaviorIT extends InventoryITBase {
```
olim7t: 👍 
I also separated null saving strategy tests in `QueryNullStrategyIT` (which btw uses a protocol version check instead of the `CassandraRequirement` annotation, we might want to change that).

Maybe we could merge the two suites. I think the way I did it in the other class is easier to follow, and it should be pretty easy to add other methods.

tolbertam: sure, I can do that.  I wanted to identify everything i needed to pull out of the existing tests first, and now that I've got that sorted out I can focus on the null-specific tests themselves, will take care of that first thing tomorrow.

tolbertam: Made the suggested change, let me know how that looks :+1:

integration-tests/src/test/java/com/datastax/oss/driver/mapper/DefaultNullSavingStrategyIT.java
```diff
@@ -0,0 +1,331 @@
...
+        break;
+      default:
+        throw new AssertionError("unhandled strategy " + expectedStrategy);
+    }
+  }
+
+  @Dao
+  public interface DaoWithNoStrategy {
+    @Query("INSERT INTO foo (k, v) values (:k, :v)")
+    void insertWithNoStrategy(Integer k, Integer v);
```
olim7t: For clarity, maybe we should rename:
```
insertWith... => queryWith...
saveWith... => insertWith...
```

olim7t (approved):
👍

integration-tests/src/test/java/com/datastax/oss/driver/mapper/DefaultNullSavingStrategyIT.java
tolbertam: will do, thanks for reviewing :)

PR #1269, opened 2019-06-19
Small enhancements to the mapper

adutra:
This is required by JAVA-2296.

### DIFF
--- a/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
+++ b/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
@@ -41,6 +41,7 @@ import com.squareup.javapoet.MethodSpec;
 import com.squareup.javapoet.ParameterizedTypeName;
 import com.squareup.javapoet.TypeName;
 import com.squareup.javapoet.TypeSpec;
+import edu.umd.cs.findbugs.annotations.NonNull;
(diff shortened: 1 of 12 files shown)


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
```diff
@@ -298,7 +299,7 @@ protected ClassName getPrincipalTypeName() {
         TypeSpec.classBuilder(implementationName)
             .addJavadoc(JAVADOC_GENERATED_WARNING)
             .addModifiers(Modifier.PUBLIC)
-            .superclass(DaoBase.class)
+            .superclass(getDaoParentClass())
```
tolbertam, line 302: This might be useful to do in `EntityHelperGenerator.getContents()` as well as we follow a similar pattern there with using an `EntityHelperBase` for `EntityHelper` implementations.

tolbertam (approved):
:+1:

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoImplementationGenerator.java
adutra: Good idea, will do.

olim7t:
This got closed when I deleted `java2078`, reopening as #1278.

PR #1270, opened 2019-06-19
JAVA-2297: Set socket options from NettyOptions#afterBootstrapInitialized

olim7t:
Motivation:

Setting socket options on the bootstrap is currently hardcoded in
ChannelFactory. This can cause warnings if the transport is overridden
with an implementation that does not support some of these options (ex:
TCP_NODELAY for Unix domain sockets).

Modifications:

Move the code to NettyOptions#afterBootstrapInitialized.

Result:

Custom NettyOptions implementations can decide which socket options to
set. Note that they will now have to copy over the code from
DefaultNettyOptions to get the default behavior.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
@@ -22,8 +22,10 @@ import com.datastax.oss.driver.shaded.guava.common.util.concurrent.ThreadFactory
 import io.netty.bootstrap.Bootstrap;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.channel.Channel;
+import io.netty.channel.ChannelOption;
 import io.netty.channel.DefaultEventLoopGroup;
 import io.netty.channel.EventLoopGroup;
+import io.netty.channel.FixedRecvByteBufAllocator;
 import io.netty.channel.nio.NioEventLoopGroup;
 import io.netty.channel.socket.nio.NioSocketChannel;
 import io.netty.util.HashedWheelTimer;
@@ -46,6 +48,7 @@ public class DefaultNettyOptions implements NettyOptions {
 
   private static final Logger LOG = LoggerFactory.getLogger(DefaultNettyOptions.class);
 
+  private final DriverExecutionProfile config;
   private final EventLoopGroup ioEventLoopGroup;
   private final EventLoopGroup adminEventLoopGroup;
   private final int ioShutdownQuietPeriod;
@@ -57,7 +60,7 @@ public class DefaultNettyOptions implements NettyOptions {
   private final Timer timer;
 
   public DefaultNettyOptions(InternalDriverContext context) {
-    DriverExecutionProfile config = context.getConfig().getDefaultProfile();
+    this.config = context.getConfig().getDefaultProfile();
     int ioGroupSize = config.getInt(DefaultDriverOption.NETTY_IO_SIZE);
     this.ioShutdownQuietPeriod = config.getInt(DefaultDriverOption.NETTY_IO_SHUTDOWN_QUIET_PERIOD);
     this.ioShutdownTimeout = config.getInt(DefaultDriverOption.NETTY_IO_SHUTDOWN_TIMEOUT);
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/context/NettyOptions.java
```diff
@@ -62,6 +66,45 @@
...
+   *
+   * <p>Implementation note: this is very similar to {@link #afterBootstrapInitialized}; the only
+   * reason to have two separate methods is backward compatibility.
+   *
+   * @param bootstrap the bootstrap to configure.
+   * @param defaultProfile the default profile of the driver's configuration (where the values to
+   *     set will typically be read from).
+   * @since 4.1.0
+   */
+  default void setSocketOptions(Bootstrap bootstrap, DriverExecutionProfile defaultProfile) {
```
tolbertam: One thing sort of awkward about this is that this method provides `DriverExecutionProfile` where the others don't. 

I was thinking maybe we could overload `afterBootstrapInitialized` with another method providing the profile as a parameter, but it gets sort of confusing to have two methods with the same name, and as you explained we don't want to break existing custom implementations which implicitly depended on the options to be set otherwise, so I think what you did here is probably our best option.

tolbertam (approved):
:+1:

core/src/main/java/com/datastax/oss/driver/internal/core/context/NettyOptions.java
```diff
@@ -62,6 +66,45 @@
...
+    boolean tcpNoDelay = defaultProfile.getBoolean(DefaultDriverOption.SOCKET_TCP_NODELAY);
+    bootstrap.option(ChannelOption.TCP_NODELAY, tcpNoDelay);
```
emerkle826: Do we want to use the same `isDefined()` check here for the Unix socket case, or is the idea that when using Unix sockets, you would provide a different implementation for `setSocketOptions` that does not try to set TCP_NODELAY?

emerkle826: I commented too soon. We are doing it this way to keep the functionality backward compatible, correct?

emerkle826 (approved):
Ignore my comments, I overthought the change too much. :+1:

core/src/main/java/com/datastax/oss/driver/internal/core/context/NettyOptions.java
olim7t: > One thing sort of awkward about this is that this method provides `DriverExecutionProfile` where the others don't.

Yes, that's the only way I found to provide a default implementation.

adutra: Do we really expect people extending or implementing `NettyOptions`? This being an internal class, I would be in favor of being more agressive with backwards compatibility.

olim7t: Very good point, for some reason I was assuming it was in the public API and didn't even check the package name.
You're right, it's fair to be a bit more aggressive here, I'll do the changes.

olim7t:
Force-pushed to do it in `afterBootstrapInitialized`. I added a note in the upgrade guide since this is still a relatively high-level feature.

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultNettyOptions.java
```diff
@@ -131,7 +134,30 @@ public ByteBufAllocator allocator() {
...
+          .option(ChannelOption.SO_RCVBUF, receiveBufferSize)
+          .option(ChannelOption.RCVBUF_ALLOCATOR, new FixedRecvByteBufAllocator(receiveBufferSize));
+    }
+    if (config.isDefined(DefaultDriverOption.SOCKET_SEND_BUFFER_SIZE)) {
+      int sendBufferSize = config.getInt(DefaultDriverOption.SOCKET_SEND_BUFFER_SIZE);
+      bootstrap.option(ChannelOption.SO_SNDBUF, sendBufferSize);
+    }
   }
 
   @Override
```
adutra, line 163: Not related to this PR, but I noticed that `PromiseCombiner` has been deprecated (used in method `onClose` below). Is there anything we can do to avoid using it?

olim7t: It's just the no-arg constructor that's deprecated. I guess we should inline its implementation: `new PromiseCombiner(ImmediateEventExecutor.INSTANCE)`; we can't use any of the driver's executors since they are being shut down.
I'll add the change in a separate commit.

olim7t: Actually `GlobalEventExecutor.INSTANCE` is a better choice (see the explanations on netty/netty@625c4e8286d3c8a36046feb1289b6b710c7541a5).

PR #1271, opened 2019-06-20
JAVA-2298: Annotate parameters corresponding to bind markers

adutra:


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
```diff
@@ -198,4 +200,55 @@ protected VariableElement findBoundStatementFunction(ExecutableElement methodEle
...
+      CqlName cqlName = parameter.getAnnotation(CqlName.class);
+      if (cqlName != null) {
+        context
+            .getMessager()
+            .error(
+                parameter,
+                "Method %s: parameter %s does not refer to a bind marker "
+                    + "and must not be annotated with @CqlName",
+                methodElement,
+                parameter.getSimpleName());
```
olim7t: This is not really blocking, I think we could just warn that the name will be ignored.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -125,19 +128,31 @@ public static void addFinalFieldAndConstructorArgument(
...
+      String parameterName;
+      if (cqlName == null) {
+        parameterName = parameter.getSimpleName().toString();
+      } else {
+        parameterName = cqlName.value();
+      }
+      bindMarkerNames.add(CodeBlock.of("$S", parameterName));
+    }
+    bindParameters(
+        parameters, bindMarkerNames, methodBuilder, enclosingClass, context, useNullSavingStrategy);
```
olim7t, line 155: Nice way to handle this 👍 
I like how this simplifies the other method.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
```diff
@@ -198,4 +200,55 @@ protected VariableElement findBoundStatementFunction(ExecutableElement methodEle
...
+      for (VariableElement parameter : parameters) {
+        CqlName cqlName = parameter.getAnnotation(CqlName.class);
+        if (cqlName == null) {
+          context
+              .getMessager()
+              .error(
+                  parameter,
+                  "Method %s: parameter %s is declared in a compiled method "
+                      + "and refers to a bind marker "
+                      + "and thus must be annotated with @CqlName",
```
olim7t: I think we should also include the name of the child interface, but this can wait for [JAVA-2302](https://datastax-oss.atlassian.net/browse/JAVA-2302) (currently we don't have a reference to `DaoImplementationGenerator.interfaceElement` in this class, and that's one of the things that ticket will address).

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoInsertMethodGenerator.java
```diff
@@ -120,27 +120,33 @@ public DaoInsertMethodGenerator(
 
     populateBuilderWithStatementAttributes(methodBodyBuilder, methodElement);
     populateBuilderWithFunction(methodBodyBuilder, boundStatementFunction);
-    String entityParameterName = parameters.get(0).getSimpleName().toString();
 
-    NullSavingStrategy nullSavingStrategy =
-        nullSavingStrategyValidation.getNullSavingStrategy(
-            Insert.class, Insert::nullSavingStrategy, methodElement, enclosingClass);
+    if (validateCqlNamesAbsent(parameters.subList(0, 1))) {
```
olim7t: Side note: this doesn't validate the optional `Function<BoundStatementBuilder, BoundStatementBuilder>` argument from [JAVA-2139](https://datastax-oss.atlassian.net/browse/JAVA-2139) because it's already been removed from the list at this point. But that's not a big deal, the annotation will be ignored silently if present.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
```diff
@@ -210,12 +210,14 @@ public DaoDeleteMethodGenerator(
...
       }
-      GeneratedCodePatterns.bindParameters(
-          parameters.subList(nextParameterIndex, parameters.size()),
-          methodBodyBuilder,
-          enclosingClass,
-          context,
-          false);
+      List<? extends VariableElement> bindMarkers =
+          parameters.subList(nextParameterIndex, parameters.size());
+      if (validateCqlNamesPresent(bindMarkers)) {
```
olim7t, line 214: We should also validate the case when a full entity is provided (previous branch).

olim7t (changes requested):
LGTM.
Two minor requests, I'll amend the PR myself.

olim7t (approved):

PR #1272, opened 2019-06-21
JAVA-2300: Add support for additional return types in Insert methods

tolbertam:
For [JAVA-2300](https://datastax-oss.atlassian.net/browse/JAVA-2300)

---

Motivation:

`@Insert`-annotated DAO methods do not currently support the capability
of returning `boolean` or `ResultSet` return types.  These are useful to
determine whether or not LWT operations were applied and also to have
access to useful data associated with `ResultSet`, such as
`ExecutionInfo`.

Modifications:

Update `DaoInsertMethodGenerator.getSupportedReturnTypes()` to include
`BOOLEAN`, `FUTURE_OF_BOOLEAN`, `RESULT_SET`, and
`FUTURE_OF_RESULT_SET`.

Result:

`@Insert` methods can now return `boolean` and `ResultSet`-based types.


adutra (changes requested):
LGTM but could you amend the javadocs of `Insert` to mention the new return types?

olim7t (approved):

PR #1273, opened 2019-06-21
JAVA-2247: PagingIterable implementations should implement spliterator()

adutra:
I ran this [small benchmark](https://gist.github.com/adutra/7d6c15a403867c4e6f2d052249ecbc46), adapted from [here](https://www.airpair.com/java/posts/parallel-processing-of-io-based-data-with-java-streams), and that gave me the below results:

```
Start processing JDK stream
          Cores: 8
       CPU time: 70.60 s
      Real time: 18.00 s
CPU utilization: 49.01%

Start processing fixed-batch stream
          Cores: 8
       CPU time: 86.84 s
      Real time: 10.93 s
CPU utilization: 99.29%
```
So it does seem to improve things, specially in the case of parallel computation.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/PagingIterableWrapper.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/PagingIterableWrapper.java
@@ -48,7 +48,7 @@ public class PagingIterableWrapper<SourceT, TargetT> implements PagingIterable<T
   /**
    * Creates a {@link PagingIterableWrapper} for the given source. If {@code sized} is {@code true},
    * spliterators for this iterable will report {@link Spliterator#SIZED} and {@link
-   * Spliterator#SIZED} and their estimated size will be {@link #getAvailableWithoutFetching()}.
+   * Spliterator#SUBSIZED} and their estimated size will be {@link #getAvailableWithoutFetching()}.
    *
    * @param source The source to wrap.
    * @param elementMapper The element mapper.
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/PagingIterableWrapper.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/PagingIterableWrapper.java
@@ -18,20 +18,48 @@ package com.datastax.oss.driver.internal.core;
 import com.datastax.oss.driver.api.core.PagingIterable;
 import com.datastax.oss.driver.api.core.cql.ColumnDefinitions;
 import com.datastax.oss.driver.api.core.cql.ExecutionInfo;
+import com.datastax.oss.driver.internal.core.cql.PagingIterableSpliterator;
 import com.datastax.oss.driver.shaded.guava.common.collect.AbstractIterator;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Spliterator;
 import java.util.function.Function;
 
 public class PagingIterableWrapper<SourceT, TargetT> implements PagingIterable<TargetT> {
 
   private final PagingIterable<SourceT> source;
+  private final boolean sized;
   private final Iterator<TargetT> iterator;
 
+  /**
+   * Creates a {@link PagingIterableWrapper} for the given source, with unknown size. Spliterators
+   * for this iterable will never report {@link Spliterator#SIZED}.
+   *
+   * @param source The source to wrap.
+   * @param elementMapper The element mapper.
+   */
   public PagingIterableWrapper(
-      PagingIterable<SourceT> source, Function<? super SourceT, ? extends TargetT> elementMapper) {
+      @NonNull PagingIterable<SourceT> source,
+      @NonNull Function<? super SourceT, ? extends TargetT> elementMapper) {
+    this(source, elementMapper, false);
+  }
+
+  /**
+   * Creates a {@link PagingIterableWrapper} for the given source. If {@code size} is {@code true},
+   * spliterators for this iterable will report {@link Spliterator#SIZED} and {@link
+   * Spliterator#SIZED} and their estimated size will be {@link #getAvailableWithoutFetching()}.
+   *
+   * @param source The source to wrap.
+   * @param elementMapper The element mapper.
+   * @param sized Whether this iterable has a known size or not.
+   */
+  public PagingIterableWrapper(
+      @NonNull PagingIterable<SourceT> source,
+      @NonNull Function<? super SourceT, ? extends TargetT> elementMapper,
+      boolean sized) {
     this.source = source;
+    this.sized = sized;
     Iterator<SourceT> sourceIterator = source.iterator();
     this.iterator =
         new AbstractIterator<TargetT>() {
@@ -71,8 +99,26 @@ public class PagingIterableWrapper<SourceT, TargetT> implements PagingIterable<T
     return source.wasApplied();
   }
 
+  @NonNull
   @Override
   public Iterator<TargetT> iterator() {
     return iterator;
+  }
+
+  @NonNull
+  @Override
+  public Spliterator<TargetT> spliterator() {
+    PagingIterableSpliterator.Builder<TargetT> builder = PagingIterableSpliterator.builder(this);
+    if (sized) {
+      builder.withEstimatedSize(getAvailableWithoutFetching());
+    }
+    return builder.build();
+  }
+
+  @NonNull
+  @Override
+  public <NewTargetT> PagingIterable<NewTargetT> map(
+      Function<? super TargetT, ? extends NewTargetT> elementMapper) {
+    return new PagingIterableWrapper<>(this, elementMapper, sized);
   }
 }
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/PagingIterableWrapper.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/PagingIterableWrapper.java
@@ -46,7 +46,7 @@ public class PagingIterableWrapper<SourceT, TargetT> implements PagingIterable<T
   }
 
   /**
-   * Creates a {@link PagingIterableWrapper} for the given source. If {@code size} is {@code true},
+   * Creates a {@link PagingIterableWrapper} for the given source. If {@code sized} is {@code true},
    * spliterators for this iterable will report {@link Spliterator#SIZED} and {@link
    * Spliterator#SIZED} and their estimated size will be {@link #getAvailableWithoutFetching()}.
    *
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/cql/PagingIterableSpliterator.java
```diff
@@ -0,0 +1,167 @@
...
+   * @param <ElementT> The iterable's element type
+   * @return a new {@link Builder}.
+   */
+  @NonNull
+  public static <ElementT> Builder<ElementT> builder(@NonNull PagingIterable<ElementT> iterable) {
+    return new Builder<>(iterable);
+  }
+
+  /** The default chunk size for {@link PagingIterableSpliterator}. */
+  public static final int DEFAULT_CHUNK_SIZE = 128;
```
adutra, line 43: I went for 128 after executing [this benchmark](https://gist.github.com/adutra/853609fead6938ff87db2b7902ecc55c). It gave me results like the ones below:

```
Benchmark                            (batchSize)  Mode  Cnt      Score       Error  Units
MyBenchmark.benchmarkPagingIterable           32  avgt    3   9948.614 ±  3767.967  ms/op
MyBenchmark.benchmarkPagingIterable          128  avgt    3   9825.762 ±  1232.689  ms/op
MyBenchmark.benchmarkPagingIterable          256  avgt    3  10804.523 ±  6801.505  ms/op
MyBenchmark.benchmarkPagingIterable          512  avgt    3  10721.407 ± 13054.991  ms/op
MyBenchmark.benchmarkPagingIterable         1024  avgt    3  11438.708 ±  3053.842  ms/op
```

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/PagingIterableSpliteratorIT.java
```diff
@@ -0,0 +1,133 @@
...
+            .getContext()
+            .getConfig()
+            .getDefaultProfile()
+            .withInt(DefaultDriverOption.REQUEST_PAGE_SIZE, pageSize);
+    ResultSet result =
+        session.execute(
+            SimpleStatement.newInstance("SELECT v FROM test where k0 = 0")
+                .setExecutionProfile(profile));
+    Spliterator<Row> spliterator = result.spliterator();
+    if (pageSize > 20_000) {
```
tomekl007, line 88: why we are using `20_000` as a threshold value?

core/src/main/java/com/datastax/oss/driver/internal/core/cql/PagingIterableSpliterator.java
```diff
@@ -0,0 +1,167 @@
...
+   */
+  public PagingIterableSpliterator(@NonNull PagingIterable<ElementT> iterable) {
+    this(iterable, Long.MAX_VALUE, DEFAULT_CHUNK_SIZE);
+  }
+
+  private PagingIterableSpliterator(
+      @NonNull PagingIterable<ElementT> iterable, long estimatedSize, int chunkSize) {
+    this.iterable = Objects.requireNonNull(iterable, "iterable cannot be null");
+    this.estimatedSize = estimatedSize;
+    this.chunkSize = chunkSize;
```
tomekl007, line 64: shouldn't we add some upper bound validation for chunkSize? What if someone will pass here Integer.MAX_VALUE?
Probably we may get some out of memory exception here:
https://github.com/datastax/java-driver/pull/1273/files#diff-09b68e1e372fede23d398783e0f7b401R106

core/src/main/java/com/datastax/oss/driver/internal/core/cql/PagingIterableSpliterator.java
```diff
@@ -0,0 +1,167 @@
...
+    if (estimatedSize < Long.MAX_VALUE) {
+      characteristics =
+          Spliterator.ORDERED
```
tomekl007, line 67: do we need to make our iterator `ORDERED`?
If the order is not important maybe we could remove that flag - I suspect that it may have a potential influence on the performance (the unordered version may be faster than ordered)

core/src/main/java/com/datastax/oss/driver/internal/core/cql/PagingIterableSpliterator.java
```diff
@@ -0,0 +1,167 @@
...
+    if (row == null) {
+      return false;
+    }
+    action.accept(row);
+    return true;
+  }
+
+  @Override
+  @Nullable
+  public Spliterator<ElementT> trySplit() {
```
tomekl007, line 90: I wonder how do we relate to docs of `trySplit`:
```
  * An ideal {@code trySplit} method efficiently (without
     * traversal) divides its elements exactly in half, allowing
     * balanced parallel computation. Many departures from this ideal
     * remain highly effective;
```
I see that we are supporting a couple of ways to split (not only split in half)
Is it needed to give that support to a client?

core/src/main/java/com/datastax/oss/driver/internal/core/PagingIterableWrapper.java
```diff
@@ -18,20 +18,48 @@
...
+   */
   public PagingIterableWrapper(
-      PagingIterable<SourceT> source, Function<? super SourceT, ? extends TargetT> elementMapper) {
+      @NonNull PagingIterable<SourceT> source,
+      @NonNull Function<? super SourceT, ? extends TargetT> elementMapper) {
+    this(source, elementMapper, false);
+  }
+
+  /**
+   * Creates a {@link PagingIterableWrapper} for the given source. If {@code size} is {@code true},
```
olim7t:
```suggestion
   * Creates a {@link PagingIterableWrapper} for the given source. If {@code sized} is {@code true},
```

core/src/main/java/com/datastax/oss/driver/internal/core/PagingIterableWrapper.java
```diff
@@ -18,20 +18,48 @@
...
+   * spliterators for this iterable will report {@link Spliterator#SIZED} and {@link
+   * Spliterator#SIZED} and their estimated size will be {@link #getAvailableWithoutFetching()}.
```
olim7t:
```suggestion
   * Spliterator#SUBSIZED} and their estimated size will be {@link #getAvailableWithoutFetching()}.
```
(I think)

core/src/main/java/com/datastax/oss/driver/internal/core/cql/PagingIterableSpliterator.java
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * Creates a new {@link Builder} for {@link PagingIterableSpliterator} instances.
+   *
+   * @param iterable The {@link PagingIterable} to create a spliterator for.
+   * @param <ElementT> The iterable's element type
+   * @return a new {@link Builder}.
+   */
+  @NonNull
+  public static <ElementT> Builder<ElementT> builder(@NonNull PagingIterable<ElementT> iterable) {
```
olim7t, line 38: Nit: do we need that level of detail in javadocs for internal classes? This block doesn't really tell anything that isn't obvious from the method signature.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/PagingIterableSpliterator.java
olim7t: I think `ORDERED` simply means that we get the elements in a deterministic order.

> Characteristic value signifying that an encounter order is defined for elements.

olim7t:
So, I don't understand how client code is supposed to use this.

I tried the following in `PagingIterableSpliteratorIT`:
```java
    SimpleStatement statement =
        SimpleStatement.builder("SELECT v FROM test where k0 = 0").setPageSize(100).build();
    ResultSet resultSet = session.execute(statement);
    Spliterator<Row> split1 = resultSet.spliterator();
    Spliterator<Row> split2 = split1.trySplit();

    int count1 = 0;
    while (split1.tryAdvance(row -> {})) {
      count1 += 1;
    }
    int count2 = 0;
    while (split2.tryAdvance(row -> {})) {
      count2 += 1;
    }

    System.out.println(count1); // 19872
    System.out.println(count2); // 128
```
The two splits are not balanced at all. Which is quite obvious since you would have to consume the whole result set to figure out the middle. This circles back to the docs that Tomasz pointed to:
> An ideal trySplit method efficiently (without traversal) divides its elements exactly in half

So how would I parallel process a result set that spans multiple pages?

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/PagingIterableSpliteratorIT.java
adutra: I just picked a number that is big enough to exercise spliterators in a large data stream.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/PagingIterableSpliterator.java
adutra: But what value? That is the question. If someone passes `Integer.MAX_VALUE` he's shooting himself in the foot.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/PagingIterableSpliterator.java
adutra: Yes, `ORDERED` is different from `SORTED`.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/PagingIterableSpliterator.java
adutra: It's impossible to split a non-`SIZED` stream in two halfs of same size. The next best thing to do is to split in chunks of equal size.

adutra:
> So, I don't understand how client code is supposed to use this.

"Normal" client code is not supposed to use spliterators directly, spliterators are just a support for creating streams. You should only be doing like below:

```
ResultSet result = session.execute(...);
long result = StreamSupport.stream(result.spliterator(), true).filter(...).map(...).count();
```

And your stream will be consumed in parallel using splits generated by the spliterator.

> The two splits are not balanced at all.

That is not the intent. Don't get fooled by the javadocs of `trySplit`: it's impossible to split a stream of unknown size in two chunks of equal size. The strategy here is to create splits of fixed size, which is why you are seeing a first split of size 128.

The algorithm used here creates an unbalanced tree of splits as follows:
```
   |
  / \
128  |
    / \
  128  |
      / \
    128 etc.
```

olim7t:
OK I think I get it, the threads who got the small chunks will steal new chunks from the original spliterator once they're done, correct?

adutra:
> OK I think I get it, the threads who got the small chunks will steal new chunks from the original spliterator once they're done, correct?

Correct with one condition: if the process is executed on a work-stealing pool such as the `ForkJoinPool` (which is the case if you call `StreamSupport.stream()`.

olim7t (approved):

PR #1274, opened 2019-06-22
Improve extensibility of plain text auth provider

olim7t:
Until we agree on something for [JAVA-2236](https://datastax-oss.atlassian.net/browse/JAVA-2236), this provides an easy first step for implementors who want to retrieve the credentials from another source.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/auth/PlainTextAuthProviderBase.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/auth/PlainTextAuthProviderBase.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.internal.core.auth;
+
+import com.datastax.oss.driver.api.core.auth.AuthProvider;
+import com.datastax.oss.driver.api.core.auth.Authenticator;
+import com.datastax.oss.driver.api.core.auth.SyncAuthenticator;
+import com.datastax.oss.driver.api.core.metadata.EndPoint;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/auth/PlainTextAuthProviderBase.java
```diff
@@ -0,0 +1,137 @@
...
+
+  @Override
+  public void close() {
+    // nothing to do
+  }
+
+  protected static class Credentials {
+
+    private final String username;
+    private final String password;
```
adutra: As an additional security measure, I would recommend storing the password as a char array, for the reasons explained [here](https://stackoverflow.com/questions/8881291/why-is-char-preferred-over-string-for-passwords). Converting from `char[]` to `byte[]` later on will be a bit annoying, but it's still doable.

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/auth/PlainTextAuthProviderBase.java
olim7t: Good point, I was not aware of that.

> Converting from char[] to byte[] later on will be a bit annoying

We might as well use `byte[]` directly then, wdyt?

adutra: I fear that users might be surprised by a password field of type `byte[]`, but I am OK going that direction if you think it's easier.

olim7t: All things considered, I'll stick to `char[]`. Using bytes forces the UTF-8 encoding on the users, that's kind of an implementation detail.

core/src/main/java/com/datastax/oss/driver/internal/core/auth/PlainTextAuthProviderBase.java
```diff
@@ -0,0 +1,166 @@
...
+
+    private ByteBuffer toUtf8Bytes(char[] charArray) {
+      CharBuffer charBuffer = CharBuffer.wrap(charArray);
+      return Charsets.UTF_8.encode(charBuffer);
+    }
+
+    @Override
+    @Nullable
+    public ByteBuffer initialResponseSync() {
+      return initialToken.duplicate();
```
olim7t, line 152: Continuing with the idea that sensitive information should be kept in memory as shortly as possible, this should probably be cleared once the protocol layer is done with it.

That's not really urgent (currently it will be GCed with the `Authenticator` / `ProtocolInitHandler`), I've logged [JAVA-2306](https://datastax-oss.atlassian.net/browse/JAVA-2306) to follow up on it.

PR #1277, opened 2019-06-24
JAVA-2303: Ignore peer rows matching the control host's RPC address

adutra:


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -162,8 +161,6 @@ public DefaultTopologyMonitor(InternalDriverContext context) {
     DriverChannel channel = controlConnection.channel();
     EndPoint localEndPoint = channel.getEndPoint();
 
-    savePort(channel);
```
adutra: I removed this altogether because I think it is racy.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -428,13 +427,46 @@ protected InetSocketAddress getBroadcastRpcAddress(@NonNull AdminRow row) {
...
       }
     }
-    return new InetSocketAddress(broadcastRpcInetAddress, broadcastRpcPort);
+    InetSocketAddress broadcastRpcAddress =
+        new InetSocketAddress(broadcastRpcInetAddress, broadcastRpcPort);
+    if (broadcastRpcAddress.equals(localEndPoint.resolve())) {
+      // JAVA-2303: if the peer is actually the control node, ignore that peer as it is likely
+      // a misconfiguration problem.
+      LOG.warn(
+          "[{}] Control node {} has an entry for itself in {}: this entry will be ignored. "
```
tolbertam, line 457: This will probably be sufficient for most cases.  If you've misconfigured your rpc_address with the most common case being using default localhost on all nodes, this will likely catch it.  Mainly, I think we just want the user to know something is configured wrong, the level of corrective action we take doesn't seem as important, but skipping the peer seems good since that'll match the behavior of 3.x

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
tolbertam: Aren't all operations on control connection always only done on the same thread, so there's no chance of two threads executing it concurrently?

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitorTest.java
```diff
@@ -239,23 +267,23 @@ public void should_refresh_node_from_peers_if_broadcast_address_is_not_present_V
   @Test
   public void should_get_new_node_from_peers() {
     // Given
-    AdminRow peer3 = mockPeersRow(3, UUID.randomUUID());
-    AdminRow peer2 = mockPeersRow(2, node2.getHostId());
-    AdminRow peer1 = mockPeersRow(1, node1.getHostId());
+    AdminRow peer3 = mockPeersRow(4, UUID.randomUUID());
```
tolbertam, line 268: Is there some significance behind this change? Just so i'm not missing anything.

tolbertam (approved):
looks good, although I suspect the logic change around the default port isn't needed, but :+1: either way.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
adutra: Sorry what I meant by racy is the fact that the control node can change while you are still updating the nodes after having queried the previous control node. It's the same caveat that is already expressed a few lines below in the javadocs of `nodeInfoBuilder`: "it would be racy to call {@code controlConnection.channel().getEndPoint()} from within this method, as the control connection may have changed its channel since."

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
adutra: What do you think about logging at WARN level? Driver 3.x logs at DEBUG level, but on one side, it seems serious enough to deserve a warning, but on the other side, I fear that the warning is going to be logged too often, wdyt?

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitorTest.java
adutra: Yes, the number you pass to the `mockPeersRow` method determines its broadcast RPC address, so for the test to pass, now you need to assign addresses distinct from the one used by the local node (which is 1 by default).

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
tolbertam: I think `WARN` is appropriate.  Something is wrong, and while things might still work...I don't picture a case where the configuration is what the user intended.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
olim7t: I don't follow either.
We retrieved the channel two lines above, and we reuse that local variable. It would be racy if we called `controlConnection.channel()` again.
Besides, this value is only used for Cassandra < 4, where the port is the same for all nodes.

adutra: The previous code would only save the port if its current value is -1. So at least in theory, we could save the port of channel A, then change the control channel, and we would keep using port from channel A to create `NodeInfo` instances retrieved from channel B. Granted, port numbers are meant to be the same across nodes with C* < 4, but I do believe that the new version is conceptually more accurate and future-proof than the previous one. And in fact, we really don't need to store this value as an instance field, it is easily obtainable from the current channel so that is another argument in favor of getting rid of this field altogether.

olim7t: Yes, that's the intended behavior: the ports are all the same, so we save it for the first channel, then reuse it.

> more accurate and future-proof

The only way the current code could fail is if nodes had different ports, but we couldn't find that information in system tables. In that situation, using the port of the control node for another node would always be wrong, whether it's the first ever control node or the current control node.

Please, let's focus this PR on JAVA-2303 and nothing else.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
olim7t: I don't see any logic to skip that check when the method is called for the local row. And indeed I see the warning even for 1-node integration tests (e.g. `ConnectKeyspaceIT`):
```
10:38:04.309 [s0-io-0] WARN  c.d.o.d.i.c.m.DefaultTopologyMonitor - [s0] Control node /127.0.0.1:9042 has an entry for itself in system.peers: this entry will be ignored. This is likely due to a misconfiguration; please verify your rpc_address configuration in cassandra.yaml on all nodes in your cluster.
```

olim7t (approved):
LGTM.

I'll note that using `localEndPoint.resolve()` is still not ideal, because it won't return what we want if there's some address translation at play. But that's the best we have since we cannot rely on `system.local.rpc_address` (because of CASSANDRA-11181).

One can reasonably assume that if users are dealing with address translators, they're past the point of making basic configuration mistakes in `cassandra.yaml`.

PR #1279, opened 2019-06-24

olim7t:
This was based on an older version of the mapper, it will require significant rebasing on top of the latest 4.x. I can help if needed.


tomekl007:
comments in closed branch: https://github.com/datastax/java-driver/pull/1245

PR #1281, opened 2019-06-26
JAVA-2236: Add method to set the auth provider in SessionBuilder

GregBestland:


core/src/main/java/com/datastax/oss/driver/internal/core/auth/RuntimePlainTextAuthProvider.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright DataStax, Inc.
```
GregBestland: This Could also be folded into the normal PlainTextAuthProvider if appropriate. That's what I did initially, but with the introduction of the PlainTextAuthProviderBase, it seemed appropriate to have a separate class.

core/revapi.json
```diff
@@ -4756,6 +4756,12 @@
...
+        "new": "method com.datastax.oss.driver.api.core.context.DriverContext com.datastax.oss.driver.api.core.session.SessionBuilder<SelfT extends com.datastax.oss.driver.api.core.session.SessionBuilder, SessionT>::buildContext(com.datastax.oss.driver.api.core.config.DriverConfigLoader, java.util.List<com.datastax.oss.driver.api.core.type.codec.TypeCodec<?>>, com.datastax.oss.driver.api.core.metadata.NodeStateListener, com.datastax.oss.driver.api.core.metadata.schema.SchemaChangeListener, com.datastax.oss.driver.api.core.tracker.RequestTracker, com.datastax.oss.driver.api.core.auth.AuthProvider, java.util.Map<java.lang.String, java.lang.String>, java.util.Map<java.lang.String, java.util.function.Predicate<com.datastax.oss.driver.api.core.metadata.Node>>, java.lang.ClassLoader)",
+        "justification": "Allow  for AuthProvider to be loaded programmatically"
```
olim7t: This method breaks every time we add a new programmatic builder option, and I think we're at a point where we should avoid that. I've opened [JAVA-2315](https://datastax-oss.atlassian.net/browse/JAVA-2315) to address this.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -223,6 +225,18 @@ public SelfT withRequestTracker(@Nullable RequestTracker requestTracker) {
...
   }
 
+  /**
+   * Register an auth provider
+   *
+   * <p>If the auth provider is specified programmatically with this method, it overrides the
+   * configuration (that is, the {@code advanced.auth-provider} option will be ignored).
+   */
+  @NonNull
+  public SelfT withAuth(@Nullable AuthProvider authProvider) {
```
olim7t: Nit: could you name this `withAuthProvider`?

olim7t: Maybe we could also add `withPlainTextCredentials(String username, String password)` that wraps the creation and registration of a `RuntimePlainTextAuthProvider`. That way the user doesn't have to deal with an internal class.

It would come with the following caveats:
* only works with OSS Cassandra's plain text authenticator
* does not support changing the credentials at runtime
* stores the credentials as strings in memory

olim7t (changes requested):
#1287 is merged, so you can rebase and you should now be able to avoid the breaking API change.

See also my comments inline. I'm not 100% sure about adding `withPlainTextCredentials`: the `AuthProvider` interface is flexible enough already, and we don't want to add too many specialized methods to the session builder, but this one sounds reasonable. If there are other opinions I'm open to hearing them :)

core/src/main/java/com/datastax/oss/driver/internal/core/auth/RuntimePlainTextAuthProvider.java
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * <pre>
+ *     SessionBuilder builder =
+ *         SessionUtils.baseBuilder()
+ *             .withAuth(new RuntimePlainTextAuthProvider("logPrefix", "username", "password"));
+ * </pre>
+ */
+@ThreadSafe
+public class RuntimePlainTextAuthProvider extends PlainTextAuthProviderBase {
+  private final String username;
```
adutra: Let's apply JAVA-2306 here and avoid storing sensitive data as strings. I suggest that you create a `Credentials` instance upfront and store it here instead of these two fields.

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/auth/RuntimePlainTextAuthProvider.java
GregBestland: :+1: that's a good idea.

GregBestland: So while this seemed like a good idea, it won't actually work. The credentials are cleared when used, and the auth provider get credentials is invoked a few times during session creation. As a result the credentials end up cleared as part of protocol negotiation. In order to accomplish this we would need a way to encrypt the user name password in memory, and recreate the credentials when needed.

adutra: Then you could implement a `Credentials.copy()` method as follows:

```
public Credentials copy() {
  return new Credentials(username.clone(), password.clone());
}
```
And use it as below:
```
public class RuntimePlainTextAuthProvider extends PlainTextAuthProviderBase {
  private final Credentials credentials;

  public RuntimePlainTextAuthProvider(String username, String password) {
    this("", username, password);
  }

  public RuntimePlainTextAuthProvider(String logPrefix, String username, String password) {
    super(logPrefix);
    credentials = new Credentials(username.toCharArray(), password.toCharArray());
  }

  @NonNull
  @Override
  protected Credentials getCredentials() {
    return credentials.copy();
  }
}
```

olim7t: But then we're still holding the credentials in memory, because the original field is never getting cleared.
I don't think there's a way to avoid it when the credentials have been provided programmatically (unless we encrypted them, but that's way off topic). I'll add a few explanations in the javadocs.

olim7t (approved):

PR #1284, opened 2019-06-27
JAVA-2312: Handle UDTs with names that clash with collection types

olim7t:


tolbertam (approved):
:+1:

adutra (commented):
Out of curiosity, what happens if the UDT is named `list` or `frozen`, will that work as expected?

olim7t:
> Out of curiosity, what happens if the UDT is named list or frozen, will that work as expected?

🤦‍♂ you're right, I don't how it didn't occur to me to check that. `list`, `map` and `frozen` are also valid UDT names, `set` isn't.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/DataTypeCqlNameParser.java
```diff
@@ -69,8 +69,23 @@ private DataType parse(
       return nativeType;
     }
 
+    if (parser.isEOS()) {
+      // No parameters => it's a UDT
+      CqlIdentifier name = CqlIdentifier.fromCql(type);
```
olim7t, line 74: It's simpler to move UDT before collections.
Note that this would accept a UDT called `set`, but it's not a big deal since that's impossible to create on the server anyway.

olim7t (commented):
Force-pushed.
This is now an entirely different fix, there was no value in keeping past history.

I'll update the 3.x PR too.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/DataTypeCqlNameParser.java
```diff
@@ -110,25 +122,17 @@ private DataType parse(
     }
 
     if (type.equalsIgnoreCase("tuple")) {
-      List<String> rawTypes = parser.parseTypeParameters();
+      if (parameters.isEmpty()) {
+        throw new IllegalArgumentException("Expecting at list one parameter for tuple, got none");
```
tolbertam, line 126: Shouldn't technically be possible since server won't allow it right?  Was able to exercise it in a unit test though so seems like a good idea :+1:.  I noticed `list<>,set<>,map<>, map<only1>` all throw similar errors too which is good.

tolbertam (approved):
:+1:

adutra (approved):

PR #1285, opened 2019-06-27
JAVA-2307: Improve @Select and @Delete by not requiring full primary key

tolbertam:
For [JAVA-2307](https://datastax-oss.atlassian.net/browse/JAVA-2307)

---

Motivation:

Currently `@Select` and `@Delete` annotated methods only allow
selecting/deleting individual entities when parameters are provided as
the mapper currently requires specifying all columns of the primary key as
parameters.

Since CQL allows you to select and delete multiple rows by providing a
subset of the primary key, these dao methods can be improved by
accepting a subset of the primary key as parameters, as long as
at least the partition keys are represented.

In addition `@Select` should be supported when providing no parameters,
which will allow querying an entire table's rows for a given entity.

Modifications:

Add `EntityHelper.selectByPrimaryKeyParts` and `deleteByPrimaryKeyParts`
methods and added generators to provide this capability.

Refactor parameter validation into EntityUtils.areParametersValid from
somewhat duplicated logic in delete and select method generators.

Add tests to exercise functionality.

Add documentation demonstrating extended capability.

Result:

`@Select` and `@Delete` now support providing a subset of primary key to
select/delete multiple entities.

`@Select` also supports providing zero parameters, which performs a
select on all rows in an entity's table.


mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/entity/EntityHelper.java
```diff
@@ -188,6 +188,23 @@
...
+   * values (the bind markers have the same names as the columns). They are listed in the natural
+   * order, i.e. partition key columns first, followed by clustering columns (in the order defined
+   * by the {@link PartitionKey} and {@link ClusteringColumn} annotations on the entity class).
+   *
+   * <p>If <code>parameterCount</code> of 0 is provided, this is the equivalent of a select query
+   * without a where clause, which queries for all rows in a table.
+   */
+  @NonNull
+  default Select selectByPrimaryKeyParts(int parameterCount) {
+    throw new UnsupportedOperationException("selectbyPrimaryKeyParts is unimplemented");
```
tolbertam: Done to avoid breaking changes, although it is a bit awkward.

changelog/README.md
```diff
@@ -2,8 +2,10 @@
 
 <!-- Note: contrary to 3.x, insert new entries *first* in their section -->
 
+
```
adutra: Unnecessary change.

changelog/README.md
```diff
@@ -2,8 +2,10 @@
...
 ### 4.2.0 (in progress)
 
+- [improvement] JAVA-2307: Improve @Select and @Delete by not requiring full primary key
```
adutra: Tip: place all words starting with `@` within backticks here and in commit messages, to avoid having Github consider that these are user handles.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/InventoryITBase.java
```diff
@@ -31,14 +33,31 @@
       new Product(UUID.randomUUID(), "Flamethrower", new InventoryITBase.Dimensions(30, 10, 8));
   protected static Product MP3_DOWNLOAD = new Product(UUID.randomUUID(), "MP3 download", null);
 
+  protected static ProductSale FLAMETHROWER_SALE_1 =
+      new ProductSale(FLAMETHROWER.getId(), "2019-06-27", 1, Uuids.startOf(1561643130), 500.00, 5);
```
olim7t: Maybe you could have the date in the constant name, to make things more obvious in tests.

manual/mapper/daos/delete/README.md
```diff
@@ -36,6 +36,19 @@ The method can operate on:
     In addition, because the entity class can't be inferred from the method signature, it must be
     specified via the annotation's `entityClass` element.
 
+* a subset of the primary key (partition key, or partition key + subset of clustering columns):
```
olim7t: It's always the first k clustering columns, correct? Like if the PK is `((pk1, pk2), cc1, cc2, cc3)` you can't do `((pk1, pk2), cc3)`. We should say it explicitly in the docs (I'm not sure how best to formulate it).

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
```diff
@@ -105,8 +105,10 @@ public DaoDeleteMethodGenerator(
     VariableElement firstParameter = parameters.get(0);
     entityElement = EntityUtils.asEntityElement(firstParameter, typeParameters);
     hasEntityParameter = (entityElement != null);
+    final int keyParameterSize;
```
olim7t: I think we can find a better name for this. It's the number of PK columns that we want in the generated query.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
```diff
@@ -120,25 +122,39 @@ public DaoDeleteMethodGenerator(
...
-                "Invalid parameter list: %s methods that do not operate on an entity instance "
-                    + "must take the partition key components in the exact order "
-                    + "(expected PK of %s: %s)",
-                Delete.class.getSimpleName(),
-                entityElement.getSimpleName(),
-                primaryKeyTypes);
+      // if a custom if clause is provided, the whole primary key must also be provided.
+      List<? extends VariableElement> keyParameters = parameters;
+      if (!annotation.customIfClause().isEmpty()) {
+        if (keyParameters.size() < entityDefinition.getPrimaryKey().size()) {
```
olim7t: This gave me pause because the check is not accurate, we don't know how many placeholders the custom IF clause has.
But it's more of a guard so that we can perform the `subList` call in the other branch; if this check succeeds it's indeed a definitive mistake (even though there could still be something wrong if the check fails, but that will be caught later).
So 👍

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/EntityUtils.java
```diff
@@ -81,4 +88,86 @@ public static TypeElement asEntityElement(
     }
     return typeElement;
   }
+
+  /**
+   * Validates that the given parameters are valid for an {@link EntityDefinition}, meaning that
+   * there are at least enough parameters provided to match the number of partition key columns and
+   * that that parameter types match the primary key types.
```
olim7t:
```suggestion
   * that parameter types match the primary key types.
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
```diff
@@ -187,7 +204,7 @@ public DaoDeleteMethodGenerator(
           entityDefinition.getPrimaryKey().stream()
               .map(PropertyDefinition::getCqlName)
               .collect(Collectors.toList());
-      List<? extends VariableElement> bindMarkers = parameters.subList(0, primaryKeyNames.size());
+      List<? extends VariableElement> bindMarkers = parameters.subList(0, keyParameterSize);
```
olim7t: I think this assertion in `GeneratedCodePatterns.bindParameters` is going to fail with this change:
```java
assert bindMarkerNames.size() == parameters.size();
```
The cleanest fix is to sublist `primaryKeyNames` as well.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/entity/EntityHelper.java
olim7t: > Done to avoid breaking changes, although it is a bit awkward.

Yes, and this only hides the breaking change by turning it into a runtime error. I've been thinking about other possible approaches:

1. introduce a sub-interface and require a cast (as we've done in other cases). The problem is that this doesn't scale well: we add `ExtendedProductHelper` now but what if we need more methods in the next version? `ExtendedExtendedProductHelper`?
2. Accept the breaking change and document it, on the assumption that clients probably don't implement this interface. The problem is in the "probably"; also, it's a slippery slope.
3. Don't add those methods to `EntityHelper`. Instead, generate the query directly from the method generators. For SELECT we already have `selectStart()`; for DELETE we have to duplicate the whole method, but it's not huge either.

    We won't make those queries available to users through the interface (e.g. for custom query providers), but they're not that hard to rewrite from scratch anyway.

My vote goes to 3.

olim7t (changes requested):

changelog/README.md
tolbertam: Thanks, I'll amend the commit when merging. It looks like github doesn't hijack markdown in the same way, but agree that we shoulds use backticks here as well.

manual/mapper/daos/delete/README.md
tolbertam: That's right.  Although if `cc1`, `cc2` and `cc3` have the same types technically the code would still generate, but it's not what user wants.  I'll document this.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/entity/EntityHelper.java
tolbertam: > Don't add those methods to EntityHelper

You know, it didn't occur to me that generated code used the implementations instead of declaring things as `EntityHelper`.  In this case, there is no reason to add something to the interface, so I agree this seems like a good course of action.

It looks like the only place (currently) that `EntityHelper` is used as an interface in other code is in `QueryProvider` and in `DaoBase`, where it's more used for setting/getting fields.

Since this method is used for query generation, I suppose it's probably fine to not put directly in `EntityHelper`.

But I do wonder if we should just consider removing all the query-generation based methods from `EntityHelper` now if we see no utility.  What do you think?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
tolbertam: good catch :+1:

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
tolbertam: That's right.  It's not legal to not provide the full primary key and use a custom if clause.

So there's possibly the case where enough parameters are provided to fulfill the number of primary key parameters, but some of those parameters were meant for the custom if clause.  In this case it's likely it won't match the primary key definition and you'll get an error which explains that he parameters don't match the primary key, which is technically correct.  If they do somehow match, there's not much we can do at compile time (I think?).  In any case, you'll get an error at run time, which is similar to what would happen if you don't provide enough parameters to match the query bind markers.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/entity/EntityHelper.java
olim7t: > You know, it didn't occur to me that generated code used the implementations instead of declaring things as `EntityHelper`.

It didn't occur to me either 😆 I was initially thinking of moving the code to the DAO implementation class, but keeping them as "internal" methods in the helper class is even better.

> removing all the query-generation based methods from `EntityHelper`

As we discussed privately, those methods can be useful for `@QueryProvider` implementors, so it's nice to have them on the interface.

manual/mapper/daos/delete/README.md
```diff
@@ -36,6 +36,28 @@ The method can operate on:
...
+    // delete all rows in partition
+    @Delete(entityClass = ProductSale.class)
+    void deleteByIdForDay(UUID productId, LocalDate day);
+
+    // delete by partition key and partial clustering key
+    @Delete(entityClass = ProductSale.class)
+    void deleteByIdForCustomer(UUID productId, LocalDate day, UUID customerId);
+   
+    /* Note that the clustering columns in your primary key definition are significant. All
+     * proceeding clustering columns must be provided if any are.
```
olim7t: > proceeding clustering columns

Did you mean "preceding"?

olim7t (approved):
LGTM

manual/mapper/daos/delete/README.md
tolbertam: :man_facepalming: yep, that's what I meant.  Good catch, i'll fix.

PR #1287, opened 2019-06-28
JAVA-2315: Improve extensibility of session builder

olim7t:
Motivation:

Every time we add a new programmatic argument to the session builder,
there is a breaking change on SessionBuilder#buildContext.

This extends to DefaultDriverContext's constructor (even though
DefaultDriverContext is an internal class, it would be wise to be
conservative about its API, because plugging something through a custom
context subclass is one of the most basic "advanced" extensions).

Modifications:

Create a new class ProgrammaticArguments to wrap the arguments to
buildContext. Also pass it to the context constructor.

Preserve backward compatibility with the previous signatures.

Result:

We can now add new arguments to ProgrammaticArguments, which is not a
breaking change.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -4,6 +4,7 @@
 
 ### 4.2.0 (in progress)
 
+- [improvement] JAVA-2315: Improve extensibility of session builder
 
 ### 4.1.0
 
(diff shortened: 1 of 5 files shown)


adutra (approved):
Nice 👍

PR #1288, opened 2019-07-01
JAVA-2204: Avoid memory leak when client holds onto a stale TableMetadata instance

olim7t:


driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
```diff
@@ -441,19 +441,29 @@ private void updateKeyspaces(
         metadata.triggerOnKeyspaceChanged(newKeyspace, oldKeyspace);
       }
       Map<String, TableMetadata> oldTables =
-          oldKeyspace == null ? new HashMap<String, TableMetadata>() : oldKeyspace.tables;
+          oldKeyspace == null
+              ? new HashMap<String, TableMetadata>()
+              : new HashMap<String, TableMetadata>(oldKeyspace.tables);
```
olim7t, line 446: Not the most elegant fix, but it has the merit of being straightforward.

I initially thought of replacing this with a simple `get` call:
```java
TableMetadata oldTable = oldTables.put(newTable.getName(), newTable);
```
But that doesn't work because during a partial refresh, `oldTables` points to the tables of the current keyspace.

driver-core/src/test/java/com/datastax/driver/core/SchemaChangesTest.java
```diff
@@ -80,13 +81,18 @@
...
-    cluster1 = builder.build();
-    cluster2 = builder.build();
+            .withQueryOptions(nonDebouncingQueryOptions())
+            .build();
+    cluster2 =
+        Cluster.builder()
+            .addContactPoints(getContactPoints())
+            .withPort(ccm().getBinaryPort())
+            .withQueryOptions(nonDebouncingQueryOptions())
+            .build();
```
olim7t, line 95: The previous version shared the same `QueryOptions` instance between the two clusters, which caused my test to perform the full refresh on the wrong cluster. Took me some time to figure this out 😞

driver-core/src/main/java/com/datastax/driver/core/SchemaParser.java
```diff
@@ -441,19 +441,29 @@ private void updateKeyspaces(
...
+          oldKeyspace == null
+              ? new HashMap<String, AggregateMetadata>()
+              : new HashMap<String, AggregateMetadata>(oldKeyspace.aggregates);
       updateAggregates(metadata, oldAggregates, newKeyspace.aggregates, null);
       Map<String, MaterializedViewMetadata> oldViews =
-          oldKeyspace == null ? new HashMap<String, MaterializedViewMetadata>() : oldKeyspace.views;
+          oldKeyspace == null
+              ? new HashMap<String, MaterializedViewMetadata>()
+              : new HashMap<String, MaterializedViewMetadata>(oldKeyspace.views);
       updateViews(metadata, oldViews, newKeyspace.views, null);
```
tomekl007, line 467: shouldn't `newKeyspace.views` also be wrapped in the `new HashMap<String, MaterializedViewMetadata>`?

olim7t: This is not necessary, `updateViews` does not mutate `newKeyspace.views`.

driver-core/src/test/java/com/datastax/driver/core/SchemaChangesTest.java
tolbertam: Oh wow, I can imagine that taking a bit to figure out :laughing:.  I guess that's a huge downside of having operations driven by changing configuration objects directly :+1:

tolbertam (approved):

adutra (approved):

PR #1289, opened 2019-07-03
JAVA-2306: Clear security tokens from memory immediately after use

olim7t:
Most of the changes are in the native protocol (datastax/native-protocol#28), this merely upgrades the dependency and adds a few javadocs.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/auth/Authenticator.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/auth/Authenticator.java
@@ -57,7 +57,11 @@ public interface Authenticator {
    * Obtain an initial response token for initializing the SASL handshake.
    *
    * @return a completion stage that will complete with the initial response to send to the server
-   *     (which may be {@code null}).
+   *     (which may be {@code null}). Note that, if the returned byte buffer is writable, the driver
+   *     will <b>clear its contents</b> immediately after use (to avoid keeping sensitive
+   *     information in memory); do not reuse the same buffer across multiple invocations.
+   *     Alternatively, if the contents are not sensitive, you can make the buffer {@linkplain
+   *     ByteBuffer#asReadOnlyBuffer() read-only} and safely reuse it.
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/auth/Authenticator.java
```diff
@@ -68,7 +70,10 @@
...
    * @param challenge the server's SASL challenge.
    * @return a completion stage that will complete with the updated SASL token (which may be null to
-   *     indicate the client requires no further action).
+   *     indicate the client requires no further action). Note that the driver will <b>clear the
+   *     contents</b> of the returned byte buffer immediately after use (to avoid keeping sensitive
+   *     information in memory); if your authenticator returns the same result for multiple
+   *     evaluations, do not reuse the same buffer.
    */
   @NonNull
   CompletionStage<ByteBuffer> evaluateChallenge(@Nullable ByteBuffer challenge);
```
adutra, line 82: As I commented in the other PR: do we need to clear the challenge as well?

core/src/main/java/com/datastax/oss/driver/api/core/auth/Authenticator.java
```diff
@@ -68,7 +70,10 @@
    *
    * @param challenge the server's SASL challenge.
    * @return a completion stage that will complete with the updated SASL token (which may be null to
-   *     indicate the client requires no further action).
+   *     indicate the client requires no further action). Note that the driver will <b>clear the
+   *     contents</b> of the returned byte buffer immediately after use (to avoid keeping sensitive
+   *     information in memory); if your authenticator returns the same result for multiple
+   *     evaluations, do not reuse the same buffer.
```
adutra: You could add: `...do not reuse the same buffer, or make it {@linkplain ByteBuffer#asReadOnlyBuffer() read-only}.`

adutra (approved):

PR #1290, opened 2019-07-03
JAVA-2308: Add customWhereClause to Delete

tolbertam:
For [JAVA-2308](https://datastax-oss.atlassian.net/browse/JAVA-2308)

---

Motivation:

CQL DELETE allows advanced capabilities for deleting a range of data
within a partition.  For example given a primary key of ((int k), int c),
one could perform the following:

DELETE FROM tbl WHERE k = 0 and c > 1 and c < 4

To support this capability in the mapper, customWhereClause should be
added to Delete in a similar way it is handled in the Select annotation.

Modifications:

Add customWhereClause to Delete annotation.

Update DaoDeleteMethodGenerator to handle customWhereClause.

Add EntityHelperDeleteStartMethodGenerator which adds a deleteStart
method, which is reused in
EntityHelperDeleteByPrimaryKeyPartsMethodGenerator and also used in the
code generation in DaoDeleteMethodGenerator

Result:

customWhereClause added to Delete.

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/mapper/DeleteIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/mapper/DeleteIT.java
@@ -24,6 +24,7 @@ import com.datastax.oss.driver.api.core.cql.AsyncResultSet;
 import com.datastax.oss.driver.api.core.cql.ResultSet;
 import com.datastax.oss.driver.api.core.cql.Row;
 import com.datastax.oss.driver.api.core.cql.SimpleStatement;
+import com.datastax.oss.driver.api.core.uuid.Uuids;
 import com.datastax.oss.driver.api.mapper.annotations.CqlName;
 import com.datastax.oss.driver.api.mapper.annotations.Dao;
 import com.datastax.oss.driver.api.mapper.annotations.DaoFactory;
@@ -84,6 +85,7 @@ public class DeleteIT extends InventoryITBase {
     saleDao.save(FLAMETHROWER_SALE_2);
     saleDao.save(FLAMETHROWER_SALE_3);
     saleDao.save(FLAMETHROWER_SALE_4);
+    saleDao.save(FLAMETHROWER_SALE_5);
     saleDao.save(MP3_DOWNLOAD_SALE_1);
   }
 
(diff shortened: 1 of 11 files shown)


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
```diff
@@ -120,43 +135,51 @@ public DaoDeleteMethodGenerator(
...
+                        + "must specify the entire primary key (expected primary keys of %s: %s)",
+                    Delete.class.getSimpleName(),
+                    entityElement.getSimpleName(),
+                    primaryKeyTypes);
+            return Optional.empty();
+          } else {
+            // restrict parameters to primary key length.
+            primaryKeyParameters =
+                primaryKeyParameters.subList(0, entityDefinition.getPrimaryKey().size());
+          }
```
tolbertam, line 167: No advanced validation is done on parameters if `customWhereClause` is provided.  We could technically look at the # of parameters and compare to partition key size (since that's required), but not the types since those could be in any order.  I figured adding size validation would be validation for the sake for validation.

tolbertam:
Another thing I realized while doing this is there is currently no mechanism for deleting individual column(s), only full rows.  Do we want that capability?  That could be another ticket.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/DeleteIT.java
```diff
@@ -236,6 +242,39 @@ public void should_delete_if_exists_sales() {
...
+    assertThat(result.wasApplied()).isFalse();
+    Row row = result.one();
+    assertThat(row).isNotNull();
+    assertThat(row.getDouble("price")).isEqualTo(500.0);
+
+    result =
+        saleDao.deleteCustomWhereCustomIf(
+            2, FLAMETHROWER.getId(), DATE_1, FLAMETHROWER_SALE_2.getTs(), 500.0);
+
+    assertThat(result.wasApplied()).isTrue();
```
tomekl007, line 275: shouldn't we check result after this call: `saleDao.all().all()` as well? Not only `wasApplied()`

olim7t:
> Another thing I realized while doing this is there is currently no mechanism for deleting individual column(s), only full rows.

I think `@Query` is the way to go for that case.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
olim7t: Agreed, no need to go into details if we can't fully validate anyway.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
```diff
@@ -215,13 +238,13 @@ public DaoDeleteMethodGenerator(
 
     // Bind any remaining parameters, assuming they are values for a custom IF clause
```
olim7t: Nit: could say "WHERE or IF clause"

olim7t (approved):
👍

integration-tests/src/test/java/com/datastax/oss/driver/mapper/DeleteIT.java
tolbertam: The other LWT tests don't do this, just knowing that it is applied is probably enough, anything else is more validating C* behavior.

PR #1291, opened 2019-07-05
JAVA-2324 & JAVA-2325: Fixes for primitive types in the mapper

adutra:


integration-tests/src/test/java/com/datastax/oss/driver/mapper/PrimitivesIT.java
```diff
@@ -0,0 +1,218 @@
...
+
+  private static TestMapper mapper;
+
+  @BeforeClass
+  public static void setup() {
+    CqlSession session = sessionRule.session();
+    session.execute(
+        SimpleStatement.builder(
+                "CREATE TABLE primitives_entity("
+                    + "id int PRIMARY KEY, "
```
tomekl007, line 64: maybe you can add a similar test for wrapper types (`Integer`, `Load`, ...)?

adutra: I don't think it's necessary, they are just normal objects so it's guaranteed that they will be treated exactly like any other type (`String`, `BigDecimal`...)

tomekl007 (approved):

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -102,21 +102,36 @@ public EntityDefinition getDefinition(TypeElement classElement) {
...
         TypeMirror typeMirror = getMethod.getReturnType();
         if (typeMirror.getKind() == TypeKind.VOID) {
           continue;
         }
-        String propertyName = Introspector.decapitalize(getMethodName.substring(3));
+
+        String getMethodName = getMethod.getSimpleName().toString();
+        boolean regularGetterName = getMethodName.startsWith("get");
+        boolean booleanGetterName =
+            getMethodName.startsWith("is") && typeMirror.getKind() == TypeKind.BOOLEAN;
```
olim7t: We should also allow `Boolean.class`. I know the JavaBeans spec only mentions `boolean`, but we're not strictly adhering to it.

olim7t (approved):
LGTM.

Usual request to avoid "wishful" commit messages:
```
JAVA-2324: Add support for primitive shorts in mapper
JAVA-2325: Allow "is" prefix for boolean getters in mapped entities
```

adutra:
> Usual request to avoid "wishful" commit messages:

OK, but I wanted to point out that imperative style imo does not work very well for bugs: `Add support for primitive shorts in mapper` makes it sound less like a bug and more like a new feature.

PR #1292, opened 2019-07-05
JAVA-2323: Driver should fire onAdd event for node with same host id but different IP address

adutra:


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/AddNodeRefresh.java
```diff
@@ -38,18 +39,18 @@ public Result compute(
...
-      Map<UUID, Node> newNodes =
-          ImmutableMap.<UUID, Node>builder()
-              .putAll(oldNodes)
-              .put(newNode.getHostId(), newNode)
-              .build();
-      return new Result(
-          oldMetadata.withNodes(newNodes, tokenMapEnabled, false, null, context),
-          ImmutableList.of(NodeStateEvent.added(newNode)));
+      Node oldNode = oldNodes.get(newNodeInfo.getHostId());
+      if (oldNode.getEndPoint().equals(newNodeInfo.getEndPoint())) {
```
adutra: This is the simplest fix I can think of, but I would like to understand: what was the initial motivation to _not_ update the metadata if the host ID is already present in it? Is it because sometimes we receive onAdd events that are false positives?

olim7t: > what was the initial motivation to not update the metadata if the host ID is already present in it

We've been asked to add a node, the node is already present => do nothing.

But the underlying assumption was that the server also treats `host_id` as the unique node identifier. Based on the reporter's scenario, I now suspect that that is wrong. Otherwise we would get `UP`, not `NEW_NODE`.

I think the right thing to do is to `copyInfos` the new data into the existing `Node` instance, and probably emit `TopologyEvent.suggestUp` instead of `NodeStateEvent.added`.

olim7t: As discussed offline, the above is still not sufficient because the node's endpoint won't be updated.
So we should make `DefaultNode.endPoint` volatile and have `copyInfos` update it.

olim7t: OK, I've dug deeper into this. I didn't set up a full Docker environment, but actually it's easy to reproduce with just a local CCM cluster by stopping a node and changing the IP manually.

The main problem if we create a new node is that the old pool isn't removed, and keeps trying to reconnect. So as agreed I went the other route and mutated the existing instance, which led me further down the rabbit hole:
* `metricUpdater` must also be switched because node metrics are prefixed by the IP.
* we can't use `endPoint` anymore in `equals`/`hashCode` (mutable field is a big no-no). Since `hostId` doesn't work either (it's not known until the first connection is established), I've switched to reference equality, which after all is quite logical since the driver always returns the same instance for a known host.

Since I had the changes locally, I figured I might as well push them.

There are two more things I saw along the way but I'll open follow-up tickets:
* when we switch the metrics, the old ones for the previous IP stay, with their last value. That's also the case for a regular `NODE_REMOVED` event. We might want to unregister them ([JAVA-2331](https://datastax-oss.atlassian.net/browse/JAVA-2331)).
* when we process a regular `NODE_REMOVED` event, we remove the node from the metadata, but I suspect we don't shut down the pool ([JAVA-2332](https://datastax-oss.atlassian.net/browse/JAVA-2332)).

adutra: > The main problem if we create a new node is that the old pool isn't removed, and keeps trying to reconnect.

Why can't you fire 2 events in `AddNodesRefresh` and force the old node down? E.g.

```
if (!existing.getEndPoint().equals(newNodeInfo.getEndPoint())) {
  copyInfos(newNodeInfo, ((DefaultNode) existing), null, context);
  assert existing.getBroadcastRpcAddress().isPresent(); // always for peer nodes
  assert newNodeInfo.getBroadcastRpcAddress().isPresent(); // always for peer nodes
  return new Result(
      oldMetadata,
      ImmutableList.of(
          TopologyEvent.forceDown(existing.getBroadcastRpcAddress().get()),
          TopologyEvent.suggestUp(newNodeInfo.getBroadcastRpcAddress().get())));
} else {
  return new Result(oldMetadata);
}
```

Wouldn't that save you from mutating the existing instance?

olim7t: Yeah I guess that could work too.
But then you get a different `Node` instance. I actually like unicity, it's nice to know that you can hold onto an instance and see the changes reflected over time.

olim7t (approved):

PR #1293, opened 2019-07-08
JAVA-2326: Reduce memory allocations during queries execution

netudima:
Fields "id" are evaluated for each request and allocate extra objects in heap but actually needed only for tracing. The values can be calculated lazy.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/Connection.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Connection.java
@@ -731,7 +731,8 @@ class Connection {
       throw new ConnectionException(address, "Connection has been closed");
     }
 
-    logger.trace("{}, stream {}, writing request {}", this, request.getStreamId(), request);
+    if (logger.isTraceEnabled())
+      logger.trace("{}, stream {}, writing request {}", this, request.getStreamId(), request);
     writer.incrementAndGet();
 
     if (DISABLE_COALESCING) {
@@ -787,8 +788,9 @@ class Connection {
                   }
                 });
         } else {
-          logger.trace(
-              "{}, stream {}, request sent successfully", Connection.this, request.getStreamId());
+          if (logger.isTraceEnabled())
+            logger.trace(
+                "{}, stream {}, request sent successfully", Connection.this, request.getStreamId());
         }
       }
     };
@@ -1103,11 +1105,11 @@ class Connection {
         }
       }
 
-      // Always flush what we have (don't artificially delay to try to coalesce more messages)
-      for (Channel channel : channels) channel.flush();
-      channels.clear();
-
       if (doneWork) {
+        // Always flush what we have (don't artificially delay to try to coalesce more messages)
+        for (Channel channel : channels) channel.flush();
+        channels.clear();
+
         runsWithNoWork = 0;
       } else {
         // either reschedule or cancel
--- a/driver-core/src/main/java/com/datastax/driver/core/Frame.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/Frame.java
@@ -179,9 +179,10 @@ class Frame {
       WARNING,
       USE_BETA;
 
+      private static final Flag[] values = Flag.values();
+
       static EnumSet<Flag> deserialize(int flags) {
         EnumSet<Flag> set = EnumSet.noneOf(Flag.class);
-        Flag[] values = Flag.values();
         for (int n = 0; n < 8; n++) {
           if ((flags & (1 << n)) != 0) set.add(values[n]);
         }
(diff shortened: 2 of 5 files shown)


driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -731,7 +731,8 @@ ResponseHandler write(
       throw new ConnectionException(address, "Connection has been closed");
     }
 
-    logger.trace("{}, stream {}, writing request {}", this, request.getStreamId(), request);
+    if (logger.isTraceEnabled())
+      logger.trace("{}, stream {}, writing request {}", this, request.getStreamId(), request);
```
adutra, line 735: Just to make sure I understand, you are trying to avoid the array allocation for the varags arguments?

driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -1103,11 +1105,11 @@ public void run() {
         }
       }
 
-      // Always flush what we have (don't artificially delay to try to coalesce more messages)
-      for (Channel channel : channels) channel.flush();
-      channels.clear();
-
       if (doneWork) {
+        // Always flush what we have (don't artificially delay to try to coalesce more messages)
+        for (Channel channel : channels) channel.flush();
```
adutra, line 1110: Makes sense, although I'm not very comfortable changing this logic.

driver-core/src/main/java/com/datastax/driver/core/Frame.java
```diff
@@ -179,9 +179,10 @@ static int lengthFor(ProtocolVersion version) {
       WARNING,
       USE_BETA;
 
+      private static final Flag[] values = Flag.values();
+
       static EnumSet<Flag> deserialize(int flags) {
         EnumSet<Flag> set = EnumSet.noneOf(Flag.class);
-        Flag[] values = Flag.values();
```
adutra, line 184: Are you sure this gives us any perf improvement? I would expect the implementation of `Enum.values()` to be highly optimized.

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -83,15 +82,13 @@
   private volatile List<Host> triedHosts;
   private volatile ConcurrentMap<InetSocketAddress, Throwable> errors;
 
-  private final Timer.Context timerContext;
```
adutra, line 86: So you are just trying to avoid the allocation of a `Timer.Context` object, right?

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -129,6 +125,14 @@ void cancel() {
     cancelPendingExecutions(null);
   }
 
+  String getId() {
+    // atomicity is not required to set the value
```
adutra, line 129: This method could be called concurrently from different speculative executions. Are you saying atomicity is not required because even if 2 threads enter the method at the same time, the effect is going to be deterministic and idempotent, i.e. field `id` will be eventually set to `Long.toString(System.identityHashCode(this))`?

adutra (commented):
Thanks for you PR. Everything makes sense although I'm not sure any of these changes will produce a significant performance improvement. Have you benchmarked your changes? I'm also a bit reluctant in introducing these changes now that the 3.x branch is being retired.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
netudima: array allocation in such places is usually optimized by escape analysis, the main issue which I see for such trace logging - hidden auto-boxing of int streamId.

The main goal of all the changes is to reduce an overall memory allocation on an application side (and to reduce GC pauses for the application correspondently). I am not expecting a visible change in request processing avg latency/throughput. Maybe some small CPU reduction will be as a bonus due to reducing some small amount of work.
Added a comment to https://datastax-oss.atlassian.net/browse/JAVA-2326

driver-core/src/main/java/com/datastax/driver/core/Frame.java
netudima: enum.values() method creates a new copy of array with values each time (due to inability to create immutable arrays in Java - https://dzone.com/articles/memory-hogging-enumvalues-method)

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
netudima: yes, trying to avoid the extra object allocation

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
netudima: yes, you are right, the result is deterministic and double evaluation is not a problem if it will happen (an atomic construction, like AtomicLong, would be more expensive here)

driver-core/src/main/java/com/datastax/driver/core/Connection.java
netudima: the main trigger for the change is to avoid HashSet iterator objects creation for empty spin loops of flushing

olim7t:
Closing per my comment on the JIRA ticket.

PR #1294, opened 2019-07-09
JAVA-2320: Expose more attributes on `@Select` for individual query clauses

olim7t:


integration-tests/src/test/java/com/datastax/oss/driver/mapper/SelectOtherClausesIT.java
```diff
@@ -0,0 +1,289 @@
...
+  @Dao
+  public interface SimpleDao {
+    @Insert
+    void insert(Simple simple);
+
+    @Select(limit = ":l")
+    PagingIterable<Simple> selectWithLimit(@CqlName("l") int l);
+
+    @Select(limit = ":l")
+    PagingIterable<Simple> selectWithLimit(int k, @CqlName("l") int l);
```
olim7t, line 152: These are examples of the new rule "if the method doesn't take the full primary key, the other parameters must be annotated with `@CqlName`". This only applies to `@Select`.

manual/mapper/daos/select/README.md
```diff
@@ -120,7 +134,13 @@ entity class and the [naming strategy](../../entities/#naming-strategy)).
 [default keyspace]:          https://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/core/session/SessionBuilder.html#withKeyspace-com.datastax.oss.driver.api.core.CqlIdentifier-
 [@ClusteringColumn]:         https://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/mapper/annotations/ClusteringColumn.html
 [@PartitionKey]:             https://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/mapper/annotations/PartitionKey.html
-[@Select]:                   https://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/mapper/annotations/Select.html
+[@Select]:                   https://docs.datastax.com/en/drivers/java/4.1/com/datastax/oss/driver/api/mapper/annotations/Select.html
+[allowFiltering()]:          https://docs.datastax.com/en/drivers/java/4.0/com/datastax/oss/driver/api/mapper/annotations/Select.html#allowFiltering--
```
olim7t: I'm aware that all the links should say 4.1. I created a separate ticket [JAVA-2327](https://datastax-oss.atlassian.net/browse/JAVA-2327) to fix that.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/SelectOtherClausesIT.java
tolbertam: Nice, when I first heard about this I had thought it would always apply, but if it's only when you don't provide the whole key, that seems very reasonable :+1:

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
```diff
@@ -79,70 +79,53 @@ protected DaoReturnType parseAndValidateReturnType(
...
-                  methodElement,
-                  "Invalid ttl value: "
-                      + "'%s' is not a bind marker name and can't be parsed as a literal integer "
-                      + "either, the generated query will probably fail",
-                  ttl);
-        }
-        methodBuilder.addCode(".usingTtl($L)", ttl);
-      }
-    }
+    maybeAddSimpleClause(ttl, Integer::parseInt, "usingTtl", "ttl", methodBuilder);
```
tolbertam, line 82: Nice refactoring :)

tolbertam (approved):
Looks great! :+1:

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoSelectMethodGenerator.java
```diff
@@ -87,32 +93,61 @@ public DaoSelectMethodGenerator(
...
     String customClause = selectAnnotation.customWhereClause();
-    // select without where criteria is ok.
-    // if parameters are provided, we must have at least enough to match partition key.
-    if (customClause.isEmpty()
-        && !parameters.isEmpty()
+    if (parameters.isEmpty()) {
+      primaryKeyParameters = freeFormParameters = Collections.emptyList();
+    } else if (customClause.isEmpty()) {
+      // If we have a partial primary key *and* free-form parameters, things get ambiguous because
+      // we don't know where the primary key ends. By convention, we require all free-form
```
tomekl007: >By convention, we require all free-form parameters to be annotated with @CqlName in those cases.

Is it true? from the logic below, it seems that we are treating the first occurrence of the value annotated with `@CqlName` parameter as the limit, and we are not validating that subsequent parameters are annotated with `@CqlName`.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoSelectMethodGenerator.java
```diff
@@ -157,22 +193,66 @@ private void generateSelectRequest(
...
+    if (annotation.allowFiltering()) {
+      methodBuilder.addCode(".allowFiltering()");
+    }
+    methodBuilder.addCode(".build();$]\n");
   }
+
+  private void addOrdering(String orderingSpec, MethodSpec.Builder methodBuilder) {
+    List<String> tokens = ON_SPACES.splitToList(orderingSpec);
+    ClusteringOrder clusteringOrder;
+    if (tokens.size() != 2 || (clusteringOrder = parseClusteringOrder(tokens.get(1))) == null) {
```
tomekl007, line 236: are we supporting the `orderBy` to have multiple statements? I.e:
```cc DESC aa DESC```
If so why we are validating only for tokens.size() != 2? Should we allow tokens to have `size() % 2 == 0`?

File: manual/mapper/daos/select/README.md [hunk 6]
```diff
@@ -51,7 +51,7 @@ public interface ProductDao {
 }
 ```
 
-If the annotation has a `customWhereClause`, it completely replaces the WHERE clause. The provided
+If the annotation has a [customWhereClause()], it completely replaces the WHERE clause. The provided
```
adutra, line 54: 1. This method has already been referenced line 16 above, I think the link would be more appropriate there than here.
2. Nit: you can use the following syntax to preserve monospace formatting inside links: 

```
[`customWhereClause()`][customWhereClause()]
```

manual/mapper/daos/select/README.md
```diff
@@ -60,6 +60,20 @@ for each, with the same name and a compatible Java type.
...
+whose values will be provided through additional method parameters. Note that it is sometimes not
+possible to determine if a parameter is a primary key component or a placeholder value; therefore
+the rule is that **if your method takes a partial primary key, all other parameters must be
+explicitly annotated with [@CqlName](../../entities/#user-provided-names)**. For example if the
+primary key is `((day int, hour int, minute int), ts timestamp)`:
+
+```java
+// Annotate 'l' so that it's not mistaken for the second PK component
+@Select(limit = ":l")
+PagingIterable<Sale> findDailySales(int day, @CqlName("l") int l);
```
adutra, line 75: I agree with the principle and am OK with it. Out of curiosity, have you considered the below alternative?

```
@Select
PagingIterable<Sale> findDailySales(int day, @Limit int limit);
```

olim7t: Yes, but we would need one such annotation for every secondary clause (`@PartitionLimit`, etc.), and this also assumes that there can only be one bind marker per clause. Currently this is the case, but this could change in the future (e.g. when `GROUP BY` accepts functions).

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoSelectMethodGenerator.java
olim7t: The annotation value is an array, you would have one element per column:
```java
@Select(orderBy = { "cc DESC", "aa DESC" })
```
I will amend the javadocs and manual to explicitly include an example.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoSelectMethodGenerator.java
olim7t: Correct. I find that "all non-PK parameters" is almost simpler to explain, but technically it's only the first one. I'll change it.

manual/mapper/daos/select/README.md
olim7t: 1. will do
2. I tend to avoid other formatting when the text is already a link, monospace in particular doesn't render very well IMO.

tomekl007 (approved):

PR #1296, opened 2019-07-10
Mapper uses Session to allow for pluggable request proccessors (e.g. GuavaSession etc)

fromanator:
Motivation:

The initial implementation of Mapper required a CqlSession be passed to it to use for queries.
However when trying to implement a custom request processor following the GuavaSession example this
caused issues:

Caused by: java.lang.IllegalArgumentException: No request processor found for com.datastax.oss.driver.internal.core.cql.DefaultPrepareRequest@2d88361b
    at com.datastax.oss.driver.internal.core.session.RequestProcessorRegistry.processorFor(RequestProcessorRegistry.java:74)
    at com.datastax.oss.driver.internal.core.session.DefaultSession.execute(DefaultSession.java:205)
    at com.datastax.oss.driver.api.core.CqlSession.prepareAsync(CqlSession.java:196)
    at com.datastax.oss.driver.internal.mapper.DaoBase.prepare(DaoBase.java:60)
    at CarDAOImpl__MapperGenerated.initAsync(CarDAOImpl__MapperGenerated.java:105)
    ... 33 more

Since I overrode the CqlSession with my own that would handle asynchronous queries using Ratpack Promies (similar to Guava ListenableFuture).

Thus I felt it made sense to change the Mapper code to use the more generic Session interface, since this allows for better support for
those who are using their own custom request processeors.

Result:

I had added a simple test in where the Mapper is configured to use the existing GuavaSession example instead of the CqlSession, and it all works.

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/example/guava/mapper/CarITBase.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/example/guava/mapper/CarITBase.java
@@ -1,0 +1,121 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.example.guava.mapper;
+
+import com.datastax.oss.driver.api.mapper.annotations.Entity;
+import com.datastax.oss.driver.api.mapper.annotations.PartitionKey;
+import com.datastax.oss.driver.api.testinfra.ccm.CcmRule;
+import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
+import java.util.List;
+import java.util.Objects;
+import java.util.UUID;
+
+public abstract class CarITBase {
+  protected static GuavaSessionGetEntityIT.Car SAAB93 =
+      new GuavaSessionGetEntityIT.Car(UUID.randomUUID(), "SAAB", "9-3", 1999);
+  protected static GuavaSessionGetEntityIT.Car SAAB900 =
+      new GuavaSessionGetEntityIT.Car(UUID.randomUUID(), "SAAB", "900", 1993);
+
+  protected static List<String> createStatements(CcmRule ccmRule) {
+    ImmutableList.Builder<String> builder =
+        ImmutableList.<String>builder()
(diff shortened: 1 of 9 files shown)


olim7t:
Hi,

That's how we had it in early mapper prototypes, but since the mapper expects the default CQL processors to be present anyway, it wasn't much of a stretch to require the `CqlSession` interface, and that makes the generated code (and user-provided `@QueryProvider` implementations) slightly simpler.

Can't you have you custom session implement `CqlSession`? I understand that you might not want to expose it everywhere, but with a hierarchy of interfaces it should be manageable:

```java
interface MySession extends CqlSession, RatpackSession
```

Either your session builder returns a `MySession`, and you pass `(RatpackSesssion)session` to client code that shouldn't have access to the CQL stuff.
Or the other way: your session builder declares `RatpackSession` as the return type for `build()`, but you know you can safely cast to `CqlSession` when you build a mapper instance.

fromanator:
Ah I think I kind of follow.

If you look at how the `GuavaSession` and `GuavaDriverContext` examples it looks like it excludes the `CqlRequestAsyncProcessor` and `CqlPrepareAsyncProcessor` from being registered as processors, and instead has the `GuavaSession` handle  `ASYNC` and `ASYNC_PREPARED`. So if we changed the `GuavaSession` to handle `LISTENABLE_FUTURE` and `LISTENABLE_FUTURE_PREPARED` and also registered the default CqlRequest processors it would work.

I think the challenge actually is (that my pr doesn't address) having the mapper return the asynchronous type that you want to use. For instance if `GuavaSession` extended off `CqlSession` so it can be used with the mapper, you would want something like this for the mapper:

```
@Select
ListenableFuture<Product> findByIdAsync(UUID productId);
```

However you're forced to have this instead:

```
@Select
CompletionStage<Product> findByIdAsync(UUID productId);
```

Thus you would have to put some glue code around your mapper calls to convert the `CompletionStage` into your preferred `ListenableFuture` return type to be used in your code.

However I think that supporting this use case of generic return types in the mapper and code generating it would be pretty challenging (but not impossible). For now I may just have to have some wrapper code to convert the `CompletionStage` returned by the mapper into my desired type.

olim7t:
As it turns out, that logic to plug custom return types already exists in the mapper 😃 (we have plans to support additional types in an upcoming release of the DSE driver). But your intuition is correct, it is pretty involved: you would have to build your own mapper artifact in order to subclass `MapperProcessor`; also it can't simply depend on the original artifact, because you want to replace the service descriptor so that only the new processor runs (we solve that by shading the original JAR). And finally you're coding against internal APIs.

To be honest, I think that's a bit too complicated for a 3rd-party implementation. It's probably more reasonable to wrap. You can either declare default methods directly on the DAO interfaces (but then the original methods are still visible to the client); or just write wrapper types for the mapper and DAOs.

fromanator:
Yeah I'll probably go the wrapping route, more stable than integrating with the internal API, and easier.

I think I will update this PR to revert changing CqlSession -> Session and instead just update the Guava example to extend off CqlSession and show how you can use the mapper with it. Hoping that the code example would be useful for others.

Right now working on the Ratpack Promise implementation as an internal company project. But with the plan of eventually open sourcing it. 

So far very happy with the direction the new driver is going, it makes it a lot more versatile and compostable for what can be implemented now versus the old 3.x codebase.

PR #1298, opened 2019-07-11
JAVA-2332: Destroy connection pool when a node gets removed

olim7t:


core/src/main/java/com/datastax/oss/driver/internal/core/session/PoolManager.java
```diff
@@ -314,10 +314,14 @@ private void processStateEvent(NodeStateEvent event) {
...
-      } else if (newState == NodeState.FORCED_DOWN) {
+      } else if (newState == null || newState == NodeState.FORCED_DOWN) {
         ChannelPool pool = pools.remove(node);
         if (pool != null) {
-          LOG.debug("[{}] {} was FORCED_DOWN, destroying pool", logPrefix, node);
+          LOG.debug(
+              "[{}] {} was {}, destroying pool",
+              logPrefix,
+              node,
+              newState == null ? "removed" : "forced down");
```
emerkle826: total nit here: Does it matter much that the debug log changes `FORCED_DOWN` to `forced down` in those cases? I realize it's just a log, and a Debug log at that. Again, total nitpick.

emerkle826 (approved):
LGTM :+1:

core/src/main/java/com/datastax/oss/driver/internal/core/session/PoolManager.java
olim7t: You're right, debug logs don't matter much but there's no reason to change it either. If anything, using the constant name gives you a hint of what to look for in the source code.
I'll change it back.

core/src/main/java/com/datastax/oss/driver/internal/core/session/PoolManager.java
```diff
@@ -314,10 +314,14 @@ private void processStateEvent(NodeStateEvent event) {
       NodeState newState = event.newState;
       if (pending.containsKey(node)) {
         pendingStateEvents.put(node, event);
-      } else if (newState == NodeState.FORCED_DOWN) {
+      } else if (newState == null || newState == NodeState.FORCED_DOWN) {
```
adutra, line 317: So when the state is `null` that means the node has been removed? Can we document this somewhere?

olim7t: Yes, `NodeStateEvent` is a bit bare-bones: added is (null, UNKNOWN) and removed is (null, null). I'll add javadocs to the fields.

PR #1300, opened 2019-07-17

tomekl007:
todo:
- [x] for the case where keyspace == null, try session.getKeyspace()
- [x] enabling schema validation through an annotation is a bit too static. It would probably be more flexible with a runtime parameter:
```
new InventoryMapperBuilder(session).setSchemaValidation(true).build()
```
- [x] trying to match a table and then falling back to a UDT is a good default. But maybe we could also have an optional annotation to provide hints to the mapper (keeping the name generic because other hints might be useful in the future):
```
@Entity
@SchemaHint(targetElement = TABLE)
public class Product { ... }
```
- [x] we could also validate the partition key and clustering columns
- [x] propagate `isSchemaValidationEnabled` in `InventoryMapperBuilder` to prevent backward compatibility problems
- [x] similarly, I think we have enough information to validate the types: we know the Java type from EntityDefinition, and the CQL type at runtime from the schema metadata. So we could call codecRegistry.codecFor at runtime, and fail if it returns no codec.
- [x] we recently introduced an EntityHelperBase parent class. There's probably opportunities to move some code there instead of generating it, for example methods such as:
void validateColumnExists(TableMetadata table, String columnName, GenericType<?> javaType)
- [x] 👍 for reporting all missing columns instead of failing on the first one. We could extend this to all the entity helpers of a given DAO, this way we validate the DAO's complete schema in one go.
- [x] docs


tomekl007:
comments in closed branch: #1245

tomekl007:
Comments:
https://github.com/datastax/java-driver/pull/1245

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DefaultMapperContext.java
```diff
@@ -107,6 +130,11 @@ public NameConverter getNameConverter(Class<? extends NameConverter> converterCl
     return customState;
   }
 
+  @Override
+  public boolean isSchemaValidationEnabled() {
```
tomekl007: @olim7t  according to your suggestion:
```
enabling schema validation through an annotation is a bit too static. It would probably be more flexible with a runtime parameter:
new InventoryMapperBuilder(session).setSchemaValidation(true).build()
```
I've moved the check into Builder. I've created a new field `schemaValidationEnabled`  but it obviously breaks the backward compatibility. I can put that schemaValidatEnabled field into `customState` and propagate it there. Let me know wdyt.
Also, I can set the default to be `true` or `false` if we want to have that behavior

tomekl007:
@olim7t 
concerning this:

>similarly, I think we have enough information to validate the types: we know the Java type from EntityDefinition, and the CQL type at runtime from the schema metadata. So we could call codecRegistry.codecFor at runtime, and fail if it returns no codec.

I was trying to achieve this but there are a couple of problems:
1. I see no way for the java type from `EntityDefinition`  to be transformed to `GenericType`. It returns `PropertyType` and there is a method `asTypeName()` that returns `TypeName`. I don't see a way for transforming `TypeName` to `GenericType`. I've tried the:
`CodecRegistry.DEFAULT.codecFor(asTypeName().class)`, it worked for simple types:
` CodecRegistry.DEFAULT.codecFor(java.lang.String.class)`
but for collections and UDTs it will not work - `CodecRegistry.DEFAULT.codecFor(java.util.Set<java.lang.String>)`
We can do some conditional logic and create `GenericType.ofSet(..)`, `ofList(...)` etc, but I think it is a very fragile solution (and still the problem with UDT handling persists)
Do you see a viable solution for it?
2. the second problem is that DriverContext is not accessible within `mapper` code. I cannot get `codecRegistry` easily. I can use `CodecRegistry.DEFAULT` but we are losing the types that were added by the user.
3. I don't get this part: ` So we could call codecRegistry.codecFor at runtime` - we can do it for types that are returned by `tableMetadata.get()).getColumns()` but this is not validation that we need. Also, assuming that we are able to transform `TypeName` to `GenericType` we can validate if the codecs are present but this has nothing to do with validation of discrepancies between Mapper entity definition and actual CQL table schema. I think that an ideal solution will be to transform types of columns defined in the mapper to `GenericType`, compare if they are matching the types in CQL tables, and if not report the columns that have different types: (for example `the c1 column defined in the Entity has type boolean but the c1 column in CQL table has type String). So this `codecRegistry.codecFor` would be used only to transform the entity defined column into CodecRegistry, and the 2nd phase of validation does the actual diff vs table. (of course, it is impossible without solving the 1st issue) 

let me know wdyt about it.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DefaultMapperContext.java
tomekl007: solved by using `customState`

tomekl007:
>👍 for reporting all missing columns instead of failing on the first one. We could extend this to all the entity helpers of a given DAO, this way we validate the DAO's complete schema in one go.

Are you sure that this is a good solution? I think that from the UX perspective it will be not-readable and bit too verbose. If someone will have multiple `@Entity` objects in the `@Dao` and we will return a huge error message with all configuration problems, a user may be overwhelmed by it.
If we are reporting problems for one `@Entity`, the content of error will be smaller and the user will be able to fix one `@Entity` problem at a time, then go the next one if needed. @olim7t wdyt?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
```diff
@@ -0,0 +1,261 @@
...
+        "$1T keyspaceId = context.getKeyspaceId() != null ? context.getKeyspaceId() : context.getSession().getKeyspace().orElse(null)",
+        CqlIdentifier.class);
+
+    // Handle case where keyspaceId = null. In such case we cannot infer and validate schema for
+    // table
+    // or udt
+    methodBuilder.beginControlFlow(
+        "if (!context.getSession().getMetadata().getKeyspace(keyspaceId).isPresent())");
+    methodBuilder.addStatement(
+        "return"); // todo maybe we should do log.warn("we cannot validate schema because keyspace
```
tomekl007: should we add `warn` for that case or fail-silently?

olim7t:
>  DriverContext is not accessible within mapper code

You can get it via the `MapperContext`: `context.getSession().getContext().getCodecRegistry()`

olim7t:
> I don't get this part: So we could call codecRegistry.codecFor at runtime

If we're able to generate this for each property:
```
DataType cqlType = ...
GenericType<...> javaType = ...
```

Then, if `codecFor(cqlType, javaType)` doesn't throw, the Java<->CQL mapping is correct.

We could also generate a `Class<...> javaType` instead for simple types, but I'm not sure if it's worth making that distinction, the validation code is not on the hot path. Generic type for all cases should be good enough, so that simplifies the processor code.

> I see no way for the java type from `EntityDefinition` to be transformed to `GenericType`.

I think it will be similar to the way we already do it when we call getters and setters. For example, this snippet from `EntityPolymorphismIT_CircleHelper__MapperGenerated`:

```java
private static final GenericType<Set<String>> GENERIC_TYPE = new GenericType<Set<String>>(){};
[...]
    if (entity.getTags() != null || nullSavingStrategy == NullSavingStrategy.SET_TO_NULL) {
      target = target.set("tags", entity.getTags(), GENERIC_TYPE);
    }
```

Which is handled [here](https://github.com/datastax/java-driver/blob/4.x/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java#L252). We do generate the construction of the `GenericType` from the `PropertyType`. From what I see your validation code also lives in the helper, so you'll be able to reuse `enclosingClass.addGenericTypeConstant` for your needs.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
olim7t: I think we should warn: if `schemaValidationEnabled == true` and the validation can't run, we should let the user know.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MapperBuilder.java
```diff
@@ -33,13 +33,17 @@
...
+      "datastax.mapper.schemaValidationEnabled";
   protected final CqlSession session;
   protected CqlIdentifier defaultKeyspaceId;
   protected Map<Object, Object> customState;
 
   protected MapperBuilder(CqlSession session) {
     this.session = session;
     this.customState = new HashMap<>();
+    // schema validation is enabled by default
+    customState.put(SCHEMA_VALIDATION_ENABLED_SETTING, true);
```
olim7t, line 46: I would have used `false` as the default, but now that I think about it I can see pros and cons on both sides:
* enabled by default: errors are reported sooner. But if users don't realize that they can disable it, the (possibly expensive?) checks will run every time a new DAO gets created.
* disabled by default: more efficient by default. But errors are only reported at usage time, and might be less clear (stack trace involving generated mapper classes). Also they will be detected one at a time.

In the end I think I agree with enabled by default.

tomekl007: I agree - I think that the overhead of checking for every `Dao` maybe not so big because it is happening only one time at the compile time.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
```diff
@@ -0,0 +1,309 @@
...
+    // In such case we cannot infer and validate schema for table or udt
+    methodBuilder.beginControlFlow(
+        "if (!context.getSession().getMetadata().getKeyspace(keyspaceId).isPresent())");
+    loggingGenerator.warn(
+        methodBuilder,
+        "[{}] Unable to validate table: {} for the entity class: {} because keyspace: {} is not present",
+        CodeBlock.of("context.getSession().getName()"),
+        CodeBlock.of("tableId"),
+        CodeBlock.of("entityClassName"),
+        CodeBlock.of("keyspaceId"));
```
olim7t: Nit: `keyspaceId` might be null. In fact, I think this will be the most common reason for this warning: when neither the session nor the DAO was created with a keyspace. So I think we should handle it separately with a more descriptive error message than "keyspace null is not present".

(The other possible reasons are: keyspace name present but wrong, or keyspace name present but metadata is out of date.)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
```diff
@@ -0,0 +1,309 @@
...
+        ArrayList.class);
+    for (CodeBlock expectedCqlName : expectedCqlNames) {
+      methodBuilder.addStatement(
+          "expectedCqlNames.add($1T.fromCql($2L))", CqlIdentifier.class, expectedCqlName);
+    }
+
+    methodBuilder.addStatement(
+        "$1T<$2T> tableMetadata = context.getSession().getMetadata().getKeyspace(keyspaceId).flatMap(v -> v.getTable(tableId))",
+        Optional.class,
+        TableMetadata.class);
```
olim7t, line 100: Nit: could you extract `context.getSession().getMetadata().getKeyspace(keyspaceId)` and reuse it for table and UDT?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
```diff
@@ -0,0 +1,309 @@
...
+    }
+  }
+
+  // Finds out missingTableCqlNames - columns that are present in Entity Mapping but NOT present in
+  // CQL table
+  private void findMissingColumnsInTable(MethodSpec.Builder methodBuilder) {
+    methodBuilder.beginControlFlow("if (tableMetadata.isPresent())");
+
+    methodBuilder.addComment("validation of missing Clustering Columns");
+    generateMissingClusteringColumnsCheck(methodBuilder);
```
olim7t, line 185: Nit: we could skip this if there are no clustering columns, that would save a few unnecessary lines in the generated code. For example in `UserHelper__MapperGenerated`:
```java
      // validation of missing Clustering Columns
      List<CqlIdentifier> expectedCqlClusteringColumns = new ArrayList<>();
      List<CqlIdentifier> missingTableClusteringColumnNames = findMissingColumns(expectedCqlClusteringColumns, tableMetadata.get().getClusteringColumns().keySet());
      if (!missingTableClusteringColumnNames.isEmpty()) {
        throw new IllegalArgumentException(String.format("The CQL ks.table: %s.%s has missing Clustering columns: %s that are defined in the entity class: %s", keyspaceId, tableId, missingTableClusteringColumnNames, entityClassName));
      }
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
```diff
@@ -0,0 +1,309 @@
...
+    methodBuilder.addStatement(
+        "$1T<$2T> missingTableCqlNames = findMissingColumnsCql(expectedCqlNames, (($3T) tableMetadata.get()).getColumns().keySet())",
+        List.class,
+        CqlIdentifier.class,
+        DefaultTableMetadata.class);
+
+    // Throw if there are any missingTableCqlNames
+    CodeBlock missingCqlColumnExceptionMessage =
+        CodeBlock.of(
+            "String.format(\"The CQL ks.table: %s.%s has missing columns: %s that are defined in the entity class: %s\", "
```
olim7t, line 219:
```suggestion
            "String.format(\"The CQL table: %s.%s has missing columns: %s that are defined in the entity class: %s\", "
```
?


Also occurs a couple of other times below.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/entity/EntityHelperBase.java
```diff
@@ -75,4 +86,70 @@ protected void throwIfKeyspaceMissing() {
...
+
+  public List<CqlIdentifier> findMissingColumns(
+      List<CqlIdentifier> entityColumns, Collection<ColumnMetadata> cqlColumns) {
+    return findMissingColumnsCql(
+        entityColumns,
+        cqlColumns.stream().map(ColumnMetadata::getName).collect(Collectors.toList()));
+  }
+
+  public List<CqlIdentifier> findMissingColumnsCql(
+      List<CqlIdentifier> entityColumns, Collection<CqlIdentifier> cqlColumns) {
```
olim7t, line 101: Naming is a bit confusing here, it's not immediately clear how `findMissingColumns` / `findMissingColumnsCql` differ.
The only reason to extract the latter is because you're passing `tableMetadata.get()).getColumns().keySet()` when validating regular columns. You could pass `tableMetadata.get()).getColumns().values()` instead, and merge the two methods.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/entity/EntityHelperBase.java
```diff
@@ -75,4 +86,70 @@ protected void throwIfKeyspaceMissing() {
               DaoKeyspace.class.getSimpleName()));
     }
   }
+
+  public List<CqlIdentifier> findMissingColumns(
+      List<CqlIdentifier> entityColumns, Collection<ColumnMetadata> cqlColumns) {
```
olim7t, line 94: I was going to comment on the fact that the generated code boxes every identifier before invoking this method:
```java
      List<CqlIdentifier> expectedCqlPKs = new ArrayList<>();
      expectedCqlPKs.add(CqlIdentifier.fromCql("userid")); // <== LIKE THIS
      List<CqlIdentifier> missingTablePksNames = findMissingColumns(expectedCqlPKs, tableMetadata.get().getPartitionKey());
```
But that's actually the right way to do it. These identifiers may be provided directly by the user with a `CqlName` annotation, so comparing them directly to `cqlColumn.asCql(...)` would not be reliable.

So nothing to change here 👍

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
```diff
@@ -0,0 +1,309 @@
...
+    CodeBlock missingCqlUdtExceptionMessage =
+        CodeBlock.of(
+            "String.format(\"The CQL ks.udt: %s.%s has missing columns: %s that are defined in the entity class: %s\", "
+                + "keyspaceId, tableId, missingTableCqlNames, entityClassName)");
+    methodBuilder.beginControlFlow("if (!missingTableCqlNames.isEmpty())");
+    methodBuilder.addStatement(
+        "throw new $1T($2L)", IllegalArgumentException.class, missingCqlUdtExceptionMessage);
+    methodBuilder.endControlFlow();
+
+    methodBuilder.endControlFlow();
```
olim7t: We should also check the types for UDTs.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
```diff
@@ -0,0 +1,309 @@
...
+    }
+    // if explicitly provided SchemaHint is UDT, then generate only UDT check
+    else if (targetElement.get().equals(TargetElement.UDT)) {
+      findMissingColumnsInUdt(methodBuilder, false);
+    }
+  }
+
+  // Finds out missingTableCqlNames - columns that are present in Entity Mapping but NOT present in
+  // CQL table
+  private void findMissingColumnsInTable(MethodSpec.Builder methodBuilder) {
```
olim7t: Nit: this also validates the types now, so the name is not accurate.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/entity/EntityHelperBase.java
```diff
@@ -75,4 +86,70 @@ protected void throwIfKeyspaceMissing() {
...
+      Map<CqlIdentifier, ColumnMetadata> cqlColumns,
+      CodecRegistry codecRegistry) {
+    List<String> missingCodecs = new ArrayList<>();
+
+    for (Map.Entry<CqlIdentifier, GenericType<?>> entityEntry : entityColumns.entrySet()) {
+      ColumnMetadata columnMetadata = cqlColumns.get(entityEntry.getKey());
+      if (columnMetadata == null) {
+        // this will not happen because it will be catch by the generateMissingColumnsCheck() method
+        throw new IllegalArgumentException(
+            "There is no cql column for entity column: " + entityEntry.getKey());
```
olim7t, line 158: For this kind of impossible case, I think `AssertionError` is the most appropriate type to throw.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/entity/EntityHelperBase.java
```diff
@@ -75,4 +86,70 @@ protected void throwIfKeyspaceMissing() {
...
+   * When the new instance of a class annotated with {@link Dao} is created an automatic check for
+   * schema validation is performed. It verifies if all {@link Dao} entity fields are present in CQL
+   * table. If not the {@link IllegalArgumentException} exception with detailed message is thrown.
+   * This check has startup overhead so once your app is stable you may want to disable it. The
+   * schema validation check is enabled by default. It can be disabled using the {@link
+   * MapperBuilder#withSchemaValidationEnabled(boolean)} method.
+   */
+  public abstract void validateEntityFields();
+
+  public static List<String> findMissingTypes(
```
olim7t: Nit: it's not really that the types are "missing", more that they don't match. Maybe `findTypeMismatches`?

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/entity/EntityHelperBase.java
```diff
@@ -75,4 +86,70 @@ protected void throwIfKeyspaceMissing() {
...
+  public void throwMissingTypesIfNotEmpty(
+      List<String> missingTypes,
+      CqlIdentifier keyspaceId,
+      CqlIdentifier tableId,
+      String entityClassName) {
+    if (!missingTypes.isEmpty()) {
+      throw new IllegalArgumentException(
+          String.format(
+              "The CQL ks.table: %s.%s defined in the entity class: %s has wrong types:\n%s",
+              keyspaceId, tableId, entityClassName, String.join("\n", missingTypes)));
```
olim7t: "has wrong types" => "declares type mappings that are not supported by the codec registry"

changelog/README.md
```diff
@@ -4,6 +4,7 @@
 
 ### 4.4.0 (in progress)
 
+- [new feature] JAVA-2263: Add optional schema validation
```
olim7t: Since there's less context here:
```suggestion
- [new feature] JAVA-2263: Add optional schema validation to the mapper
```

olim7t (changes requested):
This looks great.

Last round of comments, the "nits" are mostly cosmetic so feel free to apply at your discretion.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
tomekl007: the `ks.table` is a format of the message - in cases for UDt it is `ks.udt`. 
I think we should keep it because it gives the user information about the type of the validated entity.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/entity/EntityHelperBase.java
tomekl007: I will still need this method for udt check, see:
```
 else if (userDefinedType.isPresent()) {
      List<CqlIdentifier> columns = userDefinedType.get().getFieldNames();
      List<CqlIdentifier> missingTableCqlNames = findMissingColumnsCql(expectedCqlNames, columns);
      if (!missingTableCqlNames.isEmpty()) {
        throw new IllegalArgumentException(String.format("The CQL ks.udt: %s.%s has missing columns: %s that are defined in the entity class: %s", keyspaceId, tableId, missingTableCqlNames, entityClassName));
      }
    }
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
tomekl007: The error looks like this:
`Unable to validate table: {} for the entity class: {} because keyspace: {} is not present`
so it contains table and entity (not only ks that is `null`). Could you give me an example of how this message could look like to improve it (because I am not sure how to make it better for this case)?

tomekl007: I've expanded this check to two (https://github.com/datastax/java-driver/pull/1300/commits/0d16fe9a2133385acb4baa71446f5e19cfd1403e):
1. lookup the `keyspaceId` in the `context.getSession().getMetadata().getKeyspaces()` -> if not present, then report that provided keyspace is not pressent in metadata
2. When the `keyspaceId`  is present but returned `keyspace` is of a type `Optional.empty` then report that keyspace is not present.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
```diff
@@ -109,18 +109,25 @@ public EntityHelperSchemaValidationMethodGenerator(
...
-    methodBuilder.addStatement(
-        "throw new $1T($2L)", IllegalArgumentException.class, missingKeyspaceTableExceptionMessage);
-    methodBuilder.endControlFlow();
+    logMissingMetadata(methodBuilder);
 
     return Optional.of(methodBuilder.build());
   }
 
+  private void logMissingMetadata(MethodSpec.Builder methodBuilder) {
+    methodBuilder.addComment(
```
tomekl007, line 116: There were some failures of https://jenkins-drivers.build.dsinternal.org/job/datastax.java-driver.java2263_rebase.default/lastCompletedBuild/testReport/ `DefaultKeyspaceId`, it reported that the table is not present (it fails only on jenkins) I suspect that there is some kind of race condition in metadata and this is related to your comment:
>keyspace name present but metadata is out of date

I don't know how can we alleviate that problem - maybe we could refresh the schema explicitly?
If not, I think the viable solution is to `log.warning`:
`warn if there are is keyspace.table for defined entity - it means that table is missing, or schema it out of date.`
instead of throwing an exception and fail fast if we cannot be sure if the schema was out-dated.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
tomekl007: > keyspace name present but metadata is out of date
addressed in this commit:  https://github.com/datastax/java-driver/pull/1300/commits/9442728d5e0fa5274c283bf4c1ade0f2ed2c873d
See my comment: https://github.com/datastax/java-driver/pull/1300#discussion_r358248749
for more explanation.

olim7t: Hrmm I don't get it, the two checks are equivalent. From the generated code with the new version:
```java
if(!keyspaceNamePresent(context.getSession().getMetadata().getKeyspaces(), keyspaceId)) {
  ... // warning 1
}
Optional<KeyspaceMetadata> keyspace = context.getSession().getMetadata().getKeyspace(keyspaceId);
if (!keyspace.isPresent()) {
  ... // warning 2
}
```
The two checks to the same thing: if the key is not present in the map, then getKeyspace returns an empty result.

My suggestion was more in the lines of:
```java
CqlIdentifier keyspaceId = context.getKeyspaceId() != null ? context.getKeyspaceId() : context.getSession().getKeyspace().orElse(null);
if (keyspaceId == null) {
  ... // DAO will run unqualified requests on an unqualified session: probably a mapper configuration error, the requests will fail at runtime
}
Optional<KeyspaceMetadata> keyspace = context.getSession().getMetadata().getKeyspace(keyspaceId);
if (!keyspace.isPresent()) {
  ... // what you already have for this case
}
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
olim7t: If you just say "table" or "udt" it's enough to indicate the type of entity.

I see the intent now, but it looks a bit unusual in a plain english message. Also if you just say "the table inventory.product", it should be pretty obvious enough that "inventory" is the keyspace.

But that's just a cosmetic detail so I don't want to dwell on it too much, I'll leave the final decision to you.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/entity/EntityHelperBase.java
olim7t: I see. Maybe rename it to `findMissingIdentifiers` then...

olim7t (commented):
Hrmm I thought I only replied directly to existing comments, but Github is forcing me to create another review... Hope nothing was lost in the process.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
tomekl007: They are not equivalent. 
1. The first check is checking the case that you mentioned - so in the case of `keyspaceId == null` it will log warn and progress to unqualified requests. But besides that, it is also checking if the keyspace (if provided as not null) is present in the metadata, see:
```
 if(!keyspaceNamePresent(context.getSession().getMetadata().getKeyspaces(), keyspaceId)) {
}
.....
public boolean keyspaceNamePresent(
      Map<CqlIdentifier, KeyspaceMetadata> keyspaces, CqlIdentifier keyspaceId) {
    return keyspaces.keySet().contains(keyspaceId);
  }
```
(`null` will be not present so it will catch your first case.
Please take a look at this test: `SchemaValidationIT#should_log_warning_when_passing_not_existing_keyspace`
2. This second case:
```
 Optional<KeyspaceMetadata> keyspace = context.getSession().getMetadata().getKeyspace(keyspaceId);
    if (!keyspace.isPresent()) {
      LOG.warn("[{}] Unable to validate table: {} for the entity class: {} because keyspace: {} is not present.",
          context.getSession().getName(),
          tableId,
          entityClassName,
          keyspaceId);
      return;
    }
```
is getting the actual `KespaceMetadata` for not null `keyspaceId` (if it was null, then previous case will catch that)

So I think that the current solution is equivalent to your suggestion + it is adding the validation of the existence of the keyspace name.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
tomekl007: When writing tests and debugging it was useful to me have that `ks.udt` and `table.udt` in logs so I will leave those in the logs if you don't mind

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperSchemaValidationMethodGenerator.java
olim7t: These two conditions are equivalent:
```java
session.getMetadata().getKeyspaces().keySet().contains(keyspaceId)
session.getMetadata().getKeyspace(keyspaceId).isPresent()
```
Except if you assume that:
* the map can contain null values (which is in fact impossible because our implementation relies on Guava's `ImmutableMap`, which doesn't allow it)
* or that there is a bug in `getKeyspace()`.

So the first check already covers the second check. Try to write a test that produces the second warning.

We have two error conditions to cover:
1. the DAO was not built with sufficient information to determine a keyspace id
2. it was, but the keyspace id does not exist

Currently, the first check does both 1 and 2, and the second check does 2 (but is unreachable).

tomekl007: Ah, you right - sorry for back and forth on this.
I've switch checks (according to your suggestion here: https://github.com/datastax/java-driver/pull/1300#discussion_r364867139) to first check if `keyspaceId == null` and in that case give a detailed message. Otherwise the `keyspaceNamePresent` check.

olim7t (approved):

olim7t:
I had a few more trivial remarks, I pushed the changes directly because I don't think they are controversial.
LGTM.

PR #1301, opened 2019-07-18
JAVA-2338: Revisit toString() for data container types

olim7t:
Contrary to what I initially suggested on the ticket, I didn't sanitize the output, because the existing implementations for tuples/UDTs didn't do it. I think it's a corner case anyway, and if needed it's pretty easy to reimplement in your own utility method.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/cql/Row.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/cql/Row.java
@@ -19,6 +19,7 @@ import com.datastax.oss.driver.api.core.data.GettableById;
 import com.datastax.oss.driver.api.core.data.GettableByIndex;
 import com.datastax.oss.driver.api.core.data.GettableByName;
 import com.datastax.oss.driver.api.core.detach.Detachable;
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/cql/Row.java
```diff
@@ -33,4 +34,55 @@
...
+      String name = definition.getName().asCql(true);
+      TypeCodec<Object> codec = codecRegistry().codecFor(definition.getType());
+      Object value = codec.decode(getBytesUnsafe(i), protocolVersion());
+      result.append(name).append(':').append(codec.format(value));
+    }
+    return result.append("]").toString();
+  }
+
+  /**
+   * Returns an abstract representation of this object, <b>that does not include the row's
```
adutra: The method contracts specified here are implementation-dependent. I would suggest that we stress that these characteristics (no row's contents, default format of `toString` etc.) are only valid as long as the actual implementation is a driver built-in one.

olim7t: 👍 will do

adutra (approved):

PR #1303, opened 2019-07-20
JAVA-2302: Better target mapper errors and warnings for inherited methods

olim7t:
Motivation:

If
- the mapper processes a class or interface that inherits from another
  type
- that parent type is not part of the current compilation round (e.g.
  it's in a JAR dependency of the current project)
- that parent type is not annotated correctly

Then the errors or warnings issued by the mapper processor do not
provide enough context, because they can't be linked to a source file
and line number.

Modifications:

Wrap all field- and method-level messages in a `TypeMemberMessager`,
that detects the problematic situation, and issues the messages on the
child type instead.

Result:

The errors or warnings will appear on a source file, with a contextual
prefix like "[getId() inherited from BaseEntity]".

### DIFF
--- a/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/DecoratedMessager.java
+++ b/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/DecoratedMessager.java
@@ -15,8 +15,11 @@
  */
 package com.datastax.oss.driver.internal.mapper.processor;
 
+import edu.umd.cs.findbugs.annotations.NonNull;
 import javax.annotation.processing.Messager;
 import javax.lang.model.element.Element;
+import javax.lang.model.element.ExecutableElement;
+import javax.lang.model.element.TypeElement;
 import javax.tools.Diagnostic;
 
 /** Wraps {@link Messager} to provide convenience methods. */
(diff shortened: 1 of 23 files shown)


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/TypeMemberMessager.java
```diff
@@ -0,0 +1,97 @@
...
+      @NonNull Element member,
+      @NonNull TypeElement processedType,
+      @NonNull ProcessorContext context) {
+
+    this.delegate = context.getMessager();
+
+    // The type that declares the method (BaseDao in the example above). That's the one that might
+    // be already compiled.
+    Element declaringType = member.getEnclosingElement();
+    assert declaringType instanceof TypeElement : "member must be a field or method";
```
tomekl007: According to documentation `TypeElement`: `Represents a class or interface program element`.
So this statement: `member must be a field or method` don't seems to be true

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/TypeMemberMessager.java
```diff
@@ -0,0 +1,97 @@
...
+
+    if (processedType.equals(declaringType) || isSourceFile((TypeElement) declaringType)) {
+      this.targetElement = member;
+      this.locationInfo = "";
+    } else {
+      this.targetElement = processedType;
+      this.locationInfo =
+          String.format("[%s inherited from %s] ", member, declaringType.getSimpleName());
+    }
```
tomekl007: what will happen if we will have such inheritance:
```
interface Parent {
	@Update u();
}

interface BaseDao extends Parent{
@Select
   void select();
}
@Dao
interface ConcreteDao extends BaseDao {}
```
and:
- only the Parent belongs to Jar dependency
- BaseDao and Parent belongs to Jar dependency
- BaseDao belongs to Jar dependency and Parent does not (I think it is impossible but just asking)
Are we able to catch such a hierarchy and add error info about all classes in a chain?
If it is already supported, maybe we can add an integration test that covers this?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/TypeMemberMessager.java
olim7t: The assertion is not on `member`, but on `member.getEnclosingElement()`. If `member` is a field or method, its enclosing element is a class or interface.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/TypeMemberMessager.java
olim7t: Yes, those cases will be handled correctly (and the last one is impossible indeed).

Check out `DefaultEntityFactory` and `DaoImplementationGenerator`. In both cases, we get all the parent types as a flat structure. What matters is the class we are currently processing (`processedClass`), and the type where we are inheriting the member from (`declaringType`). How many types there are in between has no impact on the algorithm used here.

tomekl007 (approved):

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/TypeMemberMessager.java
```diff
@@ -0,0 +1,97 @@
...
+      @NonNull ProcessorContext context) {
+
+    this.delegate = context.getMessager();
+
+    // The type that declares the method (BaseDao in the example above). That's the one that might
+    // be already compiled.
+    Element declaringType = member.getEnclosingElement();
+    assert declaringType instanceof TypeElement : "member must be a field or method";
+
+    if (processedType.equals(declaringType) || isSourceFile((TypeElement) declaringType)) {
```
adutra: I'm wonder if the current design is the best. If I understand correctly, if `processedType.equals(declaringType) || isSourceFile((TypeElement) declaringType)` is true, this class will behave exactly like `DecoratedMessager`. So wouldn't be a better design the make this class extend `DecoratedMessager` and add a factory method somewhere that will determine if it should create a "vanilla" `DecoratedMessager` or the special `TypeMemberMessager` for cases where the member in question is pre-compiled?
Also wouldn't `PreCompiledMessager` be a better name?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
```diff
@@ -70,12 +72,9 @@ public DaoDeleteMethodGenerator(
     Delete annotation = methodElement.getAnnotation(Delete.class);
     assert annotation != null;
     if (annotation.ifExists() && !annotation.customIfClause().isEmpty()) {
-      context
-          .getMessager()
-          .error(
-              methodElement,
-              "Invalid annotation parameters: %s cannot have both ifExists and customIfClause",
-              Delete.class.getSimpleName());
+      methodMessager.error(
```
adutra: Possibly related to my previous comment: wouldn't it be possible to obtain the messager from the context in all cases, and let the context decide which specific implementation to use?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/TypeMemberMessager.java
olim7t: Indeed, a weakness of the current design is that it relies on us using `TypeMemberMessager` instead of the default messager when necessary.

Maybe I could instead have all methods in `DecoratedMessager` receive `Element targetElement, TypeElement processedClass`, and do the detection every time.

I'll take a shot at it.

olim7t: Done, this looks much cleaner indeed. LMK what you think.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGeneratorTest.java
```diff
@@ -145,8 +145,7 @@ public void should_fail_with_expected_error(
   @Test
   public void should_warn_when_non_bind_marker_has_cql_name() {
     should_succeed_with_expected_warning(
-        "delete(java.util.UUID,java.lang.String): parameter id does not refer "
-            + "to a bind marker, @CqlName annotation will be ignored",
+        "Parameter id does not refer " + "to a bind marker, @CqlName annotation will be ignored",
```
adutra: Nit: strings can be joined.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/DecoratedMessager.java
```diff
@@ -28,15 +31,146 @@ public DecoratedMessager(Messager messager) {
...
+        case PARAMETER:
+          ExecutableElement method =
+              (ExecutableElement) intendedTargetElement.getEnclosingElement();
+          declaringType = (TypeElement) method.getEnclosingElement();
+          if (processedType.equals(declaringType) || isSourceFile(declaringType)) {
+            this.actualTargetElement = intendedTargetElement;
+            this.locationInfo = "";
+          } else {
+            this.actualTargetElement = processedType;
+            this.locationInfo =
```
adutra, line 144: The design is indeed much better now. However I'm not seeing any specific test exercising this case branch and the case branch above for classes and interfaces.

adutra (changes requested):

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/DecoratedMessager.java
olim7t: I implemented those cases because there were likely to happen in the future, but currently there is no message in the processor that covers them.

PR #1306, opened 2019-07-24
JAVA-2351: Add a driver example for the object mapper

olim7t:
TODO:
- [x] rebase on 4.x and remove the workaround for JAVA-2358

### DIFF
--- a/examples/src/main/java/com/datastax/oss/driver/examples/mapper/KillrVideoMapperExample.java
+++ b/examples/src/main/java/com/datastax/oss/driver/examples/mapper/KillrVideoMapperExample.java
@@ -1,0 +1,192 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.examples.mapper;
+
+import com.datastax.oss.driver.api.core.CqlIdentifier;
+import com.datastax.oss.driver.api.core.CqlSession;
+import com.datastax.oss.driver.api.core.PagingIterable;
+import com.datastax.oss.driver.api.core.cql.SimpleStatement;
+import com.datastax.oss.driver.examples.mapper.killrvideo.KillrVideoMapper;
+import com.datastax.oss.driver.examples.mapper.killrvideo.user.User;
+import com.datastax.oss.driver.examples.mapper.killrvideo.user.UserDao;
+import com.datastax.oss.driver.examples.mapper.killrvideo.video.LatestVideo;
+import com.datastax.oss.driver.examples.mapper.killrvideo.video.UserVideo;
+import com.datastax.oss.driver.examples.mapper.killrvideo.video.Video;
+import com.datastax.oss.driver.examples.mapper.killrvideo.video.VideoByTag;
+import com.datastax.oss.driver.examples.mapper.killrvideo.video.VideoDao;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+import java.util.stream.Collectors;
+
+/**
+ * Uses the driver's object mapper to interact with a schema.
+ *
+ * <p>We use the data model of the <a href="https://killrvideo.github.io/">KillrVideo</a> sample
+ * application. The mapped entities and DAOs are in the {@link
+ * com.datastax.oss.driver.examples.mapper.killrvideo} package. We only cover a subset of the data
+ * model (ratings, stats, recommendations and comments are not covered).
+ *
+ * <p>Preconditions:
+ *
+ * <ul>
+ *   <li>An Apache Cassandra(R) cluster is running and accessible through the contacts points
+ *       identified by basic.contact-points (see application.conf).
+ * </ul>
+ *
+ * <p>Side effects:
+ *
+ * <ul>
+ *   <li>creates a new keyspace "killrvideo" in the session. If a keyspace with this name already
+ *       exists, it will be reused;
+ *   <li>creates the tables of the KillrVideo data model, if they don't already exist;
+ *   <li>inserts a new user, or reuse the existing one if the email address is already taken;
+ *   <li>inserts a video for that user.
+ * </ul>
+ *
+ * @see <a href="https://docs.datastax.com/en/developer/java-driver/latest/manual/mapper/">Java
+ *     Driver Mapper manual</a>
+ */
+public class KillrVideoMapperExample {
+
+  private static final CqlIdentifier KEYSPACE_ID = CqlIdentifier.fromCql("killrvideo");
+
+  public static void main(String[] args) {
+
+    try (CqlSession session = CqlSession.builder().build()) {
+
+      maybeCreateSchema(session);
+
+      KillrVideoMapper mapper =
+          KillrVideoMapper.builder(session).withDefaultKeyspace(KEYSPACE_ID).build();
+
+      // Create a new user
+      UserDao userDao = mapper.userDao();
+
+      User user =
+          new User(UUID.randomUUID(), "test", "user", "testuser@example.com", Instant.now());
+
+      if (userDao.create(user, "password123".toCharArray())) {
+        System.out.println("Created " + user);
+      } else {
+        user = userDao.getByEmail("testuser@example.com");
+        System.out.println("Reusing existing " + user);
+      }
+
+      // Creating another user with the same email should fail
+      assert !userDao.create(
+          new User(UUID.randomUUID(), "test2", "user", "testuser@example.com", Instant.now()),
+          "secret123".toCharArray());
+
+      // Simulate login attempts
+      tryLogin(userDao, "testuser@example.com", "password123");
+      tryLogin(userDao, "testuser@example.com", "secret123");
+
+      // Insert a video
+      VideoDao videoDao = mapper.videoDao();
+
+      Video video = new Video();
+      video.setUserid(user.getUserid());
+      video.setName(
+          "Getting Started with DataStax Apache Cassandra as a Service on DataStax Constellation");
+      video.setLocation("https://www.youtube.com/watch?v=68xzKpcZURA");
+      Set<String> tags = new HashSet<>();
+      tags.add("apachecassandra");
+      tags.add("nosql");
+      tags.add("hybridcloud");
+      video.setTags(tags);
+
+      videoDao.create(video);
+      System.out.printf("Created video [%s] %s%n", video.getVideoid(), video.getName());
+
+      // Check that associated denormalized tables have also been updated:
+      PagingIterable<UserVideo> userVideos = videoDao.getByUser(user.getUserid());
+      System.out.printf("Videos for %s %s:%n", user.getFirstname(), user.getLastname());
+      for (UserVideo userVideo : userVideos) {
+        System.out.printf("  [%s] %s%n", userVideo.getVideoid(), userVideo.getName());
+      }
+
+      PagingIterable<LatestVideo> latestVideos = videoDao.getLatest(todaysTimestamp());
+      System.out.println("Latest videos:");
+      for (LatestVideo latestVideo : latestVideos) {
+        System.out.printf("  [%s] %s%n", latestVideo.getVideoid(), latestVideo.getName());
+      }
+
+      PagingIterable<VideoByTag> videosByTag = videoDao.getByTag("apachecassandra");
+      System.out.println("Videos tagged with apachecassandra:");
+      for (VideoByTag videoByTag : videosByTag) {
+        System.out.printf("  [%s] %s%n", videoByTag.getVideoid(), videoByTag.getName());
+      }
+
+      // Update the existing video:
+      Video template = new Video();
+      template.setVideoid(video.getVideoid());
+      template.setName(
+          "Getting Started with DataStax Apache Cassandra® as a Service on DataStax Constellation");
+      videoDao.update(template);
+      // Reload the whole entity and check the fields
+      video = videoDao.get(video.getVideoid());
+      System.out.printf("Updated name for video %s: %s%n", video.getVideoid(), video.getName());
+    } catch (Exception e) {
+      e.printStackTrace();
+    }
+  }
+
+  private static void tryLogin(UserDao userDao, String email, String password) {
+    Optional<User> maybeUser = userDao.login(email, password.toCharArray());
+    System.out.printf(
+        "Logging in with %s/%s: %s%n",
+        email, password, maybeUser.isPresent() ? "Success" : "Failure");
+  }
+
+  private static void maybeCreateSchema(CqlSession session) throws Exception {
+    session.execute(
+        SimpleStatement.newInstance(
+                "CREATE KEYSPACE IF NOT EXISTS killrvideo WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1}")
+            .setExecutionProfileName("slow"));
+    session.execute("USE killrvideo");
+    for (String statement : getStatements("killrvideo_schema.cql")) {
+      session.execute(SimpleStatement.newInstance(statement).setExecutionProfileName("slow"));
+    }
+  }
+
+  private static List<String> getStatements(String fileName) throws Exception {
+    Path path = Paths.get(ClassLoader.getSystemResource(fileName).toURI());
+    String contents = new String(Files.readAllBytes(path), StandardCharsets.UTF_8);
+    return Arrays.stream(contents.split(";"))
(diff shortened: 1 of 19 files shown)


examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/user/PasswordHashing.java
```diff
@@ -0,0 +1,37 @@
...
+import at.favre.lib.crypto.bcrypt.BCrypt;
+
+/**
+ * Utility methods to safely store passwords in the database.
+ *
+ * <p>We rely on a third-party implementation of the bcrypt password hash function.
+ *
+ * @see <a href="https://github.com/patrickfav/bcrypt">patrickfav/bcrypt</a>
+ */
+public class PasswordHashing {
```
olim7t, line 27: A bit off-topic, but I figured we might as well show the proper way to do this.

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/user/CreateUserQueryProvider.java
```diff
@@ -0,0 +1,103 @@
...
+      if (!insertCredentialsIfNotExists(email, password, userid)) {
+        // email already exists
+        return Optional.empty();
+      }
+
+      User user = new User(userid, firstname, lastname, email, now);
+      if (insertUserIfNotExists(user)) {
+        return Optional.of(user);
+      } else {
+        // The only way the insert can fail is a UUID collision. This is highly unlikely, but handle
```
adutra: IMO there is no need for this. "Highly unlikely", according to Wikipedia, is actually virtually impossible: "for there to be a one in a billion chance of duplication, 103 trillion version 4 UUIDs must be generated." And since `UUID.random()` uses `SecureRandom` behind the scenes, I think it is safe to assume that the Java implementation has the same probability of collision.

I suggest that we either do not instruct users to care about UUID collisions, or that we switch to `timeuuid` and use `Uuids.timeBased()` to create UUIDs.

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/user/PasswordHashing.java
adutra: I like it, this is certainly going to be very useful for our users. 👍

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/user/CreateUserQueryProvider.java
```diff
@@ -0,0 +1,103 @@
...
+    this.credentialsHelper = credentialsHelper;
+
+    this.preparedInsertCredentials =
+        session.prepare(credentialsHelper.insert().ifNotExists().asCql());
+    this.preparedDeleteCredentials =
+        session.prepare(credentialsHelper.deleteByPrimaryKey().asCql());
+    this.preparedInsertUser = session.prepare(userHelper.insert().ifNotExists().asCql());
+  }
+
+  Optional<User> create(String firstname, String lastname, String email, char[] password) {
```
adutra: Why not use a logged batch to ensure atomicity?

examples/src/main/java/com/datastax/oss/driver/examples/mapper/KillrVideoMapperExample.java
```diff
@@ -0,0 +1,148 @@
...
+ * @see <a href="https://docs.datastax.com/en/developer/java-driver/latest/manual/mapper/">Java
+ *     Driver Mapper manual</a>
+ */
+public class KillrVideoMapperExample {
+
+  private static final CqlIdentifier KEYSPACE_ID = CqlIdentifier.fromCql("killrvideo");
+
+  public static void main(String[] args) {
+    try (CqlSession session = CqlSession.builder().build()) {
+      KillrVideoMapper mapper =
```
tomekl007, line 83: For other examples, we are including prepare phase, that is creating keyspace and tables in an automatic way.
For example:
```
  private static void createSchema(CqlSession session) {
    session.execute(
        "CREATE KEYSPACE IF NOT EXISTS examples "
            + "WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1}");
    session.execute(
        "CREATE TABLE IF NOT EXISTS examples.forward_paging_rest_ui("
            + "userid int, username text, "
            + "added timestamp, "
            + "videoid int, title text, "
            + "PRIMARY KEY (userid, added, videoid)"
            + ") WITH CLUSTERING ORDER BY (added DESC, videoid ASC)");
  }
```
I think we should do the same here.

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/video/LatestVideo.java
```diff
@@ -0,0 +1,63 @@
...
+import com.datastax.oss.driver.api.mapper.annotations.CqlName;
+import com.datastax.oss.driver.api.mapper.annotations.Entity;
+import com.datastax.oss.driver.api.mapper.annotations.PartitionKey;
+import java.time.Instant;
+import java.util.UUID;
+
+@Entity
+@CqlName("latest_videos")
+public class LatestVideo extends VideoBase {
+  @PartitionKey private String yyyymmdd;
```
adutra, line 28: This is semantically a CQL `date` type, why model it as a `varchar`?

examples/src/main/java/com/datastax/oss/driver/examples/mapper/KillrVideoMapperExample.java
```diff
@@ -0,0 +1,148 @@
...
+
+      // Check that associated denormalized tables have also been updated:
+      PagingIterable<UserVideo> userVideos = videoDao.getByUser(user.getUserid());
+      System.out.printf("Videos for %s %s:%n", user.getFirstname(), user.getLastname());
+      for (UserVideo userVideo : userVideos) {
+        System.out.printf("  [%s] %s%n", userVideo.getVideoid(), userVideo.getName());
+      }
+
+      String yyyymmdd =
+          DateTimeFormatter.ofPattern("yyyyMMdd").withZone(ZoneOffset.UTC).format(Instant.now());
```
tomekl007: That `Instant.now()` is a bit non-intuitive. It is also embedded within the logic of `CreateVideoQueryProvider`.
To improve design maybe we could refactor it to use a time provider? Instead of hardcoding that in the logic?

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/user/LoginQueryProvider.java
```diff
@@ -0,0 +1,72 @@
...
+  Optional<User> login(String email, char[] password) {
+    return Optional.ofNullable(session.execute(preparedSelectCredentials.bind(email)).one())
+        .flatMap(
+            credentialsRow -> {
+              String hashedPassword = credentialsRow.getString("password");
+              if (PasswordHashing.matches(password, hashedPassword)) {
+                UUID userid = credentialsRow.getUuid("userid");
+                Row userRow = session.execute(preparedSelectUser.bind(userid)).one();
+                if (userRow == null) {
+                  throw new IllegalStateException(
```
tomekl007, line 62: This method works in a functional way but we are throwing an exception from the inside of it.
Maybe we could Encapsulate user in a `Try` construct? We could use https://www.vavr.io/vavr-docs/#_try but I guess this is not worth adding an additional dependency

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/user/UserDao.java
```diff
@@ -0,0 +1,58 @@
...
+   * @return the newly created user, or {@link Optional#empty()} if the email address is already
+   *     taken.
+   */
+  @QueryProvider(
+      providerClass = CreateUserQueryProvider.class,
+      entityHelpers = {User.class, UserCredentials.class})
+  Optional<User> create(String firstName, String lastName, String email, char[] password);
+
+  /**
+   * Similar to {@link #create}, this encapsulate encryption so use a query provider.
```
adutra: Nit: "this encapsulate**s** encryption so use**s** a query provider."

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/user/CreateUserQueryProvider.java
```diff
@@ -0,0 +1,103 @@
...
+    Instant now = Instant.now();
```
tomekl007: This line should be refactored layer higher to not hard-code `now()`

examples/src/main/resources/killrvideo_schema.cql
```diff
@@ -0,0 +1,137 @@
...
+
+// Records the number of views/playbacks of a video
+CREATE TABLE IF NOT EXISTS video_playback_stats (
+    videoid uuid,
+    views counter,
+    PRIMARY KEY (videoid)
+);
+
+// Recommendations by user (powered by Spark), with the newest videos added to the site always first
+CREATE TABLE IF NOT EXISTS video_recommendations (
```
adutra, line 79: Some of these tables are not mapped in this example, should we remove their DDL declaration?

adutra (changes requested):

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/user/CreateUserQueryProvider.java
tomekl007: The fact that Optional<User>.empty() is returned when the actual user is present in the db seems a bit odd.
I wonder if there are other options or this is an idiomatic way to handle that?

examples/src/main/java/com/datastax/oss/driver/examples/mapper/KillrVideoMapperExample.java
```diff
@@ -0,0 +1,148 @@
...
+      video.setName(
+          "Getting Started with DataStax Apache Cassandra as a Service on DataStax Constellation");
+      video.setLocation("https://www.youtube.com/watch?v=68xzKpcZURA");
+      Set<String> tags = new HashSet<>();
+      tags.add("apachecassandra");
+      tags.add("nosql");
+      tags.add("hybridcloud");
+      video.setTags(tags);
+
+      videoDao.create(video);
```
tomekl007, line 122: Why the decision about having inconsistent API between `videoDao.create` and `UserDao.create`?
The first method is taking the whole Entity object whereas the second one is not taking the user as the argument but single fields of the User entity.

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/video/CreateVideoQueryProvider.java
```diff
@@ -0,0 +1,134 @@
...
+    this.latestVideoHelper = latestVideoHelper;
+    this.videoByTagHelper = videoByTagHelper;
+
+    this.preparedInsertVideo = prepareInsert(session, videoHelper);
+    this.preparedInsertUserVideo = prepareInsert(session, userVideoHelper);
+    this.preparedInsertLatestVideo = prepareInsert(session, latestVideoHelper);
+    this.preparedInsertVideoByTag = prepareInsert(session, videoByTagHelper);
+  }
+
+  void create(Video video) {
```
tomekl007, line 69: In the `CreateUserQueryProvider.create` we have logic that is able to retry in a case of a failure.
Here we are trying only once and if the batch failed the exception is just propagated. In the first case, we are guarding against `UUID` collisions but I think that in this method `video.setVideoid(UUID.randomUUID());` can also lead to a collision. Shouldn't we handle that?

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/user/CreateUserQueryProvider.java
olim7t: 👍
I wasn't sure about this (I didn't do it for videos). I'll remove the loop.

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/user/CreateUserQueryProvider.java
olim7t: A batch containing conditional updates can operate only within a single partition, we're inserting in two different tables.

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/video/LatestVideo.java
olim7t: I'm not sure, I just reused the KillrVideo data model as-is, and it's how they do it. Maybe they wanted to support legacy versions that don't have `date`.

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/video/CreateVideoQueryProvider.java
olim7t: I'm removing UUID collision handling for user creation per Alexandre's other comment.

examples/src/main/java/com/datastax/oss/driver/examples/mapper/KillrVideoMapperExample.java
olim7t: We can parse the CQL file and execute the statements I guess, I'll look into that.

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/user/CreateUserQueryProvider.java
olim7t: In a real app I would agree (in order to improve testability), but since this is just a simple example I think we can take this shortcut.

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/user/CreateUserQueryProvider.java
olim7t: The reason I wanted to return the user is because the method generates the id. But since creating an id is trivial anyway, maybe we can take it as a parameter, and the method will just return a boolean.

examples/src/main/java/com/datastax/oss/driver/examples/mapper/KillrVideoMapperExample.java
olim7t: Honestly I didn't give it much thought, I'll see if I can improve it.

examples/src/main/java/com/datastax/oss/driver/examples/mapper/KillrVideoMapperExample.java
olim7t: Yeah I admit it looks a bit weird in the middle of the client code. I'll extract it in a method and add a few explanations.

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/user/LoginQueryProvider.java
olim7t: > I guess this is not worth adding an additional dependency

Yeah I agree.

examples/src/main/resources/killrvideo_schema.cql
olim7t: I see no harm in leaving them, maybe we can expand the example later.

tomekl007 (approved):

examples/src/main/java/com/datastax/oss/driver/examples/mapper/killrvideo/user/CreateUserQueryProvider.java
```diff
@@ -0,0 +1,89 @@
...
+
+    if (!insertCredentialsIfNotExists(user.getEmail(), password, user.getUserid())) {
+      // email already exists
+      return false;
+    }
+
+    if (user.getCreatedDate() == null) {
+      user.setCreatedDate(Instant.now());
+    }
+    insertUser(user);
```
adutra: I now understand why you didn't use a batch here, but what if the user insertion fails, e.g. because of an `UnavailableException`? It would be wiser to delete the corresponding credentials.

olim7t: I've amended the code to clean up any inserted row if there is an exception.

olim7t:
Force-pushed to rebase on latest 4.x.

examples/src/main/java/com/datastax/oss/driver/examples/mapper/KillrVideoMapperExample.java
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  private static void tryLogin(UserDao userDao, String email, String password) {
+    Optional<User> maybeUser = userDao.login(email, password.toCharArray());
+    System.out.printf(
+        "Logging in with %s/%s: %s%n",
+        email, password, maybeUser.isPresent() ? "Success" : "Failure");
+  }
+
+  public static void maybeCreateSchema(CqlSession session) throws Exception {
```
adutra: Can be private

examples/src/main/java/com/datastax/oss/driver/examples/mapper/KillrVideoMapperExample.java
```diff
@@ -0,0 +1,192 @@
...
+        .filter(s -> !s.isEmpty())
+        .collect(Collectors.toList());
+  }
+
+  /**
+   * KillrVideo uses a textual timestamp to partition recent video. Build the timestamp for today to
+   * fetch our latest insertions.
+   */
+  private static String todaysTimestamp() {
+    return DateTimeFormatter.ofPattern("yyyyMMdd").withZone(ZoneOffset.UTC).format(Instant.now());
```
adutra, line 190: This is repeated in `CreateVideoQueryProvider`, can't we factor the two together?

adutra (approved):

PR #1307, opened 2019-07-24
JAVA-2358: Fix list of reserved CQL keywords

olim7t:
I've regenerated the list from the CQL spec, turns out that there were a lot of differences 😞 

I've checked each one manually:
* removed entries: I can create a table with that name (also checked a few with Cassandra 1.2, so I don't think it's a legacy thing)n
* added entries: I cannot create a table with that name.


core/src/main/java/com/datastax/oss/driver/internal/core/util/Strings.java
```diff
@@ -270,48 +269,54 @@ private Strings() {}
           "by",
           "columnfamily",
           "create",
+          "default",
```
adutra, line 272: I don't know if you used that to create this list, but there is a constant field in `org.apache.cassandra.cql3.ReservedKeywords#reservedKeywords` with the list of reserved keywords. I manually validated that our list matches theirs for the current state of the `trunk` branch.

adutra (approved):

PR #1308, opened 2019-07-25
JAVA-2365: Redeclare default constants when an enum is abstracted behind an interface

olim7t:
A few users have been tricked by this already, this would make us much more completion-friendly:
```java
BatchStatement batch = BatchStatement.newInstance(BatchType.LOGGED);
```

The downside is that it creates a circular dependency. Vanilla Java handles it well, but I think we encountered weird cases in the past (Scala?) where this would be an issue.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -4,6 +4,8 @@
 
 ### 4.2.0 (in progress)
 
+- [improvement] JAVA-2365: Redeclare default constants when an enum is abstracted behind an
+  interface
 - [improvement] JAVA-2302: Better target mapper errors and warnings for inherited methods
 - [improvement] JAVA-2336: Expose byte utility methods in the public API 
 - [improvement] JAVA-2338: Revisit toString() for data container types
--- a/core/src/main/java/com/datastax/oss/driver/api/core/ConsistencyLevel.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/ConsistencyLevel.java
@@ -26,6 +26,18 @@ import edu.umd.cs.findbugs.annotations.NonNull;
  */
 public interface ConsistencyLevel {
 
+  ConsistencyLevel ANY = DefaultConsistencyLevel.ANY;
+  ConsistencyLevel ONE = DefaultConsistencyLevel.ONE;
+  ConsistencyLevel TWO = DefaultConsistencyLevel.TWO;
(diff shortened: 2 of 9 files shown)


olim7t:
> Scala?

[JAVA-1252](https://datastax-oss.atlassian.net/browse/JAVA-1252)
I'll set up a test sbt project.

tomekl007 (commented):
Are you sure that this is a good direction(from the maintenance perspective)? 
Does the constants that we are referencing change often or not? If not I am +1 on that but if they are changed then it is a bit tricky because when adding a new constant to the target enum we need to remember to include them in the interface (because I think it won't be caught by compiler, and I am not sure if that will be caught by our tests)

adutra (approved):
Looks good, but I am somehow a bit uncomfortable here, so an extra check with at least Scala would be nice.

adutra:
> A few users have been tricked by this already

I also saw users tricked by the similarity in class names between e.g. `com.datastax.oss.driver.api.core.cql.BatchType` vs `com.datastax.oss.protocol.internal.ProtocolConstants.BatchType`.

olim7t:
> I also saw users tricked by the similarity in class names between e.g. `com.datastax.oss.driver.api.core.cql.BatchType` vs `com.datastax.oss.protocol.internal.ProtocolConstants.BatchType`.

That's the issue I was initially trying to solve.

olim7t:
Sample sbt project [here](https://gist.github.com/olim7t/1821be4861f3689e1e5186d6c2946d15).

Verdict:
- I can reproduce JAVA-1252 with Scala 2.11.12, but it's gone in 2.13.0.
- this PR doesn't cause the same issue, neither in 2.11.12 nor in 2.13.0.

olim7t:
> Are you sure that this is a good direction(from the maintenance perspective)?

The main goal is to improve completion-friendliness. Currently if users type `ProtocolVersion.` they can't autocomplete, the only thing that's hinting at `DefaultProtocolVersion` are the javadocs, and we know how much that's worth 😿 
For `BatchType` it's worse because there's a eponymous internal class that gets in the way.

> Does the constants that we are referencing change often or not?

CL, BatchType and WriteType: very rarely.
ProtocolVersion: every couple of years or so.

> when adding a new constant to the target enum we need to remember to include them in the interface

Yes, that's the main drawback. What I can do is add a comment in the enum as a reminder.

PR #1310, opened 2019-07-30
JAVA-2371: Handle null elements in collections on the decode path

olim7t:


jtgrabowski (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/MapCodec.java
```diff
@@ -134,18 +134,28 @@ public ByteBuffer encode(
       int size = input.getInt();
       Map<KeyT, ValueT> result = Maps.newLinkedHashMapWithExpectedSize(size);
       for (int i = 0; i < size; i++) {
+        KeyT key;
```
tomekl007, line 137: If we are handling Map shouldn't we handle that for Tuple and Udt as well?

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/MapCodecTest.java
```diff
@@ -118,6 +118,16 @@ public void should_decode_non_empty_map() {
...
   }
 
+  @Test
+  public void should_decode_map_with_null_elements() {
+    when(keyCodec.decode(Bytes.fromHexString("0x10"), ProtocolVersion.DEFAULT)).thenReturn("a");
+    when(valueCodec.decode(Bytes.fromHexString("0x0002"), ProtocolVersion.DEFAULT)).thenReturn(2);
+    assertThat(decode("0x" + "00000002" + "0000000110" + "FFFFFFFF" + "FFFFFFFF" + "000000020002"))
+        .containsOnlyKeys("a", null)
+        .containsEntry("a", null)
+        .containsEntry(null, 2);
```
tomekl007, line 128: I think that we are missing a test case for key and value null for the same entry but I don't know if such a test gives much value.

olim7t: It wouldn't increase coverage so I would say no.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/ListCodec.java
```diff
@@ -115,11 +115,19 @@ public ByteBuffer encode(
...
       List<ElementT> result = new ArrayList<>(size);
       for (int i = 0; i < size; i++) {
+        ElementT element;
         int elementSize = input.getInt();
-        ByteBuffer encodedElement = input.slice();
-        encodedElement.limit(elementSize);
-        input.position(input.position() + elementSize);
-        result.add(elementCodec.decode(encodedElement, protocolVersion));
+        // Allow null elements on the decode path, because Cassandra might return such collections
+        // for some computed values in the future -- e.g. SELECT ttl(some_collection)
```
adutra, line 121: I'm probably a bit paranoid but should we include the example of `ttl(collection)`? This example is DSE-specific.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/ListCodec.java
```diff
@@ -115,11 +115,19 @@ public ByteBuffer encode(
...
+        if (elementSize < 0) {
+          element = null;
+        } else {
+          ByteBuffer encodedElement = input.slice();
+          encodedElement.limit(elementSize);
+          input.position(input.position() + elementSize);
```
adutra: Nit: it would be a bit easier to understand this code if you invert these two lines:

```
element = elementCodec.decode(encodedElement, protocolVersion);
input.position(input.position() + elementSize);
```

(I initially thought that updating the position of `input` was required before the element can be decoded.)

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/ListCodecTest.java
```diff
@@ -93,6 +93,12 @@ public void should_decode_non_empty_list() {
         .containsExactly(1, 2, 3);
   }
 
+  @Test
+  public void should_decode_list_with_null_elements() {
+    when(elementCodec.decode(Bytes.fromHexString("0x0002"), ProtocolVersion.DEFAULT)).thenReturn(2);
+    assertThat(decode("0x" + "00000002" + "FFFFFFFF" + "000000020002")).containsExactly(null, 2);
```
adutra: This is very cryptic, how about:

```
assertThat(
        decode(
            "0x"
                + "00000002" // number of elements
                + "FFFFFFFF" // size of element 1 (null)
                + "00000002" // size of element 2 (2 bytes)
                + "0002" // element 2
            ))
    .containsExactly(null, 2);
```

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/MapCodec.java
olim7t: `TupleCodec` and `UdtCodec` already handle null values.

UDT field names don't appear in the encoded representation. And there's no way they can ever be null anyway.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/ListCodec.java
olim7t: [CASSANDRA-8877](https://issues.apache.org/jira/browse/CASSANDRA-8877) will likely add that feature.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/ListCodec.java
olim7t: ok I'm doing it in a separate commit

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/ListCodecTest.java
olim7t: There's a commented example 30 lines above. When you have the whole file open in an IDE it's pretty visible, I think it's fine not to repeat it in every test.

tomekl007: I agree with Alex on that, I had the same impression that it was very cryptic

adutra: Well it's not exactly the same example, here we have a `"FFFFFFFF"` block that seems strange at first glance, and `"000000020002"` is also strange since it contains both the size of element 2 and element 2 itself.

PR #1311, opened 2019-08-01
JAVA-2375: Use per-request keyspace when repreparing on the fly

olim7t:


emerkle826 (approved):
:+1:

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cql/PerRequestKeyspaceIT.java
```diff
@@ -203,4 +204,33 @@ public void should_prepare_statement_with_keyspace() {
...
+                  .setKeyspace(sessionRule.keyspace()));
+
+      // Drop and re-create the table to invalidate the prepared statement server side
+      executeDdl(session, String.format("DROP TABLE %s.bar", sessionRule.keyspace()));
+      executeDdl(
+          session,
+          String.format("CREATE TABLE %s.bar (k int primary key)", sessionRule.keyspace()));
+
+      // This will re-prepare on the fly
+      session.execute(pst.bind(0));
```
tomekl007, line 231: maybe we can check the cache to see that it was re-prepared and populated?

olim7t: I assume you mean the server-side cache. I can check `system.prepared_statements`, it will always be present since the test targets C* 4+.

PR #1313, opened 2019-08-06
JAVA-2379: Fail fast if prepared id doesn't match when repreparing on the fly

olim7t:


core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
```diff
@@ -645,30 +645,33 @@ public void onResponse(Frame responseFrame) {
...
                 throttler,
                 sessionMetricUpdater,
-                logPrefix,
-                "Reprepare " + reprepareMessage.toString());
+                logPrefix);
         reprepareHandler
             .start()
             .handle(
-                (result, exception) -> {
+                (repreparedId, exception) -> {
```
tomekl007, line 674: It is not possible that `exception != null && reprepareId != null`, right?
If it will be then I think we should log this https://github.com/datastax/java-driver/pull/1313/files#diff-3708101a1afe48621cd3e28722f732bdR703 in the exception handling as well

olim7t: It's not, because we never complete the future with null for PREPARED responses in `AdminRequestHandler.onResponse`.

PR #1316, opened 2019-08-12
JAVA-2390: Add methods to set the SSL engine factory programmatically

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/ssl/ProgrammaticSslEngineFactory.java
```diff
@@ -0,0 +1,67 @@
...
+ * @see SessionBuilder#withSslEngineFactory(SslEngineFactory)
+ * @see SessionBuilder#withSslContext(SSLContext)
+ */
+public class ProgrammaticSslEngineFactory implements SslEngineFactory {
+
+  protected final SSLContext sslContext;
+
+  public ProgrammaticSslEngineFactory(SSLContext sslContext) {
+    this.sslContext = sslContext;
+  }
```
GregBestland, line 46: This is perfect. You don't have wrap anything just hand over the context. +1

integration-tests/src/test/java/com/datastax/oss/driver/core/ssl/ProgrammaticSslIT.java
```diff
@@ -0,0 +1,94 @@
+/*
```
GregBestland, line 1: *NIT* It seems like we are inconsistent with our naming, other classnames in our project use SSL,  instead of Ssl we should try to bring the naming inline so we are consistent.

GregBestland (approved):
Other than one nit about naming, this LGTM

integration-tests/src/test/java/com/datastax/oss/driver/core/ssl/ProgrammaticSslIT.java
olim7t: Apart from `SSLContext` and `SSLEngine` (which are not ours, they come from the JDK), I think we used `Ssl` everywhere.

PR #1318, opened 2019-08-21
Added Failover Policy

tdl-jturner:
Motivation:

Failover is a critical aspect of the database layer that is missing in the current
implementation of the driver. While a cloud model does lend itself to marking down
a full site, other deployment models do not. The goal of the failover policy is to
build a simple tie in to allow for failvoer decisions to be made based on caught
exceptions as well as statement attributes.

The included DefaultFailoverPolicy allows failover to a second policy which could
then be a new dc, consistency, or many other variations.

Modifications:

Added new failover parameters, FailoverPolicy interface, and DefaultFailoverPolicy
objects. This was integrated to the CqlRequestHandler#handle method by adding a
new exceptionally stage which evaluates if a failover should occur then triggers
a new CqlRequestHandler with the modified statement.

Result:

New FailoverPolicy is available to failover statements in the event of a failure.

### DIFF
--- a/core/revapi.json
+++ b/core/revapi.json
@@ -4809,6 +4809,10 @@
         "code": "java.method.addedToInterface",
         "new": "method java.lang.String com.datastax.oss.driver.api.core.data.UdtValue::toString()",
         "justification": "False positive -- all objects implicitly have toString()"
+      },{
+        "code": "java.method.addedToInterface",
+        "new": "method java.util.Map<java.lang.String, com.datastax.oss.driver.api.core.failover.FailoverPolicy> com.datastax.oss.driver.api.core.context.DriverContext::getFailoverPolicies()",
+        "justification": "Added FailoverPolicy"
       }
     ]
   }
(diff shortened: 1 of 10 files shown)


olim7t:
Hi,

Thanks for taking the time to provide a full prototype.

Adding a new policy is not a decision to be taken lightly, as it impacts the complexity and future maintenance of the driver. Here are a few thoughts about your proposed approach:
* every request still hits the primary target every time, even if it ends up failing over. If we detect that a DC went down, wouldn't it be more efficient to modify the statements at the source (in application code), and target the alternate DC directly, instead of hitting an error and failing over every time?
* one particular example is `DriverTimeoutException`. If each request potentially executes twice, the driver does not honor `basic.request.timeout` anymore, the SLA is now twice that. It's not a big deal, but an example of the subtle impacts that users of the policy would need to understand.
* there's also a bit of overlap with existing policies. For example, if what you're after is DC-level failover like in driver 3, this could be achieved with a custom load balancing policy.

For these reasons, I don't see a strong case to add this a an out-of-the-box feature.

But note that it's still pretty easy to add from your application (or application framework) code. At the simplest, you can wrap the session and intercept `execute` calls. Or you could:
* write a custom request processor that wraps `CqlRequestAsyncProcessor` to add the retry logic
* create a custom context subclass to plug in your custom processor
* create an alternate session builder that creates a session with your custom context

You have most of this code in here already.

tdl-jturner:
Thank you for the feedback and consideration. I understand your perspective and desire to not have this as baked in functionality. This was originally written for another project wrapping the session as suggested which functions but is not nearly as integrated. This seemed an opportunity to offer the feature back to the community and thus ported and submitted it. I'll go ahead and close the PR.

PR #1322, opened 2019-09-04
JAVA-2419: Allow registration of user codecs at runtime

olim7t:


manual/core/custom_codecs/README.md
```diff
@@ -87,6 +87,18 @@ CqlSession session = CqlSession.builder()
...
+You may also add codecs to an existing session at runtime:
+
+```java
+// The cast is required for backward compatibility reasons (registry mutability was introduced in
+// 4.3.0). It is safe as long as you didn't hack the driver internals to plug a custom registry
+// implementation.
+MutableCodecRegistry registry =
+    (MutableCodecRegistry) session.getContext().getCodecRegistry();
+
+registry.register(new CqlIntToStringCodec());
```
olim7t, line 119: Canonical way to use the new feature from client code.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/registry/MutableCodecRegistry.java
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Adds the given codec to the registry.
+   *
+   * <p>This method will log a warning and ignore the codec if it collides with one already present
+   * in the registry. Note that this check is not done in a completely thread-safe manner; codecs
+   * should typically be registered at application startup, not in a highly concurrent context (if a
+   * race condition occurs, the worst possible outcome is that no warning gets logged, and the codec
+   * gets ignored silently).
+   */
+  void register(TypeCodec<?> codec);
```
olim7t, line 43: Same methods as in 3.x.
I made them void to emphasize the fact that they mutate the existing registry (since that's the only way you can operate on the instance stored in the context).

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -68,19 +69,71 @@
...
+   *     #register} methods.
+   */
+  @Deprecated
+  protected CachingCodecRegistry(
+      @NonNull String logPrefix,
+      @NonNull TypeCodec<?>[] primitiveCodecs,
+      @NonNull TypeCodec<?>[] userCodecs) {
+    this(logPrefix, primitiveCodecs);
+    register(userCodecs);
+  }
```
olim7t, line 98: Deprecating this constructor is the best way I've found to address the fact that it calls an overridable method. 3.x didn't have a way to pass user codecs to the constructor so I don't think it's a big deal to drop that.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/DefaultCodecRegistry.java
```diff
@@ -48,7 +48,24 @@
...
+    this(logPrefix, CodecRegistryConstants.PRIMITIVE_CODECS);
+  }
+
+  /**
+   * Creates a new instance that accepts user codecs, with the given built-in codecs and the default
+   * cache behavior.
+   */
+  public DefaultCodecRegistry(@NonNull String logPrefix, @NonNull TypeCodec<?>[] primitiveCodecs) {
+    this(logPrefix, 0, null, 0, null, primitiveCodecs);
+  }
```
olim7t, line 64: Unlike `CachingCodecRegistry`, I didn't preserve backward compatibility for this class's constructors. It's much less likely that someone extended it.

One catch is that the last argument of `DefaultCodecRegistry(String, TypeCodec<?>[])` now corresponds to primitive codecs (vs. user codecs in 4.2.0). But if someone runs into that things will blow out pretty fast, there's no risk that it goes undetected.

core/src/main/java/com/datastax/oss/driver/api/core/detach/AttachmentPoint.java
```diff
@@ -39,6 +40,12 @@ public CodecRegistry getCodecRegistry() {
   @NonNull
   ProtocolVersion getProtocolVersion();
 
+  /**
+   * Note that the default registry implementation returned by the driver also implements {@link
+   * MutableCodecRegistry}, which allows you to register new codecs at runtime. You can safely cast
+   * the result of this method (as long as you didn't hack the driver internals to plug a custom
```
tomekl007:
```
as long as you didn't hack the driver internals to plug a custom
   * registry implementation
```
maybe we can give an example of what exactly we want to prevent? Users may do this without being aware that they are doing this "hack"

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/registry/MutableCodecRegistry.java
```diff
@@ -0,0 +1,58 @@
...
+ * @since 4.3.0
+ */
+public interface MutableCodecRegistry extends CodecRegistry {
+
+  /**
+   * Adds the given codec to the registry.
+   *
+   * <p>This method will log a warning and ignore the codec if it collides with one already present
+   * in the registry. Note that this check is not done in a completely thread-safe manner; codecs
+   * should typically be registered at application startup, not in a highly concurrent context (if a
```
tomekl007: ` not in a highly concurrent context` - I will feel this requirement is a bit vague. What does it mean not highly concurrent? Is it 2 threads, 10?

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -68,19 +69,71 @@
...
+
+  @Override
+  public void register(TypeCodec<?> newCodec) {
+    for (TypeCodec<?> primitiveCodec : primitiveCodecs) {
+      if (primitiveCodec.accepts(newCodec.getCqlType())
```
tomekl007: I think that this check can be refactored to a method and reused by this line and https://github.com/datastax/java-driver/pull/1322/files#diff-34d4c85a5bfd4f5ccb02c01deb989c99R111

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -68,19 +69,71 @@
...
+            userCodec);
+        return;
+      }
+    }
+    // Technically this would cover the two previous cases as well, but we want precise messages.
+    try {
+      TypeCodec<?> cachedCodec =
+          getCachedCodec(newCodec.getCqlType(), newCodec.getJavaType(), false);
+      LOG.warn(
+          "[{}] Ignoring codec {} because it collides with previously generated codec {}",
```
tomekl007: what is a difference between `previously generated codec` and `previously registered codec` (from the client perspective)?

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -68,19 +69,71 @@
...
+          logPrefix,
+          newCodec,
+          cachedCodec);
+      return;
+    } catch (CodecNotFoundException ignored) {
+      // Catching the exception is ugly, but it avoids breaking the internal API (e.g. by adding a
```
tomekl007: maybe we should log a warning?

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/registry/CodecRegistry.java
```diff
@@ -40,13 +40,24 @@
...
   /**
    * An immutable instance, that only handles built-in driver types (that is, primitive types, and
    * collections, tuples, and user defined types thereof).
    */
-  CodecRegistry DEFAULT = new DefaultCodecRegistry("default");
+  CodecRegistry DEFAULT =
+      new DefaultCodecRegistry("default") {
+        @Override
+        public void register(TypeCodec<?> newCodec) {
+          throw new IllegalStateException("CodecRegistry.DEFAULT is immutable");
```
adutra: `java.lang.UnsupportedOperationException` would be a better fit here imho.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/registry/MutableCodecRegistry.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * A codec registry that can be extended with new user codecs at runtime.
+ *
+ * <p>This interface only exists to preserve backward compatibility. In practice, the default {@link
+ * CodecRegistry} implementation returned by the driver implements this interface, so it can safely
+ * be cast.
+ *
+ * <p>However {@link CodecRegistry#DEFAULT} is immutable. It implements this interface, but {@link
+ * #register(TypeCodec)} throws an {@link IllegalStateException}.
```
adutra: If you change the thrown exception as I suggested don't forget to change this line as well.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/registry/MutableCodecRegistry.java
adutra: Shouldn't we simply add a `synchronized` block to this method's implementation? This way there would be no possible race condition and, as you mentioned, this method is not meant to be called on the application's hot path, so it shouldn't be too much of a penalty to have a synchronized section here.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
adutra: It should be possible to avoid the deprecation of this constructor:

```
 protected CachingCodecRegistry(
    @NonNull String logPrefix,
    @NonNull TypeCodec<?>[] primitiveCodecs,
    @NonNull TypeCodec<?>[] userCodecs) {
  this.logPrefix = logPrefix;
  this.primitiveCodecs = primitiveCodecs;
  this.primitiveCodecsByCode = sortByProtocolCode(primitiveCodecs);
  for (TypeCodec<?> userCodec : userCodecs) {
    registerInternal(userCodec);
  }
}

@Override
public void register(TypeCodec<?> newCodec) {
  registerInternal(newCodec);
}

private void registerInternal(TypeCodec<?> newCodec) {
  ...
}
```

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
adutra: Generated means that the registry created the codec on the fly, e.g. a codec for `List<Foo>` can be automatically generated if the user registers a codec for `Foo`.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
adutra: It would be overkill imo because we are only testing if such a codec has been generated; the nominal case would be that such a code should not exist, so the exception is actually the expected outcome.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/DefaultCodecRegistry.java
```diff
@@ -61,10 +78,9 @@ public DefaultCodecRegistry(
       @Nullable BiFunction<CacheKey, TypeCodec<?>, Integer> cacheWeigher,
       int maximumCacheWeight,
       @Nullable BiConsumer<CacheKey, TypeCodec<?>> cacheRemovalListener,
-      @NonNull TypeCodec<?>[] primitiveCodecs,
-      @NonNull TypeCodec<?>[] userCodecs) {
+      @NonNull TypeCodec<?>[] primitiveCodecs) {
```
adutra: You could use varargs here.

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/registry/CodecRegistry.java
olim7t: You're right, I will change that.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
olim7t: I thought of that, but then if you have a subclass that needs to do some additional processing when you register a codec, it can only do so in the overridden method, not change how the constructor handles registration.

core/src/main/java/com/datastax/oss/driver/api/core/detach/AttachmentPoint.java
olim7t: Good point, I'll be more specific like "extend the driver context to override buildCodecRegistry".

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
olim7t: 👍 extracted `collides(TypeCodec, TypeCodec)`

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/registry/MutableCodecRegistry.java
olim7t: OK, adding synchronization.

adutra (approved):

PR #1323, opened 2019-09-05
JAVA-2196: Convert TableOptionsIT into a unit test

olim7t:
Table options vary across server versions, it's hard to write robust
integration tests. TableOptionsIT only covered two basic text options
added in Cassandra 4 (JAVA-2090), this can be addressed equally well by
a unit test.


core/src/test/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/TableParserTest.java
```diff
@@ -114,6 +116,24 @@ public void should_parse_modern_tables() {
         .containsEntry("rows_per_partition", "NONE");
   }
 
+  /** Covers two additional Cassandra 4.0 options added in JAVA-2090. */
+  @Test
```
tomekl007, line 120: My one concern with this test is that previously it was explicitly stated that we are verifying against Cassandra 4.0 (` @CassandraRequirement(min = "4.0"`).
Right now I feel that this is implicitly tested and the only way it manifests is in a variable name - `tableRow40`. 
Are we sure that this test gives us the same coverage as the previous IT test?

adutra (approved):

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/TableParserTest.java
olim7t: `DescribeIT` already has some integration-level coverage, so we should update it once Cassandra 4 is stable (there is a ticket for that: [JAVA-2184](https://datastax-oss.atlassian.net/browse/JAVA-2184)).

PR #1326, opened 2019-09-13
JAVA-2436: Extract simplified parse method in DataTypeCqlNameParser

olim7t:
/cc @RussellSpitzer 

The new method takes `AttachmentPoint`, which is just a wrapper around protocolVersion + codecRegistry.

I wonder if it wouldn't be simpler to duplicate the method in Cassandra. I'm not happy about Cassandra core depending on a low-level internal method (or depending on the driver at all for that matter -- speaking strictly of the core, I'm ok with tools like stress using it).

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/DataTypeClassNameParser.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/DataTypeClassNameParser.java
@@ -17,6 +17,7 @@ package com.datastax.oss.driver.internal.core.metadata.schema.parsing;
 
 import com.datastax.oss.driver.api.core.CqlIdentifier;
 import com.datastax.oss.driver.api.core.DefaultProtocolVersion;
+import com.datastax.oss.driver.api.core.detach.AttachmentPoint;
 import com.datastax.oss.driver.api.core.type.DataType;
 import com.datastax.oss.driver.api.core.type.DataTypes;
 import com.datastax.oss.driver.api.core.type.UserDefinedType;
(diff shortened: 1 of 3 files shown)


RussellSpitzer:
What is the Attachement point that still needs to get passed?

RussellSpitzer:
I agree with removing it, but the current usage is core, stress and sqltable loader, I can add a ticket to have it moved later but then there will still be two versions inside dse-db. The copied version for server, and the javadriver version for the tools.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/DataTypeClassNameParser.java
```diff
@@ -60,6 +61,30 @@ public DataType parse(
...
+    return parse(toParse, userTypes, context, context.getSessionName());
+  }
+
+  /**
+   * Simplified parse method for external use.
+   *
+   * <p>Note that this method is intended for use in Cassandra's UDF implementation (the current
+   * version uses the similar method from driver 3).
+   */
+  public DataType parse(String toParse, AttachmentPoint attachmentPoint) {
```
tomekl007, line 75: is the Cassandra-core implementing `AttachmentPoint` and provides own DriverContext?
Maybe we can create a simple unit test for that public method?

RussellSpitzer: We'll have to make one, it will just hold the DSE Registry and the most recent Driver version

olim7t:
Regarding the attachment point:
```java
public interface AttachmentPoint {
  @NonNull
  ProtocolVersion getProtocolVersion();

  @NonNull
  CodecRegistry getCodecRegistry();
}
```
There is no standalone implementation but it's easy to write your own. I think in your case you always use the default registry, so you can hard-code `CodecRegistry.DEFAULT`.

PR #1328, opened 2019-09-16
JAVA-2430: Use variable metadata to infer the routing keyspace on bound statements

olim7t:
The previous version was erroneously using the result metadata, which is
empty for non-conditional mutations.


emerkle826 (approved):
:+1:

adutra (approved):

PR #1329, opened 2019-09-16
JAVA-2443: Compute prepared statement PK indices for protocol v3

olim7t:
This is covered by the test in #1328 when it runs in the Cassandra 2.1 suite. Although we might want to cover more complex scenarios.

### DIFF
--- a/core/src/test/java/com/datastax/oss/driver/internal/core/cql/ConversionsTest.java
+++ b/core/src/test/java/com/datastax/oss/driver/internal/core/cql/ConversionsTest.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.internal.core.cql;
+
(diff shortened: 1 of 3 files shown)


core/src/test/java/com/datastax/oss/driver/internal/core/cql/ConversionsTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+import com.datastax.oss.driver.api.core.CqlIdentifier;
+import com.datastax.oss.driver.api.core.cql.ColumnDefinition;
+import com.datastax.oss.driver.api.core.cql.ColumnDefinitions;
+import com.datastax.oss.driver.api.core.metadata.schema.ColumnMetadata;
+import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
+import java.util.List;
+import org.junit.Test;
+
+public class ConversionsTest {
```
olim7t, line 30: I added a couple of unit tests to cover all cases: multi-column partition keys and not all PK columns bound.

No changes to production code.

core/src/test/java/com/datastax/oss/driver/internal/core/cql/ConversionsTest.java
```diff
@@ -0,0 +1,88 @@
...
+    assertThat(
+            Conversions.findIndices(
+                partitionKey("pk1", "pk2", "pk3"),
+                variables("c1", "pk2", "pk3", "c2", "pk1", "c3")))
+        .containsExactly(4, 1, 2);
+  }
+
+  @Test
+  public void should_use_first_pk_index_if_bound_multiple_times() {
+    assertThat(Conversions.findIndices(partitionKey("pk"), variables("pk", "pk")))
```
tomekl007, line 47: Is it possible to have in cql:
`Conversions.findIndices(partitionKey("pk", "pk"), variables("pk")))` 
(two partition key that has the same name)?

olim7t: No, the first parameter comes from the schema metadata, and a CQL partition key cannot contain the same column twice.

PR #1330, opened 2019-09-19
JAVA-2402

absurdfarce:


core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
```diff
@@ -0,0 +1,77 @@
+/*
```
absurdfarce, line 1: A unit test for a property setter on a builder is (arguably) overkill... especially since there was no existing test for other builder functionality.  That said, I had a pseudo-integration test for confirming that everything worked as expected.. so switching that over to a unit test was pretty trivial.

I'm good with scrapping this if it's judged to be unnecessary, but it shouldn't be... harmful really.

core/src/main/java/com/datastax/oss/driver/api/core/cql/StatementBuilder.java
```diff
@@ -157,7 +157,13 @@ public SelfT setIdempotence(@Nullable Boolean idempotent) {
   /** @see Statement#setTracing(boolean) */
   @NonNull
   public SelfT setTracing() {
-    this.tracing = true;
+    return setTracing(true);
```
olim7t, line 165: Maybe we should add a short explanation in the javadoc of this method, along the lines of: "this method exists for historical reasons, it is just a shortcut for setTracing(true)".

core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+import org.junit.Test;
+
+public class StatementBuilderTest {
+
+  private static class NullStatementBuilder
+      extends StatementBuilder<NullStatementBuilder, SimpleStatement> {
```
olim7t, line 25: I don't mind creating a class, but why not just use `SimpleStatementBuilder`? The only thing it needs is a query string.

core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
```diff
@@ -0,0 +1,77 @@
...
+    public SimpleStatement build() {
+      return null;
+    }
+  }
+
+  @Test
+  public void should_handle_set_tracing_without_args() {
+
+    NullStatementBuilder builder = new NullStatementBuilder();
+    assertFalse(builder.tracing);
```
olim7t: Nit: we prefer AssertJ for assertions.
```
import static com.datastax.oss.driver.Assertions.assertThat;
...
assertThat(builder.tracing).isFalse()
```
Not a huge difference for a simple boolean, but for more complex types like collections it's way more expressive.

olim7t (changes requested):
👍 looking good

Can you add an entry to `changelog/README.md`? Also we generally include the ticket description in the commit message, so if you plan to merge this with the "rebase and merge" option, please update the title of the pull request.

core/src/main/java/com/datastax/oss/driver/api/core/cql/StatementBuilder.java
absurdfarce: Agreed, this seems reasonable... will change.

core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
absurdfarce: The functionality in question is defined in the abstract superclass so I wanted a way to get at it as cleanly as possible.  While admittedly quite unlikely there's always the possibility that a subclass could modify the behaviour... so I stayed away from SimpleStatementBuilder.

core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
absurdfarce: This makes sense.  I saw the use of AssertJ elsewhere but I went to the JUnit ones out of muscle memory.  There's no reason to make a special case for this one test, though, so I'll update to use AssertJ.

absurdfarce:
@olim7t My thinking was to merge this via a squash merge, so in that case I can add the ticket number + description as part of the squashed commit.  That should address your concern about the title of this PR... unless you guys don't use squash merges in general, in which case I'll need to come up with another plan. :)

olim7t:
As long as this ends up as a single commit on 4.x with the subject `JAVA-2402: Add setTracing(boolean) to StatementBuilder`, any method works for me :)

olim7t (approved):

olim7t:
Nit: next time, please put the ticket description in the first line of the commit message, so that it shows in github default view / `git log --pretty=oneline`, etc.

PR #1332, opened 2019-09-23
JAVA-2454: Handle empty CQL type while parsing schema

olim7t:


adutra (approved):
Are we confident that this change doesn't require a dedicated unit test?

olim7t:
> Are we confident that this change doesn't require a dedicated unit test?

I validated the fix against the cluster where the issue was initially detected.

Beyond that, I think a unit test would not provide a lot of value, because it would be a very literal check: mocking an `empty` column type and verify that the column does not exist. A better approach would be an integration test that creates and migrates an actual Thrift schema, but that sounds a bit complicated.

tomekl007 (approved):

tomekl007:
> > Are we confident that this change doesn't require a dedicated unit test?
> 
> I validated the fix against the cluster where the issue was initially detected.
> 
> Beyond that, I think a unit test would not provide a lot of value, because it would be a very literal check: mocking an `empty` column type and verify that the column does not exist. A better approach would be an integration test that creates and migrates an actual Thrift schema, but that sounds a bit complicated.

Don't you think that even the unit test could serve as good documentation of the behavior and state clearly that this should be handled by the empty type? Right now some refactoring in the future could change that and we will not get any test failures

olim7t:
OK I'll create a follow-up ticket to cover this better. I think a good approach would be to migrate a Thrift schema manually, and emulate the resulting system tables contents in a unit test.

PR #1333, opened 2019-09-23
JAVA-2455: Improve logging of schema refresh errors

olim7t:
Motivation:

While debugging JAVA-2454, we noticed that the error wasn't surfaced in
the logs as expected. MetadataManager.startSchemaRequest has a listener
for that purpose, but it's on the wrong future; parsing errors will
surface on currentSchemaRefresh.

Modifications:

Don't try to log errors in MetadataManager. Instead, handle it on all
call sites of MetadataManager.refreshSchema, *if* the future is not
returned to the client.

Result:

If a user-initiated refresh fails, they will get back a failed future.
If an internal refresh fails, it will be logged.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
@@ -226,7 +226,20 @@ public class ControlConnection implements EventCallback, AsyncAutoCloseable {
 
   private void processSchemaChange(Event event) {
     SchemaChangeEvent sce = (SchemaChangeEvent) event;
-    context.getMetadataManager().refreshSchema(sce.keyspace, false, false);
+    context
+        .getMetadataManager()
+        .refreshSchema(sce.keyspace, false, false)
+        .whenComplete(
+            (metadata, error) -> {
+              if (error != null) {
+                Loggers.warnWithException(
+                    LOG,
+                    "[{}] Unexpected error while refreshing schema for a SCHEMA_CHANGE event, "
+                        + "keeping previous version",
+                    logPrefix,
+                    error);
+              }
+            });
   }
 
   private class SingleThreaded {
@@ -467,7 +480,20 @@ public class ControlConnection implements EventCallback, AsyncAutoCloseable {
                     // first successful refresh; make sure the LBP gets initialized (this is a no-op
                     // if it was initialized already).
                     context.getLoadBalancingPolicyWrapper().init();
-                    context.getMetadataManager().refreshSchema(null, false, true);
+                    context
+                        .getMetadataManager()
+                        .refreshSchema(null, false, true)
+                        .whenComplete(
+                            (metadata, schemaError) -> {
+                              if (schemaError != null) {
+                                Loggers.warnWithException(
+                                    LOG,
+                                    "[{}] Unexpected error while refreshing schema after a "
+                                        + "successful reconnection, keeping previous version",
+                                    logPrefix,
+                                    schemaError);
+                              }
+                            });
                   } catch (Throwable t) {
                     Loggers.warnWithException(
                         LOG, "[{}] Unexpected error on control connection reconnect", logPrefix, t);
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
@@ -106,7 +106,18 @@ public class MetadataManager implements AsyncAutoCloseable {
             || !keyspacesBefore.equals(refreshedKeyspaces)
             || (!tokenMapEnabledBefore && tokenMapEnabled))
         && isSchemaEnabled()) {
-      refreshSchema(null, false, true);
+      refreshSchema(null, false, true)
+          .whenComplete(
+              (metadata, error) -> {
+                if (error != null) {
+                  Loggers.warnWithException(
+                      LOG,
+                      "[{}] Unexpected error while refreshing schema after it was re-enabled "
+                          + "in the configuration, keeping the previous version",
+                      logPrefix,
+                      error);
+                }
+              });
     }
   }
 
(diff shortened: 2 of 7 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
```diff
@@ -404,11 +415,7 @@ private void startSchemaRequest(CompletableFuture<Metadata> future) {
             .whenComplete(
                 (v, error) -> {
                   if (error != null) {
-                    Loggers.warnWithException(
-                        LOG,
-                        "[{}] Unexpected error while refreshing schema, skipping",
-                        logPrefix,
-                        error);
+                    currentSchemaRefresh.completeExceptionally(error);
```
olim7t, line 418: It's unlikely that an error will surface here, because the previous stages should have caught it. But it doesn't hurt to keep the check.

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
```diff
@@ -226,7 +226,20 @@ private void processStatusChange(Event event) {
 
   private void processSchemaChange(Event event) {
     SchemaChangeEvent sce = (SchemaChangeEvent) event;
-    context.getMetadataManager().refreshSchema(sce.keyspace, false, false);
+    context
+        .getMetadataManager()
+        .refreshSchema(sce.keyspace, false, false)
+        .whenComplete(
```
adutra, line 232: Wouldn't be simpler to use `exceptionally`?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
```diff
@@ -106,7 +106,18 @@ private void onConfigChanged(@SuppressWarnings("unused") ConfigChangeEvent event
...
             || (!tokenMapEnabledBefore && tokenMapEnabled))
         && isSchemaEnabled()) {
-      refreshSchema(null, false, true);
+      refreshSchema(null, false, true)
+          .whenComplete(
+              (metadata, error) -> {
+                if (error != null) {
+                  Loggers.warnWithException(
+                      LOG,
+                      "[{}] Unexpected error while refreshing schema after it was re-enabled "
```
adutra, line 115: Nit: all call sites have `Unexpected error while...` except `CqlRequestHandler` which has `Error while...`.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
adutra: Out of curiosity, what is the rationale for moving the log messages outside `MetadataManager`?

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
olim7t: It's because we don't log in all cases. In my opinion, an error should either be logged or surfaced through the user API, but not both.
For example:
* if the refresh is triggered by an incoming `SCHEMA_CHANGE` event on the control connection, there is no way to surface it so we log.
* but if the user called `session.refreshSchema`, we can rethrow the error and let them handle it. I don't like to log in those cases, because it will show an error in the logs even if the application already handled it.

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
olim7t: It's equivalent. But `exceptionally` creates another future with a fallback value, which is not needed here.
I did use it in `CqlRequestHandler` though, because I needed a future.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
olim7t: 👍 I'll change this message.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
```diff
@@ -106,7 +106,18 @@ private void onConfigChanged(@SuppressWarnings("unused") ConfigChangeEvent event
             || !keyspacesBefore.equals(refreshedKeyspaces)
             || (!tokenMapEnabledBefore && tokenMapEnabled))
```
tomekl007, line 107: There is one method in this class:
`setSchemaEnabled`
that calls `return refreshSchema(null, false, true);` and doesn't handle error in `whenComplete`. Shouldn't it be handled? Or should the error be propagated and handled by the clients (same as for `refreshSchema`) since this is a public method?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
```diff
@@ -404,11 +415,7 @@ private void startSchemaRequest(CompletableFuture<Metadata> future) {
...
                   if (error != null) {
-                    Loggers.warnWithException(
-                        LOG,
-                        "[{}] Unexpected error while refreshing schema, skipping",
-                        logPrefix,
-                        error);
+                    currentSchemaRefresh.completeExceptionally(error);
                   }
                   singleThreaded.firstSchemaRefreshFuture.complete(null);
                 });
```
tomekl007, line 421: In this line:
https://github.com/datastax/java-driver/blob/b3475942c80b83c379181516db9b158d6cc72407/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java#L454
Are we still catching an exception and fail silently? Shouldn't we add the logging here?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
olim7t: `setSchemaMetadataEnabled` returns the future to the caller, so we expect it to handle potential errors.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
olim7t: We are not failing silently, but propagating the error (via the future). Callers will either surface the future to the client, or add a callback that logs the error.

PR #1335, opened 2019-09-26
Introduce shallowEquals() to KeyspaceMetadata

nastra:
This would allow subclasses to implement their own version of a shallow
comparison


core/src/main/java/com/datastax/oss/driver/api/core/metadata/schema/KeyspaceMetadata.java
```diff
@@ -245,4 +245,6 @@ default String describeWithChildren(boolean pretty) {
 
     return builder.build();
   }
+
+  boolean shallowEquals(Object other);
```
olim7t: This should be default, in order to avoid a breaking change. You can implement it in terms of the other interface methods.

olim7t (changes requested):
LGTM except for the fact that the method should be default.

I didn't see your PR before I made the comments on the ticket, but glad we came to the same conclusion.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/schema/KeyspaceMetadata.java
nastra: good idea, pushed a fix

olim7t (approved):

PR #1336, opened 2019-09-26
JAVA-2452: Allow "none" as a compression option

absurdfarce:


core/src/test/java/com/datastax/oss/driver/internal/core/context/MockedDriverContextFactory.java
```diff
@@ -0,0 +1,68 @@
...
+        ProgrammaticArguments.builder()
+            .addTypeCodecs(new TypeCodec[0])
+            .withNodeStateListener(nodeStateListener)
+            .withSchemaChangeListener(schemaChangeListener)
+            .withRequestTracker(requestTracker)
+            .withLocalDatacenters(Maps.newHashMap())
+            .withNodeFilters(Maps.newHashMap())
+            .withClassLoader(classLoader)
+            .build();
+    return new DefaultDriverContext(configLoader, args);
```
absurdfarce, line 66: With the conversion of this test to the new constructor there are only a few integration tests continuing to use the old (deprecated) constructor.  I'm happy to change those tests to use the new constructor (and remove the old constructor entirely) as part of this PR if that's judged to be useful.... I don't have enough context to know if that's a Good Idea :tm: or not.

core/src/test/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContextTest.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  @Test
+  public void should_create_noop_compressor_if_defined_as_none() {
+
+    DefaultDriverContext ctx = buildMockedContextWithDef("none");
+    Compressor<ByteBuf> compressor = ctx.getCompressor();
+    assertThat(compressor).isNotNull();
+    assertThat(compressor).isInstanceOf(NoopCompressor.class);
+  }
```
absurdfarce, line 79: We were indirectly testing the cases above via the StartupOptionsBuilderTest but it seemed worthwhile to explicitly test this as the DefaultDriverContext level as well.  As a nice side effect the additional need for a mocked DefaultDriverContext instance (perhaps given an execution profile) encouraged the move of that logic into a discrete factory impl.

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
```diff
@@ -361,20 +361,23 @@ protected EventBus buildEventBus() {
...
-    if (defaultProfile.isDefined(DefaultDriverOption.PROTOCOL_COMPRESSION)) {
-      String name = defaultProfile.getString(DefaultDriverOption.PROTOCOL_COMPRESSION);
-      if (name.equalsIgnoreCase("lz4")) {
+    if (!defaultProfile.isDefined(DefaultDriverOption.PROTOCOL_COMPRESSION)) {
+      return Compressor.none();
+    }
+
+    String name = defaultProfile.getString(DefaultDriverOption.PROTOCOL_COMPRESSION);
+    switch (name) {
+      case "lz4":
```
olim7t, line 367: The options were case-insensitive, let's preserve that behavior.
If you want to keep the switch statement, you can lower-case `name`.

core/src/test/java/com/datastax/oss/driver/internal/core/context/StartupOptionsBuilderTest.java
```diff
@@ -87,22 +61,43 @@ private void assertDefaultStartupOptions(Startup startup) {
...
+  }
+
+  @Test
+  public void should_build_startup_options_with_snappy_compression() {
+
+    doCompressorStartupOptionsTest("snappy");
+  }
+
+  @Test
+  public void should_build_startup_options_with_invalid_compression() {
```
olim7t: Nit: maybe this should be called `should_fail_to_build...`

olim7t (changes requested):

core/src/test/java/com/datastax/oss/driver/internal/core/context/MockedDriverContextFactory.java
olim7t: Yes, we didn't go over the tests when we deprecated the constructor, but it would be a good idea to do it.
I'd just like to keep that as a separate commit though, so maybe it's simpler to open a separate PR.

absurdfarce: Seems reasonable; I'll handle that in a separate PR once this goes in

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
absurdfarce: Good catch; that was a straight miss on my part. :(  Will fix, and I'll probably add some unit tests around that to validate going forward.

core/src/test/java/com/datastax/oss/driver/internal/core/context/StartupOptionsBuilderTest.java
absurdfarce: Agreed... I think that's probably a more descriptive name.

absurdfarce:
Also missed an update to the README as well; will add that when addressing the other comments.

core/src/test/java/com/datastax/oss/driver/internal/core/context/StartupOptionsBuilderTest.java
```diff
@@ -87,22 +61,43 @@ private void assertDefaultStartupOptions(Startup startup) {
...
+  @Test
+  public void should_build_startup_options_with_lz4_compression() {
+
+    doCompressorStartupOptionsTest("lz4");
+  }
+
+  @Test
+  public void should_build_startup_options_with_snappy_compression() {
+
+    doCompressorStartupOptionsTest("snappy");
```
tomekl007: I think you can replace those one line tests with one parameterized test

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
```diff
@@ -361,20 +361,23 @@ protected EventBus buildEventBus() {
   @SuppressWarnings("unchecked")
   protected Compressor<ByteBuf> buildCompressor() {
```
tomekl007, line 362: Suggestion: maybe instead of mocking whole `DriverContext` you can just test this method?
It seems that this method is only using the `DriverExecutionProfile defaultProfile` so you don't need to access anything from `driverContext` besides that. 
But to make it happen you would need to create a second method that is taking `DriverExecutionProfile` as an argument and this method is just delegating to that second method: `buildCompressor(){return buildCompressor(getConfig().getDefaultProfile())}` but I don't know if this is worth the effort.

core/src/test/java/com/datastax/oss/driver/internal/core/context/StartupOptionsBuilderTest.java
absurdfarce: You're thinking of... the built-in JUnit support for parameterized tests @tomekl007 ?  Yeah, I think that could be made to work... but there are some things I'm not super fond of there:

* The reliance on test state (via either the constructor usage or injection) isn't super-awesome
* The inability to mix and match parameterized vs. other tests in the same class seems problematic

Seems like the only way to address that last problem would be to create a distinct test just for these two cases... and that seems like an increase in complexity.

Or were you thinking of something more along the lines of https://github.com/Pragmatists/JUnitParams?  That addresses both concerns above with a much simpler API.  I could get behind that... is it a big issue to add that as another dep?

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
absurdfarce: It's an interesting idea @tomekl007 , but I'm reluctant to change the API of DefaultDriverContext to make that happen.  Your point does remind me of something I was wondering, though.

Because of the isolation built into the various components of DefaultDriverContext we probably don't _need_ all the mocks created in MockedDriverContextFactory; we can probably just pass in nulls for most of the dependencies and only do the bits that actually setup the plumbing for the config loader.  I just tested that out locally and it _does_ work.  But given that we already have a working impl now _with_ all the mocks it feels like passing in a bunch of nulls via ProgrammaticArguments would be a step backwards.

Abstracting creation of a mock DefaultDriverContext into the factory leaves open the door for future tests to build upon what's there now to provide different ways of creating these mocks in the future, which seems like something of a goodness.  Hopefully this will enable DefaultDriverContextTest to grow over time as things change in the default context. :crossed_fingers:

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
```diff
@@ -361,20 +361,23 @@ protected EventBus buildEventBus() {
...
-    if (defaultProfile.isDefined(DefaultDriverOption.PROTOCOL_COMPRESSION)) {
-      String name = defaultProfile.getString(DefaultDriverOption.PROTOCOL_COMPRESSION);
-      if (name.equalsIgnoreCase("lz4")) {
+    if (!defaultProfile.isDefined(DefaultDriverOption.PROTOCOL_COMPRESSION)) {
+      return Compressor.none();
+    }
+
+    String name = defaultProfile.getString(DefaultDriverOption.PROTOCOL_COMPRESSION);
+    switch (name.toLowerCase()) {
+      case "lz4":
```
absurdfarce, line 367: This is feeling more and more like a case for an enum.  There used to be such an enum in 3.x.... did we move away from that for a reason?

core/src/test/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContextTest.java
```diff
@@ -0,0 +1,93 @@
...
+
+  @Test
+  public void should_create_noop_compressor_if_defined_as_none() {
+
+    doCreateCompressorTest(Optional.of("none"), NoopCompressor.class);
+
+    /* Config should be case-insensitive */
+    doCreateCompressorTest(Optional.of("NONE"), NoopCompressor.class);
+    doCreateCompressorTest(Optional.of("NoNe"), NoopCompressor.class);
+    doCreateCompressorTest(Optional.of("nONe"), NoopCompressor.class);
```
absurdfarce: This (and the similar cases above) are a great argument for generative testing... but per the conversation with @tomekl007 elsewhere re: parameterized tests I didn't want to introduce a new dependency just for this case.  I'm happy to look at this more deeply if something like that is desired, though.

absurdfarce:
It looks like the re-run actually was successful; see https://travis-ci.com/datastax/java-driver/builds/129500149 for details.  The PR status apparently... just didn't update?

core/src/test/java/com/datastax/oss/driver/internal/core/context/StartupOptionsBuilderTest.java
tomekl007: We are already using parameterized tests in this project via: 
`com.tngtech.java.junit.dataprovider`
(see for example `DataTypeIT`) - I think we should stick to that library and approach of parameterized tests :)

core/src/test/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContextTest.java
tomekl007: We should use `com.tngtech.java.junit.dataprovider`, see my comment:
https://github.com/datastax/java-driver/pull/1336#discussion_r329432605

core/src/test/java/com/datastax/oss/driver/internal/core/context/StartupOptionsBuilderTest.java
absurdfarce: Nice!  I think that takes care of my concerns mentioned above... will gladly switch to that.  Thanks for the pointer @tomekl007 !

core/src/test/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContextTest.java
absurdfarce: It's not exactly equivalent to generative testing... but switching to the dataprovider seems to strike a pretty reasonable middle ground.

core/src/test/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContextTest.java
```diff
@@ -0,0 +1,85 @@
...
+    return MockedDriverContextFactory.defaultDriverContext(Optional.of(defaultProfile));
+  }
+
+  private void doCreateCompressorTest(Optional<String> configVal, Class<?> expectedClz) {
+
+    DefaultDriverContext ctx = buildMockedContext(configVal);
+    Compressor<ByteBuf> compressor = ctx.getCompressor();
+    assertThat(compressor).isNotNull();
+    assertThat(compressor).isInstanceOf(expectedClz);
+  }
```
absurdfarce, line 52: I thought about creating a data provider for both params, something along the lines of the following:

```java
  @DataProvider
  public static Object[][] createCompressorDP() {

	  Object[][] rv = new Object[12][2];
	  int idx = 0;
	  for (String name : new String[] { "lz4", "lZ4", "Lz4", "LZ4"}) {
		  rv[idx++] = new Object[] { name, Lz4Compressor.class };
	  }
	  for (String name : new String[] { "snappy", "SNAPPY", "sNaPpY", "SNapPy"}) {
		  rv[idx++] = new Object[] { name, SnappyCompressor.class };
	  }
	  for (String name : new String[] { "none", "NONE", "NoNe", "nONe"}) {
		  rv[idx++] = new Object[] { name, NoopCompressor.class };
	  }
	  return rv;
  }

  @Test
  @UseDataProvider("createCompressorDP")
  public void should_create_compressor(String configVal, Class<?> expectedClz) {

	    DefaultDriverContext ctx = buildMockedContext(Optional.of(configVal));
	    Compressor<ByteBuf> compressor = ctx.getCompressor();
	    assertThat(compressor).isNotNull();
	    assertThat(compressor).isInstanceOf(expectedClz);
	  }
```

This made the test output just a bit harder to read, though, so I settled on the middle ground above.

core/src/test/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContextTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+  @Test
+  @DataProvider({"lz4", "lZ4", "Lz4", "LZ4"})
```
tomekl007, line 55: LGTM, but to make it even more maintanable maybe you could coalesce it to one test?
```
@DataProvider({"lz4", "lZ4", "Lz4", "LZ4","snappy", "SNAPPY", "sNaPpY", "SNapPy","none", "NONE", "NoNe", "nONe" })
public void should_create_compressor()
```

changelog/README.md
```diff
@@ -9,6 +9,7 @@
 - [documentation] JAVA-2416: Update paging section in the manual
 - [improvement] JAVA-2402: Add setTracing(boolean) to StatementBuilder
 - [bug] JAVA-2466: Set idempotence to null in BatchStatement.newInstance
+- [improvement] JAVA-2452: Allow "none" as a compression option
```
adutra: Nit: changelog entries should be inserted first in their section for new drivers, and last for legacy drivers. I know, we don't make it any easier :-D

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
adutra: Unfortunately we don't have `getEnum()` in `DriverExecutionProfile`.

core/src/test/java/com/datastax/oss/driver/internal/core/context/MockedDriverContextFactory.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /* Setup remaining mocks */
+    final NodeStateListener nodeStateListener = mock(NodeStateListener.class);
+    final SchemaChangeListener schemaChangeListener = mock(SchemaChangeListener.class);
+    final RequestTracker requestTracker = mock(RequestTracker.class);
+    final ClassLoader classLoader = mock(ClassLoader.class);
+
+    ProgrammaticArguments args =
+        ProgrammaticArguments.builder()
+            .addTypeCodecs(new TypeCodec[0])
```
adutra: This line can be removed.

core/src/test/java/com/datastax/oss/driver/internal/core/context/MockedDriverContextFactory.java
```diff
@@ -0,0 +1,68 @@
...
+import com.datastax.oss.driver.api.core.config.DriverExecutionProfile;
+import com.datastax.oss.driver.api.core.metadata.NodeStateListener;
+import com.datastax.oss.driver.api.core.metadata.schema.SchemaChangeListener;
+import com.datastax.oss.driver.api.core.session.ProgrammaticArguments;
+import com.datastax.oss.driver.api.core.tracker.RequestTracker;
+import com.datastax.oss.driver.api.core.type.codec.TypeCodec;
+import com.datastax.oss.driver.shaded.guava.common.collect.Maps;
+import java.util.Optional;
+
+public class MockedDriverContextFactory {
```
adutra: Could be package-private.

core/src/test/java/com/datastax/oss/driver/internal/core/context/MockedDriverContextFactory.java
```diff
@@ -0,0 +1,68 @@
...
+
+  public static DefaultDriverContext defaultDriverContext() {
+    return defaultDriverContext(Optional.empty());
+  }
+
+  public static DefaultDriverContext defaultDriverContext(
+      Optional<DriverExecutionProfile> profileOption) {
+
+    /* If the caller provided a profile use that, otherwise make a new one */
+    final DriverExecutionProfile profile = profileOption.orElse(mock(DriverExecutionProfile.class));
```
adutra: Unnecessary `final` keywords.

adutra (changes requested):

core/src/test/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContextTest.java
absurdfarce: Unfortunately you need to match up an input name with an expected compressor class to make the assertions work out... so you'd need a bit more to make it work.  I experimented with something like what you're describing in https://github.com/datastax/java-driver/pull/1336/files#r329775123 but it seemed like splitting things out into discrete test cases kept the output readable while still keeping things maintainable.

core/src/test/java/com/datastax/oss/driver/internal/core/context/MockedDriverContextFactory.java
absurdfarce: Agreed, will fix

core/src/test/java/com/datastax/oss/driver/internal/core/context/MockedDriverContextFactory.java
absurdfarce: I'm inclined to keep them (unless there's a convention not to do so).  Explicitly marking those refs as final seems like a reasonable communication of intent.

core/src/test/java/com/datastax/oss/driver/internal/core/context/MockedDriverContextFactory.java
absurdfarce: This is a holdover from the original impl.  Seems like there might be merit in explicitly providing something here even if the current impl of provides a sane default.  Either way, it seems like the usage here should match what we do for withLocalDatacenters() and withNodeFilters(); seems like we should rely on builder defaults for none or all three, does it not?

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
absurdfarce: Sure, but you could just as easily create the enum instance within this method, something like:

```java
  enum ValidCompressor {
    LZ4,
    SNAPPY,
    NONE
  }

  protected Compressor<ByteBuf> buildCompressor() {
    DriverExecutionProfile defaultProfile = getConfig().getDefaultProfile();
    if (!defaultProfile.isDefined(DefaultDriverOption.PROTOCOL_COMPRESSION)) {
      return Compressor.none();
    }

    String name = defaultProfile.getString(DefaultDriverOption.PROTOCOL_COMPRESSION);
    ValidCompressor compressor = ValidCompressor.valueOf(name.toUpperCase());
    switch (compressor) {
      case LZ4:
        return new Lz4Compressor(this);
      case SNAPPY:
        return new SnappyCompressor(this);
      case NONE:
        return Compressor.none();
      default:
            String.format(
                "Unsupported compression algorithm '%s' (from configuration option %s)",
                name, DefaultDriverOption.PROTOCOL_COMPRESSION.getPath()));
    }
  }
```

This is closer to the ProtocolOptions.Compression enum in 3.x.  It's probably _marginally_ cleaner than relying on checks for explicit string values but I don't feel _terribly_ strongly about it.

core/src/test/java/com/datastax/oss/driver/internal/core/context/MockedDriverContextFactory.java
absurdfarce: Oh, nevermind, that's an "add" not a "with"... yeah, I see what you mean now.  Removing.

core/src/test/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContextTest.java
absurdfarce: I suppose this isn't _as_ bad now that I look at it:

```java
  @DataProvider
  public static List<List<Object>> createCompressorDP() {

	  Multimap<Class<?>,String> themap = Multimaps.newListMultimap(Maps.newHashMap(), Lists::newArrayList);
	  themap.putAll(Lz4Compressor.class, ImmutableList.of("lz4", "lZ4", "Lz4", "LZ4"));
	  themap.putAll(SnappyCompressor.class, ImmutableList.of("snappy", "SNAPPY", "sNaPpY", "SNapPy"));
	  themap.putAll(NoopCompressor.class, ImmutableList.of("none", "NONE", "NoNe", "nONe"));
	  
	  return themap.entries().stream()
			  .map(entry -> ImmutableList.<Object>of(Optional.of(entry.getValue()), entry.getKey()))
  }

  @Test
  @UseDataProvider("createCompressorDP")
  public void doCreateCompressorTest(Optional<String> configVal, Class<?> expectedClz) {

    DefaultDriverContext ctx = buildMockedContext(configVal);
    Compressor<ByteBuf> compressor = ctx.getCompressor();
    assertThat(compressor).isNotNull();
    assertThat(compressor).isInstanceOf(expectedClz);
  }
```

@tomekl007 what do you think... do you prefer the all-in-one data provider or discrete providers for discrete test methods?

tomekl007: We are using this pattern:
```
@Test
  @UseDataProvider("compressors")
  public void should_create_snappy_compressor(String name, Class<?> classT) {

    doCreateCompressorTest(Optional.of(name), classT);
  }


  @DataProvider
  public static Object[][] compressors() {
    return new Object[][] {
        new Object[]{"lz4", Lz4Compressor.class},
        new Object[]{"lZ4", Lz4Compressor.class},
        // lz4..
        new Object[]{"snappy", SnappyCompressor.class},
        // snappy..
        new Object[]{"none", NoopCompressor.class},
        //none..
    };
  }
```

absurdfarce: After mulling this over for a bit I think I'll stick with what I have now in the PR.  I think it's easier to read and grok at a glance than the alternatives.  I also like the fact that it highlights the interrelationship of the "noop if undefined" and "noop if explicitly called for" behaviours.  It's also a bit easier to identify individual test cases when running this test in an IDE.  If both params are passed along (i.e. if either of the proposed explicit provider functions are used) the message that gets printed there is kind of ugly.

core/src/test/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContextTest.java
olim7t: Yeah we've not explored generative testing on the project, I'm not sure if we have many good cases for it...
In this particular case I think it would be a bit overkill.

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
olim7t: I was going to mention extensibility as an argument against an enum, but actually if someone adds their own compression scheme, they'll have to override `buildCompressor` anyway so they could handle a special case that is not an enum constant.

So like you I don't feel terribly strongly about it either. We wouldn't expose the enum in the public API anyway, so let's keep it like this for now.

Site note: I just noticed that Cassandra added a `ZstdCompressor` recently, creating a ticket for that.

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
olim7t: One more thing I missed during my first review: config getters can take a default value, this simplifies this method a little bit. Addressing this directly in the merge.

olim7t (approved):

PR #1337, opened 2019-09-26
JAVA-2464: Fix initial schema refresh when reconnect-on-init is enabled

olim7t:
Motivation:

When reconnect-on-init is enabled and the driver does not connect
immediately, the schema metadata is empty.

This is because the initial schema refresh is chained on
ControlConnection.firstConnectionAttemptFuture. The goal was to allow
the session to initialize without waiting for the control connection (if
there is a custom topology monitor that uses another mechanism). But in
the scenario above, firstConnectionAttemptFuture is failed.

Modifications:

Chain the schema refresh to ControlConnection.initFuture (complete,
successful connection).
Remove firstConnectionAttemptFuture.

Result:

The schema is present at startup.

Note that if a custom topology monitor is in place, and it is the
metadata manager that initializes the control connection, the first
schema refresh (and therefore the whole session initialization) will
have to wait for the control connection to successfully connect. All
things considered, this is pretty logical.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
@@ -102,9 +102,8 @@ public class ControlConnection implements EventCallback, AsyncAutoCloseable {
    * @param listenToClusterEvents whether to register for TOPOLOGY_CHANGE and STATUS_CHANGE events.
    *     If the control connection has already initialized with another value, this is ignored.
    *     SCHEMA_CHANGE events are always registered.
-   * @param reconnectOnFailure whether to schedule a reconnection if the initial attempt fails (this
-   *     does not affect the returned future, which always represent the outcome of the initial
-   *     attempt only).
+   * @param reconnectOnFailure whether to schedule a reconnection if the initial attempt fails (if
+   *     true, the returned future will only complete once the reconnection has succeeded).
    * @param useInitialReconnectionSchedule if no node can be reached, the type of reconnection
    *     schedule to use. In other words, the value that will be passed to {@link
    *     ReconnectionPolicy#newControlConnectionSchedule(boolean)}. Note that this parameter is only
@@ -128,10 +127,6 @@ public class ControlConnection implements EventCallback, AsyncAutoCloseable {
 
   public boolean isInit() {
     return singleThreaded.initFuture.isDone();
-  }
-
-  public CompletionStage<Void> firstConnectionAttemptFuture() {
-    return singleThreaded.firstConnectionAttemptFuture;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
```diff
@@ -462,44 +454,50 @@ private void connect(
...
+            .refreshNodes()
+            .whenComplete(
+                (result, error) -> {
+                  if (error != null) {
+                    LOG.debug("[{}] Error while refreshing node list", logPrefix, error);
+                  } else {
+                    try {
+                      // A failed node list refresh at startup is not fatal, so this might be the
+                      // first successful refresh; make sure the LBP gets initialized (this is a
+                      // no-op
```
adutra: Nit: indentation.

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
```diff
@@ -462,44 +454,50 @@ private void connect(
     }
 
     private void onSuccessfulReconnect() {
-      // If reconnectOnFailure was true and we've never connected before, complete the future now,
-      // otherwise it's already complete and this is a no-op.
-      initFuture.complete(null);
+      // If reconnectOnFailure was true and we've never connected before, complete the future now to
+      // signal that the initialization is complete.
+      boolean isFirstConnection = initFuture.complete(null);
```
adutra, line 459: This method is only called from the lambda in `reconnect()`, which in turn is only called from the method reference created when the `Reconnection` object is created. So can this `isFirstConnection` variable ever be true, and if so, in which case?

adutra (commented):
I must say, the code is very hard to follow: Runnables, Callables, Suppliers and Consumers all over the place, all of them created as lambda expressions, and the logic flow going back and forth between this class and the `Reconnection` class.

olim7t:
> I must say, the code is very hard to follow

Yes, session initialization is probably the most complex part of driver 4. I wish it was easier to read, but a lot of it is inherent though: when you consider that everything has to be fully asynchronous, some key internal components are pluggable, and the reconnect-on-init option completely changes the behavior, that's a lot to handle.

> Runnables, Callables, Suppliers and Consumers all over the place

That's pretty standard if you're chaining futures asynchronously. If the problem is with the lambdas, I can try extracting more method references, although I'm not sure it's always an improvement.

If anything, this PR simplifies things slightly.

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
olim7t: Yes, if the reconnection is not part of the initial attempt:
- the driver connects successfully.
- later, the control node goes down.
- `onChannelClosed` is invoked, it calls `reconnection.start()`
- the `Reconnection` object starts scheduling attempts. Each attempt will invoke `reconnect()`. If the attempt fails, `Reconnection` keeps trying. Otherwise, `onSuccessfulReconnect` is invoked.

PR #1338, opened 2019-10-01
JAVA-2444: Add  method setRoutingKey(ByteBuffer...) to StatementBuilder

absurdfarce:


core/src/main/java/com/datastax/oss/driver/api/core/cql/StatementBuilder.java
```diff
@@ -123,6 +124,11 @@ public SelfT setRoutingKey(@Nullable ByteBuffer routingKey) {
     return self;
   }
 
+  @NonNull
```
adutra, line 128: Other methods in this class have this kind of javadocs:

```
  /** @see Statement#setRoutingKey(ByteBuffer) */
```

So I think we should do the same here.

core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
```diff
@@ -16,61 +16,95 @@
...
+    byte[] arr = new byte[8];
+    random.nextBytes(arr);
+    return ByteBuffer.wrap(arr);
+  }
+
+  @Test
+  public void should_match_set_routing_key_vararg() {
+
+    Random random = new Random();
+    ByteBuffer buff1 = randomBuffer(random);
```
adutra: Do we really need random byte buffers? What about just this:

```
ByteBuffer buff1 = Bytes.fromHexString("0xcafe");
ByteBuffer buff2 = Bytes.fromHexString("0xbabe");
```

core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
```diff
@@ -16,61 +16,95 @@
...
 
+import java.nio.ByteBuffer;
+import java.util.Random;
 import org.junit.Test;
 
 public class StatementBuilderTest {
 
-  private static class NullStatementBuilder
-      extends StatementBuilder<NullStatementBuilder, SimpleStatement> {
+  private static class MockSimpleStatementBuilder
```
adutra, line 28: Not sure we need this inner class? I just replaced all references to it with `SimpleStatementBuilder builder = new SimpleStatementBuilder("IRRELEVANT");` (or `builder = new SimpleStatementBuilder(template)`) and the tests passed.

adutra (changes requested):

adutra:
I forgot to mention a few things missing:

1. Please add an entry to the changelog.
2. Change the commit message to include the Jira ticket number and its title (note that I updated the title since it was a bit misleading).
3. Change the PR title before merging to also include: Jira ticket number + title.

core/src/main/java/com/datastax/oss/driver/api/core/cql/StatementBuilder.java
absurdfarce: Agreed; I saw those and simply forgot to make sure they were here.  I'll add those in for sure.

core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
absurdfarce: This came up in the PR for JAVA-2402 (which introduced the test in question).  The functionality we're testing here is implemented in StatementBuilder so my intent was to test that class directly.  Although it's not necessarily likely a subclass can always modify that behaviour; better to test as close to the functionality in question as possible.

Original comment should be at https://github.com/datastax/java-driver/pull/1330#discussion_r326422401

absurdfarce:
Good points, @adutra .  I've added this change to the changelog (for some reason I consistently forget to do that) and have updated the title of this PR to include the JIRA title.  I'll do a squash merge once everything is addressed and update the commit message to be JIRA ID + title at that time.

I think... that covers everything?

core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
absurdfarce: I don't have a strong opinion either way I guess.  I'm sympathetic to the argument for randomized testing (to potentially catch data-driven oddness).  I'm also sympathetic to the argument for predictable testing to catch regressions.

If there's a preference for a fixed value here I can make that change as well.

core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
adutra: IOW we are trying to test an abstract class, and that's why we had to introduce this concrete class. But we shouldn't be testing anything abstract in the first place, only its concrete implementations. Which are 3 currently:

1. BatchStatementBuilder
2. BoundStatementBuilder
3. SimpleStatementBuilder

So why don't you parameterize each test and make it run against an instance of each of the types above?

core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
adutra: Yeah I hear you; but I think I have a slight preference for test repeatability. Random tests, unless we are really dealing with randomness, are usually of little value; here for instance, I can't see why the actual byte buffer contents would matter – except, of course, for special values such as `null` or empty.

absurdfarce: Okay, I'll shift this over to a fixed string in the next round of changes.

core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
absurdfarce: I'm not sure I agree with the reasoning here.

There is a function implemented on the abstract super-class StatementBuilder.  That function should be tested just like any other function (so that we know it does the right thing if a sub-class doesn't override it).  We could test that by using an existing sub-class which doesn't (currently) override that method... but that test then is subject to silent changes in behaviour if that sub-class later overrides the method.

In order to avoid that concern, the unit test extends the abstract sub-class directly.

Testing all three concrete sub-classes _in this test_ seems like a bad idea for similar reasons; this test is intended to validate behaviours in the abstract super-class and the introduction of a new concrete sub-class requires an update to this unrelated test in a way that's not obvious.  Now, if you're arguing that we should have similar tests for each of the three concrete sub-classes _in addition to_ this test I could certainly go along with that.

I'm not opposed to testing the concrete sub-classes at all.  I just want to be very clear about the scope of each individual test.

olim7t (approved):
LGTM

No strong feelings either way about the tests. If we add a test for each builder implementation, we'll want to cover other methods as well so that should be a different PR.

core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
adutra: OK OK I don't feel strong about it. Creating a dummy child class to test a parent abstract class seems reasonable.

adutra (approved):

PR #1339, opened 2019-10-01
JAVA-2428: Add developer docs

olim7t:


manual/developer/admin/README.md
```diff
@@ -0,0 +1,323 @@
...
+  * the metadata manager, to read schema metadata from `system_schema.*`.
+
+It has its own reconnection mechanism (if the channel goes down, a new one will be opened to another
+node in the cluster) and some logic for initialization and shutdown.
+
+Note that the control connection is really just an implementation detail of the metadata manager and
+topology monitor: if those components are overridden with custom versions that use other means to
+get their data, the driver will detect it and not initialize the control connection (at the time of
+writing, the session also references the control connection directly, but that's a bug:
+[JAVA-2473](https://datastax-oss.atlassian.net/browse/JAVA-2473)).
```
tomekl007, line 53: I think we should add the comment and line number https://github.com/datastax/java-driver/pull/1339/files#diff-9e9f825f2c44fb5f74b7c95ae568a28aR53
in this jira to remove this from documentation once it will be fixed

manual/developer/admin/README.md
```diff
@@ -0,0 +1,323 @@
...
+* `TopologyEvent` is an external signal about the state of a node (by default, a `TOPOLOGY_CHANGE`
+  or `STATUS_CHANGE` gossip event received on the control connection). This is considered as a mere
+  suggestion, that the driver may or may not decide to follow;
+* `NodeStateEvent` is an actual decision made by the driver to change a node to a given state.
+
+`NodeStateManager` essentially transforms topology events, as well as other internal signals, into
+node state events.
+
+In general, other driver components only react to node state events, but there are a few exceptions:
+for example, if a connection pool is reconnecting and the next attempt is scheduled in 5 minutes,
```
tomekl007, line 132: I wonder if we should add the link to the actual code where this is happening?
Or we don't want to expose links to internal code because of the maintenance burden that it could cause?

manual/developer/admin/README.md
```diff
@@ -0,0 +1,323 @@
...
+         |                |        |known already |         |                 |
+         |                |        |           {d}|         |                 |
+         |                |        +------+-------+         |                 |
+         |                |               |                 |                 |
+         |                |               |     addNode     |                 |
+         |                |               +---------------->|                 |
+         |                |               |                 |  getNewNodeInfo |
+         |                |               |                 +---------------->|
+         |                |               |                 |                 |
+         |                 query(SELECT FROM system.peers)                    |
```
tomekl007, line 171: do we query only `system.peers`? What about `peers_v2` and `local`?

manual/developer/native_protocol/README.md
```diff
@@ -0,0 +1,180 @@
...
++-------------+
+|  CodecGroup |
++------+------+
+       |
+       |    +------------------------+
+       +----+ ProtocolV3ClientCodecs |
+       |    +------------------------+
+       |
+       |    +------------------------+
+       +----+ ProtocolV3ServerCodecs |
```
tomekl007, line 115: Does server codes exists only for `V3`? (`ProtocolV3ServerCodecs`) If yes, why?

tomekl007 (approved):

manual/developer/admin/README.md
```diff
@@ -0,0 +1,323 @@
...
+    the event bus, to be processed by other components;
+  * schema events are propagated directly to the metadata manager, to trigger a refresh;
+* provide a way to query system tables. In practice, this is used by:
+  * the topology monitor, to read node information from `system.local` and `system.peers`;
+  * the metadata manager, to read schema metadata from `system_schema.*`.
+
+It has its own reconnection mechanism (if the channel goes down, a new one will be opened to another
+node in the cluster) and some logic for initialization and shutdown.
+
+Note that the control connection is really just an implementation detail of the metadata manager and
```
adutra, line 49: It doesn't appear clearly from the diagram above that the control connection is just an implementation detail of the metadata manager. For instance, the topology monitor is also accessing it directly.

manual/developer/native_protocol/README.md
```diff
@@ -0,0 +1,180 @@
...
+       +---------------------------+ Message.Codec     |
+           1 codec per opcode      +-------------------+
+           and protocol version    | B encode(Message) |
+                                   | Message decode(B) |
+                                   +-------------------+
+```
+
+Most of the time, you'll want to use the full set of message codecs for a given protocol version.
+`CodecGroup` provides a convenient way to register multiple codecs at once. The project provides
+default implementations for all supported protocol version, both for clients like the driver (e.g.
```
adutra, line 101: Nit: version -> versions

manual/developer/request_execution/README.md
```diff
@@ -0,0 +1,314 @@
...
+the time. If on the other hand the fluctuations are rare and predictable (e.g. peak for holiday
+sales), then a manual configuration change is good enough.
+
+#### Wait-free
+
+To get a connection to a node, client code calls `ChannelPool.next()`. This returns the less busy
+connection, based on the the `getAvailableIds()` counter exposed by
+[InFlightHandler](netty_pipeline/#in-flight-handler).
+
+If all connections are busy, there is no queuing; the driver moves to the next node immediately. The
```
adutra, line 98: How does this articulate with load balancing policies? IOW, am I right to infer that if the LBP picks nodeA, we might end up querying nodeB if all connections to nodeA are busy?

adutra (approved):
These docs are outstanding, very good work!

PR #1340, opened 2019-10-04
JAVA-2438: Improve log messages that format Node instances

absurdfarce:


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultNode.java
```diff
@@ -180,7 +180,7 @@ public NodeMetricUpdater getMetricUpdater() {
 
   @Override
   public String toString() {
-    return String.format("%s(%s,%s)", super.toString(), hostId, endPoint);
+    return String.join("@", getHostId().toString().trim(), getEndPoint().toString().trim());
```
absurdfarce, line 183: I would've preferred to have this as a default method on the interface... but that's not allowed. :(

tomekl007: Shouldn't we have a unit test that is testing that behavior?

absurdfarce: Hmmmm, good question.  My original inclination was to say we probably don't need one since the toString() impl doesn't really affect functionality.  But I don't see how it _hurts_ anything, and it was important enough to file a ticket to address in the first place... so yeah, maybe we should.

tomekl007:
https://travis-ci.com/datastax/java-driver/jobs/244368755
Licence headers missing

absurdfarce:
Grrr, thanks @tomekl007 ... that's only the third time I've forgotten that now. :(

Pushed a fix + merge of 4.x in as well

tomekl007:
> Grrr, thanks @tomekl007 ... that's only the third time I've forgotten that now. :(
> 
> Pushed a fix + merge of 4.x in as well

Sometimes I am also forgetting that. Maybe we would need a plugin for IntelliJ? :D

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultNode.java
tomekl007, line 183: why do we need to `trim()`?

tomekl007 (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultNode.java
absurdfarce: We don't need it right now; it's more a precautionary thing than anything else.  We're at the mercy of the toString() impls of a couple other classes here and I wanted to guard against the case in which either of those somehow find themselves with whitespace issues.

PR #1341, opened 2019-10-08
JAVA-2471: Consider dsePatch version when parsing the schema

absurdfarce:


core/src/test/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/DefaultSchemaQueriesFactoryTest.java
```diff
@@ -0,0 +1,130 @@
...
+          .args("4.0.1")
+          .expect(Expected.CASS_4)
+          .args("4.1.0")
+          .expect(Expected.CASS_4)
+          .build();
+
+  private static When<Expected> whenDse =
+      When.<Expected>builder()
+          /* DSE 6.0.0 */
+          .args("4.0.0.2284")
```
tomekl007: This is failing when I run it. It returns `Expected.CASS_4`

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/DefaultSchemaQueriesFactoryTest.java
```diff
@@ -0,0 +1,130 @@
...
+  @Test
+  @UseDataProvider("expected")
+  public void should_return_correct_schema_queries_impl(Expected expected, String version) {
+
+    final Node mockNode = mock(Node.class);
+    when(mockNode.getCassandraVersion()).thenReturn(Version.parse(version));
+
+    DefaultSchemaQueriesFactory factory = buildFactory();
+    SchemaQueries queries =
+        factory.newInstance(mockNode, mock(DriverChannel.class), mock(CompletableFuture.class));
```
tomekl007, line 120: You may want to add `SupressWarninig("unchecked")` for this one:
`mock(CompletableFuture.class)`

core/src/test/java/com/datastax/oss/driver/When.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+package com.datastax.oss.driver;
+
+import com.datastax.oss.driver.shaded.guava.common.collect.Collections2;
+import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
+import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableMap;
+import java.util.Iterator;
+import java.util.List;
+
+public class When<T> implements Iterable<Iterable<Object>> {
```
tomekl007: Where it is plugin into logic? I don't see any usage of this new class besides the test

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/DefaultSchemaQueriesFactoryTest.java
absurdfarce: Yeah, this is expected.  That's the root of the underlying ticket; we're returning schema queries for DSE < 6.7.0 that don't actually match up to what that version of DSE has available.  I implemented this test as a mechanism to document and validate the underlying problem... as of this commit I haven't actually implemented the fix yet. :)

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/DefaultSchemaQueriesFactoryTest.java
absurdfarce: Agreed.  I'd noticed this in Eclipse and hadn't actually put in the fix yet.  Will add in the next round of updates.

core/src/test/java/com/datastax/oss/driver/When.java
absurdfarce: Yeah, it's only a testing thing.  I added it to make the data provider definition more explicitly declarative.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/DefaultReplicationStrategyFactory.java
```diff
@@ -41,6 +41,8 @@ public ReplicationStrategy newInstance(Map<String, String> replicationConfig) {
         return new SimpleReplicationStrategy(replicationConfig);
       case "org.apache.cassandra.locator.NetworkTopologyStrategy":
         return new NetworkTopologyReplicationStrategy(replicationConfig, logPrefix);
+      case "org.apache.cassandra.locator.EverywhereStrategy":
+        return new EverywhereReplicationStrategy();
```
absurdfarce: A drive-by fix noticed while testing this against various DSE versions.  In addition to inappropriate queries against system_virtual_schema for DSE 6.0.x (the heart of this ticket) I was also seeing exceptions for failures to find a replication strategy impl for EverywhereStrategy.  Brought in the fix from the DSE side to address this failure.

core/src/test/java/com/datastax/oss/driver/When.java
olim7t: To be honest, I'm not a fan. To me, it doesn't make the provider declarations any easier to read.

If the goal is to document what each field in the provided data is (and I agree it's a weakness in the array syntax), then I find that a simple comment at the beginning does a pretty decent job. See example [here](https://github.com/datastax/java-driver/blob/4.x/query-builder/src/test/java/com/datastax/oss/driver/api/querybuilder/BuildableQueryTest.java#L42).

core/src/main/java/com/datastax/oss/driver/api/core/metadata/NodeProperties.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core.metadata;
+
+import com.datastax.oss.driver.api.core.Version;
+
+/** The keys for the additional properties stored in {@link Node#getExtras()}. */
+public class NodeProperties {
```
olim7t: Given that this is kind of temporary (we'll use `DseNodeProperties` after we merge the drivers), I'd rather put this in an internal package.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -347,18 +349,23 @@ private String getPeerTableName() {
...
-        .withBroadcastAddress(broadcastAddress)
-        .withListenAddress(listenAddress)
-        .withDatacenter(row.getString("data_center"))
-        .withRack(row.getString("rack"))
-        .withCassandraVersion(row.getString("release_version"))
-        .withTokens(row.getSetOfString("tokens"))
-        .withPartitioner(row.getString("partitioner"))
-        .withHostId(Objects.requireNonNull(row.getUuid("host_id")))
-        .withSchemaVersion(row.getUuid("schema_version"));
+    DefaultNodeInfo.Builder rv =
```
olim7t, line 351: Nit: we don't have rules for short-lived local variable names, but why `rv`?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/DefaultSchemaQueriesFactory.java
```diff
@@ -81,6 +82,19 @@ protected SchemaQueries newInstance(
     } else if (version.compareTo(Version.V4_0_0) < 0) {
       return new Cassandra3SchemaQueries(channel, refreshFuture, config, logPrefix);
     } else {
+
+      /* A bit of custom logic for DSE 6.0.x.  These versions report a Cassandra version of 4.0.0 but don't
+       * have support for system_virtual_schema tables supported by that version.  To compensate we return
+       * the Cassandra 3 schema queries here for those versions */
```
olim7t: Nit: could you avoid block-style comments (ie. use `//` style instead)?

Sometimes when you're on a test server and all you have is the command line and vi, it's convenient to use `/* ... */` to comment parts of the code, but that only works if it doesn't already contain such comments.

olim7t (changes requested):
Looks good, a few minor comments.

Regarding `EveryWhereStrategy`: I'd _really_ like to keep this as a separate commit, and if we keep it here I'm worried we'll forget and squash everything together. So I think it would be better to open a separate PR.

absurdfarce:
No worries about the EverywhereStrategy change; I deliberately isolated that to it's own commit in case we wanted to do something along the lines of what you're describing.  Do you want a new ticket for that or just a new PR on this ticket?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/DefaultSchemaQueriesFactory.java
absurdfarce: Sure, although the block syntax is common even throughout the existing code base.  I personally prefer it (I find the double-slash syntax visually very distracting) but I don't care enough to go to the mat for it.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
absurdfarce: rv = return value.  I use it quite a bit in cases like this, although my preference would be to avoid the variable declaration all together and just return the result of the expression. That's pretty hard to do with a conditional application such as this, though.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/NodeProperties.java
absurdfarce: Makes sense, I'll move it to an equivalent spot in the internal space

core/src/test/java/com/datastax/oss/driver/When.java
absurdfarce: I personally don't find that example nearly as declarative (or as readable) as what's in this PR, but at the same time I don't want to introduce an impl that's only going to be used in some subset of unit tests.  I'll see if I can re-work my test into a middle ground between those ideas (and remove the "when" impl) but it does seem to me like there's _considerable_ room to improve how we define these providers in the future.

absurdfarce:
I think all review comments have been addressed so @tomekl007 + @olim7t this is ready for another pass at your convenience.

Also, @olim7t , let me know what you want to do re: bringing in EverywhereStrategy in another PR.  My initial thought was that a new ticket just for that might be overkill... but as I thought more about it that seemed better than just being a random PR on another random ticket.  So maybe a new standalone JIRA ticket does make sense.

adutra (approved):

absurdfarce:
EverywhereStrategy support has been implemented in https://github.com/datastax/java-driver/pull/1343

adutra:
Beware that integration tests are hanging: https://jenkins-drivers.build.dsinternal.org/view/dsbulk/job/datastax.java-driver.PR.1341.default/

olim7t (approved):

PR #1342, opened 2019-10-10
JAVA-2497: Make DefaultExecutionInfo/DefaultNode serializable

nastra:
This is necessary because both types appear in numerous Exception
classes. On the server-side one can see the below error when things
fail:

`java.io.NotSerializableException: com.datastax.oss.driver.internal.core.metadata.DefaultNode`


nastra:
Actually I haven't even looked at the details in `DefaultExecutionInfo`, so rather consider this PR a starting point for a discussion as it definitely needs more work.

adutra:
Closing in favor of https://github.com/datastax/java-driver/pull/1348

PR #1343, opened 2019-10-11
JAVA-2471: Bring fix for JAVA-1708 into OSS driver

absurdfarce:


olim7t (approved):
👍 to the name change to match other strategies.

I've added a changelog entry and used the original ticket number.

PR #1344, opened 2019-10-14
JAVA-2492: Parse vertex and edge metadata as internal identifiers

absurdfarce:
…null checks


core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
```diff
@@ -16,21 +16,36 @@
...
 
 public class CqlIdentifiers {
 
-  public static List<CqlIdentifier> wrap(Iterable<String> in) {
+  public static List<CqlIdentifier> wrap(
+      @NonNull Iterable<String> in, @NonNull Function<String, CqlIdentifier> fn) {
+
+    Preconditions.checkNotNull(in, "Input Iterable must not be null");
+    Preconditions.checkNotNull(fn, "CqlIdentifier conversion function must not be null");
     ImmutableList.Builder<CqlIdentifier> out = ImmutableList.builder();
```
tomekl007: I think that to be consistent the result of apply `fn.apply(name)` should also be checked using 
`Preconditions.checkNotNull` but maybe it is not needed?

absurdfarce: Guava's immutable list builder will definitely be unhappy if we try to add a null value here.  But this isn't a problem at the moment, at least as the code is currently written.  CqlIdentifier.fromCql() is annotated with `@NonNull`; presumably that's why the old impl was safe to use it without such a check.  And since CqlIdentifier.fromInternal() is also marked with `@NonNull` the same logic should apply for all cases of the new impl.

Obviously if those guarantees go away things become problematic... but at least at the moment it isn't a concern.

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
```diff
@@ -16,21 +16,36 @@
...
     return out.build();
   }
 
+  public static List<CqlIdentifier> wrap(Iterable<String> in) {
+    return wrap(in, CqlIdentifier::fromCql);
+  }
+
+  public static List<CqlIdentifier> wrapInternal(Iterable<String> in) {
+    return wrap(in, CqlIdentifier::fromInternal);
+  }
```
absurdfarce, line 50: I do wonder about a couple things here:

* Presumably we should annotate the args to these methods with `@NonNull` as well?  Not sure if any static checker is smart enough to realize this method is a straight cut-through and that the params of the caller should be applied here
* Presumably both of these methods (as well as the underlying wrap(Iterable, Fn) call) should also be annotated with `@NonNull`?

adutra: I think so. Anyways, it doesn't hurt to annotate everything.

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
```diff
@@ -16,21 +16,36 @@
...
 import java.util.Map;
+import java.util.function.Function;
 
 public class CqlIdentifiers {
 
-  public static List<CqlIdentifier> wrap(Iterable<String> in) {
+  public static List<CqlIdentifier> wrap(
+      @NonNull Iterable<String> in, @NonNull Function<String, CqlIdentifier> fn) {
+
+    Preconditions.checkNotNull(in, "Input Iterable must not be null");
```
adutra: Not that I care much about it, but we've been leaning towards `Objects.requireNonNull` for such cases.

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
adutra: Our contribution guidelines forbid the usage of Streams API but I often think that this is too excessive. Here, for instance, one line of code could do the job, and it's very readable (and technically speaking, this _is_ a stream transformation):

```
Streams.stream(in).map(fn).collect(ImmutableList.toImmutableList());
```

If I'm not the only one thinking so, we can relax the guidelines a bit for cases like this.

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
```diff
@@ -16,21 +16,36 @@
...
+    Preconditions.checkNotNull(fn, "CqlIdentifier conversion function must not be null");
     ImmutableList.Builder<CqlIdentifier> out = ImmutableList.builder();
     for (String name : in) {
-      out.add(CqlIdentifier.fromCql(name));
+      out.add(fn.apply(name));
     }
     return out.build();
   }
 
+  public static List<CqlIdentifier> wrap(Iterable<String> in) {
```
adutra: I would suggest that you rename this method to `wrapCql`. It's an internal API so we don't care about backwards compatibility.

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
```diff
@@ -16,21 +16,36 @@
...
+    return wrap(in, CqlIdentifier::fromCql);
+  }
+
+  public static List<CqlIdentifier> wrapInternal(Iterable<String> in) {
+    return wrap(in, CqlIdentifier::fromInternal);
+  }
+
   public static <V> Map<CqlIdentifier, V> wrapKeys(Map<String, V> in) {
```
adutra: I suggest you apply the same changes to this method as well:

```
public static <V> Map<CqlIdentifier, V> wrapKeys(
    Map<String, V> in, @NonNull Function<String, CqlIdentifier> fn) {
  Objects.requireNonNull(in, "Input Map must not be null");
  Objects.requireNonNull(fn, "CqlIdentifier conversion function must not be null");
  return in.keySet().stream().collect(ImmutableMap.toImmutableMap(fn, in::get));
}

public static <V> Map<CqlIdentifier, V> wrapKeysCql(Map<String, V> in) {
  return wrapKeys(in, CqlIdentifier::fromCql);
}

public static <V> Map<CqlIdentifier, V> wrapKeysInternal(Map<String, V> in) {
  return wrapKeys(in, CqlIdentifier::fromInternal);
}
```

adutra (dismissed):

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
absurdfarce: Agreed, I'll add those in the next push

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
absurdfarce: That actually was my preference... but I was concerned about the API modification.  Good to know that's more relaxed for the internal stuff (which does make sense).

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
absurdfarce: I actually _completely_ agree with this.  I think a blanket prohibition on usage of the stream API is excessive.  It certainly can be abused, but it's also a very powerful tool in many common cases that we're depriving ourselves of.

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
olim7t: We didn't have the nullability annotations initially. We've added them all over the public API but we yet have to do a pass on the internal packages. When we add new internal methods I agree that it doesn't hurt to have the annotations from the start.

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
olim7t: No strong feelings here, but I think we should still aim for minimal impacts even for internal APIs. Renaming this method adds 4 or 5 query builder classes to the change set, and that's not the object of this ticket.

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
olim7t: I disagree. Examples like this are what I had in mind when I wrote the guidelines.

The goal of this method is to transform a handful of identifiers. Looking at the call sites, it will probably be less than 5 most of the time. It doesn't need "a pipeline of computational operations, functional in nature, laziness-seeking and possibly unbounded".

Readability: not much of a factor because this is a trivial example, any version would be pretty easy to figure out. Yet you can't really beat "for each element, apply the function".

Performance: again it's probably very close, and it's not on the hot path so a few nanoseconds won't make a difference. But conceptually there's something about "convert to a stream, do operations and convert back to a list" that doesn't feel right for 5 elements.

One-liner: yeah it crams everything on one line, is that a good thing? The imperative version is 5 lines, probably doesn't take very long for your brain to parse either. And the typing time is probably very close too, if that even matters.


I'm not against the stream API if we find the need to apply computational steps to a large stream of elements. But I don't get what it buys us over a plain old for loop for trivial operations on small collections.

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
```diff
@@ -16,21 +16,36 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 
 public class CqlIdentifiers {
 
-  public static List<CqlIdentifier> wrap(Iterable<String> in) {
+  public static List<CqlIdentifier> wrap(
+      @NonNull Iterable<String> in, @NonNull Function<String, CqlIdentifier> fn) {
```
olim7t, line 31: You could keep this private for now if it's not used anywhere.

olim7t:
Nit: since JAVA-2492 is DSE specific, can you avoid mentioning it in the final commit message? Something generic like "Add CqlIdentifier.wrapInternal" will be sufficient.

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
absurdfarce: I considered that, but there wasn't a glaring reason to restrict it either... so I just kept it public to maximize API options.

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
absurdfarce: Sure, but those costs have to be paid anytime you decide you want to rename some methods, so it seems like the real question here is whether the proposed names are more descriptive.  And it does seem to me like they are, especially in light of the other changes being made here.

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
absurdfarce: There's clearly not a consensus on any change to the coding guidelines so for now I'm leaving this as-is.  I do very much agree that this is a debate worth having, although this PR is not the place for it... especially given the time-sensitive nature of this ticket. :)

adutra: And I disagree with the disagreement.

absurdfarce:
All: now that we have an immediate fix on the DSE side... are we okay to merge this PR?  I'm still in need of an explicit approval from somebody.

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
olim7t: The question is cost vs benefit.

Cost = breaking 3rd-party code if it happened to use that method. The rules are relaxed for internal packages, but that doesn't mean it's a free-for-all. As a [reminder](https://docs.datastax.com/en/developer/java-driver/4.2/manual/api_conventions/), the rule is:

> backward compatibility is “best-effort” only: we’ll try to preserve it as much as possible, but it’s not formally guaranteed.

In other words, I don't want it to hamper us for changes that are really necessary, but we should still avoid it as much as possible.

As a happy medium, I propose we keep the old method but deprecate it. It's probably overkill for this trivial case, but it's easy to do.

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
olim7t: If there's no reason one way or another, I prefer being conservative. It's easier to expand the visibility later than restrict it.

absurdfarce: I tend to take the opposite approach: exposing more API surface area provides greater flexibility for other developers.  But I'll defer here in the interest of getting this PR wrapped up.

core/src/main/java/com/datastax/oss/driver/internal/core/CqlIdentifiers.java
absurdfarce: I'd argue that a more coherent naming scheme is a pretty reasonable argument for an update under the criteria laid out above... but I don't feel strongly enough about it to go to the mat.  And I'd rather wind up with a clear API for developers to use rather than leave some deprecated methods around.  So I'll defer and remove the rename.

adutra:
I had to kill [yet another build](https://jenkins-drivers.build.dsinternal.org/view/java-driver/job/datastax.java-driver.java2492.default/3/) from this branch. For some reason it seems that Simulacron tests are hanging with the changes from this PR.

absurdfarce:
I reproduced the hanging failures @adutra referenced above locally; looked to be largely driven by the null host ID in Node.toString() stuff fixed in JAVA-2438.  That change had been merged into 4.x but not pulled into this branch yet.  A merge of 4.x appeared to resolve the issue on a local run... pushing to Jenkins now to confirm we're good there too.

adutra:
Looking much better indeed, so let's merge this. I will amend the DSE driver accordingly.

PR #1345, opened 2019-10-15
JAVA-2438: Improve log messages that format Node instances

absurdfarce:


adutra:
I don't have time for a proper review but I'm puzzled. This is not going into the right direction IMO. The goal was to _simplify_ the node's string representation with information that is relevant to the end user. How is a hash code going to be useful to the user?

olim7t:
> How is a hash code going to be useful to the user?

`DefaultNode` uses reference equality (we can't use the host id because it's unknown for contact points at init time, nor the endpoint because we've decided that a node can change its address, see JAVA-2323). You can have two instances with the same endpoint and id that are not equal. Including the hash code is a convenient way to detect this situation.

It's not necessarily useful for the end user, but we as developers also use toString in logs and tests.

adutra:
> It's not necessarily useful for the end user, but we as developers also use toString in logs and tests.

As you wish. I would note however that we are once more favoring developer experience over user experience.

adutra (approved):

PR #1346, opened 2019-10-17
JAVA-2459: Improve extensibility of existing load balancing policies

adutra:
This PR introduces 2 new LBPs:

*  `BasicLoadBalancingPolicy` : acts in a dc-agnostic way if no local dc explicitly provided.
* `DcInferringLoadBalancingPolicy`: infers the local dc from contact points if none explicitly provided.

`BasicLoadBalancingPolicy` is the top-level class. Both `DefaultLoadBalancingPolicy` and `DcInferringLoadBalancingPolicy` inherit from it.

A corresponding PR will be open against the DSE driver.

This PR also contains a (trivial) fix for  JAVA-2447.


jorgebay:
Shouldn't we also provide a `RoundRobinPolicy` / `BasicLoadBalacingPolicy` of some kind that, along with `withNodeFilter()`, would allow tool devs to create query plans with basic whitelisting?

This public class would not be recommended for normal users but be documented as a building block for advanced users.

adutra:
> Shouldn't we also provide a `RoundRobinPolicy` / `BasicLoadBalacingPolicy` of some kind that, along with `withNodeFilter()`, would allow tool devs to create query plans with basic whitelisting?
> 
> This public class would not be recommended for normal users but be documented as a building block for advanced users.

Isn't this covered by `LoadBalancingPolicyBase`? Would you like me to rename it to `BasicLoadBalacingPolicy` and make it concrete? Or do you have some other hierarchy in mind?

jorgebay:
I think we should expose a concrete class for this advanced users (apart from the base class). This concrete class should not validate or obtain the local data center parameter in any way, providing the same functionality as `HostFilterPolicy` in driver 3.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/BasicLoadBalancingPolicy.java
```diff
@@ -0,0 +1,370 @@
...
+ * A basic implementation of {@link LoadBalancingPolicy} that can serve as a building block for more
+ * advanced use cases.
+ *
+ * <p><b>Local datacenter inference</b>: This implementation will only infer the local datacenter if
+ * it is explicitly set either through configuration or programmatically; if the local datacenter is
+ * unspecified, this implementation will effectively act as a datacenter-agnostic load balancing
+ * policy and will consider all nodes in the cluster when creating query plans, regardless of their
+ * datacenter.
+ *
+ * <p>This class is not recommended for normal users who should always prefer {@link
```
jorgebay: +100

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/BasicLoadBalancingPolicy.java
```diff
@@ -0,0 +1,370 @@
...
+import java.util.function.Predicate;
+import net.jcip.annotations.ThreadSafe;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A basic implementation of {@link LoadBalancingPolicy} that can serve as a building block for more
+ * advanced use cases.
+ *
+ * <p><b>Local datacenter inference</b>: This implementation will only infer the local datacenter if
```
jorgebay: We could avoid talking about "inference", like:

```
<p><b>Local data center</b>: This implementation will use the local data center set
explicitly through configuration or programmatically ...
```

Additionally, we could document how it behaves: prioritizes replica nodes (shuffling) and round-robin through the rest of the nodes.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -71,255 +44,61 @@
...
-      // node previously.
-      distanceReporter.setDistance(node, NodeDistance.LOCAL);
-      if (localDcLiveNodes.add(node)) {
-        LOG.debug("[{}] {} came back UP, added to live set", logPrefix, node);
+                + " in the config, or set it programmatically with SessionBuilder.withLocalDatacenter)");
       }
     } else {
-      distanceReporter.setDistance(node, NodeDistance.IGNORED);
+      throw new IllegalStateException(
+          "You provided explicit contact points, the local DC must be explicitly set (see "
```
jorgebay: Can we provide the data center names (based on nodes metadata) or will that be tackled in a different ticket?

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/BasicLoadBalancingPolicy.java
```diff
@@ -0,0 +1,371 @@
...
+   *
+   * <p>This implementation fetches the user-supplied datacenter, if any, from the programmatic
+   * configuration API, or else, from the driver configuration. If no user-supplied datacenter can
+   * be retrieved, this implementation returns empty, effectively causing this load balancing policy
+   * to operate in a datacenter-agnostic fashion.
+   *
+   * @return The local datacenter, or empty if none found.
+   */
+  @NonNull
+  protected Optional<String> inferLocalDatacenter() {
```
jorgebay: NIT: we can avoid referring it as `infer` (as it will use the one provided by the user by default), maybe something like `calculate` / `obtain` / `fetch`

jorgebay (commented):
Looks good, I like the approach of a `BasicLoadBalancingPolicy` that provides both token-awareness and filtering capabilities.

I've included few comments below.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
adutra: I suggest that we do this within this PR since it's all entangled in the same classes.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/BasicLoadBalancingPolicy.java
adutra: Calculate is misleading too... How about `discoverLocalDatacenter`?

jorgebay: I like it!

absurdfarce:
FWIW if there is a concern about the use of inheritance here you can also do this via composition:

* Implement a DelegatingLoadBalancingPolicy which routes all methods of LoadBalancingPolicy to another impl
* DcInferring and Default policies use this to defer most methods to a contained BasicLoadBalancingPolicy
* DcInferring and Default policies wind up only needing to override the LocalDcHelper methods (as they do now)
    * Might need to expose a few getters in order to allow DcInferring and Default policies (which are now no longer subclasses) to access driver execution profile + log prefix

Not required by any means.... I mention it only because the question came up in standup

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/LocalDcHelper.java
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.internal.core.loadbalancing.helper;
+
+import java.util.Optional;
+
+public interface LocalDcHelper {
```
absurdfarce, line 28: This interface is a functional interface and probably should be annotated as such

absurdfarce, line 28: Actually this interface is fundamentally equivalent to a Supplier (or, even more generally, a Function).  Might be worth considering changing the impl of newLocalDcHelper() to return a Supplier and move the impl of the concrete classes into lambdas within the various LBP classes.  The advantage is that the entire logic for a given LBP resides within the LBP class rather than being a combination of LBP class + local DC helper class.

Again, not required by any means... just something to consider.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/BasicLocalDcHelper.java
```diff
@@ -0,0 +1,115 @@
...
+   */
+  protected void checkLocalDatacenterCompatibility(
+      @NonNull String localDc, Set<? extends Node> contactPoints) {
+    Set<Node> badContactPoints = new LinkedHashSet<>();
+    for (Node node : contactPoints) {
+      if (!Objects.equals(localDc, node.getDatacenter())) {
+        badContactPoints.add(node);
+      }
+    }
+    if (!badContactPoints.isEmpty()) {
```
jorgebay: I think we should first validate that the local dc name is valid and in case it doesn't match, error out with a message like: `"Local data center 'my_dc' was provided but only found hosts in datacenters 'dc1' and 'dc2'"`

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/DefaultLocalDcHelper.java
```diff
@@ -0,0 +1,90 @@
...
+            "The local DC could not be inferred from implicit contact point, please set it explicitly (see "
+                + DefaultDriverOption.LOAD_BALANCING_LOCAL_DATACENTER.getPath()
+                + " in the config, or set it programmatically with SessionBuilder.withLocalDatacenter)");
+      }
+    } else {
+      throw new IllegalStateException(
+          "Since you provided explicit contact points, the local DC must be explicitly set (see "
+              + DefaultDriverOption.LOAD_BALANCING_LOCAL_DATACENTER.getPath()
+              + " in the config, or set it programmatically with SessionBuilder.withLocalDatacenter). "
+              + "Current contact points are: "
```
jorgebay: We should also include the available DCs.

jorgebay (commented):
It's looking great, I've added a few notes below about surfacing the DC names in the error messages.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/BasicLoadBalancingPolicy.java
```diff
@@ -0,0 +1,287 @@
...
+  }
+
+  @Override
+  public void init(@NonNull Map<UUID, Node> nodes, @NonNull DistanceReporter distanceReporter) {
+    this.distanceReporter = distanceReporter;
+    localDc = newLocalDcHelper().discoverLocalDc().orElse(null);
+    filter = newNodeFilterHelper().createNodeFilter(localDc);
+    for (Node node : nodes.values()) {
+      if (filter.test(node)) {
+        distanceReporter.setDistance(node, NodeDistance.LOCAL);
```
tomekl007, line 130: Is it safe to set distance to `LOCAL`even if the node filter works for BasicStrategy that doesn't require dc?

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/InferringLocalDcHelper.java
```diff
@@ -0,0 +1,84 @@
...
+  public Optional<String> discoverLocalDc() {
+    Optional<String> optionalLocalDc = super.discoverLocalDc();
+    if (optionalLocalDc.isPresent()) {
+      return optionalLocalDc;
+    }
+    Set<String> datacenters = new HashSet<>();
+    Set<DefaultNode> contactPoints = context.getMetadataManager().getContactPoints();
+    for (Node node : contactPoints) {
+      String datacenter = node.getDatacenter();
+      if (datacenter != null) {
```
tomekl007, line 63: in what case the `node.getDatacenter()` can be `null`?

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/DefaultLocalDcHelper.java
```diff
@@ -0,0 +1,90 @@
...
+    Optional<String> optionalLocalDc = super.discoverLocalDc();
+    if (optionalLocalDc.isPresent()) {
+      return optionalLocalDc;
+    }
+    Set<DefaultNode> contactPoints = context.getMetadataManager().getContactPoints();
+    if (context.getMetadataManager().wasImplicitContactPoint()) {
+      // We only allow automatic inference of the local DC in this specific case
+      assert contactPoints.size() == 1;
+      Node contactPoint = contactPoints.iterator().next();
+      String localDc = contactPoint.getDatacenter();
```
tomekl007: how this edge case differ from `BasicLocalDcHelper`?  I remember from the discussion that `BasicLocalDcHelper` is used for the studio when they have only one contact point and they don't want to specify DC. For me, it looks that exactly this logic is covered in this `if`

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/LocalDcHelper.java
adutra: I didn't because it's not used as such, but OK to add it.

tomekl007: +1

tomekl007: From my experience, it is better to "name" the Supplier to give it a specific type.
So if the constructor is taking LocalDcHelper (that happens to be supplier) it is better to have the strong type to allow clients to pass only `LocalDcHelper` instead of wider type (`Supplier`). It is a similar case to having the Name of a person that is a `String` as well. For strong typing, it is better to create an explicit data class name as `class Name{}` instead of having String API.

adutra: > Might be worth considering changing the impl of newLocalDcHelper() to return a Supplier and move the impl of the concrete classes into lambdas within the various LBP classes. 

This would incur in some code duplication, e.g. if a grand-child class wants to expose the same behavior as its grand-parent, but the parent has a different behavior. This actually happens e.g. for the DSE LBPs.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/DefaultLocalDcHelper.java
jorgebay: oh man... I see that `formatNodes()` already include the DC name of each contact point, thanks Alex for pointing that out.

nvm!

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/BasicLocalDcHelper.java
jorgebay: `formatNodes()` already includes the DC name, nvm!

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/BasicLoadBalancingPolicy.java
adutra: I think so. At least this mimics the behavior of the old driver `com.datastax.driver.core.policies.RoundRobinPolicy`, which always returns `LOCAL` for all nodes in the cluster.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/InferringLocalDcHelper.java
adutra: Only in pathological cases, e.g. when `system.peers` contains corrupted data.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/DefaultLocalDcHelper.java
adutra: This LBP only infers the DC if no contact points were provided. This is never the case for Studio, where you need to specify at least (or maybe exactly, I don't know) one contact point.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/LocalDcHelper.java
adutra: > From my experience, it is better to "name" the Supplier to give it a specific type.

Agreed, and that also allows for better documentation of method contracts.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/LocalDcHelper.java
```diff
@@ -0,0 +1,47 @@
...
+   * otherwise.
+   *
+   * <p>Implementors may choose to throw {@link IllegalStateException} instead of returning {@link
+   * Optional#empty empty}, if they require a local datacenter to be defined in order to operate
+   * properly.
+   *
+   * @param nodes All the nodes that were known to exist in the cluster (regardless of their state)
+   *     when the load balancing policy was {@linkplain
+   *     com.datastax.oss.driver.api.core.loadbalancing.LoadBalancingPolicy#init(Map,
+   *     com.datastax.oss.driver.api.core.loadbalancing.LoadBalancingPolicy.DistanceReporter)
```
olim7t: Maybe import the class to make this easier to read in the IDE.
... or remove the `helper` package altogether, it's not really necessary.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/BasicLocalDcHelper.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class BasicLocalDcHelper implements LocalDcHelper {
```
olim7t: Why isn't the basic helper simply DC-agnostic? If the client is able to provide a DC, then they might as well use the "default" behavior.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/InferringLocalDcHelper.java
```diff
@@ -0,0 +1,86 @@
...
+   * @return The local datacenter; always present.
+   * @throws IllegalStateException if the local datacenter could not be inferred.
+   */
+  @NonNull
+  @Override
+  public Optional<String> discoverLocalDc(@NonNull Map<UUID, Node> nodes) {
+    Optional<String> optionalLocalDc = super.discoverLocalDc(nodes);
+    if (optionalLocalDc.isPresent()) {
+      return optionalLocalDc;
+    }
```
olim7t, line 58: Same remark as the basic helper: why honor the config/programmatic localDc at all here? If anything, it's more confusing than always inferring.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/DefaultLocalDcHelper.java
```diff
@@ -0,0 +1,94 @@
...
+import com.datastax.oss.driver.internal.core.metadata.DefaultNode;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DefaultLocalDcHelper extends BasicLocalDcHelper {
```
olim7t: I find the terminology a bit confusing, although I'm not sure I have better names to suggest.
* "inferring": that one is clear
* "basic": does not require the local DC, except if it's provided (why?). If it completely ignored it maybe it could be `NoLocalDcHelper`.
* "default": maybe `MandatoryLocalDcHelper`

olim7t (approved):
Technically this looks good to me.

Conceptually I don't like it one bit. We're effectively introducing a framework to write load balancing policies. Even though this is internal, this can and will be used by 3rd-party implementations. I can already imagine the incoming feature requests (changing the local DC at runtime, etc).
We've discussed this already so I don't want to drag this PR any longer.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/LocalDcHelper.java
absurdfarce: I'm far from attached to the idea... I probably wouldn't even have mentioned it if there hadn't been an earlier conversation about inheritance.  And in this case it's trickier since there is explicit inheritance in the helper hierarchy as well as the LBP policy directly.

That said, this could be worked around.  The existing impls largely do some additional processing if the base impl couldn't determine a DC; these could be easily implemented via fn declarations which include the originally computed value as an arg.

I understand the point about preferring named types.  But in this case the differences in the subtypes are relatively small... and you're only modifying the behaviour of another function on the interface in any meaningful ways.  It seems like using `Function<Map<UUID, Node>, Optional<String>>` to represent that functionality is a more natural (to say nothing of more obvious) way to represent the interaction going on here than an explicit type hierarchy.

Regardless, what's there now definitely does work so I'll not belabor the point. :)

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DcInferringLoadBalancingPolicy.java
```diff
@@ -0,0 +1,77 @@
+/*
```
absurdfarce, line 1: Do we really need an explicit type hierarchy for the different policies here?  The only thing the subtypes do is provide a different helper impl (which is used in a single function at initialization time within the policy).  Seems like the best answer is to just pass the helper as an arg to init(), but I assume we can't do that due to API breakage.  If that's the case can we just pass the helper as an arg to the constructor?  Are we not doing this because it... would also change the contract, this time with the configuration?

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/BasicLocalDcHelper.java
adutra: But in this case which helper should the "basic" LBP use? I agree that when the DC is provided explicitly the "basic" LBP behaves exactly like the "default" one; but when the DC is _not_ provided however, their behaviors are different. We need a helper that allows that (use the DC if set explicitly, otherwise act in a DC-agnostic way).

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/InferringLocalDcHelper.java
adutra: I think the idea was to be flexible: the same LBP can be used in situations where the DC is known, and set explicitly, or when it is unknown, and discovered dynamically. I agree that we could introduce 2 distinct LBPs: one that always gathers the DC from its config, another that always infers it from the contact points. But that would require a given user to think about whether or not the DC will be known in advance before choosing the right policy, which seemed to me a bit convoluted. @jorgebay would you feel more comfortable with 2 distinct LBPs here?

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/helper/DefaultLocalDcHelper.java
adutra: Ah names :-) `MandatoryLocalDcHelper` works for me. `NoLocalDcHelper` annoys me however because it doesn't always return an empty optional. Would `OptionalLocalDcHelper` work for you?

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DcInferringLoadBalancingPolicy.java
adutra: > The only thing the subtypes do is provide a different helper impl (which is used in a single function at initialization time within the policy)

Not exactly, the type hierarchy is also used to modify the query plan computation logic from OSS (simple replica selection) to DSE (busy node avoidance). And at least in theory, the node filter logic could be customized too. 

> Seems like the best answer is to just pass the helper as an arg to init()

I avoided that because it requires creating a secondary constructor that accepts the helpers as arguments. I'm not sure how well this plays with the reflection mechanism in the driver that instantiates the components by reflection.

adutra: You are however right about something: it would be a bit better to hide the helper classes from the internal API in the `BasicLoadBalancingPolicy` class hierarchy (concretely speaking, remove references to the helpers from protected methods). I can do that quickly.

absurdfarce: Sorry, I should've been clearer.  The only thing these subtypes do is offer a different helper implementation.  Other subtypes are free to modify other behaviours as they wish.  And in this case the DSE impl isn't actually a problem since it's not part of this type hierarchy; it just implements the LBP interface directly does it not?

Regardless, I definitely like the idea of removing the helper classes as an explicit part of the API; that addresses a large part of my concerns above.  But I can't help but wonder... now that we don't actually expose the helper classes why have them as separate classes at all?  Why not just make, say, InferringLocalDcHelper.discoverLocalDc() be the impl of DcInferringLoadBalancingPolicy.discoverLocalDc()?

adutra: > And in this case the DSE impl isn't actually a problem since it's not part of this type hierarchy; it just implements the LBP interface directly does it not?

No, the DSE LBPs both inherit from `DefaultLoadBalancingPolicy`.

> Why not just make, say, InferringLocalDcHelper.discoverLocalDc() be the impl of DcInferringLoadBalancingPolicy.discoverLocalDc()

Because we also have a `DseDcInferringLoadBalancingPolicy` that inherits its query plan logic from `DseLoadBalancingPolicy`, but exhibits a dc inference logic equal to the one exposed by `DcInferringLoadBalancingPolicy`, which is not a parent superclass. If the dc inference logic was baked into `DcInferringLoadBalancingPolicy`, it would have to be copied into `DseDcInferringLoadBalancingPolicy`.

jorgebay (approved):
Makes sense for me to have some internal building blocks for tool developers, regular users should continue to use the default load balancing policy.

Hopefully, new error messages will help them understand when local dc is incorrectly configured.

PR #1347, opened 2019-10-18
JAVA-2434: Add support for custom cipher suites and host name validation to ProgrammaticSslEngineFactory

adutra:


core/src/main/java/com/datastax/oss/driver/api/core/ssl/ProgrammaticSslEngineFactory.java
```diff
@@ -57,11 +94,19 @@ public SSLEngine newSslEngine(@NonNull EndPoint remoteEndpoint) {
       engine = sslContext.createSSLEngine();
     }
     engine.setUseClientMode(true);
+    if (cipherSuites != null) {
+      engine.setEnabledCipherSuites(cipherSuites);
+    }
+    if (requireHostnameValidation) {
+      SSLParameters parameters = engine.getSSLParameters();
```
tomekl007, line 102: does SSLParameters are always `nonNull`?

core/src/main/java/com/datastax/oss/driver/api/core/ssl/ProgrammaticSslEngineFactory.java
```diff
@@ -40,9 +41,45 @@
...
+      @NonNull SSLContext sslContext, @Nullable String[] cipherSuites) {
+    this(sslContext, cipherSuites, false);
+  }
+
+  /**
+   * Creates an instance with the given {@link SSLContext}, cipher suites and host name validation.
+   *
+   * @param sslContext the {@link SSLContext} to use.
+   * @param cipherSuites the cipher suites to use, or null to use the default ones.
+   * @param requireHostnameValidation whether to enable host name validation.
```
tomekl007: Maybe add info in docs that when hostname validation is true then SSLParameters will be taken from 
`SSLParameters parameters = engine.getSSLParameters();`?

core/revapi.json
```diff
@@ -4830,6 +4830,20 @@
         "new": "method java.util.UUID com.datastax.oss.driver.api.core.metadata.Node::getHostId()",
         "annotation": "@edu.umd.cs.findbugs.annotations.Nullable",
         "justification": "JAVA-2505: Annotate Node.getHostId() as nullable"
+      },
+      {
+        "code": "java.annotation.added",
+        "old": "parameter void com.datastax.oss.driver.api.core.ssl.ProgrammaticSslEngineFactory::<init>(===javax.net.ssl.SSLContext===)",
+        "new": "parameter void com.datastax.oss.driver.api.core.ssl.ProgrammaticSslEngineFactory::<init>(===javax.net.ssl.SSLContext===)",
+        "annotation": "@edu.umd.cs.findbugs.annotations.NonNull",
+        "justification": "JAVA-2434: added @NonNull to ProgrammaticSslEngineFactory(SSLContext) constructor"
```
tomekl007, line 4839: can we introduce not backward-compatible changes?

tomekl007 (approved):

core/src/main/java/com/datastax/oss/driver/api/core/ssl/ProgrammaticSslEngineFactory.java
adutra: Yes it cannot be null.

core/src/main/java/com/datastax/oss/driver/api/core/ssl/ProgrammaticSslEngineFactory.java
adutra: I think this is the only way, so not sure it's worth mentioning it. However I can mention that the verification algorithm is HTTPS.

core/revapi.json
adutra: In this case I need to remove the `@NonNull` annotation. But honestly adding/removing annotations doesn't really break users code.

adutra: Hmm you are right for the `throws Exception` clause, this could break user code. I will remove it.

PR #1348, opened 2019-10-18
JAVA-2497: Ensure nodes and exceptions are serializable

olim7t:


integration-tests/src/test/java/com/datastax/oss/driver/core/SerializationIT.java
```diff
@@ -0,0 +1,142 @@
...
+
+    // Then
+    ColumnDefinition columnDefinition = row.getColumnDefinitions().get("t");
+    assertThat(columnDefinition.getType()).isEqualTo(DataTypes.TEXT);
+    assertThat(row.getString("t")).isEqualTo("mock data");
+  }
+
+  // Serialization methods copied from SerializationHelper in core tests.
+
+  private static <T> byte[] serialize(T t) {
```
adutra: It's the 3rd time already that we are copying-pasting this code: I see one `SerializationHelper` in native-protocol, another one in OSS core. Why don't you re-use the one from OSS core? The test-jar is already being produced, it's just a matter of adding a dependency.

adutra (changes requested):

integration-tests/src/test/java/com/datastax/oss/driver/core/SerializationIT.java
olim7t: I got ahead of myself a little bit on this one: I think when the two drivers get merged, we should stop deploying a `core` test JAR. Maven central is a community resource and we should use as little of it as possible. Publishing a test JAR for every release just to factor a few lines of trivial code is not a good reason IMHO.

But right now we have to publish it anyway because it's used in the DSE driver, which reuses more than a few trivial things I think. I'll add the dependency as you suggest, and we can discuss this again in the driver merge PR.

PR #1349, opened 2019-10-18
JAVA-2480: Upgrade Jackson to 2.10.0

emerkle826:


examples/src/main/java/com/datastax/oss/driver/examples/json/codecs/JacksonJsonCodec.java
```diff
@@ -52,28 +52,28 @@
...
 
   /**
    * Creates a new instance for the provided {@code javaClass}, and using the provided {@link
-   * ObjectMapper}.
+   * JsonMapper}.
    *
    * @param javaClass the Java class this codec maps to.
    */
-  public JacksonJsonCodec(Class<T> javaClass, ObjectMapper objectMapper) {
+  public JacksonJsonCodec(Class<T> javaClass, JsonMapper jsonMapper) {
```
emerkle826, line 74: I know this is a public method signature change, but it is in the `examples` project, so I figured it's not a big deal.

pom.xml
```diff
@@ -69,8 +69,7 @@
     <hk2.version>2.5.0</hk2.version>
     <jax-rs.version>2.0.1</jax-rs.version>
     <jsr353-ri.version>1.1.4</jsr353-ri.version>
-    <jackson.version>2.9.9</jackson.version>
-    <jackson-databind.version>2.9.9.3</jackson-databind.version>
```
absurdfarce, line 73: It _looks like_ this property is referenced specifically in com.datastax.oss.driver.osgi.BundleOptions::jacksonBundles()... is that a problem?

emerkle826: good catch, I thought I caught all the references.

emerkle826: @adutra It looks like we separated the Jackson version from the Jackson databind version with this commit:
https://github.com/datastax/java-driver/commit/f6980df6e065bca4ed2db28297d56e8f55c8d514
Do you recall why? And do you think it's ok to upgrade to 2.10.0 for both?

adutra: Yes because versions for these two artifacts started diverging a while ago.

adutra: I think it might be easier to just keep declaring these 2 properties, even if they have the same value.

adutra:
Pushed a small fix for the databind property and confirmed that the OSGi tests pass.

adutra (approved):
I will open an equivalent PR for the DSE driver.

PR #1351, opened 2019-10-22
JAVA-2476: Improve error message when codec registry inspects a collection with a null element

olim7t:
Scheduled for 4.3.0 because this is a trivial ticket, but it can also wait for the next release.


core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistryTest.java
```diff
@@ -672,6 +701,26 @@ public void should_create_map_codec_for_java_value_when_first_element_is_a_subty
         .lookup(null, GenericType.mapOf(Inet4Address.class, Inet4Address.class), true);
   }
 
+  @Test
+  public void should_throw_for_map_codec_containing_null_element() {
```
adutra, line 705: Nit: a parameterized test would be more elegant.

adutra (approved):

PR #1352, opened 2019-10-22
JAVA-2509: Mention file-based approach for Cloud configuration in the manual

adutra:


manual/cloud/README.md
```diff
@@ -110,7 +110,63 @@ database.
...
-    c. Save and close the ConnectDatabase.java file.
+    c. **File-based configuration**. An alternative to the programmatic configuration method 
+    detailed above is to include the information required to connect in the driver's configuration 
+    file (`application.conf`). Merge the following options with any other options that you might 
+    want to include in the configuration file:
+    
+    ```hocon
+   basic {
+     # change this to match the target keyspace
+     session-keyspace = keyspace_name
```
tomekl007, line 121: does it need to be specified? I think it will be taken from `secure-connect-bundle`

adutra: We don't parse the keyspace anymore. We still parse the credentials though, even if they are now always absent.

tomekl007: Ok, but even in that case, the session-keyspace is not required, right?
The client can specify keyspace later. I am not sure if we should mention that in the docs

manual/cloud/README.md
```diff
@@ -121,3 +177,4 @@ database.
 [Download the secure connect bundle - GCP]: https://helpdocs.datastax.com/gcp/dscloud/apollo/dscloudObtainingCredentials.html
 [Download the secure connect bundle - AWS]: https://helpdocs.datastax.com/aws/dscloud/apollo/dscloudObtainingCredentials.html
 [Example pom.xml file]: ../core/integration/#minimal-project-structure
+[driver documentation]: https://docs.datastax.com/en/developer/java-driver/4.3/manual/core/configuration/
```
emerkle826: this link can be simplified to:
```
[driver documentation]: ../core/configuration/
```

emerkle826 (changes requested):

manual/cloud/README.md
adutra: I am not sure about the keyspace actually. It was said at some point that users would have access to just one keyspace. Not sure if this is valid still. But it doesn't hurt to specify a keyspace anyways, so I would leave it that way. @emerkle826 do you know by chance if a keyspace is required for cloud?

emerkle826: The last I was aware, you don't need to specify a keyspace, but I believe you are correct in that they will only have access to one keyspace (for now at least). From Jorge:
>I think there was consensus for not parsing the keyspace: users might be able to create their own keyspace in future versions of cloud and it will not be provided when switching to token-based auth

emerkle826 (approved):

manual/cloud/README.md
adutra: In this case let's leave the keyspace example, it seems that at least for the initial Cloud version it will be necessary to set it since users will only have access to that specific keyspace.

tomekl007 (approved):

PR #1353, opened 2019-10-24
JAVA-2460: Document how to determine the local DC

adutra:


tomekl007 (approved):

jorgebay:
Looks great, I would take the opportunity to remove references about no alternatives to the default, e.g., "No out-of-the-box alternatives, but you can write your own." from the doc.

adutra:
> Looks great, I would take the opportunity to remove references about no alternatives to the default, e.g., "No out-of-the-box alternatives, but you can write your own." from the doc.

Do you think we should mention the other ones (basic and dc-inferring)? I thought we would keep them "hidden"...

jorgebay:
yes, its ok not to mention them, but saying "No out-of-the-box alternatives" can be misleading.

In that case, removing the bullet point would be better IMO

jorgebay (approved):
+1

PR #1354, opened 2019-10-24
JAVA-2515: NEW_NODE and REMOVED_NODE events should trigger ADDED and REMOVED

adutra:


driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -2735,9 +2735,19 @@ public void handle(Message.Response response) {
         case TOPOLOGY_CHANGE:
           ProtocolEvent.TopologyChange tpc = (ProtocolEvent.TopologyChange) event;
           Host.statesLogger.debug("[{}] received event {}", tpc.node, tpc.change);
-          // Always do a full refresh for topology changes. This is simpler than trying to infer a
-          // new EndPoint from the rpc_address only.
-          submitNodeListRefresh();
+          // Do NOT translate the address, it will be matched against Host.getBroadcastRpcAddress()
+          // to find the target host.
+          switch (tpc.change) {
```
adutra, line 2740: This is the logic pre-cloud API.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -3070,6 +3080,8 @@ public String toString() {
             case UP:
               Host upHost = metadata.getHost(address);
               if (upHost == null) {
+                // We don't have enough information to create a new Host (we are missing it's ID)
+                // so trigger a full node refresh
                 submitNodeListRefresh();
```
adutra, line 3084: This logic changed with the Cloud API, previously we would create a new Host here calling `Metadata.newHost()` but now it's not possible anymore because we don't have the host ID.

driver-core/src/test/java/com/datastax/driver/core/StateListenerTest.java
```diff
@@ -52,7 +52,7 @@ public void should_receive_events_when_node_states_change() throws InterruptedEx
     ccm().start(1);
     listener.waitForEvent();
 
-    listener.setExpectedEvent(REMOVE);
+    listener.setExpectedEvent(DOWN);
```
adutra: The event received is now DOWN because of the changes in the NodeRefreshRequestDeliveryCallback logic. I don't know how to fix it.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
olim7t: If we do this, we should also re-add the branches for `NEW_NODE` and `REMOVED_NODE` in `NodeRefreshRequestDeliveryCallback`, see [this](https://github.com/datastax/java-driver/blob/a815196b81a6180d555b60b08bcbec2e87673fff/driver-core/src/main/java/com/datastax/driver/core/Cluster.java#L2924). Currently `NodeRefreshDebouncerTest` passes because it gets an UP event for an unknown node and turns it into a full list refresh; the addition is silently ignored.

But then we'll run into the issue that we can't create a new node for ADDED because we don't have the host id (like in the UP case where you added a comment below). This is why I changed to a full refresh in the first place.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -2735,9 +2735,19 @@ public void handle(Message.Response response) {
...
+          // Do NOT translate the address, it will be matched against Host.getBroadcastRpcAddress()
+          // to find the target host.
+          switch (tpc.change) {
+            case NEW_NODE:
+              submitNodeRefresh(tpc.node, HostEvent.ADDED);
+              break;
+            case REMOVED_NODE:
+              submitNodeRefresh(tpc.node, HostEvent.REMOVED);
+              break;
+            case MOVED_NODE:
```
olim7t: Side note: I'm not sure where this MOVED_NODE event comes from, it's not referenced anywhere in the protocol spec.
It's been in the codebase for a long time, maybe it used to be a thing.

Driver 4 does not have it (nor do I think it should).

edit - with my change it will get covered by the `default` case in the switch.

olim7t (changes requested):
So yeah as I said in the comments I think it will be hard to completely reestablish the old behavior for `onAdd`, because we don't have a host id to create the node. So I propose we keep a full refresh for that case, and relax the conditions in the test. Unfortunately full refreshes are not processed by the same debouncer as individual node events, so we can't have a single event in the test; I think this is acceptable as long as we get an ADDED notification and the host eventually goes up. 

For REMOVED_NODE you are right that we can keep the old behavior, and process it as an individual event. This fixes `StateListenerTest`.

I added a commit with those two suggestions.

driver-core/src/main/java/com/datastax/driver/core/Cluster.java
```diff
@@ -3096,6 +3107,10 @@ public String toString() {
                 }
               }
               break;
+            case REMOVED:
```
adutra, line 3110: OK I see you added this case branch that was somehow missing since the first Cloud API. But I also see that a case branch for `ADDED` existed before the Cloud API:

```
case ADDED:
              Host newHost = metadata.newHost(address);
              Host previous = metadata.addIfAbsent(newHost);
              if (previous == null) {
                futures.add(schedule(hostAdded(newHost)));
              } else if (!previous.isUp()) {
                futures.add(schedule(hostUp(previous)));
              }
              break;
```

Shouldn't we trigger a `submitNodeListRefresh()` for it? Or is it because `ADDED` is not being triggered anymore? (But in this case, maybe we can delete the `HostEvent.ADDED` enum constant.)

olim7t: > is it because ADDED is not being triggered anymore?

Yes, that.

You're right, I'll remove the unused constant.

olim7t (approved):

PR #1355, opened 2019-10-25
JAVA-2516: Enable hostname validation with Cloud

olim7t:
We'll also need to backport this to the legacy branches.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/ssl/SniSslEngineFactory.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/ssl/SniSslEngineFactory.java
@@ -37,7 +37,6 @@ public class SniSslEngineFactory implements SslEngineFactory {
   private final SSLContext sslContext;
   private final CopyOnWriteArrayList<String> fakePorts = new CopyOnWriteArrayList<>();
 
-  /** Builds a new instance from the driver configuration. */
   public SniSslEngineFactory(SSLContext sslContext) {
     this.sslContext = sslContext;
   }
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/ssl/SniSslEngineFactory.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/ssl/SniSslEngineFactory.java
@@ -21,6 +21,8 @@ import com.datastax.oss.driver.api.core.ssl.SslEngineFactory;
 import com.datastax.oss.driver.internal.core.metadata.SniEndPoint;
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.net.InetSocketAddress;
+import java.util.concurrent.CopyOnWriteArrayList;
 import javax.net.ssl.SNIHostName;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLEngine;
@@ -28,7 +30,12 @@ import javax.net.ssl.SSLParameters;
 
 public class SniSslEngineFactory implements SslEngineFactory {
 
+  // An offset that gets added to our "fake" ports (see below). We pick this value because it is the
+  // start of the ephemeral port range.
+  private static final int FAKE_PORT_OFFSET = 49152;
+
   private final SSLContext sslContext;
+  private final CopyOnWriteArrayList<String> fakePorts = new CopyOnWriteArrayList<>();
 
   /** Builds a new instance from the driver configuration. */
   public SniSslEngineFactory(SSLContext sslContext) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/ssl/SniSslEngineFactory.java
```diff
@@ -44,17 +46,40 @@ public SSLEngine newSslEngine(@NonNull EndPoint remoteEndpoint) {
...
+    //
+    // In our case, the first check will never succeed because our SNI host name is not the DNS name
+    // (we use the Cassandra host_id instead). So we *must* set the advisory peer information.
+    //
+    // However if we use the address as-is, this leads to another issue: the advisory peer
+    // information is also used to cache SSL sessions internally. All of our nodes share the same
+    // proxy address, so the JDK tries to reuse SSL sessions across nodes. But it doesn't update the
+    // SNI host name every time, so it ends up opening connections to the wrong node.
+    //
+    // To avoid that, we create a unique "fake" port for every node. We still get session reuse for
```
adutra, line 69: Very clever 👍 but how is session caching going to work in spite of this? As I see it, since the advisory port is fake, when the JDK will lookup for an existing session, it will use the real port reported by the address and won't find any cached session for that host+port combination.

core/src/main/java/com/datastax/oss/driver/internal/core/ssl/SniSslEngineFactory.java
```diff
@@ -44,17 +46,40 @@ public SSLEngine newSslEngine(@NonNull EndPoint remoteEndpoint) {
...
-    parameters.setServerNames(ImmutableList.of(new SNIHostName(sniEndPoint.getServerName())));
+    parameters.setServerNames(ImmutableList.of(new SNIHostName(sniServerName)));
+    parameters.setEndpointIdentificationAlgorithm("HTTPS");
     engine.setSSLParameters(parameters);
     return engine;
   }
 
+  private int getFakePort(String sniServerName) {
+    fakePorts.addIfAbsent(sniServerName);
+    return fakePorts.indexOf(sniServerName);
```
adutra: The fake port's actual number doesn't seem to make any difference, so instead of storing host names in a list, why don't you use an `AtomicInteger`, preferably starting at the beginning of the ephemeral port range (49152)? What makes me a bit uncomfortable is that zero has a special meaning when used as a port, and low numbers also usually mean reserved ports. You stated that the number is really only used for session caching, but who knows if this won't change in future JDK versions.

adutra: OK I saw your comments regarding using AtomicInteger. Please disregard this comment. I only hope using zero as a port number won't bite us back one day.

tomekl007: @adutra could you tell me what was the reason?

core/src/main/java/com/datastax/oss/driver/internal/core/ssl/SniSslEngineFactory.java
```diff
@@ -44,17 +46,40 @@ public SSLEngine newSslEngine(@NonNull EndPoint remoteEndpoint) {
               "Configuration error: can only use %s with SNI end points",
               this.getClass().getSimpleName()));
     }
-    SSLEngine engine;
     SniEndPoint sniEndPoint = (SniEndPoint) remoteEndpoint;
+    InetSocketAddress address = sniEndPoint.resolve();
+    String sniServerName = sniEndPoint.getServerName();
 
-    engine = sslContext.createSSLEngine();
+    // When hostname verification is enabled (with setEndpointIdentificationAlgorithm), the SSL
```
tomekl007, line 57: The thing that worries me is that we don't have any test that captures that. If somewhen in a future this method 
https://github.com/JetBrains/jdk8u_jdk/blob/master/src/share/classes/sun/security/ssl/X509TrustManagerImpl.java#L401-L445 will change we may not notice that port is taken into account. I think we should cover at least happy path

core/src/main/java/com/datastax/oss/driver/internal/core/ssl/SniSslEngineFactory.java
```diff
@@ -29,8 +31,8 @@
 public class SniSslEngineFactory implements SslEngineFactory {
```
tomekl007, line 31: One question to make it clear. This `SniSslEngineFactory` is created only once per lifetime of the application, right?
Or is it created once per session? If once per session then where is that reusability of `fakePorts`?
And `newSslEngine()` is called only at the start of the app as many times as there are endpoints in the cloud that we are connecting to, yes?
Once it will be done there are no more `newSslEngine` calls?

core/src/main/java/com/datastax/oss/driver/internal/core/ssl/SniSslEngineFactory.java
olim7t: For a given host_id, the fake port is always the same. So we will reuse SSL sessions for connections to the same node, but not across nodes. This is what we want.

If we went with a different fake port every time, then we wouldn't reuse sessions at all.

core/src/main/java/com/datastax/oss/driver/internal/core/ssl/SniSslEngineFactory.java
olim7t: > I only hope using zero as a port number won't bite us back one day.

We can always add a constant offset to the list index.
We just have to be careful not to collide with the metadata service's port, which is 30443 IIRC.

core/src/main/java/com/datastax/oss/driver/internal/core/ssl/SniSslEngineFactory.java
olim7t: I think SSL session caching is done at the `SSLContext` level (need to confirm that).

We have one `SniSslEngineFactory` per driver session, and each factory uses its own `SSLContext` (created in `CloudConfigFactory` actually). So I don't think there can be any collision across driver sessions.

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/ssl/SniSslEngineFactory.java
olim7t: > starting at the beginning of the ephemeral port range (49152)

I missed that the first time I read your comment, sounds perfect.

core/src/main/java/com/datastax/oss/driver/internal/core/ssl/SniSslEngineFactory.java
olim7t: Just to confirm, the cache is indeed [local to the SSLContext](https://github.com/JetBrains/jdk8u_jdk/blob/master/src/share/classes/sun/security/ssl/SSLContextImpl.java#L47).

tomekl007: Ok, thx for checking this.

PR #1356, opened 2019-10-25
JAVA-2517: Document incompatibility of unresolved addresses and SSL host name validation

adutra:


core/src/main/java/com/datastax/oss/driver/api/core/ssl/ProgrammaticSslEngineFactory.java
```diff
@@ -69,6 +69,24 @@ public ProgrammaticSslEngineFactory(
...
    * Creates an instance with the given {@link SSLContext}, cipher suites and host name validation.
    *
+   * <p>Note: {@linkplain
+   * com.datastax.oss.driver.api.core.config.DefaultDriverOption#SSL_HOSTNAME_VALIDATION host name
+   * validation} is not compatible with unresolved addresses. If this feature is enabled, make sure
+   * that you only provide resolved contact points:
+   *
+   * <ul>
+   *   <li>If you provided your {@linkplain
+   *       com.datastax.oss.driver.api.core.config.DefaultDriverOption#CONTACT_POINTS contact
```
tomekl007, line 79: It mean that when set `com.datastax.oss.driver.api.core.config.DefaultDriverOption#CONTACT_POINTS` we are createing unresolved addresses, right?

tomekl007 (approved):

core/src/main/java/com/datastax/oss/driver/api/core/ssl/ProgrammaticSslEngineFactory.java
adutra: It depends on `advanced.resolve-contact-points` see `ContactPoints`.

olim7t:
I don't think there's any issue here, the error is a side-effect of the fact that we use IPs in our integration tests.

Consider the following:
```java
    InetSocketAddress address;

    address = new InetSocketAddress("127.0.0.1", 9042);
    System.out.printf(
        "1) from IP, resolved: address=%s, hostString=%s, hostName=%s%n",
        address, address.getHostString(), address.getHostName());

    address = InetSocketAddress.createUnresolved("127.0.0.1", 9042);
    System.out.printf(
        "2) from IP, unresolved: address=%s, hostString=%s, hostName=%s%n",
        address, address.getHostString(), address.getHostName());

    address = InetSocketAddress.createUnresolved("localhost", 9042);
    System.out.printf(
        "3) from name, unresolved: address=%s, hostString=%s, hostName=%s%n",
        address, address.getHostString(), address.getHostName());
```

Which produces (added justification for readability):

```java
1) from IP, resolved:     address=localhost/127.0.0.1:9042, hostString=127.0.0.1, hostName=localhost
2) from IP, unresolved:   address=127.0.0.1:9042,           hostString=127.0.0.1, hostName=127.0.0.1
3) from name, unresolved: address=localhost:9042,           hostString=localhost, hostName=localhost
```

When you call `CCM_RULE.getContactPoints()`, it returns a numerical address (case 2). The SSL engine factory uses the hostName, so it will compare "127.0.0.1" to "localhost" in the certificate, hence the failure to connect (side note: you're also allowed to provide the IP as an alternative name, but that's not the case in our test certificate).

In contrast, `DefaultSslEngineFactoryHostnameValidationIT` uses resolved addresses, so it's in case 1. "localhost" matches the certificate.

I think people who use unresolved addresses will be in case 3 (case 2 makes no sense really, why would you keep it unresolved if it's already an IP).
Also, note that ITs that do hostname validation must build the CCM rule with `withSslLocalhostCn()`.

If I build the rule that way and use `InetSocketAddress.createUnresolved("localhost", 9042)` as the contact point, the connection succeeds.

tomekl007:
> I think people who use unresolved addresses will be in case 3 (case 2 makes no sense really, why would you keep it unresolved if it's already an IP).

This issue came up in KAF-80 where people are providing their `contact-points` not directly using our ProgrammaticBuilder API but through config mechanism of the tool (I think that the same situation is for dsbulk). I am observing the same behavior. When I provide the "localhost" as the contact point and create it using `createUnresolved()` hostname validation works (case 3), but when I provide IP: "127.0.0.1" and do `createUnresolved()` the hostname validation don't work (case 2) so it is the same as your experiment.

However, regarding case 2 that makes no sense - yes I agree but when you do automatic configuration loading I am not sure if you want to distinguish and detect if the provided address is an IP or hostname (and create `new InetSocketAddress` or `InsetSocetAddress.createUnresolved()`). 
It will complicate the logic and maybe a bit fragile. I wonder if there is a way to make our tests work for ccm for both cases (2 and 3)

adutra:
Ok thanks for the explanation, I'm closing this.

PR #1357, opened 2019-10-25
JAVA-2518: Reduce nightly build matrix

emerkle826:
The change proposed will reduce the matrix to the following:
jenkins_3 x_matrix

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -21,11 +21,21 @@ schedules:
   nightly:
     # Run full suite nightly on change for all primary branches if they have changes.
     schedule: nightly
+    matrix:
+      exclude:
+        # exclude builds for newer Cassandra versions and older JDKs
+        - java: openjdk6
+          cassandra: ['2.2', '3.0', '3.11']
+        - java: openjdk7
+          cassandra: ['2.1', '3.0', '3.11']
+        # exclude builds for older Cassandra versions and newer JDKs
+        - java: openjdk11
+          cassandra: ['2.1', '2.2']
     branches:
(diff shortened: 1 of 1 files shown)


build.yaml
```diff
@@ -66,7 +77,8 @@ build:
       jdk_switcher use $JAVA_VERSION
   - type: maven
     version: 3.2.5
-    goals: verify --fail-never -P$TEST_GROUP
+    # for Nightly builds, only run the "long" tests against OpenJDK8
+    goals: verify --fail-never -P$(if [ "${NIGHTLY}" = "true" -a "${JAVA_VERSION}" != "openjdk8" ]; then echo "short"; else echo ${TEST_GROUP}; fi)
```
adutra: Unfortunately this is failing on Jenkins, see: 

https://jenkins-drivers.build.dsinternal.org/view/java-driver/job/datastax.java-driver.java2518.commit/2/CASSANDRA_VERSION=3.11,JAVA_VERSION=openjdk11,OS_VERSION=ubuntu%2Fbionic64%2Fjava-driver/console

emerkle826: I was afraid of that. I'm not sure I can easily adjust the `build.yaml` to switch between `long` and `short` tests within the nightly schedule currently. I'll keep trying a few things, but if it gets to the point where changing Job Creator is the only way, I'm not sure it's worth it at that point.

emerkle826: I've reverted this and just left it a reduced matrix for nightly builds.

adutra (approved):

PR #1361, opened 2019-11-06
JAVA-2532: Add BoundStatement ReturnType for insert, update, and delete DAO methods

fromanator:
Motivation:

In the 3.x driver the object-mapper had Mapper.saveQuery(entity) and
 Mapper.deleteQuery(id) which returned a BoundStatement that you could
 later execute. This was useful for creating a list of BoundStatement
 that would be added to a LOGGED Batch statement for atomically saving
 data that is saved across several denormalized tables (to support
 different query patterns on the same data).

While the current code sort of enables that with the @SetEntity, it
 however lacks a lot of the features that the other DAO methods have,
 for instance setting a TTL, custom IF and WHERE clauses that are
 are supported by the @Insert, @Update, and @Delete methods. And not
 all of these feature may be supportable on this more generic implemenation.

Modifcations:

Added a new DaoReturnType.BOUND_STATEMENT that is practically identical to
 to the existing DaoReturnType. Since it just returns the BoundStatment
 instead of executing it.

Result:

For @Insert, @Update, @Delete, and @Query DAO methods BoundStatement is
 now a new supported return type. This is an additive change to the API
 of the Mapper so it is non-breaking.

### DIFF
--- a/manual/mapper/daos/delete/README.md
+++ b/manual/mapper/daos/delete/README.md
@@ -105,6 +105,14 @@ The method can return:
     ResultSet deleteIfDescriptionMatches(UUID productId, String expectedDescription);
     // if the condition fails, the result set will contain columns '[applied]' and 'description'
     ```
+  
+* a [BoundStatement]. This is intended for queries where you will execute this statement later
+  or in a batch.
+  
+    ```java
+    @Delete
+    BoundStatement delete(Product product);
+    ```
     
 * a [CompletionStage] or [CompletableFuture] of any of the above. The method will execute the query
   asynchronously. Note that for result sets, you need to switch to [AsyncResultSet].
@@ -141,6 +149,7 @@ entity class and the [naming strategy](../../entities/#naming-strategy)).
 [@PartitionKey]:          https://docs.datastax.com/en/drivers/java/4.3/com/datastax/oss/driver/api/mapper/annotations/PartitionKey.html
 [ResultSet]:              https://docs.datastax.com/en/drivers/java/4.3/com/datastax/oss/driver/api/core/cql/ResultSet.html
 [ResultSet#wasApplied()]: https://docs.datastax.com/en/drivers/java/4.3/com/datastax/oss/driver/api/core/cql/ResultSet.html#wasApplied--
+[BoundStatement]:         https://docs.datastax.com/en/drivers/java/4.3/com/datastax/oss/driver/api/core/cql/BoundStatement.html
 
 [CompletionStage]: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html
 [CompletableFuture]: https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html
--- a/manual/mapper/daos/query/README.md
+++ b/manual/mapper/daos/query/README.md
@@ -54,6 +54,9 @@ The method can return:
 
 * a [ResultSet]. The method will return the raw query result, without any conversion.
 
+* a [BoundStatement]. This is intended for queries where you will execute this statement later
+  or in a batch.
+
 * a [PagingIterable]. The method will convert each row into an entity instance.
 
 * a [CompletionStage] or [CompletableFuture] of any of the above. The method will execute the query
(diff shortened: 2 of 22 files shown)


manual/mapper/daos/delete/README.md
```diff
@@ -105,6 +105,15 @@ The method can return:
...
     ```
+  
+* a [BoundStatement]. This is intended for queries with custom IF clauses; where you will execute
+  this statement later or in a batch.
+  
+    ```java
+    @Delete(entityClass = Product.class, customIfClause = "description = :expectedDescription")
+    BoundStatement deleteIfDescriptionMatches(UUID productId, String expectedDescription);
+    // if the condition fails, the result set will contain columns '[applied]' and 'description'
+    ```
```
olim7t, line 115: > This is intended for queries with custom IF clauses

I don't see why, is this a copy-paste error?

If so we should also modify the example to something simpler. A simple `BoundStatement delete(Product product)` should suffice.

olim7t, line 115: This should also be added to the [annotation javadocs](https://docs.datastax.com/en/drivers/java/4.3/com/datastax/oss/driver/api/mapper/annotations/Delete.html).

There's a lot of overlap between the javadocs and the manual, we try to get the information out any way we can :)

olim7t (changes requested):
Sounds reasonable 👍 

I agree with your decision not to do it for `Select`, if the use case is batching this should not be necessary. It can always be added at a later point.

The implementation looks good, just a few minor remarks about the docs. Nit: the PR title and commit message do not mention `Query`; maybe it's not necessary to list all methods, e.g. "Add BoundStatement return type in mapper".

manual/mapper/daos/delete/README.md
fromanator: That makes sense, since the primary goal of this feature it to give you the bound statement to do what you will with it later. It's just a bonus that you can also use conditionals etc too if you choose. With how it's documented now it's slightly confusing in that you could only use this if you where using a conditional.

fromanator: Added the docs to the annotation classes too.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGeneratorTest.java
```diff
@@ -105,7 +105,7 @@ public void should_fail_with_expected_error(
       },
       {
         "Delete methods must return one of [VOID, FUTURE_OF_VOID, BOOLEAN, FUTURE_OF_BOOLEAN, "
-            + "RESULT_SET, FUTURE_OF_ASYNC_RESULT_SET]",
+            + "RESULT_SET, BOUND_STATEMENT, FUTURE_OF_ASYNC_RESULT_SET]",
```
adutra, line 108: Note to team members: we need to amend DSE mapper tests once this gets merged, some of the tests that expect a specific error message are failing with these changes.

adutra (approved):

manual/mapper/daos/delete/README.md
```diff
@@ -105,6 +105,13 @@ The method can return:
     ResultSet deleteIfDescriptionMatches(UUID productId, String expectedDescription);
     // if the condition fails, the result set will contain columns '[applied]' and 'description'
     ```
+  
+* a [BoundStatement]. This is intended for queries where you will execute this statement later or in a batch.
```
olim7t: Nit: our coding convention uses a column limit of 100 characters, could you add a line break here?

olim7t (changes requested):
One last request: could you amend the docs for the `Query` annotation as well?

fromanator:
@olim7t noticed I missed the `README.md` for `Query` as well, so I updated that along with the annotation good catch.

manual/mapper/daos/query/README.md
```diff
@@ -54,6 +54,9 @@ The method can return:
 
 * a [ResultSet]. The method will return the raw query result, without any conversion.
 
+* a [BoundStatement]. This is intended for queries where you will execute this statement later
```
fromanator, line 57: Added this because I missed it.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/Query.java
```diff
@@ -83,6 +84,8 @@
  *   <li>an {@link Optional} of an entity class. The method will extract the first row and convert
  *       it, or return {@code Optional.empty()} if the result set is empty.
  *   <li>a {@link ResultSet}. The method will return the raw query result, without any conversion.
+ *   <li>a {@link BoundStatement}. This is intended for cases where you intend to execute this
```
fromanator, line 87: Also added missing `Query` annotation java docs.

olim7t (approved):

adutra:
I added the missing changelog entry and committed to 4.x as 0f0da297c. Thank you for your contribution @fromanator !

PR #1362, opened 2019-11-08
JAVA-2527: Allow AllNodesFailedException to accept more than one error per node

adutra:
Note: this PR fixes the core issue and exposes a new getter returning a `Map<Node, List<Throwable>>` structure containing all errors (it also deprecates the old `getErrors` method).

It does not add suppressed exceptions as suggested in JAVA-2528.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
@@ -17,13 +17,18 @@ package com.datastax.oss.driver.api.core;
 
 import com.datastax.oss.driver.api.core.cql.ExecutionInfo;
 import com.datastax.oss.driver.api.core.metadata.Node;
-import com.datastax.oss.driver.shaded.guava.common.base.Joiner;
+import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableMap;
-import com.datastax.oss.driver.shaded.guava.common.collect.Iterables;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
 
 /**
  * Thrown when a query failed on all the coordinators it was tried on. This exception may wrap
@@ -31,42 +36,46 @@ import java.util.Map;
  */
 public class AllNodesFailedException extends DriverException {
 
+  /** @deprecated Use {@link #fromErrors(List)} instead. */
   @NonNull
+  @Deprecated
   public static AllNodesFailedException fromErrors(@Nullable Map<Node, Throwable> errors) {
     if (errors == null || errors.isEmpty()) {
       return new NoNodeAvailableException();
     } else {
-      return new AllNodesFailedException(ImmutableMap.copyOf(errors));
+      return new AllNodesFailedException(groupByNode(errors.entrySet()));
     }
   }
 
   @NonNull
-  public static AllNodesFailedException fromErrors(
-      @Nullable List<Map.Entry<Node, Throwable>> errors) {
-    Map<Node, Throwable> map;
+  public static AllNodesFailedException fromErrors(@Nullable List<Entry<Node, Throwable>> errors) {
     if (errors == null || errors.isEmpty()) {
-      map = null;
+      return new NoNodeAvailableException();
     } else {
-      ImmutableMap.Builder<Node, Throwable> builder = ImmutableMap.builder();
-      for (Map.Entry<Node, Throwable> entry : errors) {
-        builder.put(entry);
-      }
-      map = builder.build();
+      return new AllNodesFailedException(groupByNode(errors));
     }
-    return fromErrors(map);
   }
 
-  private final Map<Node, Throwable> errors;
+  private final Map<Node, List<Throwable>> errors;
 
+  /** @deprecated Use {@link #AllNodesFailedException(String, ExecutionInfo, Iterable)} instead. */
+  @Deprecated
   protected AllNodesFailedException(
       @NonNull String message,
       @Nullable ExecutionInfo executionInfo,
       @NonNull Map<Node, Throwable> errors) {
-    super(message, executionInfo, null, true);
-    this.errors = errors;
+    this(message, executionInfo, groupByNode(errors.entrySet()));
   }
 
-  private AllNodesFailedException(Map<Node, Throwable> errors) {
+  protected AllNodesFailedException(
+      @NonNull String message,
+      @Nullable ExecutionInfo executionInfo,
+      @NonNull Iterable<Entry<Node, List<Throwable>>> errors) {
+    super(message, executionInfo, null, true);
+    this.errors = toDeepImmutableMap(errors);
+  }
+
+  private AllNodesFailedException(Collection<Entry<Node, List<Throwable>>> errors) {
     this(
         buildMessage(
             String.format("All %d node(s) tried for the query failed", errors.size()), errors),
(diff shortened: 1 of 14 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
```diff
@@ -17,87 +17,150 @@
...
       @NonNull String message,
       @Nullable ExecutionInfo executionInfo,
       @NonNull Map<Node, Throwable> errors) {
+    this(message, executionInfo, groupByNode(errors.entrySet()));
+  }
+
+  protected AllNodesFailedException(
+      @NonNull String message,
+      @Nullable ExecutionInfo executionInfo,
+      @NonNull Iterable<Entry<Node, List<Throwable>>> errors) {
```
adutra, line 73: Note that this can't be a `Map<Node, List<Throwable>>` because the erased type `Map` would clash with the existing constructor signature. This unfortunately makes the fix not as straightforward as it would seem.

integration-tests/src/test/java/com/datastax/oss/driver/core/AllNodesFailedIT.java
```diff
@@ -0,0 +1,104 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map.Entry;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(ParallelizableTests.class)
+public class AllNodesFailedIT {
```
adutra, line 47: I tried to create a unit test first using the existing test harness, but spent too long trying to get it to work, so I went for an IT instead.

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
```diff
@@ -17,87 +17,150 @@
...
+    while (iterator.hasNext()) {
+      Entry<Node, List<Throwable>> entry = iterator.next();
+      details.append(entry.getKey()).append(": ").append(entry.getValue());
+      if (iterator.hasNext()) {
+        details.append(", ");
+      }
+    }
     return String.format(
-        baseMessage + " (showing first %d, use getErrors() for more: %s)", limit, details);
+        baseMessage + " (showing first %d, use getAllErrors() for more): %s", limit, details);
```
adutra: Nit: I allowed myself to move the closing parenthesis before the errors list, which can be very long; otherwise, it seems that the opening parenthesis is unbalanced.

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
```diff
@@ -17,87 +17,150 @@
...
     return String.format(
-        baseMessage + " (showing first %d, use getErrors() for more: %s)", limit, details);
+        "%s (showing first %d, use getAllErrors() for more): %s", baseMessage, limit, details);
   }
 
-  /** The details of the individual error on each node. */
+  /**
+   * An immutable map containing the first error on each tried node.
+   *
+   * @deprecated Use {@link #getAllErrors()} instead.
```
olim7t, line 106: Maybe we should also indicate that this version will only report the first error of each node.

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
```diff
@@ -17,87 +17,150 @@
...
   }
 
-  private static String buildMessage(String baseMessage, Map<Node, Throwable> errors) {
+  private static String buildMessage(
+      String baseMessage, Collection<Entry<Node, List<Throwable>>> errors) {
     int limit = Math.min(errors.size(), 3);
-    String details =
-        Joiner.on(", ").withKeyValueSeparator(": ").join(Iterables.limit(errors.entrySet(), limit));
-
+    Iterator<Entry<Node, List<Throwable>>> iterator = errors.iterator();
```
olim7t: We've lost the limit in the refactoring. You can use `Iterables.limit()` on this line.

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
```diff
@@ -17,87 +17,150 @@
...
+    while (iterator.hasNext()) {
+      Entry<Node, List<Throwable>> entry = iterator.next();
+      details.append(entry.getKey()).append(": ").append(entry.getValue());
+      if (iterator.hasNext()) {
+        details.append(", ");
+      }
+    }
     return String.format(
-        baseMessage + " (showing first %d, use getErrors() for more: %s)", limit, details);
+        "%s (showing first %d, use getAllErrors() for more): %s", baseMessage, limit, details);
```
olim7t: Maybe reword as `showing first %d nodes`, it's what it will do if you apply my other suggestion a few lines above.

olim7t (approved):
LGTM 👍 

Minor formatting suggestions inline.

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
```diff
@@ -17,87 +17,150 @@
...
   public AllNodesFailedException reword(String newMessage) {
     return new AllNodesFailedException(
-        buildMessage(newMessage, errors), getExecutionInfo(), errors);
+        buildMessage(newMessage, errors.entrySet()), getExecutionInfo(), errors.entrySet());
+  }
+
+  private static Set<Entry<Node, List<Throwable>>> groupByNode(
+      Iterable<Entry<Node, Throwable>> errors) {
+    // no need for immutable collections here
+    Map<Node, List<Throwable>> map = new LinkedHashMap<>();
```
tomekl007, line 145: Isn't it better to just use guaga `MultiMap`?

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
```diff
@@ -17,87 +17,150 @@
...
+          });
+    }
+    return map.entrySet();
+  }
+
+  private static Map<Node, List<Throwable>> toDeepImmutableMap(
+      Iterable<Entry<Node, List<Throwable>>> errors) {
+    ImmutableMap.Builder<Node, List<Throwable>> builder = ImmutableMap.builder();
+    for (Entry<Node, List<Throwable>> entry : errors) {
+      builder.put(entry.getKey(), ImmutableList.copyOf(entry.getValue()));
```
tomekl007, line 170: If that is a deep copy, shouldn't we make a safe copy of `entry.getKey()`? Or is the Node returned by `getKey` immutable?

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
```diff
@@ -573,20 +576,21 @@ private void forceClose() {
...
       if (errors.size() == 0) {
         return false;
       }
-      for (Throwable nodeError : errors) {
-        if (!(nodeError instanceof AuthenticationException)) {
-          authFailure = false;
-          break;
+      for (List<Throwable> nodeErrors : errors) {
+        for (Throwable nodeError : nodeErrors) {
+          if (!(nodeError instanceof AuthenticationException)) {
```
tomekl007, line 587: is this refactoring correct? Previously if any of the errors were `AuthenticationException` then this method returned `true`. Now if any of the errors is NOT `AuthenticationException` then this method returns `false`.

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
adutra: That's already stated on line 103, do you think it's worth repeating it here (line 105)?

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
adutra: Not possible, we cannot expose Guava in the API (it's shaded).

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
adutra: `DefaultNode` is mutable, but I don't want to change it in this PR because it was already mutable in the previous implementation of this class. I'm only making the values immutable.

core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
adutra: I don't agree, both algorithms look equivalent to me and both return `false` as soon as any error is NOT `AuthenticationException`.

tomekl007: ah, ok: there is a `break` there, thx for the clarification

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
adutra: The only thing that bothers me is that the signature of this method, at least in theory, cannot guarantee that exactly N distinct nodes will be printed, since duplicate entries could happen (not in the current implementation, but if someone extends this class, then it's not impossible). I will try to change the signature to accept a map of nodes instead, just to be on the safe side.

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
adutra: Done, but I also changed the signature to accept a map of nodes instead. All in all, it looks a bit more straightforward this way than before imo.

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
olim7t: I don't know how I managed to miss that 🤦‍♂
Forget this comment then.

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
olim7t: `DefaultNode` is mutable but it uses reference equality. Changing the mutable fields will never affect how two instances compare.

PR #1363, opened 2019-11-15
JAVA-2107: Add XML formatting plugin

olim7t:
I haven't run the formatter yet. It's better to do it at the last minute, to avoid unnecessary conflicts if the POMs change before we're ready to merge this:
```
mvn xml-format:xml-format
```
(incidentally, this is why the build fails)

I did a test run locally, it's pretty similar to what we have already, and overall satisfactory. There are a few whacky edge cases around comments, but I didn't find any [configuration options](https://acegi.github.io/xml-format-maven-plugin/xml-format-mojo.html) to avoid them. I don't want to spend too much time fighting the formatter, consistency is what matters.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,8 @@
 
 ## Code formatting
 
+### Java
+
 We follow the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html). See
 https://github.com/google/google-java-format for IDE plugins. The rules are not configurable.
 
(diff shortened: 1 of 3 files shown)


adutra (approved):
Looks good, I'm only worried about line endings, the default is "LF" (Use Unix and Mac style line endings) which is OK for most of us but I wonder if this isn't going to create a snafu on Windows.

UPDATE
I tried this on a Windows VM, it does format everything with LF but IntelliJ and most text editors didn't choke on it. Committing and pushing the formatted XML, then reading it from a non-Windows machine worked as expected.

PR #1364, opened 2019-11-18
JAVA-2539: Stop assuming CDC unavailable for Cassandra prior to 3.8.0

chapj2001:
Simple change to use CDC value as reported by Cassandra instead of forcing to false when Cassandra version is less than 3.8.0. This change unblocks my team from using a custom Cassandra variant which backports CDC to an older version.

FYI I'm submitting this PR to 3.2.x because that's the branch which first introduces CDC (along with the Cassandra version restriction). I'm not actually using 3.2.x; my team is currently migrating from 3.3.2 to 3.7.1. We'd be migrating to latest release except that 3.7.2 introduced a small performance improvement incompatible with latest published release of Simulacron (see [Simulacron issue 88 Please publish 0.8.11 to maven central](https://github.com/datastax/simulacron/issues/88)). Once Simulacron 0.8.11 is published to maven central my team will likely upgrade to 3.8.0.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
@@ -269,9 +269,7 @@ public abstract class AbstractTableMetadata {
         if (cassandraVersion.getMajor() > 2) {
             and(sb, formatted).append("crc_check_chance = ").append(options.getCrcCheckChance());
         }
-        if (cassandraVersion.getMajor() > 3 || (cassandraVersion.getMajor() == 3 && cassandraVersion.getMinor() >= 8)) {
-            and(sb, formatted).append("cdc = ").append(options.isCDC());
-        }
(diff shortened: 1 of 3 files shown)


adutra (commented):
I don't think we will be able to accept the proposed change. Appending `cdc = false` to all generated statements will make such statements invalid for all C* versions < 3.8.0 (without backported CDC). I just tried and got:

```
SyntaxException: <ErrorMessage code=2000 [Syntax error in CQL query] message="Unknown property 'cdc'">
```

Actually we have no way to account for backports. Someone could do the same as you and backport, say, `crc_check_chance` to C* < 3.0. But if you append `crc_check_chance = ...` to the DDL statement, that would make it invalid as well.

In driver 3 we do not store whether or not a given property was found or not in table metadata, so we cannot reliably reproduce exactly which properties were found, and which values. In driver 4 we improved this a lot, and the driver now only reports non-null properties found in the original table metadata.

If you can, please upgrade to driver 4.3.0.

PR #1366, opened 2019-11-26
JAVA-2552: RequestHandlerTestHarness improvements

adutra:


core/src/test/java/com/datastax/oss/driver/internal/core/cql/PoolBehavior.java
```diff
@@ -59,7 +58,7 @@ public PoolBehavior(Node node, boolean createChannel) {
       this.channel = mock(DriverChannel.class);
       EventLoop eventLoop = mock(EventLoop.class);
       ChannelConfig config = mock(DefaultSocketChannelConfig.class);
-      this.writePromise = GlobalEventExecutor.INSTANCE.newPromise();
```
tomekl007, line 62: what was the reason why `GlobalEventExecutor` was not performing well here?
Was it because: 
```
and stops it when there is no
 * task pending in the task queue for 1 second`
```
?

core/src/test/java/com/datastax/oss/driver/internal/core/cql/DirectThreadEventExecutor.java
```diff
@@ -0,0 +1,53 @@
...
+
+  private DirectThreadEventExecutor() {
+    super(null, MoreExecutors.directExecutor(), true);
+  }
+
+  @Override
+  protected void run() {
+    // copied from DefaultEventExecutor
+    for (; ; ) {
+      Runnable task = takeTask();
```
tomekl007: shouldn't we add some kind of delay between the invocation of this `takeTask()`?
Similar to `GlobalEventExecutor`:
`taskQueue.poll(delayNanos, TimeUnit.NANOSECONDS);` ?

core/src/test/java/com/datastax/oss/driver/internal/core/cql/PoolBehavior.java
adutra: No it's because it's a true event loop that runs all its tasks asynchronously in a separate thread. We want to run such tasks synchronously to make tests even more deterministic than they currently are.

core/src/test/java/com/datastax/oss/driver/internal/core/cql/DirectThreadEventExecutor.java
adutra: I don't see this instruction in `GlobalEventExecutor.run()`.

tomekl007: Ok, you right. This is already within `takeTask()` so we should be ok

tomekl007 (approved):

adutra:
When rebasing onto 4.x, please keep the 2 commits separate.

core/src/test/java/com/datastax/oss/driver/internal/core/util/concurrent/CapturingTimer.java
```diff
@@ -51,12 +49,7 @@ public Timeout newTimeout(TimerTask task, long delay, TimeUnit unit) {
    * tell from the returned Timeout itself.
    */
   public CapturedTimeout getNextTimeout() {
-    try {
-      return timeoutQueue.poll(100, TimeUnit.MILLISECONDS);
-    } catch (InterruptedException ie) {
-      fail("Unexpected interruption", ie);
-      throw new AssertionError();
-    }
+    return timeoutQueue.poll();
```
adutra, line 52: We don't need to wait anymore here. Since write promises are completed on the calling thread, the scheduled timeout is either already there, or will never be.

PR #1367, opened 2019-11-28
JAVA-2556: Make ExecutionInfo compatible with any Request type

adutra:


tomekl007 (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultExecutionInfo.java
```diff
@@ -65,7 +68,34 @@ public DefaultExecutionInfo(
...
+      Node coordinator,
+      int speculativeExecutionCount,
+      int successfulExecutionIndex,
+      List<Map.Entry<Node, Throwable>> errors,
+      ByteBuffer pagingState,
+      Frame frame,
+      boolean schemaInAgreement,
+      DefaultSession session,
+      InternalDriverContext context,
+      DriverExecutionProfile executionProfile) {
```
absurdfarce, line 68: Semi-nit: is there a reason not to make this the lone constructor?  Any existing usages of the Statement constructor will transparently work here (given the inheritance hierarchy)... and it avoids a mostly artificial bifurcation between the current two constructors.

adutra: I think you have a valid point here. I kept the old constructor thinking backwards compatibility, but in hindsight, this is an internal class and backwards compatibility isn't a requirement here. So I will remove the deprecated constructor as you suggested. Thanks!

absurdfarce (approved):

adutra:
I will amend the DSE driver accordingly now that this is merged.

PR #1369, opened 2019-12-04
JAVA-2473: Don't reconnect control connection if protocol is downgraded

olim7t:
Motivation:

In theory, the control connection is an implementation detail of the
topology monitor and metadata manager. If those components are
overridden with custom versions that use another way to get the data,
the rest of the driver should be able to function without initializing
the control connection at all.

This is currently not the case: DefaultSession references the control
connection, in order to reconnect if the protocol version was
downgraded after the initial connection (JAVA-1295).

Modifications:

Do not reconnect the control connection in that case; it's not really
needed because none of the control queries use any protocol-dependent
feature.

This also allows us to remove MetadataManager.firstSchemaRefreshFuture()
and slightly simplify session initialization.

Result:

DefaultSession does not reference ControlConnection.

If the protocol version is downgraded, the control connection keeps
using the initial version. It might switch to the "correct" version
later if it reconnects to another node.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
@@ -32,7 +32,6 @@ import com.datastax.oss.driver.api.core.type.reflect.GenericType;
 import com.datastax.oss.driver.internal.core.channel.DriverChannel;
 import com.datastax.oss.driver.internal.core.context.InternalDriverContext;
 import com.datastax.oss.driver.internal.core.context.LifecycleListener;
-import com.datastax.oss.driver.internal.core.control.ControlConnection;
 import com.datastax.oss.driver.internal.core.metadata.MetadataManager;
 import com.datastax.oss.driver.internal.core.metadata.NodeStateEvent;
 import com.datastax.oss.driver.internal.core.metadata.NodeStateManager;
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
```diff
@@ -378,36 +376,27 @@ private void afterInitialNodeListRefresh(CqlIdentifier keyspace) {
...
+                (metadata, error) -> {
+                  if (error != null) {
+                    Loggers.warnWithException(
+                        LOG,
+                        "[{}] Unexpected error while refreshing schema during initialization, "
+                            + "keeping previous version",
+                        logPrefix,
+                        error);
+                  }
+                  afterInitialSchemaRefresh(keyspace);
```
adutra, line 398: Should this be invoked even in case of failure? It seems this wasn't being called in case of failure before.

olim7t: Correct, but it is, it's after the error check, not in an `else` branch.

olim7t: Correct, but it does run in all cases, it's after the error check, not in an `else` branch.

adutra (approved):

PR #1371, opened 2019-12-05
JAVA-2569: Make driver compatible with Netty < 4.1.34 again

adutra:


manual/core/shaded_jar/README.md
```diff
@@ -3,6 +3,10 @@
 The default driver JAR depends on [Netty](http://netty.io/), which is
 used internally for networking.
 
+The driver is compatible with all Netty versions in the range [4.1.7, 4.2.0),
+that is, it can work with any version equal to or higher than 4.1.7, and
```
adutra, line 7: Compatibility with Netty < 4.1.7 is much trickier, so I preferred to stop here.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ConnectInitHandler.java
```diff
@@ -59,8 +58,10 @@ public void connect(
     realConnectPromise.addListener(future -> onRealConnect(ctx));
 
     // Make the caller's promise wait on the other two:
-    PromiseCombiner combiner = new PromiseCombiner(ImmediateEventExecutor.INSTANCE);
-    combiner.addAll(new Future[] {realConnectPromise, initPromise});
+    // JAVA-2569: preserve binary compatibility with Netty < 4.1.34
+    @SuppressWarnings("deprecation")
+    PromiseCombiner combiner = new PromiseCombiner();
```
tomekl007: I understand that `PromiseComibner()` constructor is deprecated, maybe we could extract this as a method and do deprecation inside of it? Then we will not need that duplication(https://github.com/datastax/java-driver/pull/1371/files#diff-08569c129701967a21d9a64ee733151eR179), and if someone will need to construct this class, he could call this method that encapsulates the logic?

tomekl007 (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ConnectInitHandler.java
adutra: You mean, create a utility class exposing a static method `PromiseCombiner newPromiseCombiner()`? I think I disagree with you for two reasons: 1) the creation of that utility class requires the same amount of lines of code than the current situation so it doesn't bring concision; and 2) I bet we would forget to use this utility class and keep instantiating `PromiseCombiner` by simply doing `new PromiseCombiner()`...

olim7t:
Actually this is not as simple as reverting to the old constructor (sorry I should have remembered that earlier). It was deprecated because it is only thread-safe if the combiner is confined to a single event loop, see netty/netty@625c4e8286d3c8a36046feb1289b6b710c7541a5.

If we use an immediate executor, the child promises will increment `PromiseCombiner.doneCount` from a different thread, and if you look at the implementation there is no synchronization.

(I'm also noticing just now that I correctly scheduled the construction of the combiner on the global executor in `DefaultNettyOptions`, but forgot to do it in `ConnectInitHandler`.)

At this stage I think our best course of action is to write our own combiner, that uses proper internal synchronization.

olim7t:
Tentative implementation:
```java
package com.datastax.oss.driver.internal.core.util.concurrent;

public class ThreadsafePromiseCombiner {
  private final Promise<Void> aggregatePromise;
  private final AtomicInteger remainingCount;
  private final AtomicReference<Throwable> aggregateFailureRef = new AtomicReference<>();

  public ThreadsafePromiseCombiner(Promise<Void> aggregatePromise, Future<?>... parents) {
    this.aggregatePromise = aggregatePromise;
    this.remainingCount = new AtomicInteger(parents.length);
    for (Future<?> parent : parents) {
      parent.addListener(this::onParentComplete);
    }
  }

  private void onParentComplete(Future<?> parent) {
    if (!parent.isSuccess()) {
      aggregateFailureRef.updateAndGet(
          aggregateFailure -> {
            if (aggregateFailure == null) {
              aggregateFailure =
                  new ExecutionException(
                      "One or more of the parent futures failed", parent.cause());
            } else {
              aggregateFailure.addSuppressed(parent.cause());
            }
            return aggregateFailure;
          });
    }

    if (remainingCount.decrementAndGet() == 0) {
      Throwable aggregateFailure = aggregateFailureRef.get();
      if (aggregateFailure != null) {
        aggregatePromise.tryFailure(aggregateFailure);
      } else {
        aggregatePromise.trySuccess(null);
      }
    }
  }
}
```

olim7t (changes requested):
See previous comments

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/PromiseCombiner.java
```diff
@@ -0,0 +1,83 @@
...
+public class PromiseCombiner {
+
+  /**
+   * Combines the given futures into the given promise, that is, ties the completion of the latter
+   * to that of the formers.
+   *
+   * @param aggregatePromise The promise that will complete when all parents complete.
+   * @param parents The parent futures.
+   */
+  public static void combine(
```
adutra, line 40: @olim7t I changed your design slightly because it felt odd to instantiate an object without doing anything with it after (that is, all the action was happening in its constructor).

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/PromiseCombiner.java
```diff
@@ -0,0 +1,83 @@
...
+      remainingCount = new AtomicInteger(numberOfParents);
+    }
+
+    @Override
+    public void operationComplete(Future<Object> future) {
+      if (!future.isSuccess()) {
+        aggregateFailureRef.updateAndGet(
+            aggregateFailure -> {
+              if (aggregateFailure == null) {
+                aggregateFailure = future.cause();
```
adutra, line 66: Not wrapping the cause in an `ExecutionException` because `Promise.get()`  already does so.

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/PromiseCombiner.java
```diff
@@ -0,0 +1,83 @@
...
+              } else {
+                aggregateFailure.addSuppressed(future.cause());
+              }
+              return aggregateFailure;
+            });
+      }
+      if (remainingCount.decrementAndGet() == 0) {
+        Throwable aggregateFailure = aggregateFailureRef.get();
+        if (aggregateFailure != null) {
+          aggregatePromise.tryFailure(aggregateFailure);
```
adutra, line 76: The aggregated promise is now completed in an arbitrary thread. I think this is OK for us since the 2 usages we had so far also mix promises from different event executors. But just wanted to point this out here for reviewers.

core/src/test/java/com/datastax/oss/driver/internal/core/util/concurrent/PromiseCombinerTest.java
```diff
@@ -0,0 +1,62 @@
...
+    Promise<Void> parent1 = executor.newPromise();
+    Promise<Void> parent2 = executor.newPromise();
+    Promise<Void> parent3 = executor.newPromise();
+    NullPointerException npe = new NullPointerException();
+    IOException ioe = new IOException();
+    // when
+    PromiseCombiner.combine(promise, parent1, parent2, parent3);
+    parent1.setSuccess(null);
+    parent2.setFailure(npe);
+    parent3.setFailure(ioe);
```
tomekl007, line 57: Is it possible to do `setFailure(null)`? If so, maybe we should add a test case for it?

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/PromiseCombiner.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @Override
+    public void operationComplete(Future<Object> future) {
+      if (!future.isSuccess()) {
+        aggregateFailureRef.updateAndGet(
+            aggregateFailure -> {
+              if (aggregateFailure == null) {
+                aggregateFailure = future.cause();
+              } else {
+                aggregateFailure.addSuppressed(future.cause());
```
tomekl007, line 68: is it possible that `future.cause()` will be `null`? or `future`? if so our logic may not work because of this check: https://github.com/datastax/java-driver/pull/1371/files/8ae9ee79a503454657dda8814ccb3b132e6e0db4..3242753311848dd161315230777949c7de750dc3#diff-ab0e22e2e333da4968ffb1f8318c49afR75

core/src/test/java/com/datastax/oss/driver/internal/core/util/concurrent/PromiseCombinerTest.java
adutra: It's not allowed, you get: `java.lang.NullPointerException: cause`.

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/PromiseCombiner.java
adutra: `future` itself cannot be null. `future.cause()` can be null if 1) the future is not completed yet, or if it completed successfully. But here these conditions are not met, since when `operationComplete` is called the future is necessarily complete, and since we tested that `future.isSuccess() == false`, we know that the future failed.

adutra:
@olim7t would you mind making a second review please?

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/PromiseCombiner.java
olim7t: There's a possibility that the exception we get here doesn't support suppressed exceptions, but I think we can ignore this corner case. The worst that can happen is that other futures' causes will be ignored when we add them at line 68.

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/PromiseCombiner.java
olim7t: Completing a promise is thread-safe so I don't think this matters.

olim7t (approved):

PR #1372, opened 2019-12-12
JAVA-2442: Don't check for schema agreement twice when completing a DDL query

olim7t:
Motivation:

When we process a SCHEMA_CHANGE response to a DDL query, we check for
schema agreement, and then trigger a schema refresh. But the refresh
itself also checks for agreement, so we end up doing it twice.

Modifications:

Only check for agreement during the refresh. Introduce a new
RefreshSchemaResult type, so that we can surface the outcome of the
agreement in addition to the new metadata.
Remove the agreement check in CqlRequestHandler, use the value surfaced
by the refresh instead.

Result:

Schema agreement runs only once.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
@@ -598,26 +598,24 @@ public class CqlRequestHandler implements Throttled {
         if (responseMessage instanceof SchemaChange) {
           SchemaChange schemaChange = (SchemaChange) responseMessage;
           context
-              .getTopologyMonitor()
-              .checkSchemaAgreement()
-              .thenCombine(
-                  context
-                      .getMetadataManager()
-                      .refreshSchema(schemaChange.keyspace, false, false)
-                      .exceptionally(
-                          error -> {
-                            Loggers.warnWithException(
-                                LOG,
-                                "[{}] Unexpected error while refreshing schema after DDL query, "
-                                    + "keeping previous version",
-                                logPrefix,
-                                error);
-                            return null;
-                          }),
-                  (schemaInAgreement, metadata) -> schemaInAgreement)
+              .getMetadataManager()
+              .refreshSchema(schemaChange.keyspace, false, false)
               .whenComplete(
-                  (schemaInAgreement, error) ->
-                      setFinalResult(schemaChange, responseFrame, schemaInAgreement, this));
+                  (result, error) -> {
+                    boolean schemaInAgreement;
+                    if (error != null) {
+                      Loggers.warnWithException(
+                          LOG,
+                          "[{}] Unexpected error while refreshing schema after DDL query, "
+                              + "keeping previous version",
+                          logPrefix,
+                          error);
+                      schemaInAgreement = false;
+                    } else {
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
```diff
@@ -598,26 +598,24 @@ public void onResponse(Frame responseFrame) {
         if (responseMessage instanceof SchemaChange) {
           SchemaChange schemaChange = (SchemaChange) responseMessage;
           context
-              .getTopologyMonitor()
-              .checkSchemaAgreement()
-              .thenCombine(
-                  context
-                      .getMetadataManager()
-                      .refreshSchema(schemaChange.keyspace, false, false)
```
olim7t, line 606: Also, `thenCombine` was wrong here, because it started the refresh before the agreement was done.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
```diff
@@ -396,33 +423,55 @@ private void acceptSchemaRequest(CompletableFuture<Metadata> future, boolean flu
...
-                  if (error != null) {
-                    currentSchemaRefresh.completeExceptionally(error);
+                (schemaInAgreement, agreementError) -> {
+                  if (agreementError != null) {
+                    refreshFuture.completeExceptionally(agreementError);
+                  } else {
+                    schemaQueriesFactory
+                        .newInstance(
+                            // TODO remove this unused parameter (see JAVA-2582)
+                            null)
```
olim7t, line 437: I could have addressed it directly, but it would introduce additional merge conflicts in the OSS/DSE merge. So I created a follow-up ticket, it's not urgent.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
adutra: I was about to ask if there are implications in refreshing the schema before agreement is reached, but it seems we've been in that boat for a while now.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/MetadataManager.java
```diff
@@ -396,33 +423,55 @@ private void acceptSchemaRequest(CompletableFuture<Metadata> future, boolean flu
...
+                                refreshFuture.completeExceptionally(metadataError);
+                              } else {
+                                refreshFuture.complete(
+                                    new RefreshSchemaResult(newMetadata, schemaInAgreement));
+                              }
+
+                              firstSchemaRefreshFuture.complete(null);
+
+                              currentSchemaRefresh = null;
+                              // If another refresh was enqueue during this one, run it now
```
adutra: Nit: enqueue -> enqueued

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
olim7t: I guess it would manifest as the control node not having the new table yet, we've never observed it in practice.

PR #1373, opened 2019-12-13
JAVA-2523: Disallow programmatic SSL methods with cloud

GregBestland:


integration-tests/src/test/java/com/datastax/oss/driver/api/core/cloud/CloudIT.java
```diff
@@ -219,4 +222,23 @@ public void should_error_when_contact_points_and_secure_bundle_used() {
...
+      CqlSessionBuilder builder =
+          CqlSession.builder()
+              .withCloudSecureConnectBundle(bundle)
+              .withAuthCredentials("cassandra", "cassandra")
+              .withSslContext(SSLContext.getInstance("SSL"));
+      assertThatThrownBy(() -> builder.build())
+          .isInstanceOf(IllegalStateException.class)
+          .hasMessage(
+              "Can't use withCloudSecureConnectBundle and addContactPoint(s). They are mutually exclusive.");
+    } catch (NoSuchAlgorithmException e) {
```
tomekl007: Maybe we could pick an existing `SSLContext` https://github.com/datastax/java-driver/pull/1373/files#diff-d001c2e83a104c3b75049085a3425839R235 to not have that `NoSuchAlgorithmException`? It makes this test a little bit less readable.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -523,6 +525,10 @@ public SessionT build() {
           throw new IllegalStateException(
               "Can't use withCloudSecureConnectBundle and addContactPoint(s). They are mutually exclusive.");
         }
+        if (sslConfigured) {
```
adutra: I think you should also check if SSL was configured through configuration, that is, if `datastax-java-driver.advanced.ssl-engine-factory.class` is defined.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cloud/CloudIT.java
adutra: I think the mistake was to catch `NoSuchAlgorithmException`: it would be better to simply have the test method declare to throw it imo.

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -525,7 +525,9 @@ public SessionT build() {
           throw new IllegalStateException(
               "Can't use withCloudSecureConnectBundle and addContactPoint(s). They are mutually exclusive.");
         }
-        if (sslConfigured) {
+        String configuredSSLFactory =
+            defaultConfig.getString(DefaultDriverOption.SSL_ENGINE_FACTORY_CLASS, null);
+        if (sslConfigured || configuredSSLFactory != null) {
```
adutra, line 530: `isDefined()` would look slightly better:

```
boolean sslDefined =
    defaultConfig.isDefined(DefaultDriverOption.SSL_ENGINE_FACTORY_CLASS);
if (sslConfigured || sslDefined) {
```

adutra (approved):

PR #1374, opened 2019-12-17
JAVA-2533: Refine connection errors for cloud

GregBestland:


core/src/main/java/com/datastax/oss/driver/internal/core/config/cloud/CloudConfigFactory.java
```diff
@@ -228,12 +230,22 @@ protected TrustManagerFactory createTrustManagerFactory(
...
-        new InputStreamReader(connection.getInputStream(), StandardCharsets.UTF_8));
+    try {
+      HttpsURLConnection connection = (HttpsURLConnection) metadataServiceUrl.openConnection();
+      connection.setSSLSocketFactory(sslContext.getSocketFactory());
+      connection.setRequestMethod("GET");
+      connection.setRequestProperty("host", "localhost");
+      return new BufferedReader(
+          new InputStreamReader(connection.getInputStream(), StandardCharsets.UTF_8));
+    } catch (ConnectException e) {
+      throw new IllegalStateException(
```
adutra, line 241: It's a matter of taste, but since this method already declares to throw `IOException` I think I would prefer to `throw new IOException("blah blah", e);`

core/src/main/java/com/datastax/oss/driver/internal/core/config/cloud/CloudConfigFactory.java
```diff
@@ -228,12 +230,22 @@ protected TrustManagerFactory createTrustManagerFactory(
...
-    return new BufferedReader(
-        new InputStreamReader(connection.getInputStream(), StandardCharsets.UTF_8));
+    try {
+      HttpsURLConnection connection = (HttpsURLConnection) metadataServiceUrl.openConnection();
+      connection.setSSLSocketFactory(sslContext.getSocketFactory());
+      connection.setRequestMethod("GET");
+      connection.setRequestProperty("host", "localhost");
+      return new BufferedReader(
+          new InputStreamReader(connection.getInputStream(), StandardCharsets.UTF_8));
+    } catch (ConnectException e) {
```
adutra, line 240: In theory we should have a test for this in `CloudConfigFactoryTest` but I guess it will be complex to emulate a `ConnectException`...

core/src/main/java/com/datastax/oss/driver/internal/core/config/cloud/CloudConfigFactory.java
GregBestland: We can certainly change this, however if you look over the rest of the CloudConfigFactory, typically when we run into something invalid we throw an IllegalStateException. I was trying to keep some consistency with the rest of the behavior. I think if we wanted to have consistency across the errors IOException would be inappropriate in some cases.

core/src/main/java/com/datastax/oss/driver/internal/core/config/cloud/CloudConfigFactory.java
GregBestland: It would be difficult and I think more importantly of somewhat limited utility.

adutra (approved):

PR #1376, opened 2019-12-20
JAVA-2583: IS NOT NULL clause should be idempotent

adutra:


query-builder/src/test/java/com/datastax/oss/driver/api/querybuilder/update/UpdateIdempotenceTest.java
```diff
@@ -148,4 +148,11 @@ public void should_not_be_idempotent_if_concatenating_to_collection() {
         .hasCql("UPDATE foo SET l=[1,2,3]+l WHERE k=?")
         .isNotIdempotent();
   }
+
+  @Test
+  public void should_be_idempotent_if_relation_does_not_have_right_operand() {
+    assertThat(update("foo").setColumn("col1", literal(42)).whereColumn("col2").isNotNull())
```
tomekl007, line 154: could you tell me what is a `rightOperand` in this test case?

adutra: There isn't. `IS NOT NULL` could be seen as a function of arity 1: `boolean isNotNull(x)` and thus it takes only one operand (the "left" operand). Other operators, such as `+`, are functions of arity 2: `int plus(x, y)`, which is why they have 2 operands, left (x) and right (y).

tomekl007 (approved):

olim7t (approved):

PR #1377, opened 2019-12-20
JAVA-2584: Ensure codec registry is able to create codecs for collections of UDTs and tuples

adutra:


core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -243,12 +253,6 @@ private boolean collides(TypeCodec<?> newCodec, TypeCodec<?> oldCodec) {
       }
     }
 
-    if (value instanceof TupleValue) {
```
adutra, line 246: This wasn't enough for collections of tuples or UDTs. We need to fall through and let `inspectType` + `getCachedCodec` do their jobs in all cases.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -273,15 +277,10 @@ private boolean collides(TypeCodec<?> newCodec, TypeCodec<?> oldCodec) {
...
-      return uncheckedCast(codecFor(((TupleValue) value).getType(), TupleValue.class));
-    } else if (value instanceof UdtValue) {
-      return uncheckedCast(codecFor(((UdtValue) value).getType(), UdtValue.class));
-    }
-
-    GenericType<?> javaType = inspectType(value, null);
+    DataType cqlType = inferCqlTypeFromValue(value);
```
adutra, line 280: We need to infer a CQL type here, if possible, before calling `getCachedCodec` because once inside this method the original value is lost, and it's impossible to guess the CQL type after that.

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistryTest.java
```diff
@@ -216,519 +146,131 @@ public void should_find_user_codec_for_custom_java_type() {
...
     assertThatThrownBy(() -> registry.codecFor(value))
         .isInstanceOf(IllegalArgumentException.class)
-        .hasMessage(
-            "Can't infer list codec because the first element is null "
-                + "(note that CQL does not allow null values in collections)");
+        .hasMessage(expected);
   }
 
   @Test
-  public void should_create_set_codec_for_cql_and_java_types() {
```
adutra, line 374: Tests for lists, sets and maps are now grouped together.

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistryTest.java
```diff
@@ -758,58 +298,47 @@ public void should_create_tuple_codec_for_cql_type() {
...
-    TypeCodec<TupleValue> codec = registry.codecFor(cqlType, value);
+    TypeCodec<?> codec = registry.codecFor(cqlType, value);
     assertThat(codec).isNotNull();
     assertThat(codec.accepts(cqlType)).isTrue();
     assertThat(codec.accepts(GenericType.TUPLE_VALUE)).isTrue();
     assertThat(codec.accepts(TupleValue.class)).isTrue();
     assertThat(codec.accepts(value)).isTrue();
-    inOrder.verify(mockCache).lookup(cqlType, GenericType.TUPLE_VALUE, false);
-
+    inOrder.verify(mockCache).lookup(cqlType, GenericType.of(DefaultTupleValue.class), true);
```
adutra, line 314: I think that `isJavaCovariant` should always be true when we are creating a codec for an actual object. The fact that `isJavaCovariant` was set to `false` for instances of `TupleValue` and `UdtValue` was somewhat artificial, and only possible because these two interfaces were being special-cased [here](https://github.com/datastax/java-driver/commit/8e4464e1634a97ab2e25a8de256be893d1cc24bd#diff-34d4c85a5bfd4f5ccb02c01deb989c99L246) and [here](https://github.com/datastax/java-driver/commit/8e4464e1634a97ab2e25a8de256be893d1cc24bd#diff-34d4c85a5bfd4f5ccb02c01deb989c99L276-L280).

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -273,15 +277,10 @@ private boolean collides(TypeCodec<?> newCodec, TypeCodec<?> oldCodec) {
...
+    GenericType<?> javaType = inspectType(value, cqlType);
     LOG.trace("[{}] Continuing based on inferred type {}", logPrefix, javaType);
-    return uncheckedCast(getCachedCodec(null, javaType, true));
+    return uncheckedCast(getCachedCodec(cqlType, javaType, true));
```
tomekl007, line 287: are we sure in this case that `true` should be passed as `isJavaCovariant` for all cases?

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -443,6 +442,105 @@ protected boolean matches(
     }
   }
 
+  @Nullable
+  protected DataType inferCqlTypeFromValue(@NonNull Object value) {
+    if (value instanceof List) {
+      List<?> list = (List<?>) value;
+      if (list.isEmpty()) {
+        return CQL_TYPE_FOR_EMPTY_LISTS;
```
tomekl007, line 454: why `CQL_TYPE_FOR_EMPTY_LISTS` carries `Boolean` as a type? is it safe? (same question for `Set`, `Map`)

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -443,6 +442,105 @@ protected boolean matches(
...
+      return DataTypes.SMALLINT;
+    } else if (Byte.class.equals(javaClass)) {
+      return DataTypes.TINYINT;
+    } else if (CqlDuration.class.equals(javaClass)) {
+      return DataTypes.DURATION;
+    } else if (UdtValue.class.isAssignableFrom(javaClass)) {
+      return ((UdtValue) value).getType();
+    } else if (TupleValue.class.isAssignableFrom(javaClass)) {
+      return ((TupleValue) value).getType();
+    }
```
tomekl007, line 542: nit: I would extract this method as `inferCqlPrimitiveTypeFromValue` or smth similar

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
adutra: It's explained [here](https://github.com/datastax/java-driver/pull/1377/files#diff-34d4c85a5bfd4f5ccb02c01deb989c99R682-R684): an empty collection is serialized in the same way for all element types.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
adutra: `isJavaCovariant` should always be true if we are dealing with a real object. This is because your object is likely to be something like `ArrayList<Inet4Address>` but the registry would return a codec that handles `List<InetAddress>` instead.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
adutra: Not sure I agree, most methods in this class deal with all CQL types in one single method.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -273,15 +277,10 @@ private boolean collides(TypeCodec<?> newCodec, TypeCodec<?> oldCodec) {
...
-    if (value instanceof TupleValue) {
-      return uncheckedCast(codecFor(((TupleValue) value).getType(), TupleValue.class));
-    } else if (value instanceof UdtValue) {
-      return uncheckedCast(codecFor(((UdtValue) value).getType(), UdtValue.class));
-    }
-
-    GenericType<?> javaType = inspectType(value, null);
+    DataType cqlType = inferCqlTypeFromValue(value);
+    GenericType<?> javaType = inspectType(value, cqlType);
     LOG.trace("[{}] Continuing based on inferred type {}", logPrefix, javaType);
```
olim7t: Nit: adjust this log to print both types

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -322,7 +321,7 @@ protected boolean matches(
   @NonNull
   protected GenericType<?> inspectType(@NonNull Object value, @Nullable DataType cqlType) {
```
olim7t, line 326: ~`cqlType` can no longer be null. We should adjust the annotation and remove the null checks in the method body (which means maybe the `JAVA_TYPE_FOR...` constants can go too?)~

edit: my mistake, it can still be null if the value is/contains a custom type (last return statement in `inferCqlTypeFromValue`, line 541).

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -273,15 +277,10 @@ private boolean collides(TypeCodec<?> newCodec, TypeCodec<?> oldCodec) {
...
 
-    if (value instanceof TupleValue) {
-      return uncheckedCast(codecFor(((TupleValue) value).getType(), TupleValue.class));
-    } else if (value instanceof UdtValue) {
-      return uncheckedCast(codecFor(((UdtValue) value).getType(), UdtValue.class));
-    }
-
-    GenericType<?> javaType = inspectType(value, null);
+    DataType cqlType = inferCqlTypeFromValue(value);
+    GenericType<?> javaType = inspectType(value, cqlType);
```
olim7t, line 281: It feels a bit awkward to have two separate methods here, since they both perform the same traversal of the value.

We could have a single method that constructs both CQL type and Java type at the same type, but it would force us to duplicate `inspectType`, and also we'd need a pair return type which is awkward in Java.

So all in all it's fine the way it is I guess.

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistryTest.java
```diff
@@ -22,58 +22,39 @@
...
-import java.util.UUID;
 import org.junit.Before;
 import org.junit.Test;
+import org.junit.runner.RunWith;
 import org.mockito.InOrder;
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 
+@RunWith(DataProviderRunner.class)
 public class CachingCodecRegistryTest {
```
olim7t, line 58: Refactoring the whole test is somewhat off topic.

To be fair it makes the code objectively better, but please try to isolate those kind of changes in separate commits. Even if it happens late in the PR, it's pretty easy to stash your changes (or move them to a branch), do the refactoring and commit it separately, and then rebase your main changes on top of it.

(and 👍 for doing it for the compilation warnings)

olim7t (approved):
LGTM

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistryTest.java
olim7t: Agreed, this was a side effect of the previous implementation.

PR #1378, opened 2019-12-23
JAVA-2592: Make reload support parameterizable for DefaultDriverConfigLoader

adutra:


core/src/test/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoaderTest.java
```diff
@@ -208,4 +208,23 @@ public void should_load_from_file() {
...
+    DefaultDriverConfigLoader loader =
+        new DefaultDriverConfigLoader(() -> ConfigFactory.parseString(configSource.get())) {
+          @Override
+          public boolean supportsReloading() {
+            return false;
+          }
+        };
+    CompletionStage<Boolean> stage = loader.reload();
+    assertThatStage(stage)
+        .isFailed(
```
tomekl007, line 219: isn't this breaking backward compatibility? 
If the `reload()` method is called anywhere in our driver code a client did override the `supportsReloading` to return `false` before, he will get the `UnsupportedOperationException` in their code (whereas it was not a case before this change)

adutra: Yes but if they did override `supportsReloading`, they should expect `reload` to _not_ reload anything and return a failed future, which is not the case currently.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -211,21 +211,27 @@ static ProgrammaticDriverConfigLoaderBuilder programmaticBuilder() {
...
+   * reload.
+   *
+   * <p>If so, it's also guaranteed that internal driver components have been notified by that time;
+   * note however that some react to the notification asynchronously, so they may not have
+   * completely applied all resulting changes yet.
+   *
+   * <p>If this loader does not support programmatic reloading &mdash; which you can check by
+   * calling {@link #supportsReloading()} before this method &mdash; the returned stage should fail
+   * immediately with an {@link UnsupportedOperationException}. The default implementation of this
+   * interface does support programmatic reloading however, and never returns a failed stage.
```
olim7t, line 251: Mandatory rant about off-topic cosmetic changes: the javadocs might look marginally better with paragraphs, but this produces a large diff that buries the actual change.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
```diff
@@ -90,9 +91,17 @@ public void onDriverInit(@NonNull DriverContext driverContext) {
   @NonNull
   @Override
   public CompletionStage<Boolean> reload() {
-    CompletableFuture<Boolean> result = new CompletableFuture<>();
-    RunOrSchedule.on(singleThreaded.adminExecutor, () -> singleThreaded.reload(result));
-    return result;
+    // this class does support reloading, but test #supportsReloading nonetheless in case
+    // a subclass overrides #supportsReloading.
+    if (supportsReloading()) {
```
olim7t: I don't get it. This class is internal, and not really designed for subclassing: the intended way to reuse it is to construct with a custom supplier, it's not immediately clear to me what more subclassing would allow.

Does this come from a concrete use case?

adutra: It comes from the Spring Boot layer (I can't share the link here but look there for `DataStaxDriverAutoConfiguration`). We don't want to support auto reloading since Spring configuration files do not reload. We initially subclassed `DefaultDriverConfigLoader` and overrode `supportsReloading` but we realized that this wasn't preventing reloads as advertised by the javadocs (we were expecting `reload` to return a failed future). In the end we implemented `DriverConfigLoader` from scratch.
I think the problem is a design problem. The API establishes a tight bond between `supportsReloading` and `reload`, but has no way to enforce that bond on implementors. Maybe a good compromise would be to parameterize whether or not `DefaultDriverConfig` should support reload or not.

adutra: I gave the idea outlined above a try in a separate commit. We can also chose to do nothing since the Spring module chose to re-implement `DriverConfigLoader` eventually.

adutra: Also it's worth mentioning that we do not support reloading for Spring configuration files for now, but we would like to let the users decide eventually. Some users may use reloading techniques such as the one outlined [here](https://www.baeldung.com/spring-reloading-properties). In these cases it would be nice to re-use `DefaultDriverConfigLoader` and avoid re-implementing the whole reload logic (the implementation is not really trivial).

olim7t: I see. This feels a bit unusual, but indeed it's generic enough to be supported by this class without introducing coupling with 3rd-party implementations.

> Maybe a good compromise would be to parameterize whether or not DefaultDriverConfig should support reload or not.

Yes, maybe a protected constructor that adds a `boolean allowReloading` parameter.

adutra:
@olim7t would you mind making a second review round please?

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
```diff
@@ -90,14 +113,20 @@ public void onDriverInit(@NonNull DriverContext driverContext) {
...
+      return CompletableFutures.failedFuture(
+          new UnsupportedOperationException(
+              "This instance of DefaultDriverConfigLoader does not support reloading"));
+    }
   }
 
   @Override
   public boolean supportsReloading() {
-    return true;
+    return supportsReloading;
```
tomekl007, line 129: I see a potential problem when user will override `supportsReloading` and also will use this constructor:
`public DefaultDriverConfigLoader(Supplier<Config> configSupplier, boolean supportsReloading) {`
Shouldn't we make this method final to prevent overriding?

adutra: We could, but at some point we cannot prevent users from doing all kinds of stupid things. My problem with final methods is that they cannot be mocked, so making it final could cause more harm than good eventually.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -211,21 +211,27 @@ static ProgrammaticDriverConfigLoaderBuilder programmaticBuilder() {
...
    */
   @NonNull
   CompletionStage<Boolean> reload();
 
   /**
    * Whether this implementation supports programmatic reloading with the {@link #reload()} method.
+   *
+   * <p>The default implementation of this interface does support programmatic reloading and always
```
tomekl007, line 259: By default, we are referring now to the usage of this class by calling parameterless `DefaultDriverConfigLoader()` constructor, right?

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
tomekl007: Yes, but it seems a bit misleading to me. 
In a case when the user will call the parameterless constructor or `public DefaultDriverConfigLoader(Supplier<Config> configSupplier)` the local field of the class `private final boolean supportsReloading;` will have the value `true` whereas `supportsReloading()` may be overridden and returns `false`.
For me, it looks like the object will be in the inconsistent state - value says one thing, the method other.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
adutra: Yes.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
adutra: Ok ok, let's make those methods final.

tomekl007 (approved):

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
adutra: Which is why the changes were confined to a separate commit, see bed37a7200356fa266be20c3cd8cb470e1fec76f.

olim7t: OK I missed that 👍

olim7t (approved):

PR #1379, opened 2019-12-23
JAVA-2555: Generate append/prepend constructs compatible with legacy C* versions

adutra:
I'm also proposing a fix for JAVA-2596 since it's entangled.

### DIFF
--- a/manual/query_builder/idempotence/README.md
+++ b/manual/query_builder/idempotence/README.md
@@ -116,6 +116,45 @@ Statement statement =
 assert !statement.isIdempotent();
 ```
 
+The generic `remove` method is however safe since collection removals are idempotent:
+
+```java
+Statement statement =
+    update("foo")
+        .remove("l", literal(Arrays.asList(1, 2, 3)))
+        .whereColumn("k").isEqualTo(bindMarker())
+        .build();
+// UPDATE foo SET l=l-[1,2,3] WHERE k=?
+assert !statement.isIdempotent();
+```
+
+When appending, prepending or removing a single element to/from a collection, it is possible to use 
+the dedicated methods listed below; their idempotence depends on the collection type (list, set or 
+map), the operation (append, prepend or removal) and on the idempotence of the element being 
+added/removed:
+
+1. `appendListElement` : not idempotent
+2. `prependListElement` : not idempotent
+3. `removeListElement` : idempotent if element is idempotent
+4. `appendSetElement` : idempotent if element is idempotent
+5. `prependSetElement` : idempotent if element is idempotent
+6. `removeSetElement` : idempotent if element is idempotent
+7. `appendMapElement` : idempotent if both key and value are idempotent
+8. `prependMapElement` : idempotent if both key and value are idempotent
+9. `removeMapElement` : idempotent if both key and value are idempotent
+
+For example, the following statement is idempotent:
+
+```java
+Statement statement =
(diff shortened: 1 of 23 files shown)


manual/query_builder/idempotence/README.md
```diff
@@ -99,7 +99,7 @@ Statement statement =
         .appendListElement("l", literal(1))
         .whereColumn("k").isEqualTo(bindMarker())
         .build();
-// UPDATE foo SET l+=[1] WHERE k=?
```
tomekl007, line 102: this change confuses me, I think that the goal of the ticket is to support `+=` syntax, but here instead we are changing the syntax to `value= value+...`

manual/query_builder/idempotence/README.md
```diff
@@ -116,6 +116,45 @@ Statement statement =
...
+The generic `remove` method is however safe since collection removals are idempotent:
+
+```java
+Statement statement =
+    update("foo")
+        .remove("l", literal(Arrays.asList(1, 2, 3)))
+        .whereColumn("k").isEqualTo(bindMarker())
+        .build();
+// UPDATE foo SET l=l-[1,2,3] WHERE k=?
+assert !statement.isIdempotent();
```
tomekl007: shouldn't the `statement.isIdempotent()` return true? - according to your docs:
>The generic `remove` method is however safe since collection removals are idempotent

manual/query_builder/idempotence/README.md
```diff
@@ -116,6 +116,45 @@ Statement statement =
...
+```
+
+When appending, prepending or removing a single element to/from a collection, it is possible to use 
+the dedicated methods listed below; their idempotence depends on the collection type (list, set or 
+map), the operation (append, prepend or removal) and on the idempotence of the element being 
+added/removed:
```
tomekl007, line 134: maybe we can expand more on this:
>and on the idempotence of the element being  added/removed

(explanation of element idempotence)?

manual/query_builder/idempotence/README.md
adutra: No the goal of the ticket is not to support `x += y`. This is the currently supported syntax, but it has the drawback of not being available for C* < 3.10. The goal is to use the old, alternative syntax `x = x + y`, which is more widely supported.

manual/query_builder/idempotence/README.md
adutra: Correct, thanks for pointing this out.

manual/query_builder/idempotence/README.md
adutra: Implemented in 5290e35.

tomekl007 (approved):

manual/query_builder/idempotence/README.md
```diff
@@ -155,6 +157,19 @@ Statement statement =
...
+```java
+SimpleStatement statement =
+    update("foo")
+        .removeListElement("l", function("myfunc"))
+        .whereColumn("k").isEqualTo(bindMarker())
+        .build();
+// UPDATE foo SET l=l-[myfunc()] WHERE k=?
+assert !statement.isIdempotent();
+```
+
```
olim7t, line 172: 👍 I had to refer to the javadocs to remember when terms weren't idempotent, so good thing to mention it in the manual too.

olim7t (approved):

PR #1381, opened 2019-12-26
JAVA-2528: Store suppressed exceptions in AllNodesFailedException

adutra:


tomekl007 (approved):

core/src/main/java/com/datastax/oss/driver/api/core/AllNodesFailedException.java
```diff
@@ -31,7 +31,8 @@
 
 /**
  * Thrown when a query failed on all the coordinators it was tried on. This exception may wrap
- * multiple errors, use {@link #getAllErrors()} to inspect individual problems on each node.
+ * multiple {@linkplain #getSuppressed() suppressed exceptions}, use {@link #getAllErrors()} to
+ * inspect individual problems on each node.
```
olim7t: Suggestion: "This exception may wrap multiple errors, that are available either as suppressed exceptions, or via getAllErrors() where they are grouped by node".

olim7t (approved):
LGTM.

Overall I think it's a good change, but I'd just like to emphasize the fact that stack traces include suppressed exceptions. Here's an
[example](https://gist.github.com/olim7t/1a50f307939c51a13275ef8dcf35fff6) with bad credentials on a 3-node cluster.

One risk is that we get backlash because stack traces are getting too long. But I think with typical deployments (number of contact points, number of retries triggered by the policy), the number of child exceptions should remain low. Also, our client-side representations of server errors (`UnavailableException`, etc.) don't fill their stack trace, so they won't take too much space.

tomekl007:
> LGTM.
> 
> Overall I think it's a good change, but I'd just like to emphasize the fact that stack traces include suppressed exceptions. Here's an
> [example](https://gist.github.com/olim7t/1a50f307939c51a13275ef8dcf35fff6) with bad credentials on a 3-node cluster.
> 
> One risk is that we get backlash because stack traces are getting too long. But I think with typical deployments (number of contact points, number of retries triggered by the policy), the number of child exceptions should remain low. Also, our client-side representations of server errors (`UnavailableException`, etc.) don't fill their stack trace, so they won't take too much space.

Maybe we could add some kind of a threshold parameter (similar to dsbulk) for setting the upper bound of number of suppressed exceptions?

adutra:
> Maybe we could add some kind of a threshold parameter (similar to dsbulk) for setting the upper bound of number of suppressed exceptions?

Valid concern, but let's wait until someone complains before making this class any more complex.

PR #1383, opened 2019-12-26
JAVA-2530: Expose shortcuts for name-based UUIDs

adutra:


core/src/main/java/com/datastax/oss/driver/api/core/uuid/Uuids.java
```diff
@@ -202,13 +235,188 @@ private static long makeClockSeqAndNode() {
...
+    // clear and set version
+    data[6] &= (byte) 0x0f;
+    data[6] |= (byte) (version << 4);
+    // clear and set variant to IETF
+    data[8] &= (byte) 0x3f;
+    data[8] |= (byte) 0x80;
+    return fromBytes(data);
+  }
+
+  private static UUID fromBytes(byte[] data) {
```
adutra, line 393: I wonder if it would be useful to expose this method and the one below as generic utilities to convert to/from UUID and byte arrays?

core/src/main/java/com/datastax/oss/driver/api/core/uuid/Uuids.java
```diff
@@ -202,13 +235,188 @@ private static long makeClockSeqAndNode() {
...
+   * @param namespace The namespace UUID to use; cannot be null.
+   * @param name The name part; cannot be null.
+   * @throws NullPointerException if <code>namespace</code> or <code>name</code> is null.
+   * @throws IllegalStateException if the {@link MessageDigest} algorithm for version 3 (MD5) is not
+   *     available on this platform.
+   */
+  @NonNull
+  public static UUID nameBased(@NonNull UUID namespace, @NonNull String name) {
+    Objects.requireNonNull(name, "name cannot be null");
+    return nameBased(namespace, name.getBytes(StandardCharsets.UTF_8));
```
tomekl007, line 260: nit: `StandardCharsets.UTF_8` can be moved to constant

changelog/README.md
```diff
@@ -4,6 +4,7 @@
 
 ### 4.4.0 (in progress)
 
+- [improvement] JAVA-2530: Expose shortcuts for version 3 UUIDs
```
tomekl007: why do we need to support both 3 and 4 versions of UUIDs in the same driver version?

core/src/main/java/com/datastax/oss/driver/api/core/uuid/Uuids.java
adutra: It's already a constant, so why would you create another constant?

changelog/README.md
adutra: We don't need to, we are just trying to be nice to users and provide them with a user-friendly API to create most kinds of UUIDs. We already had methods to create time-based UUIDs (version 1) and random UUIDs (version 4), now we are also offering methods to create name-based UUIDs (versions 3 and 5).

core/src/main/java/com/datastax/oss/driver/api/core/uuid/Uuids.java
```diff
@@ -202,13 +235,188 @@ private static long makeClockSeqAndNode() {
...
+      lsb = (lsb << 8) | (data[i] & 0xff);
+    }
+    return new UUID(msb, lsb);
+  }
+
+  private static byte[] toBytes(UUID uuid) {
+    byte[] out = new byte[16];
+    long msb = uuid.getMostSignificantBits();
+    for (int i = 0; i < 8; i++) {
+      out[i] = (byte) ((msb >> ((7 - i) * 8)) & 0xff);
```
olim7t: Nit: I think the mask is unnecessary, when you cast to `byte` the higher-order bytes will be discarded.
```suggestion
      out[i] = (byte) (msb >> ((7 - i) * 8));
```

olim7t (approved):
LGTM.
I was a bit hesitant since this is not a core driver feature, but it feels legitimate since we offer more than the standard JDK methods, in particular type 5.

core/src/main/java/com/datastax/oss/driver/api/core/uuid/Uuids.java
olim7t: IMO this is too low-level and specific to be provided as a driver feature.

core/src/main/java/com/datastax/oss/driver/api/core/uuid/Uuids.java
tomekl007: I meant to create a field in this class and reference it, but you are right.

PR #1385, opened 2020-01-03
JAVA-2541: Improve error messages during connection initialization

adutra:


core/src/main/java/com/datastax/oss/driver/internal/core/channel/ProtocolInitHandler.java
```diff
@@ -140,6 +141,8 @@ protected boolean setConnectSuccess() {
     // This class is a finite-state automaton, that sends a different query depending on the step
     // in the initialization sequence.
     private Step step;
+    private int stepNumber = 0;
```
tomekl007, line 144: just to double-check, shouldn't this be `volatile`

adutra: No, because you are inside a class that is only accessed from the event loop, thus it can only be accessed by one thread at a time.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelHandlerRequest.java
```diff
@@ -71,7 +71,9 @@ private void writeListener(Future<? super Void> writeFuture) {
       timeoutFuture =
           channel.eventLoop().schedule(this::onTimeout, timeoutMillis, TimeUnit.MILLISECONDS);
     } else {
-      fail(describe() + ": error writing ", writeFuture.cause());
+      String message =
+          String.format("%s: failed to send request (%s)", describe(), writeFuture.cause());
```
adutra, line 75: I'm adding the cause in the message string because these errors usually appear inside an `AllNodesFailedException`, and thus we don't usually see their cause printed in logs.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ProtocolInitHandler.java
```diff
@@ -150,29 +153,36 @@ protected boolean setConnectSuccess() {
 
     @Override
     String describe() {
-      return "[" + logPrefix + "] init query " + step;
+      return String.format(
+          "[%s] Protocol initialization request, step %d (%s)", logPrefix, stepNumber, request);
```
olim7t, line 157: For the record, `AuthResponse#toString()` doesn't include the token, so no risk of leaking sensitive information in the logs.

olim7t (approved):
👍
I noticed the "unexpected error at step Xxx" in some user logs and had made a note to include the cause, this will take care of it.

PR #1386, opened 2020-01-06
JAVA-2578 Fix ConnectIT when run from dse driver

GregBestland:
Default LBP needs to be set explicitly in order to work from the dse-driver's nightly runs


integration-tests/src/test/java/com/datastax/oss/driver/core/ConnectIT.java
```diff
@@ -131,6 +132,8 @@ public void should_cleanup_on_lbp_init_failure() {
     DriverConfigLoader loader =
         SessionUtils.configLoaderBuilder()
             .without(DefaultDriverOption.LOAD_BALANCING_LOCAL_DATACENTER)
+            .withClass(
```
adutra: This would force the DSE driver to use the OSS LBP. I think instead we would like to make this test work with the DSE driver using its default LBP (DseLoadBalancingPolicy).

From what I see we could do this but the incriminated test is creating a hard-coded `CqlSession` line 137. This should be changed to `SessionUtils.baseBuilder()`.

GregBestland:
PR comments addressed, I also targeted this against the oss branch.
See corresponding DSE driver PR https://github.com/riptano/java-dse-driver/pull/337

olim7t (approved):

PR #1387, opened 2020-01-15
JAVA-2600: Introduce map-backed config loader

olim7t:
This is a proposal for a more straightforward approach to programmatic configuration: no files, rely only on maps.

Sample usage:
```java
// This creates a configuration equivalent to the built-in reference.conf:
Map<String, Map<DriverOption, Object>> optionsMap =
    DriverConfigLoader.buildDefaultOptionsMap();

// Customize an option:
optionsMap
  .get(DriverExecutionProfile.DEFAULT_NAME)
  .put(DefaultDriverOption.REQUEST_TIMEOUT, Duration.ofSeconds(5));

DriverConfigLoader loader = DriverConfigLoader.fromMap(optionsMap);
CqlSession session = CqlSession.builder().withConfigLoader(loader).build();
```

See the javadocs of the new method `DriverConfigLoader.fromMap` for more details.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverConfigLoader.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverConfigLoader.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.internal.core.config.map;
+
+import com.datastax.oss.driver.api.core.config.DriverConfig;
+import com.datastax.oss.driver.api.core.config.DriverConfigLoader;
+import com.datastax.oss.driver.api.core.config.DriverOption;
+import com.datastax.oss.driver.api.core.config.OptionsMap;
+import com.datastax.oss.driver.api.core.context.DriverContext;
+import com.datastax.oss.driver.internal.core.config.ConfigChangeEvent;
+import com.datastax.oss.driver.internal.core.context.EventBus;
+import com.datastax.oss.driver.internal.core.context.InternalDriverContext;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Consumer;
+
+public class MapBasedDriverConfigLoader implements DriverConfigLoader, Consumer<OptionsMap> {
+
+  @NonNull private final OptionsMap source;
+  @NonNull private final Map<String, Map<DriverOption, Object>> rawMap;
+  private volatile EventBus eventBus;
+
+  public MapBasedDriverConfigLoader(
+      @NonNull OptionsMap source, @NonNull Map<String, Map<DriverOption, Object>> rawMap) {
+    this.source = source;
+    this.rawMap = rawMap;
+  }
+
+  @NonNull
+  @Override
+  public DriverConfig getInitialConfig() {
+    return new MapBasedDriverConfig(rawMap);
+  }
+
+  @Override
+  public void onDriverInit(@NonNull DriverContext context) {
+    eventBus = ((InternalDriverContext) context).getEventBus();
+    source.addChangeListener(this);
+  }
+
+  @Override
+  public void accept(OptionsMap map) {
+    assert eventBus != null; // listener is registered after setting this field
+    eventBus.fire(ConfigChangeEvent.INSTANCE);
+  }
+
+  @NonNull
+  @Override
+  public CompletionStage<Boolean> reload() {
+    return CompletableFuture.completedFuture(true);
+  }
+
+  @Override
+  public boolean supportsReloading() {
+    return true;
+  }
+
+  @Override
+  public void close() {
+    source.removeChangeListener(this);
+  }
+}
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverExecutionProfile.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverExecutionProfile.java
@@ -1,0 +1,332 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.internal.core.config.map;
+
+import com.datastax.oss.driver.api.core.config.DriverExecutionProfile;
+import com.datastax.oss.driver.api.core.config.DriverOption;
+import com.datastax.oss.driver.shaded.guava.common.base.Preconditions;
+import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
+import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableMap;
+import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableSortedSet;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.time.Duration;
+import java.util.AbstractMap;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.SortedSet;
+
+/** @see MapBasedDriverConfigLoader */
+public class MapBasedDriverExecutionProfile implements DriverExecutionProfile {
+
+  private static final Object NO_VALUE = new Object();
+
+  private final String profileName;
+  // Anything that was overridden in a derived profile with `withXxx` methods. Empty for non-derived
+  // profiles
+  private final Map<DriverOption, Object> overrides;
+  // The backing map for the current profile
+  private final Map<DriverOption, Object> profile;
+  // The backing map for the default profile (if the current one is not the default)
+  private final Map<DriverOption, Object> defaultProfile;
+
+  public MapBasedDriverExecutionProfile(
+      Map<String, Map<DriverOption, Object>> optionsMap, String profileName) {
+    this(
+        profileName,
+        Collections.emptyMap(),
+        optionsMap.get(profileName),
+        profileName.equals(DriverExecutionProfile.DEFAULT_NAME)
+            ? Collections.emptyMap()
+            : optionsMap.get(DriverExecutionProfile.DEFAULT_NAME));
+    Preconditions.checkArgument(
+        optionsMap.containsKey(profileName),
+        "Unknown profile '%s'. Check your configuration.",
+        profileName);
+  }
+
+  public MapBasedDriverExecutionProfile(
+      String profileName,
+      Map<DriverOption, Object> overrides,
+      Map<DriverOption, Object> profile,
+      Map<DriverOption, Object> defaultProfile) {
+    this.profileName = profileName;
+    this.overrides = overrides;
+    this.profile = profile;
+    this.defaultProfile = defaultProfile;
+  }
+
+  @NonNull
+  @Override
+  public String getName() {
+    return profileName;
+  }
+
+  @Override
+  public boolean isDefined(@NonNull DriverOption option) {
+    if (overrides.containsKey(option)) {
+      return overrides.get(option) != NO_VALUE;
+    } else {
+      return profile.containsKey(option) || defaultProfile.containsKey(option);
+    }
+  }
+
+  // Driver options don't encode the type, everything relies on the user putting the right types in
+  // the backing map, so no point in trying to type-check.
+  @SuppressWarnings({"unchecked", "TypeParameterUnusedInFormals"})
+  @NonNull
+  private <T> T get(@NonNull DriverOption option) {
+    Object value =
+        overrides.getOrDefault(option, profile.getOrDefault(option, defaultProfile.get(option)));
+    if (value == null || value == NO_VALUE) {
+      throw new IllegalArgumentException("Missing configuration option " + option.getPath());
+    }
+    return (T) value;
+  }
+
+  @Override
+  public boolean getBoolean(@NonNull DriverOption option) {
+    return get(option);
+  }
+
+  @NonNull
+  @Override
+  public List<Boolean> getBooleanList(@NonNull DriverOption option) {
+    return get(option);
+  }
+
+  @Override
+  public int getInt(@NonNull DriverOption option) {
+    return get(option);
+  }
+
+  @NonNull
+  @Override
+  public List<Integer> getIntList(@NonNull DriverOption option) {
+    return get(option);
+  }
+
+  @Override
+  public long getLong(@NonNull DriverOption option) {
+    return get(option);
+  }
+
+  @NonNull
+  @Override
+  public List<Long> getLongList(@NonNull DriverOption option) {
+    return get(option);
+  }
+
+  @Override
+  public double getDouble(@NonNull DriverOption option) {
+    return get(option);
+  }
+
+  @NonNull
+  @Override
+  public List<Double> getDoubleList(@NonNull DriverOption option) {
+    return get(option);
+  }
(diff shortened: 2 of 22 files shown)


core/src/test/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverConfigLoaderTest.java
```diff
@@ -0,0 +1,71 @@
...
+    defaultProfileMap.put(MockOptions.INT1, 2);
+    assertThat(config.getDefaultProfile().getInt(MockOptions.INT1)).isEqualTo(2);
+  }
+
+  /**
+   * Checks that, if we ask to pre-fill the default profile, then we get the same set of options as
+   * the built-in reference.conf.
+   */
+  @Test
+  public void should_fill_default_profile_like_reference_file() {
```
olim7t, line 51: This test will protect us against `buildDefaultOptionsMap()` getting out of sync with `reference.conf`.

emerkle826: This is a very good idea. :+1:

core/src/main/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverConfigLoader.java
```diff
@@ -0,0 +1,199 @@
...
+  public void close() {
+    // nothing to do
+  }
+
+  public static Map<String, Map<DriverOption, Object>> buildDefaultOptionsMap() {
+    Map<String, Map<DriverOption, Object>> optionsMap = new ConcurrentHashMap<>();
+    Map<DriverOption, Object> defaultProfileMap = new ConcurrentHashMap<>();
+    optionsMap.put(DriverExecutionProfile.DEFAULT_NAME, defaultProfileMap);
+
+    // Skip CONFIG_RELOAD_INTERVAL because the map-based config doesn't need periodic reloading
```
tomekl007: I have a concern about keeping those defaults here. Up from this point, we will need to manage two places with default settings - this class and `reference.conf`.  Isn't there a way to fill those default from `reference.conf` instead of hard-coding this?

core/src/main/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverExecutionProfile.java
```diff
@@ -0,0 +1,329 @@
...
+  // Anything that was overridden in a derived profile with `withXxx` methods. Empty for non-derived
+  // profiles
+  private final Map<DriverOption, Object> overrides;
+  // The backing map for the current profile
+  private final Map<DriverOption, Object> profile;
+  // The backing map for the default profile (if the current one is not the default)
+  private final Map<DriverOption, Object> defaultProfile;
+
+  public MapBasedDriverExecutionProfile(
+      Map<String, Map<DriverOption, Object>> optionsMap, String name) {
```
tomekl007: nit: I would rename `name` to `profileName` to make it more descriptive

core/src/main/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverExecutionProfile.java
```diff
@@ -0,0 +1,329 @@
...
+  @NonNull
+  @Override
+  public String getName() {
+    return name;
+  }
+
+  @Override
+  public boolean isDefined(@NonNull DriverOption option) {
+    if (overrides.containsKey(option)) {
+      return overrides.get(option) != NO_VALUE;
```
tomekl007, line 81: I don't get why we need to check against this special marker `NO_VALUE;` could you elaborate on that?

core/src/test/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverConfigTest.java
```diff
@@ -0,0 +1,103 @@
...
+    assertThat(base.getInt(MockOptions.INT2)).isEqualTo(43);
+    assertThat(derived.isDefined(MockOptions.INT2)).isFalse();
+  }
+
+  private void addOption(String profile, DriverOption option, Object value) {
+    addProfile(profile).put(option, value);
+  }
+
+  private Map<DriverOption, Object> addProfile(String profile) {
+    return optionsMap.computeIfAbsent(profile, name -> new ConcurrentHashMap<>());
```
tomekl007: Those two methods are nice encapsulation of profile and option. The current implementation of `MapBasedDriverConfig` makes user to thing about that nested `<Map<String,<Map, Object>` every time there is an interaction with this config loader. Maybe we could include similar methods in the `MapBasedDriverConfig` to allow users for simpler interaction with a loader?  
Or maybe instead of:
``` 
public MapBasedDriverExecutionProfile(
 Map<String, Map<DriverOption, Object>> optionsMap, String name) {
```
we could create a factory method: 
```
Map<DriverOption, Object> forProfile(String profileName)
```
that returns map only for a specific profile and does management of it underneath? wdyt?

core/src/main/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverConfigLoader.java
olim7t: I'd like to avoid a dependency to Typesafe config at runtime, i.e. the user can exclude it from their classpath if they use the map-based loader.

The unit test I commented on above ensures that we don't forget to keep the two places in sync: the build will fail if both sides don't match.

In fact I already tested that when I forgot to add the new `LOAD_BALANCING_POLICY_SLOW_AVOIDANCE` option when rebasing my PR.

core/src/main/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverExecutionProfile.java
olim7t: For an option that exists in the original profile, this allows us to distinguish between those two cases:
1. the option was not overridden
2. the option was removed with `without()`

See `MapBasedDriverConfigTest.should_create_derived_profile_unsetting_option()`.

core/src/test/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverConfigTest.java
olim7t: > The current implementation of `MapBasedDriverConfig` makes user to think about that nested `Map<String,<Map, Object>` every time there is an interaction with this config loader.

Yes, that map is all users should think about. The intent of this loader is to bare metal, using the simplest thing that could possibly work: if you know how to use maps, you know how to use the loader.
I don't want to introduce another API layer on top of the map, because as soon as we do that we open the door for confusion (as we saw with even trivial methods like `startProfile()` and `endProfile()` in the programmatic loader).

One downside is that building nested maps is not as user-friendly as it could be, but this loader is intended for tool builders, which is a very small minority of our target audience.

As for this particular unit test, I'll admit that the code is not very elegant (particularly relying on a class-level field), I went for quick and dirty because it's only 50 or so lines of test code. Creating some sort of `NestedMapBuilder` just for that feels overkill.

core/src/main/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverConfigLoader.java
tomekl007: nice idea with that unit test

olim7t:
Marking WIP because I want to revisit this PR when time allows.

Type safety would be really nice to have. We could achieve it with a separate option type that wraps the original one. This means an entire new set of options to maintain in parallel. I need to take a stab at it to see if it's reasonable or not.

adutra:
One idea that came to me when reviewing this PR is: I am not sure that `Map<DriverOption, Object>` is the best signature for the options map, maybe `Map<String, Object>` or `Map<Object, Object>`. The reason is because users may create components that read the configuration in a dynamic way, and the property keys may not be all known at compile time. Using String or Object makes it easier for users to pass configuration options programmatically for such components, whereas `DriverOption` would force them to create implementations of that interface first (if at all possible).

olim7t:
> Using String or Object makes it easier for users to pass configuration options programmatically for such components, whereas `DriverOption` would force them to create implementations of that interface first (if at all possible).

The driver uses `DriverOption` to read the configuration, so you have to create `DriverOption` instances if you want the driver to be able to read them via `DriverConfig`.

olim7t:
> Type safety would be really nice to have. We could achieve it with a separate option type that wraps the original one. This means an entire new set of options to maintain in parallel.

I pushed a commit that implements that idea. Preview of the new API:

```java
// Now a custom type instead of a raw map:
OptionsMap map = OptionsMap.driverDefaults();

// Note the new TypedDriverOption: this will fail if the value is not a Duration.
map.put(TypedDriverOption.REQUEST_TIMEOUT, Duration.ofSeconds(5));

// This is the syntax for profiles:
map.put("slow", TypedDriverOption.REQUEST_TIMEOUT, Duration.ofSeconds(30));

DriverConfigLoader loader = DriverConfigLoader.fromMap(map);
```

As expected, `TypedDriverOption` duplicates every option. This was unavoidable since there's no practical way to add type information to `DriverOption`. I think maintenance won't be that bad because we don't add options very often, and I've added a unit test that will fail if we add a new `DriverOption` but forget the corresponding `TypedDriverOption`.
The only real downside is that this might introduce confusion. `TypedDriverOption` is only used with `OptionsMap`, everywhere else it's still `DriverOption`. That being said, that "everywhere else" is mostly driver internals.

I was initially unsure whether to proceed with this, but typed options cover a useful case for which we had no solution before: injecting a set of text-based options in a generic manner (the option type tells you how to parse the value). For example if a client has its configuration in a properties file, and it wants to use the convention that every property starting with `driver.` translates to a driver option.

core/src/main/java/com/datastax/oss/driver/internal/core/config/map/DefaultOptionsMap.java
```diff
@@ -0,0 +1,189 @@
...
+    source.put(TypedDriverOption.REPREPARE_ENABLED, true);
+    source.put(TypedDriverOption.REPREPARE_CHECK_SYSTEM_TABLE, false);
+    source.put(TypedDriverOption.REPREPARE_MAX_STATEMENTS, 0);
+    source.put(TypedDriverOption.REPREPARE_MAX_PARALLELISM, 100);
+    source.put(TypedDriverOption.REPREPARE_TIMEOUT, Duration.ofMillis(500));
+    source.put(TypedDriverOption.NETTY_DAEMON, false);
+    source.put(TypedDriverOption.NETTY_IO_SIZE, 0);
+    source.put(TypedDriverOption.NETTY_IO_SHUTDOWN_QUIET_PERIOD, 2);
+    source.put(TypedDriverOption.NETTY_IO_SHUTDOWN_TIMEOUT, 15);
+    source.put(TypedDriverOption.NETTY_IO_SHUTDOWN_UNIT, "SECONDS");
```
tomekl007: nit, maybe use a static constant for repeatable values like for example `SECONDS`

core/src/main/java/com/datastax/oss/driver/internal/core/config/map/DefaultOptionsMap.java
```diff
@@ -0,0 +1,189 @@
...
+    Objects.requireNonNull(profile, "profile");
+    return map.computeIfAbsent(profile, p -> new ConcurrentHashMap<>());
+  }
+
+  // Isolate the suppressed warning for retrieval. The cast should always succeed unless the user
+  // messes with asMap() directly.
+  @SuppressWarnings({"unchecked", "TypeParameterUnusedInFormals"})
+  @Nullable
+  private <ValueT> ValueT cast(@Nullable Object value) {
+    return (ValueT) value;
```
tomekl007: I think that we are missing a test case that demonstrates type-safety of this new API (i.e: trying to put and retrieve using the different type)

core/src/main/java/com/datastax/oss/driver/api/core/config/map/OptionsMap.java
```diff
@@ -0,0 +1,64 @@
...
+  static OptionsMap driverDefaults() {
+    return DefaultOptionsMap.driverDefaults();
+  }
+
+  @Nullable
+  <ValueT> ValueT put(
+      @NonNull String profile, @NonNull TypedDriverOption<ValueT> option, @NonNull ValueT value);
+
+  @Nullable
+  default <ValueT> ValueT put(@NonNull TypedDriverOption<ValueT> option, @NonNull ValueT value) {
```
tomekl007: I wonder how this new API plays with @adutra question about the support of building map settings dynamically. Right now it seems that we are no longer exposing raw API  (to allow users to just add `object, object`). I see that we left the method `asRaw()` as a back-door, will this be enough?

adutra:
> The driver uses `DriverOption` to read the configuration, so you have to create `DriverOption` instances if you want the driver to be able to read them via `DriverConfig`.

Yes, the idea was to have the config loader create those instances for each string/object key, so that users don't need to do that in their applications.

core/src/main/java/com/datastax/oss/driver/internal/core/config/map/DefaultOptionsMap.java
```diff
@@ -0,0 +1,189 @@
...
+  private Map<DriverOption, Object> getProfileMap(@NonNull String profile) {
+    Objects.requireNonNull(profile, "profile");
+    return map.computeIfAbsent(profile, p -> new ConcurrentHashMap<>());
+  }
+
+  // Isolate the suppressed warning for retrieval. The cast should always succeed unless the user
+  // messes with asMap() directly.
+  @SuppressWarnings({"unchecked", "TypeParameterUnusedInFormals"})
+  @Nullable
+  private <ValueT> ValueT cast(@Nullable Object value) {
```
adutra: We might be able to improve things a bit:

```java
@Nullable
private <ValueT> ValueT cast(@NonNull GenericType<ValueT> expectedType, @Nullable Object value) {
  if (value == null) {
    return null;
  }
  if (expectedType.isSupertypeOf(GenericType.of(value.getClass()))) {
    ValueT cast = (ValueT) value;
    return cast;
  }
  // The cast should always succeed unless the user messes with asMap() directly.
  throw new ClassCastException("Cannot cast blah blah");
}
```

core/src/main/java/com/datastax/oss/driver/api/core/config/map/OptionsMap.java
```diff
@@ -0,0 +1,64 @@
...
+package com.datastax.oss.driver.api.core.config.map;
+
+import com.datastax.oss.driver.api.core.config.DriverExecutionProfile;
+import com.datastax.oss.driver.api.core.config.DriverOption;
+import com.datastax.oss.driver.internal.core.config.map.DefaultOptionsMap;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.Map;
+
+public interface OptionsMap {
```
adutra: Can we have overloaded methods for the default profile?

core/src/main/java/com/datastax/oss/driver/internal/core/config/map/DefaultOptionsMap.java
olim7t: Such a test won't compile.

core/src/main/java/com/datastax/oss/driver/api/core/config/map/OptionsMap.java
olim7t: The config is not meant to support fully dynamic options. The driver internals use `DriverOption` to retrieve the values, so you need pre-existing instances for every option.

`asRawMap()` is not a backdoor, it's only intended to pass the options to the `DriverConfig` implementation. If I could hide it I would.

core/src/main/java/com/datastax/oss/driver/internal/core/config/map/DefaultOptionsMap.java
olim7t: How is that better? As far as I can tell, it does exactly the same thing as the cast operator.

core/src/main/java/com/datastax/oss/driver/api/core/config/map/OptionsMap.java
olim7t: You mean for put/remove/etc? There are default methods on the interface that do that.

olim7t:
> Yes, the idea was to have the config loader create those instances for each string/object key, so that users don't need to do that in their applications.

If someone creates custom options, it's because they have custom code that reads those options:
```java
public enum MyDriverOption implements DriverOption {
  SOME_OPTION("...");
}
public class DefaultLoadBalancingPolicy implements LoadBalancingPolicy {
  ...
  config.getBoolean(MyDriverOption.SOME_OPTION);
  ...
}
```
If the loader creates the instances, what would the client code pass to methods like `getBoolean`?

core/src/main/java/com/datastax/oss/driver/api/core/config/map/OptionsMap.java
olim7t: `asRawMap()` is clunky ☹️ 
I'll try to revisit that, but it's tricky because `MapBasedDriverConfig/ExecutionProfile` does lookups by `DriverOption`, and I need a live view of the `OptionsMap`.

core/src/main/java/com/datastax/oss/driver/api/core/config/map/OptionsMap.java
adutra: Sorry I missed them.

adutra (approved):

PR #1388, opened 2020-01-20
JAVA-2618: Improve error handling in request handlers

olim7t:


core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
```diff
@@ -854,7 +866,15 @@ public void onFailure(Throwable error) {
...
         decision = RetryDecision.RETHROW;
       } else {
-        decision = retryPolicy.onRequestAborted(statement, error, retryCount);
+        try {
+          decision = retryPolicy.onRequestAborted(statement, error, retryCount);
+        } catch (Throwable cause) {
+          setFinalError(
+              new IllegalStateException("Unexpected error while invoking the retry policy", cause),
+              null,
+              execution);
```
olim7t, line 875: Other invocations of the retry policy are already protected by similar try/catch blocks. This one would propagate directly to `InFlightHandler`, I think it's better to handle it here.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/InFlightHandler.java
```diff
@@ -356,20 +350,27 @@ private void abortAllInFlight(DriverException cause) {
...
       // closing the channel
     }
   }
 
+  private void fail(ResponseCallback callback, Throwable failure) {
+    try {
+      callback.onFailure(failure);
+    } catch (Throwable throwable) {
+      // Protect against unexpected errors. We don't have anywhere to report the error (since
+      // onFailure failed), so log as a last resort.
```
olim7t, line 369: Note that all callbacks also have some kind of timeout, so they will eventually fail.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/InFlightHandler.java
```diff
@@ -356,20 +350,27 @@ private void abortAllInFlight(DriverException cause) {
...
+      LOG.error("[{}] Unexpected error while failing {}", logPrefix, callback, throwable);
```
olim7t, line 370: The level was WARN in the few places where this was already done, but I raised to ERROR because it's a serious issue.

PR #1389, opened 2020-01-20
JAVA-2614: Provide random paging as a built-in feature

olim7t:
Opening early because the implementation and tests are ready, I still need to work on the docs.

TODO:
- [x] javadocs
- [x] manual
- [x] rename to `OffsetPager`?

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/paging/OffsetPager.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/paging/OffsetPager.java
@@ -1,0 +1,319 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core.paging;
+
+import com.datastax.oss.driver.api.core.AsyncPagingIterable;
+import com.datastax.oss.driver.api.core.MappedAsyncPagingIterable;
+import com.datastax.oss.driver.api.core.PagingIterable;
+import com.datastax.oss.driver.api.core.cql.ResultSet;
+import com.datastax.oss.driver.api.core.cql.Statement;
+import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+import net.jcip.annotations.ThreadSafe;
+
+/**
+ * A utility to emulate offset queries on the client side (<b>this comes with important performance
+ * trade-offs, make sure you read and understand the whole javadocs before using this class</b>).
+ *
+ * <p>Web UIs and services often provide paginated results with random access, for example: given a
+ * page size of 20 elements, fetch page 5. Cassandra does not support this natively (see <a
+ * href="https://issues.apache.org/jira/browse/CASSANDRA-6511">CASSANDRA-6511</a>), because such
+ * queries are inherently linear: the database would have to restart from the beginning every time,
+ * and skip unwanted rows until it reaches the desired offset.
+ *
+ * <p>However, random pagination is a real need for many applications, and linear performance can be
+ * a reasonable trade-off if the cardinality stays low. This class provides a way to emulate this
+ * behavior on the client side.
+ *
+ * <h3>Performance considerations</h3>
+ *
+ * For each page that you want to retrieve:
+ *
+ * <ul>
+ *   <li>you need to re-execute the query, in order to start with a fresh result set;
+ *   <li>this class starts iterating from the beginning, and skips rows until it reaches the desired
+ *       offset.
+ * </ul>
+ *
+ * <br>
+ *
+ * <pre>
+ * String query = "SELECT ...";
+ * OffsetPager pager = new OffsetPager(20);
+ *
+ * // Get page 2: start from a fresh result set, throw away rows 1-20, then return rows 21-40
+ * ResultSet rs = session.execute(query);
+ * OffsetPager.Page&lt;Row&gt; page2 = pager.getPage(rs, 2);
+ *
+ * // Get page 5: start from a fresh result set, throw away rows 1-80, then return rows 81-100
+ * rs = session.execute(query);
+ * OffsetPager.Page&lt;Row&gt; page5 = pager.getPage(rs, 5);
+ * </pre>
+ *
+ * <h3>Establishing application-level guardrails</h3>
+ *
+ * Linear performance should be fine for the values typically encountered in real-world
+ * applications: for example, if the page size is 25 and users never go past page 10, the worst case
+ * is only 250 rows, which is a very small result set. However, we strongly recommend that you
+ * implement hard limits in your application code: if the page number is exposed to the user (for
+ * example if it is passed as a URL parameter), make sure it is properly validated and enforce a
+ * maximum, so that an attacker can't inject a large value that could potentially fetch millions of
+ * rows.
+ *
+ * <h3>Relation with protocol-level paging</h3>
+ *
+ * Protocol-level paging refers to the ability to split large response into multiple network chunks:
+ * see {@link Statement#setPageSize(int)} and {@code basic.request.page-size} in the configuration.
+ * It happens under the hood, and is completely transparent for offset paging: this class will work
+ * the same no matter how many network roundtrips were needed to fetch the result. You don't need to
+ * set the protocol page size and the logical page size to the same value.
+ */
+@ThreadSafe
+public class OffsetPager {
+
+  /** A page returned as the result of an offset query. */
+  public interface Page<ElementT> {
+
+    /** The elements in the page. */
+    @NonNull
+    List<ElementT> getElements();
+
+    /**
+     * The page number (1 for the first page, 2 for the second page, etc).
+     *
+     * <p>Note that it may be different than the number you passed to {@link
+     * #getPage(PagingIterable, int)}: if the result set was too short, this is the actual number of
+     * the last page.
+     */
+    int getPageNumber();
+
+    /** Whether this is the last page in the result set. */
+    boolean isLast();
+  }
+
+  private final int pageSize;
+
+  /**
+   * Creates a new instance.
+   *
+   * @param pageSize the number of elements per page. Must be greater than or equal to 1.
+   */
+  public OffsetPager(int pageSize) {
+    if (pageSize < 1) {
+      throw new IllegalArgumentException("Invalid pageSize, expected >=1, got " + pageSize);
+    }
+    this.pageSize = pageSize;
+  }
+
+  /**
+   * Extracts a page from a synchronous result set, by skipping rows until we get to the requested
+   * offset.
+   *
+   * @param iterable the iterable to extract the results from: typically a {@link ResultSet}, or a
+   *     {@link PagingIterable} returned by the mapper.
+   * @param targetPageNumber the page to return (1 for the first page, 2 for the second page, etc).
+   *     Must be greater than or equal to 1.
+   * @return the requested page, or the last page if the requested page was past the end of the
+   *     iterable.
+   * @throws IllegalArgumentException if the conditions on the arguments are not respected.
+   */
+  @NonNull
+  public <ElementT> Page<ElementT> getPage(
+      @NonNull PagingIterable<ElementT> iterable, final int targetPageNumber) {
+
+    throwIfIllegalArguments(iterable, targetPageNumber);
+
+    // Holds the contents of the target page. We also need to record the current page as we go,
+    // because our iterable is forward-only and we can't predict when we'll hit the end.
+    List<ElementT> currentPageElements = new ArrayList<>();
+
+    int currentPageNumber = 1;
+    int currentPageSize = 0;
+    for (ElementT element : iterable) {
+      currentPageSize += 1;
+
+      if (currentPageSize > pageSize) {
+        currentPageNumber += 1;
+        currentPageSize = 1;
+        currentPageElements.clear();
+      }
+
+      currentPageElements.add(element);
+
+      if (currentPageNumber == targetPageNumber && currentPageSize == pageSize) {
+        // The target page has the full size and we've seen all of its elements
+        break;
+      }
+    }
+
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/paging/Pager.java
```diff
@@ -0,0 +1,324 @@
...
+
+  /** What to do if the caller requests a page that is past the end of the current result set. */
+  public enum OutOfBoundsStrategy {
+    /**
+     * Fail with an {@link IndexOutOfBoundsException}.
+     *
+     * <p>As a special case, if the result set is empty and the first page is requested, this
+     * strategy will return an empty page, not fail.
+     */
+    FAIL,
```
olim7t: I wonder if it's worth exposing this strategy at all. Its utility is dubious, and it can always be emulated by using `RETURN_EMPTY_PAGE` and throwing if `page.getPageNumber() > 1 && page.getElements().isEmpty()`.

olim7t: I wonder if it's even worth exposing this strategy at all. Its utility is dubious, and it can always be emulated by using `RETURN_EMPTY_PAGE` and throwing if `page.getPageNumber() > 1 && page.getElements().isEmpty()`.

core/src/main/java/com/datastax/oss/driver/api/core/paging/Pager.java
```diff
@@ -0,0 +1,324 @@
...
+    boolean isLast();
+  }
+
+  /** What to do if the caller requests a page that is past the end of the current result set. */
+  public enum OutOfBoundsStrategy {
+    /**
+     * Fail with an {@link IndexOutOfBoundsException}.
+     *
+     * <p>As a special case, if the result set is empty and the first page is requested, this
+     * strategy will return an empty page, not fail.
```
tomekl007: I think that this doc:
```
this
     * strategy will return an empty page, not fail.
```
is not correct. The `FAIL` strategy should fail

core/src/main/java/com/datastax/oss/driver/api/core/paging/Pager.java
```diff
@@ -0,0 +1,324 @@
...
+   * @throws IllegalArgumentException if the conditions on the arguments are not respected.
+   * @throws IndexOutOfBoundsException if the requested page is past the end of the iterable, and
+   *     the out of bounds strategy is {@link OutOfBoundsStrategy#FAIL}.
+   */
+  @NonNull
+  public <ElementT> Page<ElementT> getPage(
+      @NonNull PagingIterable<ElementT> iterable, final int targetPageNumber, final int pageSize) {
+
+    Objects.requireNonNull(iterable);
+    if (targetPageNumber < 1) {
```
tomekl007: don't we want to start page number from `0`? Wouldn't starting from `1` confuse the users?

core/src/main/java/com/datastax/oss/driver/api/core/paging/Pager.java
tomekl007: I lean into the direction of not providing `OutOfBoundsStrategy` and let the caller handle those cases. We could throw an exception in that case and the user can catch it and returns empty(or last page) if he needs to. Was this `OutOfBoundsStrategy` a feature request? I have the impression that supporting this complicates the implementation a lot. wdyt?

core/src/main/java/com/datastax/oss/driver/api/core/paging/Pager.java
```diff
@@ -0,0 +1,324 @@
...
+    // Throw IllegalArgumentException directly instead of failing the stage, since it signals
+    // blatant programming errors
+    Objects.requireNonNull(iterable);
+    if (targetPageNumber < 1) {
+      throw new IllegalArgumentException(
+          "Invalid targetPageNumber, expected >=1, got " + targetPageNumber);
+    }
+    if (pageSize < 1) {
+      throw new IllegalArgumentException("Invalid pageSize, expected >=1, got " + pageSize);
+    }
```
tomekl007: This validation seems to be common with https://github.com/datastax/java-driver/pull/1389/files#diff-0e080f2b6227ab15f42e1108695bd3a4R100-R107 - maybe we can extract it to a separate method and reuse?

core/src/main/java/com/datastax/oss/driver/api/core/paging/Pager.java
olim7t: If we only keep one strategy, it should be `RETURN_LAST_PAGE`, it's the only one that can't be implemented in term of the others.

But overall I think I agree more and more with you, `RETURN_EMPTY_PAGE` is kind of whacky, I don't think it would be very common in a UI or REST API.

core/src/main/java/com/datastax/oss/driver/api/core/paging/Pager.java
olim7t: In a web UI or REST API it usually starts from 1 (see Google, etc). Since that's a pretty high-level feature we should stay closer to that. See for example how `RandomPagingRestUi` would have to adjust it every time.

core/src/main/java/com/datastax/oss/driver/api/core/paging/Pager.java
olim7t: Fair enough, done.

core/src/main/java/com/datastax/oss/driver/api/core/paging/Pager.java
```diff
@@ -0,0 +1,246 @@
...
+
+  /** A page returned as the result of an offset query. */
+  public interface Page<ElementT> {
+
+    /** The elements in the page. */
+    @NonNull
+    List<ElementT> getElements();
+
+    /**
+     * The page number.
```
adutra: Is it worth mentioning if the numbering starts at zero or 1?

core/src/main/java/com/datastax/oss/driver/api/core/paging/Pager.java
```diff
@@ -0,0 +1,246 @@
...
+     *
+     * <p>It will usually match the requested page number, except if it was past the end of the
+     * iterable. In that case, the last page is returned, and the page number will reflect this.
+     */
+    int getPageNumber();
+
+    /** Whether there are other pages after this one. */
```
adutra: Strictly speaking, this should read "whether there _aren't_ other pages after this one". Maybe -> "Whether this is the last page in the result set."

core/src/main/java/com/datastax/oss/driver/api/core/paging/Pager.java
```diff
@@ -0,0 +1,246 @@
...
+    for (ElementT element : iterable) {
+      currentPageSize += 1;
+
+      if (currentPageSize > pageSize) {
+        currentPageNumber += 1;
+        currentPageSize = 1;
+        currentPageElements.clear();
+      }
+
+      currentPageElements.add(element);
```
adutra: Do we need to add elements that do not belong to the target page? I think it should be possible to only execute the addition `if (currentPageNumber == targetPageNumber)`, which should spare us the call to `ArrayList.clear()` above.

examples/src/main/java/com/datastax/oss/driver/examples/paging/RandomPagingRestUi.java
```diff
@@ -211,7 +210,7 @@ protected void configure() {
     @PostConstruct
     @SuppressWarnings("unused")
     public void init() {
-      this.pager = new Pager(session, ITEMS_PER_PAGE);
+      this.pager = new Pager();
```
adutra: Not related to this PR but I think we need to add some examples with Spring WebFlux and reactive result sets. Created JAVA-2639 for that.

core/src/main/java/com/datastax/oss/driver/api/core/paging/Pager.java
olim7t: This is necessary if the requested page is past the end: we never know which actual page we'll end up returning, so we have to keep each page in case it's the last.

adutra: OK. Then we might be facing one of those rare cases where a `LinkedList` would perform better than `ArrayList`, since `add` is O(1) for the former vs O(n)-ish for the latter.

core/src/main/java/com/datastax/oss/driver/api/core/paging/OffsetPager.java
```diff
@@ -0,0 +1,312 @@
...
+  }
+
+  private final int pageSize;
+
+  /**
+   * Creates a new instance.
+   *
+   * @param pageSize the number of elements per page. Must be greater than or equal to 1.
+   */
+  public OffsetPager(int pageSize) {
```
olim7t, line 120: While writing some examples I realized it's more convenient to pass the page size once through the constructor. Also, it prevents accidentally inverting page size and page number.

The only downside is if you use different page sizes across your application, but this class holds no state so creating multiple instances does not add a lot of overhead.

examples/src/main/java/com/datastax/oss/driver/examples/paging/RandomPagingRestUi.java
```diff
@@ -221,41 +220,38 @@ public void init() {
...
+      Page<Row> page = pager.getPage(session.execute(statement), requestedPageNumber);
+
+      List<UserVideo> videos = new ArrayList<>(page.getElements().size());
+      for (Row row : page.getElements()) {
+        UserVideo video =
+            new UserVideo(row.getInt("videoid"), row.getString("title"), row.getInstant("added"));
+        videos.add(video);
       }
 
+      // Since we use the RETURN_LAST_PAGE strategy, the actual number could be different from the
```
tomekl007: I think that: The comment about `RETURN_LAST_PAGE` is not relevant anymore because the strategy is not present (same for test for this strategy)

core/src/main/java/com/datastax/oss/driver/api/core/paging/OffsetPager.java
```diff
@@ -0,0 +1,316 @@
...
+   * @param iterable the iterable to extract the results from: typically a {@link ResultSet}, or a
+   *     {@link PagingIterable} returned by the mapper.
+   * @param targetPageNumber the page to return (1 for the first page, 2 for the second page, etc).
+   *     Must be greater than or equal to 1.
+   * @return the requested page, or the last page if the requested page was past the end of the
+   *     iterable.
+   * @throws IllegalArgumentException if the conditions on the arguments are not respected.
+   */
+  @NonNull
+  public <ElementT> Page<ElementT> getPage(
```
tomekl007, line 140: can't we reuse the same method for sync and async version?

core/src/main/java/com/datastax/oss/driver/api/core/paging/OffsetPager.java
```diff
@@ -0,0 +1,316 @@
...
+ * // Iterate through rows 1-20 and discard them, then return rows 21-40
+ * OffsetPager.Page&lt;Row&gt; page2 = pager.getPage(rs, 2);
+ *
+ * rs = session.execute(query);
+ * // Iterate through rows 1-80 and discard them, then return rows 81-100
+ * OffsetPager.Page&lt;Row&gt; page5 = pager.getPage(rs, 5);
+ * </pre>
+ *
+ * This is fine for the values typically encountered in real-world applications: for example, if the
+ * page size is 25 and users never go past page 10, the worst case is only 250 rows, which is a very
```
adutra: Actually the number of queries you have to do to get to page N, navigating one by one, is `N * (N + 1) / 2`, that is, to get at page 10 starting from page 1, you are going to execute 1 + 2 + ... + 9 + 10 = 55 queries. You probably did this math already, but was wondering if it's worth mentioning this.

core/src/main/java/com/datastax/oss/driver/api/core/paging/OffsetPager.java
```diff
@@ -0,0 +1,316 @@
...
+ * code. If the page number is exposed to the user (for example if it is passed as a URL parameter),
+ * make sure it is properly validated and enforce a maximum, so that an attacker can't inject a
+ * large value that could potentially fetch millions of rows.
+ *
+ * <h3>Relation with protocol-level paging</h3>
+ *
+ * There is also something called "paging" at the protocol level (controlled by the config option
+ * {@code basic.request.page-size}). It simply means that the driver might receive the results in
+ * multiple "chunks"; this happens under the hood and is completely transparent for users of this
+ * class. You don't need to set the protocol page size to the same value as the logical page size.
```
adutra: Very good point. I wonder if you should stress that protocol-level pages are not guaranteed to have that exact page size, especially when/if CASSANDRA-11745 lands.

examples/src/main/java/com/datastax/oss/driver/examples/paging/RandomPagingRestUi.java
olim7t: Good catch.

core/src/main/java/com/datastax/oss/driver/api/core/paging/OffsetPager.java
olim7t: How would that work? `PagingIterable` and `AsyncPagingIterable` are unrelated types.

core/src/main/java/com/datastax/oss/driver/api/core/paging/OffsetPager.java
olim7t: I see what you mean.
I think it's better to stay generic and describe what happens for one query, in practice clients might not always start from page 1 and click through.

examples/src/main/java/com/datastax/oss/driver/examples/paging/RandomPagingRestUi.java
olim7t: > same for test for this strategy

What do you mean? I didn't find any other text reference to RETURN_LAST_PAGE.

core/src/main/java/com/datastax/oss/driver/api/core/paging/OffsetPager.java
olim7t: I feel it would be a bit distracting here since we're saying that the protocol size does not impact this particular feature. It's already covered in the manual [here](https://docs.datastax.com/en/developer/java-driver/4.5/manual/core/paging/#setting-the-page-size).

adutra (approved):

examples/src/main/java/com/datastax/oss/driver/examples/paging/RandomPagingRestUi.java
tomekl007: I meant this test:
```
  public void should_return_last_page_when_result_finishes_with_empty_frame(int fetchSize) {
```

core/src/main/java/com/datastax/oss/driver/api/core/paging/OffsetPager.java
tomekl007: maybe we could extract some common bits into one class but you are right, maybe it is better to keep them totally separate

core/src/main/java/com/datastax/oss/driver/api/core/paging/Pager.java
adutra: > if the requested page is past the end: we never know which actual page we'll end up returning

Why is that? Shouldn't we just throw an error if the requested page doesn't exist?

PR #1391, opened 2020-01-24
Fix build.yaml for jenkins 

tomekl007:


adutra:
The file was already fixed in another PR, see https://github.com/datastax/java-driver/pull/1390#issuecomment-577502713.

PR #1392, opened 2020-01-24
JAVA-2627: Avoid logging error message including stack trace in reque…

etedpet:
…st handler

Avoid logging error messages inluding stack trace from within
SpeculativeExecution, when the request "continues" by querying the next
host.

Logging level is changed to WARNING and the stack trace is removed.

The stack trace is logged in the debug message instead.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
@@ -427,7 +427,10 @@ class RequestHandler {
                 findNextHostAndQuery();
               } catch (RuntimeException e) {
                 if (connection != null) connection.release();
-                logger.error("Unexpected error while querying " + host.getEndPoint(), e);
+                logger.warn(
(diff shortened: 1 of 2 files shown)


olim7t (approved):

driver-core/src/main/java/com/datastax/driver/core/RequestHandler.java
```diff
@@ -558,7 +564,7 @@ private void retry(final boolean retryCurrent, ConsistencyLevel newConsistencyLe
     }
 
     private void logError(EndPoint endPoint, Throwable exception) {
-      logger.debug("[{}] Error querying {} : {}", id, endPoint, exception.toString());
+      logger.debug("[{}] Error querying {}", id, endPoint, exception);
```
olim7t: I think we can even leave this one as-is, I don't want to spam the debug logs too much for "normal" server error responses.
The error is surfaced via the `errors` map anyway, so it's available if someone wants to inspect it.

PR #1394, opened 2020-01-31
JAVA-2624: Expose a config option for the connect timeout

emerkle826:


emerkle826:
I tried using Simulacron to setup a situation where the client can't create a channel to the server and get the Netty connect timeout to pop, but Simulacron rejects immediately (for connections). It seems you can allow a connection and delay or prevent protocol negotiations from completing, but you can't have it delay the channel connection. If you feel I should do more to come up with a better test, let me know.

tomekl007:
> I tried using Simulacron to setup a situation where the client can't create a channel to the server and get the Netty connect timeout to pop, but Simulacron rejects immediately (for connections). It seems you can allow a connection and delay or prevent protocol negotiations from completing, but you can't have it delay the channel connection. If you feel I should do more to come up with a better test, let me know.

I wonder how hard it will be to introduce such functionality in simulacron. I imagine that it may be useful in the future to emulate connection problems and test the delay on the connection.

core/src/main/resources/reference.conf
```diff
@@ -420,6 +420,18 @@ datastax-java-driver {
...
+    #
+    # This timeout is for controlling how long the driver will wait for the underlying channel
+    # to actually connect to the server. This is not the time limit for completing protocol
+    # negotiations, only the time limit for establishing a channel connection.
+    #
+    # Required: yes
+    # Modifiable at runtime: yes, the new value will be used for connections created after the
+    #   change.
+    # Overridable in a profile: no
+    connect-timeout = 5 seconds
```
tomekl007: why `5 seconds`?

emerkle826: 5 seconds was the default for the 3.x driver: https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/SocketOptions.java#L23-L29

tomekl007: Shouldn't we increase that? In driver 3 timeouts were more fine-grain. In 4.x the request timeout is taking whole path of request into account - so I suspect it is higher. In such a case maybe connect timeout should be also higher?

emerkle826: This is just the Netty channel connection timeout though. I don't think it affects or is affected by request processing. Just how long Netty will wait before giving up trying to establish an I/O connection to the server.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
```diff
@@ -161,12 +162,20 @@ private void connect(
...
+            .getDuration(DefaultDriverOption.CONNECTION_CONNECT_TIMEOUT);
 
     Bootstrap bootstrap =
         new Bootstrap()
             .group(nettyOptions.ioEventLoopGroup())
             .channel(nettyOptions.channelClass())
             .option(ChannelOption.ALLOCATOR, nettyOptions.allocator())
+            .option(
+                ChannelOption.CONNECT_TIMEOUT_MILLIS,
+                Long.valueOf(connectTimeout.toMillis()).intValue())
```
emerkle826: @olim7t Should I move this bit into `afterBootstrapInitialized()`?

core/src/main/resources/reference.conf
olim7t: 👍 
This is unrelated to `basic.request.timeout`. It will only affect session initialization, or background reconnections. The same default as 3.x is fine.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
olim7t: Yes, let's move it there for consistency, that's where the other socket options are processed.

core/src/main/resources/reference.conf
olim7t: Nit: could you move it at the beginning of the section, either before or after the other timeouts? It feels slightly more convenient to have all the connection timeouts in the same place.

olim7t (approved):
LGTM, provided that we move the change to `NettyOptions` as discussed inline.

As for adding connect timeout to Simulacron, it would be a nice-to-have but I don't think we need to hold this PR for it, it's a pretty straightforward change. Maybe you can create a follow-up issue [here](https://github.com/datastax/simulacron/issues).

tomekl007 (approved):

emerkle826:
I'll rebase this and resolve the conflict. I've also opened up an enhancement issue for Simulacron here: https://github.com/datastax/simulacron/issues/95

PR #1396, opened 2020-02-01
JAVA-1556: Publish Maven Bill Of Materials POM

olim7t:
Validated locally against a test project:
```xml
  <dependencyManagement>
    <dependencies>
      <dependency>
        <artifactId>java-driver-bom</artifactId>
        <version>4.5.0-SNAPSHOT</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
  <dependencies>
    <dependency>
      <artifactId>java-driver-core-shaded</artifactId>
    </dependency>
  </dependencies>
```

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -1,0 +1,109 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright DataStax, Inc.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.datastax.oss</groupId>
+    <artifactId>java-driver-parent</artifactId>
+    <version>4.5.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>java-driver-bom</artifactId>
+  <packaging>pom</packaging>
+  <name>DataStax Java driver for Apache Cassandra(R) - Bill Of Materials</name>
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.datastax.oss</groupId>
+        <artifactId>java-driver-core</artifactId>
+        <version>4.5.0-SNAPSHOT</version>
+      </dependency>
+      <dependency>
+        <groupId>com.datastax.oss</groupId>
+        <artifactId>java-driver-core-shaded</artifactId>
+        <version>4.5.0-SNAPSHOT</version>
+      </dependency>
+      <dependency>
+        <groupId>com.datastax.oss</groupId>
+        <artifactId>java-driver-mapper-processor</artifactId>
+        <version>4.5.0-SNAPSHOT</version>
+      </dependency>
+      <dependency>
+        <groupId>com.datastax.oss</groupId>
+        <artifactId>java-driver-mapper-runtime</artifactId>
+        <version>4.5.0-SNAPSHOT</version>
+      </dependency>
+      <dependency>
+        <groupId>com.datastax.oss</groupId>
+        <artifactId>java-driver-query-builder</artifactId>
+        <version>4.5.0-SNAPSHOT</version>
+      </dependency>
+      <dependency>
+        <groupId>com.datastax.oss</groupId>
+        <artifactId>java-driver-test-infra</artifactId>
+        <version>4.5.0-SNAPSHOT</version>
+      </dependency>
+      <dependency>
+        <groupId>com.datastax.oss</groupId>
+        <artifactId>native-protocol</artifactId>
+        <version>1.4.9</version>
+      </dependency>
+      <dependency>
+        <groupId>com.datastax.oss</groupId>
+        <artifactId>java-driver-shaded-guava</artifactId>
+        <version>25.1-jre</version>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>flatten-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <id>flatten</id>
+            <phase>process-resources</phase>
+            <goals>
+              <goal>flatten</goal>
+            </goals>
+            <configuration>
+              <pomElements>
+                <dependencyManagement>keep</dependencyManagement>
+                <ciManagement>expand</ciManagement>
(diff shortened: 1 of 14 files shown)


bom/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+  </parent>
+  <artifactId>java-driver-bom</artifactId>
+  <packaging>pom</packaging>
+  <name>DataStax Java driver for Apache Cassandra(R) - Bill Of Materials</name>
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>${project.groupId}</groupId>
+        <artifactId>java-driver-core</artifactId>
+        <version>${project.version}</version>
```
olim7t: The original reporter of the jira ticket advised against that:
> A good practice is to NOT use project.version property when declaring a version, but the real version

But it works, so unless someone can explain why this would be a bad idea, I'm going with the simplest option.

I would even have omitted the version altogether, because we already have a `<dependencyManagement>` section in the parent POM. But apparently this one doesn't inherit from it.

adutra: @mp911de would you mind taking a look? ^

bom/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+        <groupId>${project.groupId}</groupId>
+        <artifactId>java-driver-query-builder</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>${project.groupId}</groupId>
+        <artifactId>java-driver-test-infra</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+    </dependencies>
```
adutra, line 71: I don't see the shaded Guava artifact nor native-protocol in this list, isn't the whole purpose to groupe together all driver transitive dependencies so that we make sure users are importing compatible versions of each artifact?

bom/pom.xml
mp911de: I'm not exactly sure about the full rationale behind why placeholders should not be used in a BOM declaration. From what I know is that placeholders require resolution during consumption. With the current arrangement, a consumer needs to fetch `java-driver-parent` just to be able to resolve version attributes regardless of whether dependencies are included in the project.

A typical project using the BOM directly shouldn't be impacted too much as they would depend on the Cassandra driver anyway. For frameworks (such as Spring Boot), they typically come with a wide dependency management arrangement that points to pre-defined versions while a lot of users do not use all declared dependency versions.

bom/pom.xml
mp911de: :+1: A BOM helps to resolve version ambiguities as `<dependencyManagement>` affects also versions for transitive dependencies. That being said, you should include all 1st party artifacts that you're shipping. The native protocol and the shaded guava libraries should be included.

You can see the effect in the following example: If a project uses two dependencies, that depend each on netty these dependencies use different netty versions, then importing the netty BOM ties the actually used netty version to a single consistent version.

bom/pom.xml
snicoll: The rationale is that `${project.version}` is a bit misleading in the sense of it is a standard property of the project. Right now Maven resolves that token locally to the project (so it uses the version of the bom) but if you look at `${project.version}` in a project that imports the BOM, it does not refer to the same value at all. 

That's why things like the flattened maven plugin resolves the version and hardcodes it to make it more explicit. I'd recommend using that plugin btw, especially for a bom.

bom/pom.xml
snicoll: Anything that's yours and that another bom of yours does not manage should indeed be here.

bom/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>com.datastax.oss</groupId>
+    <artifactId>java-driver-parent</artifactId>
```
snicoll, line 23: you can't really do that. The parent must be as small as possible and, of course, should not include any dependency management at all.

snicoll (changes requested):
The current parent is problematic. You can run `mvn help:effective-pom` to figure out what the actual resolved bom looks like with the current parent.

olim7t:
Thanks for the reviews and explanations.

Hard-coding the version for project modules works perfectly, the release plugin is smart enough to keep them in sync. But for native-protocol and java-driver-shaded-guava, I want to keep referencing the properties in the parent POM, otherwise we would have to keep them in sync manually.

So I kept the parent dependency and added the flatten plugin. I had to switch to full manual configuration because `flattenMode=bom` keeps some plugins in the build section. And even then, I can't get it to do exactly what I want: `dependencyManagement=interpolate` does not resolve the versions of native-protocol and guava (I think it just interpolates properties defined in the same POM, not those of the parents). `dependencyManagement=keep` keeps the section as-is, which means I have to expand the properties; but that copies every property from the parent POM, not just the two I need.

So in the end I can generate the flattened BOM, the only problem is that it has a bunch of unused properties. As far as I can tell, that's the best we can do with the flatten plugin.

Another option would be to keep the reference to the parent POM. It doesn't make a big difference because it will get downloaded as soon as a module is referenced anyway. I think the flat BOM is slightly cleaner though.

olim7t:
Looks like there's a problem with the generated BOM in Travis. I'll check tomorrow.

adutra:
> I want to keep referencing the properties in the parent POM, otherwise we would have to keep them in sync manually.

I think that the whole point of having a BOM is to get rid of that `<dependencyManagement>` section in the parent POM altogether, along with those properties. 

The only problem I foresee is that some of these properties are being injected in OSGi tests, we would probably have to figure out some other way to do this.

olim7t:
I see. I hadn't realized that the modules themselves can use the BOM. In that case we can hard-code the versions of native-protocol and guava in the BOM and use it as the only source for them.

> The only problem I foresee is that some of these properties are being injected in OSGi tests, we would probably have to figure out some other way to do this.

Maybe we could piggyback on the dependency reports generated by [JAVA-2617](https://datastax-oss.atlassian.net/browse/JAVA-2617).

snicoll:
> I hadn't realized that the modules themselves can use the BOM. 

It's more than it should otherwise you're not testing the BOM before releasing.

olim7t:
> Maybe we could piggyback on the dependency reports generated by JAVA-2617.

OK, it worked. For now it's done just for native-protocol and Guava. I think it would be nice to generalize it to all versions, but then we also need to generate a report for test dependencies. I'll create a follow-up ticket (I've spent way more time on this than expected already).

The BOM now hard-codes all versions, no more properties:
```xml
      <dependency>
        <version>4.5.0-SNAPSHOT</version>
      </dependency>
      <dependency>
        <artifactId>java-driver-core-shaded</artifactId>
        <version>4.5.0-SNAPSHOT</version>
      </dependency>
      <dependency>
        <artifactId>java-driver-mapper-processor</artifactId>
        <version>4.5.0-SNAPSHOT</version>
      </dependency>
      <dependency>
        <artifactId>java-driver-mapper-runtime</artifactId>
        <version>4.5.0-SNAPSHOT</version>
      </dependency>
      <dependency>
        <artifactId>java-driver-query-builder</artifactId>
        <version>4.5.0-SNAPSHOT</version>
      </dependency>
      <dependency>
        <artifactId>java-driver-test-infra</artifactId>
        <version>4.5.0-SNAPSHOT</version>
      </dependency>
      <dependency>
        <version>1.4.9</version>
      </dependency>
      <dependency>
        <artifactId>java-driver-shaded-guava</artifactId>
        <version>25.1-jre</version>
      </dependency>
```
I'll merge this on Monday.

PR #1397, opened 2020-02-04
JAVA-2637: Bump Netty to 4.1.45

olim7t:
This required adjustements in a few unit tests, where
`waitForPendingAdminTasks` methods relied on the fact that a task
scheduled on the event loop with a short delay would be executed after
any current immediate task.

This doesn't work anymore, possibly after the event loop refactoring in
Netty 4.1.44. It was replaced with a combination of Awaitility and
timeouts in `Mockito.verify` calls.

### DIFF
--- a/core/src/test/java/com/datastax/oss/driver/internal/core/control/ControlConnectionEventsTest.java
+++ b/core/src/test/java/com/datastax/oss/driver/internal/core/control/ControlConnectionEventsTest.java
@@ -16,6 +16,7 @@
 package com.datastax.oss.driver.internal.core.control;
 
 import static com.datastax.oss.driver.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
@@ -46,16 +47,19 @@ public class ControlConnectionEventsTest extends ControlConnectionTestBase {
 
     // When
     controlConnection.init(true, false, false);
-    waitForPendingAdminTasks();
-    DriverChannelOptions channelOptions = optionsCaptor.getValue();
 
     // Then
-    assertThat(channelOptions.eventTypes)
-        .containsExactly(
-            ProtocolConstants.EventType.SCHEMA_CHANGE,
-            ProtocolConstants.EventType.STATUS_CHANGE,
-            ProtocolConstants.EventType.TOPOLOGY_CHANGE);
-    assertThat(channelOptions.eventCallback).isEqualTo(controlConnection);
+    await()
+        .untilAsserted(
+            () -> {
+              DriverChannelOptions channelOptions = optionsCaptor.getValue();
+              assertThat(channelOptions.eventTypes)
+                  .containsExactly(
+                      ProtocolConstants.EventType.SCHEMA_CHANGE,
+                      ProtocolConstants.EventType.STATUS_CHANGE,
+                      ProtocolConstants.EventType.TOPOLOGY_CHANGE);
+              assertThat(channelOptions.eventCallback).isEqualTo(controlConnection);
+            });
   }
 
   @Test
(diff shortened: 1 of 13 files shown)


core/src/test/java/com/datastax/oss/driver/internal/core/control/ControlConnectionTest.java
```diff
@@ -155,26 +154,23 @@ public void should_reconnect_if_channel_goes_down() throws Exception {
     CompletionStage<Void> initFuture = controlConnection.init(false, false, false);
     factoryHelper.waitForCall(node1);
 
-    waitForPendingAdminTasks();
-    assertThatStage(initFuture).isSuccess();
-    assertThat(controlConnection.channel()).isEqualTo(channel1);
-    verify(eventBus).fire(ChannelEvent.channelOpened(node1));
+    assertThatStage(initFuture)
+        .isSuccess(v -> assertThat(controlConnection.channel()).isEqualTo(channel1));
```
olim7t, line 158: Note that `assertThatStage` already waits (at most 2 seconds) if the future is not done yet. A lot of tests were not taking advantage of that.

core/src/test/java/com/datastax/oss/driver/internal/core/control/ControlConnectionTest.java
```diff
@@ -155,26 +154,23 @@ public void should_reconnect_if_channel_goes_down() throws Exception {
...
-    assertThat(controlConnection.channel()).isEqualTo(channel2);
-    verify(eventBus).fire(ChannelEvent.channelClosed(node1));
-    verify(eventBus).fire(ChannelEvent.channelOpened(node2));
-    verify(metadataManager).refreshNodes();
-    verify(loadBalancingPolicyWrapper).init();
+    await().untilAsserted(() -> assertThat(controlConnection.channel()).isEqualTo(channel2));
+    verify(eventBus, VERIFY_TIMEOUT).fire(ChannelEvent.channelClosed(node1));
+    verify(eventBus, VERIFY_TIMEOUT).fire(ChannelEvent.channelOpened(node2));
+    verify(metadataManager, VERIFY_TIMEOUT).refreshNodes();
+    verify(loadBalancingPolicyWrapper, VERIFY_TIMEOUT).init();
```
olim7t, line 173: I'm being a bit heavy-handed with the verify timeouts, some events could probably piggy-back on previous timed verifications. But I prefer not to rely on subtle reasoning about ordering.

core/src/test/java/com/datastax/oss/driver/internal/core/control/ControlConnectionEventsTest.java
```diff
@@ -46,16 +47,19 @@ public void should_register_for_all_events_if_topology_requested() {
...
 
     // Then
-    assertThat(channelOptions.eventTypes)
-        .containsExactly(
-            ProtocolConstants.EventType.SCHEMA_CHANGE,
-            ProtocolConstants.EventType.STATUS_CHANGE,
-            ProtocolConstants.EventType.TOPOLOGY_CHANGE);
-    assertThat(channelOptions.eventCallback).isEqualTo(controlConnection);
+    await()
+        .untilAsserted(
```
tomekl007, line 53: Isn't `unitAsserted` too strict? There is an option that it will block indefinitely. I think it is better to have `wait().unit()` with some sane timeout.

core/src/test/java/com/datastax/oss/driver/internal/core/pool/ChannelPoolTestBase.java
```diff
@@ -111,17 +111,4 @@ DriverChannel newMockDriverChannel(int id) {
     when(driverChannel.toString()).thenReturn("channel" + id);
     return driverChannel;
   }
-
-  // Wait for all the tasks on the pool's admin executor to complete.
-  void waitForPendingAdminTasks() {
```
tomekl007, line 116: why we don't need this method anymore?

olim7t: It doesn't work anymore with Netty 4.1.45.
So I replaced it with all the timed verifies / awaitility stuff.

core/src/test/java/com/datastax/oss/driver/internal/core/control/ControlConnectionEventsTest.java
olim7t: I think by default it times out after 10 seconds, like the other awaitility methods. Will check.

olim7t: Yes, it's 10 seconds:
```java
await().untilAsserted(() -> assertThat(true).isEqualTo(false));

org.awaitility.core.ConditionTimeoutException: Assertion condition defined as a Tmp expected:<[fals]e> but was:<[tru]e> within 10 seconds.
```
But I also noticed that `untilAsserted` doesn't propertly throw a timeout exception in Awaitibility 3.1.6. I've upgraded it to 4.0.2.

PR #1398, opened 2020-02-10
JAVA-2644: Revisit channel selection when pool size > 1

olim7t:
~Very early preview because I want to share the direction this is taking ASAP. Tests do not pass yet, and I still have some cleanup and documentation to do.~ edit -- this is ready for review now

In a nutshell, `DriverChannel.preAcquireId()` must be called **exactly once** before each call to `DriverChannel.write()`. No need to explicitly release the id, this will still be done automatically in `InFlightHandler`.

`DriverChannel.getAvailableIds()` now reports the number of ids that have been pre-acquired and not yet released. `preAquireId / getAvailableIds` now have atomic semantics, this allows clients to have a guarantee that once pre-acquired succeeded they won't run into a `BusyConnectionException`.

Unfortunately this re-introduces some of the complexity we had tried to get rid of compared to 3.x: we can't allow the count to get out of sync with the actual number of ids used / about to be used. `ChannelPool.next()` and `ChannelHandlerRequest` automatically pre-acquire. For `AdminRequestHandler`, things get more complicated: while most cases need to pre-acquire (e.g. every use in the control connection), a few cases must not because they operate on a channel that was obtained with `ChannelPool.next()`, we don't want to pre-acquire twice. For now I've introduced a new boolean parameter that must be set every time, I'll see if I can do better tomorrow.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/connection/BusyConnectionException.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/connection/BusyConnectionException.java
@@ -30,12 +30,19 @@ import edu.umd.cs.findbugs.annotations.NonNull;
  */
 public class BusyConnectionException extends DriverException {
 
+  // Note: the driver doesn't use this constructor anymore, it is preserved only for backward
+  // compatibility.
+  @SuppressWarnings("unused")
   public BusyConnectionException(int maxAvailableIds) {
     this(
         String.format(
             "Connection has exceeded its maximum of %d simultaneous requests", maxAvailableIds),
         null,
         false);
+  }
+
+  public BusyConnectionException(String message) {
+    this(message, null, false);
   }
 
   private BusyConnectionException(
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRequestHandler.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRequestHandler.java
@@ -17,6 +17,7 @@ package com.datastax.oss.driver.internal.core.adminrequest;
 
 import com.datastax.oss.driver.api.core.DriverTimeoutException;
 import com.datastax.oss.driver.api.core.ProtocolVersion;
+import com.datastax.oss.driver.api.core.connection.BusyConnectionException;
 import com.datastax.oss.driver.api.core.type.codec.TypeCodecs;
 import com.datastax.oss.driver.internal.core.channel.DriverChannel;
 import com.datastax.oss.driver.internal.core.channel.ResponseCallback;
@@ -54,6 +55,7 @@ public class AdminRequestHandler<ResultT> implements ResponseCallback {
       DriverChannel channel, Query query, Duration timeout, String logPrefix) {
     return new AdminRequestHandler<>(
         channel,
+        true,
         query,
         Frame.NO_PAYLOAD,
         timeout,
@@ -78,7 +80,7 @@ public class AdminRequestHandler<ResultT> implements ResponseCallback {
       debugString += " with parameters " + parameters;
     }
     return new AdminRequestHandler<>(
-        channel, message, Frame.NO_PAYLOAD, timeout, logPrefix, debugString, Rows.class);
+        channel, true, message, Frame.NO_PAYLOAD, timeout, logPrefix, debugString, Rows.class);
   }
 
   public static AdminRequestHandler<AdminResult> query(
@@ -87,6 +89,7 @@ public class AdminRequestHandler<ResultT> implements ResponseCallback {
   }
 
   private final DriverChannel channel;
+  private final boolean shouldPreAcquireId;
   private final Message message;
   private final Map<String, ByteBuffer> customPayload;
   private final Duration timeout;
@@ -100,6 +103,7 @@ public class AdminRequestHandler<ResultT> implements ResponseCallback {
 
   protected AdminRequestHandler(
       DriverChannel channel,
+      boolean shouldPreAcquireId,
       Message message,
       Map<String, ByteBuffer> customPayload,
       Duration timeout,
@@ -107,6 +111,7 @@ public class AdminRequestHandler<ResultT> implements ResponseCallback {
       String debugString,
       Class<? extends Result> expectedResponseType) {
     this.channel = channel;
+    this.shouldPreAcquireId = shouldPreAcquireId;
     this.message = message;
     this.customPayload = customPayload;
     this.timeout = timeout;
@@ -117,7 +122,14 @@ public class AdminRequestHandler<ResultT> implements ResponseCallback {
 
   public CompletionStage<ResultT> start() {
     LOG.debug("[{}] Executing {}", logPrefix, this);
-    channel.write(message, false, customPayload, this).addListener(this::onWriteComplete);
+    if (shouldPreAcquireId && !channel.preAcquireId()) {
+      setFinalError(
+          new BusyConnectionException(
+              String.format(
+                  "%s has reached its maximum number of simultaneous requests", channel)));
+    } else {
+      channel.write(message, false, customPayload, this).addListener(this::onWriteComplete);
+    }
     return result;
   }
 
@@ -199,6 +211,8 @@ public class AdminRequestHandler<ResultT> implements ResponseCallback {
         buildQueryOptions(currentOptions.pageSize, currentOptions.namedValues, pagingState);
     return new AdminRequestHandler<>(
         channel,
+        // This is called for next page queries, so we always need to reacquire an id:
+        true,
         new Query(current.query, newOptions),
         customPayload,
         timeout,
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/channel/InFlightHandler.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/channel/InFlightHandler.java
@@ -117,17 +117,24 @@ public class InFlightHandler extends ChannelDuplexHandler {
   private void write(ChannelHandlerContext ctx, RequestMessage message, ChannelPromise promise) {
     if (closingGracefully) {
       promise.setFailure(new IllegalStateException("Channel is closing"));
+      streamIds.cancelPreAcquire();
       return;
     }
     int streamId = streamIds.acquire();
     if (streamId < 0) {
-      promise.setFailure(new BusyConnectionException(streamIds.getMaxAvailableIds()));
+      // Should not happen with the preAcquire mechanism, but handle gracefully
+      promise.setFailure(
+          new BusyConnectionException(
+              String.format(
+                  "Couldn't acquire a stream id from InFlightHandler on %s", ctx.channel())));
+      streamIds.cancelPreAcquire();
       return;
     }
 
     if (inFlight.containsKey(streamId)) {
       promise.setFailure(
           new IllegalStateException("Found pending callback for stream id " + streamId));
+      streamIds.cancelPreAcquire();
       return;
     }
 
@@ -372,6 +379,10 @@ public class InFlightHandler extends ChannelDuplexHandler {
 
   int getAvailableIds() {
     return streamIds.getAvailableIds();
+  }
+
+  boolean preAcquireId() {
+    return streamIds.preAcquire();
   }
 
   int getInFlight() {
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareHandler.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareHandler.java
@@ -283,6 +283,7 @@ public class CqlPrepareHandler implements Throttled {
       ThrottledAdminRequestHandler<ByteBuffer> handler =
           ThrottledAdminRequestHandler.prepare(
               channel,
+              false,
               message,
               request.getCustomPayload(),
               timeout,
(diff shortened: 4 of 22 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
```diff
@@ -660,6 +660,7 @@ private void processErrorResponse(Error errorMessage) {
         ThrottledAdminRequestHandler<ByteBuffer> reprepareHandler =
             ThrottledAdminRequestHandler.prepare(
                 channel,
+                true,
                 reprepareMessage,
                 repreparePayload.customPayload,
                 timeout,
```
olim7t, line 666: (this is the prepare-and-retry scenario when we get UNPREPARED)

Note that we reuse the same channel as the original query. If it is full, `reprepareHandler` will get an error when it calls `preAcquireId`.
In theory it would be better to reacquire from the pool in case there is another channel that is less busy. But that complicates the code here, and I don't think this is hugely important since the request will be retried on the next node anyway.

core/src/main/java/com/datastax/oss/driver/internal/core/session/ReprepareOnUp.java
```diff
@@ -224,37 +226,51 @@ private void startWorker() {
...
   }
 
   @VisibleForTesting
   protected CompletionStage<ByteBuffer> prepareAsync(
       Message message, Map<String, ByteBuffer> customPayload) {
-    ThrottledAdminRequestHandler<ByteBuffer> reprepareHandler =
-        ThrottledAdminRequestHandler.prepare(
-            channel, message, customPayload, timeout, throttler, metricUpdater, logPrefix);
-    return reprepareHandler.start();
+    DriverChannel channel = pool.next();
```
olim7t, line 265: Side-effect of reacquiring every time: the reprepare requests will now be distributed across pooled connections, vs using just 1 in the previous implementation. This is an improvement.

core/src/main/java/com/datastax/oss/driver/internal/core/pool/ChannelSet.java
```diff
@@ -83,18 +83,25 @@ DriverChannel next() {
...
-          int score = channel.getAvailableIds();
-          if (score > bestScore) {
-            bestScore = score;
-            best = channel;
+        while (true) {
+          DriverChannel best = null;
+          int bestScore = 0;
+          for (DriverChannel channel : snapshot) {
+            int score = channel.getAvailableIds();
+            if (score > bestScore) {
```
tomekl007, line 105: why we are assuming that a higher score means that the channel is better?

olim7t: We want to distribute incoming requests evenly across channels. Even though this is not perfect (some requests might be more expensive than others), I think it's a fair approximation.

tomekl007: so it basically means that we are trying to always pick the highest available stream id, right?

olim7t: The channel with the most available ids.

core/src/main/java/com/datastax/oss/driver/internal/core/pool/ChannelSet.java
```diff
@@ -83,18 +83,25 @@ DriverChannel next() {
...
+        return onlyChannel.preAcquireId() ? onlyChannel : null;
       default:
-        DriverChannel best = null;
-        int bestScore = 0;
-        for (DriverChannel channel : snapshot) {
-          int score = channel.getAvailableIds();
-          if (score > bestScore) {
-            bestScore = score;
-            best = channel;
+        while (true) {
```
olim7t: Maybe we should add a limit to the number of iterations, to avoid a starvation condition if a thread consistently sees channels with available ids, but the best one always fills up by the time it tries to acquire.

olim7t: After testing locally, I think we are pretty safe. Even with a high level of contention (1000 threads competing for 5 connections with 20 requests/connection), the highest number of iterations I've observed is 3, and this happens 5 times over a 2-minute test period. One iteration in the loop usually takes about 200ns.

integration-tests/src/test/java/com/datastax/oss/driver/core/PoolBalancingIT.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.atomic.AtomicReference;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
+
+@Category(ParallelizableTests.class)
+public class PoolBalancingIT {
```
olim7t, line 40: Pretty basic, but it reproduces the error immediately on 4.x.

integration-tests/src/test/java/com/datastax/oss/driver/core/PoolBalancingIT.java
```diff
@@ -0,0 +1,97 @@
...
+                session.execute("SELECT release_version FROM system.local");
+              } catch (Throwable t) {
+                unexpectedErrorRef.compareAndSet(null, t);
+                // Even a single error is a failure, no need to continue
+                done.countDown();
+              }
+            }
+          });
+    }
+    timeoutExecutor.schedule(done::countDown, 5, SECONDS);
```
tomekl007: so this test will wait up to 5 seconds, and if there is no error during that time it will finish, right?
Is there a particular reason why it is 5 seconds?

core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRequestHandler.java
```diff
@@ -117,7 +122,14 @@ protected AdminRequestHandler(
 
   public CompletionStage<ResultT> start() {
     LOG.debug("[{}] Executing {}", logPrefix, this);
-    channel.write(message, false, customPayload, this).addListener(this::onWriteComplete);
+    if (preAcquireId && !channel.preAcquireId()) {
```
tomekl007: This is a bit confusing to me, from what I understand, the `preAcquireId` means that is should pre-acquire, and then `channel.preAcquireId()` is doing actual acquisition of stream id, right? If so, I propose to rename local field `preAcquireId` to `shouldPreAcquireId`, wdyt?

core/src/main/java/com/datastax/oss/driver/internal/core/channel/InFlightHandler.java
```diff
@@ -117,17 +117,24 @@ public void write(ChannelHandlerContext ctx, Object in, ChannelPromise promise)
...
       promise.setFailure(new IllegalStateException("Channel is closing"));
+      streamIds.cancelPreAcquire();
       return;
     }
     int streamId = streamIds.acquire();
     if (streamId < 0) {
-      promise.setFailure(new BusyConnectionException(streamIds.getMaxAvailableIds()));
+      // Should not happen with the preAcquire mechanism, but handle gracefully
+      promise.setFailure(
+          new BusyConnectionException(
```
tomekl007, line 127: don't we want to include this `streamIds.getMaxAvailableIds()` in the failure any more? Why?

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareHandler.java
```diff
@@ -283,6 +283,7 @@ private void setFinalResult(Prepared prepared) {
       ThrottledAdminRequestHandler<ByteBuffer> handler =
           ThrottledAdminRequestHandler.prepare(
               channel,
+              false,
```
tomekl007, line 286: why we don't need to pre-acquire here?

core/src/main/java/com/datastax/oss/driver/internal/core/pool/ChannelSet.java
```diff
@@ -83,18 +94,27 @@ DriverChannel next() {
       case 0:
         return null;
       case 1:
-        return snapshot[0];
+        DriverChannel onlyChannel = snapshot[0];
```
tomekl007, line 97: question not related to this PR: the `snapshot` is only a reference to the array, so if something in the `private volatile DriverChannel[] channels;` will change it will be reflected in the `snapshot` - so to me it seems like not a snapshot (it would need to be an immutable copy to be a real snapshot) Is it intended? Isn't it causing us some problems?

core/src/main/java/com/datastax/oss/driver/internal/core/pool/ChannelSet.java
```diff
@@ -32,6 +34,15 @@
...
 @ThreadSafe
 class ChannelSet implements Iterable<DriverChannel> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(ChannelSet.class);
+  /**
+   * The maximum number of iterations in the busy wait loop in {@link #next()} when there are
+   * multiple channels. This is a backstop to protect against thread starvation, in practice we've
+   * never observed more than 3 iterations in tests.
+   */
+  private static final int MAX_ITERATIONS = 50;
```
tomekl007, line 44: maybe we should add a unit test for that `MAX_ITERATIONS`?

integration-tests/src/test/java/com/datastax/oss/driver/core/PoolBalancingIT.java
olim7t: Yes. 5 seconds is enough to reproduce the issue on 4.x. In fact it happens much faster, so maybe I can lower this timeout.

core/src/main/java/com/datastax/oss/driver/internal/core/adminrequest/AdminRequestHandler.java
olim7t: Yeah I see what you mean, I will rename the boolean field.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/InFlightHandler.java
olim7t: It's not easily obtainable from other places where the constructor is now called, and I thought it was pretty clear from the configuration.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareHandler.java
olim7t: Because the channel was obtained from `session.getChannel`, which already pre-acquires (as part of selecting the channel from the pool, see `ChannelSet.next()`).

That's the kind of complexity we have to deal with after this change, unfortunately.

core/src/main/java/com/datastax/oss/driver/internal/core/pool/ChannelSet.java
olim7t: Right, I think it can be covered by a test.

core/src/main/java/com/datastax/oss/driver/internal/core/pool/ChannelSet.java
olim7t: Mutation operations (`add()` and `remove()`) always replace the whole array, they don't switch elements in place. So `snapshot` is guaranteed to represent a consistent view of the set at some point in recent time.

The worst that can happen is that `add()` or `remove()` is called concurrently right after we take the snapshot.
- add: we're missing a new channel that just got added, no big deal (it's as if we had done the selection a few nanoseconds before)
- remove: the stale snapshot might contain a channel that is now closed. At the very worst we select it, the client will run into an error when it tries to write, and try the next node.

tomekl007: thanks for the explanation

integration-tests/src/test/java/com/datastax/oss/driver/core/PoolBalancingIT.java
```diff
@@ -49,49 +49,56 @@
...
-                done.countDown();
-              }
-            }
-          });
+      reschedule(null, null);
     }
-    timeoutExecutor.schedule(done::countDown, 5, SECONDS);
-
-    done.await();
+    SECONDS.sleep(5);
```
tomekl007: Introduction 5 seconds delay in a test seems like a lot, maybe instead we could create a new `AtomicInteger successfullRequests = simultaneousRequests`, increment it in every successful session execute, and change the sleep to:
`awaitUntil( () successfullRequests == simultaneousRequests)`? The test will keep the characteristic because we will be waiting until all succeed OR error will happen. wdyt?

olim7t: `simultaneousRequests` is the concurrent amount at any given point in time, we execute way more than that.

I'll lower the delay to 1 second, that's more than enough to reproduce the bug on 4.x.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareHandler.java
adutra: Wouldn't it be possible for `DriverChannel` to keep track of whether or not an id has been pre-acquired (and clear that flag once `write` has been called)?

core/src/main/java/com/datastax/oss/driver/api/core/connection/BusyConnectionException.java
```diff
@@ -30,6 +30,9 @@
  */
 public class BusyConnectionException extends DriverException {
 
+  // Note: the driver doesn't use this constructor anymore, it is preserved only for backward
+  // compatibility.
+  @SuppressWarnings("unused")
```
adutra, line 35: We usually mark those as deprecated, why not here?

adutra (commented):
Sorry I came very late to the party, feel free to ignore my comments or address them with low priority.

PR #1399, opened 2020-02-10
JAVA-2630: Correctly handle custom classes in IndexMetadata.describe

emerkle826:


core/src/test/java/com/datastax/oss/driver/internal/core/metadata/schema/IndexMetadataTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.Test;
+
+public class IndexMetadataTest {
+
+  private IndexMetadata indexMetadata;
+
+  @Before
+  public void before() {
+    final Map<String, String> options = new HashMap<>();
+    options.put("class_name", "com.datastax.MyClass");
```
tomekl007: what is a reason for doing this in `@Before?` From what I see this is used in only one `@Test` so I guess it belongs to that test (given section)

emerkle826: Good point. I started it in a `@Before` method as I thought I was going to have to mock things. If we decide to add more tests here, maybe there will be a more obvious reason for a `@Before` method, but for now I can just move it all inside the one test method.

olim7t (approved):
👍

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/schema/IndexMetadataTest.java
olim7t: 👍 to move it to the test method, it's a bit overkill right now.

PR #1400, opened 2020-02-25
JAVA-2634: Don't throw error on graph metadata corruption

GregBestland:


adutra:
After discussion with the team we agreed that we should keep the current behavior of failing fast when the schema metadata is corrupted.

PR #1401, opened 2020-02-25
JAVA-2184: Refactor DescribeIT to improve maintenability

olim7t:
As explained on the JIRA ticket:
* don't duplicate create statements in the test code. Instead, load the CQL file from `src/test/resources`, parse the statements it contains and execute them.
* split the CQL files into OSS and DSE variants
* always use the file with the highest version that is less than or equal to the server version running in the test. For example, if we're running against DSE 6.7.x, use `dse/5.1.cql`.


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/RelationParser.java
```diff
@@ -83,6 +87,26 @@ protected RelationParser(SchemaRows rows, InternalDriverContext context) {
     return builder.build();
   }
 
+  /**
+   * Handle a few oddities in Cassandra 4: some options still appear in system_schema.tables, but
+   * they are not valid in CREATE statements anymore. We need to exclude them from our metadata,
+   * otherwise {@link Describable#describe(boolean)} will generate invalid CQL.
```
olim7t, line 93: I'm not sure if this is normal, maybe we should open a Cassandra ticket to remove those options from the system tables.

olim7t: [CASSANDRA-15604](https://issues.apache.org/jira/browse/CASSANDRA-15604)

olim7t:
Force-pushed to rebase on latest 4.x, I had other test fixes there so this branch should hopefully pass CI now.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/parsing/RelationParser.java
olim7t: After some discussion on the C* ticket, we agreed to not fix it, so we'll keep this workaround.

integration-tests/src/test/java/com/datastax/oss/driver/core/metadata/DescribeIT.java
```diff
@@ -50,207 +51,112 @@
...
+      } else {
+        LOG.debug("best so far");
+        bestVersion = candidateVersion;
+        bestFile = candidate;
       }
     }
-    return "";
+    if (bestFile == null) {
+      fail("Could not find create script with version <= %s in %s", serverVersion, scriptsDir);
+      throw new AssertionError(); // never reached
```
emerkle826: nit: We don't need this, do we? The test will fail on the line above if we get here.

emerkle826 (approved):
This looks good, just one really minor nit.

integration-tests/src/test/java/com/datastax/oss/driver/core/metadata/DescribeIT.java
olim7t: It's only here to avoid a compile-time error, the compiler doesn't know that `fail()` always throws.

But as I'm writing this I realize that there is a clearer way to do it: `assertThat(bestFile).isNotNull()`. I'll change it.

emerkle826: That works too :+1:

PR #1402, opened 2020-03-02
JAVA-2667: Add ability to fail the build when integration tests fail

adutra:
And at the same time, this PR makes it possible to finish the entire integration test suite before failing the build, if any of the 3 executions (parallel, serial, isolated) fails.


olim7t (approved):
Makes perfect sense 👍

PR #1403, opened 2020-03-04
JAVA-2664: Add a callback to inject the session in node state listeners

olim7t:
Work in progress, I'm pushing early to have something concrete to discuss on the JIRA ticket.

TODO
- [x] tests
- [x] do the same on `SchemaChangeListener`.
  - I think we also need the replay mechanism, because the control connection can receive schema events before the session is fully initialized.
  - One thing to warn users about is to not do schema-altering queries in a listener. Although it seems unlikely that they would do it in a way that causes infinite recursion (e.g. every callback invocation creates a table).
- [x] maybe do the same on `RequestTracker`. But here the risk of recursion is much higher, users must actively defend against it (e.g. by using a profile) otherwise every `session.execute` call from within the listener will in turn invoke a callback.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/metadata/SafeInitNodeStateListener.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/metadata/SafeInitNodeStateListener.java
@@ -1,0 +1,190 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core.metadata;
+
+import com.datastax.oss.driver.api.core.session.Session;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.BiConsumer;
+import net.jcip.annotations.GuardedBy;
+
+/**
+ * A node state listener wrapper that delays (or ignores) init events until after the session is
+ * ready.
+ *
+ * <p>By default, the driver calls node state events, such as {@link #onUp} and {@link #onAdd},
+ * before the session is ready; see {@link NodeStateListener#onSessionReady(Session)} for a detailed
+ * explanation. This can make things complicated if your listener implementation needs the session
+ * to process those events.
+ *
+ * <p>This class wraps another implementation to shield it from those details:
+ *
+ * <pre>
+ * NodeStateListener delegate = ... // your listener implementation
+ *
+ * SafeInitNodeStateListener wrapper =
+ *     new SafeInitNodeStateListener(delegate, true);
+ *
+ * CqlSession session = CqlSession.builder()
+ *     .withNodeStateListener(wrapper)
+ *     .build();
+ * </pre>
+ *
+ * With this setup, {@code delegate.onSessionReady} is guaranteed to be invoked first, before any
+ * other method. The second constructor argument indicates what to do with the method calls that
+ * were ignored before that:
+ *
+ * <ul>
+ *   <li>if {@code true}, they are recorded, and replayed to {@code delegate} immediately after
+ *       {@link #onSessionReady}. They are guaranteed to happen in the original order, and before
+ *       any post-initialization events.
+ *   <li>if {@code false}, they are discarded.
+ * </ul>
+ *
+ * @since 4.6.0
+ */
+public class SafeInitNodeStateListener implements NodeStateListener {
+
+  private final NodeStateListener delegate;
+  private final boolean replayInitEvents;
+
+  // Write lock: recording init events or setting sessionReady
+  // Read lock: reading init events or checking sessionReady
+  private final ReadWriteLock lock = new ReentrantReadWriteLock();
+
+  @GuardedBy("lock")
+  private boolean sessionReady;
+
+  @GuardedBy("lock")
+  private final List<InitEvent> initEvents = new ArrayList<>();
+
+  /**
+   * Creates a new instance.
+   *
+   * @param delegate the wrapped listener, to which method invocations will be forwarded.
+   * @param replayInitEvents whether to record events during initialization and replay them to the
+   *     child listener once it's created, or just ignore them.
+   */
+  public SafeInitNodeStateListener(@NonNull NodeStateListener delegate, boolean replayInitEvents) {
+    this.delegate = Objects.requireNonNull(delegate);
+    this.replayInitEvents = replayInitEvents;
+  }
+
+  @Override
+  public void onSessionReady(@NonNull Session session) {
+    lock.writeLock().lock();
+    try {
+      if (!sessionReady) {
+        sessionReady = true;
+        delegate.onSessionReady(session);
+        if (replayInitEvents) {
+          for (InitEvent event : initEvents) {
+            event.invoke(delegate);
+          }
+        }
+      }
+    } finally {
+      lock.writeLock().unlock();
+    }
+  }
+
+  @Override
+  public void onAdd(@NonNull Node node) {
+    onEvent(node, InitEvent.Type.ADD);
+  }
+
+  @Override
+  public void onUp(@NonNull Node node) {
+    onEvent(node, InitEvent.Type.UP);
+  }
+
+  @Override
+  public void onDown(@NonNull Node node) {
+    onEvent(node, InitEvent.Type.DOWN);
+  }
+
+  @Override
+  public void onRemove(@NonNull Node node) {
+    onEvent(node, InitEvent.Type.REMOVE);
+  }
+
+  private void onEvent(Node node, InitEvent.Type eventType) {
+
+    // Cheap case: the session is ready, just delegate
+    lock.readLock().lock();
+    try {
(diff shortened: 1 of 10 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/metadata/SessionAwareNodeStateListener.java
```diff
@@ -0,0 +1,65 @@
...
+ * <ul>
+ *   <li>First the driver shuffles the contact points, and tries each one sequentially. For any
+ *       contact point that can't be reached, {@link #onDown(Node)} is invoked; for the one that
+ *       eventually succeeds, {@link #onUp(Node)} is invoked and that node becomes the control node
+ *       (if none succeeds, the session initialization fails and the process stops here).
+ *   <li>The control node's {@code system.peers} table is inspected to discover the remaining nodes
+ *       in the cluster. For any node that wasn't already a contact point, {@link #onAdd(Node)} is
+ *       invoked; for any contact point that doesn't have a corresponding entry in the table, {@link
+ *       #onRemove(Node)} is invoked;
+ *   <li>The load balancing policy computes the nodes' {@link NodeDistance distances}, and, for each
```
adutra: Nit: linkplain would look better.

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
```diff
@@ -431,7 +433,7 @@ private void afterInitialSchemaRefresh(CqlIdentifier keyspace) {
       }
     }
 
-    private void notifyLifecycleListeners() {
+    private void notifyListeners() {
       for (LifecycleListener lifecycleListener : context.getLifecycleListeners()) {
         try {
           lifecycleListener.onSessionReady();
```
adutra, line 439: Shouldn't we pass the session here too?

adutra (approved):

amichair:
First of all, I have to say I really appreciate the quick, thorough and super helpful responses... it's quite remarkable :-)

I have a few thoughts about the design and implementation in this PR. First of all, in general it looks correct and solves the problem. However, I think while the implementation may be good, perhaps it would be better to keep as much of it internal (blackbox) as possible, and leave the public API more simple and consistent. A few examples:
- Instead of the factory method passed to SafeInitNodeStateListenerWrapper accepting a Session, it could be no-arg and just provide the listener instance. The session can be passed via the already existing onSessionReady listener method if needed, to stay consistent (but in this case it would of course be called first before the other events).
- It may not be necessary for SafeInitNodeStateListenerWrapper to accept a factory method at all, rather it can just take a listener instance. This is more consistent with existing code (withNodeStateListener) that just accepts a listener, which allows the calling class to itself implement a listener interface and just pass 'this' (as we currently do, at least in our code). I'm not sure what use case the lazy initialization is needed for.
- SafeInitNodeStateListenerWrapper can stay internal, and just provide an overloaded builder.withNodeStateListener(SessionAwareNodeStateListener, boolean) to the caller that uses the wrapper internally. That's a minimal change to the API, very consistent with the existing one, that simply extends it with the newly added functionality, and less client boilerplate code. Then the implementation class is of no concern to the caller.

As for the implementation, just a couple of potential refactoring ideas:
- The initEvents lazy initialization can be replaced with regular field initialization to slightly simplify the code. There is no performance improvement to be gained here.
- Each Type enum constant can contain its associated BiConsumer listener method as a field. Then you don't need to pass it around explicitly, you can do away with the switch statement, etc. I think it'll be cleaner and shorter that way.

Those are just some ideas... in any case it looks good :-)

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
olim7t: This has worked without it so far, so I see no reason to change it now. If we reuse `LifecycleListener` in the future, and the new use case requires the session, we can change it then.

olim7t:
> Instead of the factory method passed to SafeInitNodeStateListenerWrapper accepting a Session, it could be no-arg and just provide the listener instance. The session can be passed via the already existing onSessionReady listener method if needed [ ... ] It may not be necessary for SafeInitNodeStateListenerWrapper to accept a factory method at all, rather it can just take a listener instance

The goal of the factory is to pass the session to the child listener as a constructor parameter. This way it can be stored in a final field and the implementation doesn't have to deal concurrency at all (volatile, etc).

> SafeInitNodeStateListenerWrapper can stay internal, and just provide an overloaded builder.withNodeStateListener(SessionAwareNodeStateListener, boolean) to the caller that uses the wrapper internally

Not everyone will want/need the wrapper. Some listeners may be legitimately interested in getting the events exactly when they are emitted.

> The initEvents lazy initialization can be replaced with regular field initialization to slightly simplify the code

Yes. I wanted to avoid allocating the list if it's not going to be used (`replayInitEvents == false`), but that's negligible, I agree code simplicity is more important.

> Each Type enum constant can contain its associated BiConsumer listener method as a field. Then you don't need to pass it around explicitly, you can do away with the switch statement, etc

Right, I see what you mean, will do.

amichair:
> 
> The goal of the factory is to pass the session to the child listener as a constructor parameter. This way it can be stored in a final field and the implementation doesn't have to deal concurrency at all (volatile, etc).
> 
Personally I don't think making it volatile instead of final is a problem, and it's likely that most of what is done in the event listener methods needs to deal with concurrency anyway in whatever else they do (they are invoked by a different non-user thread, not sure about re-entrant or concurrent invocation guarantees). Using a driver-invoked constructor also means you can't easily configure or inject other things into it simply (without additional workarounds), or use existing listener code. I would vote for keeping the API simple, backwards-compatible and consistent and use what's already there - but maybe that's just me :-) It's also possible to support both ways (factory and instance), but I'm not sure that's needed either. it's up to you :-)
> > SafeInitNodeStateListenerWrapper can stay internal, and just provide an overloaded builder.withNodeStateListener(SessionAwareNodeStateListener, boolean) to the caller that uses the wrapper internally
> 
> Not everyone will want/need the wrapper. Some listeners may be legitimately interested in getting the events exactly when they are emitted.
> 
Yep, that's why I suggested an overloaded method. If you want previous behavior, use the existing method (single listener parameter), fully backward-compatible. if you want session and ignore all previous events, use new method with false parameter, if you want session first and all previous events replayed, use new method with true parameter. Or use three one-arg methods with slightly different descriptive names to avoid the boolean if you prefer that (withReplayingNodeStatelistener or whatever). One-liner in the client code either way, using the same builder construct, without dealing with additional classes and code. btw for even more flexibility the new method can accept the base NodeStateListener, and only call onSessionReady if it's an instance of SessionAwareNodeStateListener (an improvement on what I wrote above).

amichair:
btw refactorings look good!

I just noticed you can take it a step further and remove the enum altogether, use the method reference directly instead, although it's a matter of taste. Shorter code, but the explicit enum might be useful in the future or more readable to some. Either way is fine.

olim7t:
Back on this PR, I've force-pushed to rebase on the latest 4.x.

In hindsight I agree that my design was a bit convoluted. I've made the following simplifications:
- `onSessionReady` is now directly on `NodeStateListener` with a default empty implementation. This avoids introducing another interface.
- the wrapper passes the session to the delegate via `onSessionReady`, like for a normal listener.
- the wrapper takes the delegate as a parameter instead of a factory.

olim7t:
Done for `SchemaChangeListener`.

I'm tempted to leave `RequestTracker` alone for now: the use case is less obvious, and it's really a recipe for disaster if the listener executes requests that recurse back into itself indefinitely.
We can always add it later following the same template.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/SafeInitNodeStateListener.java
```diff
@@ -0,0 +1,188 @@
...
+  public SafeInitNodeStateListener(@NonNull NodeStateListener delegate, boolean replayInitEvents) {
+    this.delegate = Objects.requireNonNull(delegate);
+    this.replayInitEvents = replayInitEvents;
+  }
+
+  @Override
+  public void onSessionReady(@NonNull Session session) {
+    lock.writeLock().lock();
+    try {
+      sessionReady = true;
```
adutra: Should we check `sessionReady` to prevent multiple invocations of the delegate's methods (especially double delivery of init events)? I know the driver won't do this, but application code might.

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/api/core/metadata/SafeInitNodeStateListener.java
olim7t: Hrmm not sure if I follow, the boolean is flipped only once at the end of init, how could it prevent double delivery?

adutra: I was just suggesting the following idiom:
```
lock.writeLock().lock();
try {
	if (!sessionReady) {
		sessionReady = true;
		delegate.onSessionReady(session);
		if (replayInitEvents) {
			for (InitEvent event : initEvents) {
				event.invoke(delegate);
			}
		}
	}
} finally {
	lock.writeLock().unlock();
}
```

olim7t: I see. Yeah I'll add it, as you noted it's not supposed to happen, but it doesn't hurt to have an extra check.

core/src/main/java/com/datastax/oss/driver/internal/core/session/SchemaListenerNotifier.java
```diff
@@ -33,6 +34,11 @@
   private final SchemaChangeListener listener;
   private final EventExecutor adminExecutor;
 
+  // It is technically possible that a schema change could happen in the middle of session
+  // initialization. Don't forward events in this case, it would likely do more harm than good if a
+  // listener implementation doesn't expect it.
+  private boolean sessionReady;
```
olim7t, line 40: Unlike node events, I think it's highly unlikely that anyone would be interested in schema changes that happened before the session was ready. So I changed the driver internals to ignore them, instead of introducing a replay mechanism.

olim7t:
I ended up doing the request tracker as well, with a warning about the infinite recursion pitfall.

This should be good to go.

adutra (approved):

PR #1404, opened 2020-03-04
JAVA-1861: Add Metadata.getClusterName()

emerkle826:


core/src/main/java/com/datastax/oss/driver/api/core/metadata/Metadata.java
```diff
@@ -114,4 +114,10 @@
    */
   @NonNull
   Optional<TokenMap> getTokenMap();
+
+  /** The cluster name for this cluster. */
```
olim7t: "for this session" to be more accurate.

Please also explain that this corresponds to the value of `system.local.cluster_name` on the server, just to lift any ambiguity.

Finally, also mention that this method has a default implementation only for the sake of backward compatibility: in the unlikely case that someone implements the interface in their code (e.g. a wrapper), they are expected to override it.
(btw I agree with the decision to make it default, it's simpler than introducing yet-another-subinterface)

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/InitialNodeListRefresh.java
```diff
@@ -62,7 +62,12 @@ public Result compute(
 
     ImmutableMap.Builder<UUID, DefaultNode> newNodesBuilder = ImmutableMap.builder();
 
+    String clusterName = null;
+
     for (NodeInfo nodeInfo : nodeInfos) {
+      if (clusterName == null) {
+        clusterName = nodeInfo.getClusterName();
```
olim7t: The cluster name is also available from `context.getChannelFactory()` (currently package-private but we can add a getter for it): we re-check it on every new connection to ignore foreign nodes if there are stale entries in the peers table ([JAVA-397](https://datastax-oss.atlassian.net/browse/JAVA-397)).

I think it might be easier to grab it from there, this avoids all the impacts on the topology monitor.

olim7t (commented):
Looks good but I think we can simplify a bit, see inline comment.

integration-tests/src/test/java/com/datastax/oss/driver/core/metadata/SchemaIT.java
```diff
@@ -127,6 +127,7 @@ public void should_enable_schema_programmatically_when_disabled_in_config() {
               CqlIdentifier.fromInternal("system"),
               CqlIdentifier.fromInternal("system_traces"),
               sessionRule.keyspace());
+      assertThat(session.getMetadata().getClusterName().get()).isEqualTo("ccm_1");
```
olim7t: This feels out of place in this schema test. I don't see anywhere else to put it, so I'll create a separate class.

olim7t (approved):
👍 
One last minor remark on the integration test, but I'll address it directly and merge.

integration-tests/src/test/java/com/datastax/oss/driver/core/metadata/SchemaIT.java
emerkle826: I wanted to include an IT test that verified the cluster name, so I added it here. I meant to revisit it and either find a better place or create a new IT test but forgot. Thanks for catching it.

PR #1405, opened 2020-03-05
JAVA-2670: Provide base class for mapped custom codecs

adutra:


core/src/main/java/com/datastax/oss/driver/api/core/type/codec/MappingCodec.java
```diff
@@ -0,0 +1,156 @@
...
+ * <p>See the <a
+ * href="https://docs.datastax.com/en/developer/java-driver/latest/manual/core/custom_codecs/">driver
+ * documentation on custom codecs</a> for more information and a few usage examples.
+ *
+ * @param <InnerT> The "inner" Java type; must be a driver supported Java type (that is, there must
+ *     exist a codec registered for it).
+ * @param <OuterT> The "outer", or target Java type; this codec will handle the mapping to and from
+ *     {@code InnerT} and {@code OuterT}.
+ */
+public abstract class MappingCodec<InnerT, OuterT> implements TypeCodec<OuterT> {
```
tomekl007, line 44: maybe instead of `InnerT` and `OuterT` we could simplify it a bit?
so `InnerT` would be `InT` and `OUter` would be `OutT`

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/MappingCodec.java
```diff
@@ -0,0 +1,156 @@
...
+   * the {@linkplain #encode(Object, ProtocolVersion) encode} method is covariant with {@code
+   * JavaTypeT}.
+   */
+  @Override
+  public boolean accepts(@NonNull Object value) {
+    Preconditions.checkNotNull(value);
+    if (getJavaType().getRawType().isAssignableFrom(value.getClass())) {
+      try {
+        @SuppressWarnings("unchecked")
+        OuterT outer = (OuterT) value;
```
tomekl007: if someone will create a custom codec and mess with types he will get `CastException` right? Maybe we could treat that `Exception` differently and log it?

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/MappingCodec.java
adutra: Shorter is not necessarily better, but is often less clear. Here "InT" and "OutT" could be understood as : "IncomingT" and "OutgoingT" which is not exactly right.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/MappingCodec.java
adutra: The cast here cannot fail, since we have a check for `getJavaType().getRawType().isAssignableFrom(value.getClass())` earlier.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/MappingCodec.java
```diff
@@ -0,0 +1,156 @@
...
+ *
+ * <p>This codec can be used to provide support for Java types that are not natively handled by the
+ * driver, as long as there is a conversion path to and from another supported Java type.
+ *
+ * <p>See the <a
+ * href="https://docs.datastax.com/en/developer/java-driver/latest/manual/core/custom_codecs/">driver
+ * documentation on custom codecs</a> for more information and a few usage examples.
+ *
+ * @param <InnerT> The "inner" Java type; must be a driver supported Java type (that is, there must
+ *     exist a codec registered for it).
```
olim7t, line 34: You could link to https://docs.datastax.com/en/developer/java-driver/latest/manual/core/#cql-to-java-type-mapping for the exhaustive list.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/MappingCodec.java
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * Creates a new mapping codec providing support for {@code OuterT} based on an existing codec for
+   * {@code InnerT}.
+   *
+   * @param innerCodec The inner codec to use to handle instances of InnerT; must not be null.
+   * @param outerJavaType The outer Java type; must not be null.
+   */
+  public MappingCodec(
+      @NonNull TypeCodec<InnerT> innerCodec, @NonNull GenericType<OuterT> outerJavaType) {
```
olim7t, line 57: nit: maybe make this constructor protected (although I don't see any real issue with it being public)

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/MappingCodec.java
```diff
@@ -0,0 +1,156 @@
...
+  /** @return The type of {@code InnerT}. */
+  public GenericType<InnerT> getInnerJavaType() {
+    return innerCodec.getJavaType();
+  }
+
+  /**
+   * Whether this codec is capable of encoding the given Java object.
+   *
+   * <p>This implementation first checks if the value's runtime type is compatible with {@code
+   * OuterT}, then checks if the converted value's runtime type is compatible with {@code InnerT}.
```
olim7t: Is that second check really necessary? It costs us an additional invocation of `outerToInner()`. The only thing that could cause a mismatch would be a defective implementation of that method.

integration-tests/src/test/java/com/datastax/oss/driver/core/type/codec/registry/CodecRegistryIT.java
```diff
@@ -463,4 +433,120 @@ public void should_be_able_to_handle_empty_collections() {
...
+    @Override
+    protected UdtValue outerToInner(@Nullable Coordinates value) {
+      return value == null
+          ? null
+          : getCqlType().newValue().setInt("x", value.x).setInt("y", value.y);
+    }
+  }
+
+  @Test
+  public void should_be_able_to_register_and_use_custom_codec_for_user_defined_type() {
```
olim7t: nit: when a test method starts with `should_be_able_to_xxx`, it can generally be shortened to `should_xxx`

manual/core/custom_codecs/README.md
```diff
@@ -191,6 +191,106 @@ String s1 = row.getString("anIntColumn");         // int -> String, will decode
 String s2 = row.get("anIntColumn", specialCodec); // int -> String, will decode with specialCodec
 ``` 
 
+### Creating custom Java-to-CQL mappings with `MappingCodec`
+
+The above example, `CqlIntToStringCodec`, could be rewritten to leverage [MappingCodec], an abstract 
+class that ships with the driver. This class has been designed for situations where we want to 
+represent a CQL type with a different Java type than the Java type natively supported by the driver,
+and the conversion between the former and the latter is straightforward.
```
olim7t, line 199: 👍 I think it's a good idea to keep the first example as-is, and only introduce the mapping codec in a later section.

manual/core/custom_codecs/README.md
```diff
@@ -258,4 +358,6 @@ private static String formatRow(Row row) {
 [CodecRegistry]: https://docs.datastax.com/en/drivers/java/4.5/com/datastax/oss/driver/api/core/type/codec/registry/CodecRegistry.html
 [GenericType]:   https://docs.datastax.com/en/drivers/java/4.5/com/datastax/oss/driver/api/core/type/reflect/GenericType.html
 [TypeCodec]:     https://docs.datastax.com/en/drivers/java/4.5/com/datastax/oss/driver/api/core/type/codec/TypeCodec.html
-[SessionBuilder.addTypeCodecs]: https://docs.datastax.com/en/drivers/java/4.5/com/datastax/oss/driver/api/core/session/SessionBuilder.html#addTypeCodecs-com.datastax.oss.driver.api.core.type.codec.TypeCodec...-
\ No newline at end of file
+[MappingCodec]:     https://docs.datastax.com/en/drivers/java/4.5/com/datastax/oss/driver/api/core/type/codec/MappingCodec.html
+[SessionBuilder.addTypeCodecs]: https://docs.datastax.com/en/drivers/java/4.5/com/datastax/oss/driver/api/core/session/SessionBuilder.html#addTypeCodecs-com.datastax.oss.driver.api.core.type.codec.TypeCodec...-
+[object mapping framework]: https://docs.datastax.com/en/developer/java-driver/4.5/manual/mapper/
```
olim7t: This link should be relative.

manual/core/custom_codecs/README.md
```diff
@@ -191,6 +191,106 @@ String s1 = row.getString("anIntColumn");         // int -> String, will decode
...
+}
+```
+
+Then the new mapping codec could be registered as follows:
+
+```java
+CqlSession session = ...
+CodecRegistry codecRegistry = session.getContext().getCodecRegistry();
+// The target user-defined type
+UserDefinedType coordinatesUdt = ...;
```
olim7t: Maybe show the actual code to retrieve the type from schema metadata, even if that involves a dummy keyspace name.

olim7t (approved):
LGTM, minor nits inline.

Maybe we could also add a driver example for UDT mapping.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/MappingCodec.java
adutra: I think it is, at least for correctness. With the example of `CqlIntToStringCodec` in mind, without this extra check the method would return true for the String `"NaN"`, but that string cannot be encoded as a CQL int.

That said, I agree that if we remove the extra check, there would probably be no visible consequence: let's assume that we are calling `CodecRegistry.codecFor(DataTypes.INT, "123")`. The built-in codec for `DataTypes.INT` does not accept `"123"` and would not be selected. Only a user codec accepting both `DataTypes.INT` and an instance of String can be selected, and so `CqlIntToStringCodec` is the only codec that could be returned, regardless if the extra check is present or not.

adutra:
Commit af353cd adds examples for UDTs and tuples, both simple and mapped. We might want to keep this commit separate.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/MappingCodec.java
adutra: I'm revisiting this before the merge and I think I agree we can avoid the extra conversion. The concrete consequence is that `codec.accepts("not an int")` will return true for a `MappingCodec<Integer,String>`. I'm going to merge this because I need it for JAVA-2691, we can discuss this offline if you wish.

PR #1406, opened 2020-03-05
Add V4+ features: `DATE_TYPE` and `SMALLINT_AND_TINYINT_TYPES`

dalaro:
This would be nice-to-have (though not essential) for the Spark connector.  The connector code necessarily checks the protocol version in several places when generating queries and thinking about schema.  These checks currently look like regular integer-primitive comparisons, which isn't ideal for readability.  Descriptively-named features are more readable, when available.

Following @adutra's  helpful preliminary comment, I split these features along OSS *C issue lines:

* https://jira.apache.org/jira/browse/CASSANDRA-7523 - date type support
* https://jira.apache.org/jira/browse/CASSANDRA-8951 - combined smallint & tinyint type support

This passed the coveo formatter enforcement and local testing (`mvn -pl core test -Dtest=DefaultProtocolVersionRegistryTest`).

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/DefaultProtocolFeature.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/DefaultProtocolFeature.java
@@ -35,5 +35,19 @@ public enum DefaultProtocolFeature implements ProtocolFeature {
    * @see <a href="https://issues.apache.org/jira/browse/CASSANDRA-10145">CASSANDRA-10145</a>
    */
   PER_REQUEST_KEYSPACE,
+
+  /**
+   * Support for smallint and tinyint types.
+   *
+   * @see <a href="https://jira.apache.org/jira/browse/CASSANDRA-8951">CASSANDRA-8951</a>
+   */
+  SMALLINT_AND_TINYINT_TYPES,
+
+  /**
+   * Support for the date type.
+   *
+   * @see <a href="https://jira.apache.org/jira/browse/CASSANDRA-7523">CASSANDRA-7523</a>
+   */
+  DATE_TYPE,
   ;
 }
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/DefaultProtocolFeature.java
```diff
@@ -35,5 +35,19 @@
...
+   * @see <a href="https://jira.apache.org/jira/browse/CASSANDRA-8951">CASSANDRA-8951</a>
+   */
+  SMALLINT_AND_TINYINT_TYPES,
+
+  /**
+   * Support for smallint and tinyint types.
+   *
+   * @see <a href="https://jira.apache.org/jira/browse/CASSANDRA-7523">CASSANDRA-7523</a>
+   */
+  DATE_TYPE,
```
olim7t, line 51: nit: wrong javadoc (copy-paste)

olim7t (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/DefaultProtocolFeature.java
dalaro: Thanks, whoops.  Looks like I introduced that [when splitting `DATE_SMALLINT_TINYINT_TYPES` into two separate enum values](https://github.com/dalaro/java-driver/commit/b3fffeef6c0a2a6704b9f537f5610b7b730e92ed).

I force-pushed a fix (and also rebased onto the latest 4.x).

adutra (approved):

olim7t:
Squashed and rebased on new 4.5.x branch: since this has very little impact, we're targeting our upcoming 4.5.1 patch release.

PR #1408, opened 2020-03-06
JAVA-2673: Fix mapper generated code for UPDATE with TTL and IF condition

emerkle826:


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoUpdateMethodGenerator.java
```diff
@@ -195,9 +195,10 @@ private void generatePrepareRequest(
         methodBuilder, requestName, helperFieldName, annotation.customWhereClause());
     maybeAddTtl(annotation.ttl(), methodBuilder);
     maybeAddTimestamp(annotation.timestamp(), methodBuilder);
+    methodBuilder.addCode(")");
     maybeAddIfClause(methodBuilder, annotation);
 
-    methodBuilder.addCode(").asCql()");
+    methodBuilder.addCode(".asCql()");
```
emerkle826, line 201: I think this is the right fix, to close the DefaultUpdate cast before starting the conditional created with the "if clause".

olim7t: Yes, that works.

I was afraid that throwing a timestamp on top of TTL + customIf was going to create another issue, but actually it works out because `usingTimestamp` can be chained after `usingTtl`.

Decomposing the generated code:
```java
// The declared type of productHelper is the implementation class (not EntityHelper<Product>),
//  and that class declares DefaultUpdate as the return type
DefaultUpdate u1 = productHelper.updateByPrimaryKey()

UpdateStart u2 = u1.usingTtl(..);

// Normal chained call, usingTimestamp is visible from UpdateStart
UpdateStart u3 = u2.usingTimestamp(...);

// Now we run into an issue because ifRaw is not visible on UpdateStart
DefaultUpdate u4 = (DefaultUpdate) u3;
Update u5 = u4.ifRaw(...);
```

This is a bit messy: on the one hand the progressive interfaces are nice to guide users when they are writing code by hand, on the other they get in the way when we need to generate the clauses out of order in generated code.

olim7t (approved):
LGTM
Let's hold the merge until we figure out if we want to do a patch release for this.

tomekl007 (approved):

PR #1409, opened 2020-03-08
JAVA-2633: Add execution profile argument to DAO factory method

GregBestland:


integration-tests/src/test/java/com/datastax/oss/driver/mapper/ProfileIT.java
```diff
@@ -0,0 +1,295 @@
...
+                        com.datastax.oss.simulacron.common.codec.ConsistencyLevel.ONE,
+                        com.datastax.oss.simulacron.common.codec.ConsistencyLevel.ANY),
+                    params,
+                    paramTypes))
+                .then(noRows()));
+  }
+
+  @Mapper
+  public interface InventoryMapper {
+    @DaoFactory
```
tomekl007, line 250: Don't we want to allow users to pass the `DriverExecutionProfile` type to the `@DaoProfile` as well?

integration-tests/src/test/java/com/datastax/oss/driver/mapper/ProfileIT.java
```diff
@@ -0,0 +1,295 @@
...
+    SIMULACRON_RULE
+        .cluster()
+        .prime(
+            when(query(
+                    "DELETE FROM ks.simple WHERE pk=:pk",
+                    Lists.newArrayList(
+                        com.datastax.oss.simulacron.common.codec.ConsistencyLevel.ONE,
+                        com.datastax.oss.simulacron.common.codec.ConsistencyLevel.ANY),
+                    params,
+                    paramTypes))
```
tomekl007, line 193: I think that it can be refactored to a new method with only one parameter - string query. 
This is common logic for other `prime*Query` methods

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
```diff
@@ -146,6 +146,13 @@ protected void populateBuilderWithFunction(
     }
   }
 
+  protected void populateBuilderWithProfile(CodeBlock.Builder builder) {
+    builder.beginControlFlow("if(context.getExecutionProfileName() !=null)");
```
tomekl007: nit, missing white space after `!=`

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/mapper/MapperDaoFactoryMethodGenerator.java
```diff
@@ -118,22 +120,30 @@ public MapperDaoFactoryMethodGenerator(
...
                 DaoKeyspace.class.getSimpleName(),
-                DaoTable.class.getSimpleName());
+                DaoTable.class.getSimpleName(),
+                DaoProfile.class.getSimpleName());
         return Optional.empty();
       }
     }
     boolean isCachedByKeyspaceAndTable =
-        (keyspaceArgumentName != null || tableArgumentName != null);
+        (keyspaceArgumentName != null || tableArgumentName != null || executionProfileName != null);
```
tomekl007, line 155: I don't understand this change:
`executionProfileName != null` - could you explain?

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MapperContext.java
```diff
@@ -45,6 +45,13 @@
   @Nullable
   CqlIdentifier getTableId();
 
+  /**
+   * If this context belongs to a DAO that was built with a execution-profile-parameterized mapper
+   * method, the value of that parameter. Otherwise null.
+   */
+  @Nullable
+  String getExecutionProfileName();
```
tomekl007, line 53: you can propagate that setting in the `Map<Object, Object> getCustomState();` to not introduce a method that breaks backward compatibility.
See `MapperBuilder` and how's this is done for `SCHEMA_VALIDATION_ENABLED_SETTING`:
```

  protected MapperBuilder(CqlSession session) {
    this.session = session;
    this.customState = new HashMap<>();
    // schema validation is enabled by default
    customState.put(SCHEMA_VALIDATION_ENABLED_SETTING, true);
  }
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
```diff
@@ -146,6 +146,13 @@ protected void populateBuilderWithFunction(
...
+    builder.addStatement(
+        "boundStatementBuilder = boundStatementBuilder.setExecutionProfileName(context.getExecutionProfileName())");
```
tomekl007, line 152: are we handling `SimpleStatement`s as well? (I see logic only for boundStatements)

integration-tests/src/test/java/com/datastax/oss/driver/mapper/ProfileIT.java
GregBestland: I hope not...

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/mapper/MapperDaoFactoryMethodGenerator.java
GregBestland: The dao cache key now includes the execution profile name, as part of it. If any one or any combination of these parameters are defined we need to use the DaoCacheKey as the means of retrieval , if you only defined @DaoProfile without defining @DaoTable, or @DaoKeyspace, @DaoProfile would not be honored.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/ProfileIT.java
GregBestland: I added this, it turned out to not be a very large change

integration-tests/src/test/java/com/datastax/oss/driver/mapper/ProfileIT.java
GregBestland: It might look that way at first glance, but I think if you look again you will see this is more trouble than it's worth. Parameters and return types change

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MapperContext.java
GregBestland: I guess I'm questioning how important it is to maintain backwards compatibility with the MapperContext interface... I could stick this into a custom state, my issue with that is that then the support for the @Daoprofile annotation is implicit instead of explicit.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
GregBestland: The simple statements you are seeing are during the Dao initialization, these are prepared, however they are never executed directly. The prepared statements are cached in the DAO, and retrieved later to make bound statements then executed. I believe this is working correctly. If you can find a specific counter example please let me know... I couldn't find any instances of them being executed directly.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MapperContext.java
tomekl007: I don't think it needs to be implicit. Please see how it is handled for `schema-validation`, there is an explicit setting: 
```
public MapperBuilder<MapperT> withSchemaValidationEnabled(boolean enableSchemaValidation) {
```

integration-tests/src/test/java/com/datastax/oss/driver/mapper/ProfileIT.java
tomekl007: Thx, please add an IT test similar to this one `ProfileIT.SimpleDao simpleDao(@DaoProfile String executionProfile);`

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MapperContext.java
GregBestland: I discussed this with Olivier, he's okay with my current implementation.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/ProfileIT.java
GregBestland: Done

tomekl007:
@GregBestland jenkins build is not finishing because of the revapi change:
https://jenkins-drivers-old.build.dsinternal.org/view/java-dse-driver/job/datastax.java-driver.java2633.commit/3/CASSANDRA_VERSION=4.0,JABBA_VERSION=oracle_jdk8,OS_VERSION=ubuntu%2Fbionic64%2Fjava-driver/console

GregBestland:
> @GregBestland jenkins build is not finishing because of the revapi change:
> https://jenkins-drivers-old.build.dsinternal.org/view/java-dse-driver/job/datastax.java-driver.java2633.commit/3/CASSANDRA_VERSION=4.0,JABBA_VERSION=oracle_jdk8,OS_VERSION=ubuntu%2Fbionic64%2Fjava-driver/console

I think the revapi change was added yesterday.... I ran it locally and it was fine. I restarted the build. Let's wait for that to finish.

GregBestland:
I think everything should be good to go now.

tomekl007 (approved):
LGTM

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DefaultMapperContext.java
```diff
@@ -73,7 +92,19 @@ public DefaultMapperContext withKeyspaceAndTable(
             && Objects.equals(newTableId, this.tableId))
         ? this
         : new DefaultMapperContext(
-            session, newKeyspaceId, newTableId, nameConverterCache, customState);
+            session, newKeyspaceId, newTableId, null, nameConverterCache, customState);
+  }
+
+  public DefaultMapperContext withExecutionProfile(@Nullable String newExecutionProfile) {
+    return newExecutionProfile.equals(this.executionProfile)
```
olim7t, line 99: IDEA warning: Method invocation 'equals' may produce 'NullPointerException'

You can use `Objects.equals()` instead.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DefaultMapperContext.java
```diff
@@ -73,7 +92,19 @@ public DefaultMapperContext withKeyspaceAndTable(
...
+        ? this
+        : new DefaultMapperContext(
+            session, keyspaceId, tableId, newExecutionProfile, nameConverterCache, customState);
+  }
+
+  public DefaultMapperContext withExecutionProfile(
+      @Nullable DriverExecutionProfile newExecutionProfile) {
+    return withExecutionProfile(newExecutionProfile.getName());
```
olim7t, line 107: This won't work with [derived profiles](https://docs.datastax.com/en/developer/java-driver/4.5/manual/core/configuration/#derived-profiles): the name references the profile they were copied from, it won't reflect any option that was changed dynamically.

We need a separate `DriverExecutionProfile` field in the context. It's the same as statement types in the core driver.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/DaoProfile.java
```diff
@@ -0,0 +1,46 @@
...
+ * <p>*
+ *
+ * <pre>
+ *  * &#64;Mapper
+ *  * public interface InventoryMapper {
+ *  *   ProductDao productDao(@DaoTable String executionProfile);
+ *  * }
+ *  * </pre>
+ *
+ * The annotated parameter can be a {@link String}. If it is present, the value will be injected in
```
olim7t, line 38: It can also be a `DriverExecutionProfile`.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/mapper/MapperDaoFactoryMethodGenerator.java
```diff
@@ -176,33 +210,64 @@ public MapperDaoFactoryMethodGenerator(
...
-              annotation.getSimpleName());
+              String.class.getSimpleName(),
+              CqlIdentifier.class.getSimpleName());
+      return null;
+    }
+    return candidate.getSimpleName().toString();
+  }
+
+  private String validateExecutionProfile(
+      VariableElement candidate, String previous, Class<?> annotation, ProcessorContext context) {
```
olim7t, line 235: IDEA warning: Actual value of parameter 'annotation' is always 'DaoProfile.class'

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/mapper/MapperDaoFactoryMethodGenerator.java
olim7t: Maybe we should rename the variable for clarity.

`isCachedByKeyspaceAndTableAndExecutionProfile` is a mouthful, so maybe `isCachedByMethodArguments` instead.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/mapper/MapperDaoFactoryMethodGenerator.java
```diff
@@ -160,14 +179,29 @@ public MapperDaoFactoryMethodGenerator(
...
       }
-      overridingMethodBuilder
-          .addCode(");\n")
-          .addStatement(
-              "return $L.computeIfAbsent(key, "
-                  + "k -> $T.$L(context.withKeyspaceAndTable(k.getKeyspaceId(), k.getTableId())))",
-              fieldName,
-              daoImplementationName,
-              isAsync ? "initAsync" : "init");
+      overridingMethodBuilder.addCode(", ");
```
olim7t, line 182: nit: the comment at lines 168-169 is obsolete

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/mapper/MapperDaoFactoryMethodGenerator.java
```diff
@@ -160,14 +179,29 @@ public MapperDaoFactoryMethodGenerator(
...
+      if (executionProfileName == null) {
+        overridingMethodBuilder.addCode("($T)null", String.class);
+      } else {
+
+        if (!executionProfileIsClass) {
+          overridingMethodBuilder.addCode("$L", executionProfileName);
+        } else {
+          overridingMethodBuilder.addCode("$L.getName()", executionProfileName);
```
olim7t, line 190: See my other comment in `DefaultMapperContext`.

olim7t (commented):
Sorry for the late review, but I think there are a couple of things we need to address. This can be done in a follow-up PR.

PR #1410, opened 2020-03-12
JAVA-2657: Add ability to specify the class loader to use for application-specific classpath resources

emerkle826:


core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -58,13 +62,33 @@
...
         () -> {
           ConfigFactory.invalidateCaches();
           Config config =
               ConfigFactory.defaultOverrides()
-                  .withFallback(ConfigFactory.parseResourcesAnySyntax(resourceBaseName))
-                  .withFallback(ConfigFactory.defaultReference())
+                  .withFallback(
+                      ConfigFactory.parseResourcesAnySyntax(
+                          resourceBaseName,
+                          ConfigParseOptions.defaults().setClassLoader(appClassLoader)))
```
emerkle826, line 88: For the static methods `fromClasspath`, `fromPath`, `fromFile`, and `fromUrl`, the idea here is to use the App ClassLoader provided for the calls to `ConfigFactory.parseXYZ()` because the ClassLoader for the App will be the one that can access any OSGi bundled resrouces.

None of these methods should be needed in a web context (with the app class loader typically being a child of the container/server class loader).

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -58,13 +62,33 @@
...
+                  .withFallback(ConfigFactory.defaultReference(DRIVER_CLASS_LOADER))
```
emerkle826: Here, we use the Driver's ClassLoader (pulled from `CqlSession.class.getClassLoader()` above) as it should be the one that will have access to the Driver's defaults (i.e. `reference.conf`)

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticDriverConfigLoaderBuilder.java
```diff
@@ -37,7 +37,9 @@
     implements ProgrammaticDriverConfigLoaderBuilder {
 
   public static final Supplier<Config> DEFAULT_FALLBACK_SUPPLIER =
-      () -> ConfigFactory.defaultApplication().withFallback(ConfigFactory.defaultReference());
+      () ->
+          ConfigFactory.defaultApplication()
+              .withFallback(ConfigFactory.defaultReference(DriverConfigLoader.DRIVER_CLASS_LOADER));
```
emerkle826: This should fix the programmatic config case where users will provide overrides programmatically for all settings that should override the Driver's defaults, and should correctly be able to access the Driver's defaults by specifying its class loader.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticDriverConfigLoaderBuilder.java
```diff
@@ -46,6 +48,25 @@
...
+   * {@code application.conf}.
+   *
+   * @param appClassLoader The application's ClassLoader from which to load application
+   *     configuration defaults.
+   */
+  public static Supplier<Config> getDefaultConfigSupplier(ClassLoader appClassLoader) {
+    return () ->
+        ConfigFactory.defaultApplication(appClassLoader)
+            .withFallback(ConfigFactory.defaultReference(DriverConfigLoader.DRIVER_CLASS_LOADER));
+  }
```
emerkle826: This will behave identically to the `DEFAULT_FALLBACK_SUPPLIER` above, except that it will not rely on the current Context ClassLoader to find application overrides (i.e. `application.conf`), and will instead use the supplied ClassLoader for that.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticDriverConfigLoaderBuilder.java
```diff
@@ -62,6 +83,18 @@ public DefaultProgrammaticDriverConfigLoaderBuilder() {
...
+   * An instance of a programmatic builder that uses application specific config defaults from the
+   * specified application CLassLoader to override Driver config defaults for config that is not
+   * supplied programmatically.
+   *
+   * @param appClassLoader The application's ClassLoader from which to load application
+   *     configuration defaults.
+   */
+  public DefaultProgrammaticDriverConfigLoaderBuilder(ClassLoader appClassLoader) {
+    this(getDefaultConfigSupplier(appClassLoader), DefaultDriverConfigLoader.DEFAULT_ROOT_PATH);
+  }
```
emerkle826: There wasn't an easy way to automatically get a hold of an OSGi bundle's ClassLoader from within the Driver's code as the Driver is a separate bundle and doesn't know which bundle's might be using it. For this to work in an OSGi container, an application bundle using the programmatic config loader will have to pass in a reference to its ClassLoader here. This allows us to create a Config supplier that can locate application bundled resources, instead of letting TypeSafe try to use the context ClassLoader, which isn't the same class loader as the OSGi app bundle ClassLoader.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -58,13 +62,33 @@
    */
   @NonNull
   static DriverConfigLoader fromClasspath(@NonNull String resourceBaseName) {
+    return fromClasspath(resourceBaseName, DRIVER_CLASS_LOADER);
+  }
+
+  /**
+   * Just like {@link #fromClasspath(java.lang.String)} except that the loader will use the provided
```
adutra: 1. I would avoid the word "loader" since we are talking about a config loader that uses a class loader... to avoid confusion I would rephrase to `... except that application configuration resources will be located using the provided {@link ClassLoader} instead of the driver's {@linkplain DriverConfigLoader#DRIVER_CLASS_LOADER default one}`. Just a suggestion, you don't have to agree :-)
2. I wonder if the second paragraph is really necessary, given that this is very hard to explain and, arguably, only of concern to a very small set of users.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -58,13 +62,33 @@
...
+   *
+   * <p>This is intended for use when an application's ClassLoader is different from the Driver's
+   * ClassLoader, and the current Context ClassLoader is not the same CLassLoader that loaded the
+   * application (eg. OSGi bundles). For web deployments not using OSGi, use of this method is
+   * normally not necessary, as the Context ClassLoader in that environment can usually access
+   * application bundled resources.
+   */
+  @NonNull
+  static DriverConfigLoader fromClasspath(
+      @NonNull String resourceBaseName, ClassLoader appClassLoader) {
```
adutra: Missing `@NonNull`.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -92,7 +116,22 @@ static DriverConfigLoader fromClasspath(@NonNull String resourceBaseName) {
...
+   * {@code ClassLoader}, as opposed to the Driver's {@code ClassLoader}, to find the {@code Path}.
+   *
+   * <p>This is intended for use when an application's ClassLoader is different from the Driver's
+   * ClassLoader, and the current Context ClassLoader is not the same CLassLoader that loaded the
+   * application (eg. OSGi bundles). For web deployments not using OSGi, use of this method is
+   * normally not necessary, as the Context ClassLoader in that environment can usually access
+   * application bundled resources.
+   */
+  @NonNull
+  static DriverConfigLoader fromPath(@NonNull Path file, ClassLoader appClassLoader) {
```
adutra: For `Path`, `File` and `URL` you don't need a class loader. The contents are retrieved by other mechanisms that do not depend on the class loader being used, so I think the overloaded methods are not useful.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticDriverConfigLoaderBuilder.java
adutra: Hmm I think this isn't enough. You need to use `defaultApplication(java.lang.ClassLoader)` instead: 

```
ConfigFactory.defaultApplication(DriverConfigLoader.DRIVER_CLASS_LOADER)
              .withFallback(ConfigFactory.defaultReference(DriverConfigLoader.DRIVER_CLASS_LOADER))
```

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticDriverConfigLoaderBuilder.java
adutra: I don't think we should expose such a method, I would instead inline this code in the constructor below.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -221,6 +294,22 @@ static ProgrammaticDriverConfigLoaderBuilder programmaticBuilder() {
...
+   * configuration defaults.
+   *
+   * <p>This is intended for use when an application's ClassLoader is different from the Driver's
+   * ClassLoader, and the current Context ClassLoader is not the same CLassLoader that loaded the
+   * application (eg. OSGi bundles). For web deployments not using OSGi, use of this method is
+   * normally not necessary, as the Context ClassLoader in that environment can usually access
+   * application bundled resources.
+   */
+  @NonNull
+  static ProgrammaticDriverConfigLoaderBuilder programmaticBuilder(ClassLoader appClassLoader) {
```
adutra: Missing `@NonNull`

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticDriverConfigLoaderBuilder.java
```diff
@@ -62,6 +83,18 @@ public DefaultProgrammaticDriverConfigLoaderBuilder() {
     this(DEFAULT_FALLBACK_SUPPLIER, DefaultDriverConfigLoader.DEFAULT_ROOT_PATH);
   }
 
+  /**
+   * An instance of a programmatic builder that uses application specific config defaults from the
```
adutra: Again I would reduce the verbosity to avoid confusing 99% of the users that don't need this: 

```
Creates an instance of {@link DefaultProgrammaticDriverConfigLoaderBuilder} that locates application configuration resources using the provided {@link ClassLoader} instead of the driver's {@linkplain DriverConfigLoader#DRIVER_CLASS_LOADER default one}.
```

adutra (changes requested):
Very good job. First round of remarks, will finish tomorrow.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticDriverConfigLoaderBuilder.java
emerkle826: I wasn't sure if it was a good idea to expose this. I will do as you suggest.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticDriverConfigLoaderBuilder.java
emerkle826: I did think about doing that, but in a web environment, it may be more limiting. If we do not provide a ClassLoader to TypeSafe's `ConfigFactory.defaultApplication()`, it will use the current Thread's Context ClassLoader. Please correct me where I'm wrong:
1) In a normal Java application, the class loaders are all the same as the context class loader.
2) In a web environment, the context class loader should be able to load resources from the app bundle ("WEB-INF/lib") as well as the web container ("/lib"). If we specify the Driver's ClassLoader here, then we could be limiting the ability to find application config if the Driver jar is in the web container's system folder ("/lib"), and the app jar is in the app folder ("WEB-INF/lib").
3) In an OSGi container, neither of these will work for finding application config in the application bundle as the Context ClassLoader isn't the ClassLoader for the application, and the Driver's ClassLoader is in a different bundle.

Also, I don't think we really need to look for Application config specifically in the Driver's ClassLoader since the Driver doesn't ship with an "application.conf". It only works if the Driver's ClassLoader is the ClassLoader for the app jar (or is a child of the app jar's ClassLoader).

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
amichair: AIUI, the problem is that the fallback default configuration isn't found - so this needs to be fixed anywhere defaultReference() is called to load the default config, regardless of how the 'overriding' configuration is loaded.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -34,6 +36,8 @@
  */
 public interface DriverConfigLoader extends AutoCloseable {
 
+  static final ClassLoader DRIVER_CLASS_LOADER = CqlSession.class.getClassLoader();
```
adutra: Maybe add javadocs here and explain briefly that this is used to load classpath config resources and that most people should not care about it.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticDriverConfigLoaderBuilder.java
adutra: You are absolutely right 👍

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
```diff
@@ -55,7 +55,8 @@
   public static final Supplier<Config> DEFAULT_CONFIG_SUPPLIER =
       () -> {
         ConfigFactory.invalidateCaches();
-        return ConfigFactory.load().getConfig(DEFAULT_ROOT_PATH);
+        return ConfigFactory.load(DriverConfigLoader.DRIVER_CLASS_LOADER)
```
adutra: Isn't this going to use `DRIVER_CLASS_LOADER` to load both application and reference files? Maybe we need to make this identical to `Default ProgrammaticDriverConfigLoaderBuilder.DEFAULT_FALLBACK_SUPPLIER`.

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -58,13 +62,27 @@
    */
   @NonNull
   static DriverConfigLoader fromClasspath(@NonNull String resourceBaseName) {
+    return fromClasspath(resourceBaseName, DRIVER_CLASS_LOADER);
```
adutra: Thinking about your comment [here](https://github.com/datastax/java-driver/pull/1410/files#r391900433): I wonder if this is the best class loader we can pass here and in other similar locations; maybe the context class loader is better because for web deployments it will cover more cases (and for OSGi, as you said, neither class loader would work anyways).

adutra:
@emerkle826 I pushed a branch with a few commits to implement my own suggestions, let me know what you think: https://github.com/datastax/java-driver/compare/java2657...java2657-adu?expand=1

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -58,13 +62,27 @@
...
+   * Just like {@link #fromClasspath(java.lang.String)} except that application configuration
+   * resources will be located using the provided {@link ClassLoader} instead of the driver's
+   * {@linkplain DriverConfigLoader#DRIVER_CLASS_LOADER default one}.
+   */
+  @NonNull
+  static DriverConfigLoader fromClasspath(
+      @NonNull String resourceBaseName, @NonNull ClassLoader appClassLoader) {
     return new DefaultDriverConfigLoader(
         () -> {
           ConfigFactory.invalidateCaches();
```
adutra, line 82: This is unrelated but I think we should improve cache invalidation, it's being done more than once in certain cases. I think it's the config loader's responsibility to invalidate the caches, not the config suppliers'.

emerkle826:
> @emerkle826 I pushed a branch with a few commits to implement my own suggestions, let me know what you think: https://github.com/datastax/java-driver/compare/java2657...java2657-adu?expand=1

I like the changes you've made. :+1:

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
emerkle826: Do you want me to create a new ticket for this, or address it in this PR?

adutra: New ticket, because this PR should be released in 4.5.1. BTW, I think you need to change the base branch to 4.5.x, right @olim7t ?

emerkle826: Will do to both.

changelog/README.md
```diff
@@ -4,6 +4,7 @@
 
 ### 4.6.0 (in progress)
 
+- [bug] JAVA-2657: DriverConfigLoader.programmaticBuilder() doesn't work in OSGi container
```
adutra: We might try to come up with a better name here, suggestion: "Ability to specify the class loader to use for application-specific classpath resources".

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
emerkle826: JAVA-2689 created for the cache improvements.

adutra: Thank you!

adutra (approved):
lgtm, I suppose the proper test for this fix is going to be JAVA-2658? Problem is, we might not be able to finish 2658 before 4.5.1 is released.

emerkle826:
> lgtm, I suppose the proper test for this fix is going to be JAVA-2658? Problem is, we might not be able to finish 2658 before 4.5.1 is released.

I'm working on it now. I have tested this some in Karaf, but that was just enough to make sure things get loaded from where I expected. I will try to get JAVA-2658 done ASAP and ready for Monday.

manual/osgi/README.md
```diff
@@ -70,6 +70,46 @@ CqlSession session = CqlSession.builder()
     .build();
 ```
 
+### Using a custom `ClassLoader` for Application Bundled configuration
+
+In addition to specifying a `ClassLoader` when constructing a `Session`, you can also specify
+a `ClassLoader` instance on certain `DriverConfigLoader` methods for cases when your OSGi
+application bundle provides overrides to Driver configuration defaults. This is typically done by
+including an `application.conf` file in your application bundle. For example, if you want to use
+[DriverConfigLoader.fromClasspath] to specify a different resource name (other than "application"):
```
adutra, line 79: Actually it would be nicer if we started by explaining the general case: using the default config loader with a custom class loader:

```
CqlSession session = CqlSession.builder()
    .withClassLoader(classLoader)
    .withConfigLoader(new DefaultDriverConfigLoader(classLoader))
    .build();
```

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -404,8 +413,20 @@ public SelfT withKeyspace(@Nullable String keyspaceName) {
...
-   * <p>If null, the driver attempts to use the same {@link ClassLoader} that loaded the core driver
-   * classes, which is generally the right thing to do.
+   * <p>Unless you define a custom {@link #configLoader}, this class loader will also be used to
+   * locate application-specific configuration resources.
+   *
+   * <p>If you do not provide any custom class loader, the driver will attempt to use the following
+   * ones:
+   *
+   * <ol>
+   *   <li>When reflectively loading class names defined in configuration: same class loader that
```
adutra, line 423: I think we can improve the defaults, I'm bootstrapping an internal debate on that. Stay tuned.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -617,7 +638,13 @@ public SessionT build() {
   @NonNull
   protected final CompletionStage<CqlSession> buildDefaultSessionAsync() {
     try {
-      DriverConfigLoader configLoader = buildIfNull(this.configLoader, this::defaultConfigLoader);
+
+      ProgrammaticArguments programmaticArguments = programmaticArgumentsBuilder.build();
+
+      DriverConfigLoader configLoader =
```
adutra, line 644: Propagate the CL if no config loader is defined.

adutra: This way we avoid having to specify the CL twice as we currently have to:

```
CqlSession session = CqlSession.builder()
    .withClassLoader(classLoader)
    .withConfigLoader(new DefaultDriverConfigLoader(classLoader))
    .build();
```

manual/osgi/README.md
adutra: I pushed a small commit that propagates the session builder's CL to the config loader, iif the user doesn't specify their own CL. SO the above snippet can be shortened to:

```
CqlSession session = CqlSession.builder()
    .withClassLoader(classLoader)
    .build();
```

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -38,6 +40,11 @@
    * Builds an instance using the driver's default implementation (based on Typesafe config), except
    * that application-specific options are loaded from a classpath resource with a custom name.
    *
+   * <p>The class loader used to locate application-specific classpath resources is {@linkplain
```
olim7t, line 44: I think we can be even more specific, to avoid any ambiguity:
```suggestion
   * <p>The class loader used to locate {@code <resourceBaseName>.*} is {@linkplain
```
And then maybe move this paragraph after the list.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -209,6 +230,11 @@ static DriverConfigLoader fromUrl(@NonNull URL url) {
    * Note that {@code application.*} is entirely optional, you may choose to only rely on the
    * driver's built-in {@code reference.conf} and programmatic overrides.
    *
+   * <p>The class loader used to locate application-specific classpath resources is {@linkplain
```
olim7t, line 237: Same as other method:
```suggestion
   * <p>The class loader used to locate {@code application.*} is {@linkplain
```

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
olim7t: sgtm 👍

olim7t (approved):
LGTM so far.
We also discussed reverting JAVA-2598, this can be done either here in a standalone commit, or in a separate PR, as you prefer.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
olim7t: This has shifted to 4.6.0 now, don't forget to change the base branch again.

emerkle826: :+1:  I'll rebase this branch with the next commits I push, and then change the PR target branch back to 4.x

emerkle826:
@adutra My last commit here: https://github.com/datastax/java-driver/pull/1410/commits/b098828e575ad41cf4c47f8d545668b3a6e6305d is very much a work in progress. Feel free to revert it, but it was my attempt at using the forked Pax exam libraries (instead of the native libraries) to try to adjust tests so they would catch classloading issues. Basically the work that would be covered by JAVA-2658.

adutra:
I fixed an issue with the programmatic builder and rebased this branch. Unfortunately I was unable to make any progress using the forked container. I found it slightly less practical than native since we can't have logs, and we can't debug the remote JVM (tried adding a `vmOptions`  config option, but it didn't work).
We are spending too much time in OSGi tests for little benefit. I suggest that instead of Pax Exam we create a sample OSGi application using Karaf and use that to validate OSGi.

adutra:
I created #1424 with a revised fix for 2657 + fixes for 2658 and 2301. I think we should close this one if everybody agrees to switch to that PR instead for more OSGi fun.

emerkle826:
> I created #1424 with a revised fix for 2657 + fixes for 2658 and 2301. I think we should close this one if everybody agrees to switch to that PR instead for more OSGi fun.

I have no objection to closing this. I'll try to review the new PR ASAP.

adutra:
Closing in favor of #1424.

PR #1411, opened 2020-03-16
JAVA-2679: Add port information to QueryTrace and TraceEvent

GregBestland:


GregBestland:
I took the approach of adding a separate port int.. another approach would have to used an InetSocketAddress and set the port there. However that would involve me modifying an existing interface method, instead of simply adding to one. Let me know what you think.

integration-tests/src/test/java/com/datastax/oss/driver/core/cql/QueryTraceIT.java
```diff
@@ -90,4 +91,23 @@ public void should_fetch_trace_when_tracing_enabled() {
...
+                    .setTracing()
+                    .build())
+            .getExecutionInfo();
+
+    assertThat(executionInfo.getTracingId()).isNotNull();
+
+    QueryTrace queryTrace = executionInfo.getQueryTrace();
+    assertThat(queryTrace.getCoordinatorPort()).isEqualTo(7000);
+    assertThat(queryTrace.getEvents().get(0).getSourcePort()).isEqualTo(7000);
+  }
```
olim7t, line 103: I think it would be better to include that in the main test, `should_fetch_trace_when_tracing_enabled`. The condition for ports to be present is:
```java
    boolean expectPorts =
        CCM_RULE.getCassandraVersion().nextStable().compareTo(Version.V4_0_0) >= 0
            && !CCM_RULE.getDseVersion().isPresent();
```
That way we can also assert that they're set to 0 when not expected.

core/src/main/java/com/datastax/oss/driver/api/core/cql/QueryTrace.java
```diff
@@ -43,6 +43,12 @@
   @NonNull
   InetAddress getCoordinator();
 
+  /**
+   * The port of the node that coordinated the query. Prior to C* 4.0 this is not set and will
+   * default to 0.
+   */
+  int getCoordinatorPort();
+
```
olim7t, line 65: > I took the approach of adding a separate port int.. another approach would have to used an InetSocketAddress and set the port there. However that would involve me modifying an existing interface method, instead of simply adding to one. Let me know what you think.

Agreed, but we can have the new method return `InetSocketAddress` if it has a different name. I suggest `getCoordinatorAddress` (for lack of a better idea). We can also make it default to avoid the breaking change:

```java
  @NonNull
  @Deprecated
  InetAddress getCoordinator();

  @NonNull
  default InetSocketAddress getCoordinatorAddress() {
    return new InetSocketAddress(getCoordinator(), 0);
  }
```
It would still be overridden in the default implementation, to return the correct port if available.

olim7t (changes requested):

core/src/main/java/com/datastax/oss/driver/api/core/cql/QueryTrace.java
GregBestland: I took your suggestion, and converted the port into an InetSocketAddress, I did this for both the TraceEvent and the QueryTrace. I was able to remove the revapi exception. Let me know if anything else needs to be addressed,  but I think this is buttoned up.

olim7t (approved):
LGTM.
A couple of last remarks but I addressed them directly in follow-up commits. This is ready to squash and merge.

PR #1412, opened 2020-03-19
JAVA-2698: TupleCodec and UdtCodec give wrong error message when parsing fails

adutra:


core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/TupleCodec.java
```diff
@@ -186,15 +187,25 @@ public TupleValue parse(@Nullable String value) {
...
-      tuple = tuple.set(i, codec.parse(fieldValue), codec);
+      Object parsed;
+      try {
+        parsed = codec.parse(fieldValue);
+      } catch (Exception e) {
+        throw new IllegalArgumentException(
+            String.format(
+                "Cannot parse tuple value from \"%s\", invalid CQL value at character %d: %s",
+                value, position, e.getMessage()),
+            e);
```
tomekl007: This can be extracted to some method because the logic of constructing this message is duplicated in a couple places

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/TupleCodec.java
```diff
@@ -186,15 +187,25 @@ public TupleValue parse(@Nullable String value) {
       } catch (IllegalArgumentException e) {
         throw new IllegalArgumentException(
             String.format(
-                "Cannot parse tuple value from \"%s\", invalid CQL value at character %d",
-                value, position),
+                "Cannot parse tuple value from \"%s\", invalid CQL value at character %d: %s",
+                value, position, e.getMessage()),
```
tomekl007: Shouldn't we throw the IllegalArgumentException with suppressed exception and pass `e` to it?

adutra: But we are passing `e` as the cause already.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/TupleCodec.java
adutra: There are only 2 occurrences of this exact message, the others are slightly different, so I think it's not worth it.

olim7t, line 213: Since we also know the field index, it would be nice to include it in the message as well:
> invalid CQL value at field 0 (character 1)

Same with the field name for UDTs.

olim7t (approved):
👍 
Minor suggestion to improve the messages even further.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/TupleCodec.java
adutra: Good idea. Done. But that made the whole changeset more involved than I initially thought it would be. Hope that's ok.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/ParseUtils.java
```diff
@@ -25,7 +25,7 @@
    * @return the index of the first character in toParse from idx that is not a "space.
    */
   public static int skipSpaces(String toParse, int idx) {
-    while (isBlank(toParse.charAt(idx)) && idx < toParse.length()) ++idx;
+    while (idx < toParse.length() && isBlank(toParse.charAt(idx))) ++idx;
```
adutra, line 28: Reported by the IDE, and indeed the order of checks would prevent this method from working in the case where `idx >= toParse.length()`.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/TupleCodec.java
```diff
@@ -161,59 +161,90 @@ public TupleValue parse(@Nullable String value) {
...
-      return tuple;
-    }
-
     CodecRegistry registry = cqlType.getAttachmentPoint().getCodecRegistry();
 
-    int i = 0;
-    while (position < value.length()) {
+    int field = 0;
+    while (position < length) {
+      if (value.charAt(position) == ')') {
```
adutra, line 181: Moved here to take care of the empty tuple case.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/TupleCodec.java
```diff
@@ -161,59 +161,90 @@ public TupleValue parse(@Nullable String value) {
...
       int n;
       try {
         n = ParseUtils.skipCQLValue(value, position);
       } catch (IllegalArgumentException e) {
         throw new IllegalArgumentException(
             String.format(
-                "Cannot parse tuple value from \"%s\", invalid CQL value at character %d",
-                value, position),
+                "Cannot parse tuple value from \"%s\", invalid CQL value at field %d (character %d)",
+                value, field, position),
```
adutra, line 198: There was no point in adding the exception's message, it is always empty.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/UdtCodec.java
```diff
@@ -165,24 +165,45 @@ public UdtValue parse(@Nullable String value) {
...
+    position++;
     position = ParseUtils.skipSpaces(value, position);
 
-    if (value.charAt(position) == '}') {
-      return udt;
+    if (position == length) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Cannot parse UDT value from \"%s\" at character %d: expecting CQL identifier or '}', got EOF",
+              value, position));
```
adutra, line 185: We don't have the field name yet, so I had to fail fast here.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/UdtCodec.java
```diff
@@ -165,24 +165,45 @@ public UdtValue parse(@Nullable String value) {
...
     }
 
     CodecRegistry registry = cqlType.getAttachmentPoint().getCodecRegistry();
 
-    while (position < value.length()) {
+    CqlIdentifier id = null;
+    while (position < length) {
+      if (value.charAt(position) == '}') {
```
adutra, line 192: Moved here for same reasons.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/TupleCodec.java
```diff
@@ -161,59 +161,90 @@ public TupleValue parse(@Nullable String value) {
...
-    int i = 0;
-    while (position < value.length()) {
+    int field = 0;
+    while (position < length) {
+      if (value.charAt(position) == ')') {
+        position++;
+        if (position == length) {
+          return tuple;
+        } else {
+          position = ParseUtils.skipSpaces(value, position);
```
adutra: These cases weren't handled correctly before.

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/TupleCodecTest.java
```diff
@@ -169,7 +171,33 @@ public void should_parse_null_tuple() {
...
+    assertThat(tuple.isNull(1)).isTrue();
+    assertThat(tuple.isNull(2)).isTrue();
+
+    verifyNoMoreInteractions(intCodec);
+    verifyNoMoreInteractions(doubleCodec);
+    verifyNoMoreInteractions(textCodec);
+  }
+
+  @Test
+  public void should_parse_partial_tuple() {
```
adutra, line 187: We don't have tests for partial tuples and udts so I added a few.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/TupleCodec.java
```diff
@@ -161,70 +161,90 @@ public TupleValue parse(@Nullable String value) {
...
+          if (position == length) {
+            return tuple;
+          }
+          throw new IllegalArgumentException(
+              String.format(
+                  "Cannot parse tuple value from \"%s\", at character %d expecting EOF or blank, but got \"%s\"",
+                  value, position, value.substring(position)));
+        }
+      }
```
olim7t, line 190: This can be simplified a little bit, given that `skipSpaces` is a no-op if there are no spaces:
```java
    while (position < length) {
      if (value.charAt(position) == ')') {
        position = ParseUtils.skipSpaces(value, position + 1);
        if (position == length) {
          return tuple;
        }
            String.format(
                "Cannot parse tuple value from \"%s\", at character %d expecting EOF or blank, but got \"%s\"",
                value, position, value.substring(position)));
      }
      int n;
```

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/TupleCodec.java
```diff
@@ -161,70 +161,90 @@ public TupleValue parse(@Nullable String value) {
...
 
       position = n;
-      i += 1;
 
       position = ParseUtils.skipSpaces(value, position);
+      if (position == length) {
+        throw new IllegalArgumentException(
+            String.format(
+                "Cannot parse tuple value from \"%s\", at field %d (character %d) expecting ',' or ')', but got EOF",
+                value, field, position));
```
olim7t, line 224: Maybe you could just break here, in order to hit the `throw` statement after the loop. It makes sense since `position == length` is the exit condition of the loop.

adutra: But the error message would have to be adapted since here we are expecting a comma or the closing `)` and at the other location we are expecting a CQL literal or `)`. I think it would add complexity to factorize the two error messages into one `throw` statement.

olim7t: I hadn't noticed the difference. Makes sense then 👍

PR #1413, opened 2020-03-19
JAVA-2691: Provide driver 4 support for extra codecs

adutra:
- [x] TODO: amend documentation on custom codecs
- [x] TODO: amend examples custom codecs


core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodecs.java
```diff
@@ -78,39 +98,171 @@
...
+  /**
+   * A codec that handles Apache Cassandra(R)'s timestamp type and maps it to Java's {@link
+   * Instant}, using the system's {@linkplain ZoneId#systemDefault() default time zone} as its
+   * source of time zone information. If you need a different time zone, consider other constants in
+   * this class, or call {@link #timestampAt(ZoneId)} instead.
+   *
+   * @see #TIMESTAMP_UTC
+   * @see #timestampAt(ZoneId)
+   */
   public static final TypeCodec<Instant> TIMESTAMP = new TimestampCodec();
```
adutra: I'm following the established pattern of proposing 2 constants and 1 method for groups of related timestamp codecs:

1. Constant for system default zone: `FOO_SYSTEM`
2. Constant for UTC: `FOO_UTC`
3. Factory method for other zones: `fooAt`

Following this logic the constant `TIMESTAMP` should have been named `TIMESTAMP_SYSTEM` but it's too late for that.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/extras/enums/EnumNameCodec.java
```diff
@@ -0,0 +1,58 @@
...
+import com.datastax.oss.driver.api.core.type.codec.MappingCodec;
+import com.datastax.oss.driver.api.core.type.codec.TypeCodecs;
+import com.datastax.oss.driver.api.core.type.reflect.GenericType;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.Nullable;
+import java.util.Objects;
+import net.jcip.annotations.Immutable;
+
+/**
+ * A codec that serializes {@link Enum} instances as CQL {@code varchar}s representing their
```
adutra, line 27: I find it a bit disappointing that javadocs for codecs are not published. They contain very valuable information. Don't you think we should start publishing javadocs for internal classes as well?

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodecs.java
```diff
@@ -153,6 +335,35 @@
...
 
+  /**
+   * Returns a codec that handles Apache Cassandra(R)'s timestamp type and maps it to Java's {@link
+   * Instant}, using the supplied {@link ZoneId} as its source of time zone information.
+   *
+   * @see #TIMESTAMP
+   * @see #TIMESTAMP_UTC
+   */
+  @NonNull
+  public static TypeCodec<Instant> timestampAt(@NonNull ZoneId timeZone) {
```
adutra: I think we forgot this method, so I added it now.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodecs.java
```diff
@@ -163,9 +374,114 @@
...
+
+  /**
+   * Returns a codec that maps the given Java type to JSON strings, using the provided Jackson
+   * {@link JsonMapper} to perform serialization and deserialization of JSON objects.
+   *
+   * @see <a href="http://wiki.fasterxml.com/JacksonHome">Jackson JSON Library</a>
+   */
+  @NonNull
+  public static <T> TypeCodec<T> json(
+      @NonNull GenericType<T> javaType, @NonNull JsonMapper jsonMapper) {
```
adutra: The exposure of `JsonMapper` causes strange revapi exceptions. We might want to revisit this. Besides, Jackson (non-shaded) is included in the core driver because of the Insights client, which may be removed in the future.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/extras/OptionalCodec.java
```diff
@@ -0,0 +1,70 @@
...
+  public OptionalCodec(@NonNull TypeCodec<T> innerCodec) {
+    super(
+        Objects.requireNonNull(innerCodec, "innerCodec must not be null"),
+        GenericType.optionalOf(innerCodec.getJavaType()));
+  }
+
+  @Override
+  public boolean accepts(@NonNull Object value) {
+    Objects.requireNonNull(value);
+    if (value instanceof Optional) {
```
adutra, line 46: We agreed not to push the check too deep in `MappingCodec`, but I figured that for Optionals, it was easy enough to do dit so I did.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/extras/array/ByteArrayCodec.java
```diff
@@ -0,0 +1,95 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.nio.ByteBuffer;
+import java.util.Objects;
+import net.jcip.annotations.Immutable;
+
+/**
+ * A codec that maps the CQL type {@code list<byte>} to the Java type {@code byte[]}.
+ *
+ * <p>Note that this codec is not suitable for reading CQL blobs as byte arrays; you should use
+ * {@link SimpleBlobCodec} for that.
```
adutra: We have 3 different codecs dealing with bytes:

1. `BlobCodec` : `blob <-> ByteBuffer`
1. `SimpleBlobCodec` : `blob <-> byte[]`
1. `ByteArrayCodec` : `list<tinyint> <-> byte[]`

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/extras/json/JsonCodec.java
```diff
@@ -0,0 +1,182 @@
...
+ *   <groupId>com.fasterxml.jackson.core</groupId>
+ *   <artifactId>jackson-databind</artifactId>
+ *   <version>LATEST</version>
+ * </dependency>
+ * }</pre>
+ *
+ * @see <a href="http://wiki.fasterxml.com/JacksonHome">Jackson JSON Library</a>
+ * @param <T> The Java type that this codec serializes from and deserializes to, from JSON strings.
+ */
+public class JsonCodec<T> implements TypeCodec<T> {
```
adutra, line 53: I'm proposing one single `JsonCodec` using Jackson. I think this will satisfy 90% of the users. I therefore removed the `JacksonJsonCodec` classes from the query-builder and examples modules, and made these modules use this one instead.

The examples module has another Json codec using JSR-353. I left that codec there because 1) I think there is no real demand for that and 2) it drags a lot of new dependencies.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/extras/time/PersistentZonedTimestampCodec.java
```diff
@@ -0,0 +1,103 @@
...
+ * <p>Since Cassandra's <code>timestamp</code> type does not store any time zone, by using a <code>
+ * tuple&lt;timestamp,varchar&gt;</code> a timezone can be persisted in the <code>varchar
+ * </code> field of such tuples, and so when the value is deserialized the original timezone is
+ * preserved.
+ *
+ * <p>Note: if you want to retrieve CQL timestamps as {@link ZonedDateTime} instances but don't need
+ * to persist the time zone to the database, you should rather use {@link ZonedTimestampCodec}.
+ */
+@Immutable
+public class PersistentZonedTimestampCodec extends MappingCodec<TupleValue, ZonedDateTime> {
```
adutra, line 47: This is the "old" `ZonedTimestampCodec`, but that name is already taken. We used to have users using this codec before, see [JAVA-1952](https://datastax-oss.atlassian.net/browse/JAVA-1952).

adutra: So we have the following codecs now:

1. `TimestampCodec`: `timestamp <-> Instant` (uses pre-defined zone for parsing only)
1. `TimestampMillisCodec`: `timestamp <-> long`  (uses pre-defined zone for parsing only)
1. `ZonedTimestampCodec`: `timestamp <-> ZonedDateTime`  (uses pre-defined zone for conversion and parsing)
1. `LocalTimestampCodec`: `timestamp <-> LocalDateTime` (uses pre-defined zone for conversion and parsing)
1. `PersistentZonedTimestampCodec`: `tuple<timestamp,string> <-> ZonedDateTime` (zone is persisted in tuple)

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/extras/time/PersistentZonedTimestampCodec.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  @NonNull
+  @Override
+  public String format(@Nullable ZonedDateTime value) {
+    if (value == null) {
+      return "NULL";
+    }
+    // Use TIMESTAMP_UTC for a better-looking format
+    return "("
```
adutra, line 74: This allows the usage of `TypeCodecs.TIMESTAMP_UTC` instead of the default `TypeCodecs.TIMESTAMP`, which produces better-looking results such as `('2020-03-19T12:23:56.789Z','+02:00')` – instead of the awkward looking output e.g. if your system TZ is `America/New_York`: `('2020-03-19T08:23:56.789Z [America/New_York]','+02:00')`.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/extras/time/ZonedTimestampCodec.java
```diff
@@ -13,19 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.datastax.oss.driver.internal.core.type.codec;
+package com.datastax.oss.driver.internal.core.type.codec.extras.time;
```
adutra, line 16: I moved this codec to `com.datastax.oss.driver.internal.core.type.codec.extras.time` to align with other temporal codecs added in this PR.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/extras/time/ZonedTimestampCodec.java
```diff
@@ -34,17 +32,17 @@
...
+ * PersistentZonedTimestampCodec} instead.
  *
  * <p>This codec shares its logic with {@link TimestampCodec}. See the javadocs of this codec for
  * important remarks about implementation notes and accepted timestamp formats.
  *
  * @see TimestampCodec
  */
 @ThreadSafe
-public class ZonedTimestampCodec implements TypeCodec<ZonedDateTime> {
+public class ZonedTimestampCodec extends MappingCodec<Instant, ZonedDateTime> {
```
adutra, line 44: Also retrofitted to extend `MappingCodec`.

examples/src/main/java/com/datastax/oss/driver/examples/json/jsr/Jsr353JsonCodec.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.datastax.oss.driver.examples.json.codecs;
+package com.datastax.oss.driver.examples.json.jsr;
```
adutra, line 16: Moved this codec here since he was alone in its old package and there was no reason why we wouldn't include it here along with the examples that use it.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/extras/array/BooleanArrayCodec.java
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * A codec that maps the CQL type {@code list<boolean>} to the Java type {@code boolean[]}.
+ *
+ * <p>Note that this codec is designed for performance and converts CQL lists <em>directly</em> to
+ * {@code boolean[]}, thus avoiding any unnecessary boxing and unboxing of Java primitive {@code
+ * boolean} values; it also instantiates arrays without the need for an intermediary Java {@code
+ * List} object.
+ */
+@Immutable
+public class BooleanArrayCodec extends AbstractPrimitiveArrayCodec<boolean[]> {
```
adutra: The legacy driver extras module had codecs for int[], long[], float[] and double[] only. For completeness, I'm also including codecs for boolean[], byte[] and short[].

core/revapi.json
```diff
@@ -4991,6 +4991,31 @@
...
+      },
+      {
+        "code": "java.class.nonPublicPartOfAPI",
+        "new": "class com.fasterxml.jackson.databind.type.TypeParser.MyTokenizer",
+        "justification": "Caused by the exposure of JsonMapper as a parameter in TypeCodecs.json()"
+      },
+      {
+        "code": "java.class.nonPublicPartOfAPI",
+        "new": "class com.fasterxml.jackson.databind.util.PrimitiveArrayBuilder.Node<T extends java.lang.Object>",
+        "justification": "Caused by the exposure of JsonMapper as a parameter in TypeCodecs.json()"
```
olim7t: 👍 I verified that other `TypeCodecs` methods and constant can still be referenced even if Jackson is excluded from the classpath.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodecs.java
```diff
@@ -78,39 +98,171 @@
...
+   * @see #TIMESTAMP_UTC
+   * @see #timestampAt(ZoneId)
+   */
   public static final TypeCodec<Instant> TIMESTAMP = new TimestampCodec();
 
+  /**
+   * A codec that handles Apache Cassandra(R)'s timestamp type and maps it to Java's {@link
+   * Instant}, using {@link ZoneOffset#UTC} as its source of time zone information. If you need a
+   * different time zone, consider other constants in this class, or call {@link
+   * #timestampAt(ZoneId)} instead.
```
olim7t: Maybe we should emphasize that the time zone is only used for parsing and formatting. It's a bit surprising at first because neither `Instant` nor `timestamp` are zoned.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodecs.java
```diff
@@ -78,39 +98,171 @@
...
+   * <p>If you are looking for a codec mapping the CQL type {@code blob} to the Java type {@link
+   * ByteBuffer}, you should use {@link #BLOB} instead.
+   *
+   * <p>If you are looking for a codec mapping the CQL type {@code list<tinyint} to the Java type
+   * {@code byte[]}, you should use {@link #BYTE_ARRAY} instead.
+   *
+   * @see #BLOB
+   * @see #BYTE_ARRAY
+   */
+  public static final TypeCodec<byte[]> BLOB_SIMPLE = new SimpleBlobCodec();
```
olim7t: Not a fan of the name, although I'm not sure I have a better proposal. Maybe `BLOB_TO_ARRAY`?

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/extras/array/BooleanArrayCodec.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.Objects;
+import net.jcip.annotations.Immutable;
+
+/**
+ * A codec that maps the CQL type {@code list<boolean>} to the Java type {@code boolean[]}.
+ *
+ * <p>Note that this codec is designed for performance and converts CQL lists <em>directly</em> to
+ * {@code boolean[]}, thus avoiding any unnecessary boxing and unboxing of Java primitive {@code
+ * boolean} values; it also instantiates arrays without the need for an intermediary Java {@code
+ * List} object.
```
olim7t: Those comments would be more useful on the constants and methods in `TypeCodecs`.

The codec implementations themselves are internal, the end user does not usually reference them directly.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodecs.java
```diff
@@ -163,9 +374,114 @@
...
+   * Returns a codec mapping Java Enums to CQL ints, according to their {@linkplain Enum#ordinal()
+   * ordinals}.
+   *
+   * @see #enumNamesOf(Class)
+   */
+  @NonNull
+  public static <EnumT extends Enum<EnumT>> TypeCodec<EnumT> enumOrdinalsOf(
+      @NonNull Class<EnumT> enumClass) {
+    return new EnumOrdinalCodec<>(enumClass);
+  }
```
olim7t: I wonder if we should provide this out of the box. Using ordinals is a terrible idea, let's not encourage it.

olim7t (changes requested):
This looks good, only a few minor remarks.

I wonder if we shouldn't add a separate `ExtraTypeCodecs` class. Having everything directly on `TypeCodecs` is a bit overwhelming, especially for time types. On the other hand, it makes it a bit harder to find a given codec.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodecs.java
olim7t: Yeah I'm not a big fan of piggybacking on that dependency, indeed we don't know how long it will be there. There's also a chance thats someone uses an incompatible version.

How about moving JSON to the examples?

adutra: I find it a bit disappointing because moving this to the examples module would require users to copy-paste code. I would like to see something more straightforward. I'm tempted to re-introduce the "extras" module for everything related to Json.

adutra: After giving this some more thought, I take back my previous statement. I think it is still better for users if everything is available from driver-core. If we ever remove Insights, we could switch Jackson to an optional dependency.

adutra:
Sorry for the last-minute commit, 2d59830: I just realized that I've been using `JsonMapper` everywhere, but it was introduced in Jackson 2.10, which means that we wound't be compatible with any prior version of Jackson. I am now using the more traditional `ObjectMapper` instead. It doesn't change anything for the codec itself, but has a much broader compatibility range.

adutra:
Can we get this in 4.6.0?

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodecs.java
adutra: But we already provided this in driver 3.x, do you think we should remove this functionality then?

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodecs.java
adutra: Agreed. I think I will also change the names of fields and classes dealing with list <-> arrays.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodecs.java
adutra: I'm in favor of leaving it. Ordinals can save space on disk, if the enum never changes I don't see why it's so horrible.

adutra:
Rebased on 4.x.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/TypeCodecs.java
olim7t: Sorry to rekindle the debate, I was about to concede on the grounds of 3.x compatibility, but this really bothers me:

> if the enum never changes I don't see why it's so horrible

That's a very big if. The enum always changes. It just takes one careless developer inserting a new constant in the middle, or reordering the constants to their liking, and you start inserting different codes, that overlap with the old ones. Your database is trashed, and it's not recoverable (beyond restoring a backup, or maybe a script that would track the writetime, but imagine the mess).

It's far from evident: the enum might come from a different layer of the application, or even a different project, where people are not even aware that it's being inserted in a database.

It's almost impossible to enforce: you can add a comment ("the order of these constants matters"), but you don't know if it will be read. Tools like Clirr or Revapi will warn you, but I'm not sure how widespread their use is.

So all in all it's a very dangerous feature, I think we are doing our users a disservice by providing it. It was indeed in 3.x, but I think it was a mistake to include it then.
If people want to save bytes, the better solution is to have an explicit `fromCode` / `toCode` conversion layer, and write your own codec, which should be relatively straightforward with `MappingCodec`.

adutra: >  It just takes one careless developer inserting a new constant in the middle, or reordering the constants to their liking

... or changing an enum constant name. And your database is screwed just the same. In this case we shouldn't provide any codec for enums. And then we would be back to careful users complaining that our driver is hard to use because they need to create their codecs themselves.

There is demand for storing enums in the database. Someone in DataStax Community recently was even [asking for an enum type in Cassandra itself](https://community.datastax.com/questions/5844/enum-datatype-for-cassandra-column.html). One of the reasons invoked is storage efficiency, and enum ordinals would be better at that.

I would like us to provide both codecs, just like in driver 3. We are not responsible for careless developers, we have warnings in the docs explaining the caveats of both approaches, so we're covered imo.

olim7t: > or changing an enum constant name. And your database is screwed just the same

No, it's much more obvious: client code will stop compiling; and even after that, the codec will fail to decode the old name. Also, you would likely pick an entirely new name, not swap existing constants.

Let's say you store card suits by name, but one of the names is wrong: `enum Suit { HEARTS, DIAMONDS, SPADES, CLOVERS }`. Someone fixes `CLOVERS` => `CLUBS`. Assuming no one picked up on the compile errors, and your tests didn't catch it either, the worst that can happen is that you only find out in production. This will fail hard and fast as soon as the codec tries to decode `CLOVERS`. You revert the upgrade, and you also need an annoying but doable migration of any corrupted writes: find any row with `CLUBS` and replace it with `CLOVERS`.

On the other hand, you store suits by ordinal, and have `enum Suit { HEARTS, DIAMONDS, SPADES, CLUBS }`. But someone decides that enum constants should be sorted in alphabetical order: `enum Suit { CLUBS, DIAMONDS, HEARTS, SPADES }`. The code still compiles. Tests won't catch it if they use enum constants for both reads & writes. Production code won't fail fast, it still happily reads and writes data. It's more likely that a human will detect the issue, which might take significantly longer. When all is said and done, you're left to reconcile rows where 0 means hearts, 2 means spades and 3 means clubs, with rows where 0 means clubs, 2 means hearts and 3 means spades. This is an order of magnitude worse.

> We are not responsible for careless developers

No, but we shouldn't give users a ticking timebomb in a nice giftwrap either.

olim7t: @csplinter any opinion on this?
I'm not dead-set on not having an enum ordinal codec out of the box, but there are strong arguments against it.

olim7t (approved):
We decided to leave the enum ordinal codec in. I'll add a warning in the javadocs.

Rebasing and merging this now.

olim7t:
OK this should be good to squash and merge now.

Not doing it right now because I did a number of changes on the manual: I tried to make the time codecs stuff more palatable by splitting it in smaller sections and adding examples. Let me know how that looks.

I'll merge on Monday.

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/ExtraTypeCodecs.java
```diff
@@ -0,0 +1,482 @@
...
+
+  /**
+   * A codec that maps CQL type {@code list<boolean>} to Java's {@code boolean[]}.
+   *
+   * <p>Note that this codec is designed for performance and converts CQL lists <em>directly</em> to
+   * {@code boolean[]}, thus avoiding any unnecessary boxing and unboxing of Java primitive {@code
+   * boolean} values; it also instantiates arrays without the need for an intermediary Java {@code
+   * List} object.
+   */
+  public static final TypeCodec<boolean[]> BOOLEAN_ARRAY = new BooleanListToArrayCodec();
```
adutra: @olim7t I'm overall 👍 on the new class but why did you change the name of this field and not the others below?

manual/core/custom_codecs/README.md
```diff
@@ -39,8 +39,220 @@ Sometimes you might want to use different mappings, for example:
...
+#### Mapping CQL lists to Java arrays
+
+By default, the driver maps CQL `list` to Java's [java.util.List]. If you prefer to deal with 
+arrays, the driver offers the following codecs:
+
+1. For primitive types:
+
+    | Codec | CQL type | Java type |
+    |---|---|---|
+    | [ExtraTypeCodecs.BOOLEAN_LIST_TO_ARRAY] | `list<boolean>` | `boolean[]` |
```
adutra, line 65: The field name does not correspond.

manual/core/custom_codecs/README.md
```diff
@@ -39,8 +39,220 @@ Sometimes you might want to use different mappings, for example:
 
 Custom codecs allow you to define those dedicated mappings, and plug them into your session.
 
+### Using alternative codecs provided by the driver
```
adutra, line 42: Awesome rewrite, kudos!

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/ExtraTypeCodecs.java
olim7t: 🤔  I thought I had renamed it to match the others (hence the correct name in the manual). Will do now.

PR #1414, opened 2020-03-20
JAVA-2105: Transient Replication Support

GregBestland:
I think the solution proposed here is correct. I don't believe we want to consider transient replica's full replica's for the purposes of token aware load balancing. 

I believe at this time only NetworkTopologyStrategy and Simple Strategy are supported for transient replication

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationFactor.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationFactor.java
@@ -15,26 +15,22 @@ package com.datastax.oss.driver.internal.core.metadata.token;
  * limitations under the License.
  */
 import com.datastax.oss.driver.shaded.guava.common.base.Preconditions;
-import com.datastax.oss.driver.shaded.guava.common.base.Splitter;
-import java.util.List;
 import java.util.Objects;
 
 // This class is a subset of server version at org.apache.cassandra.locator.ReplicationFactor
 public class ReplicationFactor {
   private final int allReplicas;
   private final int fullReplicas;
+  private final int transientReplicas;
 
-  private ReplicationFactor(int allReplicas, int transientReplicas) {
+  public ReplicationFactor(int allReplicas, int transientReplicas) {
     this.allReplicas = allReplicas;
+    this.transientReplicas = transientReplicas;
     this.fullReplicas = allReplicas - transientReplicas;
   }
 
-  public static ReplicationFactor withTransient(int allReplicas, int transientReplicas) {
-    return new ReplicationFactor(allReplicas, transientReplicas);
-  }
-
-  public static ReplicationFactor fullOnly(int allReplicas) {
-    return new ReplicationFactor(allReplicas, 0);
+  public ReplicationFactor(int allReplicas) {
+    this(allReplicas, 0);
   }
 
   public int fullReplicas() {
@@ -42,7 +38,7 @@ public class ReplicationFactor {
   }
 
   public int transientReplicas() {
-    return allReplicas - fullReplicas;
+    return transientReplicas;
   }
 
   public boolean hasTransientReplicas() {
@@ -51,28 +47,32 @@ public class ReplicationFactor {
 
   public static ReplicationFactor fromString(String s) {
     if (s.contains("/")) {
-      List<String> parts = Splitter.on('/').splitToList(s);
+
+      int slash = s.indexOf('/');
+      String allPart = s.substring(0, slash);
+      String transientPart = s.substring(slash + 1);
       Preconditions.checkArgument(
-          parts.size() == 2, "Replication factor format is <replicas> or <replicas>/<transient>");
-      return new ReplicationFactor(Integer.parseInt(parts.get(0)), Integer.parseInt(parts.get(1)));
+          allPart != null && transientPart != null,
+          "Replication factor format is <replicas> or <replicas>/<transient>");
+      return new ReplicationFactor(Integer.parseInt(allPart), Integer.parseInt(transientPart));
     } else {
       return new ReplicationFactor(Integer.parseInt(s), 0);
     }
   }
 
-  public String toParseableString() {
+  @Override
+  public String toString() {
     return allReplicas + (hasTransientReplicas() ? "/" + transientReplicas() : "");
   }
 
   @Override
-  public String toString() {
-    return "rf(" + toParseableString() + ')';
-  }
-
-  @Override
   public boolean equals(Object o) {
-    if (this == o) return true;
-    if (!(o instanceof ReplicationFactor)) return false;
+    if (this == o) {
+      return true;
+    }
+    if (!(o instanceof ReplicationFactor)) {
+      return false;
+    }
     ReplicationFactor that = (ReplicationFactor) o;
     return allReplicas == that.allReplicas && fullReplicas == that.fullReplicas;
   }
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationFactor.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationFactor.java
@@ -29,12 +29,12 @@ public class ReplicationFactor {
     this.fullReplicas = allReplicas - transientReplicas;
   }
 
-  public static ReplicationFactor withTransient(int totalReplicas, int transientReplicas) {
-    return new ReplicationFactor(totalReplicas, transientReplicas);
+  public static ReplicationFactor withTransient(int allReplicas, int transientReplicas) {
+    return new ReplicationFactor(allReplicas, transientReplicas);
   }
 
-  public static ReplicationFactor fullOnly(int totalReplicas) {
-    return new ReplicationFactor(totalReplicas, 0);
+  public static ReplicationFactor fullOnly(int allReplicas) {
+    return new ReplicationFactor(allReplicas, 0);
   }
 
   public int fullReplicas() {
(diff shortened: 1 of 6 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationFactor.java
```diff
@@ -0,0 +1,84 @@
...
+import com.datastax.oss.driver.shaded.guava.common.base.Splitter;
+import java.util.List;
+import java.util.Objects;
+
+// This class is a subset of server version at org.apache.cassandra.locator.ReplicationFactor
+public class ReplicationFactor {
+  private final int allReplicas;
+  private final int fullReplicas;
+
+  private ReplicationFactor(int allReplicas, int transientReplicas) {
```
tomekl007: nit: `allReplicas` and `totalReplicas` names are used for the same thing, maybe we should normalize it?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationFactor.java
```diff
@@ -0,0 +1,84 @@
...
+
+  public int transientReplicas() {
+    return allReplicas - fullReplicas;
+  }
+
+  public boolean hasTransientReplicas() {
+    return allReplicas != fullReplicas;
+  }
+
+  public static ReplicationFactor fromString(String s) {
```
tomekl007, line 48: I think it will be beneficial to create a unit test for this method

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationFactor.java
```diff
@@ -0,0 +1,84 @@
...
+      Preconditions.checkArgument(
+          parts.size() == 2, "Replication factor format is <replicas> or <replicas>/<transient>");
+      return new ReplicationFactor(Integer.valueOf(parts.get(0)), Integer.valueOf(parts.get(1)));
+    } else {
+      return new ReplicationFactor(Integer.valueOf(s), 0);
+    }
+  }
+
+  public String toParseableString() {
+    return String.valueOf(allReplicas) + (hasTransientReplicas() ? "/" + transientReplicas() : "");
```
tomekl007: same here (unit test)

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/SimpleReplicationStrategy.java
```diff
@@ -63,9 +67,9 @@ private static Token getTokenWrapping(int i, List<Token> ring) {
     return ring.get(i % ring.size());
   }
 
-  private static int extractReplicationFactor(Map<String, String> replicationConfig) {
+  private static ReplicationFactor extractReplicationFactor(Map<String, String> replicationConfig) {
     String factorString = replicationConfig.get("replication_factor");
     Preconditions.checkNotNull(factorString, "Missing replication factor in " + replicationConfig);
-    return Integer.parseInt(factorString);
+    return ReplicationFactor.fromString(factorString);
```
tomekl007, line 69: to sum up this change, the transient replicas are only loaded and parsed, but currently, they are ignored by the `ReplicationStrategy`, right?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/SimpleReplicationStrategy.java
```diff
@@ -30,22 +30,26 @@
...
+  private final ReplicationFactor replicationFactor;
 
   SimpleReplicationStrategy(Map<String, String> replicationConfig) {
     this(extractReplicationFactor(replicationConfig));
   }
 
   @VisibleForTesting
   SimpleReplicationStrategy(int replicationFactor) {
+    this(ReplicationFactor.fullOnly(replicationFactor));
+  }
```
olim7t: Rather than preserving this constructor, we can have unit tests create a replication factor explicitly and call `SimpleReplicationStrategy(ReplicationFactor)`.

Then maybe we can parameterize them to assert that a non-zero `transientReplicas` produces the same results.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationFactor.java
```diff
@@ -0,0 +1,84 @@
...
+  public boolean hasTransientReplicas() {
+    return allReplicas != fullReplicas;
+  }
+
+  public static ReplicationFactor fromString(String s) {
+    if (s.contains("/")) {
+      List<String> parts = Splitter.on('/').splitToList(s);
+      Preconditions.checkArgument(
+          parts.size() == 2, "Replication factor format is <replicas> or <replicas>/<transient>");
+      return new ReplicationFactor(Integer.valueOf(parts.get(0)), Integer.valueOf(parts.get(1)));
```
olim7t: Nit: use `Integer.parseInt` to avoid the IDEA warnings (redundant boxing)

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationFactor.java
olim7t: Nit: unnecessary `String.valueOf` call (IDEA warning)

olim7t (changes requested):
LGTM. A few comments but they're only minor improvements. Also please add a changelog entry.

> I don't believe we want to consider transient replica's full replica's for the purposes of token aware load balancing.

Agreed.

And I don't think we need to expose transient replicas in the metadata (token map), at least for now. I'm not sure this will ever be useful for a client; this is an experimental feature anyway, so let's wait and see if the need arises.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/SimpleReplicationStrategy.java
olim7t: Yes. They won't be used for token-aware routing, nor surfaced in token metadata. I think that's the right thing to do (see my other comment).

GregBestland: That is correct see documentation here for more information on how this is working under the covers.

http://cassandra.apache.org/blog/2018/12/03/introducing-transient-replication.html

GregBestland:
I think this should address all the concerns raised here. Let me know if further changes are required. I will squash and merge upon approval.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/SimpleReplicationStrategy.java
```diff
@@ -30,22 +29,21 @@
...
 class SimpleReplicationStrategy implements ReplicationStrategy {
 
-  private final int replicationFactor;
+  private final ReplicationFactor replicationFactor;
 
   SimpleReplicationStrategy(Map<String, String> replicationConfig) {
     this(extractReplicationFactor(replicationConfig));
   }
 
-  @VisibleForTesting
```
olim7t, line 39: Nit: you can keep this annotation, the constructor is still only exposed for tests

olim7t (approved):
👍 LGTM

tomekl007 (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationFactor.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.Objects;
+
+// This class is a subset of server version at org.apache.cassandra.locator.ReplicationFactor
+public class ReplicationFactor {
+  private final int allReplicas;
+  private final int fullReplicas;
+
+  private ReplicationFactor(int allReplicas, int transientReplicas) {
+    this.allReplicas = allReplicas;
+    this.fullReplicas = allReplicas - transientReplicas;
```
adutra, line 29: Why don't you store `transientReplicas` as a field? You are computing `fullReplicas = allReplicas - transientReplicas` here, but in `transientReplicas()` we need the inverse computation: `allReplicas - fullReplicas`.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationFactor.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  @Override
+  public String toString() {
+    return "rf(" + toParseableString() + ')';
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) return true;
```
adutra: missing braces

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationFactor.java
```diff
@@ -0,0 +1,84 @@
...
+public class ReplicationFactor {
+  private final int allReplicas;
+  private final int fullReplicas;
+
+  private ReplicationFactor(int allReplicas, int transientReplicas) {
+    this.allReplicas = allReplicas;
+    this.fullReplicas = allReplicas - transientReplicas;
+  }
+
+  public static ReplicationFactor withTransient(int allReplicas, int transientReplicas) {
```
adutra: Not convinced by the utility of these factory methods vs usage of the raw constructor (+ maybe an overloading constructor that assigns 0 transient replicas).

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationFactor.java
```diff
@@ -0,0 +1,84 @@
...
+    return allReplicas - fullReplicas;
+  }
+
+  public boolean hasTransientReplicas() {
+    return allReplicas != fullReplicas;
+  }
+
+  public static ReplicationFactor fromString(String s) {
+    if (s.contains("/")) {
+      List<String> parts = Splitter.on('/').splitToList(s);
```
adutra: `splitToList` is nice but marked as beta, we might want to use a good ol' 

```
int slash = s.indexOf('/');
String s1 = s.substring(0, slash);
String s2 = s.substring(slash + 1);
```

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationFactor.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  public static ReplicationFactor fromString(String s) {
+    if (s.contains("/")) {
+
+      int slash = s.indexOf('/');
+      String allPart = s.substring(0, slash);
+      String transientPart = s.substring(slash + 1);
+      Preconditions.checkArgument(
+          allPart != null && transientPart != null,
```
adutra, line 55: The result of a substring cannot be null, you should be checking `allPart.isEmpty()` instead – although the check isn't really necessary imo, the `parseInt` method below would fail if the strings contain anything but valid integers.

adutra:
This was merged a bit too soon.

PR #1415, opened 2020-03-27
JAVA-2711: Fix handling of UDT keys in the mapper

olim7t:


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
```diff
@@ -333,12 +337,17 @@ public static void setValue(
...
+              enclosingClass.addGenericTypeConstant(type.asRawTypeName()));
+      if (useNullSavingStrategy) {
+        methodBuilder.nextControlFlow(
+            "else if ($L == $T.$L)",
+            NULL_SAVING_STRATEGY,
+            NullSavingStrategy.class,
+            NullSavingStrategy.SET_TO_NULL);
+      } else {
+        methodBuilder.nextControlFlow("else");
+      }
```
tomekl007, line 350: nit: This code is the same as:
https://github.com/datastax/java-driver/pull/1415/files#diff-8992e8a0bae64eda9aaf4ff5012fa95fR300-R309
maybe we could extract method?

integration-tests/src/test/java/com/datastax/oss/driver/mapper/UdtKeyIT.java
```diff
@@ -0,0 +1,196 @@
...
+    // Given
+    List<Key> key = ImmutableList.of(new Key(1), new Key(2));
+    dao.saveMulti(new MultiKeyRecord(key, 42));
+
+    // When
+    MultiKeyRecord record = dao.findMultiByKey(key);
+
+    // Then
+    assertThat(record.getValue()).isEqualTo(42);
+  }
```
tomekl007, line 97: I don't see tests that examine the `useNullSavingStrategy`, should we add those?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/util/generation/GeneratedCodePatterns.java
olim7t: It's just 3 lines of code, it doesn't buy much and I think it would make the code harder to read.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/UdtKeyIT.java
olim7t: Technically it's not related to this ticket, it would have to be in `NullSavingStrategyIT`. But why not.

olim7t: Actually, `NestedUdtIT` already covers those cases.

tomekl007 (approved):

adutra (approved):

PR #1417, opened 2020-03-30
JAVA-2054: Add now_in_seconds to protocol v5 query messages

olim7t:


bom/pom.xml
```diff
@@ -61,7 +61,7 @@
       <dependency>
         <groupId>com.datastax.oss</groupId>
         <artifactId>native-protocol</artifactId>
-        <version>1.4.9</version>
+        <version>1.4.10-SNAPSHOT</version>
```
olim7t, line 64: This requires datastax/native-protocol#34.

core/src/main/java/com/datastax/oss/driver/api/core/cql/Statement.java
```diff
@@ -348,6 +348,20 @@ default SelfT setRoutingKey(@NonNull ByteBuffer... newRoutingKeyComponents) {
...
+   *
+   * <p>This method's default implementation returns the statement unchanged. The only reason it
+   * exists is to preserve binary compatibility. Internally, the driver overrides it to record the
+   * new value.
+   */
+  @NonNull
+  @SuppressWarnings("unchecked")
+  default SelfT setNowInSeconds(int nowInSeconds) {
+    return (SelfT) this;
+  }
```
olim7t, line 394: Since this feature is mainly intended for tests, I think default implementations are a good compromise to avoid breaking the API.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -142,6 +142,12 @@ public static Message toMessage(
     ProtocolVersion protocolVersion = context.getProtocolVersion();
     ProtocolVersionRegistry protocolVersionRegistry = context.getProtocolVersionRegistry();
     CqlIdentifier keyspace = statement.getKeyspace();
+    int nowInSeconds = statement.getNowInSeconds();
+    if (nowInSeconds != Integer.MIN_VALUE
```
tomekl007: I would extract `Integer.MIN_VALUE` and name it as a constant, for example:
`NOW_IN_SECONDS_UNSET` because it is used in a lot of places

integration-tests/src/test/java/com/datastax/oss/driver/core/cql/NowInSecondsIT.java
```diff
@@ -0,0 +1,122 @@
...
+@CassandraRequirement(min = "4.0")
+public class NowInSecondsIT {
+
+  private static final CcmRule CCM_RULE = CcmRule.getInstance();
+
+  private static final SessionRule<CqlSession> SESSION_RULE = buildSessionRule();
+
+  private static SessionRule<CqlSession> buildSessionRule() {
+    // Reminder to revisit the test when V5 comes out of beta: remove the custom config loader and
+    // inline this method.
```
tomekl007, line 53: Do you mean when the `V5` will be set as default? (so we will not need to do:
`.withString(DefaultDriverOption.PROTOCOL_VERSION, "V5")`)

olim7t: I mean when `v5.isBeta()` becomes false.

tomekl007 (approved):

core/src/main/java/com/datastax/oss/driver/api/core/cql/Statement.java
```diff
@@ -67,6 +68,20 @@
   GenericType<CompletionStage<AsyncResultSet>> ASYNC =
       new GenericType<CompletionStage<AsyncResultSet>>() {};
 
+  /**
+   * A special value for {@link #getQueryTimestamp()} that means "no value".
+   *
+   * <p>It is equals to {@link Long#MIN_VALUE}.
```
adutra: Nit: "it is equal to". Same for the field below.

adutra (dismissed):

PR #1420, opened 2020-04-01
JAVA-2717 & JAVA-2730: Add Cassandra 4.0 to 3.x driver builds and add support for new table options

emerkle826:
This PR introduces a Jenkinsfile to have the `3.x` branch build on the new CloudBees server, and adds Cassandra 4.0 to the 3.x driver build matrix. The build for this branch can be found here:
https://jenkins-drivers.build.dsinternal.org/job/drivers/job/java/job/oss/job/java2717-jenkinsfile/

This PR also addresses JAVA-2730 and adds support for Table Options `read_repair` and `additional_write_policy`


driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
```diff
@@ -1089,7 +1093,12 @@ public void run() {
     }
 
     private static boolean isThriftSupported(VersionNumber cassandraVersion) {
-      return cassandraVersion.compareTo(VersionNumber.parse("4.0")) < 0;
+      // Thrift is removed from some pre-release 4.x versions, make the comparison work for those
+      return cassandraVersion.compareTo(VersionNumber.parse("4.0-a")) < 0;
```
adutra: See my comment in the other PR, you should be able to use `VersionNumber.nextStable()` for that.

adutra (changes requested):

driver-core/src/test/java/com/datastax/driver/core/TestUtils.java
```diff
@@ -986,7 +986,7 @@ public static Level setLogLevel(String logger, Level newLevel) {
    * @param ccm cluster to check against
    */
   public static void compactStorageSupportCheck(CCMAccess ccm) {
-    if (ccm.getCassandraVersion().compareTo(VersionNumber.parse("4.0.0")) >= 0) {
+    if (ccm.getCassandraVersion().compareTo(VersionNumber.parse("4.0-a")) >= 0) {
```
emerkle826: again, change this to use `nextStable()`

driver-core/src/main/java/com/datastax/driver/core/AbstractTableMetadata.java
```diff
@@ -258,14 +258,19 @@ protected StringBuilder appendOptions(StringBuilder sb, boolean formatted) {
...
-    and(sb, formatted)
-        .append("dclocal_read_repair_chance = ")
-        .append(options.getLocalReadRepairChance());
+    if (cassandraVersion.getMajor() < 4)
+      sb.append("read_repair_chance = ").append(options.getReadRepairChance());
+    else sb.append("read_repair = 'BLOCKING'");
+    if (cassandraVersion.getMajor() < 4)
+      and(sb, formatted)
+          .append("dclocal_read_repair_chance = ")
+          .append(options.getLocalReadRepairChance());
```
emerkle826: I have created https://datastax-oss.atlassian.net/browse/JAVA-2730 for addressing this.

changelog/README.md
```diff
@@ -21,6 +21,7 @@
 - [documentation] JAVA-2504: Migrate Cloud "getting started" page to driver manual.
 - [improvement] JAVA-2516: Enable hostname validation with Cloud
 - [bug] JAVA-2515: NEW_NODE and REMOVED_NODE events should trigger ADDED and REMOVED.
+- [improvement[ JAVA-2730: Add support for Cassandra® 4.0 table options
```
emerkle826: typo

driver-core/src/main/java/com/datastax/driver/core/TableOptionsMetadata.java
```diff
@@ -23,9 +23,10 @@
 public class TableOptionsMetadata {
 
   private static final String COMMENT = "comment";
-  private static final String READ_REPAIR = "read_repair_chance";
-  private static final String DCLOCAL_READ_REPAIR = "dclocal_read_repair_chance";
-  private static final String LOCAL_READ_REPAIR = "local_read_repair_chance";
+  private static final String READ_REPAIR_CHANCE = "read_repair_chance";
```
emerkle826, line 26: I renamed a few private variables:
`READ_REPAIR` to `READ_REPAIR_CHANCE`
`DCLOCAL_READ_REPAIR` to `DCLOCAL_READ_REPAIR_CHANCE`
`LOCAL_READ_REPAIR` to `LOCAL_READ_REPAIR_CHANCE`
so that I could introduce `READ_REPAIR`.

Luckily the public access methods were already `getXYZchance()`

driver-core/src/main/java/com/datastax/driver/core/TableOptionsMetadata.java
```diff
@@ -206,6 +218,19 @@ public String getComment() {
    * @return the read repair chance set for table (in [0.0, 1.0]).
    */
   public double getReadRepairChance() {
+    return readRepairChance;
+  }
+
+  /**
+   * Returns the read_repair option for this table. <b>NOTE:</b> this is a Cassandra® 4.0 and newer
+   * option (described here<a
```
adutra: Nit: spurious word "here" before &lt;a> tag.

driver-core/src/main/java/com/datastax/driver/core/TableOptionsMetadata.java
```diff
@@ -206,6 +218,19 @@ public String getComment() {
...
+   * href="http://cassandra.apache.org/doc/latest/operating/read_repair.html">here:
+   * http://cassandra.apache.org/doc/latest/operating/read_repair.html</a>). Possible values are
```
adutra, line 228: The link is included in the link's text, is that on purpose?

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
```diff
@@ -1089,7 +1097,16 @@ public void run() {
     }
 
     private static boolean isThriftSupported(VersionNumber cassandraVersion) {
-      return cassandraVersion.compareTo(VersionNumber.parse("4.0")) < 0;
+      // Thrift is removed from some pre-release 4.x versions, make the comparison work for those
+      return cassandraVersion.nextStable().compareTo(VersionNumber.parse("4.0")) < 0;
+    }
+
+    private static boolean isMaterializedViewsSupported(VersionNumber cassandraVersion) {
+      return cassandraVersion.nextStable().compareTo(VersionNumber.parse("4.0")) >= 0;
```
adutra, line 1105: MVs are available for C* 3+, but indeed they are disabled by default in C* 4+.

If the intent of this method is to detect whether or not we need to include `enable_materialized_views` in C* options, then its name is misleading. I suggest `isMaterializedViewsDisabledByDefault`.

See JAVA-2183.

driver-core/src/test/java/com/datastax/driver/core/MissingRpcAddressTest.java
```diff
@@ -67,10 +67,18 @@ private void deleteNode2RpcAddressFromNode1() throws Exception {
                         Policies.defaultLoadBalancingPolicy(), Lists.newArrayList(firstHost)))
                 .build());
     Session session = cluster.connect();
-    String deleteStmt =
-        String.format(
-            "DELETE rpc_address FROM system.peers WHERE peer = '%s'",
-            ccm().addressOfNode(2).getHostName());
+    String deleteStmt;
+    if (ccm().getCassandraVersion().nextStable().compareTo(VersionNumber.parse("4.0")) < 0) {
+      deleteStmt =
```
adutra: Shouldn't we instead always delete from system.peers, and additionally, also delete from system.peers_v2 if this table exists?

driver-tests/osgi/src/test/java/com/datastax/driver/osgi/BundleOptions.java
```diff
@@ -165,7 +165,16 @@ public static CompositeOption defaultOptions() {
                 mavenBundle("ch.qos.logback", "logback-classic", getVersion("logback.version")),
                 mavenBundle("ch.qos.logback", "logback-core", getVersion("logback.version")),
                 mavenBundle("io.dropwizard.metrics", "metrics-core", getVersion("metrics.version")),
-                mavenBundle("org.yaml", "snakeyaml", getVersion("snakeyaml.version")),
+                mavenBundle(
```
adutra, line 168: 1. So snakeyaml was being included for nothing?
2. Why do we need jackson in all OSGi configs now? Isn't this going to be a problem for the shaded jar?

adutra (changes requested):

driver-core/src/main/java/com/datastax/driver/core/TableOptionsMetadata.java
emerkle826: Actually, I intended "here" to not be part of the link, but either way I've got it twice which was not the intent.

driver-core/src/main/java/com/datastax/driver/core/TableOptionsMetadata.java
emerkle826: Yes, my intent was to link to the Cassandra documentation, and have the link show the full address.

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
emerkle826: Good point, this makes more sense.

driver-core/src/test/java/com/datastax/driver/core/MissingRpcAddressTest.java
emerkle826: Yes, I think this is correct. I missed that detail when focusing on fixing this for 4.0 as deleting from peers wasn't enough.

driver-tests/osgi/src/test/java/com/datastax/driver/osgi/BundleOptions.java
emerkle826: I'm not exactly sure about this, but the Cloud API support is where this changed. The commit [here](https://github.com/datastax/java-driver/commit/aa5d6d859a5490e21a5216edf5aef54d9f6cd7c7) introduced Jackson dependencies, but also added snakeyaml to the OSGi BundleOptions.

I'm guessing snakeyaml was added as an attempt to avoid adding Jackson, and put in the OSGi BundleOptions and subsequently forgotten. If I recall correctly, we tried snakeyaml for the config at first, but it was easier to just use Jackson. So maybe later, Jackson got added, but the OSGi stuff was left alone.

It seems the Jackson libraries are only added for testing though. I'm still digging in on the shaded jar to see....

adutra: OK but let's not spend an insane amount of time on OSGi tests, given that JAVA-2658 will change all of this radically.

olim7t (approved):
LGTM.

Note that the last fixup commit contains changes that applies to both 2717 and 2730, you might have to do some editing to squash things in the right place if you want to keep them separate.

Also, I find it a bit weird that "fix tests for C* 4.0 pre-releases" hardcodes some values in `AbstractTableMetadata`, we could avoid that if we move it last.

emerkle826:
> LGTM.
> 
> Note that the last fixup commit contains changes that applies to both 2717 and 2730, you might have to do some editing to squash things in the right place if you want to keep them separate.
> 
> Also, I find it a bit weird that "fix tests for C* 4.0 pre-releases" hardcodes some values in `AbstractTableMetadata`, we could avoid that if we move it last.

I'll re-work the commits so that there's only 2, one for 2717 and one for 2730, with all the review feedback in the appropriate commit.

adutra (approved):

PR #1421, opened 2020-04-03
JAVA-2718: Warn when the number of sessions exceeds a configurable threshold

olim7t:


core/src/main/resources/reference.conf
```diff
@@ -341,6 +341,20 @@ datastax-java-driver {
...
+  # sessions and/or do not close them correctly. The driver keeps track of the number of live
+  # sessions in a static variable; if it gets over this threshold, a warning will be logged for
+  # every new session.
+  #
+  # If the value is less than or equal to 0, the feature is disabled: no warning will be issued.
+  #
+  # Required: yes
+  # Modifiable at runtime: yes, the new value will be used for sessions created after the change.
+  # Overridable in a profile: no
+  advanced.session-leak.threshold = 4
```
olim7t, line 356: I think the default should be aggressive. Most clients probably use 1 anyway, so it's better to warn fast.

integration-tests/src/test/resources/application.conf
```diff
@@ -13,6 +13,7 @@ datastax-java-driver {
     graph.timeout = 10 seconds
   }
   advanced {
+    session-leak.threshold = 16
```
olim7t: I'm not sure if this is the right value for integration test, but I'll run the suite in CI and we can adjust accordingly.

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
```diff
@@ -517,7 +548,6 @@ private void onChildrenClosed(List<CompletionStage<Void>> childrenCloseStages) {
                 if (!f.isSuccess()) {
                   closeFuture.completeExceptionally(f.cause());
                 } else {
-                  LOG.debug("[{}] Shutdown complete", logPrefix);
```
tomekl007, line 525: why do you remove this log?

tomekl007 (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
olim7t: It's replaced by the one on Line 372 (which covers more cases btw, since it's implemented as a generic callback on the close future).

olim7t:
I ended up reverting the custom `session-leak.threshold` for integration tests. Whatever value I use, I never see any warning in the logs. I suspect that parallel test execution might not be working correctly. I don't want to go down a Maven rabbithole now, so I'll create a follow-up ticket.

PR #1422, opened 2020-04-03
JAVA-2662: Refactoring of Native + converting gettimeofday() to use jnr-posix

absurdfarce:
Rather than explicitly loading libc and using jnr-posix


core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -15,67 +15,32 @@
...
-import jnr.posix.POSIXFactory;
-import jnr.posix.util.DefaultPOSIXHandler;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /** A gateway to perform system calls. */
 public class Native {
 
-  private static final Logger LOG = LoggerFactory.getLogger(Native.class);
+  private static NativeImpl impl = new JnrNativeImpl();
```
absurdfarce: The key to the refactoring of this class: I'm laying the groundwork here for JAVA-2663 as well.  It's fairly easy to see how a GraalNativeImpl would fit in here.

core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrNativeImpl.java
```diff
@@ -0,0 +1,154 @@
...
+
+    Supplier<IllegalStateException> exceptionSupplier =
+        () -> {
+          throw new IllegalStateException(
+              "Native call not available. "
+                  + "Check isCurrentTimeMicrosAvailable() before calling this method.");
+        };
+
+    Timeval tv = this.posix.map(POSIX::allocateTimeval).orElseThrow(exceptionSupplier);
+    int rv = this.posix.map(p -> p.gettimeofday(tv)).orElseThrow(exceptionSupplier);
```
absurdfarce: I'm being conservative here.  It's hard to imagine a case in which we'd be able to allocate a new Timeval but somehow get an exception (as opposed to a non-zero return val) when we go to make the actual syscall... but you never know.

core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrNativeImpl.java
```diff
@@ -0,0 +1,154 @@
...
+        return Optional.empty();
+      }
+    } catch (Throwable t) {
+
+      LOG.debug("Error calling gettimeofday()", t);
+      return Optional.empty();
+    }
+
+    return Optional.of(posix);
+  }
```
absurdfarce: There's a _slight_ difference in behaviour from the old impl here.  We've moved both calls into the jnr-posix orbit but don't maintain discrete vals for the validation of each discrete call; we just validate the whole thing by confirming that we can call _both_ native functions.  Logically this means that we could find ourselves in a situation where getpid() succeeded but gettimeofday() didn't, and in that case we'd prevent users from subsequent calls to _either_ syscall when (in theory) we could allow them to call getpid() and just throw on gettimeofday() calls.

Note that gettimeofday() "failing" above means throwing an exception (i.e.a failure in the Java infrastructure) not just a non-zero return val (which indicates that the syscall was made and something went wrong within the kernel).

I kinda like this idea even in the abstract; if we can't validate that your POSIX instance can do _everything_ we expect it's in an untrustworthy state and we don't allow you to do anything.  More practically, it seems _extremely_ unlikely that we'd get a Java exception with one syscall but not another; the whole point of POSIX is that we expect syscalls with these names and params to be available.  If one is there and another isn't we've got some kind of very odd libc we shouldn't trust much anyway.

absurdfarce:
The one _major_ problem I see here is the lack of testing.  We didn't have much (any?) testing around Native as it stands and this PR doesn't add any.  I can appreciate the complexity there: should you test this on multiple platforms or just be okay if a test passes with our OpenStack images?

I've used a driver with this change locally and confirmed that the native calls seem to do what you'd expect them to do... but that's (a) manual and (b) hardly exhaustive.  I'm open to other ideas here.

tomekl007:
> The one _major_ problem I see here is the lack of testing. We didn't have much (any?) testing around Native as it stands and this PR doesn't add any. I can appreciate the complexity there: should you test this on multiple platforms or just be okay if a test passes with our OpenStack images?
> 
> I've used a driver with this change locally and confirmed that the native calls seem to do what you'd expect them to do... but that's (a) manual and (b) hardly exhaustive. I'm open to other ideas here.

I think that we need to think about what is the goal of this ticket and test it. Our new implementation that will not be using JNR should work without that library, so our test should assert that getPid and getTimestamp are working with JNR dependencies excluded. Maybe we could use OSGI tests for that.

The same for the existing code (that you moved to `JnrNativeImpl`). Maybe we should create two tests:
- one that tests behavior when JNR is on the classpath.
- 2nd that verifies that code still work when JNR is not on the classpath.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
tomekl007: How do you plan to parameterize it and inject different implementations? We need to move it higher in the call chain (here probably constructor injection) and pass it from somewhere? Maybe parameterized `DriverContext`?

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -15,67 +15,32 @@
  */
 package com.datastax.oss.driver.internal.core.os;
 
-import com.datastax.oss.driver.internal.core.util.Reflection;
-import java.lang.reflect.Method;
-import jnr.ffi.LibraryLoader;
 import jnr.ffi.Platform;
```
tomekl007: this class cannot have any dependency on `jnr.ffi`, otherwise quarkus native image build will fail. We need to remove this dependency.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: I'm envisioning something like the following:

```java
  private static NativeImpl impl =
          /* Hope is that Graal build-time analysis will detect this case and follow the right path
          * (i.e. exclude JnrNativeImpl from consideration in the build).  Obviously this needs to be
          * tested.
          *
          * Constant values here come from org.graalvm.nativeimage.ImageInfo and are guaranteed not to change
          * in future releases.  Use the string literals here in order to avoid pulling in an import of GraalVM
          * code in this class. */
          System.getProperty("org.graalvm.nativeimage.imagecode").equals("buildtime") ?
            new GraalNativeImpl() :
              new JnrNativeImpl();
```

Obviously that'll need to be tested in a real case before we call it good.  That work will be handled in JAVA-2663.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: Platform is a different case from the other jnr-ffi code.  It doesn't actually execute any native calls... it just builds up some state based on sysprops and exposes that.  I think there's a comment or two to this effect at https://github.com/absurdfarce/graal-cassandra-driver.

Regardless, we probably can expose this information in some other way.  I'll dig into the Platform code a bit to see what it's looking at to populate this data.  If we can pull the info from the same place and just include a default impl in the NativeImpl interface that would be awesome. :)

core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrNativeImpl.java
olim7t: > More practically, it seems extremely unlikely that we'd get a Java exception with one syscall but not another

💯 
I'm fine with the change.

core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrNativeImpl.java
```diff
@@ -0,0 +1,154 @@
...
+        () -> {
+          throw new IllegalStateException(
+              "Native call not available. "
+                  + "Check isCurrentTimeMicrosAvailable() before calling this method.");
+        };
+
+    Timeval tv = this.posix.map(POSIX::allocateTimeval).orElseThrow(exceptionSupplier);
+    int rv = this.posix.map(p -> p.gettimeofday(tv)).orElseThrow(exceptionSupplier);
+    if (rv != 0) {
+      throw new IllegalStateException("Expected 0 return valu from gettimeofday(), observed " + rv);
```
olim7t: Small typo:
```suggestion
      throw new IllegalStateException("Expected 0 return value from gettimeofday(), observed " + rv);
```

core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrNativeImpl.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+    return tv.sec() * 1000000 + tv.usec();
```
olim7t: nit:
```suggestion
    return tv.sec() * 1_000_000 + tv.usec();
```

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: I wound up folding the jnr.ffi.Platform code into Native and doing away with our dependence on Platform all together.  As mentioned above our usage of Platform is really just an eval of a sysprop, so now we just do all of that inline.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -84,11 +49,7 @@ public static int getProcessId() {
    * @return {@code true} if JNR {@link Platform} class is loaded.
    */
   public static boolean isPlatformAvailable() {
```
olim7t: Unfortunate naming here, this should be called `isCpuAvailable`.

We could take advantage of this ticket to rename it, it's internal code so it won't break the API.

absurdfarce:
I agree with you on the testing question @tomekl007, but on the plus side the current situation was better than I thought.  When I changed the API a bit as part of my most recent PR I broke an existing unit test for Native proper.  I've since revised that test a bit so now we have the following:

* NativeTest - tests the bits actually in Native, really just extracting CPU info
* JnrNativeImplTest - tests the jnr-based impl of NativeImpl

This _should_ be adequate for now.  I agree that we'll need something to test the Graal side of this once we get to implementing that but that test will be handled as part of JAVA-2663.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
olim7t: Suggestion: make `Native` the interface directly, in order to avoid the additional abstraction.
```java
public interface Native {
  static Native INSTANCE = new JnrNative();
  boolean isCurrentTimeMicrosAvailable();
  ...
}
public class JnrNative implements Native {
  @Override
  public boolean isCurrentTimeMicrosAvailable() { ... }
  ...
}
```
This will just require a small change to every call site: `Native.currentTimeMicros()` => `Native.INSTANCE.currentTimeMicros()`. But again this is internal code so we're allowed to do it, and it's cleaner.

core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrNativeImpl.java
```diff
@@ -0,0 +1,154 @@
...
+import jnr.posix.util.DefaultPOSIXHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class JnrNativeImpl implements NativeImpl {
+
+  private static final Logger LOG = LoggerFactory.getLogger(JnrNativeImpl.class);
+
+  private final Optional<POSIX> posix;
+  private final Optional<Platform> platform;
```
olim7t: We allow users to [exclude JNR dependencies manually](https://docs.datastax.com/en/developer/java-driver/4.5/manual/core/integration/#native-libraries). If those JARs are missing from the classpath, `Native` should still load correctly, even though all `isXxxAvailable()` methods will return false.

This doesn't work with this version, if I exclude JNR I get:
```
java.lang.NoClassDefFoundError: jnr/posix/POSIXHandler
	at com.datastax.oss.driver.internal.core.os.Native.<clinit>(Native.java:23)
```

I suggest we simply copy over the code from 4.x, with the `PosixLoader` and `PlatformLoader` inner classes. It's a bit less pretty, but it's tried and tested.

olim7t (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
olim7t: > Platform is a different case from the other jnr-ffi code. It doesn't actually execute any native calls...

Wow I never realized that. In that case, 👍 to remove it, it's not worth the extra dependency. We'll also need to update the [integration page](https://docs.datastax.com/en/developer/java-driver/4.5/manual/core/integration/#native-libraries) in the manual.

jnr-ffi is ASL so copying the code is fine. However I think it would be a good idea to reproduce their header in our code, along the lines of what I did in [CountingIterator](https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/util/CountingIterator.java#L40-L57).

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: I agree re: unfortunate naming and I'm happy to change it.

Oh wait, nevermind, that's right... this method actually disappears because we're loading CPU information directly rather than relying on Platform to do it. :man_facepalming:

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: :100: agree with the idea of maximizing attribution so I'll add the jnr-ffi license as requested.  I'm also updating our docs to clearly indicate that jnr-posix rather than jnr-ffi is the guiding force behind the underlying native calls.

Note, however, that we won't wind up removing jnr-ffi in it's entirety as it's required by jnr-posix.  But we can remove our declaration of it and just let jnr-posix pull in whatever it needs.  Apparently this has come up before (JAVA-1797) with Andy landing on a similar idea.

My only concern with that approach is making sure we're okay on the shaded JAR; I'll check that before I put all this in.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -15,191 +15,173 @@
...
-    return PlatformLoader.PLATFORM.getCPU().toString();
+    return CPU.toString();
   }
 
-  /**
-   * If jnr-ffi is not in the classpath at runtime, we'll fail to initialize the static fields
-   * below, but we still want {@link Native} to initialize successfully, so use an inner class.
+  /* The remainder of this class is copied from jnr.ffi.Platform in jnr-ffi version 2.1.10.
+   * We copy it manually here in order to avoid introducing an extra dependency merely for the sake of
+   * evaluating some system properties.
```
absurdfarce: In isolation this comment looks a bit odd.  We already have JnrNativeImpl which hides the dependency on jnr-posix... why not do something similar with jnr-ffi and avoid copying code?  This suggestion breaks down when we consider adding Graal support in JAVA-2663.  In that case we'll still want to provide this functionality for _both_ impls but ideally we wouldn't mix jnr-ffi and the Graal impl.  With this in mind the cleanest answer was to promote this functionality to the Native level and implement it by copyig code.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: In isolation I'd probably agree with you, but keep in mind we'll be adding a different NativeImpl for Graal in JAVA-2663.  With that in mind I'd argue it's cleaner to maintain Native as a top-level facade and swap the composed class (the NativeImpl implementation) based on Graal/not-Graal.

core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrNativeImpl.java
absurdfarce: Ah, that explains the NoClassDefFoundError catch blocks in the old code.  I thought that was handling something else but now it makes considerably more sense.

There's no need to copy in the old loader code; all we really need here is a loader which can handle exceptions coming off attempts to create instances.  I'd argue that's actually a cleaner way to represent what's occurring in this case; the old way (loading everything fine and then having NCDFE checks around the "isXAvailable" calls) is far less clear (as evidenced by my confusion above).

I'll put together a quick change to handle this... it shouldn't be too bad.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -34,172 +24,172 @@
...
+  private static class ImplLoader {
+
+    public NativeImpl load() {
+      try {
+        return new JnrNativeImpl();
+      } catch (Throwable t) {
+        LOG.info(
+            "Unable to load JNR native implementation.  This could be normal if JNR is excluded from the classpath",
+            t);
+        return new EmptyNativeImpl();
```
absurdfarce: Manual testing confirmed that (a) I get the error message above and (b) other driver functions appear to work fine (without native support) if JNR is excluded.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
olim7t: As far as adding another implementation, this doesn't change anything:
```java
public class GraalNative implements Native {
  ...
}
```
The `Native.impl` or `Native.INSTANCE` singleton is held in a static variable in both cases, so it shouldn't be any harder to initialize it based on some environment condition.

core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrNativeImpl.java
olim7t: OK, I saw your empty implementation approach, that works for me.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -34,172 +24,172 @@
...
+    AARCH64,
+
+    /**
+     * Unknown CPU architecture. A best effort will be made to infer architecture specific values
+     * such as address and long size.
+     */
+    UNKNOWN;
+  }
+
+  private static CPU determineCPU() {
```
tomekl007: This is the way you try to make It work with quarkus, right (when JNR is missing)?
Shouldn't we do a similar thing for `gettimeofday`?

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -34,172 +24,172 @@
...
+    String archString = System.getProperty("os.arch");
```
tomekl007: maybe we could extract this whole logic and provide a 3rd implementation of `AbstractNativeImpl`?
So we would have: `jnr`, `unknown` and `graal` implementations?

core/src/main/java/com/datastax/oss/driver/internal/core/os/AbstractNativeImpl.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.internal.core.os;
+
+import java.util.function.Supplier;
+
+public abstract class AbstractNativeImpl implements NativeImpl {
+
+  Supplier<IllegalStateException> gettimeofdaySupplier =
```
adutra: 1. These suppliers are only used in one of the child classes, maybe move them there?
2. Storing functional interfaces as lambdas is precisely one of the new error-prone rules, I wonder if these won't generate a compile error?
3. The error messages mention `isCurrentTimeMicrosAvailable` and `isGetProcessIdAvailable` but these methods are not declared in this class hierarchy.
4. If the errors go away I think this class could be removed altogether.

core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrNativeImpl.java
```diff
@@ -0,0 +1,103 @@
...
+import jnr.posix.Timeval;
+import jnr.posix.util.DefaultPOSIXHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class JnrNativeImpl extends AbstractNativeImpl {
+
+  private static final Logger LOG = LoggerFactory.getLogger(JnrNativeImpl.class);
+
+  private final Optional<POSIX> posix;
```
adutra: I observe that you strived to give a functional look to this class, but I'm not convinced the result looks better than plain old Java.

1. Storing `Optional` as a field: my IDE complains, and since `Optional` is not serializable, I would rather avoid this situation, thinking that projects like Spark would love to see the entire driver serializable :-) 
2. The functional style is probably a matter of taste, but in the present case we must admit it's more verbose than imperative style, e.g. (also applying my suggestion to change `NativeImpl` API):

```
  @Nullable private final POSIX posix;

  public JnrNativeImpl() {
    this.posix = loadPosix();
  }

  private static POSIX loadPosix() {
    POSIX posix;
    try {
      posix = POSIXFactory.getPOSIX(new DefaultPOSIXHandler(), true);
    } catch (Throwable t) {
      LOG.debug("Error loading POSIX", t);
      return null;
    }
    try {
      posix.getpid();
    } catch (Throwable t) {
      LOG.debug("Error calling getpid()", t);
      return null;
    }
    try {
      gettimeofdayInternal(posix);
    } catch (Throwable t) {
      LOG.debug("Error calling gettimeofday()", t);
      return null;
    }
    return posix;
  }

  @Override
  public OptionalLong gettimeofday() {
    if (posix == null) {
      return OptionalLong.empty();
    }
    return OptionalLong.of(gettimeofdayInternal(posix));
  }

  private static long gettimeofdayInternal(@NonNull POSIX posix) {
    Timeval tv = posix.allocateTimeval();
    int rv = posix.gettimeofday(tv);
    if (rv != 0) {
      throw new IllegalStateException(
          "Expected 0 return value from gettimeofday(), observed " + rv);
    }
    return tv.sec() * 1_000_000 + tv.usec();
  }

  @Override
  public OptionalInt getpid() {
    if (posix == null) {
      return OptionalInt.empty();
    }
    return OptionalInt.of(posix.getpid());
  }
```

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -34,172 +24,172 @@
...
+        LOG.info(
+            "Unable to load JNR native implementation.  This could be normal if JNR is excluded from the classpath",
+            t);
+        return new EmptyNativeImpl();
+      }
+    }
+  }
+
+  /* Copied from equivalent op in jnr.ffi.Platform.  We have to have this here as it has to be defined
+   * before it's (multiple) uses in determineCpu() */
```
adutra: Nit: "before its (multiple)"

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -34,172 +24,172 @@
...
+  private static final Locale LOCALE = java.util.Locale.ENGLISH;
```
adutra: Nit: Can we use regular imports or is there a requirement to use FQCNs here?

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -34,172 +24,172 @@
...
   private static final Logger LOG = LoggerFactory.getLogger(Native.class);
 
+  private static class ImplLoader {
+
+    public NativeImpl load() {
+      try {
+        return new JnrNativeImpl();
+      } catch (Throwable t) {
+        LOG.info(
+            "Unable to load JNR native implementation.  This could be normal if JNR is excluded from the classpath",
```
adutra: Nit: two spaces after the dot.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -34,172 +24,172 @@
...
+     * Unknown CPU architecture. A best effort will be made to infer architecture specific values
+     * such as address and long size.
+     */
+    UNKNOWN;
+  }
+
+  private static CPU determineCPU() {
+    String archString = System.getProperty("os.arch");
+    assert (archString != null);
+    assert (archString.toUpperCase(LOCALE) != null);
```
adutra: My IDE reports this assertion and the next one (lines 153 and 154) as useless because of the assertion line 152.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
adutra: Ok but I'm slightly worried about maintainability. How frequently should we expect to get our hands dirty in this part of the code? 

> In that case we'll still want to provide this functionality for both impls

Why? Can't we just return UNKNOWN?

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -34,172 +24,172 @@
...
+    assert (archString.toLowerCase(LOCALE) != null);
+    if (equalsIgnoreCase("x86", archString)
+        || equalsIgnoreCase("i386", archString)
+        || equalsIgnoreCase("i86pc", archString)
+        || equalsIgnoreCase("i686", archString)) {
+      return CPU.I386;
```
adutra: All return statements raise compiler warnings: 

> Static member 'com.datastax.oss.driver.internal.core.os.Native.CPU.I386' accessed via instance reference 

This is because the field name and its type are both `CPU`.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -34,172 +24,172 @@
...
+    for (CPU cpu : CPU.values()) {
+      if (equalsIgnoreCase(cpu.name(), archString)) {
+        return cpu;
+      }
+    }
+
+    return CPU.UNKNOWN;
+  }
+
+  private static boolean equalsIgnoreCase(String s1, String s2) {
```
adutra: Is this required? Seems useless.

core/src/main/java/com/datastax/oss/driver/internal/core/os/NativeImpl.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.internal.core.os;
+
+public interface NativeImpl {
```
adutra: It's a bit funny to see an interface named `Impl` but I get why.

core/src/main/java/com/datastax/oss/driver/internal/core/os/AbstractNativeImpl.java
adutra: To fix item 3 above I would suggest to simplify the `NativeImpl` interface to just:

```
  OptionalLong gettimeofday();
  OptionalInt getpid();
```

Then you would deal of absent optionals inside `Native` class and throw the errors from there, where they more naturally belong imo.

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
olim7t: The enum should be called `Cpu` in accordance to our [naming conventions](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case), so that will indirectly address this problem.

We should also rename `getCPU` to `getCpu`.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: Not exactly.  This logic is copied here to replace the old use of Platform.  That impl just parses some system properties and computes some values for them; it doesn't actually execute any native code at all.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: Eventually we will get to a Graal impl of NativeImpl, but we'll also need to support getCPU() in that case.  That's why we copied the relevant bits of Platform in here and put it in an abstract parent class... since it's functionality will be common across NativeImpl instances.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: Bah, that's a consequence of copying the Platform code in... will fix.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: Ah, the eternal war about how many spaces to put after a period. :)  Regardless, the exception text uses one so I'll change it here.

core/src/main/java/com/datastax/oss/driver/internal/core/os/NativeImpl.java
olim7t: We can refactor to use regular interface inheritance, see [this comment](https://github.com/datastax/java-driver/pull/1422#discussion_r404418703).

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: :man_facepalming: Those were added when debugging a weird runtime error and I forgot to take them back out after I fixed the error.  Will remove.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: The Cpu name across the board does seem more sensible... I'll update this as well.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: It's part of what came along in the copy of the Platform code.  I'm _assuming_ they want to handle non-English locales and still get valid comparisons... but that's just speculation on my part.

I'm happy to just switch the tests above to use String.equalsIgnoreCase() if desired.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: I suppose we... could just return UNKNOWN, but I'm not sure why we'd take that path.  The functionality to determine the return val of getCpu() isn't based on JNR; it's a general manipulation of standard system properties.  Why _wouldn't_ we support that (and support it in the same way) for all NativeImpls?

core/src/main/java/com/datastax/oss/driver/internal/core/os/AbstractNativeImpl.java
absurdfarce: AbstractNativeImpl was a fairly late addition to the process, and it's entire reason for existence is to ensure a common exception experience across all concrete instances of NativeImpl.  I have no problem removing it if we can find an alternate way to address it's behaviours.

I also very much like the idea of removing the "is available" methods and just having getX() return an optional type, but note the consequence here.  We can no longer distinguish between a case in which the concrete type couldn't complete it's setup (i.e. when JNR wasn't present or failed to initialize) vs. when a specific function call failed.  In each case we'll return an Optional.empty() and Native will have to do... something.  We could chuck the IllegalStateException in that case, but if we do that the text will have to be tweaked (since it doesn't necessarily indicate that the native call isn't available).

A reasonable (to me) answer here would be to log the specific error in JnrNativeImpl and then throw the IllegalStateException at the Native level saying something like "native call did not succeed, consult logs for specific errors".  An Either type would also be great here but let's not hope for wild and crazy things. :)

core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrNativeImpl.java
absurdfarce: @adutra I'll speak to the functional style point shortly, but I'm curious on the first point... what specifically is your IDE complaining about re: storing an Optional type as a field?  This is... pretty common.  Is it a question of naming or complaints about the types?

Re: serializable classes... the concern there in a Spark context has to do with DTOs (and their equivalents) being serializable.  There's no need for every class in the driver to be serializable, and there's no call (that I've heard of) to make Native serializable.  The old version certainly was not so it's not clear to me why we'd need to change it here.

olim7t: No strong feeling on functional or not (I find that optional often falls short in Java, but that's entirely personal opinion).

With regard to serializability, I don't think this will be an issue since the only instance of this class is a static field in a utility class.

core/src/main/java/com/datastax/oss/driver/internal/core/os/AbstractNativeImpl.java
absurdfarce: I took a stab at what this might look like in https://github.com/datastax/java-driver/pull/1425.  Take a look and see if this is preferred... I can definitely see some benefit there.

core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrNativeImpl.java
absurdfarce: Re: the functional question... I readily agree this is largely a matter of style.  That said, my implementation included an attempt to be more explicit about what was going on, specifically around the loading vs. validation steps.  If we collapse those operations (as you do in your implementation @adutra) I'd change my version a fair amount as well.

It's also worth noting that you're still doing several checks along the lines of "if this val is null return some an empty option, otherwise return a non-empty option".  This indicates that you aren't fully leveraging the notion of the optional type.  I also note that you have several repeated try/catch and blocks which follow a very similar pattern but wind up contributing to code cruft.  The more functional/optional-friendly version allows us to filter those out.  My take on this (see https://github.com/datastax/java-driver/blob/52573a54ccc896c912e67640cd0679af062624e2/core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrNativeImpl.java) is (in my opinion) quite a bit more concise than the verbose Java example above.

Again, though, I readily agree that this question is largely a matter of taste/style.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -34,172 +24,169 @@
...
+    }
+  }
+
+  /* Copied from equivalent op in jnr.ffi.Platform.  We have to have this here as it has to be defined
+   * before its (multiple) uses in determineCpu() */
+  private static final Locale LOCALE = Locale.ENGLISH;
+
+  private static final NativeImpl IMPL = new ImplLoader().load();
+
+  @SuppressWarnings("VariableNameSameAsType")
```
adutra: Do we still need this annotation?

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -34,172 +24,169 @@
...
-   * @return {@code true} if JNR {@link Platform} class is loaded.
+   * @return the current processor architecture.
    */
-  public static boolean isPlatformAvailable() {
-    try {
-      return PlatformLoader.PLATFORM != null;
-    } catch (NoClassDefFoundError e) {
-      return false;
-    }
+  public static String getCpu() {
```
adutra, line 72: Since this is class is becoming a shell that delegates all of its operations to a `NativeImpl` instance, wouldn't it be a better design to also include `getCpu()` in `NativeImpl`?

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
adutra: Ok, fair enough.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
adutra: Or better yet: since determining the CPU does not depend on any native call, just on system properties. Shouldn't we then move all CPU stuff to its own class?

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
adutra: Ok to keep it.

core/src/main/java/com/datastax/oss/driver/internal/core/os/AbstractNativeImpl.java
absurdfarce: This has now been merged into this feature branch

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: No, we shouldn't anymore... good catch!

core/src/main/java/com/datastax/oss/driver/internal/core/os/NativeImpl.java
absurdfarce: I am going to change the name to something that makes it's role in this operation a bit clearer.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: Since we're preserving Native as the main entry point into this functionality I'd argue for keeping the functionality here.  Given it's nature this functionality won't change in the various NativeImpl/Libc implementations so I'd argue against pushing it down to that level.

Your point about moving it to it's own class is interesting, though.  I'd be fine moving this functionality to a discrete class which is then used by Native.getCpu() to implement the necessary functionality.  I'll take a quick stab at that and see what it looks like.

absurdfarce: Done; all this logic has now been moved to the CpuInfo class

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: This approach is isomorphic onto the static methods of Native in use now.  I'd argue this change doesn't really buy you much; you remove a layer of abstraction but (hopefully) the rename of NativeImpl to Libc makes the relationship between the two layers much clearer anyway.  Given that and a desire to minimize further churn on this PR I'd argue for keeping things as they are for now.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Libc.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package com.datastax.oss.driver.internal.core.os;
+
+import java.util.Optional;
+
+public interface Libc {
+
+  /* Maintained to allow Native.isXAvailable() functionality without trying to make a native call if
+   * the underlying support _is_ available. */
+  public boolean available();
```
olim7t: Nit: we don't specify the `public` modifier for interfaces, it's implied.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -34,172 +23,53 @@
...
+      } catch (Throwable t) {
+        LOG.info(
+            "Unable to load JNR native implementation. This could be normal if JNR is excluded from the classpath",
+            t);
+        return new EmptyLibc();
+      }
+    }
+  }
+
+  private static final Libc IMPL = new ImplLoader().load();
```
olim7t: We can rename to `LIBC` and `LibcLoader` now that we have a better name for the interface.

manual/core/integration/README.md
```diff
@@ -315,10 +315,6 @@ are not available on your platform, you can exclude the following dependencies:
   <artifactId>java-driver-core</artifactId>
   <version>${driver.version}</version>
   <exclusions>
-    <exclusion>
-      <groupId>com.github.jnr</groupId>
-      <artifactId>jnr-ffi</artifactId>
-    </exclusion>
```
olim7t, line 321: 👍 

Nit: the text above mentions "libraries" / "dependencies", it should be singular now.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Libc.java
absurdfarce: I've always liked specifying it; I find it easier to have the access specification at hand rather than remembering whether you're looking at an interface or not.  But I'm fine with removing it if that's preferred.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: Agreed

manual/core/integration/README.md
absurdfarce: Good catch, I'll fix the references above as well.

olim7t (approved):

absurdfarce:
Thanks @olim7t .  @adutra are you good with the current state of things?

adutra (approved):

adutra:
>  @adutra are you good with the current state of things?

Yup, I just approved the PR, Thanks!

PR #1423, opened 2020-04-06
JAVA-2706: Add now_in_seconds to protocol v5 query messages

olim7t:


driver-core/src/main/java/com/datastax/driver/core/Requests.java
```diff
@@ -250,27 +254,33 @@ public String toString() {
...
+      for (QueryFlag flag : values()) {
+        if ((flags & flag.mask) != 0) set.add(flag);
       }
       return set;
     }
 
     static void serialize(EnumSet<QueryFlag> flags, ByteBuf dest, ProtocolVersion version) {
       int i = 0;
-      for (QueryFlag flag : flags) i |= 1 << flag.ordinal();
+      for (QueryFlag flag : flags) i |= flag.mask;
```
adutra, line 283: Much better 👍

adutra (approved):

PR #1424, opened 2020-04-07
JAVA-2657, JAVA-2658, JAVA-2301: OSGi improvements

adutra:
To reviewers: we discussed the idea of getting rid of OSGi tests altogether and replace them by an example application.

It turns out we already have examples: https://github.com/datastax/java-driver-examples-osgi . I do plan on updating this repository later to use the new driver. However I struggled to find a way to get rid of Pax Exam while still being able to at least run a simple integration test. There are simply no alternatives :-(

So I'm back to the original plan of making our OSGi tests more reliable and debuggable. I'm finally proposing a PR for these long-awaited OSGi improvements:

There are a few "preparatory" commits:

1. 572a5d1 Exclude GraalVM from OSGi bundle manifests.
2. 3ed1fea Use driver's default class loader when user-supplied loader fails: this is motivated by my own findings where depending on the OSGi container, the user-supplied CL may or may not be able to load a built-in driver class. I gave up on researching and adopted this more flexible approach.
3. 1505287: Isolate mapper reactive capabilities in separate packages: improves the OSGi manifest by making Reactive Streams an optional dependency. Unfortunately that's not so easy as Reactive Streams classes must be in separate packages.
4. eabd6e5: Consider Simulacron and Commons Exec as optional dependencies : improvements to test-infra. It removes the dependency to Netty and considers Simulacron and commons-exec as optional.

And 2 main commits:

1. dbf62f5 JAVA-2657: Ability to provide a class loader to load configuration resources
2. d95d91b JAVA-2658 and JAVA-2301: Refactor OSGi tests and introduce OSGi tests for the mapper.

You can review each commit separately.


adutra:
Rebased on top of 4.x and updated the test dependencies to include Awaitility.

tomekl007:
There are some cloudIT failures on this branch, see: 
https://jenkins-drivers.build.dsinternal.org/job/drivers/job/java/job/oss/job/java2658/4/#showFailuresLink

adutra:
> There are some cloudIT failures on this branch, see:
> https://jenkins-drivers.build.dsinternal.org/job/drivers/job/java/job/oss/job/java2658/4/#showFailuresLink

All unrelated, and the UpdateIT failed test is unrelated too.

core/src/main/java/com/datastax/oss/driver/internal/core/util/Reflection.java
```diff
@@ -59,9 +59,19 @@
...
       LOG.trace("Successfully loaded {}", className);
       return clazz;
-    } catch (ClassNotFoundException | LinkageError | SecurityException e) {
-      LOG.debug(String.format("Could not load %s: %s", className, e), e);
-      return null;
+    } catch (LinkageError | Exception e) {
+      // Note: only ClassNotFoundException, LinkageError and SecurityException
+      // are declared to be thrown; however some class loaders (Apache Felix)
+      // may throw other checked exceptions, which cannot be caught directly
+      // because that would cause a compilation failure.
```
olim7t, line 66: I think some of the commit message would be useful here as well to indicate that this is intended for OSGi, it might not occur to readers to do a `git blame`, or the commit could get buried under unrelated changes (refactoring, reformating, etc).

core/src/main/java/com/datastax/oss/driver/internal/core/util/Reflection.java
```diff
@@ -59,9 +59,19 @@
...
+      LOG.debug(
+          String.format("Could not load %s with loader %s: %s", className, classLoader, e), e);
+      if (classLoader == null) {
+        return null;
+      } else {
+        // try with the driver's default class loader
+        return loadClass(null, className);
```
olim7t, line 77: Trying to think of a scenario where this could have unintended consequences:
* the user provided a classloader while building the session
* they specified policy class A, that can't be found in their classloader
* the driver's default classloader also contains a class A, but that's not the one we want to load

That seems pretty convoluted, I think we're safe.

core-shaded/pom.xml
```diff
@@ -316,15 +316,18 @@
                   not required at runtime.
                   -->!net.jcip.annotations.*, !edu.umd.cs.findbugs.annotations.*,
                   <!--
-                  3) Use resolution:=optional whenever possible in order to make it possible to use
+                  3) Don't include GraalVM.
+                  -->!org.graalvm.*,
```
olim7t: 👍  this one is easy

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
```diff
@@ -76,6 +76,17 @@ public DaoDeleteMethodGenerator(
...
+  public boolean requiresReactive() {
+    // Validate the return type:
+    DaoReturnType returnType =
+        parseAndValidateReturnType(getSupportedReturnTypes(), Delete.class.getSimpleName());
+    if (returnType == null) {
+      return false;
+    }
+    return returnType.requiresReactive();
+  }
+
```
olim7t, line 89: This feels a bit redundant since `generate` does the same thing, approximately at the same time. Maybe we could add a method on `DaoImplementationSharedCode`, this is how method generators usually communicate information back to the parent DAO class generator.
That's not high priority though: the duplicated section is short, it's compile time so a few more CPU cycles won't make a big difference, and everything is internal so we can revisit it later.

changelog/README.md
```diff
@@ -4,6 +4,7 @@
 
 ### 4.7.0 (in progress)
 
+- [bug] JAVA-2657: Add ability to specify the class loader to use for application-specific classpath resources
```
olim7t, line 9: I approved this when it was a standalone PR (#1410), I imagine the code hasn't changed much.

osgi-tests/README.md
```diff
@@ -0,0 +1,41 @@
...
+    mvn clean verify
+    
+Note that all other driver modules must have been previously compiled, that is,
+their respective `target/classes` directory must be up-to-date and contain
+not only the class files, but also an up-to-date OSGi manifest.
+
+If that is not the case you should first run on the parent module's root 
+directory:
+
+    mvn clean package -DskipTests
```
olim7t: I had to use `install` instead of `package` before I could get a successful run. Not sure why.

olim7t (approved):
LGTM

osgi-tests/README.md
adutra: You are right, these instructions seem a bit unreliable. I changed them to something more robust. Note that I spotted a small bug related to skipping IT tests, see 199e6ba6f on the 4.x branch.

olim7t:
Force-pushed to rebase on latest 4.x.
Letting CI run because `clean verify` still wasn't working in my local workspace.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGenerator.java
adutra: Created [JAVA-2807](https://datastax-oss.atlassian.net/browse/JAVA-2807) to revist that later.

adutra:
The IT tests weren't even ran because there were test failures before :-( I merged #1453 and kicked-off another build, hopefully this time we'll know.

adutra:
JAVA-2803 introduced a regression, there is a new dependency `org.graalvm.nativeimage` which exports a new package `com.oracle.svm.core.annotate`. I'm preparing a fix now.

adutra:
The CI build passes now. @olim7t leaving open just in case you'd want to have a final look. The only commit I had to amend to fix the regression is https://github.com/datastax/java-driver/pull/1424/commits/590ac020c59286cdee04cb35c05ee7e7b3d788ca.

core-shaded/pom.xml
```diff
@@ -316,17 +316,20 @@
...
-                  4) Don't import packages imported by shaded classes, if they are not used by the
+                  5) Don't import packages imported by shaded classes, if they are not used by the
                   driver bundle.
-                  -->!com.google.protobuf.*, !com.jcraft.jzlib.*, !com.ning.compress.*, !lzma.sdk.*, !net.jpountz.xxhash.*, !org.bouncycastle.*, !org.conscrypt.*, !org.apache.commons.logging.*, !org.apache.log4j.*, !org.apache.logging.log4j.*, !org.eclipse.jetty.*, !org.jboss.marshalling.*, !sun.misc.*, !sun.security.*, !com.oracle.svm.core.annotate.*,!com.barchart.udt.*, !com.fasterxml.aalto.*, !com.sun.nio.sctp.*, !gnu.io.*, !org.xml.sax.*, !org.w3c.dom.*, !reactor.blockhound.*, *
+                  -->!com.google.protobuf.*, !com.jcraft.jzlib.*, !com.ning.compress.*, !lzma.sdk.*, !net.jpountz.xxhash.*, !org.bouncycastle.*, !org.conscrypt.*, !org.apache.commons.logging.*, !org.apache.log4j.*, !org.apache.logging.log4j.*, !org.eclipse.jetty.*, !org.jboss.marshalling.*, !sun.misc.*, !sun.security.*, !com.barchart.udt.*, !com.fasterxml.aalto.*, !com.sun.nio.sctp.*, !gnu.io.*, !org.xml.sax.*, !org.w3c.dom.*, !reactor.blockhound.*, *
```
adutra, line 332: For some reason the shaded jar was already excluding the new package, but I harmonized the exclusions in both drivers to just `!com.oracle.svm.*`.

PR #1425, opened 2020-04-08
An experiment: simplifying the NativeImpl interface

absurdfarce:
…ion throws to Native (where they belong)


core/src/main/java/com/datastax/oss/driver/internal/core/os/NativeImpl.java
```diff
@@ -15,13 +15,15 @@
...
-  public boolean gettimeofdayAvailable();
+public interface NativeImpl {
 
-  public long gettimeofday();
+  /* Maintained to allow Native.isXAvailable() functionality without trying to make a native call if
+   * the underlying support _is_ available. */
+  public boolean available();
 
-  public boolean getpidAvailable();
+  public Optional<Long> gettimeofday();
```
adutra, line 26: Why not `OptionalLong` and `OptionalInt`?

core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrNativeImpl.java
```diff
@@ -34,70 +35,54 @@ public JnrNativeImpl() {
...
+  private Optional<Long> gettimeofdayImpl(POSIX posix) {
 
-    Timeval tv = this.posix.map(POSIX::allocateTimeval).orElseThrow(gettimeofdaySupplier);
-    int rv = this.posix.map(p -> p.gettimeofday(tv)).orElseThrow(gettimeofdaySupplier);
+    Timeval tv = posix.allocateTimeval();
+    int rv = posix.gettimeofday(tv);
     if (rv != 0) {
-      throw new IllegalStateException(
-          "Expected 0 return value from gettimeofday(), observed " + rv);
+      LOG.info("Expected 0 return value from gettimeofday(), observed " + rv);
```
adutra: `Log.debug` maybe?

adutra (changes requested):
Looking good!

core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrNativeImpl.java
absurdfarce: Agreed, will change

core/src/main/java/com/datastax/oss/driver/internal/core/os/NativeImpl.java
absurdfarce: Because OptionalLong and OptionalInt do not subclass Optional; they are their own type hierarchy.  As a result the map/flatMap ops off Optional<POSIX> can't generate these types.

adutra: Oh you are flatmapping them down the line. Ok then.

absurdfarce:
@adutra Can you think of any reason this _shouldn't_ be merged into the JAVA-2662 feature branch?  I do think this is a stronger API and removing the abstract super-class makes me smile as well.

adutra:
> @adutra Can you think of any reason this _shouldn't_ be merged into the JAVA-2662 feature branch? I do think this is a stronger API and removing the abstract super-class makes me smile as well.

No, go ahead and merge it. And please feel empowered to just go ahead and do whatever you think is best, no need to wait for my input :-)

PR #1426, opened 2020-04-10
JAVA-2549: Deprecate ConditionChecker in favor of Awaitility

olim7t:
The defaults are different:
||Polling Interval|Timeout|
|---|---|---|
|ConditionChecker|500ms|60s|
|Awaitility|100ms|10s|

I've updated the calls accordingly, to keep the exact same behavior (except for 1 second waits, where I've kept the 100ms default).

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/core/metadata/NodeStateIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/core/metadata/NodeStateIT.java
@@ -17,6 +17,7 @@ package com.datastax.oss.driver.core.metadata;
 
 import static com.datastax.oss.driver.assertions.Assertions.assertThat;
 import static com.datastax.oss.driver.assertions.Assertions.fail;
+import static org.awaitility.Awaitility.await;
 import static org.mockito.Mockito.inOrder;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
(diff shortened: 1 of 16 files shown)


integration-tests/src/test/java/com/datastax/oss/driver/core/metadata/NodeStateIT.java
```diff
@@ -148,13 +148,13 @@ public void setup() {
             driverContext.getLoadBalancingPolicy(DriverExecutionProfile.DEFAULT_NAME);
 
     // Sanity check: the driver should have connected to simulacron
-    ConditionChecker.checkThat(
+    await()
+        .alias("Connections established")
```
tomekl007, line 152: do you know why some `await()` have an alias, and others don't?

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/utils/ConditionChecker.java
```diff
@@ -25,12 +25,21 @@
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.BooleanSupplier;
 
+/**
```
tomekl007, line 28: we could change the internal API of `ConditionChecker` to use the `await()` underneath instead of our own solution. With such a solution there will be no need for deprecation, and the callers API will not change. But I am not sure we should hide the `await()` API this way.

tomekl007 (approved):

integration-tests/src/test/java/com/datastax/oss/driver/core/metadata/NodeStateIT.java
olim7t: The alias is just an informational message to make it easier to understand the error. We've not been using it consistently in our tests.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/utils/ConditionChecker.java
olim7t: The goal is to get rid of `ConditionChecker` entirely. I would even have deleted it, but it's part of an "API" package and I want to be rigorous about backward compatibility (even though there's a good chance that nobody is using it outside of the driver).

Changing the internals is a good idea, I'll take a quick look at it.

olim7t: I'm not going to do it. It's doable, but not a trivial one-line change, we'd have to make sure the builder API still works. This is dead code now, so any time spent on it is a waste, even if it's just 10 minutes.

adutra (approved):

PR #1428, opened 2020-04-10
JAVA-2723: Enable parallelization of graph tests

olim7t:


integration-tests/src/test/java/com/datastax/dse/driver/api/core/graph/GraphTestSupport.java
```diff
@@ -16,32 +16,35 @@
...
       CustomCcmRule.builder()
           .withDseWorkloads("graph")
           .withDseConfiguration("graph.max_query_params", 32)
           .withDseConfiguration(
               "graph.gremlin_server.scriptEngines.gremlin-groovy.config.sandbox_enabled", "false");
 
-  /** CCM Rule builder for general Graph workload tests. */
-  public static final CustomCcmRule.Builder GRAPH_CCM_RULE_BUILDER =
-      CustomCcmRule.builder().withDseWorkloads("graph");
-
```
olim7t, line 36: I don't understand why this separate constant existed, I tried running all tests with the other one and it didn't seem to change anything. So unless there's some performance impact, I assume it's safe to run all tests with `sandbox_enabled=false`. That's what I did in `DefaultCcmBridgeBuilderCustomizer`.

integration-tests/src/test/java/com/datastax/dse/driver/api/core/graph/statement/GraphTraversalITBase.java
```diff
@@ -294,7 +294,7 @@ public void should_return_zero_results() {
    */
   @Test
   public void should_return_zero_results_graphson_2() {
-    Assumptions.assumeThat(isGraphBinary()).isFalse();
+    assumeThat(isGraphBinary()).isFalse();
```
olim7t, line 297: Side effect of adding the static import in the code that I changed.
I can revert that to avoid impacting unrelated lines.

integration-tests/src/test/java/com/datastax/dse/driver/api/core/graph/statement/GraphTraversalITBase.java
```diff
@@ -595,10 +595,8 @@ public void should_allow_use_of_dsl_graph_binary() throws Exception {
    */
   @Test
   public void should_return_correct_results_when_bulked() {
-    Assumptions.assumeThat(
-            ccmRule().getCcmBridge().getDseVersion().get().compareTo(Version.parse("5.1.2")) > 0)
-        .isTrue();
-
+    assumeThat(ccmRule().getDseVersion())
+        .hasValueSatisfying(version -> assumeThat(version).isGreaterThan(Version.parse("5.1.2")));
```
olim7t, line 599: I had to change this in order to migrate the return type of `ccmRule()`: child classes now return `CcmRule`, which doesn't have a `getCcmBridge()` method.
And `getDseVersion()` is a simpler way to do it anyway.

integration-tests/src/test/java/com/datastax/dse/driver/api/core/graph/GraphPagingIT.java
```diff
@@ -57,7 +57,8 @@
 @RunWith(DataProviderRunner.class)
 public class GraphPagingIT {
 
-  private static final CustomCcmRule CCM_RULE = GraphTestSupport.GRAPH_CCM_RULE_BUILDER.build();
+  // Can't use the shared CCM rule because we pause/resume a node
+  private static final CustomCcmRule CCM_RULE = GraphTestSupport.CCM_BUILDER_WITH_GRAPH.build();
```
olim7t, line 61: Of all the graph tests impacted by this changeset, this is the only one that didn't migrate to the shared rule.

olim7t (commented):
Marking WIP as I want to see how this goes in Jenkins.

Also, I just noticed that I missed some other tests that create the rule themselves instead of using `GraphTestSupport`, for example `GraphAuthenticationIT`. I'll address them in a follow-up commit.

adutra:
Against DSE 6.7 we have a few graph failures, see [here](https://jenkins-drivers-old.build.dsinternal.org/view/java-driver/job/datastax.java-driver.java2723.commit/CASSANDRA_VERSION=dse-6.7,JABBA_VERSION=oracle_jdk8,OS_VERSION=ubuntu%2Fbionic64%2Fjava-driver/lastCompletedBuild/testReport/). But it's looking pretty good, and from what I saw many failures are just timeout problems.
However the situation is less clear to me against DSE 6.8, see [here](https://jenkins-drivers-old.build.dsinternal.org/view/java-driver/job/datastax.java-driver.java2723.nightly/CASSANDRA_VERSION=dse-6.8,JABBA_VERSION=oracle_jdk8,OS_VERSION=ubuntu%2Fbionic64%2Fjava-driver/2/#showFailuresLink). Lots of unrelated failures.

olim7t:
Yes, even locally I'm seeing a lot of tests time out that didn't on 4.x. I've tried to raise `graph.timeout` to 30 seconds to see if that makes a difference, but I'm now also seeing instances of:
```
com.datastax.oss.driver.api.core.servererrors.InvalidQueryException: Shared data commit failed due to concurrent modification: /
```
Which reminds me that graph doesn't do well with concurrent schema modifications... Maybe parallelizing those tests is not such a good idea after all.

adutra:
> Which reminds me that graph doesn't do well with concurrent schema modifications... 

This is surprising, what tests do concurrent schema modifications? Can we just exclude the ones that do so and keep the rest parallelizable?

PR #1430, opened 2020-04-16
JAVA-2700: Revisit execution profile parameters in mapper

olim7t:
This amends JAVA-2633 (5e1e5107c91c3546e5cc81a29cbccdbbaff0f82b).

See also [this review](https://github.com/datastax/java-driver/pull/1409#pullrequestreview-377027405).

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/cql/StatementBuilder.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/cql/StatementBuilder.java
@@ -91,6 +91,9 @@ public abstract class StatementBuilder<
   @NonNull
   public SelfT setExecutionProfileName(@Nullable String executionProfileName) {
     this.executionProfileName = executionProfileName;
+    if (executionProfileName != null) {
+      this.executionProfile = null;
+    }
     return self;
   }
 
@@ -98,7 +101,9 @@ public abstract class StatementBuilder<
   @NonNull
   public SelfT setExecutionProfile(@Nullable DriverExecutionProfile executionProfile) {
     this.executionProfile = executionProfile;
-    this.executionProfileName = null;
+    if (executionProfile != null) {
+      this.executionProfileName = null;
+    }
     return self;
   }
 
--- a/integration-tests/src/test/java/com/datastax/oss/driver/mapper/ProfileIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/mapper/ProfileIT.java
@@ -20,11 +20,12 @@ import static com.datastax.oss.simulacron.common.stubbing.PrimeDsl.query;
 import static com.datastax.oss.simulacron.common.stubbing.PrimeDsl.when;
 import static org.assertj.core.api.Assertions.assertThat;
 
+import com.datastax.oss.driver.api.core.ConsistencyLevel;
 import com.datastax.oss.driver.api.core.CqlSession;
-import com.datastax.oss.driver.api.core.DefaultConsistencyLevel;
 import com.datastax.oss.driver.api.core.config.DefaultDriverOption;
-import com.datastax.oss.driver.api.core.config.DriverConfigLoader;
 import com.datastax.oss.driver.api.core.config.DriverExecutionProfile;
+import com.datastax.oss.driver.api.core.cql.BoundStatementBuilder;
+import com.datastax.oss.driver.api.mapper.MapperBuilder;
 import com.datastax.oss.driver.api.mapper.annotations.Dao;
 import com.datastax.oss.driver.api.mapper.annotations.DaoFactory;
 import com.datastax.oss.driver.api.mapper.annotations.DaoProfile;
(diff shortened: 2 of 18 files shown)


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
```diff
@@ -146,13 +146,6 @@ protected void populateBuilderWithFunction(
     }
   }
 
-  protected void populateBuilderWithProfile(CodeBlock.Builder builder) {
-    builder.beginControlFlow("if(context.getExecutionProfileName() != null)");
-    builder.addStatement(
-        "boundStatementBuilder = boundStatementBuilder.setExecutionProfileName(context.getExecutionProfileName())");
-    builder.endControlFlow();
-  }
-
```
olim7t, line 155: One more thing I hadn't caught previously: it's easier to do this in `DaoBase`.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
```diff
@@ -60,6 +60,11 @@
 
   protected static CompletionStage<PreparedStatement> prepare(
       SimpleStatement statement, MapperContext context) {
+    if (context.getExecutionProfileName() != null) {
+      statement = statement.setExecutionProfileName(context.getExecutionProfileName());
+    } else if (context.getExecutionProfile() != null) {
+      statement = statement.setExecutionProfile(context.getExecutionProfile());
+    }
     return context.getSession().prepareAsync(statement);
```
tomekl007, line 68: Do we prevent users from setting both `profileName` and profile (`class`)?
If no, did we document the fact that `getExecutionProfileName` has priority over `getExecutionProfile`?

mapper-runtime/revapi.json
```diff
@@ -60,8 +60,13 @@
       {
         "code": "java.method.addedToInterface",
         "new": "method java.lang.String com.datastax.oss.driver.api.mapper.MapperContext::getExecutionProfileName()",
-        "justification": "JAVA-2633 Adding execution profile to mapper"
+        "justification": "JAVA-2633: Add execution profile argument to DAO factory method (accept API break -- it's unlikely that MapperContext will be implemented outside of the driver)"
```
tomekl007, line 63: Can't we use `customState` to propagate this? We would not need to break backward compatibility in such a case

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/DaoProfile.java
```diff
@@ -26,18 +27,16 @@
  *
  * <p>Example:
  *
- * <p>*
- *
  * <pre>
  *  * &#64;Mapper
  *  * public interface InventoryMapper {
  *  *   ProductDao productDao(@DaoTable String executionProfile);
```
adutra: Shouldn't this be `ProductDao productDao(@DaoProfile String executionProfile);`?

adutra (changes requested):

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
olim7t: It's enforced implicitly by the mapper processor: only one `@DaoProfile`-annotated parameter is allowed per DAO factory method.

I'll add a comment on the context methods to explain that.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/annotations/DaoProfile.java
olim7t: 👍 Good catch, will fix

mapper-runtime/revapi.json
olim7t: Already discussed [on the original PR](https://github.com/datastax/java-driver/pull/1409#discussion_r389558576).

To expand: `MapperContext` can't be implemented by the user in any useful way (they can't inject their implementation in mapper-generated code), so we deemed it acceptable to break the API.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DaoBase.java
olim7t: Now that I think of it, we could also allow injection of the profile via `MapperBuilder`, like we do with the default keyspace. I'll add that.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/ProfileIT.java
```diff
@@ -35,37 +36,55 @@
...
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
+import org.junit.rules.RuleChain;
+import org.junit.rules.TestRule;
 
 @Category(ParallelizableTests.class)
 public class ProfileIT {
```
olim7t, line 67: I've completely rearranged this test to cover all combinations of mapper-level, dao-level, and method-level (via the existing statement attributes mechanism).

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DefaultMapperContext.java
```diff
@@ -70,11 +56,6 @@ public DefaultMapperContext(
         NullAllowingImmutableMap.copyOf(customState));
   }
 
-  public DefaultMapperContext(
-      @NonNull CqlSession session, @NonNull Map<Object, Object> customState) {
-    this(session, null, customState);
-  }
-
```
olim7t, line 73: This constructor wasn't used anymore, no need to keep it.

core/src/main/java/com/datastax/oss/driver/api/core/cql/StatementBuilder.java
```diff
@@ -91,6 +91,7 @@ protected StatementBuilder(StatementT template) {
   @NonNull
   public SelfT setExecutionProfileName(@Nullable String executionProfileName) {
     this.executionProfileName = executionProfileName;
+    this.executionProfile = null;
     return self;
   }
```
olim7t, line 98: This solved a bug when combining DAO-level profile and `@StatementAttributes`, but I think it also makes sense for general usage of statement builders.

tomekl007 (approved):

adutra (approved):

PR #1431, opened 2020-04-16
Fix various issues with tests

emerkle826:
- Fixes  OSGi tests since Awaitility was added
- Fixes GraphPagingIT tests against DSE 6.8
- Increases timeout in MetadataManagerTests to try to make the tests less flaky

### DIFF
--- a/core/src/test/java/com/datastax/oss/driver/internal/core/metadata/MetadataManagerTest.java
+++ b/core/src/test/java/com/datastax/oss/driver/internal/core/metadata/MetadataManagerTest.java
@@ -17,7 +17,7 @@ package com.datastax.oss.driver.internal.core.metadata;
 
 import static com.datastax.oss.driver.Assertions.assertThat;
 import static com.datastax.oss.driver.Assertions.assertThatStage;
-import static org.assertj.core.api.Assertions.fail;
+import static org.awaitility.Awaitility.await;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.timeout;
 import static org.mockito.Mockito.verify;
@@ -36,21 +36,18 @@ import com.datastax.oss.driver.internal.core.metadata.schema.queries.SchemaQueri
 import com.datastax.oss.driver.internal.core.metrics.MetricsFactory;
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableSet;
-import com.datastax.oss.driver.shaded.guava.common.util.concurrent.Uninterruptibles;
 import io.netty.channel.DefaultEventLoopGroup;
-import io.netty.util.concurrent.Future;
 import java.net.InetSocketAddress;
 import java.time.Duration;
 import java.util.Collections;
 import java.util.List;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
@@ -144,7 +141,7 @@ public class MetadataManagerTest {
 
     // When
     CompletionStage<Void> refreshNodesFuture = metadataManager.refreshNodes();
-    waitForPendingAdminTasks();
+    waitForPendingAdminTasks(() -> metadataManager.refreshes.size() == 1);
 
     // Then
     assertThatStage(refreshNodesFuture).isSuccess();
@@ -173,7 +170,7 @@ public class MetadataManagerTest {
 
     // When
     CompletionStage<Void> refreshNodesFuture = metadataManager.refreshNodes();
-    waitForPendingAdminTasks();
+    waitForPendingAdminTasks(() -> metadataManager.refreshes.size() == 1);
 
     // Then
     assertThatStage(refreshNodesFuture).isSuccess();
@@ -228,7 +225,7 @@ public class MetadataManagerTest {
 
     // When
     metadataManager.addNode(broadcastRpcAddress);
-    waitForPendingAdminTasks();
+    waitForPendingAdminTasks(() -> metadataManager.addNodeCount == 1);
 
     // Then
     assertThat(metadataManager.refreshes).hasSize(1);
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/datastax/dse/driver/internal/core/graph/ContinuousGraphRequestHandler.java
```diff
@@ -99,7 +99,7 @@ protected Duration getGlobalTimeout() {
   @NonNull
   @Override
   protected Duration getPageTimeout(int pageNumber) {
-    return Duration.ZERO;
+    return this.globalTimeout;
```
emerkle826: I have concerns about the full effect of this change. Currently, a value of 0 means that the driver will wait forever for a response to successive page requests (the first page timeout is controlled by the globalTimeout).

I changed the return value here to be the globalTimeout value as the handler doesn't have a separate page timeouts (like [ContinuousCqlRequestHandler](https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/dse/driver/internal/core/cql/continuous/ContinuousCqlRequestHandler.java#L51-L52)).

The test that fails is [here](https://github.com/datastax/java-driver/pull/1431/files#diff-c96ea9340f349d35ca1642eb4f50c76aR490), as it expects the second page request to timeout while CCM has been paused. But with this method returning a Duration of 0, it never times out the request. Instead, we get a Heartbeat timeout 30 seconds later. This also generally causes many of the DatatProvider tests in that class to fail as more queries are attempted before the session has been fully re-established after CMM is resumed, which causes the session to die with an AllNodesFailed exception.

I'm not sure if this is the best solution, as maybe GraphPaging needs to support setting a timeout for `otherPages` like the ContiuousCqlRequestHandler does now.

integration-tests/src/test/java/com/datastax/dse/driver/api/core/graph/GraphPagingIT.java
```diff
@@ -69,6 +69,9 @@
                   .withStringList(
                       DefaultDriverOption.METRICS_NODE_ENABLED,
                       Collections.singletonList(DseNodeMetric.GRAPH_MESSAGES.getPath()))
+                  // Increase heartbeat interval to avoid interferences since we pause/stop the
+                  // server in this test
+                  .withDuration(DefaultDriverOption.HEARTBEAT_INTERVAL, Duration.ofMinutes(10))
```
emerkle826: If we remove the invalid test below, I don't think we need to increase the heartbeat here. The rest of the tests in this class pass fine without even coming close to tripping the 30 second heartbeat timeout.

integration-tests/src/test/java/com/datastax/dse/driver/api/core/graph/GraphPagingIT.java
```diff
@@ -461,38 +464,6 @@ public void should_trigger_global_timeout_async() throws InterruptedException {
     }
   }
 
-  @Test
-  public void should_trigger_global_timeout_async_after_first_page() throws InterruptedException {
```
tomekl007, line 465: Why do you want to remove this test? It exercises the global timeout for async paging.

adutra: Discussed in Slack: TLDR is this test is too broken to be kept.

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/MetadataManagerTest.java
```diff
@@ -308,7 +308,7 @@ private void waitForPendingAdminTasks() {
     // This works because the event loop group is single-threaded
     Future<?> f = adminEventLoopGroup.schedule(() -> null, 5, TimeUnit.NANOSECONDS);
     try {
-      Uninterruptibles.getUninterruptibly(f, 100, TimeUnit.MILLISECONDS);
+      Uninterruptibles.getUninterruptibly(f, 500, TimeUnit.MILLISECONDS);
```
adutra: Are we waiting for some specific event to occur? If so, would it be possible to use Awaitility to wait on that condition?

integration-tests/src/test/java/com/datastax/oss/driver/osgi/support/BundleOptions.java
```diff
@@ -167,6 +168,14 @@ public static CompositeOption simulacronBundles() {
                 simulacronVersion));
   }
 
+  public static CompositeOption awaitilityBundles() {
```
adutra, line 171: Note to myself: I need to port this to JAVA-2658.

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/MetadataManagerTest.java
emerkle826: I'm not sure,but I'll investigate further. With the timeout set to 100, I couldn't reproduce the failure locally, and it failed sporadically on Jenkins/CloudBees.
I can lower the value and reproduce it and figure out what I need to be waiting on.

integration-tests/src/test/java/com/datastax/oss/driver/osgi/support/BundleOptions.java
adutra: Done.

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/MetadataManagerTest.java
```diff
@@ -290,6 +287,8 @@ public void should_remove_node() {
   private static class TestMetadataManager extends MetadataManager {
 
     private List<MetadataRefresh> refreshes = new CopyOnWriteArrayList<>();
+    private int addNodeCount = 0;
```
adutra: These 2 fields probably should be volatile.

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/MetadataManagerTest.java
```diff
@@ -301,18 +300,24 @@ Void apply(MetadataRefresh refresh) {
...
-    Future<?> f = adminEventLoopGroup.schedule(() -> null, 5, TimeUnit.NANOSECONDS);
-    try {
-      Uninterruptibles.getUninterruptibly(f, 100, TimeUnit.MILLISECONDS);
-    } catch (ExecutionException e) {
-      fail("unexpected error", e.getCause());
-    } catch (TimeoutException e) {
-      fail("timed out while waiting for admin tasks to complete", e);
-    }
+  private void waitForPendingAdminTasks(Callable<Boolean> condition) {
+    await().atMost(500, TimeUnit.MILLISECONDS).until(condition);
```
adutra, line 325: Nice rework! 👍

adutra (changes requested):

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/MetadataManagerTest.java
emerkle826: Good catch. I did try to verify that the tests run in parallel don't trample each other here, and it seemed like they didn't, but `volatile` here makes sense.

adutra (approved):

PR #1432, opened 2020-04-16
Initial work to get ks/table metadata serializable

absurdfarce:


absurdfarce:
An initial idea for what might be involved with making keyspace + table metadata serializable (in order to support SPARKC-577).  I tried to use the same mechanism used with JAVA-1858, specifically applying this to a few impl classes that weren't covered in the earlier PR + the keyspace/table classes themselves.

I also added a basic test to vet the functionality against an artificially populated C* instance.  That test should properly be an IT but t'was implemented as a unit test for now running against a manually configured local instance.  To make the Travis builds happy I'll need to change that.

In a prior conversation @adutra had mentioned filing a JAVA ticket for the driver side of this work and then leveraging that effort in whatever changes are made for SPARKC-577.  I couldn't recall the outcome of that conversation so I haven't taken that course yet... thought it more useful to get this up for people to look at.  @adutra please let me know if we want to pursue that course and I'll file the ticket.

core/src/test/java/com/datastax/oss/driver/api/core/metadata/SerializableMetadataTest.java
```diff
@@ -0,0 +1,77 @@
...
+import org.apache.commons.lang3.SerializationUtils;
+import org.junit.Test;
+
+import java.net.InetSocketAddress;
+import java.util.Optional;
+
+public class SerializableMetadataTest {
+
+    /**
+     * Assuming the test below is run against the following:
```
tomekl007, line 19: why do you mean assuming? I think that it should the schema (ks, tables) should be created in the `beforeAll` instead of the comment. (CREATE KEYSPACE IF NO EXISTS, CREATE TABLE IF NOT EXISTS)

adutra:
> @adutra please let me know if we want to pursue that course and I'll file the ticket.

Please file a ticket. Thanks!

absurdfarce:
Closing, moving this work to a PR for the driver-side ticket (https://github.com/datastax/java-driver/pull/1433).  @tomekl007 re: the test... I had a comment above saying what's in place at the moment is a temporary thing to validate the work done so far.  I plan to convert that to an IT which would be responsible for populating the necessary data as part of the test.

adutra:
We did you have to close this PR and open another one?

absurdfarce:
@adutra This PR was using a branch name based on SPARKC-577... I wanted to get something more explicitly tied to JAVA-2741.  Wasn't a big deal but I thought it would be cleaner.

PR #1433, opened 2020-04-17
JAVA-2741: Extending serialization to keyspace/table metadata classes

absurdfarce:


integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SerializableMetadataIT.java
```diff
@@ -0,0 +1,105 @@
...
+    assertTrue(tableOption.isPresent());
+    TableMetadata table = tableOption.get();
+    assertTrue(table instanceof DefaultTableMetadata);
+
+    /* Validate that the table metadata is fully populated */
+    assertFalse(table.getPartitionKey().isEmpty());
+    assertFalse(table.getClusteringColumns().isEmpty());
+    assertFalse(table.getColumns().isEmpty());
+    // TODO: What kind of objects can be set for options here?  How do we guarantee they're
+    // serializable?
```
absurdfarce: Another odd case.  Options are presently defined as Objects and at the moment it looks like all of them are serializable (based on the contents of TableParser).  But this is hardly a guarantee going forward.  It'd be nice if we could change RelationMetadata.getOptions() to return a Map<CqlIdentifier, Serializable> but I can already hear @olim7t pointing out that that would be an API change (and with good reason).

Presumably we can just do the same thing here we did with the interfaces... instances of these interfaces are serializable as long as the concrete type is itself serializable.  In this case we'd have something in the docs saying the default table metadata impl is serializable as long as the options used are (which should be the common case).  Does that seem adequate?

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SerializableMetadataIT.java
```diff
@@ -0,0 +1,105 @@
...
+  @Test
+  public void should_contain_serializable_metadata() {
+
+    Optional<KeyspaceMetadata> ksOption =
+        sessionRule.session().getMetadata().getKeyspace("numbers");
+    assertTrue(ksOption.isPresent());
+    KeyspaceMetadata ks = ksOption.get();
+    assertTrue(ks instanceof DefaultKeyspaceMetadata);
+
+    /* Validate that the keyspace metadata is fully populated */
```
absurdfarce: "fully populated" here means "all the maps in DefaultKeyspaceMetadata have data in them".  This definition is based on the impl of the class and therefore isn't super-flexible.  I'd love to have a definition based on the class (i.e. if the concrete instance had some kind of isFullyPopulated() method or something) but doing so seemed to lead down some unpleasant avenues... so I decided to leave it at this for now.

I'm certainly open to better suggestions here.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/DefaultAggregateMetadata.java
```diff
@@ -40,7 +41,7 @@
   @NonNull private final DataType returnType;
   @NonNull private final FunctionSignature stateFuncSignature;
   @NonNull private final DataType stateType;
-  @NonNull private final TypeCodec<Object> stateTypeCodec;
+  @NonNull private final transient TypeCodec<Object> stateTypeCodec;
```
absurdfarce: Perhaps the most significant change in an otherwise straightforward PR.  It didn't seem reasonable to make the codec serializable so I went with this approach right now.  The IT still passes even though the codec won't be populated after deserialization becuase stateTypeCodec isn't included in the equals() impl.

My working assumption was that this change wouldn't be terribly significant in practice.  A missing codec here will cause AggregateMetadata.describe() will misbehave on deserialized instances.  But the presumed motivation for making these classes serializable is (at least largely) to use them directly as DTOs... and their performance in that role should still be intact here.

Perhaps another not worth documenting... but not much more?

absurdfarce: Correction: AggregateMetadata.describe() will be fine in this case.  It will make an attempt to use the codec but if that results in an exception (which it will in this case by way of an NPE) it falls back to toString() instead.  Still seems worthwhile adding an explicit test for this functionality as a guard against future breakage.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/DefaultAggregateMetadata.java
```diff
@@ -109,6 +110,13 @@ public DataType getStateType() {
     if (initCond == null) {
       return Optional.empty();
     }
+    if (stateTypeCodec == null) {
+      LOG.debug(
+          String.format(
+              "Null codec for INITCOND for %s.%s, using toString instead.  This often indicates use of a deserialized AggregateMetadata instance.",
+              keyspace.asInternal(), signature.getName().asInternal()));
+      return Optional.of(initCond.toString());
+    }
```
absurdfarce: This is not strictly necessary; the try/catch block below will catch the NPE in cases where stateTypeCodec is null from deserialization.  I added the explicit case here only to avoid adding messages at WARN to the logging infrastructure... DEBUG seemed like a more reasonable level for this (admittedly probably rare in practice) case.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SerializableMetadataIT.java
```diff
@@ -0,0 +1,152 @@
...
+    // TODO: What kind of objects can be set for options here?  How do we guarantee they're
+    // serializable?
+    assertFalse(table.getOptions().isEmpty());
+    assertFalse(table.getIndexes().isEmpty());
+
+    SerializationUtils.roundtrip((DefaultKeyspaceMetadata) ks);
+  }
+
+  @Test
+  public void should_support_describe_for_aggregates_with_deserialized_instances() {
```
absurdfarce: This test already passed due to the exception handling in DefaultAggregateMetadata.describe().  Adding this as an explicit integration test as a guard against future breakage in functionality (since that's the only area affected by the transient declaration).

adutra:
Shouldn't you add a `serialVersionUID` field to the modified classes?

absurdfarce:
Yes, I should @adutra.  This was a mistake on my part.  I didn't see the additions on the PR for JAVA-1858 and missed them in spot checks on the affected classes so I thought the convention was to not include them.  I'll add them in the next round of fixes.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/DefaultAggregateMetadata.java
olim7t: Maybe we should pre-compute the result of `formatInitCond()` in the constructor, and store that as a field instead of the codec.
The original intent in storing the codec was to avoid formatting eagerly if it's never going to be used, but I'll take that over a deserialized object that doesn't describe correctly.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SerializableMetadataIT.java
```diff
@@ -0,0 +1,153 @@
...
+  public void should_contain_serializable_metadata() {
+
+    KeyspaceMetadata ks = getAndValidateKsMetadata();
+
+    /* Validate that the keyspace metadata is fully populated */
+    assertFalse(ks.getUserDefinedTypes().isEmpty());
+    assertFalse(ks.getTables().isEmpty());
+    assertFalse(ks.getViews().isEmpty());
+    assertFalse(ks.getFunctions().isEmpty());
+    assertFalse(ks.getAggregates().isEmpty());
```
olim7t: Nit: we've pretty much settled on AssertJ as de facto standard by now, please use it instead of JUnit assertions. As a bonus, it has more fluent alternatives for most of those checks:
```java
assertThat(ks.getUserDefinedTypes()).isEmpty();
```
\+ the optional example from my other comment

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SerializableMetadataIT.java
```diff
@@ -0,0 +1,153 @@
...
+    /* Validate that the table metadata is fully populated */
+    assertFalse(table.getPartitionKey().isEmpty());
+    assertFalse(table.getClusteringColumns().isEmpty());
+    assertFalse(table.getColumns().isEmpty());
+    // TODO: What kind of objects can be set for options here?  How do we guarantee they're
+    // serializable?
+    assertFalse(table.getOptions().isEmpty());
+    assertFalse(table.getIndexes().isEmpty());
+
+    SerializationUtils.roundtrip((DefaultKeyspaceMetadata) ks);
```
olim7t: Nit -- we have a built-in utility for that: `SerializationHelper.serializeAndDeserialize`.

It's essentially the same thing, but we wrote it at a time when commons-lang wasn't yet available in the classpath (I'm not even sure how it landed here TBH). Let's use it everywhere for the sake of consistency.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SerializableMetadataIT.java
```diff
@@ -0,0 +1,153 @@
...
+            "CREATE OR REPLACE FUNCTION avgFinal ( state tuple<int,bigint> ) CALLED ON NULL INPUT RETURNS double LANGUAGE java AS"
+                + " 'double r = 0; if (state.getInt(0) == 0) return null; r = state.getLong(1); r/= state.getInt(0); return Double.valueOf(r);'",
+            "CREATE AGGREGATE IF NOT EXISTS average ( int ) SFUNC avgState STYPE tuple<int,bigint> FINALFUNC avgFinal INITCOND (0,0)",
+            "TRUNCATE all",
+            "INSERT INTO all (val,doub,str) VALUES (1,2,'one')",
+            "INSERT INTO all (val,doub,str) VALUES (2,4,'two')",
+            "INSERT INTO all (val,doub,str) VALUES (3,6,'three')",
+            "CREATE MATERIALIZED VIEW IF NOT EXISTS big AS SELECT * FROM all"
+                + " WHERE val is not null and doub is not null and val > 2"
+                + " PRIMARY KEY (val,doub)");
```
olim7t: Schema-related testing is always annoying because it has to work against all the server versions in our CI suite. In particular, the functions, aggregates and views used here are going to be a problem with legacy Cassandra versions.

I don't like to piggyback on existing tests, but for this particular case we could avoid a lot of headaches by reusing `DescribeIT`, which already has the logic to deal with backward compatibility. It could be as simple as throwing a serialization roundtrip in the middle of the existing method, or we could add a separate one.

As a bonus, comparing the `describeWithChildren` output of the top-level keyspace gives us a pretty reliable deep comparison, without having to enumerate all the children.

olim7t (changes requested):
Looks pretty good, just a couple of minor remarks.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SerializableMetadataIT.java
olim7t: > I can already hear @olim7t pointing out that that would be an API change

I've thought it so loud that you heard me 😉 

+1 to add a note in the javadocs.

Note that this should remain mostly the responsibility of the driver; it's technically possible to plug custom schema logic that parses custom table options, but I don't think it's very likely to happen. So it will be up to us to keep that in mind and test it correctly if new option types get added.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SerializableMetadataIT.java
olim7t: Would that notion of "fully populated" be of any use outside of tests though?

If this is more about how deep we should go with the checks, see my suggestion in another comment to merge this with `DescribeIT`.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/DefaultAggregateMetadata.java
absurdfarce: Yeah, I think I agree.  I mean the describe() op does have a fallback to toString() if the codec chucks an exception while doing it's formatting... but I agree it's probably better to avoid the transient/nullable stuff if we can.

I'll make the change.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SerializableMetadataIT.java
absurdfarce: No, the idea of "fully populated" doesn't really have any meaning outside of tests.  That's part of the reason I'm not lobbying for it... it feels like a @VisibleForTesting kind of thing of limited utility

I'll look into this more as I look to migrating this stuff to DescribeIT

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SerializableMetadataIT.java
absurdfarce: Hmmm, interesting... I'll look more at DescribeIT and see what's involved in such a migration.  Doesn't sound too bad on the face of it.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SerializableMetadataIT.java
absurdfarce: Bah, that's on me.  I know AssertJ is preferred... I used the JUnit checks when I was getting this test going and forgot to convert before I threw this open to review. :facepalm: 

Will fix.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SerializableMetadataIT.java
absurdfarce: > I've thought it so loud that you heard me :wink: 

I mean, there's also the fact that I agree with you... the two work together nicely :)

integration-tests/src/test/java/com/datastax/oss/driver/api/core/metadata/SerializableMetadataIT.java
absurdfarce: Looks like it's Gremlin's fault:

```
[INFO] +- org.apache.tinkerpop:gremlin-core:jar:3.4.5:compile
[INFO] |  +- org.apache.tinkerpop:gremlin-shaded:jar:3.4.5:compile
[INFO] |  +- commons-configuration:commons-configuration:jar:1.10:compile
[INFO] |  |  \- commons-lang:commons-lang:jar:2.6:compile
[INFO] |  +- commons-collections:commons-collections:jar:3.2.2:compile
[INFO] |  +- org.apache.commons:commons-lang3:jar:3.8.1:compile
[INFO] |  +- org.yaml:snakeyaml:jar:1.15:compile
[INFO] |  +- com.carrotsearch:hppc:jar:0.7.1:compile
[INFO] |  +- com.jcabi:jcabi-manifests:jar:1.1:compile
[INFO] |  |  \- com.jcabi:jcabi-log:jar:0.14:compile
[INFO] |  +- com.squareup:javapoet:jar:1.11.1:compile
[INFO] |  +- net.objecthunter:exp4j:jar:0.4.8:compile
[INFO] |  \- org.slf4j:jcl-over-slf4j:jar:1.7.25:compile
```

I love that the same artifact brings in both commons-lang2 (via a transitive dep) and commons-lang3 directly. :rage:

absurdfarce:
Previous standalone IT has now been moved over to DescribeIT per suggestion above.  As a nice bonus I added an explicit clustering column declaration to one of the tables in the CQL files used by that test; it was needed for the serialization test but also should help the describe test by more fully exercising that code.  Also added some version gates around some of the "is this metadata object fully populated" tests in order to avoid checking for functionality in prior versions which wasn't supported yet.

I _believe_ everything referenced in the initial round of reviews has now been addressed... somebody correct me if I missed something.

integration-tests/src/test/java/com/datastax/oss/driver/core/metadata/DescribeIT.java
```diff
@@ -158,4 +202,35 @@ private File getScriptFile() {
...
+    for (int i = 1; i < statements.size(); i++) {
+      String statement = statements.get(i);
+      try {
+        SESSION_RULE.session().execute(statement);
+      } catch (Exception e) {
+        fail("Error executing statement %s (%s)", statement, e);
+      }
+    }
+
+    return scriptContents;
```
olim7t: nit: return value is not used

olim7t (changes requested):
Sorry I missed something during the last review 🤦 : there are custom metadata implementations for DSE. I guess we need to make them serializable too. You can test that by running `DescribeIT` against DSE (e.g. `-Dccm.dse=true -Dccm.version=6.8.0`).

Also nit: please add an entry in the changelog.

integration-tests/src/test/java/com/datastax/oss/driver/core/metadata/DescribeIT.java
absurdfarce: Yeah, this was put in there to make IntelliJ happy.  I'll change the method to just return a list of failed statements and do the fail() at the caller.

olim7t: Why not just make the method void?

olim7t (approved):

PR #1434, opened 2020-04-18
JAVA-2740: Ignore peer entries with data_center null

zmalik:
This PR intention is to provide a fix [JAVA-2740](https://datastax-oss.atlassian.net/projects/JAVA/issues/JAVA-2740) to ignore peer entry when a null entry for `schema_version` is found.
Right now it will just throw a `java.lang.IllegalArgumentException: Multiple entries with same key`

edit: After @adutra review, I have moved to the check from `schema_version` to `data_center` column

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
@@ -539,13 +539,16 @@ public class DefaultTopologyMonitor implements TopologyMonitor {
    * node's broadcast RPC address and host ID; otherwise the driver may not work properly.
    */
   protected boolean isPeerValid(AdminRow peerRow) {
-    boolean hasPeersRpcAddress = peerRow.getInetAddress("rpc_address") != null;
+    boolean hasPeersRpcAddress = !peerRow.isNull("rpc_address");
     boolean hasPeersV2RpcAddress =
-        peerRow.getInetAddress("native_address") != null
-            && peerRow.getInteger("native_port") != null;
+        !peerRow.isNull("native_address") && !peerRow.isNull("native_port");
     boolean hasRpcAddress = hasPeersV2RpcAddress || hasPeersRpcAddress;
-    boolean hasHostId = peerRow.getUuid("host_id") != null;
-    boolean valid = hasRpcAddress && hasHostId;
+    boolean valid =
+        hasRpcAddress
+            && !peerRow.isNull("host_id")
+            && !peerRow.isNull("data_center")
+            && !peerRow.isNull("rack")
+            && !peerRow.isNull("tokens");
     if (!valid) {
       LOG.warn(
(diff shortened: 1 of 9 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitor.java
```diff
@@ -545,7 +545,8 @@ protected boolean isPeerValid(AdminRow peerRow) {
             && peerRow.getInteger("native_port") != null;
     boolean hasRpcAddress = hasPeersV2RpcAddress || hasPeersRpcAddress;
     boolean hasHostId = peerRow.getUuid("host_id") != null;
-    boolean valid = hasRpcAddress && hasHostId;
+    boolean hasSchemaVersion = peerRow.getUuid("schema_version") != null;
```
adutra: As I said in the Jira ticket, we should either extend the check to all other columns that might be null and create issues down the line, or chose a better column (datacenter?).
Also, this is not enough to fix the generic problem with duplicate keys in the nodes map.

zmalik: thanks, @adutra! 
I'm not sure if other columns being `null` would qualify the peer row being invalid. For example, `preferred_ip` can be `null` and entry will be still valid. But I agree datacenter feels more relevant. Let me change this for the datacenter. 

>Also, this is not enough to fix the generic problem with duplicate keys in the nodes map. 

when reviewing on `system.peers` row this will be hard. And we would need some check when building that map? because just ignoring duplicate keys could get us into trouble also if we ignore the correct key. 

I would say the driver can get better in ignoring invalid entries where we are certain those are `bogus` entries, and this is already an improvement. If we get two valid entries and we need to ignore one of them we just ignore any?

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/DefaultTopologyMonitorTest.java
```diff
@@ -407,6 +407,9 @@ public void should_skip_invalid_peers_row_v2(String columnToCheck) {
       case "host_id":
         when(peer2.getUuid(columnToCheck)).thenReturn(null);
         break;
+      case "data_center":
+        when(peer2.getUuid(columnToCheck)).thenReturn(null);
```
adutra: This should be `when(peer2.getDatacenter(columnToCheck)).thenReturn(null);`

EDIT sorry:

`when(peer2.getString(columnToCheck)).thenReturn(null);`

zmalik: nice catch! thanks!

zmalik:
rebased from `4.x` branch

olim7t (approved):
As mentioned on the jira ticket, I amended your fix to include rack and tokens.

I also added a commit to deal with duplicate host_ids in case the peer check lets a corrupt row slip through.

PR #1435, opened 2020-04-18
JAVA-2742: Expose session health diagnostics

adutra:


bom/pom.xml
```diff
@@ -61,7 +61,7 @@
       <dependency>
         <groupId>com.datastax.oss</groupId>
         <artifactId>native-protocol</artifactId>
-        <version>1.4.10-SNAPSHOT</version>
+        <version>1.4.10-alpha1</version>
```
adutra, line 64: This si temporary, to allow spring-boot builds relying on artifacts not released yet.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/token/TokenRange.java
```diff
@@ -144,4 +144,9 @@
    */
   @NonNull
   TokenRange mergeWith(@NonNull TokenRange that);
+
+  /** Formats this range as a string of the following generic shape: {@code ]start,end]}. */
+  default String format() {
+    return String.format("]%s,%s]", getStart().format(), getEnd().format());
```
alexott, line 150: why it's selected to represent this way? why not as in mathematics `(start,end]` ? Maven uses such notation as well...

adutra: Both notations are international standards, see [here](https://en.wikipedia.org/wiki/Interval_(mathematics)#Notations_for_intervals). But I can change for the Maven one if everybody agrees.

alexott: I'm not mathematician :-)

P.S. I'll continue to look through PR tomorrow - thank you for implementing it - it's really useful functionality.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/NodeGroupDiagnostic.java
```diff
@@ -0,0 +1,88 @@
...
+  @Override
+  default Status getStatus() {
+    if (getTotal() == 0 || getDown() == getTotal()) {
+      return Status.UNAVAILABLE;
+    } else if (getDown() == 0 && getUnknown() == 0) {
+      return Status.AVAILABLE;
+    } else {
+      return Status.PARTIALLY_AVAILABLE;
+    }
+  }
```
absurdfarce: Nit: the else blocks actually aren't necessary here.  You're doing an early return (👍 to that BTW) so this can be simplified:

```java
  @NonNull
  @Override
  default Status getStatus() {
    if (getTotal() == 0 || getDown() == getTotal()) {
      return Status.UNAVAILABLE;
    }
    if (getDown() == 0 && getUnknown() == 0) {
      return Status.AVAILABLE;
    }
    return Status.PARTIALLY_AVAILABLE;
  }
```

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/Status.java
```diff
@@ -0,0 +1,45 @@
...
+  /** The diagnosed system is in an unknown state. */
+  UNKNOWN;
+
+  /**
+   * Merges the given {@link Status} with this one, in a pessimistic way: unavailable statuses take
+   * precedence over available ones.
+   */
+  public Status mergeWith(Status that) {
+    return this.compareTo(that) > 0 ? this : that;
+  }
```
absurdfarce, line 48: Isn't this... just a re-implementation of natural ordering?  I don't think you actually need this.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TokenRingDiagnostic.java
```diff
@@ -0,0 +1,71 @@
...
+   * <p>The ring is considered available when all its {@linkplain TokenRange token ranges} are
+   * {@linkplain TokenRangeDiagnostic#isAvailable() available}.
+   */
+  @NonNull
+  @Override
+  default Status getStatus() {
+    return getTokenRangeDiagnostics().stream()
+        .map(TokenRangeDiagnostic::getStatus)
+        .reduce(Status::mergeWith)
+        .orElse(Status.UNKNOWN);
```
absurdfarce, line 72: Per comment above... unless I'm missing something this is identical to:

```java
  @NonNull
  @Override
  default Status getStatus() {
    return getTokenRangeDiagnostics().stream()
        .map(TokenRangeDiagnostic::getStatus)
        .max(Comparator.naturalOrder())
        .orElse(Status.UNKNOWN);
  }
```

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/diagnostic/ring/DefaultTokenRingDiagnostic.java
```diff
@@ -0,0 +1,167 @@
...
+    builder.put("availableRanges", available.size()).put("unavailableRanges", unavailable.size());
+    if (!unavailable.isEmpty()) {
+      addTop10UnavailableRanges(builder, unavailable);
+    }
+  }
+
+  /**
+   * Extracts up to 10 of the most severely unavailable ranges and returns a map of availability
+   * reports keyed by token ranges, formatted for display.
+   */
```
absurdfarce, line 139: Semi-nit: this comment isn't quite right.  This is a void method so it doesn't return anything... it updates the input builder. :)

Perhaps this method could return an ImmutableMap.Builder<String, Object> type and just return the builder it gets as an arg (after modification)?  That would make this comment correct and would allow you to avoid the side-effect bits here, at least at an API level.  Similar argument applies to addBasicDetails(builder) and addTokenRangeDiagnosticDetails(builder).

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TokenRingDiagnostic.java
```diff
@@ -0,0 +1,71 @@
...
+   * established. For all other consistency levels, returns {@link Optional#empty()}.
+   */
+  Optional<String> getDatacenter();
+
+  /**
+   * @return the {@link TokenRangeDiagnostic}s for each individual {@linkplain TokenRange token
+   *     range}.
+   */
+  @NonNull
+  SortedSet<TokenRangeDiagnostic> getTokenRangeDiagnostics();
```
absurdfarce, line 58: Is there a reason this returns a SortedSet rather than some other type which has a more natural notion of ordering like a List?  It doesn't look like we're actually exploiting the set semantics at all, at least at the moment.  Did you have some future functionality in mind w.r.t. making this a set?

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TokenRingDiagnostic.java
adutra: Yes but you are not expressing the intent of the operation anymore. The `mergeWith` method has a more meaningful name that clearly outlines why we need a `reduce` here. The fact that `mergeWith` is implemented according to the enum's natural ordering is an implementation detail.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/diagnostic/ring/DefaultTokenRingDiagnostic.java
adutra: The comment is from a previous version of the method. I'm fine with side effects here, so I'll just change the javadocs.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TokenRingDiagnostic.java
adutra: This is to order diagnostics by their token range's natural ordering.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/NodeGroupDiagnostic.java
```diff
@@ -0,0 +1,88 @@
...
+
+  /**
+   * Returns the status of this node group. The status will be {@link Status#AVAILABLE} if all nodes
+   * are up; {@link Status#UNAVAILABLE} if all nodes are down, or if the group has no node at all.
+   * In all other cases the status will be {@link Status#PARTIALLY_AVAILABLE}.
+   */
+  @NonNull
+  @Override
+  default Status getStatus() {
+    if (getTotal() == 0 || getDown() == getTotal()) {
```
tomekl007: Are we providing happens-before relation here? I mean that the `getTotal() == 0` may not necessary mean down. It may be the situation when the `getStatus()` was executed before there was underlying health check that incremented `getTotal`. Or is it not possible? Maybe when `getTotal()==0` we should return UNKNOWN state?

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TokenRingDiagnostic.java
adutra: But I agree that `TopologyDiagnostic#getLocalDiagnostics` does not need to return a `SortedMap`.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/NodeGroupDiagnostic.java
adutra: This is... really really minor. But OK to change if that pleases you.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Metadata.java
```diff
@@ -126,4 +133,51 @@
...
+   * best-effort only, and are not meant to replace a proper operational surveillance tool.
+   *
+   * @param keyspace the {@link KeyspaceMetadata keyspace} to analyze.
+   * @param consistencyLevel the {@link ConsistencyLevel consistency level} to use.
+   * @param datacenter the datacenter name; only required for datacenter-local consistency levels,
+   *     may be null otherwise.
+   * @see DefaultDriverOption#METADATA_TOKEN_MAP_ENABLED
+   * @see #getKeyspace(String)
+   */
+  default Optional<TokenRingDiagnostic> generateTokenRingDiagnostic(
```
adutra: I wonder if throwing an error would be better than returning `Optional.empty()`.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/NodeGroupDiagnostic.java
adutra: > that the getTotal() == 0 may not necessary mean down

yes it does, cf. javadocs: 

```
The status will be {@link Status#AVAILABLE} if all nodes are up; {@link Status#UNAVAILABLE} if all nodes are down, or if the group has no node at all.
```

> It may be the situation when the getStatus() was executed before there was underlying health check that incremented getTotal.

That shouldn't be allowed, and our implementations are all immutable, so this cannot happen.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Metadata.java
```diff
@@ -126,4 +133,51 @@
...
+      @NonNull KeyspaceMetadata keyspace,
```
olim7t: We should take the keyspace identifier instead. It feels weird to call `Metadata` to obtain a `KeyspaceMetadata`, only to pass it to another operation on `Metadata`.

Also, it will eliminate the possibility of passing a stale instance.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/Status.java
```diff
@@ -0,0 +1,45 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core.metadata.diagnostic;
+
+/** The status of a {@link Diagnostic}. */
+public enum Status {
+
+  // the order of enums matter, see mergeWith
```
olim7t: Then I would advocate changing the code to make it not matter.

It doesn't cost a lot to add an explicit private `precedence` field, and it will make the code easier to maintain. Also if we add a constant in a later version, we'll have to put it last for binary compatibility, but that might not necessarily match its precedence.

I ran into a silly bug in the 3.x protocol code the other day because of something like this.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TokenRingDiagnostic.java
```diff
@@ -0,0 +1,133 @@
...
+   */
+  @NonNull
+  ConsistencyLevel getConsistencyLevel();
+
+  /**
+   * If this diagnostic was produced for a {@linkplain ConsistencyLevel#isDcLocal()
+   * datacenter-local} consistency level, returns the datacenter for which this diagnostic was
+   * established. For all other consistency levels, returns {@link Optional#empty()}.
+   */
+  Optional<String> getDatacenter();
```
olim7t, line 51: Nit: missing nullability annotation (unless there's a rule about omitting it for optional results?)

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Metadata.java
```diff
@@ -126,4 +133,51 @@
...
     return Optional.empty();
   }
+
+  /**
+   * Generates a new {@link TopologyDiagnostic} for this metadata. See the javadocs of {@link
+   * TopologyDiagnostic} for a detailed explanation on the contents of this diagnostic.
+   *
+   * <p>Note: diagnostics rely on Gossip events received by the driver. But Gossip events are not
+   * 100% reliable, and therefore, the accuracy of reported health statuses should be considered
+   * best-effort only, and are not meant to replace a proper operational surveillance tool.
```
olim7t: Is this specifically about ignored nodes being in the `UNKNOWN` state? Maybe we should get into more details, to give users better expectations.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/diagnostic/topology/DefaultTopologyDiagnostic.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+    DefaultTopologyDiagnostic that = (DefaultTopologyDiagnostic) o;
+    return globalDiagnostic.equals(that.globalDiagnostic)
+        && localDiagnostics.equals(that.localDiagnostics);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(globalDiagnostic, localDiagnostics);
+  }
```
olim7t, line 142: How much effort would it take to add a custom `toString()` to all diagnostic objects?
I feel like it could be useful for quick-and-dirty testing.

EDIT-- it looks like `getDetails().toString()` is what I'm looking for.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TopologyDiagnostic.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  /**
+   * A health {@link Diagnostic} for a group of nodes, detailing how many nodes were found in total,
+   * how many were up, and how many were down.
+   *
+   * <p>The breadth of the report's scope depends on how it was created; it may refer to the entire
+   * cluster, or to just a datacenter.
+   */
+  interface NodeGroupDiagnostic extends Diagnostic {
```
olim7t: How about making this a top-level type? Some IDEs will import it by default as `TopologyDiagnostic.NodeGroupDiagnostic`, not a big fan.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Metadata.java
```diff
@@ -126,4 +133,51 @@
...
+   */
+  default TopologyDiagnostic generateTopologyDiagnostic() {
```
olim7t, line 149: 👍 to naming, it makes it clear that those are not pre-computed

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TopologyDiagnostic.java
```diff
@@ -0,0 +1,135 @@
...
+     * at all. In all other cases the status will be {@link Status#PARTIALLY_AVAILABLE}.
+     */
+    @NonNull
+    @Override
+    default Status getStatus() {
+      if (getTotal() == 0 || getDown() == getTotal()) {
+        return Status.UNAVAILABLE;
+      }
+      if (getDown() == 0 && getUnknown() == 0) {
+        return Status.AVAILABLE;
```
olim7t: So... the big question of this PR is how we are going to handle the UNKNOWN state.

With the default LBP, any multi-DC cluster currently shows as partially available, even when all nodes are up:
```
{status=PARTIALLY_AVAILABLE, total=4, up=2, down=0, unknown=2,
  dc1={total=2, up=2, down=0, unknown=0},
  dc2={total=2, up=0, down=0, unknown=2}}
```
This is guaranteed to confuse users (legitimately), and we'll get the question over and over until the end of times: "why does it say partially available, I see that all my nodes are up". On the other hand, if we take the optimistic approach, we're in the opposite situation and will report fully available when some nodes are down.

For the topology diagnostic, I think optimistic is the way to go. One could argue that "availability" in this context means the ability to find a coordinator, so ignored nodes should not matter.

For the token ring diagnostic, it gets more complicated, because ignored nodes _can_ impact the availability of data for non-local CLs.
I don't have a good answer here. This reflects what you said in one of the comments: the driver is not a general purpose monitoring tool... which maybe raises the question of how far a built-in feature should go.

PS: this has nothing to do with UNKNOWN itself; I completely stand by my decision to introduce that new state in driver 4. Driver 3 just hid the problem, it had optimism built-in, without any other possibility.

olim7t (commented):
First round of comments, I'm still looking at the token ring diagnostic implementations.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/token/TokenRange.java
olim7t: Let's stick to what we've used so far (I didn't give it much thought at the time, but glad it's  a standard :smile:).

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/diagnostic/ring/TokenRingDiagnosticGeneratorFactory.java
```diff
@@ -0,0 +1,159 @@
...
+  protected Optional<TokenRingDiagnosticGenerator> createForSimpleStrategy(
+      KeyspaceMetadata keyspace, ConsistencyLevel consistencyLevel) {
+    ConsistencyLevel filteredConsistencyLevel =
+        ConsistencyLevels.filterForSimpleStrategy(consistencyLevel);
+    if (filteredConsistencyLevel != consistencyLevel) {
+      LOG.warn(
+          String.format(
+              "Consistency level %s is not compatible with the SimpleStrategy replication configured for keyspace %s."
+                  + "Token ring health reports will assume %s instead.",
+              consistencyLevel, keyspace.getName().asCql(true), filteredConsistencyLevel));
```
olim7t: We could fail here, it's up to the user to pass an applicable CL.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TopologyDiagnostic.java
adutra: > This is guaranteed to confuse users (legitimately), and we'll get the question over and over until the end of times: "why does it say partially available, I see that all my nodes are up". On the other hand, if we take the optimistic approach, we're in the opposite situation and will report fully available when some nodes are down.

I think you are right about keeping an optimistic approach for the topology diagnostic. I went for "PARTIALLY_AVAILABLE" when there are nodes in UNKNOWN state, but as you said, this is going to confuse users. I can change that.

(On a side note, I though about introducing a `float getConfidenceLevel()` property to `Diagnostic` and only set it to 1.0 when there are no nodes in UNKNOWN state. The more nodes in that state, the lower the confidence level.)

> For the token ring diagnostic, it gets more complicated, because ignored nodes can impact the availability of data for non-local CLs.
I don't have a good answer here. This reflects what you said in one of the comments: the driver is not a general purpose monitoring tool... which maybe raises the question of how far a built-in feature should go.

Absolutely. I mulled this over and over myself, and in the end, I think that any cure would be worse than the disease. On one hand, I genuinely believe that such a diagnostic API belongs in the driver, it is just too complex to be left for users to implement. On the other hand, the driver does not have a reliable way to tell if a node is up or down, and that makes the whole API implementation a bit fragile. The only proper answer to this issue is if/when Cassandra itself decides to expose an API to interrogate node states. When that happens, we might be able to change the current implementation to something 100% (or almost) reliable. In the meanwhile I think the only way to go is to implement this API with our current tooling, and put a big fat warning on top of it, clearly stating that this is implemented as best-effort. I'm totally fine to make the current warnings even scarier if you like.

>  I completely stand by my decision to introduce that new state in driver 4

And I thank you for that, I now understand much better how remote/ignored nodes are handled by the driver.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/diagnostic/ring/AbstractTokenRingDiagnosticGenerator.java
```diff
@@ -0,0 +1,82 @@
...
+      TokenMap tokenMap = metadata.getTokenMap().get();
+      tokenRangeDiagnostics = generateTokenRangeDiagnostics(tokenMap);
+    } else {
+      tokenRangeDiagnostics = Collections.emptyNavigableSet();
+    }
+    return generateRingDiagnostic(tokenRangeDiagnostics);
+  }
+
+  protected Set<TokenRangeDiagnostic> generateTokenRangeDiagnostics(TokenMap tokenMap) {
+    Set<TokenRangeDiagnostic> reports = new HashSet<>();
```
olim7t: nit: this can be presized

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/diagnostic/ring/AbstractTokenRingDiagnosticGenerator.java
```diff
@@ -0,0 +1,82 @@
...
+          allReplicas.stream().filter(this::isAlive).collect(Collectors.toSet());
+      TokenRangeDiagnostic report = generateTokenRangeDiagnostic(range, aliveReplicas);
+      reports.add(report);
+    }
+    return reports;
+  }
+
+  protected boolean isAlive(Node node) {
+    // Be optimistic and count nodes in unknown state as alive
+    return node.getState() == NodeState.UP || node.getState() == NodeState.UNKNOWN;
```
olim7t, line 77: This answers my other comment on the topology diagnostic, we take the optimistic route here instead.

As I said, I don't have a good answer: "here's a report, but it might be completely wrong because it's based on information that the driver doesn't really have" 😕

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/diagnostic/ring/EachQuorumTokenRingDiagnosticGenerator.java
```diff
@@ -0,0 +1,86 @@
...
+    super(metadata, keyspace);
+    Objects.requireNonNull(replicationFactorsByDc, "replicationFactorsByDc cannot be null");
+    this.requiredReplicasByDc =
+        replicationFactorsByDc.entrySet().stream()
+            .collect(
+                ImmutableMap.toImmutableMap(
+                    Entry::getKey,
+                    entry ->
+                        ConsistencyLevels.requiredReplicas(
+                            ConsistencyLevel.EACH_QUORUM, entry.getValue())));
```
olim7t: Sorry to rehash this, but I have to say it at least once: this stuff is so hard to read.

olim7t (commented):
Not a lot more to add. The implementation is sound, my issues are more conceptual.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Metadata.java
olim7t: 👍 There are a lot of error cases where you just log warnings and return empty optionals/collections. I think it makes it harder to realize that something went wrong, it would be more efficient to surface them directly as exceptions.

absurdfarce: But... they aren't always exception cases.  In some cases the empty optional is returned here if something is _disabled_ or you have a local CL.  How are those worth an exception?

I'd argue the Optional usage is right, at least for that case.  If you have something that's truly worthy of an exception you could do that.  You _might_ want to throw an exception for the unsupported replication strategy case, for instance... I could see that more readily than the "token map is disabled" case.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/Status.java
absurdfarce: If we're going to go down that path, though, isn't precedence a context-sensitive operation?  Perhaps the ordering shouldn't be built into the type at all but should be implemented as a Comparator which would allow different callers to define whatever prioritization makes sense to them.  At the moment there's a single consumer of this functionality so moving that logic into a Comparator at the current call site seems pretty straightforward.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TokenRingDiagnostic.java
absurdfarce: You still should be able to annotate whether the Optional instance itself is ever null or not; we certainly do this in other areas of the driver code.  👍

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TokenRingDiagnostic.java
absurdfarce: That's fine... but you're dealing with an internal implementation here.  If you're going to use an ordering built into the type mergeWith() is identical to the natural order of the enum... which to me seems more confusing.

I'd actually argue the right answer here is to define a Comparator specific to this use case and not build precedence into the type.  But if we don't do that I'd rather just use natural order directly than do it through one level of remove.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TokenRingDiagnostic.java
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  /**
+   * A health {@link Diagnostic} for a given {@link TokenRange}, detailing whether or not the
+   * configured consistency level can be achieved for the configured keyspace on it.
+   *
+   * <p>A {@link TokenRangeDiagnostic} is generated as part of a broader {@link
+   * TokenRingDiagnostic}.
+   */
+  interface TokenRangeDiagnostic extends Diagnostic, Comparable<TokenRangeDiagnostic> {
```
absurdfarce, line 82: Is there a reason this isn't a top-level interface (like TokenRingDiagnostic)?

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TopologyDiagnostic.java
olim7t: > if/when Cassandra itself decides to expose an API to interrogate node states

One relatively low-impact way to do that would be to add a new protocol message that triggers the server to send an immediate UP/DOWN event for every node. The driver would send it after the initial node list refresh (before that would be more complicated, so we can't make it a simple STARTUP option unfortunately).

core/src/main/java/com/datastax/oss/driver/api/core/metadata/Metadata.java
olim7t: The problem with the empty optional is that it doesn't surface the error to the caller. Your only option is to go sift through the logs to find the message (hoping warn level is enabled). There's no convenient way to do this programmatically either: for example if the diagnostic was requested from a GUI and you'd like to let the user know why it's not showing up, it's going to be a nightmare to extract that message.

I realize that this problem already exists in some parts of the API: for example, it might have been useful to know why `Metadata.getTokenMap()` is absent... But here it seems more obvious because it's an operation that's invoked explicitly, not something the driver does in the background.

PS- before someone brings it up: no, we don't need a Try monad 😉

adutra: OK I'm changing this to throw exceptions.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TopologyDiagnostic.java
adutra: Ah I made them inner interfaces precisely so that only top-level sub-interfaces show up in code navigators displaying the code tree.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TopologyDiagnostic.java
adutra: > One relatively low-impact way to do that would be to add a new protocol message that triggers the server to send an immediate UP/DOWN event for every node. The driver would send it after the initial node list refresh (before that would be more complicated, so we can't make it a simple STARTUP option unfortunately).

That's a very good idea that's worth suggesting to the C* team. The only complication that I see is that sometimes, according to the manual testing that I did for this PR, the driver loses connection to the cluster entirely, and when it reconnects, the cluster state is usually not in sync anymore and many nodes show up as DOWN while they shouldn't. So I think for correctness sake we would have to interrogate the cluster at periodic intervals, or after a successful reconnection attempt.

C* team is thinking about a specific API for such feature, or virtual tables. I wonder if virtual tables wouldn't be even more flexible, as it doesn't require a protocol change.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/diagnostic/ring/AbstractTokenRingDiagnosticGenerator.java
adutra: I improved the javadocs to be more explicit about this. And how about adding a `Status.getAccuracy()` property? This would help mitigate the problem imo.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/diagnostic/ring/EachQuorumTokenRingDiagnosticGenerator.java
adutra: I find it very readable, but OK to change.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TokenRingDiagnostic.java
adutra: See my previous comment, I wanted only top-level diagnostics as top-level interfaces.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/Status.java
adutra: Changed to make the order irrelevant.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/diagnostic/ring/TokenRingDiagnosticGeneratorFactory.java
adutra: I thought about that but I wonder if this wouldn't prevent most development environments from being able to generate a diagnostic. Most developers usually develop against a 1-node cluster and a keyspace with SimpleStrategy; however the driver default CL is LOCAL_ONE. I wanted these cases to be able to generate a diagnostic in spite of the CL not being compatible.

adutra: In hindsight I think you are right, I'm going to change as you suggested.

adutra:
After giving this some more thought, I propose that we switch to a pessimistic approach for token ring diagnostics. I added a last commit that abort diagnostic generation if the presence of unknown state nodes would compromise the diagnostic's reliability.

Hopefully we're still going to be able to generate diagnostics for the most common case: local consistency level achieved on local dc with NTS. And also on single-dc clsuters.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TokenRingDiagnostic.java
adutra: I'm using `mergeWith` in Spring Boot as well, to merge status from the topology diagnostic and from token ring diagnostic.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TopologyDiagnostic.java
adutra: Actually I'm not a big fan of `NodeGroupDiagnostic` altogether, I think it should be merged into `TopologyDiagnostic`. Will work on it today.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TopologyDiagnostic.java
```diff
@@ -0,0 +1,135 @@
...
+     * <p>Nodes may be in an unknown state if the driver hasn't connected to them at all, and it has
+     * not received any Gossip event indicating their actual state. Most of such nodes should
+     * actually be up.
+     *
+     * @return the number of nodes in this group that are in unknown state.
+     */
+    int getUnknown();
+
+    /**
+     * Returns the status of this node group. The status will be {@link Status#AVAILABLE} if all
```
adutra: After changing topology diagnostics to adopt a pessimistic approach to unknown nodes, I think we should do the same here. I will work on it today.

core/src/main/java/com/datastax/oss/driver/api/core/metadata/diagnostic/TopologyDiagnostic.java
olim7t: > I wonder if virtual tables wouldn't be even more flexible, as it doesn't require a protocol change

Right. Protocol v5 is close to completion so this wouldn't make it in, and we don't want something that needs to wait for v6. We could reuse an existing message, but that feels hackish.
Querying a virtual table is not much harder, and it makes more sense for general use.

adutra:
I think we should close this without merging, any objections? 

I suggest that instead we go with just the notion of topology diagnostics, and in that case, I think we don't need any specific API in the driver, inspecting `Metadata.getNodes()` is enough.

adutra:
As agreed offline, I'm just converting this to draft for now.

PR #1436, opened 2020-04-19
JAVA-2744: Recompute token map when node is added

adutra:
I don't know how to unit-test this because the `Node` interface does not expose the node's raw tokens.

So i went for integration tests, but I had to amend `CcmBridge` and introduce a new `add()` method.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/AddNodeRefresh.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/AddNodeRefresh.java
@@ -40,7 +40,7 @@ public class AddNodeRefresh extends NodesRefresh {
     Node existing = oldNodes.get(newNodeInfo.getHostId());
     if (existing == null) {
       DefaultNode newNode = new DefaultNode(newNodeInfo.getEndPoint(), context);
-      copyInfos(newNodeInfo, newNode, null, context);
+      copyInfos(newNodeInfo, newNode, context);
       Map<UUID, Node> newNodes =
           ImmutableMap.<UUID, Node>builder()
               .putAll(oldNodes)
@@ -54,7 +54,7 @@ public class AddNodeRefresh extends NodesRefresh {
       // an addition, even though the host_id hasn't changed :(
       // Update the existing instance and emit an UP event to trigger a pool reconnection.
       if (!existing.getEndPoint().equals(newNodeInfo.getEndPoint())) {
-        copyInfos(newNodeInfo, ((DefaultNode) existing), null, context);
+        copyInfos(newNodeInfo, ((DefaultNode) existing), context);
         assert newNodeInfo.getBroadcastRpcAddress().isPresent(); // always for peer nodes
         return new Result(
             oldMetadata,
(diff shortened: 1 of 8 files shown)


integration-tests/src/test/java/com/datastax/oss/driver/core/session/RemovedNodeIT.java
```diff
@@ -43,17 +46,23 @@
   public void should_signal_and_destroy_pool_when_node_gets_removed() {
     RemovalListener removalListener = new RemovalListener();
     try (CqlSession session = CqlSession.builder().withNodeStateListener(removalListener).build()) {
-      CCM_RULE.getCcmBridge().nodetool(2, "decommission");
+      assertThat(session.getMetadata().getTokenMap()).isPresent();
```
adutra, line 51: This is to also test the opposite effect, that is, to assert that the token map is correctly shrank when a node is decommissioned.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
```diff
@@ -298,6 +298,15 @@ public void stop(int n) {
     execute("node" + n, "stop");
   }
 
+  public void add(int n, String dc) {
+    execute("add", "-i",  ipPrefix + n, "-d", dc, "node" + n);
+    start(n);
+  }
+
+  public void decommission(int n) {
```
adutra, line 306: Added this method just to have the symmetry add/decommission.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/AddNodeRefresh.java
```diff
@@ -54,7 +54,7 @@ public Result compute(
       // an addition, even though the host_id hasn't changed :(
       // Update the existing instance and emit an UP event to trigger a pool reconnection.
       if (!existing.getEndPoint().equals(newNodeInfo.getEndPoint())) {
-        copyInfos(newNodeInfo, ((DefaultNode) existing), null, context);
```
tomekl007, line 57: Was the problem that we were passing null and because of that there was no recalculation of this condition:
https://github.com/datastax/java-driver/pull/1436/files#diff-7a875990776d2f9525a97a8561fa1cfaL61
?

adutra: Yes correct, although when the node is new, you hit line 43, not 57. But in both cases, `newNodeInfo` contains the new node's tokens, but that information wasn't being copied to `existing` or `newNode`.

olim7t:
Yes, this is definitely a bug. Hard-coding the factory to null in `AddNodeRefresh` makes no sense. I think there might be other cases where it was intended to handle a case where the factory could not be determined in a previous refresh (unexpected error or token map disabled), but becomes available later; but there are better ways to do that, I'll check and add a commit if needed.

OK to remove the parameter.

olim7t:
> there might be other cases where it was intended to handle a case where the factory could not be determined in a previous refresh

In fact that doesn't even apply. If we were never able to determine the token factory before, then the token map is empty, so it will be fully recomputed the next time, regardless of the tokens.

olim7t (approved):
LGTM.
I've extracted the `CcmBridge` change as a separate commit (to move first when we rebase). There was also a small issue in the tests.

PR #1437, opened 2020-04-21
JAVA-2704: Remove protocol v5 beta status

nastra:
> We also need to investigate whether or not a connection with v5 is possible if there is a mismatch between driver and server regarding the beta status of this protocol. If a connection is not possible, we should also instruct users to upgrade to this driver version if they want to use protocol v5 with C* 4.0.

I haven't done this part yet. I would suggest to wait until `V5` is officially out of beta in C* to decide what to do here.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/DefaultProtocolVersion.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/DefaultProtocolVersion.java
@@ -30,14 +30,17 @@ public enum DefaultProtocolVersion implements ProtocolVersion {
   /** Version 4, supported by Cassandra 2.2 and above. */
   V4(ProtocolConstants.Version.V4, false),
 
+  /** Version 5, supported by Cassandra 4.0 and above. */
+  V5(ProtocolConstants.Version.V5, false),
+
   /**
-   * Version 5, currently supported as a beta preview in Cassandra 3.10 and above.
+   * Version 6, currently supported as a beta preview in Cassandra 4.0 and above.
    *
    * <p>Do not use this in production.
    *
    * @see ProtocolVersion#isBeta()
    */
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/DefaultProtocolVersion.java
```diff
@@ -30,14 +30,17 @@
...
   /**
-   * Version 5, currently supported as a beta preview in Cassandra 3.10 and above.
+   * Version 6, currently supported as a beta preview in Cassandra 4.0 and above.
    *
    * <p>Do not use this in production.
    *
    * @see ProtocolVersion#isBeta()
    */
-  V5(ProtocolConstants.Version.V5, true),
+  V6(ProtocolConstants.Version.V5, true),
```
adutra: `ProtocolConstants.Version` should be `V6`, you will probably have to add it to `native-protocol`.

nastra: added in https://github.com/datastax/native-protocol/pull/35

adutra:
Rebased and pushed a few more commits. Let's see how CI goes.

adutra:
CI is failing against 4.0-beta4 due to https://issues.apache.org/jira/browse/CASSANDRA-14973.

nastra:
@adutra does somebody from your team have time to look at https://issues.apache.org/jira/browse/CASSANDRA-14973?focusedCommentId=17289022&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17289022 ?

adutra:
CI failures are expected and will be fixed by JAVA-2925. Thank you @nastra !

PR #1438, opened 2020-04-21
JAVA-2702: Transient Replication Support for Cassandra 4.0 (3.x)

emerkle826:
This is almost a straight prop of JAVA-2105 (from the 4.x branch)

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/ReplicationFactor.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/ReplicationFactor.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.driver.core;
+
+// This class is a subset of server version at org.apache.cassandra.locator.ReplicationFactor
+
+import com.google.common.base.Preconditions;
+
+class ReplicationFactor {
+  private final int allReplicas;
+  private final int fullReplicas;
+  private final int transientReplicas;
+
+  ReplicationFactor(int allReplicas, int transientReplicas) {
+    this.allReplicas = allReplicas;
+    this.transientReplicas = transientReplicas;
+    this.fullReplicas = allReplicas - transientReplicas;
+  }
+
+  ReplicationFactor(int allReplicas) {
+    this(allReplicas, 0);
+  }
+
+  int fullReplicas() {
+    return fullReplicas;
+  }
+
+  int transientReplicas() {
+    return transientReplicas;
+  }
+
+  boolean hasTransientReplicas() {
+    return transientReplicas > 0;
+  }
+
+  static ReplicationFactor fromString(String s) {
+    if (s.contains("/")) {
+
+      int slash = s.indexOf('/');
+      String allPart = s.substring(0, slash);
+      String transientPart = s.substring(slash + 1);
+      Preconditions.checkArgument(
+          allPart != null && transientPart != null,
+          "Replication factor format is <replicas> or <replicas>/<transient>");
+      return new ReplicationFactor(Integer.parseInt(allPart), Integer.parseInt(transientPart));
+    } else {
+      return new ReplicationFactor(Integer.parseInt(s), 0);
+    }
+  }
+
+  @Override
+  public String toString() {
+    return allReplicas + (hasTransientReplicas() ? "/" + transientReplicas() : "");
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (!(o instanceof ReplicationFactor)) {
+      return false;
+    }
+    ReplicationFactor that = (ReplicationFactor) o;
+    return allReplicas == that.allReplicas && transientReplicas == that.transientReplicas;
+  }
+
+  @Override
+  public int hashCode() {
+    return allReplicas ^ transientReplicas;
+  }
+}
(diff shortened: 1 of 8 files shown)


driver-core/src/test/java/com/datastax/driver/core/ReplicationFactorTest.java
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.driver.core;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.testng.annotations.Test;
+
```
tomekl007, line 21: Is it possible to add an Integration test with a transient replica that runs only on > `OSS 4.0.0` and validates that the replication factor is parsed correctly?

emerkle826: I should be able to do that.

driver-core/src/main/java/com/datastax/driver/core/ReplicationFactor.java
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  static ReplicationFactor fromString(String s) {
+    if (s.contains("/")) {
+
+      int slash = s.indexOf('/');
+      String allPart = s.substring(0, slash);
+      String transientPart = s.substring(slash + 1);
+      Preconditions.checkArgument(
+          allPart != null && transientPart != null,
```
adutra: This is an old version of driver 4's `ReplicationFactor`. This logic can/should be simplified to

```
    if (s.contains("/")) {
      int slash = s.indexOf('/');
      String allPart = s.substring(0, slash);
      String transientPart = s.substring(slash + 1);
      return new ReplicationFactor(Integer.parseInt(allPart), Integer.parseInt(transientPart));
    } else {
      return new ReplicationFactor(Integer.parseInt(s), 0);
    }
```

driver-core/src/main/java/com/datastax/driver/core/ReplicationFactor.java
```diff
@@ -0,0 +1,85 @@
...
+    if (!(o instanceof ReplicationFactor)) {
+      return false;
+    }
+    ReplicationFactor that = (ReplicationFactor) o;
+    return allReplicas == that.allReplicas && transientReplicas == that.transientReplicas;
+  }
+
+  @Override
+  public int hashCode() {
+    return allReplicas ^ transientReplicas;
```
adutra, line 77: That's a fancy hash code, shouldn't we go for the more mainstream ` Objects.hash(allReplicas, fullReplicas)`?

adutra (changes requested):

driver-core/src/main/java/com/datastax/driver/core/ReplicationFactor.java
emerkle826: I did use that at first, but there is a JDK6 check in the compile for 3.x and `Objects.hash()` was introduced in JDK7, so it errors the compile with this:
```
[INFO] --- animal-sniffer-maven-plugin:1.15:check (check-jdk6) @ cassandra-driver-core ---
[INFO] Checking unresolved references to org.codehaus.mojo.signature:java16:1.0
[ERROR] /home/emerkle/workspace/java-driver-3.x/driver-core/src/main/java/com/datastax/driver/core/ReplicationFactor.java:85: Undefined reference: int java.util.Objects.hash(Object[])
```

So I went with the old-school hash computation of XOR'ing the 2 attributes used to construct the object.

driver-core/src/main/java/com/datastax/driver/core/ReplicationFactor.java
emerkle826: I'll update it.

driver-core/src/main/java/com/datastax/driver/core/ReplicationFactor.java
adutra: Ah right, thanks for the explanation.

driver-core/src/test/java/com/datastax/driver/core/ReplicationFactorTest.java
adutra: I think that would be welcome on 4.x branch as well. But you can do this in a follow-up ticket.

emerkle826: Added: https://datastax-oss.atlassian.net/browse/JAVA-2746

driver-core/src/main/java/com/datastax/driver/core/schemabuilder/TableOptions.java
```diff
@@ -1362,4 +1413,38 @@ public String value() {
...
+    }
+
+    @Override
+    public String toString() {
+      return value;
+    }
+  }
+
+  /** Additional Write Policy. Default value is 99p */
+  public static class AdditionalWritePolicyValue {
```
tomekl007, line 1446: Shouldn't we create an enum with possible options for this `AdditionalWritePolicyValue`?
(similar to https://github.com/datastax/java-driver/pull/1438/files/175ee2f219caf1a8a475d8ad005935aa6b93d3ca..62d5b0cad267e3746b316858037fcf540ab8b7b2#diff-89e075b9ba33eed1af4b80ebefe597ccR1419-R1420)

emerkle826: I would, but there are many possible values. This `additional_write_policy` value, from what I can tell, is parsed and treated like `speculative_retry`, to the point that invalid values result in an error message like this:
```
ConfigurationException: Invalid value ALL for option 'speculative_retry'
```
Valid values can be percentiles ('99p', '50p', '32p'), time ('400ms', '200ms'), 'ALWAYS' and 'NEVER', so I don't think it's feasible to create an enum for all values.

That said, we could add more support to `SchemaBuilder` for this and do something similar to what we do for speculative retry:

SchemaBuilder.[noSpeculativeRetry()](https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/schemabuilder/SchemaBuilder.java#L321-L323)
SchemaBuilder.[always()](https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/schemabuilder/SchemaBuilder.java#L331-L333)
SchemaBuilder.[percentile(int percentile)](https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/schemabuilder/SchemaBuilder.java#L341-L347)
SchemaBuilder.[millisecs(int millisecs)](https://github.com/datastax/java-driver/blob/3.x/driver-core/src/main/java/com/datastax/driver/core/schemabuilder/SchemaBuilder.java#L355-L361)

olim7t (approved):

adutra (approved):

PR #1439, opened 2020-04-28
JAVA-2663: Pure Graal support for native functions

absurdfarce:


absurdfarce:
Building and running a version of the keyspace counter app as a native image with Java driver 4.7.0-SNAPSHOT as a dependency I get the following:

```
[mersault@localhost countapp]$ build/native-image/countapp debug
...
23 [s0-admin-0] DEBUG com.datastax.oss.driver.internal.core.util.Reflection  - Building from unqualified name AtomicTimestampGenerator                                            
23 [s0-admin-0] DEBUG com.datastax.oss.driver.internal.core.util.Reflection  - Trying with default package com.datastax.oss.driver.internal.core.time.AtomicTimestampGenerator    23 [s0-admin-0] INFO com.datastax.oss.driver.internal.core.os.Native  - Using Graal-native functions for Graal run time                                                           
23 [s0-admin-0] INFO com.datastax.oss.driver.internal.core.time.Clock  - Using native clock for microsecond precision 
...
PID: 14429
PID (from driver): 14429
```

The first value is obtained from a local call to the Graal C API, the second via a call to Native.getProcessId().

core/src/main/java/com/datastax/oss/driver/internal/core/os/GraalGetpid.java
```diff
@@ -0,0 +1,45 @@
...
+    public List<String> getHeaderFiles() {
+
+      return Collections.unmodifiableList(Arrays.asList("<unistd.h>"));
+    }
+
+    /* TODO: I don't know that these are required... */
+    @Override
+    public List<String> getMacroDefinitions() {
+
+      return Collections.unmodifiableList(Arrays.asList("_GNU_SOURCE", "_LARGEFILE64_SOURCE"));
```
absurdfarce: We explicitly avoid Guava in these methods since this code will be evaluated by Graal analysis at build time... and we don't want to force app authors to include a bunch of Guava classes as build-time requiremenets.

tomekl007: Isn't the shaded guava already included in the build via java-driver dependency?

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -25,8 +25,24 @@
...
+    private static final String GRAAL_BUILDTIME_STATUS = "buildtime";
+    private static final String GRAAL_RUNTIME_STATUS = "runtime";
+
     public Libc load() {
       try {
+        if (isGraalBuildTime()) {
+          LOG.info("Using Graal-native functions for Graal build time");
+          return new GraalLibc();
+        }
+        if (isGraalRunTime()) {
```
tomekl007: I don't understand why do we need to create two conditions. 
If the graal is available in the build time, it is also available in the build time, isn't it?
If so we can coalesce those into one if that checks:  
```
return System.getProperty(GRAAL_STATUS_PROP).equals(    return System.getProperty(GRAAL_STATUS_PROP).equals(GRAAL_BUILDTIME_STATUS) || 
return System.getProperty(GRAAL_STATUS_PROP).equals(GRAAL_RUNTIME_STATUS)
```

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -35,6 +51,14 @@ public Libc load() {
         return new EmptyLibc();
       }
     }
+
+    private boolean isGraalBuildTime() {
```
tomekl007: Is this https://github.com/datastax/java-driver/pull/1439/files#diff-357e4854869b2e21c38b1b437f11095aR130-R134 dependency always present if one of those settings evaluates to true?

pom.xml
```diff
@@ -377,6 +378,11 @@
         <artifactId>wiremock</artifactId>
         <version>2.25.0</version>
       </dependency>
+      <dependency>
+        <groupId>org.graalvm.sdk</groupId>
+        <artifactId>graal-sdk</artifactId>
+        <version>${graal-sdk.version}</version>
+      </dependency>
```
olim7t, line 384: nit: extracting the version to a property was done for dependencies that are used in OSGi tests (and those are going away soon anyway).

When only one dependency uses the version, we can keep things simple and inline it.

core/src/main/java/com/datastax/oss/driver/internal/core/os/GraalLibc.java
```diff
@@ -0,0 +1,38 @@
...
+  @Override
+  public boolean available() {
+    return true;
+  }
+
+  /* Substrate includes a substitution for Linux + Darwin which redefines System.nanoTime() to use
+   * gettimeofday() (unless platform-specific higher-res clocks are available, which is even better). */
+  @Override
+  public Optional<Long> gettimeofday() {
+    return Optional.of(Math.round(System.nanoTime() / 1_000d));
```
olim7t: `gettimeofday()` has microsecond resolution. If `nanoTime()` uses it, it likely returns a multiple of 1000. So an integer division should be sufficient.

core/src/main/java/com/datastax/oss/driver/internal/core/os/GraalGetpid.java
```diff
@@ -0,0 +1,45 @@
...
+
+@CContext(GraalGetpid.Directives.class)
+public class GraalGetpid {
+
+  static class Directives implements CContext.Directives {
+
+    @Override
+    public List<String> getHeaderFiles() {
+
+      return Collections.unmodifiableList(Arrays.asList("<unistd.h>"));
```
olim7t: I get an IDE warning suggesting this:
```suggestion
      return Collections.unmodifiableList(Collections.singletonList("<unistd.h>"));
```

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -25,8 +25,24 @@
 
   private static class LibcLoader {
 
+    /* These values come from Graal's imageinfo API which aims to offer the ability to detect
+    * when we're in the Graal build/run time via system props.  The maintainers of Graal have
+    * agreed that this API will not change over time.  We reference these props as literals
+    * to avoid introducing a dependency on Graal code for non-Graal users here. */
```
olim7t:
```suggestion
     * when we're in the Graal build/run time via system props.  The maintainers of Graal have
     * agreed that this API will not change over time.  We reference these props as literals
     * to avoid introducing a dependency on Graal code for non-Graal users here. */
```
To make the formatter happy 🙂

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -35,6 +51,14 @@ public Libc load() {
         return new EmptyLibc();
       }
     }
+
+    private boolean isGraalBuildTime() {
+      return System.getProperty(GRAAL_STATUS_PROP).equals(GRAAL_BUILDTIME_STATUS);
```
olim7t: This throws an NPE outside of Graal, `getProperty` returns null if the property does not exist.

olim7t (changes requested):
Small bug in the property detection, but apart from that I agree with the principles.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
olim7t: Yeah I tend to agree. Maybe if we needed different implementations at some point in the future, but that sounds far-fetched and we could always make the change later...

core/src/main/java/com/datastax/oss/driver/internal/core/os/GraalGetpid.java
olim7t: It is, but we're just creating a simple unmodifiable list here, either approach will do.

core/src/main/java/com/datastax/oss/driver/internal/core/os/GraalGetpid.java
absurdfarce: Hey, I'm happy to do what I can to make the IDE happy. :)

core/src/main/java/com/datastax/oss/driver/internal/core/os/GraalGetpid.java
absurdfarce: It's not a question of whether the classes are available @tomekl007 ... you're correct, they certainly are in the classpath.  The point is that this code will be eval'd at build-time (since we have to build C source off of it).  And if we include the Guava classes here you'll get errors in the native image build operation saying that some Guava classes were unexpectedly initialized at build-time, which you then have to work around by specifying the --initialize-at-build-time flag for each of those classes.  I'm trying to avoid all of that by keeping Guava out of this section of the code base.

core/src/main/java/com/datastax/oss/driver/internal/core/os/GraalLibc.java
absurdfarce: gettimeofday() does indeed have microsecond resolution, but note the comment above about trying to use the higher resolution Linux clocks if they are available.  That's the clock_gettime() call in https://github.com/oracle/graal/blob/master/substratevm/src/com.oracle.svm.core.posix/src/com/oracle/svm/core/posix/linux/LinuxSubstitutions.java#L47-L51.  If that succeeds then we'd need rounding to get to microseconds.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: Like I said earlier, man... I'm all about making the tools happy. ;)

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: :facepalm: Fixed, good catch

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: Yes, that's correct.  It has to be the case since at build-time the annotations are eval'd for C generation and at run-time the necessary infrastructure for the C calls has to be in place.

pom.xml
absurdfarce: My working assumption was that we kept the versions in props like this to allow users to override them at build-time.  It seemed like it _might_ be desirable to support something like this for the Graal functionality... although that's just intuition with no real data behind it.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: My original thinking was that we'd need two checks to address the following cases:

* At build-time we want Graal's build-time analysis logic to follow that path rather than worry about the jnr path
* At run-time we wan to select the impl in order to access the underlying System/libc functionality

I don't think the run-time point above is terribly controversial... but as I reflect on it more I think the build-time requirement is bogus.  Built into my logic above is a belief that the build-time analysis done by the native image build is aware of these system properties and can detect branching logic based on them.  This is... somewhat silly.  A much more plausible explanation is that the analysis process just considers _both_ paths (realizing that the actual selection will be made at run-time).  As a result there's no real reason to worry about build-time at all.

Local testing seems to confirm this hypothesis, so I'll remove the build-time constraint all together.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
```diff
@@ -25,8 +25,19 @@
...
   private static class LibcLoader {
 
+    /* These values come from Graal's imageinfo API which aims to offer the ability to detect
+     * when we're in the Graal build/run time via system props.  The maintainers of Graal have
+     * agreed that this API will not change over time.  We reference these props as literals
+     * to avoid introducing a dependency on Graal code for non-Graal users here. */
+    private static final String GRAAL_STATUS_PROP = "org.graalvm.nativeimage.imagecode";
+    private static final String GRAAL_RUNTIME_STATUS = "runtime";
+
     public Libc load() {
```
absurdfarce, line 36: I was considering adding a config option that would allow the user to disable this functionality even for Graal VM apps if they desired.  A user might want to do something like this if they don't want to install the LLVM toolchain for some reason (I'm not sure why that would be an issue but let's assume it is) or they don't want to rely on a newer feature.

Is this worthwhile?  I can see pretty clear arguments for and against... and the more I think about it the more it seems like it might not be worth it.

core/src/main/java/com/datastax/oss/driver/internal/core/os/GraalGetpid.java
```diff
@@ -0,0 +1,45 @@
...
+
+  static class Directives implements CContext.Directives {
+
+    @Override
+    public List<String> getHeaderFiles() {
+
+      return Collections.unmodifiableList(Collections.singletonList("<unistd.h>"));
+    }
+
+    /* TODO: I don't know that these are required... */
```
absurdfarce: After pondering this for a few minutes it became pretty clear that they shouldn't be... it's not like you have to specify these when writing apps in C and it might be weird in the Darwin case.  Regardless things seem to work well without them so I'm removing them.

pom.xml
olim7t: > My working assumption was that we kept the versions in props like this to allow users to override them at build-time

No, it's mostly for our internal use (apart from OSGi, one other use was testing against different versions of Guava on 3.x, but that's not a thing anymore).
I don't think many users build from source, and if they start changing dependency versions it's almost already a fork anyway, we don't need to have built-in support for this.

No strong opinion here, but as a heads up I might do a pass on all those properties and remove unnecessary ones as part of #1424.

core/src/main/java/com/datastax/oss/driver/internal/core/os/GraalLibc.java
olim7t: I had missed that, makes sense then 👍

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
olim7t: Yeah I agree the use case sounds contrived... Let's keep it like this for now, if someone needs it they will ask for it.

olim7t (approved):

absurdfarce:
I did some additional testing to confirm that this code works for native images generated using Graal 19.2.1, 19.3.1 and 20.0.  These represent the most recent of the three most recent minor versions so it looks like we've avoided any version-specific issues.

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
tomekl007: I think you need to leave the `build-time`. Without it, the quarkus app that is using JNR will not compile with incomplete classpath errors.

tomekl007 (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java
absurdfarce: It shouldn't result in incomplete classpath errors, but you do raise an interesting point @tomekl007 ... it's quite possible my sample app isn't doing build-time initialization of any of the native classes.  Which means we might not be exercising everything we need for testing here.

Yup, looks like that might be the case.  Quarkus does apparently trigger build-time init of these classes (via Uuids) which leads us to the following:

```
Error: Class initialization of com.datastax.oss.driver.api.core.uuid.Uuids failed. Use the option --initialize-at-run-time=com.datastax.oss.driver.api.core.uuid.Uuids to explicit
ly request delayed initialization of this class.                                                                                                    
Detailed message:                                                      
                                                                                                
com.oracle.svm.core.util.UserError$UserException: Class initialization of com.datastax.oss.driver.api.core.uuid.Uuids failed. Use the option --initialize-at-run-time=com.datastax
.oss.driver.api.core.uuid.Uuids to explicitly request delayed initialization of this class. 
Detailed message:                                                                                  
                                                                                              
        at com.oracle.svm.core.util.UserError.abort(UserError.java:75)                     
        at com.oracle.svm.hosted.FallbackFeature.reportAsFallback(FallbackFeature.java:221)                                                                                               at com.oracle.svm.hosted.NativeImageGenerator.runPointsToAnalysis(NativeImageGenerator.java:736)
        at com.oracle.svm.hosted.NativeImageGenerator.doRun(NativeImageGenerator.java:530)                                                                                       
        at com.oracle.svm.hosted.NativeImageGenerator.lambda$run$0(NativeImageGenerator.java:445)                                                                                
        at java.util.concurrent.ForkJoinTask$AdaptedRunnableAction.exec(ForkJoinTask.java:1386)                   
        at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)                                                                                                       
        at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1056)                                                                                           
        at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)               
        at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)                                                                                          
Caused by: com.oracle.graal.pointsto.constraints.UnsupportedFeatureException: Class initialization of com.datastax.oss.driver.api.core.uuid.Uuids failed. Use the option --initial
ize-at-run-time=com.datastax.oss.driver.api.core.uuid.Uuids to explicitly request delayed initialization of this class.                                                          
Detailed message:                                                                                                                                                                
                                                                                                                                                                                 
        at com.oracle.graal.pointsto.constraints.UnsupportedFeatures.report(UnsupportedFeatures.java:126)                                                                        
        at com.oracle.svm.hosted.NativeImageGenerator.runPointsToAnalysis(NativeImageGenerator.java:733)
        ... 7 more                                                        
Caused by: java.lang.UnsatisfiedLinkError: com.datastax.oss.driver.internal.core.os.GraalGetpid.getpid()I
        at com.datastax.oss.driver.internal.core.os.GraalGetpid.getpid(Native Method)                                                                                            
        at com.datastax.oss.driver.internal.core.os.GraalLibc.getpid(GraalLibc.java:36)                                                                                          
        at com.datastax.oss.driver.internal.core.os.Native.getProcessId(Native.java:82)  
        at com.datastax.oss.driver.api.core.uuid.Uuids.getProcessPiece(Uuids.java:193)        
        at com.datastax.oss.driver.api.core.uuid.Uuids.makeNode(Uuids.java:163)             
        at com.datastax.oss.driver.api.core.uuid.Uuids.makeClockSeqAndNode(Uuids.java:226)                                              
        at com.datastax.oss.driver.api.core.uuid.Uuids.<clinit>(Uuids.java:116)      
        at sun.misc.Unsafe.ensureClassInitialized(Native Method)            
...
```

Looks like Uuids attempt to get at the PID is leading to a build-time attempt to access the native code which likely isn't available yet.

I've got an idea for how to work around this, need to do some more testing.

core/src/main/java/com/datastax/oss/driver/api/core/uuid/Uuids.java
```diff
@@ -19,6 +19,7 @@
 import com.datastax.oss.driver.internal.core.util.Loggers;
```
absurdfarce: Changes in his file are an attempt to handle the UnsatisfiedLinkErrors reported elsewhere on this PR.  By deferring initialization of state necessary to support time-based UUIDs until run-time we prevent the build-time analysis from detecting any attempt to use the linked LLVM native code.  The drawback here is that we've replaced a simple field access with a method call to a Supplier plus some (at the moment manual) unboxing into a primitive, so there is some slight performance degradation here.

@olim7t are you okay with this?

absurdfarce: Additional note: I originally tried to implement this via some Graal substitutions for the underlying methods in Uuids, but the complexity of that approach escalated quickly.  You'd want to delete the field performing static initialization and inject a new field which uses the supplier-based approach here, but you can't inject static fields.  You also want to preserve access to the underlying functionality (to use in your supplier) without copying code... which proves to be difficult/impossible to do.

tomekl007: I think it is possible to substitute static fields, see `@Alias` annotation and it's usages in the `quarkus` repo

absurdfarce:
In this case it's not a problem of the upfront static analysis @tomekl007.  Here the Native class is actually getting initialized at build-time which causes all the settings for initial values to actually execute when the native image is built.  That init in turn causes an attempt to access the generated native library code which either hasn't been generated yet or isn't otherwise accessible.... thus the UnsatisfiedLinkError

The key is that we don't actually call getpid() at build-time.  And the Supplier accomplishes that goal.

I confirmed this by running this version of the code against the quickstart app.  Without this change I get the stack trace above, but with it everything seems to work normally.

core/src/main/java/com/datastax/oss/driver/internal/core/os/GraalLibc.java
adutra: 1. What happens if Graal changes its substitution for `System.nanoTime()` to use something that is not a wall-clock time? It looks to me that we are relying on implementation details that could change later.
2. What happens for platforms that do not have a similar substitution, Windows for example? Is this method going to return the actual call to `System.nanoTime()`?

absurdfarce: Both legit points @adutra.  We've talked about this offline but I'll summarize those conversations here for the record.

Re: relying on Graal's substitution impl... I wrestled with this as well, and I don't love the idea of relying on a third party like this.  In practice, though, the logic in the substitution actually _improves on_ just relying on gettimeofday() and what we'd really be worried about here is if the Graal team made a change in the future which somehow _lessened_ the accuracy of their substitution.  This is certainly not impossible, but it seems fairly unlikely.

What we have right now seems "good enough" for an initial iteration.  We discussed filing a follow-up ticket to explore this in the future in order to see if it's worthwhile to remove this "dependency" [1] on the substitution.  I've filed JAVA-2768 to consider this change.

Re: platforms without the substitution... it seems to make sense to make the Native.isXAvailable() checks return a result based on platform in order to account for this.  If the platform is observed to be LInux or OS X we return true, otherwise false.  This approach has the advantage of playing nicely with how the rest of the driver codebase uses Native.

[1] "dependency" is (sort of) the wrong phrase here.  We're relying on System.nanoTime() doing the right thing, not any additional dependency.  Hypothetically any JDK version could also change this behaviour in it's internals... which in turn actually supports the idea of calling getimeofday() directly in order to avoid reliance on volatile parts of the platform.

core/src/main/java/com/datastax/oss/driver/api/core/uuid/Uuids.java
olim7t: I'm okay on the principle, but I agree that introducing a supplier and unboxing is unfortunate. If all we need is execution on first actual use, I think the lazy initialization holder idiom will work as well:
```java
public final class Uuids {
  private static class ClockSeqAndNodeHolder {
    private static final long CLOCK_SEQ_AND_NODE = makeClockSeqAndNode();

    // not strictly necessary, but it feels cleaner to move this method into the holder as well:
    private static long makeClockSeqAndNode() { ... }
  }
  ...
}
```

absurdfarce: Weirdly that actually _doesn't_ work in this case @olim7t.  That pattern relies on the lazy initialization of classes (i.e. not initialized until loaded), but the problem here is that the class in question (api.core.uuid.Uuids) is being initialized at build time.  As a result the loader code is exec'd immediately, voiding any benefit.

To validate, I implemented the lazy init holder pattern via the following diff: 

```diff
diff --git a/core/src/main/java/com/datastax/oss/driver/api/core/uuid/Uuids.java b/core/src/main/java/com/datastax/oss/driver/api/core/uuid/Uuids.java
index 0d633df15..d95146b23 100644
--- a/core/src/main/java/com/datastax/oss/driver/api/core/uuid/Uuids.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/uuid/Uuids.java
@@ -19,7 +19,6 @@ import com.datastax.oss.driver.internal.core.os.Native;
 import com.datastax.oss.driver.internal.core.util.Loggers;
 import com.datastax.oss.driver.shaded.guava.common.annotations.VisibleForTesting;
 import com.datastax.oss.driver.shaded.guava.common.base.Charsets;
-import com.datastax.oss.driver.shaded.guava.common.base.Suppliers;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.lang.management.ManagementFactory;
 import java.net.InetAddress;
@@ -40,7 +39,6 @@ import java.util.Set;
 import java.util.TimeZone;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.function.Supplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -115,8 +113,17 @@ public final class Uuids {
   private Uuids() {}
 
   private static final long START_EPOCH = makeEpoch();
-  private static final Supplier<Long> CLOCK_SEQ_AND_NODE =
-      Suppliers.memoize(Uuids::makeClockSeqAndNode);
+
+  private static class ClockSeqAndNodeContainer {
+
+    private static final long VALUE = makeClockSeqAndNode();
+
+    private synchronized long get() {
+      return VALUE;
+    }
+  }
+
+  private static final ClockSeqAndNodeContainer CLOCK_SEQ_AND_NODE = new ClockSeqAndNodeContainer();
 
   // The min and max possible lsb for a UUID.
   //
@@ -440,7 +447,7 @@ public final class Uuids {
    */
   @NonNull
   public static UUID timeBased() {
-    return new UUID(makeMsb(getCurrentTimestamp()), CLOCK_SEQ_AND_NODE.get().longValue());
+    return new UUID(makeMsb(getCurrentTimestamp()), CLOCK_SEQ_AND_NODE.get());
   }
 
   /**
```

This worked exactly as expected when running the app directly via a JVM (no Graal):

```
817 [s0-admin-0] DEBUG com.datastax.oss.driver.internal.core.util.Reflection  - Trying with default package com.datastax.oss.driver.internal.core.time.AtomicTimestampGenerator
1337 [s0-admin-0] INFO com.datastax.oss.driver.internal.core.time.Clock  - Using native clock for microsecond precision
...
1631 [s0-io-2] DEBUG com.datastax.oss.driver.internal.core.channel.ProtocolInitHandler  - [s0|control|id: 0x5da58562, L:/127.0.0.1:44588 - R:localhost/127.0.0.1:9042] step OPTION
S received response opcode=SUPPORTED                                                                                                                                              
1657 [s0-io-2] INFO com.datastax.oss.driver.api.core.uuid.Uuids  - PID obtained through native call to getpid(): 23083                                                            
1662 [s0-io-2] DEBUG com.datastax.oss.driver.internal.core.channel.InFlightHandler  - [s0|control|id: 0x5da58562, L:/127.0.0.1:44588 - R:localhost/127.0.0.1:9042] Got last respon
se on in-flight stream id 0, completing and releasing  
```

and when building and running my local keyspace counting app via Graal (no Quarkus):

```
11 [s0-admin-0] DEBUG com.datastax.oss.driver.internal.core.util.Reflection  - Trying with default package com.datastax.oss.driver.internal.core.time.AtomicTimestampGenerator    
11 [s0-admin-0] INFO com.datastax.oss.driver.internal.core.os.Native  - Using Graal-specific native functions                                                                     
11 [s0-admin-0] INFO com.datastax.oss.driver.internal.core.time.Clock  - Using native clock for microsecond precision   
...
46 [s0-io-2] DEBUG com.datastax.oss.driver.internal.core.channel.ProtocolInitHandler  - [s0|control|id: 0xcd094fd9, L:/127.0.0.1:44638 - R:localhost/127.0.0.1:9042] step OPTIONS
received response opcode=SUPPORTED
47 [s0-io-2] INFO com.datastax.oss.driver.api.core.uuid.Uuids  - PID obtained through native call to getpid(): 23957
48 [s0-io-2] DEBUG com.datastax.oss.driver.internal.core.channel.InFlightHandler  - [s0|control|id: 0xcd094fd9, L:/127.0.0.1:44638 - R:localhost/127.0.0.1:9042] Got last response
 on in-flight stream id 0, completing and releasing
```

Note that in each of these cases UUID initialization is clearly deferred as expected.  Things fall apart when I try to build a Quarkus app, though.  We already know that Quarkus is somehow triggering build-time init of the Uuids class, and that eventually leads me here:

```
[INFO] [io.quarkus.deployment.pkg.steps.NativeImageBuildStep] /work/local/graal-19.3.1/bin/native-image -J-Djava.util.logging.manager=org.jboss.logmanager.LogManager -J-Dsun.nio.
ch.maxUpdateArraySize=100 -J-Dvertx.logger-delegate-factory-class-name=io.quarkus.vertx.core.runtime.VertxLogDelegateFactory -J-Dvertx.disableDnsResolver=true -J-Dio.netty.leakDe
tection.level=DISABLED -J-Dio.netty.allocator.maxOrder=1 -J-Duser.language=en -J-Dfile.encoding=UTF-8 --initialize-at-build-time= -H:InitialCollectionPolicy=com.oracle.svm.core.g
enscavenge.CollectionPolicy$BySpaceAndTime -H:+JNI -jar quickstart-1.0.0-SNAPSHOT-runner.jar -H:FallbackThreshold=0 -H:+ReportExceptionStackTraces -H:-AddAllCharsets -H:-IncludeA
llTimeZones -H:EnableURLProtocols=http -H:NativeLinkerOption=-no-pie --no-server -H:-UseServiceLoaderFeature -H:+StackTrace quickstart-1.0.0-SNAPSHOT-runner                     
[quickstart-1.0.0-SNAPSHOT-runner:24275]    classlist:  24,108.06 ms                                                                                                             
[quickstart-1.0.0-SNAPSHOT-runner:24275]        (cap):   1,713.83 ms                                                                                                             
[quickstart-1.0.0-SNAPSHOT-runner:24275]        setup:   5,122.67 ms
10:19:30,365 INFO  [com.dat.oss.dri.int.cor.DefaultMavenCoordinates] DataStax Java driver for Apache Cassandra(R) (com.datastax.oss:java-driver-core) version 4.7.0-SNAPSHOT
10:19:33,420 INFO  [org.jbo.threads] JBoss Threads version 3.0.1.Final              
10:19:58,203 INFO  [com.dat.oss.dri.int.cor.os.Native] Using Graal-specific native functions                                                                                     
10:19:58,207 INFO  [com.dat.oss.dri.api.cor.uui.Uuids] PID obtained through JMX: 24275 
[quickstart-1.0.0-SNAPSHOT-runner:26961]   (typeflow):  60,717.82 ms
```

Note the initialization of Uuids at build-time here.  This disappears entirely when I rollback the changes in the diff and go back to the Supplier-based impl currently in the PR:

```
[INFO] [io.quarkus.deployment.pkg.steps.NativeImageBuildStep] /work/local/graal-19.3.1/bin/native-image -J-Dsun.nio.ch.maxUpdateArraySize=100 -J-Djava.util.logging.manager=org.jb
oss.logmanager.LogManager -J-Dio.netty.leakDetection.level=DISABLED -J-Dio.netty.allocator.maxOrder=1 -J-Dvertx.logger-delegate-factory-class-name=io.quarkus.vertx.core.runtime.VertxLogDelegateFactory -J-Dvertx.disableDnsResolver=true -J-Duser.language=en -J-Dfile.encoding=UTF-8 --initialize-at-build-time= -H:InitialCollectionPolicy=com.oracle.svm.core.genscavenge.CollectionPolicy$BySpaceAndTime -H:+JNI -jar quickstart-1.0.0-SNAPSHOT-runner.jar -H:FallbackThreshold=0 -H:+ReportExceptionStackTraces -H:-AddAllCharsets -H:-IncludeAllTimeZones -H:EnableURLProtocols=http -H:NativeLinkerOption=-no-pie --no-server -H:-UseServiceLoaderFeature -H:+StackTrace quickstart-1.0.0-SNAPSHOT-runner
[quickstart-1.0.0-SNAPSHOT-runner:26961]    classlist:  20,907.58 ms
[quickstart-1.0.0-SNAPSHOT-runner:26961]        (cap):   1,187.25 ms
[quickstart-1.0.0-SNAPSHOT-runner:26961]        setup:   3,987.47 ms
10:42:59,570 INFO  [com.dat.oss.dri.int.cor.DefaultMavenCoordinates] DataStax Java driver for Apache Cassandra(R) (com.datastax.oss:java-driver-core) version 4.7.0-SNAPSHOT
10:43:01,875 INFO  [org.jbo.threads] JBoss Threads version 3.0.1.Final
10:43:32,368 INFO  [com.dat.oss.dri.int.cor.os.Native] Using Graal-specific native functions
[quickstart-1.0.0-SNAPSHOT-runner:26961]   (typeflow):  60,717.82 ms
```

Note that lack of Uuids initialization here

It _is_ true that the Quarkus build case does succeed now rather than throwing the UnsatisfiedLinkError as it did before.  That's because the recent changes to this PR cause GraalLibc.available() to return true iff it's able to execute a getpid() and gettimeofday() native call without errors or exceptions.  Those calls fail in the Quarkus build case (due to the UnsatisfiedLinkError discussed above) so Native is "unavailable" in this context.  The initialization logic of Uuids knows to try JMX next in that case and that succeeds.  So there's a semi-plausible argument that we just live with this constraint and move on.

Note, however, that there's a built-in constraint here; we'd now be saying that _no_ logic which might get executed build-time could use os.Native without some fallback option (like JMX for getting the PID in this case).  That may be okay but it is worth highlighting.

absurdfarce: Follow-up: I'm pretty sure I can adapt the memoized Supplier code to avoid the Supplier dependency and (un)boxing issues... working on that now.

absurdfarce:
I believe all the outstanding issues have been addressed now so if everybody would like to take another pass we can (hopefully) get this sucker in soon.

olim7t (approved):
LGTM
Last request, could you amend the comment above `ClockSeqAndNodeContainer` to explain that this is for Graal? Just for the sake of future readers who might not have the whole picture in mind.

adutra (approved):

PR #1442, opened 2020-05-16
update dependency on io.dropwizard.metrics from 3.2.2 to latest 4.1.7…

iandebeer:
… and include io.dropwizard.metrics.jmx to fix backward compatibility for com.codahale.metrics.JmxReporter -> com.codahale.metrics.jmx.JmxReporter

### DIFF
--- a/driver-core/pom.xml
+++ b/driver-core/pom.xml
@@ -48,6 +48,11 @@
             <groupId>io.dropwizard.metrics</groupId>
             <artifactId>metrics-core</artifactId>
         </dependency>
+        
+        <dependency>
+            <groupId>io.dropwizard.metrics</groupId>
+            <artifactId>metrics-jmx</artifactId>
+        </dependency>
 
         <dependency>
             <groupId>org.slf4j</groupId>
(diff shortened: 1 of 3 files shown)


adutra:
Hi, thanks for your contribution!

Unfortunately, driver 3.x must remain compatible with Java 6 and 7, however Dropwizard 4 is only compatible with Java 8+.

When I attempt to compile your branch with Java 6, I get:

```
[ERROR] driver-core/src/main/java/com/datastax/driver/core/Metrics.java:[18,28] cannot access com.codahale.metrics.Counter
bad class file: com/codahale/metrics/Counter.class(com/codahale/metrics:Counter.class)
class file has wrong version 52.0, should be 50.0
Please remove or make sure it appears in the correct subdirectory of the classpath.
```

For this reason, it's not going to be possible to accept your contribution.

You have a few options:

1. Do this change in your own code. You simply need to override the Dropwizard version and use your own.
2. Have you considered upgrading to driver 4.6.1? This is our last version, and only versions in the 4.x line are actively maintained.

Good luck!

PR #1444, opened 2020-05-20
JAVA-2772: Add protocol v5 checksumming to driver 3

olim7t:
The new type is provisionally called `Segment` to avoid name clashes. I'll probably align with server-side naming when a consensus is reached. Here are the new components for low-level encoding:
- `SegmentBuilder`: accumulates messages and emits segments as they fill up. This is called from the flusher.
- `SegmentCodec`: low-level byte encoding. Called on the Netty pipeline by `SegmentToBytesEncoder`, a simple wrapper.
- `BytesToSegmentDecoder` and `SegmentToFrameDecoder`: two new handlers that `Frame.Decoder` for incoming responses. After that, decoded messages can be fed to the existing components.

To do:
- [x] plug all that with the rest of the driver, and flip the switch at the right moment.
- [x] revisit the encoding of the response to the first `STARTUP` message, see [this comment](https://issues.apache.org/jira/browse/CASSANDRA-15299?focusedCommentId=17112780&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17112780). It's actually the server code that has to change, the driver already implements what's described in the comment.
- [x] double-check that Netty buffers are released properly, both on the outgoing and incoming path.
- [x] check how CRC errors thrown from the decoders are surfaced and handled. Our goal is to drop the connection.
- [ ] Run a long test with the Netty buffer leak detector enabled (will wait for the server PR to be more mature).

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/BytesToSegmentDecoder.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/BytesToSegmentDecoder.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.driver.core;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.LengthFieldBasedFrameDecoder;
+import java.nio.ByteOrder;
+
+/**
+ * Decodes {@link Segment}s from a stream of bytes.
+ *
+ * <p>This works like a regular length-field-based decoder, but we override {@link
+ * #getUnadjustedFrameLength} to handle two peculiarities: the length is encoded on 17 bits, and we
+ * also want to check the header CRC before we use it. So we parse the whole segment header ahead of
+ * time, and store it until we're ready to build the segment.
+ */
+class BytesToSegmentDecoder extends LengthFieldBasedFrameDecoder {
+
+  private final SegmentCodec segmentCodec;
+  private SegmentCodec.Header header;
+
+  BytesToSegmentDecoder(SegmentCodec segmentCodec) {
+    super(
+        // max length (Netty wants this to be the overall length including everything):
+        segmentCodec.headerLength()
+            + SegmentCodec.CRC24_LENGTH
+            + Segment.MAX_PAYLOAD_LENGTH
+            + SegmentCodec.CRC32_LENGTH,
+        // offset and size of the "length" field: that's the whole header
+        0,
+        segmentCodec.headerLength() + SegmentCodec.CRC24_LENGTH,
+        // length adjustment: add the trailing CRC to the declared length
+        SegmentCodec.CRC32_LENGTH,
+        // bytes to skip: the header (we've already parsed it while reading the length)
+        segmentCodec.headerLength() + SegmentCodec.CRC24_LENGTH);
+    this.segmentCodec = segmentCodec;
+  }
+
+  @Override
+  protected Object decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception {
+    try {
+      ByteBuf payloadAndCrc = (ByteBuf) super.decode(ctx, in);
+      if (payloadAndCrc == null) {
+        return null;
+      } else {
+        assert header != null;
+        Segment segment = segmentCodec.decode(header, payloadAndCrc);
+        header = null;
+        return segment;
+      }
+    } catch (Exception e) {
+      // Don't hold on to a stale header if we failed to decode the rest of the segment
+      header = null;
+      throw e;
+    }
+  }
+
+  @Override
+  protected long getUnadjustedFrameLength(ByteBuf buffer, int offset, int length, ByteOrder order) {
+    // The parent class calls this repeatedly for the same "frame" if there weren't enough
+    // accumulated bytes the first time. Only decode the header the first time:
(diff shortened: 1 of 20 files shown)


olim7t:
EDIT -- negotiation is fixed on the server side; as of [b5afbe9d71](https://github.com/beobal/cassandra/commit/b5afbe9d71), the hack below is not needed anymore.

-----
Here's a temporary hack of the server code to simulate not wrapping the response to the STARTUP message. Apply on top of [1a8a79b](https://github.com/beobal/cassandra/commit/1a8a79b1f5ff6f863263a28b1160431b8ba6eef5):
```
diff --git a/src/java/org/apache/cassandra/transport/Server.java b/src/java/org/apache/cassandra/transport/Server.java
index 28331e4f03..243ab695ee 100644
--- a/src/java/org/apache/cassandra/transport/Server.java
+++ b/src/java/org/apache/cassandra/transport/Server.java
@@ -80,6 +80,7 @@ import org.apache.cassandra.service.*;
 import org.apache.cassandra.transport.messages.ErrorMessage;
 import org.apache.cassandra.transport.messages.EventMessage;
 import org.apache.cassandra.transport.messages.OptionsMessage;
+import org.apache.cassandra.transport.messages.ReadyMessage;
 import org.apache.cassandra.transport.messages.StartupMessage;
 import org.apache.cassandra.transport.messages.SupportedMessage;
 import org.apache.cassandra.utils.FBUtilities;
@@ -603,6 +604,8 @@ public class Server implements CassandraDaemon.Server

                     if (frame.header.version.isGreaterOrEqualTo(ProtocolVersion.V5))
                     {
+                        // Temporary hack to send the response to the first STARTUP without an outer frame, as agreed on the ticket.
+                        ctx.writeAndFlush(Message.ProtocolEncoder.instance.encodeMessage(new ReadyMessage(), frame.header.version));
                         try
                         {
                             configureModernPipeline(ctx, decoder, frame.header.bodySizeInBytes, message, payloadTracker);
```
This is not at all the right way to do this, but it produces the expected outcome in the nominal case (client and server both support v5, no authentication).

With this in place, I can connect to the server and execute a request over the new protocol.

olim7t:
Rebased and force-pushed to clean up the history.

driver-core/src/main/java/com/datastax/driver/core/BytesToSegmentDecoder.java
```diff
@@ -0,0 +1,72 @@
...
+        segmentCodec.headerLength() + SegmentCodec.CRC24_LENGTH);
+    this.segmentCodec = segmentCodec;
+  }
+
+  @Override
+  protected Object decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception {
+    ByteBuf payloadAndCrc = (ByteBuf) super.decode(ctx, in);
+    assert header != null;
+    if (payloadAndCrc == null) {
+      return null;
```
beobal:
```suggestion
      header = null;
      return null;
```

I think you need to also null out the header here in cases where the input buffer doesn't have enough readable bytes to decode the full payload. As it is, the next time `super.decode` is called it hits the assertion in `getUnadjustedFrameLength`

olim7t: Right, I was assuming that `getUnadjustedFrameLength` was called only once for each frame, but indeed it's called every time until we have enough bytes to read.

Fixing it, thanks.

driver-core/src/main/java/com/datastax/driver/core/BytesToSegmentDecoder.java
```diff
@@ -52,11 +52,18 @@
...
   @Override
   protected Object decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception {
-    ByteBuf payloadAndCrc = (ByteBuf) super.decode(ctx, in);
-    assert header != null;
+    ByteBuf payloadAndCrc;
+    try {
+      payloadAndCrc = (ByteBuf) super.decode(ctx, in);
+    } catch (Exception e) {
+      // Don't hold on to a stale header if the parent method discarded the "frame"
+      header = null;
```
olim7t, line 67: Actually this doesn't even matter currently, because we always drop the connection when an error happens here.
But just being safe in case that ever changes.

driver-core/src/main/java/com/datastax/driver/core/BytesToSegmentDecoder.java
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  @Override
+  protected Object decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception {
+    try {
+      ByteBuf payloadAndCrc = (ByteBuf) super.decode(ctx, in);
+      if (payloadAndCrc == null) {
+        return null;
+      } else {
+        assert header != null;
```
olim7t, line 60: That assertion was in the wrong place too. The header can still be null after `super.decode()` if we didn't even have 6 bytes to read.

driver-core/src/main/java/com/datastax/driver/core/BytesToSegmentDecoder.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+
+  @Override
+  protected long getUnadjustedFrameLength(ByteBuf buffer, int offset, int length, ByteOrder order) {
+    // The parent class calls this repeatedly for the same "frame" if there weren't enough
+    // accumulated bytes the first time. Only decode the header the first time:
+    if (header == null) {
+      header = segmentCodec.decodeHeader(buffer.slice(offset, length));
+    }
```
olim7t, line 78: I didn't want to decode every time, it feels inefficient even if it's a small operation.

This is safe because the only way `super.decode()` gives up on a frame is by throwing an exception, and we handle that in our own `decode()`.

olim7t:
I think I can do the same thing as in driver 4 and move segment construction to the pipeline. It would be simpler because we wouldn't need any change to the flusher. I'll give it a try.

olim7t:
Done, it's much cleaner that way.

olim7t:
@ifesdjeen just a heads up that I did a bit of refactoring ☝️ 
No functional changes, but you might want to switch to the latest version for your server tests.

olim7t:
I'm merging this now, for inclusion in 3.10. The frame/segment naming is still provisional, but we can revisit that later since it does not impact the public API (see [JAVA-2851](https://datastax-oss.atlassian.net/browse/JAVA-2851)).

PR #1445, opened 2020-05-28
JAVA-2793: Add composite config loader

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -265,6 +266,28 @@ static DriverConfigLoader fromMap(@NonNull OptionsMap source) {
...
+   *
+   * <p>The new config is reloadable if at least one of the input configs is. If you invoke {@link
+   * DriverConfigLoader#reload()} on the new loader, it will reload whatever is reloadable, or fail
+   * if nothing is. If the input loaders have periodic reloading built-in, each one will reload at
+   * its own pace, and the changes will be reflected in the new config.
+   */
+  @NonNull
+  static DriverConfigLoader compose(
+      @NonNull DriverConfigLoader primaryConfigLoader,
+      @NonNull DriverConfigLoader fallbackConfigLoader) {
```
olim7t, line 287: I was pondering adding a warning in the javadocs.

The feature works and I think it's good to have for completeness, but still it feels pretty convoluted.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverExecutionProfile.java
```diff
@@ -201,4 +219,109 @@ default Duration getDuration(@NonNull DriverOption option, @Nullable Duration de
    */
   @NonNull
   SortedSet<Map.Entry<String, Object>> entrySet();
+
+  @NonNull
+  @Override
+  default DriverExecutionProfile withBoolean(@NonNull DriverOption option, boolean value) {
+    return DerivedExecutionProfile.with(this, option, value);
+  }
```
olim7t, line 227: I realized the logic for derived profiles could be extracted in a generic implementation, so now it doesn't have to be done for every implementation, which is nice.

I still kept the overrides for the Typesafe implementation, because it can be done a bit better there.

core/src/main/java/com/datastax/oss/driver/internal/core/config/composite/CompositeDriverConfig.java
```diff
@@ -0,0 +1,60 @@
...
+                : v.refresh());
+  }
+
+  @NonNull
+  @Override
+  public Map<String, ? extends DriverExecutionProfile> getProfiles() {
+    // The map is updated lazily, if we want all the profiles we need to fetch them explicitly
+    for (String name :
+        Sets.union(primaryConfig.getProfiles().keySet(), fallbackConfig.getProfiles().keySet())) {
+      getProfile(name);
```
tomekl007, line 56: it looks a bit odd that we are ignoring the result from `getProfile()` call. Can we refactor it to not have side-effects, but to actually use the returned value? Or otherwise, change the method name to smth like: `refreshProfile()`?

core/src/main/java/com/datastax/oss/driver/internal/core/config/composite/CompositeDriverConfigLoader.java
```diff
@@ -0,0 +1,85 @@
...
+      return CompletableFutures.failedFuture(
+          new UnsupportedOperationException(
+              "Reloading is not supported (this is a composite config, "
+                  + "and neither the primary nor the fallback are reloadable)"));
+    } else if (!primaryConfigLoader.supportsReloading()) {
+      return fallbackConfigLoader.reload();
+    } else if (!fallbackConfigLoader.supportsReloading()) {
+      return primaryConfigLoader.reload();
+    } else {
+      return fallbackConfigLoader
```
tomekl007: - is there a specific reason for this order of reloading? ( I mean that fallback first, then primary)
- to double-check: the `thenCompose` will execute also if the fallback `reload()` will fail, right?
- Should we log it if there will be an error while reloading? (for example this:
https://github.com/datastax/java-driver/pull/1445/files#diff-dc57c9b73b57019d14f9babae5c5d7b2R72-R77)

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverExecutionProfile.java
```diff
@@ -269,7 +270,7 @@ public DriverExecutionProfile without(@NonNull DriverOption option) {
   @NonNull
   @Override
   public Object getComparisonKey(@NonNull DriverOption option) {
-    // No need to cache this, it's only used for policy initialization
+    // Override the default implementation because we can do much simpler
```
tomekl007: nit: maybe could you expand on that comment? When I am reading code it does not give me a lot of info. I think that there is too little context here.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
adutra, line 287: While at it, why not introduce a varargs parameter here and allow to compose more than 2 loaders together?

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverExecutionProfile.java
```diff
@@ -182,12 +184,28 @@ default Duration getDuration(@NonNull DriverOption option, @Nullable Duration de
...
+   * their options differ, two separate instances will be created.
+   *
+   * <p>The runtime return type does not matter, as long as identical sections (same options with
+   * same values, regardless of order) compare as equal and have the same {@code hashCode()}. The
+   * default implementation builds a map based on the entries from {@link #entrySet()}, it should be
+   * good for most cases.
    */
   @NonNull
-  Object getComparisonKey(@NonNull DriverOption option);
+  default Object getComparisonKey(@NonNull DriverOption option) {
```
adutra, line 198: The implementation in `MapBasedDriverExecutionProfile` is now identical. I think you can remove it, revapi seems to accept it.

core/src/main/java/com/datastax/oss/driver/internal/core/config/composite/CompositeDriverConfig.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+public class CompositeDriverConfig implements DriverConfig {
+
+  private final DriverConfig primaryConfig;
+  private final DriverConfig fallbackConfig;
+  private final Map<String, CompositeDriverExecutionProfile> profiles = new ConcurrentHashMap<>();
+
+  public CompositeDriverConfig(
+      @NonNull DriverConfig primaryConfig, @NonNull DriverConfig fallbackConfig) {
```
adutra, line 34: Again, if we are going through all this hassle, I feel like we should allow composition of N configs, not just 2.

core/src/main/java/com/datastax/oss/driver/internal/core/config/composite/CompositeDriverConfigLoader.java
```diff
@@ -0,0 +1,85 @@
...
+  @Override
+  public void onDriverInit(@NonNull DriverContext context) {
+    fallbackConfigLoader.onDriverInit(context);
+    primaryConfigLoader.onDriverInit(context);
+  }
+
+  @NonNull
+  @Override
+  public CompletionStage<Boolean> reload() {
+    if (!primaryConfigLoader.supportsReloading() && !fallbackConfigLoader.supportsReloading()) {
```
adutra, line 56: I find it a bit odd that a reload can be successful if only one of the inner loaders supports reloading, but would fail if both don't. Is there a specific reason why you think this is desirable? Imo the logic would seem easier to grasp if we failed as soon as one loader doesn't support reloading.

core/src/main/java/com/datastax/oss/driver/internal/core/config/composite/CompositeDriverExecutionProfile.java
```diff
@@ -0,0 +1,196 @@
...
+
+  /**
+   * Fetches the underlying profiles again from the two backing configs. This is because some config
+   * implementations support adding/removing profiles at runtime.
+   *
+   * <p>For efficiency reasons this is only done when the user fetches the profile again from the
+   * main config, not every time an option is fetched from the profile.
+   */
+  public CompositeDriverExecutionProfile refresh() {
+    // There's no `hasProfile()` in the public API because it didn't make sense until now. So
```
adutra, line 64: Can/should we add it?

core/src/main/java/com/datastax/oss/driver/internal/core/config/composite/CompositeDriverExecutionProfile.java
```diff
@@ -0,0 +1,196 @@
...
+  private volatile DriverExecutionProfile fallbackProfile;
+
+  public CompositeDriverExecutionProfile(
+      @NonNull DriverConfig primaryConfig,
+      @NonNull DriverConfig fallbackConfig,
+      @NonNull String profileName) {
+    this.primaryConfig = Objects.requireNonNull(primaryConfig);
+    this.fallbackConfig = Objects.requireNonNull(fallbackConfig);
+    this.profileName = Objects.requireNonNull(profileName);
+    refresh();
```
adutra: Calling a public method from a constructor is error-prone, could you extract the default implementation of `refresh()` to a private method?.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverExecutionProfile.java
adutra: The implementation here is a one-liner, vs 8 lines in the default impl, that's what the comment is referring to.

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverExecutionProfile.java
olim7t: That was my intention, I forgot. Thanks for pointing it out.

core/src/main/java/com/datastax/oss/driver/internal/core/config/composite/CompositeDriverConfigLoader.java
olim7t: The order does not matter, but you bring a good point about error handling, `thenCompose` does not execute the callback if the first future is failed. I'll fix the code to reload both independently and only combine the result after.

If an error occurs, it will be passed to the caller via the resulting stage, so we should not log it.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
olim7t: Because it's overkill IMHO.

Frankly I already have doubts about composing two loaders (per my previous comment). Composing more than two sounds even more wrong.

You can always chain `compose()` calls.

core/src/main/java/com/datastax/oss/driver/internal/core/config/composite/CompositeDriverExecutionProfile.java
olim7t: Outside of this implementation I don't see a lot of value in the general case. I think users create profiles in their configuration and use them, it's not like you get a random opaque config and have to figure out what's in it.

core/src/main/java/com/datastax/oss/driver/internal/core/config/composite/CompositeDriverExecutionProfile.java
olim7t: Will do.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverExecutionProfile.java
olim7t: 👍 I'll rephrase the comment.

core/src/main/java/com/datastax/oss/driver/internal/core/config/composite/CompositeDriverConfigLoader.java
olim7t: I guess it's a matter of interpretation, but the way I see it is that if something has dynamic parts and static parts, then the thing as a whole is still dynamic. Kind of like a dynamic web page, you wouldn't instruct the browser to cache it indefinitely because some of the resources are static.

core/src/main/java/com/datastax/oss/driver/internal/core/config/composite/CompositeDriverConfig.java
olim7t: It is a bit odd, the map acts as a loading cache, we need to fetch to make sure everything is up-to-date, but we end up using the map itself to build the result.

It's all private and within 15 lines of code though, I think it will be fine.

adutra (approved):

PR #1446, opened 2020-05-31
JAVA-2792: Allow custom results in the mapper

olim7t:
TODO:
- [x] add a processor flag to opt out of custom types. I think having them on by default is the right choice, but if someone is not using them, it could be nice to have a way to go back to the stricter compile-time checks.
- [x] javadocs
- [x] manual. This deserves a dedicated page.
- [x] add nullability annotations
- [x] document custom return types in the annotations' javadocs


mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MappedResultProducer.java
```diff
@@ -0,0 +1,31 @@
...
+import com.datastax.oss.driver.api.core.cql.Statement;
+import com.datastax.oss.driver.api.core.type.reflect.GenericType;
+import com.datastax.oss.driver.api.mapper.entity.EntityHelper;
+
+public interface MappedResultProducer {
+
+  boolean canProduce(GenericType<?> resultType);
+
+  @SuppressWarnings("TypeParameterUnusedInFormals")
+  <EntityT> Object execute(
```
tomekl007: For the JAVA-2754 the type that we want to support is `MutinyMappedReactiveResultSet<EntityT>`. The goal is to wrap the `Publisher<>` (in fact the `MappedReactiveResultSet` that extends `Publisher`) within the `MutinyMappedReactiveResultSet` API, see: (https://github.com/datastax/cassandra-quarkus/pull/11/files#diff-81a76325271bcd182f8cf5f0cfc5dd45R43 and https://github.com/datastax/cassandra-quarkus/pull/11/files#diff-833ec0655b43a8e6649bceb026d9a14dR41)

Ideally, I would like to be able to execute the `executeReactiveAndMap(entityHelper, statement)` and wrap the result within the `new DefaultMutinyMappedReactiveResultSet`. 
Currently, it is possible but requires some code duplication:
https://github.com/datastax/cassandra-quarkus/pull/11/commits/c73996f2d0fd401c261bd2ece6c5b013fd73c0d8#diff-e788c33b4c50e6b07ff072f711ce4532R45-R51
and
https://github.com/datastax/cassandra-quarkus/pull/11/commits/c73996f2d0fd401c261bd2ece6c5b013fd73c0d8#diff-e788c33b4c50e6b07ff072f711ce4532R36-R38

I was thinking that maybe the `MappedResultProducer` should extend the `DaoBase` somehow to allow the client to use those methods without duplicating them?

integration-tests/src/test/java/com/datastax/oss/driver/mapper/CustomResultTypeIT.java
```diff
@@ -0,0 +1,243 @@
...
+
+    @DaoFactory
+    ProductDao productDao(@DaoKeyspace CqlIdentifier keyspace);
+
+    static MapperBuilder<InventoryMapper> builder(CqlSession session) {
+      return new CustomResultTypeIT_InventoryMapperBuilder(session);
+    }
+  }
+
+  public abstract static class ListenableFutureProducer implements MappedResultProducer {
```
tomekl007: For JAVA-2754 we need to provide producers for a client to use them in the mapper:
https://github.com/datastax/cassandra-quarkus/pull/11/commits/c73996f2d0fd401c261bd2ece6c5b013fd73c0d8#diff-e788c33b4c50e6b07ff072f711ce4532R28
To make it work, the client will need to pass the `GenericType` of a required entity class to the Producer instance:
https://github.com/datastax/cassandra-quarkus/pull/11/commits/c73996f2d0fd401c261bd2ece6c5b013fd73c0d8#diff-3f2336e51cf1fcd26707d79b6d633cb5R38-R41
I wonder if we can simplify it somehow (make `MappedResultProducer` generic?) or are we ok with this requirement of using `GenericType` in the clients of `MappedResultProducer` classes?

Or using the `resultType.isSubtypeOf` of `?` is a required way of using this:
- https://github.com/datastax/cassandra-quarkus/pull/11/commits/83e47bb8f0e30703f7267b1643d4fcbdec267e21#diff-e788c33b4c50e6b07ff072f711ce4532R39
- https://github.com/datastax/cassandra-quarkus/pull/11/commits/83e47bb8f0e30703f7267b1643d4fcbdec267e21#diff-e788c33b4c50e6b07ff072f711ce4532R30

?

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MappedResultProducer.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.mapper;
+
+import com.datastax.oss.driver.api.core.cql.Statement;
+import com.datastax.oss.driver.api.core.type.reflect.GenericType;
+import com.datastax.oss.driver.api.mapper.entity.EntityHelper;
+
+public interface MappedResultProducer {
```
tomekl007: NIt is the `Mapped` a proper prefix here? It can be used also for not mapped results sets, see this conversion from `CompletionStage` to `Uni`:
https://github.com/datastax/cassandra-quarkus/pull/11/commits/ea171cb528a3dc5d75f8e2062b765f0b897b688e#diff-91bdea213ffff0bc78a4bc81a5f6ed1fR37

integration-tests/src/test/java/com/datastax/oss/driver/mapper/CustomResultTypeIT.java
```diff
@@ -0,0 +1,243 @@
...
+    CqlSession session = SESSION_RULE.session();
+
+    for (String query : createStatements(CCM_RULE)) {
+      session.execute(
+          SimpleStatement.builder(query).setExecutionProfile(SESSION_RULE.slowProfile()).build());
+    }
+
+    InventoryMapper mapper =
+        InventoryMapper.builder(SESSION_RULE.session())
+            .withResultProducers(
```
tomekl007: Can we abstract away that `withResultProducers`? I am thinking that it could be more UX friendly if their will be set on the `*Mapper` class itself.

For JAVA-2754 the use case is that if someone would like to use provided mapped producers (`toMulti` and `toUni`) he needs to remember that and call `withResultProducer` every time the Mapper is created:
https://github.com/datastax/cassandra-quarkus/pull/11/files#diff-3f2336e51cf1fcd26707d79b6d633cb5R36-R38 
There could be a use case for more than 2 mapped producers as well.

It could be alleviated by providing a way for a user to just extends their mapper (via `@Mapper` or just class extends) so instead of this:
```
@Mapper
public interface FruitMapper {
}
```
and then when constructing this:
```
 return new FruitMapperBuilder(cqlSession)
        .withResultProducers(new MutinyMappedResultProducer(), new UniMappedResultSetProducer())
        .build();
```
the quarkus framework would provide a class 
```
interface ResultSetProducers{
  MappedResultSetProducer producers();
}
class QuarkusResultSetProducers implement ResultSetProducers{
    @Override
    MappedResultSetProducer producers(){
       return Arrays.ofList(new MutinyMappedResultProducer(), new UniMappedResultSetProducer())
    }
}
```
and then a user when creating a Mapper could do 
```
@Mapper(withProducers = QuarkusResultSetProducers.class)
public interface FruitMapper {
}
```
or:
```
public interface FruitMapper extend QuarkusResultSetProducers{
}
```
The object-mapper would need to use the returned `producers()` instead of the builder method `withResultProducers`.
Then there will be no need for the client to remember and add all mappedProducers that he needs to use in:
```
.withResultProducers(new MutinyMappedResultProducer(), new UniMappedResultSetProducer())
```

The pros of this solution that I see are:
- the mappedProducers are encapsulated in the quarkus framework.
- adding a new mapped producer in the future would require no work from the client-side.

tomekl007:
Side question:
when using a new approach in Quarkus **Native** mode the processor is not found:
```
2020-06-01 17:18:22,526 ERROR [org.jbo.res.res.i18n] (executor-thread-1) RESTEASY002020: Unhandled asynchronous exception, sending back 500: java.lang.IllegalArgumentException: No request processor found for com.datastax.oss.driver.internal.core.cql.DefaultBoundStatement@10d28ad08
	at com.datastax.oss.driver.internal.core.session.RequestProcessorRegistry.processorFor(RequestProcessorRegistry.java:54)
	at com.datastax.oss.driver.internal.core.session.DefaultSession.execute(DefaultSession.java:227)
	at com.datastax.dse.driver.api.core.cql.reactive.ReactiveSession.executeReactive(ReactiveSession.java:60)
	at com.datastax.oss.quarkus.runtime.internal.session.DefaultQuarkusCqlSession.executeReactive(DefaultQuarkusCqlSession.java:65)
	at com.datastax.oss.quarkus.runtime.CassandraClientProducer_ProducerMethod_createDefaultCassandraClient_56bb2685a5ebdaa8256d5379e1323898239219fa_ClientProxy.executeReactive(CassandraClientProducer_ProducerMethod_createDefaultCassandraClient_56bb2685a5ebdaa8256d5379e1323898239219fa_ClientProxy.zig:249)
	at com.datastax.oss.quarkus.runtime.api.session.QuarkusCqlSession.executeReactive(QuarkusCqlSession.java:30)
	at com.datastax.oss.quarkus.runtime.api.reactive.mapper.MutinyMappedResultProducer.executeReactive(MutinyMappedResultProducer.java:36)
	at com.datastax.oss.quarkus.runtime.api.reactive.mapper.MutinyMappedResultProducer.execute(MutinyMappedResultProducer.java:47)
	at com.datastax.oss.quarkus.dao.ProductDaoReactiveImpl__MapperGenerated.findById(ProductDaoReactiveImpl__MapperGenerated.java:77)
	at com.datastax.oss.quarkus.CassandraEndpoint.getProductReactive(CassandraEndpoint.java:65)
	at java.lang.reflect.Method.invoke(Method.java:498)
```
maybe you have an idea why it may not work in the Quarkus now?
The code path that is failing in native mode is starting here:
https://github.com/datastax/cassandra-quarkus/pull/11/files#diff-76fba74b75a4ea99dd5614d70f6c0bbaR69

The code path that does not go through the code path using `mappedProducer` seems to work in the `Native` mode.
**EDIT:**
it was unrelated to this PR, sorry for the noise

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MappedResultProducer.java
olim7t: Even after inlining callees, `executeReactiveAndMap` is just two lines of trivial code:
```java
ReactiveResultSet source = context.getSession().executeReactive(statement);
return new DefaultMappedReactiveResultSet<>(source, entityHelper::get);
```
So there's not much duplication.
Besides, `DaoBase` is internal, I'd rather have implementors use the public session API.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/CustomResultTypeIT.java
olim7t: The client should not register a `MutinyMappedReactiveResultSet<Fruit>`, but instead a producer that can handle a reactive result set of _any entity_. See `CustomResultTypeIT.SingleEntityListenableFutureProducer` for an example.

The idea is that we would provide those Mutiny producers out of the box with the Quarkus extension, users should not have to implement them.

PS:  There's not any good way to make the interface generic, and at the end of the day it all happens at runtime anyway, there would always be an unchecked cast somewhere.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MappedResultProducer.java
olim7t: I was meant as "a result produced by the mapper".
Maybe `MapperResultProducer` would be a better name. I'll give it more thought, I wasn't completely satisfied with naming.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/CustomResultTypeIT.java
olim7t: See my other answer, we can write producers that will handle any kind of entity.

tomekl007: yes, I agree but this is not related. We still need to provide two `mapperResultsProducers` and my suggestion was to hide it from the client:
(this call `.withResultProducers(new MutinyMappedResultProducer(), new UniMappedResultSetProducer())`)

adutra: Aren't result producers good candidates for loading through the Service Loader API?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoReturnTypeKind.java
```diff
@@ -32,9 +33,10 @@
    *
    * @param methodBuilder the method to add the code to.
    * @param helperFieldName the name of the helper for entity conversions (might not get used for
-   *     certain kinds, in that case it's ok to pass null).
+   * @param returnTypeName
```
adutra: The previous param description was accidentally truncated it seems.

mapper-processor/src/test/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoDeleteMethodGeneratorTest.java
```diff
@@ -163,4 +171,32 @@ public void should_warn_when_non_bind_marker_has_cql_name() {
...
+        TypeSpec.interfaceBuilder(ClassName.get("test", "ProductDao"))
+            .addModifiers(Modifier.PUBLIC)
+            .addAnnotation(Dao.class)
+            .addMethod(methodSpec)
+            .build();
+
+    for (List<String> compilerOptions :
+        ImmutableList.of(
+            ImmutableList.of("-Acom.datastax.oss.driver.mapper.customResults.enabled=true"),
+            // The option default to true, so it should also work without explicit options:
```
adutra: Nit: defaults to true.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DefaultMapperContext.java
```diff
@@ -140,6 +150,20 @@ public NameConverter getNameConverter(Class<? extends NameConverter> converterCl
     return customState;
   }
 
+  @NonNull
+  @Override
+  public MapperResultProducer getResultProducer(GenericType<?> resultToProduce) {
+    for (MapperResultProducer resultProducer : resultProducers) {
```
adutra: This is a bit inefficient, shouldn't we cache the results?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
```diff
@@ -234,4 +244,30 @@ protected boolean isFromClassFile() {
...
+   *
+   * @param createStatementBlock the code that creates the statement. It must store it into a
+   *     variable named "boundStatement".
+   */
+  protected Optional<MethodSpec> crudMethod(
+      CodeBlock.Builder createStatementBlock, DaoReturnType returnType, String helperFieldName) {
+
+    MethodSpec.Builder method = GeneratedCodePatterns.override(methodElement, typeParameters);
+    TypeName returnTypeName = null;
+    if (returnType.getKind() == CUSTOM) {
```
adutra, line 258: it's a bit odd the the `producer` variable declaration is generated here, but only accessed by code generated inside `addExecuteStatement`. Instead of special-casing CUSTOM here, can't you generate the `producer` variable delcaration inside `DefaultDaoReturnTypeKind.CUSTOM.addExecuteStatement()` ?

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MapperResultProducer.java
```diff
@@ -0,0 +1,107 @@
...
+   * @param entityHelper if the type to produce contains a mapped entity (e.g. {@code
+   *     ListenableFuture<Product>}), an instance of the helper class to manipulate that entity. In
+   *     particular, {@link EntityHelper#get(GettableByName) entityHelper.get()} allows you to
+   *     convert rows into entity instances. If the type to produce does not contain an entity, this
+   *     will be {@code null}.
+   * @return the object to return from the DAO method. This must match the type that this producer
+   *     was selected for, there will be an unchecked cast at runtime.
+   */
+  @SuppressWarnings("TypeParameterUnusedInFormals")
+  <EntityT> Object execute(
```
adutra: Maybe add nullability annotations in this interface?

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MapperResultProducer.java
```diff
@@ -0,0 +1,107 @@
...
+   * Surfaces any error encountered in the DAO method (either in the generated mapper code that
+   * builds the statement, or during invocation of {@link #execute}).
+   *
+   * <p>For some result types, it is expected that errors will be wrapped in some sort of container
+   * instead of thrown directly; for example a failed future or publisher.
+   *
+   * <p>If rethrowing is the right thing to do, then it is perfectly fine to do so from this method
+   * (note however that you'll need to wrap checked exceptions).
+   */
+  Object wrapError(Throwable error);
```
adutra: Wouldn't it be a bit more user-friendly if this method declared to throw `Exception`? We could then catch it in call sites and convert the thrown error into something more appropriate.

adutra (changes requested):

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoMethodGenerator.java
olim7t: The output of `addExecuteStatement()` is enclosed in the `try` block. We also need the producer in the `catch` block to wrap errors.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MapperResultProducer.java
olim7t: Right, I forgot.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MapperResultProducer.java
olim7t: Great idea, I hadn't thought of doing that in the generated code. We could even check if the DAO method declares any exceptions, and rethrow those without wrapping.

integration-tests/src/test/java/com/datastax/oss/driver/mapper/CustomResultTypeIT.java
olim7t: Sorry @tomekl007 I glanced over your initial comment a bit too quickly. The idea of a `@Mapper` attribute is interesting, but I find the service loader idea even more compelling: we could publish specialized modules like `mapper-mutiny`,  `mapper-rxjava`, etc, and all users would have to do is put them in their classpath. I'm going to explore that route.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DefaultMapperContext.java
```diff
@@ -155,4 +184,12 @@ private static NameConverter buildNameConverter(Class<? extends NameConverter> c
           e);
     }
   }
+
+  private static List<MapperResultProducer> getResultProducers() {
+    ImmutableList.Builder<MapperResultProducer> result = ImmutableList.builder();
+    ServiceLoader<MapperResultProducerService> loader =
+        ServiceLoader.load(MapperResultProducerService.class);
+    loader.iterator().forEachRemaining(provider -> result.addAll(provider.getProducers()));
```
adutra, line 192: Do we really need the `MapperResultProducerService` interface? We could load instances of `MapperResultProducer` directly using the Service Loader API.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/result/MapperResultProducerService.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Provides the custom mapper result types that will be used in an application.
+ *
+ * <p>This class is loaded with the Java Service Provider Interface mechanism, you must reference it
+ * via a service descriptor: create a file {@code
+ * META-INF/services/com.datastax.oss.driver.api.mapper.result.MapperResultProducerService}, with a
+ * single line that contains the name of the implementing class.
+ */
+public interface MapperResultProducerService {
```
adutra, line 26: See my other comment, not sure if this interface is required.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/result/MapperResultProducerService.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+package com.datastax.oss.driver.api.mapper.result;
+
+/**
+ * Provides the custom mapper result types that will be used in an application.
+ *
+ * <p>This class is loaded with the Java Service Provider Interface mechanism, you must reference it
+ * via a service descriptor: create a file {@code
+ * META-INF/services/com.datastax.oss.driver.api.mapper.result.MapperResultProducerService}, with a
+ * single line that contains the name of the implementing class.
```
adutra: You can have more than one line, if there are more than one implementations to load.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/MapperContext.java
```diff
@@ -83,4 +85,18 @@
    */
   @NonNull
   Map<Object, Object> getCustomState();
+
+  /**
+   * Returns a component that will execute a statement and convert it into a custom result of the
+   * given type.
+   *
+   * <p>These components must be registered at build time with {@link
+   * MapperBuilder#withResultProducers(MapperResultProducer...)}.
```
adutra: This method was removed.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/result/MapperResultProducerService.java
adutra: Oh it just occurred to me: is it because you want to have a deterministic order when loading producers? In this case, OK, I'm fine with this interface.

adutra (changes requested):

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DefaultMapperContext.java
olim7t: Indeed, I hadn't thought of that. If the order in the descriptor is deterministic (it probably is), then I'm fine loading producers directly. I'll check.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DefaultDaoReturnTypeKind.java
```diff
@@ -18,236 +18,433 @@
...
+    @Override
+    public void addExecuteStatement(
+        CodeBlock.Builder methodBuilder,
+        String helperFieldName,
+        ExecutableElement methodElement,
+        Map<Name, TypeElement> typeParameters) {
+      TypeName returnTypeName =
+          GeneratedCodePatterns.getTypeName(methodElement.getReturnType(), typeParameters);
+      methodBuilder.addStatement(
+          "return ($T) producer.execute(boundStatement, context, $L)",
```
adutra: This statement produces compiler warnings:

```
[INFO] [...] FruitDaoReactiveImpl__MapperGenerated.java uses unchecked or unsafe operations.
[INFO] [...]  Recompile with -Xlint:unchecked for details.
---
```

Would it be possible to extract the result to a local variable and annotate it with `@SuppressWarnings("unchecked")`? Alternatively, maybe we should consider annotating the whole class with `@SuppressWarnings("all")`.

olim7t: I think we can extract local variables in the code generated by `DefaultDaoReturnTypeKind.CUSTOM`, and annotate them individually.

olim7t: Done.
I wasn't getting any warnings when compiling our integration tests, but maybe the project is configured differently.

mapper-runtime/src/main/java/com/datastax/oss/driver/internal/mapper/DefaultMapperContext.java
olim7t: The order is not formally defined by the `ServiceLoader` javadocs.

In practice, the Oracle JDK implementation parses the lines and puts them in a list, so they come out in the original order. But to play devil's advocate, it's not unrealistic to imagine another implementation that would use a hash set in order to eliminate duplicates (which are supposed to be skipped). 

So just to err on the safe side I think it's better to keep the top-level service interface, it doesn't add much complexity.

adutra (approved):

olim7t:
OK, I think everything has been addressed. I'll leave this open until tomorrow so that @tomekl007 can take one last look.

tomekl007 (approved):
LGTM

PR #1448, opened 2020-06-05
JAVA-2792 follow-up fixes

olim7t:


adutra (approved):

adutra:
A few other considerations on the new `MapperResultProducer` API:

1. The signature `<EntityT> Object execute(... EntityHelper<EntityT> entityHelper)` has a type parameter that is probably never going to be useful. Even the most generic implementations could get away without that type parameter. For instance, both versions below compile without warnings:

```
public <EntityT> Object execute(Statement<?> statement, MapperContext context, EntityHelper<EntityT> entityHelper) {
    ResultSet rows = context.getSession().execute(statement);
    Iterable<EntityT> entities = Iterables.transform(rows, entityHelper::get);
    return new Container<>(entities);
}

public Object execute(Statement<?> statement, MapperContext context, EntityHelper<?> entityHelper) {
    ResultSet rows = context.getSession().execute(statement);
    Iterable<Object> entities = Iterables.transform(rows, entityHelper::get);
    return new Container<>(entities);
}
```

However if your implementation only works for a given particular entity type, then both versions generate warnings just the same:

```
public <EntityT> Object execute(Statement<?> statement, MapperContext context, EntityHelper<EntityT> entityHelper) {
    ResultSet rows = context.getSession().execute(statement);
    // unchecked cast
    Iterable<Foo> entities = (Iterable<Foo>) Iterables.transform(rows, entityHelper::get);
    return new FooContainer(entities);
}

public Object execute(Statement<?> statement, MapperContext context, EntityHelper<?> entityHelper) {
    ResultSet rows = context.getSession().execute(statement);
    // unchecked cast
    Iterable<Foo> entities = (Iterable<Foo>) Iterables.transform(rows, entityHelper::get);
    return new FooContainer(entities);
}
```
2. I am doubtful about the usefulness of `wrapError`.
    2.a Since `wrapError` is only ever called when `execute` throws, we could then consider that it is the responsibility of `execute` to also handle errors, and return a wrapped error when appropriate, or throw. I suggest that we remove `wrapError` and amend `execute` to declare that it `throws Exception`.
    2.b Even if we keep it, it is probably not a good idea to pass a `Throwable` to it, as even the simplest implementation wouldn't compile:
```
public Object wrapError(@NonNull Throwable error) throws Exception {
  throw error; // doesn't compile
}
```

2.c Expanding on this idea, it's probably not a good idea to catch `Throwable` in generated code. I suggest to simplify this generated code:
```
try {
  BoundStatementBuilder boundStatementBuilder = ...
  BoundStatement boundStatement = boundStatementBuilder.build();
  @SuppressWarnings("unchecked") Container result =
	  (Container) producer.execute(boundStatement, context, stockHelper);
  return result;
} catch (Throwable t) {
  try {
	@SuppressWarnings("unchecked") Container result =
		(Container) producer.wrapError(t);
	return result;
  } catch (Exception e) {
	Throwables.throwIfUnchecked(e);
	throw new RuntimeException(e);
  }
}
```
to just this:
```
try {
  BoundStatementBuilder boundStatementBuilder = ...
  BoundStatement boundStatement = boundStatementBuilder.build();
  // handles errors and throws checked exceptions
  @SuppressWarnings("unchecked") Container result =
	  (Container) producer.execute(boundStatement, context, stockHelper);
  return result;
} catch (Exception e) {
    Throwables.throwIfUnchecked(e);
    throw new RuntimeException(e);
}
```

olim7t:
> The signature `<EntityT> Object execute(... EntityHelper<EntityT> entityHelper)` has a type parameter that is probably never going to be useful.

It's useful if you use a covariant return type that is parameterized by the entity type, see the [manual example](https://github.com/datastax/java-driver/commit/d523e9a95e63a740bc6af4645130e438fba248df#diff-0efdde1134a367fa87a2244a66b4fd5dR113-R130):
```java
  public <EntityT> ListenableFuture<EntityT> execute(
      Statement<?> statement, MapperContext context, EntityHelper<EntityT> entityHelper)
```
In that situation it avoids an unchecked cast, and makes the code slightly nicer.
But yeah that's a pretty thin use case, as soon as you factor things (as I did in the integration tests), that goes away and the parameter is useless.
Maybe that one use case is not worth complicating the method signature, WDYT?

> `wrapError` is only ever called when `execute` throws

It also covers some mapper-generated code before `execute`: binding parameters, handling statement attributes...
Granted, most errors thrown from that are likely mapper configuration errors. But I still think it's cleaner to encapsulate them.

>  it is probably not a good idea to pass a `Throwable`

Agreed, I will make the changes you suggest.

adutra:
The example in the mapper imo would look just as elegant without the generics:

```
public ListenableFuture<?> execute(
		Statement<?> statement, MapperContext context, EntityHelper<?> entityHelper) {
	SettableFuture<Object> result = SettableFuture.create();
	CqlSession session = context.getSession();
	session
			.executeAsync(statement)
			.whenComplete(
					(resultSet, error) -> {
						if (error != null) {
							result.setException(error);
						} else {
							Row row = resultSet.one();
							result.set((row == null) ? null : entityHelper.get(row));      // (2)
						}
					});
	return result;
}
```

So yes I agree that it is not worth complicating the method signature.

adutra:
> > `wrapError` is only ever called when `execute` throws
> 
> It also covers some mapper-generated code before `execute`: binding parameters, handling statement attributes...

Ok fair point, I agree to keep that method.
> Granted, most errors thrown from that are likely mapper configuration errors. But I still think it's cleaner to encapsulate them.
> 
> > it is probably not a good idea to pass a `Throwable`
> 
> Agreed, I will make the changes you suggest.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/result/MapperResultProducer.java
```diff
@@ -103,5 +103,5 @@
    * declares them, or wrapped into a {@link RuntimeException} otherwise.
    */
   @Nullable
-  Object wrapError(@NonNull Throwable error) throws Exception;
+  Object wrapError(@NonNull Exception e) throws Exception;
```
adutra, line 105: Could be made default.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DefaultDaoReturnTypeKind.java
```diff
@@ -423,14 +423,14 @@ public CodeBlock wrapWithErrorHandling(
       // (note: manually a multi-catch would be cleaner, but from here it's simpler to generate
       // separate clauses)
       for (TypeMirror thrownType : methodElement.getThrownTypes()) {
-        callWrapError.nextControlFlow("catch ($T e)", thrownType).addStatement("throw e");
+        callWrapError.nextControlFlow("catch ($T e2)", thrownType).addStatement("throw e2");
```
adutra, line 426: Nit: could be called `e1`.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/result/MapperResultProducer.java
olim7t: How? The implementation will be specific to each custom result, for example `ListenableFuture<EntityT>` vs `Flowable<EntityT>`.

adutra: By throwing:

```
  @Nullable
  default Object wrapError(@NonNull Exception e) throws Exception {
    throw e;
  }
```

olim7t: Throwing is not appropriate for any future or reactive type.

PR #1450, opened 2020-06-07
fix typo in manual regarding named vs positional setters

RyanQuey:
In the prepared statements documentation, there is a place where the comments mix up "named" and "positional", and this pr fixes that


adutra (approved):
Thank you for your contribution @RyanQuey!

PR #1452, opened 2020-06-10
JAVA-2803: Move Graal substitutions for protocol compression from cassandra-quarkus into java-driver

absurdfarce:


core/src/main/java/com/datastax/oss/driver/internal/core/protocol/Lz4Present.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class Lz4Present implements BooleanSupplier {
+
+  private static final String LZ4_CLZ_NAME = "net.jpountz.lz4.LZ4Compressor";
+
+  @Override
+  public boolean getAsBoolean() {
+    try {
+      Class.forName(LZ4_CLZ_NAME);
+      return false;
```
emerkle826: The logic here seems reversed. The implementation here seems to return `true` if the LZ4 compression library is _not_ found on the classpath. But I think this is desired as the Substitution class below keys off of this class. If this class returns `true`, then the Substitution will be applied, which indicates that LZ4 is _not_ built into the native image. Is that correct?

If so, maybe rename this class to `Lz4Absent`?

absurdfarce: :100: , I outright got the name of this class wrong.  Fixing.

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/Lz4Substitution.java
```diff
@@ -0,0 +1,55 @@
...
+
+import com.datastax.oss.driver.api.core.context.DriverContext;
+import com.oracle.svm.core.annotate.Inject;
+import com.oracle.svm.core.annotate.Substitute;
+import com.oracle.svm.core.annotate.TargetClass;
+import io.netty.buffer.ByteBuf;
+
+@TargetClass(
+    className = "com.datastax.oss.driver.internal.core.protocol.Lz4Compressor",
+    onlyWith = Lz4Present.class)
```
emerkle826: I think you need to use `Lz4Missing` here since the rename.

absurdfarce: :facepalm: Man, apparently I shouldn't even have gotten out of bed this morning...

emerkle826: It's all good. That's what reviews are for. :)

emerkle826 (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrLibcSubstitution.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+package com.datastax.oss.driver.internal.core.os;
+
+import com.oracle.svm.core.annotate.Substitute;
+import com.oracle.svm.core.annotate.TargetClass;
+import java.util.Optional;
+
+/**
+ * Add an explicit Graal substitution for {@link JnrLibc}. If we don't implement something like the
+ * analysis done at Graal native image build time will discover the jnr references in JnrLibc even
```
adutra:
```suggestion
 * analysis done at Graal native image build time, we'll discover the jnr references in JnrLibc even
```

adutra: (you might need to reformat this class file)

adutra (approved):
Just found a small typo in the javadocs (I think). Otherwise 👍

core/src/main/java/com/datastax/oss/driver/internal/core/os/JnrLibcSubstitution.java
absurdfarce: "will" actually is what I wanted here, but as I read through this sentence I think it's missing a few other words that make it, you know, comprehensible. :facepalm:

Fixing now.

PR #1453, opened 2020-06-11
Do not fail build if there are test failures

adutra:


emerkle826 (approved):
This is fine until we (hopefully) get all of the tests passing.

PR #1456, opened 2020-06-17
JAVA-2818: Remove root path only after merging non-programmatic configs

adutra:
This PR reverts most of the changes introduced by JAVA-2657 to the programmatic config loader builder.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -2,8 +2,9 @@
 
 <!-- Note: contrary to 3.x, insert new entries *first* in their section -->
 
-### 4.8.0 (in progress)
+### 4.7.1 (in progress)
 
+- [bug] JAVA-2818: Remove root path only after merging non-programmatic configs
 
 ### 4.7.0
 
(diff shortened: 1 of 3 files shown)


olim7t (approved):
LGTM

For context, the "root path" mechanism exists so that different sessions can be configured with one config file, see [this](https://docs.datastax.com/en/developer/java-driver/4.7/manual/core/configuration/#changing-the-config-prefix).

adutra:
> LGTM
> 
> For context, the "root path" mechanism exists so that different sessions can be configured with one config file, see [this](https://docs.datastax.com/en/developer/java-driver/4.7/manual/core/configuration/#changing-the-config-prefix).

Interesting, I never thought about that possibility. I think I would have gone with different config files in that case (e.g. `application1.conf` and `application2.conf`). Parameterizing the prefix path seems a bit more involved.

Reading the docs you linked, it seems that this example is outdated:

```
DriverConfigLoader session1ConfigLoader =
    new DefaultDriverConfigLoader(
        () -> loadConfig("session1"), DefaultDriverOption.values());
``` 

I will look into it tomorrow.

PR #1458, opened 2020-06-24
JAVA-2773: Add protocol v5 checksumming to driver 4

olim7t:
Requires datastax/native-protocol#36 installed in the local repository (also why the Travis checks fail).

TODO:
- [ ] integration tests (in particular with compression)
- [ ] add metrics to track segment sizes in bytes and frames
- [ ] handle Snappy compression, since it's not supported by the server we should probably fail fast

### DIFF
--- a/core/src/main/java/com/datastax/dse/driver/internal/core/protocol/TinkerpopBufferPrimitiveCodec.java
+++ b/core/src/main/java/com/datastax/dse/driver/internal/core/protocol/TinkerpopBufferPrimitiveCodec.java
@@ -23,6 +23,7 @@ import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
+import java.util.zip.CRC32;
 import org.apache.tinkerpop.gremlin.structure.io.Buffer;
 
 /**
@@ -88,6 +89,16 @@ public class TinkerpopBufferPrimitiveCodec implements PrimitiveCodec<Buffer> {
   }
 
   @Override
+  public void markReaderIndex(Buffer source) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public void resetReaderIndex(Buffer source) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
   public byte readByte(Buffer source) {
     return source.readByte();
   }
@@ -95,6 +106,11 @@ public class TinkerpopBufferPrimitiveCodec implements PrimitiveCodec<Buffer> {
   @Override
   public int readInt(Buffer source) {
     return source.readInt();
+  }
+
+  @Override
+  public int readInt(Buffer source, int offset) {
+    throw new UnsupportedOperationException();
   }
 
   @Override
@@ -146,6 +162,16 @@ public class TinkerpopBufferPrimitiveCodec implements PrimitiveCodec<Buffer> {
   public String readLongString(Buffer source) {
     int length = readInt(source);
     return readString(source, length);
+  }
+
+  @Override
+  public Buffer readRetainedSlice(Buffer source, int sliceLength) {
+    throw new UnsupportedOperationException();
+  }
+
+  @Override
+  public void updateCrc(Buffer source, CRC32 crc) {
+    throw new UnsupportedOperationException();
   }
 
   @Override
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
@@ -100,8 +100,10 @@ import com.datastax.oss.driver.internal.core.util.concurrent.LazyReference;
 import com.datastax.oss.driver.shaded.guava.common.base.Ticker;
 import com.datastax.oss.protocol.internal.Compressor;
 import com.datastax.oss.protocol.internal.FrameCodec;
+import com.datastax.oss.protocol.internal.PrimitiveCodec;
 import com.datastax.oss.protocol.internal.ProtocolV3ClientCodecs;
 import com.datastax.oss.protocol.internal.ProtocolV5ClientCodecs;
+import com.datastax.oss.protocol.internal.SegmentCodec;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import io.netty.buffer.ByteBuf;
@@ -166,8 +168,12 @@ public class DefaultDriverContext implements InternalDriverContext {
       new LazyReference<>("eventBus", this::buildEventBus, cycleDetector);
   private final LazyReference<Compressor<ByteBuf>> compressorRef =
       new LazyReference<>("compressor", this::buildCompressor, cycleDetector);
+  private final LazyReference<PrimitiveCodec<ByteBuf>> primitiveCodecRef =
+      new LazyReference<>("primitiveCodec", this::buildPrimitiveCodec, cycleDetector);
   private final LazyReference<FrameCodec<ByteBuf>> frameCodecRef =
       new LazyReference<>("frameCodec", this::buildFrameCodec, cycleDetector);
+  private final LazyReference<SegmentCodec<ByteBuf>> segmentCodecRef =
+      new LazyReference<>("segmentCodec", this::buildSegmentCodec, cycleDetector);
   private final LazyReference<ProtocolVersionRegistry> protocolVersionRegistryRef =
       new LazyReference<>(
           "protocolVersionRegistry", this::buildProtocolVersionRegistry, cycleDetector);
@@ -439,15 +445,23 @@ public class DefaultDriverContext implements InternalDriverContext {
     }
   }
 
+  protected PrimitiveCodec<ByteBuf> buildPrimitiveCodec() {
+    return new ByteBufPrimitiveCodec(getNettyOptions().allocator());
+  }
+
   protected FrameCodec<ByteBuf> buildFrameCodec() {
     return new FrameCodec<>(
-        new ByteBufPrimitiveCodec(getNettyOptions().allocator()),
+        getPrimitiveCodec(),
         getCompressor(),
         new ProtocolV3ClientCodecs(),
         new ProtocolV4ClientCodecsForDse(),
         new ProtocolV5ClientCodecs(),
         new DseProtocolV1ClientCodecs(),
         new DseProtocolV2ClientCodecs());
+  }
+
+  protected SegmentCodec<ByteBuf> buildSegmentCodec() {
+    return new SegmentCodec<>(getPrimitiveCodec(), getCompressor());
   }
 
   protected ProtocolVersionRegistry buildProtocolVersionRegistry() {
@@ -784,8 +798,20 @@ public class DefaultDriverContext implements InternalDriverContext {
 
   @NonNull
   @Override
+  public PrimitiveCodec<ByteBuf> getPrimitiveCodec() {
+    return primitiveCodecRef.get();
+  }
+
+  @NonNull
+  @Override
   public FrameCodec<ByteBuf> getFrameCodec() {
     return frameCodecRef.get();
+  }
+
+  @NonNull
+  @Override
+  public SegmentCodec<ByteBuf> getSegmentCodec() {
+    return segmentCodecRef.get();
   }
 
   @NonNull
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufCompressor.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufCompressor.java
@@ -25,21 +25,39 @@ public abstract class ByteBufCompressor implements Compressor<ByteBuf> {
 
   @Override
   public ByteBuf compress(ByteBuf uncompressed) {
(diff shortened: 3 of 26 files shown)


olim7t:
I'm having second thoughts about metrics. I was initially going to track a few session-level histograms:
- length (in bytes) of uncompressed payloads
- length (in bytes) of compressed payloads
- number of small frames in self-contained payloads
- number of slices that large frames are split into

If we measure incoming and outgoing separately, that's 8 additional metrics, with the corresponding entries in `reference.conf`, at least one new config option for the recording interval, etc.

Compression can already be measured indirectly: run the same load test with and without, then compare the existing `bytes-sent` and `bytes-received` metrics across the two runs.

As for the others, they would be interesting for us to assess how well v5 framing works, but they're not really actionable for a regular end user: the max payload length is hard-coded, there's no knob that you can turn to change how segments are constructed. So it would be anecdotal at best, for low-level details that users probably don't care that much about.
We can run an instrumented fork of the driver for our initial tests, but I'm not sure if those metrics are relevant for long-term inclusion.

core/src/main/java/com/datastax/dse/driver/internal/core/protocol/TinkerpopBufferPrimitiveCodec.java
```diff
@@ -87,6 +88,16 @@ public Buffer concat(Buffer left, Buffer right) {
     return rv;
   }
 
+  @Override
+  public void markReaderIndex(Buffer source) {
+    throw new UnsupportedOperationException();
```
tomekl007, line 93: Can we add some explanation to all `UnsupportedOperationException`s?

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
```diff
@@ -165,8 +167,12 @@
       new LazyReference<>("eventBus", this::buildEventBus, cycleDetector);
   private final LazyReference<Compressor<ByteBuf>> compressorRef =
       new LazyReference<>("compressor", this::buildCompressor, cycleDetector);
+  private final LazyReference<PrimitiveCodec<ByteBuf>> primitiveCodecRef =
+      new LazyReference<>("primitiveCodec", this::buildPrimitiveCodec, cycleDetector);
```
tomekl007, line 172: `primitiveCodec` was not build before?

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufCompressor.java
```diff
@@ -25,21 +25,39 @@
...
-  protected abstract ByteBuf compressDirect(ByteBuf input);
+  @Override
+  public ByteBuf compressWithoutLength(ByteBuf uncompressed) {
+    return uncompressed.isDirect()
+        ? compressDirect(uncompressed, false)
+        : compressHeap(uncompressed, false);
+  }
 
-  protected abstract ByteBuf compressHeap(ByteBuf input);
+  protected abstract ByteBuf compressDirect(ByteBuf input, boolean prependWithUncompressedLength);
```
tomekl007, line 40: nit: switch boolean to not have double negation (false - uncompressed)  - for me, it will be more readable if it will be `preparedWithCompressedLenght`

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufSegmentBuilder.java
```diff
@@ -0,0 +1,174 @@
...
+  @NonNull
+  protected ChannelPromise mergeStates(@NonNull List<ChannelPromise> framePromises) {
+    // We concatenate multiple frames into one segment. When the segment is written, all the frames
+    // are written.
+    ChannelPromise segmentPromise = context.newPromise();
+    segmentPromise.addListener(
+        future -> {
+          if (future.isSuccess()) {
+            for (ChannelPromise framePromise : framePromises) {
+              framePromise.setSuccess();
```
tomekl007, line 68: will it wait for all `framePromises`? Could you explain the flow of this merge logic for me?

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufSegmentBuilder.java
```diff
@@ -0,0 +1,174 @@
...
+        if (future.isSuccess()) {
+          remainingSlices -= 1;
+          if (remainingSlices == 0) {
+            parentPromise.setSuccess();
+          }
+        } else {
+          // No need to wait for the outcome of the other slices really. Fail immediately and we'll
+          // ignore them.
+          parentPromise.setFailure(future.cause());
+          alreadyFailed = true;
```
tomekl007: why do we need this `alreadyFailed` if we are setting `setFailure` and this works on the same eventLoop? Isn't there a way to just break the flow to prevent subsequent calls to `operationComplete`?

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/BytesToSegmentDecoder.java
```diff
@@ -0,0 +1,95 @@
...
+  public BytesToSegmentDecoder(@NonNull SegmentCodec<ByteBuf> segmentCodec) {
+    super(
+        // max length (Netty wants this to be the overall length including everything):
+        segmentCodec.headerLength()
+            + SegmentCodec.CRC24_LENGTH
+            + Segment.MAX_PAYLOAD_LENGTH
+            + SegmentCodec.CRC32_LENGTH,
+        // offset and size of the "length" field: that's the whole header
+        0,
+        segmentCodec.headerLength() + SegmentCodec.CRC24_LENGTH,
```
tomekl007, line 51: why only `segmentCodec.headerLength() + SegmentCodec.CRC24_LENGTH,` and not 
`segmentCodec.headerLength() + SegmentCodec.CRC24_LENGTH  + Segment.MAX_PAYLOAD_LENGTH  + SegmentCodec.CRC32_LENGTH`
?

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/BytesToSegmentDecoder.java
```diff
@@ -0,0 +1,95 @@
...
+        // length adjustment: add the trailing CRC to the declared length
+        SegmentCodec.CRC32_LENGTH,
+        // bytes to skip: the header (we've already parsed it while reading the length)
+        segmentCodec.headerLength() + SegmentCodec.CRC24_LENGTH);
```
tomekl007, line 55: nit: 
`segmentCodec.headerLength() + SegmentCodec.CRC24_LENGTH`
so maybe this calculation should be extracted to a simple method?

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/BytesToSegmentDecoder.java
```diff
@@ -0,0 +1,95 @@
...
+    this.segmentCodec = segmentCodec;
+  }
+
+  @Override
+  protected Object decode(ChannelHandlerContext ctx, ByteBuf in) throws Exception {
+    try {
+      ByteBuf payloadAndCrc = (ByteBuf) super.decode(ctx, in);
+      if (payloadAndCrc == null) {
+        return null;
+      } else {
```
tomekl007, line 65: maybe instead of an ` assert header != null;`
we could do `else if(header != null){}`?

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/BytesToSegmentDecoder.java
```diff
@@ -0,0 +1,95 @@
...
+        assert header != null;
+        try {
+          Segment<ByteBuf> segment = segmentCodec.decode(header, payloadAndCrc);
+          header = null;
+          return segment;
+        } catch (com.datastax.oss.protocol.internal.CrcMismatchException e) {
```
tomekl007, line 71: Why do we need two `CrcMismatchException` if both look very similar? (both are taking message)

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/SegmentToFrameDecoder.java
```diff
@@ -0,0 +1,116 @@
...
+      LOG.trace(
+          "[{}] Decoded response frame {} from {} slices",
+          logPrefix,
+          frame.streamId,
+          accumulatedSlices.size());
+      out.add(frame);
+      // Reset our state
+      targetLength = UNKNOWN_LENGTH;
+      accumulatedSlices.clear();
+      accumulatedLength = 0;
```
tomekl007, line 113: nit, maybe extract to method `resetState()`

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufCompressor.java
olim7t: The parameter does not mean "uncompressed vs. compressed", it means "prepend vs. not prepend". If a length gets prepended, it is always the uncompressed one.

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
olim7t: It was used only in one place, so we were building it directly in `buildFrameCodec`. See the diff on that method.

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufSegmentBuilder.java
olim7t: The upper layer (`InFlightHandler` to be specific) is giving us a set of frames. With each frame comes a promise that we'll have to complete to say "I'm done writing".
As it turns out, we decide to process all those frames in a single write. So we craft a single compound promise that will propagate its outcome to all the frame promises. Then we use that promise when writing (`ByteBufSegmentBuilder.processSegment`):
```java
context.write(segment, segmentPromise);
```
When Netty is done writing the segment, it will complete `segmentPromise`, and therefore all the frame promises.

It's the same logic in reverse for large frame: one frame promise becomes multiple segment promises.

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufSegmentBuilder.java
olim7t: Good point. Maybe it's possible to cancel the other promises, I'm not sure how Netty handles that. I'll try.

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/BytesToSegmentDecoder.java
olim7t: Short answer: it's the value that makes the tests pass :trollface: 
Long answer: it's how many bytes the parent class needs to read before it can invoke `getUnadjustedFrameLength`. What we do in there is `segmentCodec.decodeHeader`, which just needs the header and header CRC.

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/BytesToSegmentDecoder.java
olim7t: An assertion error signals a blatant error in the code: if the header is null at this point, there's a flaw in the algorithm.

Skipping the branch would be equivalent to ignoring that error silently.

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/BytesToSegmentDecoder.java
olim7t: One is from the native-protocol project, I don't want to surface it.

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufSegmentBuilder.java
olim7t: Yep, Netty detects cancelled promises and doesn't write out the bytes:
```java
EmbeddedChannel channel = new EmbeddedChannel();
ChannelPromise promise1 = channel.newPromise();
channel.write(1, promise1);
ChannelPromise promise2 = channel.newPromise();
channel.write(2, promise2);
promise2.cancel(false);
channel.flush();
System.out.println(channel.outboundMessages());
// [1, EmptyByteBufBE]
```
This happens in [ChannelOutboundBuffer](https://github.com/netty/netty/blob/4.1/transport/src/main/java/io/netty/channel/ChannelOutboundBuffer.java#L149-L153), so it applies to all `AbstractChannel` implementations.

Not sending the remaining slices makes sense, since the other end won't be able to reassemble the segment anyway. I'll make the change.

olim7t: Done. It's still better to handle reentry, because `operationComplete` is invoked even for a cancellation, but it occurred to me that I can check the parent promise instead of using a separate boolean.

olim7t:
I'm merging this now, for inclusion in 4.8.0. The frame/segment naming is still provisional, but we can revisit that later since it does not impact the public API (see [JAVA-2852](https://datastax-oss.atlassian.net/browse/JAVA-2852)).

PR #1459, opened 2020-06-26
JAVA-2828: Add safe paging state wrapper

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/cql/ExecutionInfo.java
```diff
@@ -110,16 +112,45 @@ default Request getRequest() {
...
+   * The paging state of the query, in a safe wrapper that checks if it's reused on the right
+   * statement.
+   *
+   * <p>This represents the next page to be fetched if this query has multiple page of results. It
+   * can be saved and reused later on the same statement.
+   *
+   * @return the paging state, or {@code null} if there is no next page.
+   */
+  @Nullable
+  default PagingState getSafePagingState() {
```
olim7t, line 138: Unfortunately we can't rename the other one.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultPagingState.java
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  @NonNull
+  @Override
+  public ByteBuffer getRawPagingState() {
+    return rawPagingState;
+  }
+
+  // Hashes a statement's query string and parameters. In retrospect we probably didn't need to
+  // include the paging state in the hash, but keeping it to have exactly the same output as driver
```
olim7t: > we probably didn't need to include the paging state in the hash

Actually I think I remember now, it provides some additional protection if a malicious actor tries to forge a serialized form. If the state is not included in the hash, you can replace it with anything and it will get accepted.
I'll amend the comment.

core/src/main/java/com/datastax/oss/driver/api/core/cql/PagingState.java
```diff
@@ -0,0 +1,75 @@
...
+  byte[] toBytes();
+
+  /**
+   * Checks if this paging state can be safely reused for the given statement. Specifically, the
+   * query string and any bound values must match.
+   *
+   * <p>Note that, if {@code statement} is a {@link SimpleStatement} with bound values, those values
+   * must be encoded in order to perform the check. This method uses the default codec registry and
+   * default protocol version. This might fail if you use custom codecs; in that case, use {@link
+   * #matches(Statement, Session)} instead.
```
olim7t, line 59: That whole thing with simple statements is annoying, it adds a bunch of complexity for a minuscule edge case. Maybe we should have attached simple statements to the session (or gotten rid of them altogether), but that would have been yet another difference compared to driver 3.

core/src/main/java/com/datastax/oss/driver/api/core/cql/PagingState.java
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * <p>This class performs additional checks to fail fast if the paging state is not reused on the
+ * same query, and it provides utility methods for conversion to/from strings and byte arrays.
+ */
+public interface PagingState {
+
+  /** Parses an instance from a string previously generated with {@code toString()}. */
+  @NonNull
+  static PagingState fromString(@NonNull String string) {
+    return DefaultPagingState.fromString(string);
```
tomekl007, line 40: It seems a bit odd that the interface is referring to the method that is on the implementation. I wonder if that dependency may introduce some problems in the future?

core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultPagingState.java
```diff
@@ -0,0 +1,160 @@
...
+  //     size of raw state|size of hash|raw state|hash|protocol version
+  //
+  // The protocol version might be absent, in which case it defaults to V2 (this is for backward
+  // compatibility with 2.0.10 where it is always absent).
+  public static DefaultPagingState fromBytes(byte[] bytes) {
+    ByteBuffer buffer = ByteBuffer.wrap(bytes);
+    short rawPagingStateLength = buffer.getShort();
+    short hashLength = buffer.getShort();
+    if (rawPagingStateLength + hashLength != buffer.remaining()
+        && rawPagingStateLength + hashLength + 2 != buffer.remaining()) {
```
tomekl007: Is this `+ 2` a case of absent protocol version?

core/src/main/java/com/datastax/oss/driver/api/core/cql/PagingState.java
olim7t: I don't think so, the public contract only states that it will return an instance of the interface, the actual class is an implementation detail.
We already do this in a number of places (`SimpleStatement.newInstance`, etc).

core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultPagingState.java
olim7t: Yes, we test two different length because the protocol version may or may not be there.

olim7t: I've rearranged the code to clarify this.

core/src/main/java/com/datastax/oss/driver/api/core/cql/Statement.java
```diff
@@ -302,6 +302,50 @@ default SelfT setRoutingKey(@NonNull ByteBuffer... newRoutingKeyComponents) {
...
+  @NonNull
+  @CheckReturnValue
+  default SelfT setPagingState(@Nullable PagingState newPagingState, @Nullable Session session) {
+    if (newPagingState == null) {
+      return setPagingState((ByteBuffer) null);
+    } else if (newPagingState.matches(this, session)) {
+      return setPagingState(newPagingState.getRawPagingState());
+    } else {
+      throw new IllegalArgumentException(
+          "Paging state mismatch, "
```
tomekl007, line 343: Is it possible to give more info about what's wrong with the non-matched paging state for the debug purpose?
(I.e. which variable is missing)

core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultPagingState.java
```diff
@@ -0,0 +1,160 @@
...
+    buffer.putShort((short) hash.length);
+    buffer.put(rawPagingState.duplicate());
+    buffer.put(hash);
+    buffer.putShort((short) protocolVersion);
+    buffer.rewind();
+    return buffer.array();
+  }
+
+  public static DefaultPagingState fromString(String string) {
+    byte[] bytes = Bytes.getArray(Bytes.fromHexString("0x" + string));
```
tomekl007, line 92: Why do we need to remove `0x` from the output?
I see that this is removed at the reading and writing side (`fromString` and `toString`)
So maybe it's better to keep it?

core/src/main/java/com/datastax/oss/driver/api/core/cql/Statement.java
olim7t: That information is not available, the hash is all we have left about the old statement.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultPagingState.java
olim7t: Short answer is we're keeping the format identical to driver 3 :) But I'd say the motivation then was to not include unnecessary information in the string form, it's supposed to be used in URLs and such, and it's already pretty long.

PR #1460, opened 2020-06-27
JAVA-2832: Ability to create uninitialized sessions

adutra:
Opening this PR just to start the conversation. All the tests pass, but I didn't create any specific tests for lazy sessions so far.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
@@ -23,6 +23,7 @@ import com.datastax.oss.driver.api.core.config.DefaultDriverOption;
 import com.datastax.oss.driver.api.core.config.DriverConfigLoader;
 import com.datastax.oss.driver.api.core.config.DriverExecutionProfile;
 import com.datastax.oss.driver.api.core.context.DriverContext;
+import com.datastax.oss.driver.api.core.cql.LazyCqlSession;
 import com.datastax.oss.driver.api.core.loadbalancing.LoadBalancingPolicy;
 import com.datastax.oss.driver.api.core.metadata.EndPoint;
 import com.datastax.oss.driver.api.core.metadata.Node;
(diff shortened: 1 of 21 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -633,73 +637,103 @@ public SessionT build() {
...
+    } catch (Throwable t) {
+      // We construct the session synchronously (until the init() call), but async clients expect a
+      // failed future if anything goes wrong. So wrap any error from that synchronous part.
+      return CompletableFutures.failedFuture(t);
+    }
+  }
 
-      ProgrammaticArguments programmaticArguments = programmaticArgumentsBuilder.build();
+  @NonNull
+  protected LazyCqlSession buildDefaultSessionLazy() {
```
adutra, line 674: `buildDefaultSessionAsync` is final, but I don't see why it should be; and making this method final would make revapi angry.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -633,73 +637,103 @@ public SessionT build() {
...
         addContactEndPoints(cloudConfig.getEndPoints());
         withLocalDatacenter(cloudConfig.getLocalDatacenter());
         withSslEngineFactory(cloudConfig.getSslEngineFactory());
         withCloudProxyAddress(cloudConfig.getProxyAddress());
         if (cloudConfig.getAuthProvider().isPresent()) {
           withAuthProvider(cloudConfig.getAuthProvider().get());
         }
-        programmaticArguments = programmaticArgumentsBuilder.build();
+      } catch (IOException e) {
+        throw new UncheckedIOException(e);
```
adutra, line 715: I'm changing this slightly so that this method does not throw checked exceptions; it seems overkill to impose to catch checked exceptions on callers of `buildLazy()`.

olim7t (changes requested):
Strong 👎  on this. I've explained my reasons on the Jira ticket.

adutra:
OK, closing.

PR #1461, opened 2020-07-01
JAVA-2837: make StringCodec strict about unicode in ascii

TvdW:
Somewhere between 3.x and 4.x the StringCodec implementation started
using String.getBytes(charset), which has the caveat of translating
unmappable characters into a charset-dependent replacement character,
which for ascii is '?'. In other words, if you were to put unicode data
into an ascii field, it will just insert a lot of question marks.

With this patch the driver will throw an InvalidArgumentException if
that happens.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/StringCodec.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/StringCodec.java
@@ -20,22 +20,47 @@ import com.datastax.oss.driver.api.core.type.DataType;
 import com.datastax.oss.driver.api.core.type.codec.TypeCodec;
 import com.datastax.oss.driver.api.core.type.reflect.GenericType;
 import com.datastax.oss.driver.internal.core.util.Strings;
-import com.datastax.oss.protocol.internal.util.Bytes;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
+import io.netty.util.concurrent.FastThreadLocal;
 import java.nio.ByteBuffer;
+import java.nio.CharBuffer;
+import java.nio.charset.CharacterCodingException;
 import java.nio.charset.Charset;
+import java.nio.charset.CharsetDecoder;
+import java.nio.charset.CharsetEncoder;
+import java.nio.charset.CodingErrorAction;
 import net.jcip.annotations.ThreadSafe;
 
 @ThreadSafe
 public class StringCodec implements TypeCodec<String> {
 
   private final DataType cqlType;
-  private final Charset charset;
+  private final FastThreadLocal<CharsetEncoder> charsetEncoder;
+  private final FastThreadLocal<CharsetDecoder> charsetDecoder;
 
   public StringCodec(@NonNull DataType cqlType, @NonNull Charset charset) {
     this.cqlType = cqlType;
-    this.charset = charset;
+    charsetEncoder =
+        new FastThreadLocal<CharsetEncoder>() {
+          @Override
+          protected CharsetEncoder initialValue() throws Exception {
+            return charset
+                .newEncoder()
+                .onMalformedInput(CodingErrorAction.REPORT)
+                .onUnmappableCharacter(CodingErrorAction.REPORT);
+          }
+        };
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/StringCodec.java
```diff
@@ -74,7 +95,11 @@ public String decode(@Nullable ByteBuffer bytes, @NonNull ProtocolVersion protoc
     } else if (bytes.remaining() == 0) {
       return "";
     } else {
-      return new String(Bytes.getArray(bytes), charset);
+      try {
+        return decoder.decode(bytes.duplicate()).toString();
```
TvdW: Not a fan of the .duplicate() here, but without it some other tests start failing. Maybe a maintainer could make the necessary additional changes in case the duplicate() is indeed unwanted

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/StringCodec.java
```diff
@@ -20,22 +20,36 @@
...
+  private final CharsetDecoder decoder;
 
   public StringCodec(@NonNull DataType cqlType, @NonNull Charset charset) {
     this.cqlType = cqlType;
-    this.charset = charset;
+    this.encoder =
+        charset
+            .newEncoder()
+            .onMalformedInput(CodingErrorAction.REPORT)
+            .onUnmappableCharacter(CodingErrorAction.REPORT);
```
TvdW: Hm, starting to think this is wrong, because these encoders aren't thread-safe. Will update the PR in a bit

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/StringCodec.java
TvdW: It was definitely wrong, so I went with the FastThreadLocal like in the TimestampCodec

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/StringCodec.java
TvdW: It was definitely wrong, so I went with FastThreadLocal like in the TimestampCodec.

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/StringCodecTest.java
```diff
@@ -78,4 +78,33 @@ public void should_accept_object() {
     assertThat(codec.accepts("hello")).isTrue();
     assertThat(codec.accepts(Integer.MIN_VALUE)).isFalse();
   }
+
+  public static class ASCIIStringCodecTest extends CodecTestBase<String> {
```
adutra: Really minor: our coding style requires this to be renamed to `AsciiStringCodecTest`.

adutra (approved):
Excellent 👍

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/StringCodecTest.java
TvdW: Makes sense. Fixed!

olim7t (approved):
👍  LGTM

I have two trivial remarks (changelog entry, and extracting the new ASCII test as a top-level class), but I'll address them in the merge directly.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/StringCodec.java
olim7t: No it's the right way to do it. The codec operates directly on the bytebuffer stored in the data structure (`Row`, `UdtValue`, etc.), and the contract is that getters don't consume the data, users should be able to call `getString`, `getBytesUnsafe`, etc, multiple times on the same column.
Other codec implementations like `ListCodec` also duplicate the buffer.

olim7t:
I don't know what's wrong with the build: everything looks fine locally, the base 4.x passes and there's nothing in this PR that should change that.
Merging anyway and I'll check how the 4.x build goes.

PR #1463, opened 2020-07-08
JAVA-2431: Set all occurrences when bound variables are used multiple times

olim7t:
Draft to see what it takes to restore the 3.x behavior. It's not as bad as I thought, see inline comments regarding the tradeoffs to avoid breaking the API.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/cql/ColumnDefinitions.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/cql/ColumnDefinitions.java
@@ -18,7 +18,10 @@ package com.datastax.oss.driver.api.core.cql;
 import com.datastax.oss.driver.api.core.CqlIdentifier;
 import com.datastax.oss.driver.api.core.data.AccessibleByName;
 import com.datastax.oss.driver.api.core.detach.Detachable;
+import com.datastax.oss.driver.internal.core.util.Loggers;
 import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Collections;
+import java.util.List;
 
 /**
  * Metadata about a set of CQL columns.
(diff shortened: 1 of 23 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/cql/ColumnDefinitions.java
```diff
@@ -97,22 +100,59 @@ default ColumnDefinition get(@NonNull CqlIdentifier name) {
...
+   *     will only return the first occurrence. Therefore it also logs a warning.
+   *     <p>Implementors should always override this method (all built-in driver implementations
+   *     do).
+   */
+  @NonNull
+  default List<Integer> allIndicesOf(@NonNull String name) {
+    Loggers.COLUMN_DEFINITIONS.warn(
+        "Implementors should always override this method, "
+            + "the default implementation only returns the first occurrence");
+    return Collections.singletonList(firstIndexOf(name));
```
olim7t, line 122: I wouldn't use this as a general solution to breaking API changes, but a number of things make me think it's OK for this case:
* this only matters for prepared and bound statements (UDTs can't have duplicate field names, and rows are not settable).
* there are probably very few (if any) custom implementations of `PreparedStatement` and `BoundStatement` outside of the driver.
* even so, the implementation is only incorrect when there are multiple occurrences of the same variable, which I still think is an edge case.
* finally, the default implementation is not completely broken, it just functions like previous 4.x releases.

pom.xml
```diff
@@ -714,6 +714,10 @@ limitations under the License.]]></inlineHeader>
           <doclint>all,-missing</doclint>
           <excludePackageNames>com.datastax.*.driver.internal*</excludePackageNames>
           <tags>
+            <tag>
+              <name>apiNote</name>
+              <placement>A</placement>
+            </tag>
```
olim7t, line 716: This is supposed to be a "semi-official" tag in Java 8, IDEA recognizes it but the build doesn't.

core/src/main/java/com/datastax/oss/driver/api/core/data/SettableById.java
```diff
@@ -59,7 +59,12 @@
   @NonNull
   @CheckReturnValue
   default SelfT setBytesUnsafe(@NonNull CqlIdentifier id, @Nullable ByteBuffer v) {
-    return setBytesUnsafe(firstIndexOf(id), v);
+    SelfT result = null;
+    for (Integer i : allIndicesOf(id)) {
+      result = (result == null ? this : result).setBytesUnsafe(i, v);
+    }
+    assert result != null; // allIndices throws if there are no results
+    return result;
```
olim7t, line 67: A couple of comments on performance:

* for bound statements (immutable), this will create a new copy for each iteration in the loop. That's fine, because repeated names are rare anyway, and if they happen it will probably be 2-3 times at most. Also `BoundStatementBuilder` is available as a workaround.
* this quirky pattern to deal with the self type is copy-pasted in every method. I could extract it to a utility method that takes a `BiFunction<SelfT, Integer, SelfT>`, but creating a capturing lambda for every method call bothers me.

integration-tests/src/test/java/com/datastax/oss/driver/core/cql/BoundStatementCcmIT.java
```diff
@@ -365,6 +366,44 @@ public void should_compute_routing_key_when_indices_randomly_distributed() {
...
+    should_set_all_occurrences_of_variable(ps.boundStatementBuilder().setInt(id, 12).build());
+  }
+
+  private void should_set_all_occurrences_of_variable(BoundStatement bs) {
+    assertThat(bs.getInt(0)).isEqualTo(12);
+    assertThat(bs.getInt(1)).isEqualTo(12);
+    assertThat(bs.getInt(2)).isEqualTo(12);
+
+    // Nothing should be shared internally (this would be a bug if the client later retrieves a
+    // buffer with getBytesUnsafe and modifies it)
```
olim7t, line 388: I want emphasize this. I think reusing the same buffer would be a mistake, but just to check everyone agrees.

core/src/main/java/com/datastax/oss/driver/api/core/cql/ColumnDefinitions.java
```diff
@@ -97,22 +100,59 @@ default ColumnDefinition get(@NonNull CqlIdentifier name) {
...
+   * @apiNote the default implementation only exists for backward compatibility. It wraps the result
+   *     of {@link #firstIndexOf(String)} in a singleton list, which is not entirely correct, as it
+   *     will only return the first occurrence. Therefore it also logs a warning.
+   *     <p>Implementors should always override this method (all built-in driver implementations
+   *     do).
+   */
+  @NonNull
+  default List<Integer> allIndicesOf(@NonNull String name) {
+    Loggers.COLUMN_DEFINITIONS.warn(
+        "Implementors should always override this method, "
```
adutra: Nit: "this method" will look a bit ambiguous in a log file, shouldn't we name it explicitly?

core/src/main/java/com/datastax/oss/driver/internal/core/util/Loggers.java
```diff
@@ -38,4 +43,10 @@ public static void warnWithException(Logger logger, String format, Object... arg
       }
     }
   }
+
+  // Loggers for API interfaces, declared here in order to keep them internal.
+  public static Logger COLUMN_DEFINITIONS = LoggerFactory.getLogger(ColumnDefinitions.class);
```
adutra, line 48: Very clever 👍

integration-tests/src/test/java/com/datastax/oss/driver/core/cql/BoundStatementCcmIT.java
```diff
@@ -365,6 +366,44 @@ public void should_compute_routing_key_when_indices_randomly_distributed() {
     }
   }
 
+  @Test
+  public void should_set_all_occurrence_of_variable() {
```
adutra:
```suggestion
  public void should_set_all_occurrences_of_variable() {
```

integration-tests/src/test/java/com/datastax/oss/driver/core/cql/BoundStatementCcmIT.java
adutra: Agreed, while using names or ids you would always modify the values of all occurrences, when using indices you still can modify just one of the occurrences.

pom.xml
adutra: This tag is not being rendered in generated javadocs, is that on purpose?

adutra (changes requested):

pom.xml
olim7t: No, I thought it would but I didn't check. I might switch back to a regular paragraph.

core/src/main/java/com/datastax/oss/driver/api/core/cql/ColumnDefinitions.java
olim7t: Agreed, my brain was in "exception message" mode, but we don't have a stacktrace for that log.

pom.xml
olim7t: OK, it just needed an additional `<head>` element.

adutra (approved):

PR #1464, opened 2020-07-10
JAVA-2808: Provide metrics bindings for Micrometer and Microprofile

emerkle826:


core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOption.java
```diff
@@ -808,6 +808,20 @@
...
+   * <p>Value-type: {@link String}
+   */
+  METRICS_FACTORY_CLASS("advanced.metrics.factory.class"),
+
+  /**
+   * The fully-qualified classname of the desired metrics registry implementation.
+   *
+   * <p>Value-type: {@link String}
+   */
+  METRICS_FACTORY_REGISTRY_CLASS("advanced.metrics.factory.registry.class"),
```
tomekl007: do we really need those two parameters? I think that `METRICS_FACTORY_CLASS` should be enough.
You wrote:
```
https://github.com/datastax/java-driver/pull/1464/files#diff-8b8ff589cb369ea902b39dcf4601a449R1242-R1245
```
could you elaborate more on that?

metrics/micrometer/src/main/java/com/datastax/oss/driver/metrics/micrometer/MicrometerSessionMetricUpdater.java
```diff
@@ -0,0 +1,130 @@
...
+                ++count;
+              }
+            }
+            return count;
+          });
+    }
+    if (enabledMetrics.contains(DefaultSessionMetric.THROTTLING_QUEUE_SIZE)) {
+      this.registry.gauge(
+          buildFullName(DefaultSessionMetric.THROTTLING_QUEUE_SIZE, null),
+          context,
```
tomekl007: nit: refactor to a separate method

metrics/micrometer/src/main/java/com/datastax/oss/driver/metrics/micrometer/MicrometerSessionMetricUpdater.java
```diff
@@ -0,0 +1,130 @@
...
+            LOG.warn(
+                "[{}] Metric {} does not support {}, it will always return 0",
+                logPrefix,
+                DefaultSessionMetric.THROTTLING_QUEUE_SIZE.getPath(),
+                requestThrottler.getClass().getName());
+            return 0;
+          });
+    }
+    if (enabledMetrics.contains(DefaultSessionMetric.CQL_PREPARED_CACHE_SIZE)) {
+      this.registry.gauge(
```
tomekl007: nit: refactor to a separate method

metrics/microprofile/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+  <parent>
+    <groupId>com.datastax.oss</groupId>
+    <artifactId>java-driver-parent</artifactId>
+    <version>4.8.0-SNAPSHOT</version>
+  </parent>
+  <artifactId>java-driver-metrics-microprofile</artifactId>
+  <packaging>bundle</packaging>
+  <name>DataStax Java driver for Apache Cassandra(R) - Metrics - microprofile</name>
+  <properties>
+    <microprofile.version>2.3</microprofile.version>
```
tomekl007: shouldn't it work for provided dependencies to not hardcode versions of the library in this module? (same for micro profile)

tomekl007 (commented):
unit and integration tests are missing

core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOption.java
emerkle826: My explanation is long, the TL;DR is this:
I wanted to try to avoid having a hard-coded metrics registry implementation for MicroProfile (i.e. smallrye). I'm not sure it's worth the effort.

Longer explanation:
I added this to allow for configuring a class that could be instantiated as a metric registry, mostly for the case of MicroProfile. Each metrics framework (DropWizard, Micrometer and MicroProfile) has a registry for holding the collected metrics, and each requires access to that registry in order to create the desired metrics.

In the case of DropWizzard and Micrometer, a default or global registry is provided with the framework, so the MetricsFactory implementation for those frameworks does not need a configured registry class.

DropWizard: https://github.com/datastax/java-driver/pull/1464/files#diff-a204e0c3b6601fba49c1b8f3414563cbR69

Micrometer: https://github.com/datastax/java-driver/pull/1464/files#diff-7823a53975fd907c9990f1e6a327fd56R71

But for MicroProfile, there is no included concrete implementation of a MeterRegistry. We use Smallye for the implementation, but there are other libraries that could be used. Currently, I have the registry hardcoded here:

https://github.com/datastax/java-driver/pull/1464/files#diff-c3b4e1da8e315c17af71d25d32983dafR71

However, if the registry class is configurable, I can remove the Smallrye dependency and allow users to configure whatever registry they want to use. But as I'm typing this explanation, I'm realizing that if we want this MicroProfile module to be drop-in, it will need an implementation of the registry, so smallrye will have to be included.

So the only reason to keep the registry class configurable is to allow for different implementations of a given metrics framework's registry, which may not really be applicable outside of MicroProfile anyway.

metrics/microprofile/pom.xml
emerkle826: Maven won't compile if I remove the dependency version. I can change it to a range and it will pick the latest.

core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOption.java
olim7t: I agree that two options sounds a bit complicated. Maybe there should be one factory per available MicroProfile implementation, e.g. `MicroprofileSmallRyeMetricsFactory implements MetricsFactory`. Are there any other implementation than SmallRye out there, and is it likely that we'd need to support them?

olim7t (commented):
Not much to say on the implementation.

On the principle, exposing this in the config bothers me: the config is "user land" and this feels too low level. I think what application developers will need most of the time is to send metrics data to a particular destination: collectd, graphite... For that you have dropwizard reporters, there's no need to replace the whole implementation. There's already a lot of stuff in `reference.conf` (one could probably argue too much).

Replacing the metrics library is for framework developers (Quarkus, Spring boot). This is a much smaller audience. I think it's fair to require the changes to be a bit more involved.

So maybe we could provide the `MetricsFactory` implementations, but in order to install them you still have to extend the context.

core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOption.java
emerkle826: > Are there any other implementation than SmallRye out there, and is it likely that we'd need to support them?

I think this is the list: https://wiki.eclipse.org/MicroProfile/Implementation#MP_Metrics_implementations
So just 2 others that implement the latest version of MicroProfile metrics.

emerkle826: So it seems the other two MicroProfile implementations are other apps, not necessarily libraries intended to be used as a metrics implementation. So, it looks like smallrye is the only one we will need to handle for now.

metrics/micrometer/src/test/java/com/datastax/oss/driver/metrics/micrometer/MicrometerMetricsFactoryIT.java
```diff
@@ -0,0 +1,91 @@
...
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(ParallelizableTests.class)
+public class MicrometerMetricsFactoryIT {
+
+  @ClassRule public static final CcmRule CCM_RULE = CcmRule.getInstance();
+  private static final MeterRegistry METER_REGISTRY = new SimpleMeterRegistry();
+
```
tomekl007: I think that the IT test should be more extensive and validate all possible metrics types.
Please see the test in the spring starter

metrics/microprofile/src/test/java/com/datastax/oss/driver/netrics/microprofile/MicroProfileMetricsFactoryIT.java
```diff
@@ -0,0 +1,91 @@
...
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(ParallelizableTests.class)
+public class MicroProfileMetricsFactoryIT {
+
+  @ClassRule public static final CcmRule CCM_RULE = CcmRule.getInstance();
+  private static final MetricRegistry METRIC_REGISTRY = new MetricsRegistryImpl();
+
```
tomekl007: Same here, the IT test does not validate all metric types.
It is beneficial to have such a test - we may catch some errors.

metrics/micrometer/src/main/java/com/datastax/oss/driver/metrics/micrometer/MicrometerDriverContext.java
```diff
@@ -0,0 +1,42 @@
...
+package com.datastax.oss.driver.metrics.micrometer;
+
+import com.datastax.oss.driver.api.core.config.DriverConfigLoader;
+import com.datastax.oss.driver.api.core.context.DriverContext;
+import com.datastax.oss.driver.api.core.metrics.MetricsFactory;
+import com.datastax.oss.driver.api.core.session.ProgrammaticArguments;
+import com.datastax.oss.driver.internal.core.context.DefaultDriverContext;
+import io.micrometer.core.instrument.MeterRegistry;
+
+/** Implementation of {@link DriverContext} that provides for a Micrometer {@link MeterRegistry}. */
```
tomekl007: The `DefaultDriverContext` is using `Dropwizard` metrics. 
Maybe to make it consistent we should create a `DropwizardDriverContext extends DefaultDriverContext ` to provide our users with a consistent API for using metrics:
- `DropwizardDriverContext`
- `MicrometerDriverContext`
- `MicroProfileDriverContext`
If not, then we should at least document that.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricsFactory.java
```diff
@@ -49,9 +50,9 @@
   @Nullable private final Metrics metrics;
   private final SessionMetricUpdater sessionUpdater;
 
-  public DropwizardMetricsFactory(InternalDriverContext context) {
+  public DropwizardMetricsFactory(DriverContext context) {
     this.logPrefix = context.getSessionName();
-    this.context = context;
+    this.context = (InternalDriverContext) context;
```
emerkle826: This change can be reverted as we're not exposing a config option to instantiate a MetricsFactory

margelatu:
> Not much to say on the implementation.
> 
> On the principle, exposing this in the config bothers me: the config is "user land" and this feels too low level. I think what application developers will need most of the time is to send metrics data to a particular destination: collectd, graphite... For that you have dropwizard reporters, there's no need to replace the whole implementation. There's already a lot of stuff in `reference.conf` (one could probably argue too much).
> 
> Replacing the metrics library is for framework developers (Quarkus, Spring boot). This is a much smaller audience. I think it's fair to require the changes to be a bit more involved.
> 
> So maybe we could provide the `MetricsFactory` implementations, but in order to install them you still have to extend the context.

As a longtime user of Dropwizard Metrics and Micrometer, I cannot fully agree with this assessment. Yes, most of the time an application developer is interested in making sure the metrics get to a particular destination, whether it is Graphite, Prometheus or something else. However, this doesn't mean that using DW or Micrometer to perform that job is equivalent.

DW Metrics performs client-side aggregation; these metrics cannot be aggregated in the metrics back-end to provide a view across hosts / containers -- the results would be statistical nonsense.

Metric libraries that support server-side aggregation, such as Micrometer, provide users with the ability to collect data in an intermediate, non-aggregated form which can then be used in the metrics backend to perform aggregation.

My point is that even app developers should have the ability to choose, since the implementations are not equivalent in their output.

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerMetricsFactory.java
```diff
@@ -0,0 +1,124 @@
...
+      this.sessionUpdater = NoopSessionMetricUpdater.INSTANCE;
+    } else {
+      this.registry = registry;
+      this.sessionUpdater =
+          new MicrometerSessionMetricUpdater(enabledSessionMetrics, this.registry, this.context);
+    }
+  }
+
+  @Override
+  public Optional<com.datastax.oss.driver.api.core.metrics.Metrics> getMetrics() {
```
adutra: Any reason to not import `Metrics`?

metrics/microprofile/src/main/java/com/datastax/oss/driver/internal/metrics/microprofile/MicroProfileMetricUpdater.java
```diff
@@ -0,0 +1,87 @@
...
+import com.datastax.oss.driver.internal.core.metrics.MetricUpdater;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import net.jcip.annotations.ThreadSafe;
+import org.eclipse.microprofile.metrics.MetricRegistry;
+
+@ThreadSafe
+public abstract class MicroProfileMetricUpdater<MetricT> implements MetricUpdater<MetricT> {
+
+  public static final String CASSANDRA_METRICS_PREFIX = "cassandra";
```
adutra: Why this? isn't the metrics prefix computed in subclasses enough?

metrics/microprofile/src/test/java/com/datastax/oss/driver/api/netrics/microprofile/MicroProfileMetricsTest.java
```diff
@@ -0,0 +1,213 @@
...
+import org.eclipse.microprofile.metrics.Metric;
+import org.eclipse.microprofile.metrics.MetricID;
+import org.eclipse.microprofile.metrics.MetricRegistry;
+import org.eclipse.microprofile.metrics.Timer;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(ParallelizableTests.class)
+public class MicroProfileMetricsTest {
```
adutra: The tests are nice, however they are integration tests. Following our conventions, they should be moved to the integration tests module.
I wonder how to add true unit tests here. The Dropwizard factory is only tested indirectly.

metrics/micrometer/src/main/java/com/datastax/oss/driver/api/metrics/micrometer/MicrometerMetricsSessionBuilder.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Override
+  protected CqlSession wrap(CqlSession defaultSession) {
+    return defaultSession;
+  }
+
+  @Override
+  protected DriverContext buildContext(
+      DriverConfigLoader configLoader, ProgrammaticArguments programmaticArguments) {
+    return new MicrometerDriverContext(configLoader, programmaticArguments, registry);
```
adutra: What happens if `registry` is null? Afaict we'll have a NPE later on. I think it's better to throw the NPE here. Same for Microprofile.

adutra (changes requested):
Looking good, I will push a few minor fixes directly.

metrics/micrometer/src/main/java/com/datastax/oss/driver/api/metrics/micrometer/MicrometerMetricsSessionBuilder.java
adutra: If the registry is required to be non-null, maybe it's better to inject it in the constructor.

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerMetricsFactory.java
emerkle826: I think at one point I was also referencing Micrometer's `Metrics` class and already had that imported. So one of them was going to need to be fully qualified. It doesn't look like I am referencing Micrometer's class any more, so I can change this now.

metrics/microprofile/src/main/java/com/datastax/oss/driver/internal/metrics/microprofile/MicroProfileMetricUpdater.java
emerkle826: This was just the pattern I followed from what we did with the quarkus extension. I can remove it. Just to clarify, removing this would result in metrics paths going from something like:
```
cassandra.s0.cql-requests
```
to something like:
```
s0.cql-requests
```

adutra: I see. Is it possible to let the user "nest" the driver metrics inside a general prefix of their choice?

metrics/micrometer/src/main/java/com/datastax/oss/driver/api/metrics/micrometer/MicrometerMetricsSessionBuilder.java
emerkle826: Good catch, I'll fix it.

metrics/microprofile/src/test/java/com/datastax/oss/driver/api/netrics/microprofile/MicroProfileMetricsTest.java
emerkle826: I'll move these to the integration test module. True unit tests were difficult to design as the metrics are pegged throughout the driver code with access to the Session and Node Updater classes. Making a unit test that simulates that is just testing the metric/meter marking code, so it doesn't provide much value. I'll try to come up with something that provides at least a little value.

metrics/microprofile/src/main/java/com/datastax/oss/driver/internal/metrics/microprofile/MicroProfileMetricUpdater.java
emerkle826: It is to the extent that a user application can override the Session name. But this implementation is just a fixed static prefix of "cassandra".

metrics/micrometer/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+      <groupId>com.datastax.oss</groupId>
+      <artifactId>java-driver-core</artifactId>
+      <exclusions>
+        <exclusion>
+          <groupId>io.dropwizard.metrics</groupId>
+          <artifactId>metrics-core</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
```
adutra, line 56: A few unnecessary dependencies here, I will fix myself.

metrics/micrometer/src/test/java/com/datastax/oss/driver/internal/metrics/micrometer/LoggerTest.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.mockito.Mockito.mock;
+
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.Appender;
+import org.mockito.ArgumentCaptor;
+import org.slf4j.LoggerFactory;
+
+public class LoggerTest {
```
adutra: Not required, you can use the one from driver-core. I will fix myself.

metrics/microprofile/pom.xml
adutra: You can also use `RELEASE`, but usually people prefer a fixed version because variable versions or version ranges tend to produce non-repeatable builds.

metrics/microprofile/src/test/java/com/datastax/oss/driver/internal/metrics/microprofile/LoggerTest.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.mockito.Mockito.mock;
+
+import ch.qos.logback.classic.Level;
+import ch.qos.logback.classic.Logger;
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.Appender;
+import org.mockito.ArgumentCaptor;
+import org.slf4j.LoggerFactory;
+
+public class LoggerTest {
```
adutra: Same here.

pom.xml
```diff
@@ -396,6 +399,11 @@
         <artifactId>svm</artifactId>
         <version>${graalapi.version}</version>
       </dependency>
+      <dependency>
```
adutra, line 401: Now that you declared smallrye here, I think it would make sense to also declare microprofile and micrometer here. I will fix myself.

adutra (commented):
Looking really good, and thanks for adapting to JAVA-2331 so quickly.

adutra (approved):

adutra:
> My point is that even app developers should have the ability to choose, since the implementations are not equivalent in their output.

Thanks for this comment, personally I must say it was very insightful for me. 

I suggest that we go forward with the current approach as we are running out of time for the next release. I will also log a follow-up ticket to make using alternative metrics frameworks a bit more user-friendly than it currently is.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricsFactory.java
```diff
@@ -136,36 +128,4 @@ public NodeMetricUpdater newNodeUpdater(Node node) {
       return dropwizardNodeMetricUpdater;
     }
   }
-
-  protected Set<SessionMetric> parseSessionMetricPaths(List<String> paths) {
```
olim7t, line 140: Nit: these overridable methods were a hook to add new metrics by subclassing. In particular this is how the DSE driver did it when it was a separate project.
Now that OSS and DSE are unified this is less of a concern, but maybe still nice to have. We can keep the protected methods but simply have them delegate to the utility class.

metrics/micrometer/revapi.json
```diff
@@ -0,0 +1,22 @@
...
+            "com\\.datastax\\.oss\\.simulacron(\\..+)?",
+            // Don't re-check sibling modules that this module depends on
+            "com\\.datastax\\.(oss|dse)\\.driver\\.api\\.core(\\..+)?",
+            "com\\.datastax\\.(oss|dse)\\.driver\\.api\\.querybuilder(\\..+)?"
+          ]
+        }
+      }
+    }
+  }
+}
```
olim7t: The two new modules only contain internal classes, so this won't match anything. We can simply skip the plugin in `pom.xml`:

```xml
      <plugin>
        <groupId>org.revapi</groupId>
        <artifactId>revapi-maven-plugin</artifactId>
        <configuration>
          <skip>true</skip>
        </configuration>
      </plugin>
```

pom.xml
```diff
@@ -77,6 +79,9 @@
     <apacheds.version>2.0.0-M19</apacheds.version>
     <surefire.version>2.22.2</surefire.version>
     <graalapi.version>20.0.0</graalapi.version>
+    <micrometer.version>1.5.0</micrometer.version>
+    <microprofile.version>2.3</microprofile.version>
+    <smallrye.version>2.4.2</smallrye.version>
```
olim7t: Nit: these properties are only used once, they can be inlined.

olim7t (approved):
LGTM, a couple of trivial remarks.

adutra:
@margelatu I created [JAVA-2855](https://datastax-oss.atlassian.net/browse/JAVA-2855) to improve the user experience around switching metrics frameworks. Please vote for it and/or comment. Thanks!

PR #1465, opened 2020-07-13
JAVA-2589: Gracefully handle TCP backpressure

tomekl007:
todo:
- [x] resume `runOnEventLoop()` when `channelWritabilityChanged` is triggered
- [x] more IT tests
- [x] performance tests?


tomekl007:
I am opening this for a discussion (please ignore `System.out.println()` for now).
I think that the best place for checking `isWritable` and doing TCP flow control is `DefaultWriteCoalescer.java`.
I am using node-js PR as a base for this: https://github.com/riptano/nodejs-dse-driver/pull/60/files#diff-e226e5d82736c8762c5caa6b825109b7R139

The crucial information needed for testing that I need to extract from `Coalescer#Flusher` is `writeQueueSize`:
https://github.com/datastax/java-driver/pull/1465/files#diff-fdd11ed2581b19e1b4ad3623efec6131R72-R75

I've created an IT test (https://github.com/datastax/java-driver/pull/1465/files#diff-e9c5dac42053ea49210fc42d6ff7ec55R49) that:
- start N big size requests
- pause the nodes in a cluster to simulate that the tcp buffer is full
- validates that write queue size is growing (https://github.com/datastax/java-driver/pull/1465/files#diff-e9c5dac42053ea49210fc42d6ff7ec55R83) since requests are buffered in the memory (in the `Coalescer`)
- resume simuacron cluster and verifies that all buffered requests are processed.

The `runOnEventLoop` is resumed when the new event is arriving and the `isWritable` returns false. Because of that to make the coalescer `flush()` all requests I need to send one artificial request in the test:
https://github.com/datastax/java-driver/pull/1465/files#diff-e9c5dac42053ea49210fc42d6ff7ec55R165-R167

This obviously handles only one case: when the new event arrives.
The 2nd case that we need to handle is to capture the signal send from netty `channelWritabilityChanged()` and restart the `runOnEventLoop`. This is problematic because `channelWritabilityChanged` is available on the `https://netty.io/4.1/api/io/netty/channel/ChannelInboundHandler.html#channelWritabilityChanged-io.netty.channel.ChannelHandlerContext-` https://netty.io/4.1/api/io/netty/channel/ChannelInboundHandler.html#channelWritabilityChanged-io.netty.channel.ChannelHandlerContext-, not on the netty `Channel` class.

We need somehow propagate the signal from one of our handlers (maybe `InFlightHandler.java` https://github.com/datastax/java-driver/pull/1465/files#diff-49203fa42e5be83a4abc84b27fe0ea7cR94-R100) to `DriverChannel` and `DefaultWriteCoalescer`. Is it possible in our current design? Can we pass the `DriverChannel` to the `InFlightHandler` constructor and restart the event loop from there? wdyt?

core/src/main/java/com/datastax/oss/driver/internal/core/channel/DefaultWriteCoalescer.java
```diff
@@ -84,8 +94,10 @@ private Flusher(EventLoop eventLoop) {
 
     private void enqueue(Write write) {
       boolean added = writes.offer(write);
+      System.out.println("writes size: " + writes.size() + " for Flusher: " + this);
       assert added; // always true (see MpscLinkedAtomicQueue implementation)
-      if (running.compareAndSet(false, true)) {
+      if (write.isWritable() && running.compareAndSet(false, true)) {
+        System.out.println("runOnEventLoop#100");
         eventLoop.execute(this::runOnEventLoop);
```
olim7t: The problem is that there is no rescheduling here: if `!write.isWritable()`, the writes will simply be left pending.

Of course if another request is executed after the channel becomes writable again, it will dequeue all the accumulated writes. But this is not guaranteed to happen; if it doesn't the requests will hang forever.

olim7t (changes requested):
I wonder if we shouldn't simply fail `Write.writePromise` on every element when the channel is not writable.

If a node is overloaded, it's better to retry the query immediately on another node, rather than waiting for the overloaded node. If we fail the promise, the driver already retries by default (for example `CqlRequestHandler.NodeResponseCallback.operationComplete`).

That won't cover the case where *all* nodes are overloaded (the queries will fail with `AllNodesFailedException`), but do we really want the driver to keep enqueuing requests in such a scenario?

core/src/main/java/com/datastax/oss/driver/internal/core/channel/DefaultWriteCoalescer.java
tomekl007: yes, that is why we need to register on the `channelWritabilityChanged()` callback - it signals that `isWritable()` changed; in such a case we should start the event loop

tomekl007:
> I wonder if we shouldn't simply fail `Write.writePromise` on every element when the channel is not writable.
> 
> If a node is overloaded, it's better to retry the query immediately on another node, rather than waiting for the overloaded node. If we fail the promise, the driver already retries by default (for example `CqlRequestHandler.NodeResponseCallback.operationComplete`).
> 
> That won't cover the case where _all_ nodes are overloaded (the queries will fail with `AllNodesFailedException`), but do we really want the driver to keep enqueuing requests in such a scenario?

Looking at the existing implementation in nodeJs and debate about TCP flow control I think that this is a different solution that we decided to abandon (see: `With overload errors, the driver will retry on next host until the query plan is exhausted and a NoHostAvailableException is thrown`.)

From what I observed in tests I see that it won't be an ideal solution because `channel.isWritable()` returns true quite often. Usually, the underlying TCP buffer is full and gets back to processing very fast - failing such a request and re-executing it on a different node may have a negative performance impact.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/DefaultWriteCoalescer.java
tomekl007: I've propagated the `channelWritabilityChanged` in the: https://github.com/datastax/java-driver/pull/1465/commits/341e7d7927e6f0021094c1324386f94d998e4261 - please take a look. The requests are no left pending - when the buffer is able to accept the messages, the event loop is restarted

core/src/main/java/com/datastax/oss/driver/internal/core/channel/InFlightHandler.java
```diff
@@ -91,6 +95,21 @@ public void channelActive(ChannelHandlerContext ctx) throws Exception {
...
+  @Override
+  public void channelWritabilityChanged(ChannelHandlerContext ctx) throws Exception {
+    System.out.println("Channel writability changed " + ctx.channel().isWritable());
+    if (ctx.channel().isWritable()) {
+      driverChannel.thenAccept(
+          d -> {
+            WriteCoalescer writeCoalescer = d.getWriteCoalescer();
+            if (writeCoalescer instanceof DefaultWriteCoalescer) {
+              ((DefaultWriteCoalescer) writeCoalescer).restartFlushers();
+            }
```
tomekl007: It seems that this TCP needs to be implemented only in the `DefaultWriteCoalescer`.
In such a case, we don't need to change the interface of the flusher

olim7t:
> `channel.isWritable()` returns true quite often. Usually, the underlying TCP buffer is full and gets back to processing very fast - failing such a request and re-executing it on a different node may have a negative performance impact.

OK, then it makes sense to enqueue at the channel level.

But I don't think we need to implement it ourselves, Netty already does it. From the javadoc of [Channel.isWritable](https://netty.io/4.1/api/io/netty/channel/Channel.html#isWritable--):

> Any write requests made when this method returns false are queued until the I/O thread is ready to process the queued write requests.

tomekl007:
> > `channel.isWritable()` returns true quite often. Usually, the underlying TCP buffer is full and gets back to processing very fast - failing such a request and re-executing it on a different node may have a negative performance impact.
> 
> OK, then it makes sense to enqueue at the channel level.
> 
> But I don't think we need to implement it ourselves, Netty already does it. From the javadoc of [Channel.isWritable](https://netty.io/4.1/api/io/netty/channel/Channel.html#isWritable--):
> 
> > Any write requests made when this method returns false are queued until the I/O thread is ready to process the queued write requests.

Does it mean that TCP flow control is already supported by the netty? 
I wonder if using `Channel.isWritable` explicitly and buffering in our code (coalescer) gives us any benefit comparing to the netty buffering?  

>  when this method returns false are queued 

Assuming that we will rely on the default behavior, do you know how to extract this queue from netty? (at least size of the queue tot validate the correctness)

olim7t:
> Does it mean that TCP flow control is already supported by the netty?
> I wonder if using `Channel.isWritable` explicitly and buffering in our code (coalescer) gives us any benefit comparing to the netty buffering?

It depends what we mean by flow control. Netty's behavior is to buffer the writes, but if the channel does not become writable in time it will OOM, there is no backpressure. From what I understand there wouldn't be any backpressure in our solution either, so I don't see any advantage reimplementing it outside of Netty.

> do you know how to extract this queue from netty? (at least size of the queue tot validate the correctness)

I'm not sure, try to check how the event loop's pending task count evolves when the channel becomes non writable:
```java
((SingleThreadEventExecutor) channel.eventLoop()).pendingTasks()
```
But maybe it happens at a lower level, we might have to dig into the Netty code to find out.

tomekl007:
> 
> But maybe it happens at a lower level, we might have to dig into the Netty code to find out.

The `pendingTasks` queue does not grow so it seems that this is not the best way to test the behavior.
From my debugging it seems that the best method to validate buffer size is:
`channel.unsafe().outboundBuffer().size()`

To validate that logic I wrapped `channel.flush()` in debug logs:
```
  for (Channel channel : channels) {
        System.out.println("1.buffer: " + channel.unsafe().outboundBuffer().size());
        System.out.println("1.pending tasks:" + ((SingleThreadEventLoop) eventLoop).pendingTasks());
        channel.flush();
        System.out.println("2.buffer: " + channel.unsafe().outboundBuffer().size());
        System.out.println("2.pending tasks:" + ((SingleThreadEventLoop) eventLoop).pendingTasks());
      }
```
With the custom solution (`isWritable` and `channelWritabilityChanged`), the output is following:
```
1.buffer: 0
1.pending tasks:0
1.buffer: 0
1.pending tasks:0
1.buffer: 0
1.pending tasks:0
2.buffer: 0
2.pending tasks:0
2.buffer: 0
2.pending tasks:0
2.buffer: 0
2.pending tasks:0
```
so the buffer is always empty, meaning that buffering takes place in our code.
When I did remove the custom buffering code, the output is different:
```
1.buffer: 0
2.buffer: 655
1.buffer: 0
1.buffer: 0
2.buffer: 654
2.buffer: 656
```
so, at the beginning the buffer is empty, but once the channel is not writable it becomes nonempty - meaning that the buffering happens in the netty code. 

I am able to catch this behavior in the IT test. I've changed the `getWriteQueueSize` to return `channel.unsafe().outboundBuffer().size`:
https://github.com/datastax/java-driver/pull/1465/files#diff-e9c5dac42053ea49210fc42d6ff7ec55R223-R234

So it seems that we don't need to change any production code logic. We can only leave IT tests that are validating the netty buffering.
@olim7t @adutra let me know wdyt.

integration-tests/src/test/java/com/datastax/oss/driver/core/TCPFlowControlIT.java
```diff
@@ -0,0 +1,253 @@
...
+    ByteBuffer buffer10Kb = ByteBuffer.wrap(buffer);
+
+    String queryString =
+        String.format("INSERT INTO table1 (id) VALUES (%s)", ByteUtils.toHexString(buffer10Kb));
+    Query query = new Query(queryString, emptyList(), emptyMap(), emptyMap());
+
+    SIMULACRON_RULE.cluster().prime(when(query).then(noRows()));
+
+    try (CqlSession session =
+        SessionUtils.newSession(SIMULACRON_RULE, SessionUtils.configLoaderBuilder().build())) {
```
adutra: I _think_ this can be simplified to `SessionUtils.newSession(SIMULACRON_RULE)`.

integration-tests/src/test/java/com/datastax/oss/driver/core/TCPFlowControlIT.java
```diff
@@ -0,0 +1,253 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.stream.Collectors;
+import org.awaitility.Awaitility;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
+@Category(IsolatedTests.class)
+public class TCPFlowControlIT {
```
adutra, line 58: I wouldn't mind keep these tests – they are very nicely designed btw – but:

1. We are testing rather Netty than the driver here;
2. I wonder how deterministic this test really is? I think it could fail in many ways depending on how the underlying socket is implemented. I'm a bit worried that this test will be flaky forever.

tomekl007: 1. I don't have strong opinion here. But since the netty is the core of our driver, sometimes it may be beneficial to test the feature of netty that we strongly rely on. Should the netty change the behavior of TCP buffering, we will notice it immediately. I see it a bit similar to performance tests that are we are running if the netty version is incremented. Here, we are not testing performance directly but another important aspect like buffering (it can have impact on performance as well). So maybe it is beneficial to have it automated via IT?
2. I run them a lot of times and they are constantly passing. The tests are also green on Jenkins. I didn't observe the flaky behavior of those ITs.

adutra: Good arguments. Let's have the test then, and if we ever notice any flakiness we can revisit this later.

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/channel/DriverChannel.java
```diff
@@ -271,6 +271,10 @@ public String toString() {
     return channel.toString();
   }
 
+  public Channel getChannel() {
```
olim7t, line 274: Can you annotate this with `@VisibleForTesting`?

integration-tests/src/test/java/com/datastax/oss/driver/core/TCPFlowControlIT.java
```diff
@@ -0,0 +1,251 @@
...
+      SIMULACRON_RULE.cluster().node(2).pauseRead();
+
+      List<CompletionStage<AsyncResultSet>> pendingRequests = new ArrayList<>();
+      for (int i = 0; i < NUMBER_OF_SUBMITTED_REQUESTS; i++) {
+        pendingRequests.add(session.executeAsync(SimpleStatement.newInstance(queryString)));
+      }
+
+      // Non-paused nodes should process the requests correctly
+      for (Node n : nonPausedNodes) {
+        Awaitility.await()
```
olim7t: Nit: you can static import this.

integration-tests/src/test/java/com/datastax/oss/driver/core/TCPFlowControlIT.java
```diff
@@ -0,0 +1,251 @@
...
+              .outboundBuffer()
+              .size();
+    }
+    return writeQueueSize;
+  }
+
+  private void waitForWriteQueueToStabilize(CqlSession cqlSession) throws InterruptedException {
+    final Integer[] lastWriteQueueValue = {getWriteQueueSize(cqlSession)};
+    // initial delay
+    Thread.sleep(500);
```
olim7t: Isn't that also done by the `await().pollDelay()` below?

integration-tests/src/test/java/com/datastax/oss/driver/core/TCPFlowControlIT.java
```diff
@@ -0,0 +1,251 @@
...
+      assertThat(writeQueueSize).isGreaterThan(1);
+      SIMULACRON_RULE.cluster().stop();
+    }
+  }
+
+  @SuppressWarnings("unchecked")
+  private int getInFlightRequests(CqlSession session, Node n) {
+    return ((Gauge<Integer>)
+            session.getMetrics().get().getNodeMetric(n, DefaultNodeMetric.IN_FLIGHT).get())
+        .getValue();
```
olim7t: You can pass an explicit type argument to `getNodeMetric` call to avoid the unchecked cast. Also using the optional API:
```suggestion
    return session
        .getMetrics()
        .flatMap(metrics -> metrics.<Gauge<Integer>>getNodeMetric(n, DefaultNodeMetric.IN_FLIGHT))
        .map(Gauge::getValue)
        .orElseThrow(() -> new AssertionError("Expected in-flight metric to be enabled"));
```

integration-tests/src/test/java/com/datastax/oss/driver/core/TCPFlowControlIT.java
```diff
@@ -0,0 +1,251 @@
...
+
+  @Test
+  public void should_not_write_more_requests_to_the_socket_after_the_server_paused_reading()
+      throws InterruptedException, ExecutionException, TimeoutException {
+    byte[] buffer = new byte[10240];
+    Arrays.fill(buffer, (byte) 1);
+    ByteBuffer buffer10Kb = ByteBuffer.wrap(buffer);
+
+    String queryString =
+        String.format("INSERT INTO table1 (id) VALUES (%s)", ByteUtils.toHexString(buffer10Kb));
```
olim7t: These 4 lines are repeated at the beginning of every test, you can extract them to a constant. I think this will produce the same string:
```java
  private static final String QUERY_STRING =
      String.format("INSERT INTO table1 (id) VALUES (0x%s)", Strings.repeat("01", 10240));
```

integration-tests/src/test/java/com/datastax/oss/driver/core/TCPFlowControlIT.java
```diff
@@ -0,0 +1,251 @@
...
+    Query query = new Query(queryString, emptyList(), emptyMap(), emptyMap());
+
+    SIMULACRON_RULE.cluster().prime(when(query).then(noRows()));
```
olim7t, line 72: This is also repeated in every test. You could move it to a `@BeforeClass` method.

integration-tests/src/test/java/com/datastax/oss/driver/core/TCPFlowControlIT.java
```diff
@@ -0,0 +1,251 @@
...
+      List<CompletionStage<AsyncResultSet>> pendingRequests = new ArrayList<>();
+      for (int i = 0; i < NUMBER_OF_SUBMITTED_REQUESTS; i++) {
+        pendingRequests.add(session.executeAsync(SimpleStatement.newInstance(queryString)));
+      }
+
+      // Assert that there are still requests that haven't been written
+      waitForWriteQueueToStabilize(session);
+      assertThat(getWriteQueueSize(session)).isGreaterThan(1);
+
+      SIMULACRON_RULE.cluster().resumeRead();
```
olim7t, line 94: I was wondering about the risk of query timeouts while the server is paused. But I see that the integration-test module's `application.conf` sets `request.timeout` to 10 seconds, so I think we're good.

integration-tests/src/test/java/com/datastax/oss/driver/core/TCPFlowControlIT.java
```diff
@@ -0,0 +1,251 @@
...
+
+      int numberOfFinished = 0;
+      for (CompletionStage<AsyncResultSet> result : pendingRequests) {
+        result.toCompletableFuture().get(1, TimeUnit.SECONDS);
+        numberOfFinished = numberOfFinished + 1;
```
olim7t: Nit: you don't need to count the number of futures, the code will throw if one of them is failed or still incomplete. If you reach the final assertion it's guaranteed to succeed.

Also, `com.datastax.oss.driver.Assertions` has an `assertThatStage` with a few convenient methods. Here you could use `isSuccess`, which already has a built-in 2-second timeout:

```java
assertThatStage(result).isSuccess();
```

integration-tests/src/test/java/com/datastax/oss/driver/core/TCPFlowControlIT.java
```diff
@@ -0,0 +1,251 @@
...
+      int hostIndex = 2;
+      BoundNode pausedNode = SIMULACRON_RULE.cluster().node(hostIndex);
+      SocketAddress pausedHostAddress = pausedNode.getAddress();
+      List<Node> nonPausedNodes =
+          session.getMetadata().getNodes().values().stream()
+              .filter(n -> !pausedHostAddress.equals(n.getBroadcastRpcAddress().get()))
+              .collect(Collectors.toList());
+      assertThat(nonPausedNodes.size()).isEqualTo(2);
+
+      SIMULACRON_RULE.cluster().node(2).pauseRead();
```
olim7t: It's the right value but I guess it would be better to reuse the existing variable?
```suggestion
      SIMULACRON_RULE.cluster().node(hostIndex).pauseRead();
```

integration-tests/src/test/java/com/datastax/oss/driver/core/TCPFlowControlIT.java
```diff
@@ -0,0 +1,251 @@
...
+    try (CqlSession session = SessionUtils.newSession(SIMULACRON_RULE)) {
+      SIMULACRON_RULE.cluster().pauseRead();
+
+      // The TCP send and receive buffer size depends on the OS
+      // We don't know how much data is needed to be flushed in order for OS to signal as full (and
+      // Channel#isWritable return false)
+      // Send 20Mb+ to each node
+      List<CompletionStage<AsyncResultSet>> pendingRequests = new ArrayList<>();
+      for (int i = 0; i < NUMBER_OF_SUBMITTED_REQUESTS; i++) {
+        pendingRequests.add(session.executeAsync(SimpleStatement.newInstance(queryString)));
```
olim7t:
```suggestion
        pendingRequests.add(session.executeAsync(queryString));
```

integration-tests/src/test/java/com/datastax/oss/driver/core/TCPFlowControlIT.java
```diff
@@ -0,0 +1,251 @@
...
+
+      // Assert that there are still requests that haven't been written
+      waitForWriteQueueToStabilize(session);
+      assertThat(getWriteQueueSize(session)).isGreaterThan(1);
+
+      // do not resumeRead
+      for (CompletionStage<AsyncResultSet> result : pendingRequests) {
+        assertThatThrownBy(() -> result.toCompletableFuture().get())
+            .isInstanceOf(ExecutionException.class)
+            .hasCauseInstanceOf(DriverTimeoutException.class);
```
olim7t, line 176: Using the same driver class that I mentioned earlier:
```java
        assertThatStage(result)
            .isFailed(e -> assertThat(e).isInstanceOf(DriverTimeoutException.class));
```
Or you could do it in pure AssertJ:
```java
        assertThat(result).hasFailedWithThrowableThat().isInstanceOf(DriverTimeoutException.class);
```

integration-tests/src/test/java/com/datastax/oss/driver/core/TCPFlowControlIT.java
```diff
@@ -0,0 +1,251 @@
...
+      }
+
+      // write queue size is still non empty
+      int writeQueueSize = getWriteQueueSize(session);
+      assertThat(writeQueueSize).isGreaterThan(1);
+      SIMULACRON_RULE.cluster().stop();
```
olim7t: I think this is a leftover, the rule will take care of that.

olim7t (changes requested):
> Let's have the test then, and if we ever notice any flakiness we can revisit this later.

Yeah I agree that this is borderline out of scope, but the work is done so let's keep it.

See a few comments inline.

Also, the `should_timeout...` test occasionally hangs on my machine, the session close future never completes. I'm looking into it.

olim7t:
> the `should_timeout...` test occasionally hangs on my machine

My bad, I was running with some local changes. My earlier suggestion:
```java
        assertThatStage(result)
            .isFailed(e -> assertThat(e).isInstanceOf(DriverTimeoutException.class));
```
Only passes if the future fails within a short timeout. But here we need to wait the whole 10 seconds, so it throws (I don't understand why that makes the test hang though...)

We can either add a 10-second sleep before the loop, or keep your original version.

integration-tests/src/test/java/com/datastax/oss/driver/core/TCPFlowControlIT.java
adutra: None of the above versions worked for me, the test hangs. As I don't have much time to investigate I will just leave the code as is now.

adutra: Actually this test hangs very often. An assertion fails, then there is a problem with the session closing sequence that never ends.

adutra (commented):
I don't think it's a good idea to merge this, one of the tests hangs more often than not and this will cause trouble in our CI infrastructure.

adutra:
> We can either add a 10-second sleep before the loop, or keep your original version.

Sorry, I just realize that you already investigated the issue. I think there is something fishy with the session closing sequence.

PR #1467, opened 2020-07-14
allow injection of CodecRegistry on session builder

wimtie:


olim7t:
Yeah that makes sense, right now we allow injection of individual codecs, but for the whole registry there's no easy way apart from extending the driver context.

I wonder if we need a factory though. Another option would be to add `SessionBuilder.withCodecRegistry()`, and then inject it via `ProgrammaticArguments` the way it's already done for nodeStateListener, etc.

The rationale for config-based initialization is that some components need a reference to the context to fetch internal dependencies at construction time. For the codec registry I don't think this will be a common need, in general you know ahead of time which codecs you'll need.

Can you share more about your use case? Does your custom implementation need the context, or could it be created before the session?

wimtie:
> Yeah that makes sense, right now we allow injection of individual codecs, but for the whole registry there's no easy way apart from extending the driver context.
> 
> I wonder if we need a factory though. Another option would be to add `SessionBuilder.withCodecRegistry()`, and then inject it via `ProgrammaticArguments` the way it's already done for nodeStateListener, etc.
> 
> The rationale for config-based initialization is that some components need a reference to the context to fetch internal dependencies at construction time. For the codec registry I don't think this will be a common need, in general you know ahead of time which codecs you'll need.
> 
> Can you share more about your use case? Does your custom implementation need the context, or could it be created before the session?

Hi, thanks for reviewing!

Initially I implemented this as SessionBuilder.withCodecRegistryBuilder(), the factory pattern is there so the creation method can still receive the session name and codecs like in `buildCodecRegistry(this.sessionName, codecs)`. Do you think it's better to keep this on SessionBuilder rather than config?

Our use case doesn't specifically need the context although I could imagine there are cases where that would be useful. We specifically are interested in overriding `CodecRegistry.createCodec()` to create specific MappingCodecs in certain cases.

Cheerio,
Tim

olim7t:
Let's just do the builder method for now. And I don't think we need a builder, just taking a registry instance should suffice:
```java
withCodecRegistry(CodecRegistry)
```
We can always add a config option later if the need arises.

wimtie:
I've updated the patch to create a `withCodecRegistry(CodecRegistry)` method on SessionBuilder rather than the previous implementation which used driver options.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -488,6 +489,12 @@ public SelfT withCloudSecureConnectBundle(@NonNull Path cloudConfigPath) {
     return self;
   }
 
+  /** Registers a CodecRegistry to use for the session. */
```
olim7t: Can you explain the interaction with `addTypeCodecs`, when both are called?

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
```diff
@@ -563,9 +562,12 @@ protected RequestProcessorRegistry buildRequestProcessorRegistry() {
     return new RequestProcessorRegistry(logPrefix, processors.toArray(new RequestProcessor[0]));
   }
 
-  protected CodecRegistry buildCodecRegistry(String logPrefix, List<TypeCodec<?>> codecs) {
-    MutableCodecRegistry registry = new DefaultCodecRegistry(logPrefix);
-    registry.register(codecs);
+  protected CodecRegistry getCodecRegistry(ProgrammaticArguments arguments) {
```
olim7t: Why change the method name? This is not a huge deal, but the rest of the class follows the convention `buildXxx` for the method that creates the object (either at construction time or when the lazy reference gets accessed), and `getXxx` for the public method that clients call to get the component. We already have a no-arg `getCodecRegistry` for the latter.

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
```diff
@@ -563,9 +562,12 @@ protected RequestProcessorRegistry buildRequestProcessorRegistry() {
...
+    MutableCodecRegistry registry = (MutableCodecRegistry) arguments.getCodecRegistry();
```
olim7t: So, that `MutableCodecRegistry` type is a bit awkward, it only exists for backward compatibility issue (maybe you've seen that in the javadocs already). While all registry implementations *should* extend it, it's possible for someone to pass an implementation that doesn't, we can't risk an exception here.

One option is to do an `instanceof` check, and skip the registration of the codecs if it fails. We should also log a warning, because if someone provided codecs they expect them to work.

Another is to require the mutable type directly at build time: `SessionBuilder.withCodecRegistry(MutableCodecRegistry)`.

I prefer the latter, WDYT?

olim7t (changes requested):
This looks good. A few more comments but they are minor.

Can you also add an entry in `changelog/README.md`? This qualifies as "[improvement]". We don't have a jira ticket, so just put the description directly.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
wimtie: I've expanded the javadoc to cover this.

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
wimtie: Fully agreed, undid the rename.

core/src/main/java/com/datastax/oss/driver/internal/core/context/DefaultDriverContext.java
wimtie: The former option feels more correct, while the second adds less clutter to the code. I find it hard to decide, so I'll defer to you and go with MutableCodecRegistry all the way.

olim7t: Sounds good. Another way to put it is that MutableCodecRegistry is the type that should have been exposed publicly from the start.

wimtie: All the changes should be there, thanks again for reviewing!

olim7t (approved):

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -488,6 +489,18 @@ public SelfT withCloudSecureConnectBundle(@NonNull Path cloudConfigPath) {
...
   }
 
+  /**
+   * Registers a CodecRegistry to use for the session.
+   *
+   * <p>When both this and {@link #addTypeCodecs(TypeCodec[])} are called, the added type codecs
+   * will be registered on the provided CodecRegistry.
+   */
+  @NonNull
+  public SelfT withCodecRegistry(@Nullable MutableCodecRegistry codecRegistry) {
```
olim7t, line 499: I changed the parameter from NonNull to Nullable. This would allow someone to clear the registry on a builder where it was configured previously. Not very likely but the implementation allows it, so why not.

(there are other attributes on the builder that don't follow this pattern, I think it's an oversight on our side, I'll take a look)

olim7t:
Merged, thanks for your contribution!

PR #1469, opened 2020-07-16
JAVA-2846: Load system settings when application.conf is also provided

tomekl007:
The config loader bug manifests when system setting is provided AND `application.conf`

It seems that the problematic part is in `DefaultDriverConfigLoader` (the `defaultApplication()`):
```
  public static final Supplier<Config> DEFAULT_CONFIG_SUPPLIER =
      () -> {
        // The thread's context class loader will be used for application classpath resources,
        // while the driver class loader will be used for reference classpath resources.
        return ConfigFactory.defaultApplication()
            .withFallback(ConfigFactory.defaultReference(CqlSession.class.getClassLoader()))
            .getConfig(DEFAULT_ROOT_PATH);
      };
```
?
When changing this to:
```
  public static final Supplier<Config> DEFAULT_CONFIG_SUPPLIER =
      () -> {
        // The thread's context class loader will be used for application classpath resources,
        // while the driver class loader will be used for reference classpath resources.
        return ConfigFactory.defaultOverrides()
            .withFallback(ConfigFactory.defaultReference(CqlSession.class.getClassLoader()))
            .getConfig(DEFAULT_ROOT_PATH);
      };
```
(please note `ConfigFactory.defaultOverrides()`) then the `-D` settings are picked up.

So if both are provided, the setting from `application.conf` has a priority over `-D` setting; according to the doc of `defaultApplication` :
`@return the default application.conf or system-property-configured configuration`
it should be another way around - so the system-property should have the higher property and override everything else, isn' i?

@adutra do you remember the context of those changes?

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
@@ -58,8 +58,10 @@ public class DefaultDriverConfigLoader implements DriverConfigLoader {
         ConfigFactory.invalidateCaches();
         // The thread's context class loader will be used for application classpath resources,
         // while the driver class loader will be used for reference classpath resources.
-        return ConfigFactory.defaultApplication()
+        return ConfigFactory.defaultOverrides()
+            .withFallback(ConfigFactory.defaultApplication())
             .withFallback(ConfigFactory.defaultReference(CqlSession.class.getClassLoader()))
+            .resolve()
             .getConfig(DEFAULT_ROOT_PATH);
       };
 
@@ -95,8 +97,10 @@ public class DefaultDriverConfigLoader implements DriverConfigLoader {
     this(
         () -> {
           ConfigFactory.invalidateCaches();
-          return ConfigFactory.defaultApplication(appClassLoader)
+          return ConfigFactory.defaultOverrides()
+              .withFallback(ConfigFactory.defaultApplication(appClassLoader))
               .withFallback(ConfigFactory.defaultReference(CqlSession.class.getClassLoader()))
+              .resolve()
               .getConfig(DEFAULT_ROOT_PATH);
         });
   }
--- a/core/src/test/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverConfigLoaderTest.java
+++ b/core/src/test/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverConfigLoaderTest.java
@@ -15,6 +15,7 @@
  */
 package com.datastax.oss.driver.internal.core.config.map;
 
+import static com.typesafe.config.ConfigFactory.defaultReference;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.fail;
 
@@ -30,6 +31,7 @@ import com.datastax.oss.driver.internal.core.config.MockOptions;
 import com.datastax.oss.driver.internal.core.config.MockTypedOptions;
 import com.datastax.oss.driver.internal.core.config.typesafe.DefaultDriverConfigLoader;
 import com.typesafe.config.ConfigException;
+import com.typesafe.config.ConfigFactory;
 import java.util.Optional;
 import org.junit.Test;
 
@@ -58,7 +60,14 @@ public class MapBasedDriverConfigLoaderTest {
     DriverExecutionProfile mapBasedConfig =
         DriverConfigLoader.fromMap(optionsMap).getInitialConfig().getDefaultProfile();
     DriverExecutionProfile fileBasedConfig =
-        new DefaultDriverConfigLoader().getInitialConfig().getDefaultProfile();
+        new DefaultDriverConfigLoader(
+                () -> {
+                  // Only load reference.conf since we are focusing on driver defaults
+                  ConfigFactory.invalidateCaches();
+                  return defaultReference().getConfig(DefaultDriverConfigLoader.DEFAULT_ROOT_PATH);
+                })
+            .getInitialConfig()
+            .getDefaultProfile();
 
     // Make sure we're not missing any options. -1 is for CONFIG_RELOAD_INTERVAL, which is not
     // defined by OptionsMap because it is irrelevant for the map-based config.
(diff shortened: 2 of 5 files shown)


adutra:
Yep, I guess I forgot ` ConfigFactory.defaultOverrides()` :-( 

But the correct call chain would be:

```
  public static final Supplier<Config> DEFAULT_CONFIG_SUPPLIER =
      () -> {
        // The thread's context class loader will be used for application classpath resources,
        // while the driver class loader will be used for reference classpath resources.
        return ConfigFactory.defaultOverrides()
            .withFallback(
                ConfigFactory.defaultApplication()
                    .withFallback(ConfigFactory.defaultReference(CqlSession.class.getClassLoader()))
                    .getConfig(DEFAULT_ROOT_PATH));
      };
```

We need to fix that for 4.8.0.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
```diff
@@ -58,8 +60,10 @@
         ConfigFactory.invalidateCaches();
         // The thread's context class loader will be used for application classpath resources,
         // while the driver class loader will be used for reference classpath resources.
-        return ConfigFactory.defaultApplication()
```
tomekl007, line 61: We have also a similar pattern in the `DefaultProgrammaticDriverConfigLoaderBuilder`:
```
  public static final Supplier<Config> DEFAULT_FALLBACK_SUPPLIER =
      () ->
          ConfigFactory.defaultApplication()
              // Do not remove root path here, it must be done after merging configs
              .withFallback(ConfigFactory.defaultReference(CqlSession.class.getClassLoader()));
```
and 
```
  public DefaultProgrammaticDriverConfigLoaderBuilder(@NonNull ClassLoader appClassLoader) {
    this(
        () ->
            ConfigFactory.defaultApplication(appClassLoader)
                .withFallback(ConfigFactory.defaultReference(CqlSession.class.getClassLoader())),
        DefaultDriverConfigLoader.DEFAULT_ROOT_PATH);
  }
```
I am not sure if progrmatticConfigLoader should also accept `ConfigFactory.defaultOverrides()`, wdyt?

adutra: No it should not change, the default overrides are applied later on, see [here](https://github.com/datastax/java-driver/blob/1427be868bcbb115258e87cccb1cffcf54c82e04/core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticDriverConfigLoaderBuilder.java#L250).

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
```diff
@@ -58,8 +60,10 @@
...
-            .withFallback(ConfigFactory.defaultReference(CqlSession.class.getClassLoader()))
+        return ConfigFactory.defaultOverrides()
+            .withFallback(
+                ConfigFactory.defaultApplication()
+                    .withFallback(defaultReference(CqlSession.class.getClassLoader())))
```
adutra: I just noticed: we are also missing a `.resolve()` at the end.

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
```diff
@@ -95,8 +99,11 @@ public DefaultDriverConfigLoader(@NonNull ClassLoader appClassLoader) {
     this(
         () -> {
           ConfigFactory.invalidateCaches();
-          return ConfigFactory.defaultApplication(appClassLoader)
-              .withFallback(ConfigFactory.defaultReference(CqlSession.class.getClassLoader()))
+          return ConfigFactory.defaultOverrides()
+              .withFallback(
+                  ConfigFactory.defaultApplication(appClassLoader)
+                      .withFallback(
+                          ConfigFactory.defaultReference(CqlSession.class.getClassLoader())))
```
adutra: Same here we need to call `.resolve()`

adutra (changes requested):

core/src/test/java/com/datastax/oss/driver/internal/core/config/map/MapBasedDriverConfigLoaderTest.java
```diff
@@ -58,7 +60,14 @@ public void should_fill_default_profile_like_reference_file() {
     DriverExecutionProfile mapBasedConfig =
         DriverConfigLoader.fromMap(optionsMap).getInitialConfig().getDefaultProfile();
     DriverExecutionProfile fileBasedConfig =
-        new DefaultDriverConfigLoader().getInitialConfig().getDefaultProfile();
+        new DefaultDriverConfigLoader(
+                () -> {
+                  // Only load reference.conf since we are focusing on driver defaults
+                  ConfigFactory.invalidateCaches();
```
adutra, line 66: required because the addition of an application.conf to the classpath breaks this test.

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoader.java
```diff
@@ -58,8 +60,11 @@
         ConfigFactory.invalidateCaches();
         // The thread's context class loader will be used for application classpath resources,
         // while the driver class loader will be used for reference classpath resources.
-        return ConfigFactory.defaultApplication()
-            .withFallback(ConfigFactory.defaultReference(CqlSession.class.getClassLoader()))
+        return ConfigFactory.defaultOverrides()
+            .withFallback(
+                ConfigFactory.defaultApplication()
+                    .withFallback(defaultReference(CqlSession.class.getClassLoader())))
```
olim7t: Nit: withFallback calls can be chained, this is more readable. I'll fix it directly as part of the merge.

olim7t (approved):

olim7t:
Merged, thanks for getting to the bottom of this.

PR #1470, opened 2020-07-16
JAVA-2811: Provide method aliases for smoother transition from driver 3

olim7t:


core/src/main/java/com/datastax/oss/driver/api/core/data/GettableById.java
```diff
@@ -196,6 +196,11 @@ default boolean getBoolean(@NonNull CqlIdentifier id) {
     return getBoolean(firstIndexOf(id));
   }
 
+  /** An alias for {@link #getBoolean(CqlIdentifier)} (for smoother transition from driver 3). */
+  default boolean getBool(@NonNull CqlIdentifier id) {
+    return getBoolean(id);
+  }
```
olim7t, line 206: This one in particular I'm not sure about, I'm really not a fan.

olim7t (commented):
We don't have to do everything, they're each in one individual commit.

adutra:
LGTM. 2 general questions:

1. I'm not sure about 769583e and the execution overloads, it's so simple to call `.build()` when you're done building your statement that I don't see this being very helpful?
2. Shouldn't we deprecate the 3.x method names? We want to smooth the upgrade to 4.x but also incentivize users to adopt the 4.x official API.

olim7t:
> I'm not sure about 769583e and the execution overloads, it's so simple to call `.build()` when you're done building your statement that I don't see this being very helpful?

One case where it would help is if you want to convert driver 3 code while keeping a mutable style.
```java
// driver 3
SimpleStatement s = new SimpleStatement("SELECT ...");
s.setString(0, "a");
session.execute(s);

// driver 4
SimpleStatementBuilder s = SimpleStatement.builder("SELECT ...");
s.setString(0, "a");
session.execute(s);
```

> Shouldn't we deprecate the 3.x method names? We want to smooth the upgrade to 4.x but also incentivize users to adopt the 4.x official API.

OK

adutra:
> One case where it would help is if you want to convert driver 3 code while keeping a mutable style.

I'm still not sure, users have to change their code from `SimpleStatement` to `SimpleStatementBuilder`, so the legacy code already requires some adaptation even if we introduce the new `execute` method.

But I'll let you decide and merge this at your convenience.

adutra (approved):

olim7t:
OK, I'm not sure anymore either, I'll revert that part.

PR #1472, opened 2020-07-17
JAVA-2819: Add DriverConfigLoader.fromString

olim7t:


core/src/test/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoaderTest.java
```diff
@@ -294,4 +294,17 @@ public void should_load_setting_from_system_property_when_application_conf_is_al
       System.clearProperty("datastax-java-driver.basic.request.timeout");
     }
   }
+
+  @Test
+  public void should_create_from_string() {
+    DriverExecutionProfile config =
+        DriverConfigLoader.fromString("datastax-java-driver.basic.request.timeout = 1 millisecond")
```
tomekl007: The string needs to be in the HOCON format, right? 
Are entries should be after newline (`/n`)?

tomekl007 (approved):

core/src/test/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultDriverConfigLoaderTest.java
olim7t: Yes. I should probably explain that in the javadocs.

core/src/main/java/com/datastax/oss/driver/api/core/config/DriverConfigLoader.java
```diff
@@ -200,6 +200,46 @@ static DriverConfigLoader fromUrl(@NonNull URL url) {
...
+   * (first-listed are higher priority):
+   *
+   * <ul>
+   *   <li>the properties in {@code contents}
+   *   <li>{@code reference.conf} (all resources on classpath with this name). In particular, this
+   *       will load the {@code reference.conf} included in the core driver JAR, that defines
+   *       default options for all mandatory options.
+   * </ul>
+   *
+   * Note that this variant does take into account system properties.
```
adutra: Why?

olim7t: The original use case was tools like Zeppelin that take the configuration from a GUI field, so I thought this would not be applicable.
But you're right, there's no reason to have that arbitrary limitation. I'll add it back.

adutra (approved):

PR #1474, opened 2020-07-17
JAVA-2800: Avoid SLF4J warnings when using the mapper processor

olim7t:
This is the easy option with slf4j-nop.

I've pushed another branch [java2800_exclude](https://github.com/datastax/java-driver/tree/java2800_exclude) with my attempt to exclude SLF4J from the processor path, but as I said in JIRA I couldn't make it work.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -4,6 +4,7 @@
 
 ### 4.8.0 (in progress)
 
+- [bug] JAVA-2800: Avoid SLF4J warnings when using the mapper processor
 - [bug] JAVA-2846: Give system properties the highest precedence in DefaultDriverConfigLoader
 - [new feature] JAVA-2691: Provide driver 4 support for extra codecs
 - [improvement] Allow injection of CodecRegistry on session builder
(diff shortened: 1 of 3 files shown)


olim7t:
Moving to draft, this still needs some work: it gets rid of the warning when the processor is configured with `<annotationProcessorPaths>`, but when it's configured the other way (regular dependency with scope `provided`), a new warning appears:
```java
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/Users/oliviermichallat/.m2/repository/org/slf4j/slf4j-nop/1.7.26/slf4j-nop-1.7.26.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/Users/oliviermichallat/.m2/repository/ch/qos/logback/logback-classic/1.2.3/logback-classic-1.2.3.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.slf4j.helpers.NOPLoggerFactory]
```

olim7t:
Closed in favor of #1475.

PR #1475, opened 2020-07-17
JAVA-2800: Exclude SLF4J from mapper-processor dependencies

olim7t:
This time I think it's the right one.


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/DefaultProcessorContext.java
```diff
@@ -115,4 +111,27 @@ public LoggingGenerator getLoggingGenerator() {
   public boolean areCustomResultsEnabled() {
     return customResultsEnabled;
   }
+
+  // Simple lazy reference implementation. We don't want to use the driver's because it depends on
+  // SLF4J.
+  static class LazyReference<ValueT> {
```
adutra: From what I see it's only `CycleDetector` that has a reference to SLF4J.

1. Do we really need those debug messages in `CycleDetector`? Maybe we could get rid of them?
2. Alternatively, could we envisage a log-agnostic implementation? Something like below:

```
public class CycleDetector {
  private static final boolean ENABLED =
      Boolean.getBoolean("com.datastax.oss.driver.DETECT_CYCLES");

  private final String errorMessage;
  private final boolean enabled;
  private final MutableValueGraph<String, String> graph;
  private final Consumer<String> log;

  public CycleDetector(String errorMessage, Consumer<String> log) {
    this(errorMessage, ENABLED, log);
  }

  @VisibleForTesting
  CycleDetector(String errorMessage, boolean enabled, Consumer<String> log) {
    this.errorMessage = errorMessage;
    this.enabled = enabled;
    this.graph = enabled ? ValueGraphBuilder.directed().build() : null;
    this.log = log;
  }

  void onTryLock(LazyReference<?> reference) {
    if (enabled) {
      synchronized (this) {
        Thread me = Thread.currentThread();
        log.accept(String.format("%s wants to initialize %s", me, reference.getName()));
        graph.putEdgeValue(me.getName(), reference.getName(), "wants to initialize");
        log.accept(graph.toString());
        if (Graphs.hasCycle(graph.asGraph())) {
          throw new IllegalStateException(errorMessage + " " + graph);
        }
      }
    }
  }

  void onLockAcquired(LazyReference<?> reference) {
    if (enabled) {
      synchronized (this) {
        Thread me = Thread.currentThread();
        log.accept(String.format("%s is initializing %s", me, reference.getName()));
        String old = graph.removeEdge(me.getName(), reference.getName());
        assert "wants to initialize".equals(old);
        graph.putEdgeValue(reference.getName(), me.getName(), "is getting initialized by");
      }
    }
  }

  void onReleaseLock(LazyReference<?> reference) {
    if (enabled) {
      synchronized (this) {
        Thread me = Thread.currentThread();
        log.accept(String.format("%s is done initializing %s", me, reference.getName()));
        graph.removeEdge(reference.getName(), me.getName());
      }
    }
  }
}
```

Then you can instantiate with or without SLF4J:

```
  protected final CycleDetector cycleDetector =
      new CycleDetector("Detected cycle in context initialization", LOG::debug);
  protected final CycleDetector cycleDetector =
      new CycleDetector("Detected cycle in context initialization", s -> {});
```

adutra: (An abstract class with abstract methods to log debug messages would work too)

olim7t: Hrmm right, the driver's original `LazyReference` is even designed to function without a cycle detector (I should know, I wrote it that way 🤦 ).
The processor context only has two references and they don't depend on each other, no need to get fancy.

PR #1476, opened 2020-07-20
JAVA-2813: Don't fail when secure bundle is specified together with other options

tomekl007:
todo:
- [x] don't fail for local-dc


core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -671,8 +675,11 @@ public SessionT build() {
...
         if (!programmaticContactPoints.isEmpty() || !configContactPoints.isEmpty()) {
-          throw new IllegalStateException(
-              "Can't use withCloudSecureConnectBundle and addContactPoint(s). They are mutually exclusive.");
+          LOG.info(
+              "The withCloudSecureConnectBundle and addContactPoint(s) were provided. They are mutually exclusive. The addContactPoint(s) setting will be ignored.");
+          // clear the contact points provided in the setting file and via addContactPoints
+          configContactPoints = Collections.emptyList();
+          programmaticContactPoints = new HashSet<>();
         }
         String configuredSSLFactory =
```
tomekl007: I am not sure if we should ignore (not fail) for SSL as well.
The `defaultConfig.getString(DefaultDriverOption.SSL_ENGINE_FACTORY_CLASS, null);` 
setting is used in multiple places, for example:
`DefaultDriverContext`:
```
protected Optional<SslEngineFactory> buildSslEngineFactory(SslEngineFactory factoryFromBuilder) {
    return (factoryFromBuilder != null)
        ? Optional.of(factoryFromBuilder)
        : Reflection.buildFromConfig(
            this,
            DefaultDriverOption.SSL_ENGINE_FACTORY_CLASS,
            SslEngineFactory.class,
            "com.datastax.oss.driver.internal.core.ssl");
  }
```
Ignoring this property will work now, but it seems fragile and may lead to some problems in the future (If we would change one of the usages of `DefaultDriverOption.SSL_ENGINE_FACTORY_CLASS`)
For `contactPoints` ignoring is not problematic because usage of `configContactPoints` and `programmaticContactPoints` is encapsulated in the `SessionBuilder`, wdty?

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -671,8 +675,11 @@ public SessionT build() {
           defaultConfig.getStringList(DefaultDriverOption.CONTACT_POINTS, Collections.emptyList());
       if (cloudConfigInputStream != null) {
         if (!programmaticContactPoints.isEmpty() || !configContactPoints.isEmpty()) {
-          throw new IllegalStateException(
-              "Can't use withCloudSecureConnectBundle and addContactPoint(s). They are mutually exclusive.");
+          LOG.info(
+              "The withCloudSecureConnectBundle and addContactPoint(s) were provided. They are mutually exclusive. The addContactPoint(s) setting will be ignored.");
```
adutra:
```suggestion
              "Both a secure connect bundle and contact points were provided. These are mutually exclusive. All contact points will be ignored.");
```

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
adutra: Any SSL configuration will be overridden by line 694 if a secure connect bundle is present:

```
withSslEngineFactory(cloudConfig.getSslEngineFactory());
```

So I think we can just replace lines 687-688 by a warning and be done with it :-)

adutra: However we should add a warning as well for local datacenter. Same pattern as for contact points.

examples/src/main/java/com/datastax/oss/driver/examples/astra/AstraReadCassandraVersion.java
```diff
@@ -57,7 +56,6 @@ public static void main(String[] args) {
             .withCloudSecureConnectBundle(Paths.get("/path/to/secure-connect-database_name.zip"))
             // Change the user_name and password here for the Astra instance
             .withAuthCredentials("user_name", "fakePasswordForTests")
-            .withConfigLoader(DriverConfigLoader.fromClasspath("application-astra"))
```
adutra, line 60: Nice that you remembered to change this 👍

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cloud/CloudIT.java
```diff
@@ -210,18 +216,32 @@ public void should_connect_to_proxy_using_url_with_http_protocol_provided_in_the
   }
 
   @Test
-  public void should_error_when_contact_points_and_secure_bundle_used() {
+  public void should_connect_and_log_info_when_contact_points_and_secure_bundle_used() {
```
adutra: Add a similar test for local datacenter and SSL.

integration-tests/src/test/java/com/datastax/oss/driver/api/core/cloud/CloudIT.java
```diff
@@ -210,18 +216,32 @@ public void should_connect_to_proxy_using_url_with_http_protocol_provided_in_the
...
     // given
+    LoggerTest.LoggerSetup logger = setupTestLogger(SessionBuilder.class, Level.INFO);
+
     Path bundle = proxyRule.getProxy().getBundleWithoutCredentialsPath();
-    CqlSessionBuilder builder =
+
+    ResultSet set;
```
adutra: Why declare this variable outside of the try-catch block?

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
tomekl007: It is a bit problematic for contact points because it fails with:
```
 Multiple entries with same key: default=dc1 and default=dc-ignored
```
It is using the 
`
private ImmutableMap.Builder<String, String> localDatacentersBuilder = ImmutableMap.builder();
`
underneath and it does not support removing. Are we ok with changing this?

olim7t: You can add a new method `ProgrammaticArguments.Builder.clearDatacenters()`.

olim7t (approved):
Implementation looks good to me (pending the additional local DC check requested by Alexandre).

I'm not sure I agree with the change though: we decided to fail fast precisely because we didn't want to miss the fact that some options were going to be ignored. So it sounds consistent that tools would have to add logic to work around that.

adutra:
> I'm not sure I agree with the change though: we decided to fail fast precisely because we didn't want to miss the fact that some options were going to be ignored. So it sounds consistent that tools would have to add logic to work around that.

The issue is with frameworks that provide some form of support for creating CqlSession instances in a more or less automated way. Take Spring Boot, for instance: the default `CassandraProperties` bean [exposes](https://github.com/spring-projects/spring-boot/blob/42817224e1ace9d3cca0090843358dc43399b5bb/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cassandra/CassandraProperties.java#L57-L68) contact points and local datacenter. These are then [automatically applied](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cassandra/CassandraAutoConfiguration.java#L130-L133) in `CassandraAutoConfiguration`. If you are using Astra, the defaults in Spring Boot won't let you connect, and your only option is to actually bypass `CassandraAutoConfiguration` completely – which, admittedly, defeats the purpose of using Spring Boot in the first place.

(And it's not realistic to offer to add support to Astra directly in Spring Boot: Astra being proprietary and commercial, it's likely that such a contribution would meet some resistance to say the least.)

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -671,18 +677,27 @@ public SessionT build() {
           defaultConfig.getStringList(DefaultDriverOption.CONTACT_POINTS, Collections.emptyList());
       if (cloudConfigInputStream != null) {
         if (!programmaticContactPoints.isEmpty() || !configContactPoints.isEmpty()) {
-          throw new IllegalStateException(
-              "Can't use withCloudSecureConnectBundle and addContactPoint(s). They are mutually exclusive.");
+          LOG.info(
+              "Both a secure connect bundle and contact points were provided. These are mutually exclusive. The contact points from a secure bundle will have priority.");
```
adutra:
```suggestion
              "Both a secure connect bundle and contact points were provided. These are mutually exclusive. The contact points from the secure bundle will have priority.");
```

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -671,18 +677,27 @@ public SessionT build() {
...
+          configContactPoints = Collections.emptyList();
+          programmaticContactPoints = new HashSet<>();
         }
         String configuredSSLFactory =
             defaultConfig.getString(DefaultDriverOption.SSL_ENGINE_FACTORY_CLASS, null);
         if (sslConfigured || configuredSSLFactory != null) {
-          throw new IllegalStateException(
-              "Can't use withCloudSecureConnectBundle and explicitly specify ssl configuration. They are mutually exclusive.");
+          LOG.info(
+              "Both withCloudSecureConnectBundle and explicitly specified ssl configuration were provided. They are mutually exclusive. The ssl settings from a secure bundle will have priority.");
```
adutra:
```suggestion
              "Both a secure connect bundle and SSL options were provided. They are mutually exclusive. The SSL options from the secure bundle will have priority.");
```

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -671,18 +677,27 @@ public SessionT build() {
...
         }
         CloudConfig cloudConfig =
             new CloudConfigFactory().createCloudConfig(cloudConfigInputStream.call());
         addContactEndPoints(cloudConfig.getEndPoints());
+
+        if (localDatacenterConfigured) {
```
adutra: I think you should also inspect the configuration (same pattern done above for contact points and SSL).

But there is one difference: since you are clearing all datacenters, you should inspect all profiles for a datacenter.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -671,18 +677,27 @@ public SessionT build() {
...
+          LOG.info(
+              "Both withCloudSecureConnectBundle and explicitly specified local datacenter configuration were provided. They are mutually exclusive. The local datacenter settings from a secure bundle will have priority.");
```
adutra:
```suggestion
              "Both a secure connect bundle and a local datacenter were provided. They are mutually exclusive. The local datacenter from the secure bundle will have priority.");
```

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -671,18 +677,27 @@ public SessionT build() {
...
-              "Can't use withCloudSecureConnectBundle and addContactPoint(s). They are mutually exclusive.");
+          LOG.info(
+              "Both a secure connect bundle and contact points were provided. These are mutually exclusive. The contact points from a secure bundle will have priority.");
+          // clear the contact points provided in the setting file and via addContactPoints
+          configContactPoints = Collections.emptyList();
+          programmaticContactPoints = new HashSet<>();
         }
         String configuredSSLFactory =
             defaultConfig.getString(DefaultDriverOption.SSL_ENGINE_FACTORY_CLASS, null);
         if (sslConfigured || configuredSSLFactory != null) {
```
adutra: Minor suggestion: rename these variables to match the variable names used for contact points:

```
sslConfigured -> programmaticSslFactory
configuredSSLFactory -> configSslFactory
```

(Also: `configuredSSLFactory` doesn't respect naming rules)

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
adutra: Following my suggestion above: rename to `programmaticLocalDatacenter`.

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
tomekl007: `programmaticSslFactory` for a boolean doesn't seem to be a good name. Maybe `isProgrammaticSslFactorySet` or similar?

adutra (approved):

PR #1478, opened 2020-07-20
JAVA-2331: Unregister old metrics when a node gets removed or changes RPC address

tomekl007:
todo:
- [x] parameterize cache eviction time 
- [x] docs


core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricUpdater.java
```diff
@@ -33,46 +40,80 @@
...
+  protected DropwizardMetricUpdater(
+      Set<MetricT> enabledMetrics, MetricRegistry registry, Ticker ticker) {
+    this(enabledMetrics, registry, ticker, Duration.ofHours(1));
+  }
+
+  protected DropwizardMetricUpdater(
+      Set<MetricT> enabledMetrics, MetricRegistry registry, Ticker ticker, Duration expiresAfter) {
     this.enabledMetrics = enabledMetrics;
     this.registry = registry;
+    metricsCache =
```
tomekl007: I followed @adutra suggestion from this jira comment: https://datastax-oss.atlassian.net/browse/JAVA-2331?focusedCommentId=51936 to implement eviction of metrics using cache but I have doubts about this approach:
- How to approach configuring the eviction time? 
- This solution is eventually consistent - our users will need to tolerate some non-negligible amount of time when there will be stale metrics - are we ok with it?
- Are all metrics frequently updated? If we will pick 1 hour as a default eviction policy, can we assure that all metrics have a shorter update period?
- What if the user will set the eviction time to very short amount? (i.e. 1 second) There will be a situation in which metrics will be removed when they shouldn't. We should probably restrict the lowest value that could be set to some high value (i.e. 1 hour)
- keeping all metrics for all nodes in a cluster in the guava cache will mean that the cache can grow to thousands of entries - are we ok with that additional memory footprint?
wdyt?

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricUpdater.java
```diff
@@ -33,46 +40,80 @@
...
   protected final Set<MetricT> enabledMetrics;
+
   protected final MetricRegistry registry;
 
-  protected DropwizardMetricUpdater(Set<MetricT> enabledMetrics, MetricRegistry registry) {
+  protected final Cache<String, Metric> metricsCache;
+
+  protected DropwizardMetricUpdater(
+      Set<MetricT> enabledMetrics, MetricRegistry registry, Ticker ticker) {
+    this(enabledMetrics, registry, ticker, Duration.ofHours(1));
```
tomekl007: `Duration.ofHours(1)` - this will be parameterized (to be able to set via `reference.conf`)

olim7t:
I like the idea of a time-based eviction mechanism.

> our users will need to tolerate some non-negligible amount of time when there will be stale metrics - are we ok with it?

A short period where the metrics of a node that is gone are still visible. Yes, I think that's OK.

> Are all metrics frequently updated? If we will pick 1 hour as a default eviction policy, can we assure that all metrics have a shorter update period?

Unfortunately no.
Node-level metrics include error counts, like `errors.request.write-timeouts`. In an ideal situation where errors never happen, you expect the metric to stay at 0, but with the current mechanism it will get deleted because it's not getting updated.

There's another problem with having the cache at the `MetricUpdater` level: the Guava cache handles eviction [synchronously, during other operations](https://github.com/google/guava/wiki/CachesExplained#when-does-cleanup-happen). If a node gets removed, its `MetricUpdater` will eventually stop receiving calls, which means the Guava cache itself will stop getting hits. So the metric evictions might never happen, because they have no cache operations to piggyback on.

I think the solution to both issues is to have a single cache at the `MetricsFactory` level.
```java
Cache<Node, DropwizardNodeMetricUpdater> cache;
```
Each single cache entry represents *all* the metrics of a given node. Whenever `newNodeUpdater` gets invoked, we create a new entry.
But note that after that, the factory is not invoked again: the updater is stored in the `Node` object, and accessed from there directly.
So if we want the cache access times to update, we need to trigger cache hits ourselves, for every call to `NodeMetricUpdater`'s `incrementCounter`/`updateHistogram`/`markMeter`/`updateTimer` methods. We pass down a component that those methods will invoke; I think a simple `Runnable signalMetricUpdated` (that does a `cache.get`) would work.

Then we just need to unregister all metrics when an entry gets evicted. This can be done by adding a method on `DropwizardNodeMetricUpdater`, or maybe one of its parent interfaces.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardNodeMetricUpdater.java
```diff
@@ -116,4 +154,40 @@ private void initializePoolGauge(
...
+    registry.remove(buildFullName(DefaultNodeMetric.IGNORES_ON_WRITE_TIMEOUT, null));
+    registry.remove(buildFullName(DefaultNodeMetric.IGNORES_ON_UNAVAILABLE, null));
+    registry.remove(buildFullName(DefaultNodeMetric.IGNORES_ON_OTHER_ERROR, null));
+    registry.remove(buildFullName(DefaultNodeMetric.SPECULATIVE_EXECUTIONS, null));
+    registry.remove(buildFullName(DefaultNodeMetric.CONNECTION_INIT_ERRORS, null));
+    registry.remove(buildFullName(DefaultNodeMetric.AUTHENTICATION_ERRORS, null));
+    registry.remove(buildFullName(DseNodeMetric.GRAPH_MESSAGES, profileName));
+
+    registry.remove(buildFullName(DefaultNodeMetric.BYTES_RECEIVED, null));
+    registry.remove(buildFullName(DefaultNodeMetric.BYTES_SENT, null));
```
olim7t: I think you can remove all the metrics in one line:
```java
registry.removeMatching((name, metric) -> name.startsWith(metricNamePrefix));
```

core/src/main/resources/reference.conf
```diff
@@ -1611,6 +1611,15 @@ datastax-java-driver {
...
       }
+
+      # The time after the node level metrics will be evicted.
+      # The eviction will happen only if none of the enabled node-level metrics is updated for a given node for a given time.
+      # When the eviction-time will elapse, all metrics for a given node are removed.
+      #
+      # Required: no (defaults to 1 hour)
+      # Modifiable at runtime: no
+      # Overridable in a profile: no
+      eviction-time = 1 hour
```
olim7t: I wonder if we need to expose this in the configuration. Maybe 1 hour is a reasonable one-size-fit-all value that we could hard-code instead.

We've had that philosophy of "when in doubt, make it configurable" in driver 4, but the config file is getting huge. I think we should be more opinionated sometimes.

olim7t (commented):
👍  it's what I had in mind

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardNodeMetricUpdater.java
tomekl007: nice, thx

core/src/main/resources/reference.conf
tomekl007: Regarding one hour, I am not sure if this is `one-size-fit-all`. 
- Maybe there will be some clients that will not tolerate stale metrics for one hour but also tolerate some risk of flickering and set it to N minutes?
- Or on the other hand, maybe someone that uses our driver in a context of batch jobs (i.e. spark) that are executed every N hours will need to increment that parameters to couple hours to retain metrics between those batch job runs?

core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOption.java
```diff
@@ -517,7 +517,6 @@
    * <p>Value-type: {@link java.time.Duration Duration}
    */
   METRICS_NODE_CQL_MESSAGES_INTERVAL("advanced.metrics.node.cql-messages.refresh-interval"),
-
```
adutra: This change looks spurious.

core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOption.java
```diff
@@ -808,6 +807,14 @@
    * <p>Value-type: int
    */
   SESSION_LEAK_THRESHOLD("advanced.session-leak.threshold"),
+  /**
+   * The time after the node level metrics will be evicted. The eviction will happen only if none of
```
adutra:
```suggestion
   * The period of inactivity after which the node level metrics will be evicted. The eviction will happen only if none of
```

core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOption.java
```diff
@@ -808,6 +807,14 @@
...
+   * the enabled node-level metrics is updated for a given node for a given time. When the
+   * eviction-time will elapse, all metrics for a given node are removed.
```
adutra:
```suggestion
   * the enabled node-level metrics is updated for a given node within this time window.
```

The last sentence seems redundant.

core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOption.java
```diff
@@ -808,6 +807,14 @@
...
+   *
+   * <p>Value-type: {@link java.time.Duration Duration}
+   */
+  METRICS_NODE_EVICTION_TIME("advanced.metrics.node.eviction-time"),
```
adutra:
```suggestion
  METRICS_NODE_EXPIRE_AFTER("advanced.metrics.node.expire-after"),
```

I don't like much the usage of the word "time" in the sense of "expiration interval". Also "eviction" is a cache vocabulary, but users don't need to know we are using a cache with eviction based on activity, this is an implementation detail.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricsFactory.java
```diff
@@ -73,6 +96,20 @@ public DropwizardMetricsFactory(InternalDriverContext context) {
     }
   }
 
+  private Duration getAndValidateEvictionTime(DriverExecutionProfile config) {
+    Duration evictionTime = config.getDuration(DefaultDriverOption.METRICS_NODE_EVICTION_TIME);
+
+    if (evictionTime.compareTo(LOWEST_ACCEPTABLE_EVICTION_TIME) < 0) {
+      LOG.warn(
+          "The {} setting was provided with too low value. Consider increasing it to at least {} hour. "
+              + "Having lower value may cause disappearing and reappearing of your node-level metrics.",
```
adutra:
```suggestion
              + "Having lower values may cause your node-level metrics to keep disappearing and reappearing.",
```

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricsFactory.java
```diff
@@ -73,6 +96,20 @@ public DropwizardMetricsFactory(InternalDriverContext context) {
     }
   }
 
+  private Duration getAndValidateEvictionTime(DriverExecutionProfile config) {
+    Duration evictionTime = config.getDuration(DefaultDriverOption.METRICS_NODE_EVICTION_TIME);
+
+    if (evictionTime.compareTo(LOWEST_ACCEPTABLE_EVICTION_TIME) < 0) {
+      LOG.warn(
+          "The {} setting was provided with too low value. Consider increasing it to at least {} hour. "
```
adutra: Shouldn't we display the offending value in the message?

integration-tests/src/test/java/com/datastax/oss/driver/core/metrics/MetricsSimulacronIT.java
```diff
@@ -0,0 +1,271 @@
...
+import org.awaitility.Awaitility;
+import org.junit.Before;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.runner.RunWith;
+
+@Category(ParallelizableTests.class)
+@RunWith(DataProviderRunner.class)
+public class MetricsSimulacronIT {
```
adutra, line 50: Nice tests 👍

adutra (changes requested):

core/src/main/resources/reference.conf
```diff
@@ -1611,6 +1611,15 @@ datastax-java-driver {
         significant-digits = 3
         refresh-interval = 5 minutes
       }
+
+      # The time after the node level metrics will be evicted.
+      # The eviction will happen only if none of the enabled node-level metrics is updated for a given node for a given time.
```
olim7t: Nit: could you reformat to avoid lines of more than 100 characters?

core/src/main/resources/reference.conf
```diff
@@ -1611,6 +1611,15 @@ datastax-java-driver {
         significant-digits = 3
         refresh-interval = 5 minutes
       }
+
+      # The time after the node level metrics will be evicted.
```
olim7t:
```suggestion
      # The time after which the node level metrics will be evicted.
```

core/src/main/resources/reference.conf
```diff
@@ -1611,6 +1611,15 @@ datastax-java-driver {
...
+      # The eviction will happen only if none of the enabled node-level metrics is updated for a given node for a given time.
+      # When the expire-after will elapse, all metrics for a given node are removed.
+      #
```
olim7t, line 1620: Also specify that it will be forced to 1 hour (currently, see my other comment) if you set a lower value.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricsFactory.java
```diff
@@ -41,24 +46,42 @@
 public class DropwizardMetricsFactory implements MetricsFactory {
 
   private static final Logger LOG = LoggerFactory.getLogger(DropwizardMetricsFactory.class);
+  private static final Duration LOWEST_ACCEPTABLE_EVICTION_TIME = Duration.ofHours(1);
```
olim7t: That seems a bit aggressive. They are probably many applications where hundreds of requests per second is the norm, I can see how someone might want to go lower in those cases.

We want to prevent absurd values, like less than a few seconds. But we could go as low as 5 minutes, WDYT?

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricsFactory.java
```diff
@@ -41,24 +46,42 @@
...
+    metricsCache =
+        CacheBuilder.newBuilder()
+            .ticker(ticker)
+            .expireAfterAccess(evictionTime)
+            .removalListener(
+                (RemovalNotification<Node, DropwizardNodeMetricUpdater> notification) -> {
+                  LOG.debug(
+                      "Removing metrics for node: {} from cache after {}",
+                      notification.getKey(),
+                      evictionTime);
```
olim7t:
```suggestion
                  LOG.debug(
                      "[{}] Removing metrics for node: {} from cache after {}",
                      logPrefix,
                      notification.getKey(),
                      evictionTime);
```

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricsFactory.java
```diff
@@ -73,6 +96,21 @@ public DropwizardMetricsFactory(InternalDriverContext context) {
...
+  private Duration getAndValidateEvictionTime(DriverExecutionProfile config) {
+    Duration evictionTime = config.getDuration(DefaultDriverOption.METRICS_NODE_EXPIRE_AFTER);
+
+    if (evictionTime.compareTo(LOWEST_ACCEPTABLE_EVICTION_TIME) < 0) {
+      LOG.warn(
+          "The {} setting was provided with too low value: {}. Consider increasing it to at least {} hour. "
+              + "Having lower values may cause your node-level metrics to keep disappearing and reappearing.",
+          DefaultDriverOption.METRICS_NODE_EXPIRE_AFTER,
+          evictionTime,
+          LOWEST_ACCEPTABLE_EVICTION_TIME.toHours());
```
olim7t:
```suggestion
      LOG.warn(
          "[{}] Value too low for {}: {}. Forcing to {} instead.",
          logPrefix,
          DefaultDriverOption.METRICS_NODE_EXPIRE_AFTER.getPath(),
          evictionTime,
          LOWEST_ACCEPTABLE_EVICTION_TIME);
```
Log prefix + using the option path instead of the programmatic name + shortening the error message (I don't think it's necessary to explain the consequences) + indicating that we force the value, it's not just a suggestion + keep the raw duration, per my other comment to maybe use less than one hour.

Note that we have a pending ticket to improve the formatting of durations: [JAVA-2845](https://datastax-oss.atlassian.net/browse/JAVA-2845).

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricsFactory.java
```diff
@@ -41,24 +46,42 @@
...
   private static final Logger LOG = LoggerFactory.getLogger(DropwizardMetricsFactory.class);
+  private static final Duration LOWEST_ACCEPTABLE_EVICTION_TIME = Duration.ofHours(1);
 
   private final String logPrefix;
   private final InternalDriverContext context;
   private final Set<NodeMetric> enabledNodeMetrics;
   private final MetricRegistry registry;
   @Nullable private final Metrics metrics;
   private final SessionMetricUpdater sessionUpdater;
+  private Cache<Node, DropwizardNodeMetricUpdater> metricsCache;
```
olim7t:
```suggestion
  private final Cache<Node, DropwizardNodeMetricUpdater> metricsCache;
```

integration-tests/src/test/java/com/datastax/oss/driver/core/metrics/MetricsSimulacronIT.java
```diff
@@ -0,0 +1,274 @@
...
+                assertThat(metrics.<Meter>getSessionMetric(DefaultSessionMetric.BYTES_SENT))
+                    .hasValueSatisfying(
+                        bytesSent -> assertThat(bytesSent.getCount()).isGreaterThan(0));
+                assertThat(metrics.<Meter>getSessionMetric(DefaultSessionMetric.BYTES_RECEIVED))
+                    .hasValueSatisfying(
+                        bytesReceived -> assertThat(bytesReceived.getCount()).isGreaterThan(0));
+              });
+
+      // and node metrics are evicted
+      Awaitility.await()
```
olim7t: You can add a static import for the method. I forgot to add it in the contribution guidelines, but we've done it in other tests.

integration-tests/src/test/java/com/datastax/oss/driver/core/metrics/MetricsSimulacronIT.java
```diff
@@ -0,0 +1,274 @@
...
+                return metrics.<Meter>getNodeMetric(node, DefaultNodeMetric.BYTES_SENT).isPresent()
+                    && metrics
+                        .<Meter>getNodeMetric(node, DefaultNodeMetric.ABORTED_REQUESTS)
+                        .isPresent();
+              });
+    }
+  }
+
+  @Test
+  public void should_log_warning_when_provided_eviction_time_setting_is_too_low() {
```
olim7t: If you make the method static in `DropwizardMetricsFactory`,  you can cover this (and the next method) in a unit test instead.

olim7t (changes requested):
Looking great, last round of remarks and then I think we're done.

Can you also add a changelog entry?

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricsFactory.java
tomekl007: good idea, it is too strict now. Changed

core/src/main/java/com/datastax/oss/driver/api/core/config/TypedDriverOption.java
```diff
@@ -701,6 +701,10 @@ public String toString() {
       new TypedDriverOption<>(
           DseDriverOption.METRICS_NODE_GRAPH_MESSAGES_INTERVAL, GenericType.DURATION);
 
+  /** The time after the node level metrics will be evicted. */
```
adutra:
```suggestion
  /** The time after which the node level metrics will be evicted. */
```

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricsFactory.java
```diff
@@ -73,6 +99,23 @@ public DropwizardMetricsFactory(InternalDriverContext context) {
...
+    Duration evictionTime = config.getDuration(DefaultDriverOption.METRICS_NODE_EXPIRE_AFTER);
+
+    if (evictionTime.compareTo(LOWEST_ACCEPTABLE_EXPIRE_AFTER) < 0) {
+      LOG.warn(
+          "[{}] Value too low for {}: {} (It should be higher than {}). Forcing to {} instead.",
+          logPrefix,
+          DefaultDriverOption.METRICS_NODE_EXPIRE_AFTER.getPath(),
+          evictionTime,
+          LOWEST_ACCEPTABLE_EXPIRE_AFTER,
+          DEFAULT_EXPIRE_AFTER);
```
adutra, line 113: If the user value is lesser than the lowest acceptable one, I think it would instead make sense to force it to be equal to the lowest acceptable value no? IOW, force to 5 minutes instead of 1 hour.

core/src/main/resources/reference.conf
```diff
@@ -1611,6 +1611,17 @@ datastax-java-driver {
         significant-digits = 3
         refresh-interval = 5 minutes
       }
+
+      # The time after which the node level metrics will be evicted.
+      # The eviction will happen only if none of the enabled node-level
+      # metrics is updated for a given node for a given time.
+      # When the expire-after will elapse, all metrics for a given node are removed.
```
adutra:
```suggestion
      # When this interval elapses, all metrics for the idle node are removed.
```

core/src/main/resources/reference.conf
```diff
@@ -1611,6 +1611,17 @@ datastax-java-driver {
...
+      # If you set it to a value lower than 5 minutes, it will be forced to a default 1 hour.
```
adutra, line 1619: See my other comment; I find this logic a bit twisted. This would make more sense imo: `If you set it to a value lower than 5 minutes, it will be forced to 5 minutes.`

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/DropwizardMetricsFactory.java
tomekl007: Yes, I was thinking about it but on the other hand, if someone provides the wrong value maybe it is better to just set it to default?

adutra: If someone overrides the default value to, say, 1 minute, it's probably because they really want an aggressive eviction policy. I think it would make sense to provide them with the most aggressive interval that we support, so 5 minutes, not 1 hour.

tomekl007: It does make sense. Done

adutra (approved):

PR #1479, opened 2020-07-20
3.x vs. 4.x benchmarks

absurdfarce:


fallout/benchmarks/benchmark.yaml
```diff
@@ -0,0 +1,81 @@
+ensemble:
+  server:
+    node.count: 3
+    provisioner:
+      name: ctool
+      properties:
```
absurdfarce: Per conversation with @adutra we likely want to explicitly locate these servers in the "performance" tenant

fallout/benchmarks/benchmark.yaml
```diff
@@ -0,0 +1,81 @@
...
+        cloud.provider: nebula
+        cloud.instance.type: m3.large
```
absurdfarce: Confirm with Fallout team whether this is the appropriate size for servers in this case

fallout/benchmarks/benchmark.yaml
```diff
@@ -0,0 +1,81 @@
...
+      - name: ctool
+        properties:
+          install.maven: true
+      - name: gatling
+        properties:
+          git.repository: git@github.com:riptano/dse-benchmarks.git
+          # master uses a Gatling DSE plugin based on 3.x
+          #git.branch: master
+          # This feature branch uses a Gatling DSE plugin based on the unified driver (>= 4.5.x)
+          git.branch: DSP-17972-unified-driver
```
absurdfarce: Branch used for a given test run should be converted to a Fallout parameter

absurdfarce:
This test should now be available at https://fallout.sjc.dsinternal.org/tests/ui/bret.mcguire@datastax.com/gatling_timeseries_latency

fallout/gatling/gatling_timeseries_latency.yaml
```diff
@@ -0,0 +1,84 @@
...
+        cloud.instance.type: m3.xlarge
+    configuration_manager:
+      - name: ctool
+        properties:
+          install.maven: true
+      - name: gatling
+        properties:
+          git.repository: git@github.com:riptano/dse-benchmarks.git
+          git.branch: {{branch}}
+
```
tomekl007, line 51: don't you want to add the `observer` node with grafana and send results from Gatling there?
Or are we ok with auto-generated HTML files that Gatling creates after every test run?

fallout/gatling/gatling_timeseries_latency.yaml
```diff
@@ -0,0 +1,84 @@
...
+workload:
+  phases:
+    # This simulation isn't necessary in a single-node case; Timeseries_Write will
+    # create the infrastructure that's needed.  We do need it in the multi-node
```
tomekl007: `create the infrastructure that's needed.` - what does it mean? Will it setup keyspace/tables?

fallout/gatling/gatling_timeseries_latency.yaml
```diff
@@ -0,0 +1,84 @@
...
+              scala:testCompile
+              gatling:execute
+              -Dgatling.disableCompiler
+              -Dgatling.simulationClass=core.timeseries.Timeseries_Schema
+    - timeseries-write:
+        module: gatling
+        properties:
+          maven_args: >
+              scala:testCompile
+              gatling:execute
```
tomekl007, line 71: where the results(HTML file with charts) from the gatling run will be stored?
Don't we want to persist it between fallout restarts? If yes, they should be copied into `${FALLOUT_ARTIFACT_DIR}`

tomekl007:
> This test should now be available at https://fallout.sjc.dsinternal.org/tests/ui/bret.mcguire@datastax.com/gatling_timeseries_latency

I don't see any results

fallout/gatling/gatling_timeseries_latency.yaml
```diff
@@ -0,0 +1,84 @@
...
+              -Dgatling.disableCompiler
+              -Dgatling.simulationClass=core.timeseries.Timeseries_Write
+              -DconnectionsPerHost=8
```
tomekl007: I think that the base test should rely on the default parameters of the driver under test. Why change it to non-default 8?

fallout/gatling/gatling_timeseries_latency.yaml
```diff
@@ -0,0 +1,84 @@
...
+              -DrowsPerPartition=1000
+              -DinjectionType=constant
+              -Dthroughput=10000
+              -DbenchmarkDurationInMinutes=10
```
tomekl007, line 77: 10 minutes as a time for performance tests seems too low. I think that we should increase it to have more durable/long-running tests.

fallout/gatling/gatling_timeseries_latency.yaml
```diff
@@ -0,0 +1,84 @@
...
+    - timeseries-read:
+        module: gatling
+        properties:
+          maven_args: >
+              scala:testCompile
+              gatling:execute
+              -Dgatling.disableCompiler
+              -Dgatling.simulationClass=core.timeseries.Timeseries_Read
+              -DconnectionsPerHost=8
```
tomekl007: same comments as for `Timeseries_Write`

fallout/gatling/gatling_timeseries_latency.yaml
```diff
@@ -0,0 +1,84 @@
...
+        properties:
+          maven_args: >
+              scala:testCompile
+              gatling:execute
+              -Dgatling.disableCompiler
+              -Dgatling.simulationClass=core.timeseries.Timeseries_Write
+              -DconnectionsPerHost=8
+              -DrowsPerPartition=1000
+              -DinjectionType=constant
+              -Dthroughput=10000
```
tomekl007, line 76: This parameter should be a lot higher. Kafka or dsbulk are able to generate traffic that is > 200_000 per second. We should validate at least this number of requests

fallout/gatling/gatling_timeseries_latency.yaml
```diff
@@ -0,0 +1,84 @@
...
+      - name: ctool
+        properties:
+          install.maven: true
+      - name: gatling
+        properties:
+          git.repository: git@github.com:riptano/dse-benchmarks.git
+          git.branch: {{branch}}
+
+workload:
+  phases:
```
tomekl007, line 53: you may want to increase the timeout of this fallout step

tomekl007 (changes requested):

fallout/gatling/gatling_timeseries_latency.yaml
absurdfarce: Gatling records HDRHistogram data files of it's run that can be used to compute a histogram of request latencies as well as throughput.  So my original plan was to just rely on those files.  I still think that will work for latency but it doesn't really get at the desire to step-wise increase throughput and record when things fall off the rails.

My _new_ thinking is very much along the lines of what you describe: use the approach I have right now for latency histograms but change the throughput to a step-wise increasing function, then observe throughput as reported to an external monitor, likely Grafana.

fallout/gatling/gatling_timeseries_latency.yaml
absurdfarce: Understood.  Most of the knobs here (timeouts, throughput, etc.) will need another pass or two of refinement once I get the details of the mechanism fully sorted out.

fallout/gatling/gatling_timeseries_latency.yaml
absurdfarce: Yes, that's correct.  Both Timeseries_Write and Timeseries_Read will create the keyspace and execute any required DDL statements.  Obviously the read test won't do much of interest unless data has previously been populated.

I'll re-work this line to make that clearer.

fallout/gatling/gatling_timeseries_latency.yaml
absurdfarce: Test artifacts are preserved even across Fallout restarts aren't they?  They were in my testing when using a local Fallout server so for now I think what's there is fine.  This may change if/when the Grafana server is introduced.

fallout/gatling/gatling_timeseries_latency.yaml
absurdfarce: This prop was set in a few other Fallout runs using Timeseries_(Read|Write) so I preserved it here for sake of continuity.  I don't think it's required by any means.

fallout/gatling/gatling_timeseries_latency.yaml
absurdfarce: This value (like the timeout mentioned above) will also need some tweaking once the mechanism is fully in place and working.

fallout/gatling/gatling_timeseries_latency.yaml
absurdfarce: Same as above: this value will likely change once I'm satisfied with the mechanism.

absurdfarce:
@tomekl007 The tests hadn't been run yet when I posted the original message.  I've since completed a run for both branches of dse-benchmarks in order to prove out the move to the public Fallout server so results are available now.

fallout/gatling/gatling_timeseries_latency.yaml
tomekl007: Yes, if you want to persists results this is exactly why `${FALLOUT_ARTIFACT_DIR}` was created. 
After each test run, the fallout is creating an artifact directory persisting all data there.

fallout/benchmarks/benchmark.yaml
absurdfarce: These servers have now been moved to the moonshot tenant

fallout/benchmarks/benchmark.yaml
absurdfarce: Current size is ms2.small based on recommendations from the Fallout team

fallout/gatling/gatling_timeseries_latency.yaml
```diff
@@ -0,0 +1,99 @@
...
+        properties:
+          graphite.create_server: true
+  server:
+    node.count: 3
+    provisioner:
+      name: ctool
+      properties:
+        cloud.provider: nebula
+        cloud.tenant: moonshot-v2
+        cloud.instance.type: ms2.small
```
adutra, line 26: Nice that we are using the new moonshot-v2 tenant, we'll have to eventually migrate everything there.

fallout/gatling/gatling_timeseries_latency.yaml
```diff
@@ -0,0 +1,99 @@
...
+---
+
+ensemble:
+  observer:
+    node.count: 1
+    provisioner:
+      name: ctool
+      properties:
+        cloud.provider: nebula
+        cloud.instance.type: m3.large
```
adutra, line 14: No tenant/project specified, do you assume that the observer can live in whichever project is the default one for the user?

fallout/gatling/gatling_timeseries_latency.yaml
```diff
@@ -0,0 +1,99 @@
...
+        cloud.instance.type: ms2.small
+    configuration_manager:
+      - name: ctool
+        properties:
+          product.type: cassandra
+          product.install.type: tarball
+          product.version: "3.11.6"
+      - name: ctool_monitoring
+        properties:
+          components: OS,JVM,GRAPH
```
adutra, line 35: Does `Graph` in this context mean "DSE Graph"? If so maybe we can remove it.

adutra: However according to the docs [here](https://fallout.sjc.dsinternal.org/components/configurationmanagers/ctool_monitoring) maybe we should add `cassandra`? We are not really interested in the cluster metrics, but maybe it's a nice to have.

fallout/gatling/gatling_timeseries_latency.yaml
```diff
@@ -0,0 +1,99 @@
...
+  client:
+    node.count: 1
+    provisioner:
+      name: ctool
+      properties:
+        cloud.provider: nebula
+        cloud.instance.type: m3.xlarge
```
adutra, line 42: same here, no tenant.

fallout/gatling/gatling_timeseries_latency.yaml
adutra: Hmm I'm unable to access this tenant; how did you create your app credentials?

adutra: Nevermind, I found it, it's on https://nebula.sjc.dsinternal.org/dashboard/identity/application_credentials/

fallout/gatling/gatling_timeseries_latency.yaml
```diff
@@ -0,0 +1,99 @@
...
+        module: gatling
+        properties:
+          maven_args: >
+              scala:testCompile
+              gatling:execute
+              -Dgatling.disableCompiler
+              -Dgatling.simulationClass=core.timeseries.Timeseries_Write
+              -DconnectionsPerHost=8
+              -DrowsPerPartition=1000
+              -DinjectionType=constant
```
adutra, line 75: I've been experimenting with throughput parameters [here](https://github.com/riptano/dse-benchmarks/blob/417eb6b411a34d583f970644c1a3825b9f41d14c/src/test/scala/core/ThroughputParameters.scala#L8-L35). It seems `linear-ramp-up` is not a valid value for time series. However `custom-ramp-up` should be close to what we want. I will experiment more when Fallout is back online (it's currently being redeployed).

fallout/gatling/gatling_timeseries_latency.yaml
absurdfarce: Yes, that was the intent.  Only the servers really need to be in the moonshot tenant, everything else can just run in whatever the default tenant might be.

fallout/gatling/gatling_timeseries_latency.yaml
absurdfarce: This entry is here largely to make sure the C* servers aren't being overloaded as part of this test.  I think "OS" and "JVM" are probably enough to accomplish that goal for now so I agree "GRAPH" should be removed.  I wouldn't say we _need_ "CASSANDRA" at this point... that seems like something we could add in later to diagnose what's causing the overload (if/when we see it).

fallout/gatling/gatling_timeseries_latency.yaml
absurdfarce: The "custom-ramp-up" injection type is exactly what I was planning on using for the real benchmarks.  In local runs something like the following does do exactly what we'd like:

```
mvn clean scala:testCompile gatling:execute -Dgatling.disableCompiler -Dgatling.simulationClass=core.timeseries.Timeseries_Write -DconnectionsPerHost=8 -DrowsPerPartition=1000 -DinjectionType=custom-ramp-up -Dthroughput.start=10 -Dthroughput.step=10 -Dthroughput.windows.count=3 -Dthroughput.windows.duration=2
```

07242020_1_start10_step10_count3

PR #1480, opened 2020-07-22
JAVA-2841: Raise timeouts during connection initialization

tomekl007:
I've tested the new timeout by connecting to Astra cluster that is in us-west from central Europe using this code:
```
public class AstraTest {
  public static void main(String[] args) {

    for (int i = 0; i < 100; i++) {
      Path bundle = Paths.get("/Users/tomaszlelek/Downloads/secure-connect-cass-dc.zip");

      try (CqlSession session =
          CqlSession.builder()
              .withCloudSecureConnectBundle(bundle)
              .withAuthCredentials("datastax", "datastax")
              .build()) {
        ResultSet set = session.execute("select * from system.local");
        System.out.println(set.all());
      }
    }
  }
}
```
with the previous timeout of 500 milliseconds, it was constantly failing to connect.
For the new timeout of 5 seconds, it has successfully connected 100 times.
I don't see a clear reason to increase `connect-timeout` timeout as well - @adutra can you elaborate more on that suggestion?

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/config/OptionsMap.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/config/OptionsMap.java
@@ -242,6 +242,8 @@ public class OptionsMap implements Serializable {
   }
 
   protected static void fillWithDriverDefaults(OptionsMap map) {
+    Duration initQueryTimeout = Duration.ofSeconds(5);
+
     // Sorted by order of appearance in reference.conf:
 
     // Skip CONFIG_RELOAD_INTERVAL because the map-based config doesn't need periodic reloading
@@ -255,8 +257,8 @@ public class OptionsMap implements Serializable {
     map.put(TypedDriverOption.LOAD_BALANCING_POLICY_SLOW_AVOIDANCE, true);
     map.put(TypedDriverOption.SESSION_LEAK_THRESHOLD, 4);
     map.put(TypedDriverOption.CONNECTION_CONNECT_TIMEOUT, Duration.ofSeconds(5));
-    map.put(TypedDriverOption.CONNECTION_INIT_QUERY_TIMEOUT, Duration.ofMillis(500));
-    map.put(TypedDriverOption.CONNECTION_SET_KEYSPACE_TIMEOUT, Duration.ofMillis(500));
+    map.put(TypedDriverOption.CONNECTION_INIT_QUERY_TIMEOUT, initQueryTimeout);
+    map.put(TypedDriverOption.CONNECTION_SET_KEYSPACE_TIMEOUT, initQueryTimeout);
     map.put(TypedDriverOption.CONNECTION_POOL_LOCAL_SIZE, 1);
     map.put(TypedDriverOption.CONNECTION_POOL_REMOTE_SIZE, 1);
     map.put(TypedDriverOption.CONNECTION_MAX_REQUESTS, 1024);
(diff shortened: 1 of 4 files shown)


adutra:
@olim7t knows better than me but afaik `connect-timeout` is for the duration of the whole connection opening process, whereas `init-query-timeout` applies to each individual request made during connection initialization. IOW, `connect-timeout` is necessarily much larger than `init-query-timeout` since it encompasses many init queries. That's why I figured that setting both timeouts to the same value doesn't make sense.

olim7t:
Actually `connect-timeout` only applies to opening the socket, it does not span protocol negotiation. So I think the current value is fine.

core/src/main/java/com/datastax/oss/driver/api/core/config/OptionsMap.java
```diff
@@ -255,7 +255,7 @@ protected static void fillWithDriverDefaults(OptionsMap map) {
     map.put(TypedDriverOption.LOAD_BALANCING_POLICY_SLOW_AVOIDANCE, true);
     map.put(TypedDriverOption.SESSION_LEAK_THRESHOLD, 4);
     map.put(TypedDriverOption.CONNECTION_CONNECT_TIMEOUT, Duration.ofSeconds(5));
-    map.put(TypedDriverOption.CONNECTION_INIT_QUERY_TIMEOUT, Duration.ofMillis(500));
+    map.put(TypedDriverOption.CONNECTION_INIT_QUERY_TIMEOUT, Duration.ofSeconds(5));
```
olim7t: `reference.conf` uses HOCON substitutions to define a few other options in terms of this one. You need to update them too: `CONNECTION_SET_KEYSPACE_TIMEOUT`, `HEARTBEAT_TIMEOUT`, `CONTROL_CONNECTION_TIMEOUT` and `REPREPARE_TIMEOUT`.

In fact we should probably extract a variable:
```java
    Duration initQueryTimeout = Duration.ofSeconds(5);
    map.put(TypedDriverOption.CONNECTION_INIT_QUERY_TIMEOUT, initQueryTimeout);
    map.put(TypedDriverOption.CONNECTION_SET_KEYSPACE_TIMEOUT, initQueryTimeout);
    ...
```

Can you also do a pass to handle all other substitutions the same way, and commit that separately?

`MapBasedDriverConfigLoaderTest.should_fill_default_profile_like_reference_file` will let you know if everything is in sync.

olim7t (changes requested):
Don't forget the changelog entry.

nit: could you follow the project's branch naming convention for future PRs? `java2841`

core/src/main/java/com/datastax/oss/driver/api/core/config/OptionsMap.java
tomekl007: done, what do you mean in a separate commit? Do you want to have two commits in the `4.x` after this branch will be merged?

adutra: Yes, because the refactoring to handle other substitutions by extracting a local variable is unrelated to this PR.

adutra (approved):
looks good, but as Olivier suggested, please do a pass to extract local variables for similar cases of hocon references in `OptionsMap`.

tomekl007:
> looks good, but as Olivier suggested, please do a pass to extract local variables for similar cases of hocon references in `OptionsMap`.

It is already done in:
https://github.com/datastax/java-driver/pull/1480/commits/acc08c04922cf6d6a99257ebb8d28de2eae3ed8b

adutra:
> It is already done in: acc08c0

No, I believe he was asking to apply the same refactoring to _other cases of hocon references_  found in reference.conf.

tomekl007:
> > It is already done in: [acc08c0](https://github.com/datastax/java-driver/commit/acc08c04922cf6d6a99257ebb8d28de2eae3ed8b)
> 
> No, I believe he was asking to apply the same refactoring to _other cases of hocon references_ found in reference.conf.

You are right, done in https://github.com/datastax/java-driver/pull/1480/commits/7ad50aafabc60e47b69aa6aa7999ea6ef9564ea8

olim7t (approved):

PR #1482, opened 2020-07-24
JAVA-2834: Node metrics for localhost cause IllegalArgumentException

tomekl007:


core/src/test/java/com/datastax/oss/driver/internal/core/metadata/DefaultEndPointTest.java
```diff
@@ -41,7 +45,32 @@ public void should_create_from_literal_ipv4_address() {
...
+  public void should_create_from_ipv6_exact_type_address() throws UnknownHostException {
+    // given
+    byte[] ipAddr = new byte[] {1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16};
+    InetAddress inet4Address = Inet6Address.getByAddress("localhost", ipAddr);
+    // when
+    DefaultEndPoint defaultEndPoint =
+        new DefaultEndPoint(new InetSocketAddress(inet4Address, 9042));
+    // then
+    assertThat(defaultEndPoint.asMetricPrefix())
+        .isEqualTo("localhost:102:304:506:708:90a:b0c:d0e:f10:9042");
```
tomekl007, line 73: without this change, this was evaluating to `locahost:9042` as well, causing `llegalArgumentException` when trying to register duplicated metric name in the `MetricRegistry`:
```
 public <T extends Metric> T register(String name, T metric) throws IllegalArgumentException {
        if (metric instanceof MetricSet) {
            registerAll(name, (MetricSet) metric);
        } else {
            final Metric existing = metrics.putIfAbsent(name, metric);
            if (existing == null) {
                onMetricAdded(name, metric);
            } else {
                throw new IllegalArgumentException("A metric named " + name + " already exists");
            }
        }
        return metric;
    }
```
- This change is covering only the case when one IP address is resolving to two: ipv4 and ipv6. I am not sure if it is possible that one address will resolve to more addresses with the same port - I think it is not. 
- We have the same logic in the `SniEndPoint#SniEndPoint` but I think that this problem should not occur there.

let me know wdyt

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultEndPoint.java
```diff
@@ -81,7 +82,16 @@ private static String buildMetricPrefix(InetSocketAddress address) {
       throw new IllegalArgumentException(
           "Could not extract a host string from provided address " + address);
     }
-    // Append the port since Cassandra 4 supports nodes with different ports
-    return hostString.replace('.', '_') + ':' + address.getPort();
+    if (address.getAddress() instanceof Inet6Address) {
+      // append getCanonicalHostName() to distinguish ip6 address from ip4
+      return hostString.replace('.', '_')
+          + ':'
+          + address.getAddress().getHostAddress()
```
adutra, line 89: `getAddress()` may return null if the address is unresolved.

I was rather thinking about something like below:

```
    String hostString =
        address.isUnresolved() ? address.getHostString() : address.getAddress().getHostAddress();
```

IOW: if the address is unresolved, there is not much we can do about it, so use `getHostString()`; if the address is resolved however, instead of relying on the original host string as provided by the user, switch to the raw IP address, since it is unique. This will solve the localhost problem.

I also think that using raw IP addresses whenever possible for metric prefixes may be beneficial since host names tend to create extremely long metric names (I tried recently with Astra, the metric names were a nightmare to read).

adutra (changes requested):

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultEndPoint.java
adutra: Also, it's worth noting that apart from contact points, any other nodes (i.e., those discovered in system.peers) will have raw IP addresses anyways. So my change above would also standardize the metrics representation of contact points with that of other nodes.

adutra: @olim7t what's your opinion on this? Is it still time to include this in 4.8?

adutra: @olim7t what's your opinion on using raw IP addresses for metric names, including for contact points? As I see it, this would have the double benefit of fixing the original issue _and_ producing homogenous metric names.
If possible let's include this in 4.9.

adutra:
Closing due to inactivity.

PR #1483, opened 2020-07-24
JAVA-2331: fix null pointer in tests that are using MockedDriverContext

tomekl007:
…tFactory


adutra (approved):
SGTM but change the commit message to: "...tests that _are_ using..."

PR #1486, opened 2020-07-30
JAVA-2721: Add counter support in the mapper

olim7t:
~Starting as draft because I'm not 100% sure about the API.~

~The DAO in the integration test uses the same entity class for all operations:~
- ~when you write an instance with `@Increment`, the fields represent the **deltas** to apply to the counters.~
- ~but when you retrieve an instance with `@Select`, you get the counter **totals**.~

EDIT- we switched to a different approach: the method declares the entity PK and each counter increment as discrete parameters. See the integration tests.

TODO:
- [x] don't default to `NullSavingStrategy.SET_TO_NULL` for legacy server versions
- [x] add manual page

### DIFF
--- a/manual/mapper/daos/increment/README.md
+++ b/manual/mapper/daos/increment/README.md
@@ -1,0 +1,86 @@
+## Increment methods
+
+Annotate a DAO method with [@Increment] to generate a query that updates a counter table that is
+mapped to an entity:
+
+```java
+// CREATE TABLE votes(article_id int PRIMARY KEY, up_votes counter, down_votes counter);
+
+@Entity
+public class Votes {
+  @PartitionKey private int articleId;
+  private long upVotes;
+  private long downVotes;
+  ... // constructor(s), getters and setters, etc.
+}
+
+@Dao
+public interface VotesDao {
+  @Increment(entityClass = Votes.class)
+  void incrementUpVotes(int articleId, long upVotes);
+
+  @Increment(entityClass = Votes.class)
+  void incrementDownVotes(int articleId, long downVotes);
+  
+  @Select
+  Votes findById(int articleId);
+}
+```
+
+### Parameters
+
+The entity class must be specified with `entityClass` in the annotation.
+
+The method's parameters must start with the [full primary key](../../entities/#primary-key-columns),
+in the exact order (as defined by the [@PartitionKey] and [@ClusteringColumn] annotations in the
+entity class). The parameter names don't necessarily need to match the names of the columns, but the
+types must match. Unlike other methods like [@Select](../select/) or [@Delete](../delete/), counter
+updates cannot operate on a whole partition, they need to target exactly one row; so all the
+partition key and clustering columns must be specified.
+
+Then must follow one or more parameters representing counter increments. Their type must be
+`long` or `java.lang.Long`. The name of the parameter must match the name of the entity
+property that maps to the counter (that is, the name of the getter without "get" and
+decapitalized). Alternatively, you may annotate a parameter with [@CqlName] to specify the
+raw column name directly; in that case, the name of the parameter does not matter:
+
+```java
+@Increment(entityClass = Votes.class)
+void incrementUpVotes(int articleId, @CqlName("up_votes") long foobar);
+```
+
+When you invoke the method, each parameter value is interpreted as a **delta** that will be applied
+to the counter. In other words, if you pass 1, the counter will be incremented by 1. Negative values
+are allowed. If you are using Cassandra 2.2 or above, you can use `Long` and pass `null` for some of
+the parameters, they will be ignored (following [NullSavingStrategy#DO_NOT_SET](../null_saving/)
+semantics). If you are using Cassandra 2.1, `null` values will trigger a runtime error.
+
+A `Function<BoundStatementBuilder, BoundStatementBuilder>` or `UnaryOperator<BoundStatementBuilder>`
+can be added as the **last** parameter. It will be applied to the statement before execution. This
+allows you to customize certain aspects of the request (page size, timeout, etc) at runtime. See
+[statement attributes](../statement_attributes/).
+
+### Return type
+
+The method can return `void`, a void [CompletionStage] or [CompletableFuture], or a
+[ReactiveResultSet].
+
+### Target keyspace and table
+
+If a keyspace was specified [when creating the DAO](../../mapper/#dao-factory-methods), then the
+generated query targets that keyspace. Otherwise, it doesn't specify a keyspace, and will only work
+if the mapper was built from a session that has a [default keyspace] set.
+
+If a table was specified when creating the DAO, then the generated query targets that table.
+Otherwise, it uses the default table name for the entity (which is determined by the name of the
+entity class and the naming convention).
+
+[@Increment]:        https://docs.datastax.com/en/drivers/java/4.8/com/datastax/oss/driver/api/mapper/annotations/Increment.html
+[ReactiveResultSet]: https://docs.datastax.com/en/drivers/java/4.8/com/datastax/dse/driver/api/core/cql/reactive/ReactiveResultSet.html
+[default keyspace]:  https://docs.datastax.com/en/drivers/java/4.8/com/datastax/oss/driver/api/core/session/SessionBuilder.html#withKeyspace-com.datastax.oss.driver.api.core.CqlIdentifier-
+[@ClusteringColumn]:      https://docs.datastax.com/en/drivers/java/4.8/com/datastax/oss/driver/api/mapper/annotations/ClusteringColumn.html
+[@PartitionKey]:          https://docs.datastax.com/en/drivers/java/4.8/com/datastax/oss/driver/api/mapper/annotations/PartitionKey.html
+[@CqlName]:             https://docs.datastax.com/en/drivers/java/4.8/com/datastax/oss/driver/api/mapper/annotations/CqlName.html
+
+[CompletionStage]:   https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html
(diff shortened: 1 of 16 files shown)


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoIncrementMethodGenerator.java
```diff
@@ -0,0 +1,190 @@
...
+    String entityParameterName = parameters.get(0).getSimpleName().toString();
+
+    // All entity properties are bound in the generated request: primary key columns in the WHERE
+    // clause, and other columns in the SET clause.
+    updateStatementBlock.addStatement(
+        "$1L.set($2L, boundStatementBuilder, $3T.$4L)",
+        helperFieldName,
+        entityParameterName,
+        NullSavingStrategy.class,
+        NullSavingStrategy.DO_NOT_SET);
```
olim7t, line 206: If there are multiple counters in the table, some of the entity fields might left to zero, I didn't filter them out here. Cassandra handles 0-increments well, but I'm not sure if that has performance implications, from a quick look at the source I don't see anywhere where it would skip them. I'll ask around.

Alternatively:
* if the server version supports unset values, you can declare your fields as `Long` and leave them to null
* or you can have one separate entity class for each combination of counters to increment

I'll document that in the manual.

adutra:
The dissymmetry between updates and retrievals is indeed concerning, especially that the same entity type is used to express deltas in the former, and absolute values in the latter.

I've been trying to think of something in Java that would express the notion of increments without ambiguity, and the best I can think of is `LongUnaryOperator`. 

Here is a suggestion: instead of passing entity instances to an `@Increment` method, pass the partition key and counter columns one by one. Counter columns should be expressed as `LongUnaryOperator` instances. Example:

```
@Increment
void increment(
    UUID productId,
    LongUnaryOperator oneStarDelta,
    LongUnaryOperator twoStarsDelta,
    LongUnaryOperator threeStarsDelta,
    LongUnaryOperator fourStarsDelta,
    LongUnaryOperator fiveStarsDelta);
```

Example of invocation:

```
dao.increment(
    productId1,
    oneStar -> oneStar + 1,
    twoStars -> twoStars,
    threeStars -> threeStars - 2,
    fourStars -> fourStars,
    fiveStars -> fiveStars + 3);
```

I believe this would make the intent of the method very clear: it's impossible to reason in absolute values inside each of the lambdas.

The generated DAO method would then contain the following code:

```
boundStatementBuilder.set("product_id", productId, UUID.class);
boundStatementBuilder = boundStatementBuilder.setLong("one_star", oneStarDelta.applyAsLong(0));
boundStatementBuilder = boundStatementBuilder.setLong("two_star", oneStarDelta.applyAsLong(0));
boundStatementBuilder = boundStatementBuilder.setLong("three_star", oneStarDelta.applyAsLong(0));
boundStatementBuilder = boundStatementBuilder.setLong("four_star", oneStarDelta.applyAsLong(0));
boundStatementBuilder = boundStatementBuilder.setLong("five_star", oneStarDelta.applyAsLong(0));
```

`LongSupplier` would work too I guess: 

```
@Increment
void increment(
    UUID productId,
    LongSupplier oneStarDelta,
    LongSupplier twoStarsDelta,
    LongSupplier threeStarsDelta,
    LongSupplier fourStarsDelta,
    LongSupplier fiveStarsDelta);

dao.increment(productId1, () -> +1, () -> 0, () -> -2, () -> 0, () -> +3);
```

But imo `LongUnaryOperator` produces a more expressive code.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoIncrementMethodGenerator.java
```diff
@@ -0,0 +1,191 @@
...
+
+    populateBuilderWithStatementAttributes(updateStatementBlock, methodElement);
+    populateBuilderWithFunction(updateStatementBlock, boundStatementFunction);
+
+    String entityParameterName = parameters.get(0).getSimpleName().toString();
+
+    // All entity properties are bound in the generated request: primary key columns in the WHERE
+    // clause, and other columns in the SET clause.
+    updateStatementBlock.addStatement(
+        "$1L.set($2L, boundStatementBuilder, isProtocolVersionV3 ? $3T.$4L : $3T.$5L)",
```
adutra: I noticed that the `set` method is implemented as below for the IT test above:

```
@Override
public <SettableT extends SettableByName<SettableT>> SettableT set(
    IncrementIT.ProductRating entity, SettableT target, NullSavingStrategy nullSavingStrategy) {

  if (entity.getProductId() != null || nullSavingStrategy == NullSavingStrategy.SET_TO_NULL) {
    target = target.set("product_id", entity.getProductId(), UUID.class);
  }

  target = target.setLong("one_star", entity.getOneStar());
  // ...
  target = target.setLong("five_star", entity.getFiveStar());

  return target;
}
```

The `if` block around the partition key column seemed strange to me. I checked quickly with CCM and it does not seem possible to leave the primary key unspecified, the statement has to contain the primary key.

olim7t: We do the same for regular `@Insert` and `@Update` methods, although it's less visible because it's buried in the entity helper's `set()` method.

I think it's more of a consequence of not having a special case for PK columns, but is it really a problem? If you try to pass an entity with null PK columns, the query will fail at runtime because those columns are absent, that's pretty consistent.

olim7t:
> The dissymmetry between updates and retrievals is indeed concerning, especially that the same entity type is used to express deltas in the former, and absolute values in the latter.
> Here is a suggestion: instead of passing entity instances to an `@Increment` method, pass the partition key and counter columns one by one.

Yeah taking the entity for increments definitely feels awkward now. We can have the method take the primary key, but the annotation will have to take the entity class. We already support that for `@Delete`. Also, taking the counters one by one allows us to allow a partial list, which solves the issue of 0-increments.

I like your idea of using lambdas, but remember that increments need to be bound to the statement. So we would need to extract the raw value, we can do that by applying the lambda to 0, but that seems a bit wasteful. Maybe a raw parameter is not so bad:
```java
@Increment(entityClass = ProductRating.class)
void increment(UUID productId, long oneStar)
```

adutra:
> Maybe a raw parameter is not so bad:
> ```
> @Increment(entityClass = ProductRating.class)
> void increment(UUID productId, long oneStar)
> ```

I can live with that, but let's be prepared to see users calling `increment(product.getId(), product.getOneStar())`. That's what I was trying to avoid.

adutra:
> applying the lambda to 0

Maybe `LongSupplier` is even better then, as you don't need to apply a function to a seed value, and they really force you to reason in terms of increments.

olim7t:
Done. This is so much cleaner.

I sticked to a simple numeric parameter. It should be pretty clear what `increment(id, value)` does.

manual/mapper/daos/increment/README.md
```diff
@@ -0,0 +1,86 @@
...
+  ... // constructor(s), getters and setters, etc.
+}
+
+@Dao
+public interface VotesDao {
+  @Increment(entityClass = Votes.class)
+  void incrementUpVotes(int articleId, long upVotes);
+
+  @Increment(entityClass = Votes.class)
+  void incrementDownVotes(int articleId, long upVotes);
```
adutra:
```suggestion
  void incrementDownVotes(int articleId, long downVotes);
```

manual/mapper/daos/increment/README.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+When you invoke the method, each parameter value will be interpreted as a **delta** that will be
+applied to the counter. In other words, if you pass 1, the counter will be incremented by 1.
+Negative values are allowed. If the parameter type is `java.lang.Long`, and the method uses
+[NullSavingStrategy#DO_NOT_SET](../null_saving/), you can pass `null` for some of the parameters,
+and they will be ignored.
+
+A `Function<BoundStatementBuilder, BoundStatementBuilder>` or `UnaryOperator<BoundStatementBuilder>`
+can be added as the **second** parameter. It will be applied to the statement before execution. This
```
adutra:
```suggestion
can be added as the **last** parameter. It will be applied to the statement before execution. This
```

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoIncrementMethodGenerator.java
```diff
@@ -0,0 +1,313 @@
...
+              Increment.class.getSimpleName());
+      return Optional.empty();
+    } else {
+      entityDefinition = context.getEntityFactory().getDefinition(entityElement);
+    }
+
+    // Validate the parameters:
+    // - all the PK components of the entity, in order.
+    // - one or more increment parameters that must match non-PK columns.
+    // - a Function<BoundStatementBuilder, BoundStatementBuilder> can be added in second position.
```
adutra:
```suggestion
    // - a Function<BoundStatementBuilder, BoundStatementBuilder> can be added in last position.
```

adutra (approved):

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/DaoIncrementMethodGenerator.java
olim7t: This is moot point with new behavior: you can declare multiple methods with however many different combinations you need, the mapper will prepare a different CQL request for each.
```java
@Increment(entityClass = ProductRating.class)
void increment(UUID productId, long counter1, long counter2);

@Increment(entityClass = ProductRating.class)
void increment(UUID productId, long counter2, long counter3);
```

PR #1487, opened 2020-07-31
JAVA-2816: Support immutability and fluent accessors in the mapper

olim7t:
This needs tests, but most of the implementation should be here.

It works on an immutable Java class, and a basic Scala case class. I need to experiment a bit more with Scala, because it seems to generate a different implementation if some fields have default values. Also I need to check what happens when some properties are marked transient, because the mapper won't have enough values to call the all-arg constructor. Maybe if the corresponding fields have default values there will be another constructor we can call, or we could require that users define one explicitly.

TODO:
- [x] finish integration tests
- [x] manual
- [x] document Scala integration
- [x] investigate support for immutable Kotlin data classes
- [x] investigate support for Java 14 records


integration-tests/src/test/java/com/datastax/oss/driver/mapper/ImmutableIT.java
```diff
@@ -0,0 +1,52 @@
...
+import com.datastax.oss.driver.api.mapper.annotations.Entity;
+import com.datastax.oss.driver.api.mapper.annotations.IntrospectionStrategy;
+import com.datastax.oss.driver.api.mapper.annotations.PartitionKey;
+import java.util.UUID;
+
+public class ImmutableIT extends InventoryITBase {
+
+  @Entity
+  @IntrospectionStrategy(getterStyle = SHORT, mutable = false)
+  public static class ImmutableProduct {
```
olim7t: No tests yet, but the generated entity helper looks correct.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -83,6 +86,15 @@ public EntityDefinition getDefinition(TypeElement processedClass) {
       typeHierarchy.add((TypeElement) context.getTypeUtils().asElement(type));
     }
 
+    boolean isScalaCaseClass = isScalaCaseClass(typeHierarchy);
+
+    Optional<IntrospectionStrategy> introspectionStrategy = getIntrospectionStrategy(typeHierarchy);
```
tomekl007: From this logic, it seems that when the class is Scala case class, the `GetterStyle` should be `SHORT` and mutable should be `false`.  Shouldn't we validate and forbid the usage of those two parameters IF the `isScalaCaseClass` is true? This will be convention over configuration. 
On the other hand, it is possible that the scala case class has a custom getter (but for sure very uncommon) so we may want to keep that.
Regarding `mutable` I think that `case class` cannot be mutable to this property could be dropped if `isScalaCaseClass == true`

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -111,31 +123,57 @@ public EntityDefinition getDefinition(TypeElement processedClass) {
...
-        boolean regularGetterName = getMethodName.startsWith("get");
-        boolean booleanGetterName =
-            getMethodName.startsWith("is")
-                && (typeMirror.getKind() == TypeKind.BOOLEAN
-                    || context.getClassUtils().isSame(typeMirror, Boolean.class));
-        if (!regularGetterName && !booleanGetterName) {
+
+        // Skip a few false positives when we don't require a "get" prefix:
+        if (getterStyle == GetterStyle.SHORT
+            && !mutable
```
tomekl007: why do we need `!mutable` check here?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -111,31 +123,57 @@ public EntityDefinition getDefinition(TypeElement processedClass) {
...
+            && (getMethodName.equals("toString") || getMethodName.equals("hashCode"))) {
+          continue;
+        }
+        if (isScalaCaseClass
+            && (getMethodName.equals("productPrefix")
+                || getMethodName.equals("productArity")
+                || getMethodName.equals("productIterator")
+                || getMethodName.startsWith("copy$default$"))) {
```
tomekl007: don't we also need `canEqual` and `productElement`? (https://www.scala-lang.org/api/2.12.4/scala/Product.html)

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -111,31 +123,57 @@ public EntityDefinition getDefinition(TypeElement processedClass) {
...
         }
+
         // skip properties we've already encountered.
         if (encounteredPropertyNames.contains(propertyName)) {
           continue;
         }
 
         ExecutableElement setMethod = findSetMethod(typeHierarchy, setMethodName, typeMirror);
-        if (setMethod == null) {
+        if (setMethod == null && mutable) {
```
tomekl007: If there is no `setMethod` doesn't it mean that class is `!mutable`?

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
adutra: These methods have parameters and are being excluded line 117.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
```diff
@@ -111,31 +123,57 @@ public EntityDefinition getDefinition(TypeElement processedClass) {
...
-        } else {
-          propertyName = Introspector.decapitalize(getMethodName.substring(2));
-          setMethodName = getMethodName.replaceFirst("is", "set");
+        switch (getterStyle) {
+          case SHORT:
+            propertyName = getMethodName;
+            setMethodName = "set" + Capitalizer.capitalize(getMethodName);
+            break;
+          case JAVABEANS:
+            boolean regularGetterName = getMethodName.startsWith("get");
```
adutra: Maybe add `&& getMethodName.length() > 3` since `get()` is a valid method name but not a valid JavaBeans getter name. Same applies for line 151 below.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
```diff
@@ -56,84 +58,123 @@ public EntityHelperGetMethodGenerator(
...
+        getBuilder.addStatement("$L.$L($L)", resultName, setterName, propertyValueName);
+      }
+    }
+
+    if (mutable) {
+      // We've already created an instance and filled the properties as we went
+      getBuilder.addStatement("return returnValue");
+    } else {
+      // Assume an all-arg constructor exists, and call it with all the temporary variables
+      getBuilder.addCode("$[return new $T(", returnType);
```
adutra, line 172: 1. Shouldn't we validate that this constructor exists? Otherwise this will create a compilation failure in a generated class.
2. How do you determine the order of parameters? Shouldn't we allow any order and match the parameters by type? Also, what is you current take on entities having many properties of the same type? I think this will inevitably lead to the creation of an annotation to disambiguate parameters à la Jackson.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
adutra: I don't think so, a mutable entity could have a method `getFoo()` but no corresponding `setFoo(Foo x)` method.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/PropertyDefinition.java
```diff
@@ -52,7 +53,8 @@
 
   /**
    * @return The name of the "set" method associated with this property used to update the value of
-   *     the property on the entity.
+   *     the property on the entity, or {@code null} if the entity was marked as {@link
+   *     IntrospectionStrategy#immutable()}.
    */
   String getSetterName();
```
adutra, line 65: Should we convert to `Optional<String>` instead?

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/entity/naming/GetterStyle.java
```diff
@@ -0,0 +1,45 @@
...
+   * "JavaBeans" style: the method name must start with "get". The name of the property is the
+   * getter name without a prefix, and decapitalized, for example {@code int getFoo() => foo}.
+   */
+  JAVABEANS,
+
+  /**
+   * "Short" style: any name will match (as long as the no-arg, not-void rule also holds), and is
+   * considered to be the property name without any prefix. For example {@code int foo() => foo}.
+   *
+   * <p>Note that this is how compiled Scala case classes appear to the annotation processor.
```
adutra:
```suggestion
   * <p>Note that this is how method names in compiled Scala case classes appear to the annotation processor.
```

adutra: Maybe add a note that this is the default if the processor detects a Scala case class?

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/entity/naming/GetterStyle.java
```diff
@@ -0,0 +1,45 @@
...
+ * <p>Note that introspection always starts by looking for getters first: no-arg, non-void methods
+ * that follow the configured style. Then the mapper will try to find a matching field (which is not
+ * required), and, if the entity is mutable, a setter.
+ *
+ * @see IntrospectionStrategy
+ */
+public enum GetterStyle {
+
+  /**
+   * "JavaBeans" style: the method name must start with "get". The name of the property is the
```
adutra: Or `is` for primitive booleans.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: The processor sets `getterStyle = SHORT, mutable = false` by default for Scala classes. But if an explicit `@IntrospectionStrategy` annotation is present, it will override those defaults. See how the two variables below are initialized:
```java
    boolean mutable =
        introspectionStrategy.map(IntrospectionStrategy::mutable).orElse(!isScalaCaseClass);
```
This way we default to the most likely behavior for Scala, but you still have the flexibility to override whatever you want.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/DefaultEntityFactory.java
olim7t: Indeed it's not strictly needed. But if the entity is mutable, the two methods would be skipped anyway because they don t have corresponding setters `setToString() / setHashCode()`, so the check is unnecessary.

Along the same lines, `getterStyle == GetterStyle.SHORT` is not mandatory either. I might just skip the two methods solely based on their names, that will be easier to follow.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
olim7t: 1. Yes. I was going to say that the validation error will manifest as a compile error as well, but it will be clearer since we can issue it on the entity type with a more precise message. I'll add that to the entity factory (and also check for a no-arg constructor for mutable entities while we're at it).
2. Not a definitive decision, but I was going for order of declaration, as parsed by the entity factory. Matching by type sounds a bit complicated.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/PropertyDefinition.java
olim7t: The method is never called for an immutable entity. I'm not convinced that we should use optional everywhere for internal stuff.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/entity/EntityHelperGetMethodGenerator.java
adutra: Is the parsing order fully deterministic? And what is that order in the presence of type inheritance?

olim7t:
I wrote and end-to-end Scala example, there are a few friction points in the build but it works. I'm not sure yet where to put it, so right now it's a standalone repo in my account: [scala-object-mapper](https://github.com/olim7t/scala-object-mapper).

integration-tests/src/test/java/com/datastax/oss/driver/mapper/ImmutableIT.java
28Smiles: I may not be qualified to do suggestions here, but I am using the driver from kotlin and in this context I was excitedly following this Issue and pull request.

Couldn't the mutabillity be infered?
 - Search for getters => extract member names
 - Search for setters => extract member names
 - mutable = getters.memberNames == setters.memberNames 

This would make all kotlin and java immutables compatible out of the box, without providing a specification via annotation. 

--
In this context java dropped the classical getter/setter naming convention in their [record types](https://openjdk.java.net/jeps/359) aswell. To say so with the introduction of java 14, anyone using records has to use this annotation too, but I have no idea how one could tackle this issue.

olim7t: Hi,

Community feedback is definitely welcome 😄 

I thought about auto-detection but it seems hard to come up with a reliable set of rules. Consider the following edge cases:

```java
// Uses Javabeans getter style, but there is also a random method that masquerades as a short-style getter
// If I encounter `int a()` first, I could legitimately assume that I'm dealing with a
// short-style immutable entity.
@Entity class Product {
  private int id;
  public int a() { return 42; };
  public int getId() { return id; }
  public void setId(int id) { this.id = id; }
  }
}

// Short-style getter, but the property name happens to look like a Javabeans getter.
// When I encounter `getMethodName()`, I could legitimately assume that I'm dealing with a Javabeans-style
// immutable entity that has a property called `methodName`.
@Entity class Field {
  private String getMethodName;
  public String getMethodName() { return getMethodName; }
  public void setGetMethodName(String getMethodName) { this.getMethodName = getMethodName; }
}
```
These are definitely convoluted cases, but since we're dealing with arbitrary input we must deal with them. Knowing the strategy beforehand makes this trivial, we can take decisions by looking at each method in isolation. It does require explicit annotation by the user, but I think that's a small price to pay since you can put the `@PropertyStrategy` annotation on a parent interface and inherit it on all your entities. We can also make educated guesses for specific languages: the PR already does so with `scala.Product` for case classes, I will check if what can be done for Kotlin data classes.

Thanks for bringing up Java 14 records too, I hadn't thought about that.

olim7t: Good news, there's a `kotlin.Metadata` annotation that gets added to every Kotlin-generated class since 1.3, and it's visible from the annotation processor. We should probably default to `mutable=false`, because that's how Kotlin users are likely using data classes.

I'll revisit the code to introduce a more generic language detection mechanism.

28Smiles: Kotlin data classes use the java bean getter style, so a mutable/immutable autodetection would be enough and totaly posible if you specified the getter/setter style, since you then know the column names. Detecting the getter style could indeed be problematic, you must either juggle around with the different types of getter/setters/constructors and find a fitting one or find a different ground trouth for the column names, so one can check if either the bean style or the short getters or setters are "column complete".

28Smiles: for java 14 records it will be the "java/lang/Record" supertype, so language detection would work there too

28Smiles: I also tested the mapper with kotlin data classes in one of our projects (thanks to jitpack) and they seem to work fine.

olim7t: Yes, Kotlin works particularly well since it's tightly integrated with the annotation processing mechanism. I don't know if you were aware of it, but we're already covering it in the [manual](https://docs.datastax.com/en/developer/java-driver/4.8/manual/mapper/config/#kotlin). The only thing that will change with this PR is that immutability will become the default, I think it makes sense because it's what people will expect with data classes.

I get your point about inferring mutability, but I think there would still be ambiguous situations, e.g. if one property has a setter, but another doesn't. One nice aspect of the current model is that we can immediately take a decision as we process each individual method. If we introduce rules like "if at least one property has a setter, then the entity is mutable and we should ignore properties that don't have setters", then we'd have to do stuff in two passes, I'm not sure if it's worth adding that complexity.

> for java 14 records it will be the "java/lang/Record" supertype, so language detection would work there too

👍 
Done. There's just a small hack when we validate the annotation's target, because they introduced a new `ElementKind.RECORD`. But apart from that it's transparent for the rest of the processor.

28Smiles: > I get your point about inferring mutability, but I think there would still be ambiguous situations, e.g. if one property has a setter, but another doesn't. One nice aspect of the current model is that we can immediately take a decision as we process each individual method. If we introduce rules like "if at least one property has a setter, then the entity is mutable and we should ignore properties that don't have setters", then we'd have to do stuff in two passes, I'm not sure if it's worth adding that complexity.

Agree. 

Thank you for this

manual/mapper/entities/README.md
```diff
@@ -37,19 +37,105 @@ public class Product {
...
+  * *may* have a corresponding field: matching name and type. You'll probably want to make that
+    field final (although that has no impact on the mapper-generated code).
+* the type **must** expose a non-private constructor that takes every
+  non-[transient](#transient-properties) property, in the declaration order.
   
-The class must expose a no-arg constructor that is at least package-private.
+When the mapper reads an immutable entity from the database, it will first read all properties, then
+invoke the "all columns" constructor to materialize the instance.
+
+Note: the "all columns" constructor must take the properties in the order that they are declared in
```
adutra, line 100: Nice! I was about to ask this question. Glad to know that the error message is explicit enough.

manual/mapper/entities/README.md
```diff
@@ -37,19 +37,105 @@ public class Product {
...
+the entity. If the entity inherits properties from parent types, those must come last in the
+constructor signature, ordered from the closest parent to the farthest. If things get too
+complicated, a good trick is to deliberately omit the constructor to let the mapper processor fail:
+the error message describes the expected signature.
+
+#### Getter style
+
+By default, the mapper looks for JavaBeans-style getters prefixed with "get":
```
adutra: ... or `is` for primitive booleans.

manual/mapper/entities/README.md
```diff
@@ -37,19 +37,105 @@ public class Product {
...
+@PropertyStrategy(getterStyle = SHORT)
+public class Product {
+  @PartitionKey private UUID productId;
+  
+  public UUID productId() { return productId; }
+  public void setProductId(UUID productId) { this.productId = productId; }
+}
+```
+
+The getter name is just the property name, without any prefix. The setter name is unchanged.
```
adutra: Shouldn't we expect the setter name to be short as well? As in `public void productId(UUID productId)`.

Thinking even further, I sometimes wonder if the notion of "property strategy" is really independent from the notion of "Immutable entity", because short getters will only really happen when the entity is also immutable. So shouldn't we call this `@ImmutableGetterStyle`?

adutra (approved):

manual/mapper/entities/README.md
olim7t: > Shouldn't we expect the setter name to be short as well? 

I've never seen short setters before, is that a thing?

> short getters will only really happen when the entity is also immutable

That's the case for our two built-in use cases (Scala case classes and Java 14 records), but I don't think we should make it a rule. I can see someone using mutable POJOs with short getters; that's a bit unusual but why go out of our way to prevent it?

adutra: OK, but right now your example of mutable short getters looks like a mix of two accessor styles: short getters and Java Beans setters. I don't know if short setters are a thing, but I can hardly imagine that someone would write a class with this specific combination of styles for getters and setters.

28Smiles: Sorry, for intervening again. I get emails of every conversation and can not resist intervening.

> I've never seen short setters before, is that a thing?

```scala
case class Stuff(name: String, var value: String)
```
will generate a "short setter" for value. So they are a thing. But nearly nobody should use this, since there is the `copy` function:
```scala
val s1 = Stuff("bashan", "a")
val s2 = s1.copy(value = "hello")
```

Also Java 14 records afaik will use short setters as well.

> OK, but right now your example of mutable short getters looks like a mix of two accessor styles

Mixing styles is indeed questionable.

olim7t: I think mutability in Scala is rare enough that we don't need to address it. But anyway case class setters (seen from the annotation processor) use yet another convention:
```java
public void value_$eq(final String x$1) {
   this.value = x$1;
}
```

Java 14 records are always immutable as far as I can tell. There's no mention of mutability in the javadocs.

> I don't know if short setters are a thing, but I can hardly imagine that someone would write a class with this specific combination of styles for getters and setters.

I would :trollface: Having two methods with the same name that do different things is the one that looks weird to me.

But to be fair, this seems to be a thing known as "fluent" accessor in some DDD circles. Also I just noticed that Lombok has an option for that, and it shortens both methods.
So I guess we're in for a `SetterStyle`. I'll change `SHORT` to `FLUENT` too.

manual/mapper/config/lombok/README.md
```diff
@@ -0,0 +1,90 @@
...
+import lombok.Data;
+import lombok.experimental.Accessors;
+import com.datastax.oss.driver.api.mapper.annotations.*;
+import com.datastax.oss.driver.api.mapper.entity.naming.*;
+
+@Data
+@Accessors(fluent = true)
+@Entity
+@PropertyStrategy(getterStyle = GetterStyle.FLUENT, setterStyle = SetterStyle.FLUENT)
+public static class Product {
```
olim7t, line 52: I did a quick test with the example Lombok project to confirm that this works.

mapper-runtime/src/main/java/com/datastax/oss/driver/api/mapper/entity/naming/SetterStyle.java
```diff
@@ -0,0 +1,44 @@
...
+ * The style of setter that the mapper will look for when introspecting a mutable entity class.
+ *
+ * <p>Note that introspection always starts by looking for getters first (see {@link GetterStyle}).
+ * Once a getter has been found, and if the entity is declared as {@link PropertyStrategy#mutable()
+ * mutable}, the mapper will try to find a matching setter: name inferred as described below,
+ * exactly one argument matching the property type, and the return type does not matter.
+ *
+ * @see PropertyStrategy
+ */
+public enum SetterStyle {
```
olim7t, line 30: I deliberately went for a distinct enum, even though the constants are the same right now. This gives us flexibility, for example if we were to implement Scala case class setters at some point in the future.

olim7t (commented):
Setter styles.

adutra (approved):

olim7t:
Merged, thanks everyone for the feedback.

PR #1488, opened 2020-08-04
JAVA-2857: Fix NPE when built statements without parameters are logged at TRACE level

adutra:


driver-core/src/test/java/com/datastax/driver/core/QueryLoggerTest.java
```diff
@@ -414,6 +414,64 @@ public void should_log_normal_queries() throws Exception {
...
+  @CassandraVersion("2.0.0")
+  public void should_log_simple_statements_without_parameters() throws Exception {
+    // given
+    normal.setLevel(TRACE);
+    queryLogger = QueryLogger.builder().build();
+    cluster().register(queryLogger);
+
+    // when
+    String query = "UPDATE test SET c_int = 42 WHERE pk = 42";
+    SimpleStatement stmt = new SimpleStatement(query);
```
tomekl007, line 428: I think that those tests could be refactored to one parameterized test. All logic besides statement creation is the same.

adutra: It's very hard to prepare a statement inside a `@DataProvider` method because the method is static. Frankly let's not complicate this PR too much, it's a very simply fix that needs to be shipped asap.

adutra:
CI looking good, failures are unrelated. `QueryLoggerTest` passed for all backends.

tomekl007 (approved):

emerkle826 (approved):

PR #1489, opened 2020-08-04
JAVA-2843: Successfully parse DSE table schema in OSS driver

adutra:


emerkle826 (approved):

Jenkinsfile
```diff
@@ -457,7 +457,11 @@ pipeline {
             values '2.1',       // Legacy Apache CassandraⓇ
                    '3.0',       // Previous Apache CassandraⓇ
                    '3.11',      // Current Apache CassandraⓇ
-                   '4.0'        // Development Apache CassandraⓇ
+                   '4.0',       // Development Apache CassandraⓇ
+                   'dse-5.1',   // Legacy DataStax Enterprise
+                   'dse-6.0',   // Previous DataStax Enterprise
+                   'dse-6.7',   // Previous DataStax Enterprise
+                   'dse-6.8.0'  // Current DataStax Enterprise
```
emerkle826, line 486: Around line 218-222, we list out the Cassandra versions that you can choose from for triggering a specific build. It would be more complete to add the DSE versions there as well.

emerkle826 (changes requested):
Just one small update to the Jenkins file for completeness.

tomekl007 (commented):
Could you please add at least one unit test that captures the problem and solution?

emerkle826 (approved):

adutra:
> Could you please add at least one unit test that captures the problem and solution?

It is unfortunately almost impossible to create unit tests for the schema parser in driver 3. The `SchemaParser` class itself needs access to a `Cluster` object and many metadata classes, and all of them are hard to mock because they are concrete classes. Furthermore, the `SchemaParser` instance is accessed from `ControlConnection` which is impossible to mock as well. That's why all the tests for metadata classes are integration tests with CCM.

So the test you are looking for is `com.datastax.driver.core.TableMetadataTest#should_parse_table_options`; if you try to run this test from the 3.x branch with DSE 6.0 or DSE 6.7, it will fail. In this branch it succeeds.

driver-core/src/main/java/com/datastax/driver/core/TableOptionsMetadata.java
```diff
@@ -183,11 +182,12 @@
...
-    if (is380OrHigher) this.cdc = isNullOrAbsent(row, CDC) ? DEFAULT_CDC : row.getBool(CDC);
-    else this.cdc = DEFAULT_CDC;
+    this.cdc = isNullOrAbsent(row, CDC) ? DEFAULT_CDC : row.getBool(CDC);
 
-    if (is400OrHigher) this.additionalWritePolicy = row.getString(ADDITIONAL_WRITE_POLICY);
-    else this.additionalWritePolicy = DEFAULT_ADDITIONAL_WRITE_POLICY;
+    this.additionalWritePolicy =
+        isNullOrAbsent(row, ADDITIONAL_WRITE_POLICY)
+            ? DEFAULT_ADDITIONAL_WRITE_POLICY
+            : row.getString(ADDITIONAL_WRITE_POLICY);
```
absurdfarce: One (quite possibly minor) point: we're losing a bit of insurance against aberrant cases by excluding the version guards, right?  The original code _expected_ fields to be present for certain versions.  The new code will silently handle missing fields silently due to isNullOrAbsent() even if a field should be present for a specific version.

This should only matter if, say, a server bug causes expected fields to not be returned for some reason.  I'm merely mentioning it to say we do lose that notion of version-based expectation here.

absurdfarce: On a related point: it seems a bit odd that after this change some fields still have version guards while some don't.  Perhaps we should make this change wholesale to all fields examined by this class?

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
```diff
@@ -1047,13 +1047,15 @@ public CCMBridge build() {
...
         cassandraConfiguration.remove("thrift_prepared_statements_cache_size_mb");
       }
-      if (isMaterializedViewsDisabledByDefault(cassandraVersion)) {
-        // enable materialized views
-        cassandraConfiguration.put("enable_materialized_views", true);
-      }
-      if (isSasiConfigEnablementRequired(cassandraVersion)) {
-        // enable SASI indexing in config (disabled by default in C* 4.0)
-        cassandraConfiguration.put("enable_sasi_indexes", true);
+      if (!dse) {
```
absurdfarce, line 1054: Perhaps the DSE check should be folded into isMaterializedViewsDisabledByDefault() and isSasiConfigEnablementRequired() instead?  Seems like that would make those two methods more accurate in general.

driver-core/src/main/java/com/datastax/driver/core/TableOptionsMetadata.java
adutra: > I'm merely mentioning it to say we do lose that notion of version-based expectation here.

Correct, but as you pointed out, I don't see how this would matter unless the server is misbehaving.

> it seems a bit odd that after this change some fields still have version guards while some don't.

Indeed, but for these fields, we distinguish 3 states: 

1. the option is supported and present: use the value in the row
2. the option is supported but not present: use the default value instead
3. the option is not supported: use null

The version guard is required for item 3.

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
adutra: That was actually my first version of this snippet :-) But the implementations felt clumsy to me as both would start with a `if (dse) return false;` statement that seemed disconnected from the rest. Anyways, it doesn't matter much, there is no version of DSE as of today that support these config options.

driver-core/src/main/java/com/datastax/driver/core/TableOptionsMetadata.java
absurdfarce: Fair point on the remaining version bounds.  And I don't feel super-strongly about the fact that we're no longer considering whether a field _should be_ provided for the changes here.  I agree that it should only figure in when the server is misbehaving... my concern was that if that were to happen it might be better for my client to fail loudly.

Regardless, I should've prefixed all of this with the "nit" label; none of it is fatal.  What you have should be fine.

absurdfarce (approved):
Validated issue with a simple connect using 3.10.x, confirmed errors were no longer present on this branch and queries returned as expected.  Mentioned a couple of small nits observed while looking at the code but I believe @adutra has satisfactorily addressed all of them.

driver-core/src/test/java/com/datastax/driver/core/CCMBridge.java
absurdfarce: It's the fact that no DSE version uses these configs that formed the foundation of my concern... that's why you want those methods to be as accurate as possible.  But we're talking about private methods in CCMBridge so it's hard to get too concerned... this also falls squarely under the category of "nit".

PR #1492, opened 2020-08-19
JAVA-2863: Reintroduce mapper processor dependency to SLF4J

olim7t:
Motivation:

In JAVA-2800 we removed that dependency to avoid SLF4J's "defaulting to
NOP" warning when the processor in configured via `-processorpath`.
However, the processor depends on driver core classes that sometimes
declare a logger. If SLF4J is missing, this will cause a
ClassNotFoundException while processing annotations.

Some more background:

- in theory the processor should be completely isolated from SLF4J (its
  only output is compiler messages). But sometimes reusing core classes
  is convenient, we wouldn't want to rewrite them from scratch. A NOP
  implementation that ignores the logs is what we want, it's just the
  warning that is undesirable.
- we can't declare slf4j-nop as an explicit dependency of the processor,
  because then it would collide with the application's binding if the
  processor is configured via the regular classpath instead of
  `-processorpath`.

Modifications:

Revert JAVA-2800 to reintroduce the dependency.
Add a snippet in the manual to explain how users can add slf4j-nop to
their processorpath to avoid the "defaulting to NOP" warning.

Result:

No more risk of ClassNotFoundException.
The warning is back, but it's not harmful and users have a workaround to
remove it.


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/dao/LoggingGenerator.java
```diff
@@ -49,8 +46,12 @@ public void addLoggerField(TypeSpec.Builder classBuilder, ClassName className) {
...
       classBuilder.addField(
           FieldSpec.builder(
-                  LOGGER_CLASS_NAME, "LOG", Modifier.PRIVATE, Modifier.FINAL, Modifier.STATIC)
-              .initializer("$T.getLogger($T.class)", LOGGER_FACTORY_CLASS_NAME, className)
+                  ClassName.get(Logger.class),
+                  "LOG",
+                  Modifier.PRIVATE,
+                  Modifier.FINAL,
+                  Modifier.STATIC)
+              .initializer("$T.getLogger($T.class)", LoggerFactory.class, className)
```
olim7t: Not mandatory, but I figured we might as well revert this too, since SLF4J types will now be available at processing time.

adutra: I'm unsure about this, we are importing a class just to feed it to javapoet code generator (as opposed to using it to actually log messages). The previous version seemed a bit more accurate to me.

olim7t: Either way works for me, I will revert this part.

PR #1493, opened 2020-08-20
JAVA-2864: Revisit mapper processor's messaging

olim7t:
TODO:
- [x] inline two deprecated methods in DecoratedMessager

### DIFF
--- a/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/DecoratedMessager.java
+++ b/mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/DecoratedMessager.java
@@ -15,18 +15,19 @@
  */
 package com.datastax.oss.driver.internal.mapper.processor;
 
-import com.datastax.oss.driver.internal.core.util.Reflection;
-import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.Set;
 import javax.annotation.processing.Messager;
 import javax.lang.model.element.Element;
-import javax.lang.model.element.ExecutableElement;
-import javax.lang.model.element.TypeElement;
 import javax.tools.Diagnostic;
 
 /** Wraps {@link Messager} to provide convenience methods. */
 public class DecoratedMessager {
 
   private final Messager messager;
+  private final Set<MessageId> emittedMessages = new HashSet<>();
 
   public DecoratedMessager(Messager messager) {
     this.messager = messager;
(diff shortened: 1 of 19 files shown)


mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/DecoratedMessager.java
```diff
@@ -37,135 +39,93 @@ public void warn(String template, Object... arguments) {
...
-  /**
-   * Emits a warning for a program element that might be inherited from another type.
-   *
-   * @param targetElement the element to target.
-   * @param processedType the type that we were processing when we detected the issue.
-   */
+  /** @deprecated TODO inline this once JAVA-2721 and JAVA-2816 are merged */
+  @Deprecated
   public void warn(
       Element targetElement, TypeElement processedType, String template, Object... arguments) {
```
olim7t: I've left these two methods to avoid unrelated impacts for now, but they should be inlined after this is merged together with the other pending mapper PRs.

adutra:
Could you give us some examples of messages that would change with this PR? I'm struggling to understand the changes.

olim7t:
> Could you give us some examples of messages that would change with this PR? I'm struggling to understand the changes.

You can generate a warning by adding this on `InventoryITBase.Product`:
```java
  @NamingStrategy(
      convention = {NamingConvention.SNAKE_CASE_INSENSITIVE, NamingConvention.UPPER_SNAKE_CASE})
  public static class Product {
```

Then:
```
cd mapper-processor
mvn install -DskipTests
cd ../integration-tests
mvn test-compile
```

Without the fix, you get 4 of these:
```
[WARNING] /path/to/com/datastax/oss/driver/mapper/InventoryITBase.java:[106,17]
Too many naming conventions: NamingStrategy must have at most one 'convention' argument (will use the first one: SNAKE_CASE_INSENSITIVE)
```

With the fix, you get 1 of these:
```
[WARNING] /path/to/com/datastax/oss/driver/mapper/InventoryITBase.java:[106,17]
[Product] Too many naming conventions: NamingStrategy must have at most one 'convention' argument (will use the first one: SNAKE_CASE_INSENSITIVE)
```

If the processor is operating on class files, you don't get the `InventoryITBase.java:[106,17]` part, so the important part is to have `[Product]` in the message itself.
I didn't use the fully-qualified name to keep it short; in our tests this is ambiguous, but in real applications I think there won't be multiple entities with the same name, so the simple name should be clear enough.

mapper-processor/src/main/java/com/datastax/oss/driver/internal/mapper/processor/DecoratedMessager.java
```diff
@@ -37,135 +39,93 @@ public void warn(String template, Object... arguments) {
...
+    @Override
+    public boolean equals(Object other) {
+      if (other == this) {
+        return true;
+      } else if (other instanceof MessageId) {
+        MessageId that = (MessageId) other;
+        return this.level == that.level
+            && Objects.equals(this.element, that.element)
+            && Objects.equals(this.template, that.template)
+            && Arrays.deepEquals(this.arguments, that.arguments);
```
tomekl007, line 105: will this `deepEquals` give us expected behavior? Maybe it would be better to convert the 
`Object[] arguments ` to `List` and do `Objects.equals()` on it?

adutra: Why wouldn't `deepEquals` give "expected behavior"? Its contract is very well defined:

> Two array references are considered deeply equal if both are null, or if they refer to arrays that contain the same number of elements and all corresponding pairs of elements in the two arrays are deeply equal.

adutra (approved):

tomekl007 (approved):

olim7t:
I tried with the Scala example, it's a lifesaver since there is no context whatsoever when you process bytecode:
```
[info] Processing Object Mapper annotations in compiled Scala classes...
error: [UserDao.increment] Missing entity class: Increment methods must always have an 'entityClass' argument
1 error
[error] Failed to run Object Mapper processor
```

PR #1494, opened 2020-08-21
JAVA-2855: Provide user-friendly way of selecting metrics framework

emerkle826:
Motivation:

As of JAVA-2808, the driver can integrate with Micrometer or MicroProfile metrics frameworks (as well as the default DropWizard metrics framework). However, the process to integrate with a non-DropWizard metrics framework requires applications to implement a custom SessionBuilder that returns the desired metrics framework enabled DriverContext. While this method is not overly burdensome, it would be easier for application developers to integrate their desired metrics framework with less code to override driver mechanisms.

Proposal:

This PR adds a configuration option for a MetricsFactory implementation class to enable the desired metrics framework.


core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOption.java
```diff
@@ -816,6 +816,13 @@
    * <p>Value-type: {@link java.time.Duration Duration}
    */
   METRICS_NODE_EXPIRE_AFTER("advanced.metrics.node.expire-after"),
+
+  /**
+   * The fully-qualified classname of the desired MetricsFactory implementation.
```
emerkle826: The implementation does not need a fully-qualified classname for implementations that exist in `com.datastax.oss.driver.internal.core.metrics`. Remove `fully-qualified` from the javadoc.

core/src/main/java/com/datastax/oss/driver/api/core/config/TypedDriverOption.java
```diff
@@ -705,6 +705,10 @@ public String toString() {
   public static final TypedDriverOption<Duration> METRICS_NODE_EXPIRE_AFTER =
       new TypedDriverOption<>(DefaultDriverOption.METRICS_NODE_EXPIRE_AFTER, GenericType.DURATION);
 
+  /** The fully-qualified classname of the desired MetricsFactory implementation. */
```
emerkle826: remove `fully-qualified`.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -650,6 +650,12 @@ public SelfT withApplicationVersion(@Nullable String applicationVersion) {
     return self;
   }
 
+  @NonNull
```
emerkle826, line 670: Needs Javadocs.

core/src/main/java/com/datastax/oss/driver/internal/core/context/InternalDriverContext.java
```diff
@@ -180,4 +180,7 @@
    */
   @NonNull
   RequestLogFormatter getRequestLogFormatter();
+
+  @Nullable
```
emerkle826, line 195: Needs Javadocs.

core/src/main/java/com/datastax/oss/driver/api/core/session/ProgrammaticArguments.java
```diff
@@ -289,6 +298,12 @@ public Builder withCodecRegistry(@Nullable MutableCodecRegistry codecRegistry) {
       return this;
     }
 
+    @NonNull
+    public Builder withMetricRegistry(@Nullable Object metricRegistry) {
```
adutra, line 302: How is this design going to solve the Spring Boot issue? You would need access to the metric registry bean at session build time.

adutra: A possible solution would be to provide a `Supplier<Object>`.

adutra: Oooh I see. `SessionBuilderCustomizer` FTW! Very smart :-)

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerMetricsFactory.java
```diff
@@ -87,9 +91,16 @@ public MicrometerMetricsFactory(
...
-      this.sessionUpdater =
-          new MicrometerSessionMetricUpdater(enabledSessionMetrics, this.registry, this.context);
+      // try to get the mertic registry from the context
+      Object possibleMetricRegistry = context.getMetricRegistry();
+      if (possibleMetricRegistry instanceof MeterRegistry) {
+        this.registry = (MeterRegistry) possibleMetricRegistry;
+        this.sessionUpdater =
+            new MicrometerSessionMetricUpdater(enabledSessionMetrics, this.registry, this.context);
+      } else {
+        this.registry = null;
```
adutra: Maybe we should throw an error if an object was present but was not of the expected type?

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -650,6 +651,25 @@ public SelfT withApplicationVersion(@Nullable String applicationVersion) {
     return self;
   }
 
+  /**
+   * The metrics registry object for storing driver metrics.
```
adutra: Very good javadocs 👍

core/src/main/resources/reference.conf
```diff
@@ -1228,6 +1228,15 @@ datastax-java-driver {
...
 
   advanced.metrics {
+    # Metrics Factory configuration.
+    factory {
+      # The class for the metrics factory.
+      #
+      # The driver provides Dropwizard metrics out of the box. If you would like to use a
+      # different metrics framework, change the factory class to the fully-qualified name of a
+      # class that implements com.datastax.oss.driver.internal.core.metrics.MetricsFactory.
+      class = DropwizardMetricsFactory
```
adutra, line 1251: I think you could add: 

```
    # Required: yes
    # Modifiable at runtime: no
```

Also, when you say "The driver provides Dropwizard metrics out of the box" I think you should also mention the 2 alternative factories that are also provided now (micrometer et microprofile).

core/src/main/java/com/datastax/oss/driver/internal/core/context/InternalDriverContext.java
```diff
@@ -180,4 +180,20 @@
...
+   * <p>This will return the object from {@link
+   * SessionBuilder#withMetricRegistry(java.lang.Object)}. Access to this registry object is only
+   * intended for {@link MetricsFactory} implementations that need to expose a way to specify the
+   * registry external to the Factory implementation itself.
+   *
+   * <p>The default metrics framework used by the Driver is DropWizard and does not need an external
+   * metrics registry object.
+   */
+  @Nullable
+  default Object getMetricRegistry() {
```
adutra, line 196: I'm going to sound a bit pedantic but I wonder if we shouldn't put "metrics" in the plural everywhere: `getMetricsRegistry`, etc.. Dropwizard and Microprofile have `MetricsRegistry` while Microprofile is the only one to have `MetricRegistry` in the singular.

metrics/microprofile/src/main/java/com/datastax/oss/driver/internal/metrics/microprofile/MicroProfileMetricsFactory.java
```diff
@@ -86,9 +90,16 @@ public MicroProfileMetricsFactory(
...
-      this.sessionUpdater =
-          new MicroProfileSessionMetricUpdater(enabledSessionMetrics, this.registry, this.context);
+      Object possibleMetricRegistry = context.getMetricRegistry();
+      if (possibleMetricRegistry instanceof MetricRegistry) {
+        this.registry = (MetricRegistry) possibleMetricRegistry;
+        this.sessionUpdater =
+            new MicroProfileSessionMetricUpdater(
+                enabledSessionMetrics, this.registry, this.context);
+      } else {
+        this.registry = null;
```
adutra: Same here.

adutra (changes requested):
Excellent work. I don't find it invasive at all to be honest.

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerMetricsFactory.java
emerkle826: Will do. But I just also realized that the registry could be NULL, which will cause an NPE here as well. I'll fix this and write some unit tests to catch it.

core/src/main/java/com/datastax/oss/driver/internal/core/context/InternalDriverContext.java
olim7t: I never shy from a pedantry contest 😄 

Dropwizard actually uses the singular form: [MetricRegistry](https://www.javadoc.io/doc/io.dropwizard.metrics/metrics-core/4.1.2/com/codahale/metrics/MetricRegistry.html). I like it a bit better, it's like apple pie, ClassLoader, LoggerFactory, etc.

PS -- I did write a `MetricsFactory` interface in the driver, but it returns an instance of the `Metrics` object.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -650,6 +651,25 @@ public SelfT withApplicationVersion(@Nullable String applicationVersion) {
...
+   *
+   * <p>If using a non-default metrics framework, it is necessary to call this method <b>only if</b>
+   * the {@link MetricsFactory} implementation does not provide its own registry for metrics
+   * storage.
+   *
+   * <p>If using Micrometer or MicroProfile driver implementations, an acceptable registry object is
+   * expected to be set with this method.
+   */
+  @NonNull
+  public SelfT withMetricsRegistry(@Nullable Object metricRegistry) {
```
olim7t: OK, we lose a bit of precision on the type but the ability to pass a registry even with alternative implementations is worth it.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
```diff
@@ -650,6 +651,25 @@ public SelfT withApplicationVersion(@Nullable String applicationVersion) {
     return self;
   }
 
+  /**
+   * The metrics registry object for storing driver metrics.
+   *
+   * <p>The default driver metrics framework is DropWizard and does not need an external metrics
+   * registry object to be provided.
```
olim7t: It does not _need_ it, but maybe we should use the object if it is provided. Right now our implementation creates an instance no matter what, but Dropwizard can publish to an existing registry like the other frameworks.

core/src/main/resources/reference.conf
```diff
@@ -1228,6 +1228,25 @@ datastax-java-driver {
...
   advanced.metrics {
+    # Metrics Factory configuration.
+    factory {
+      # The class for the metrics factory.
+      #
+      # The driver provides Dropwizard, Micrometer and MicroProfile metrics out of the box.
+      # To use Dropwizard, this value should be set to "DropwizardMetricsFactory". To use
+      # Micrometer or MicroProfile, this value needs to be set to the fully qualified class
+      # name "com.datastax.oss.driver.internal.metrics.micrometer.MicrometerMetricsFactory"
+      # or "com.datastax.oss.driver.internal.metrics.microprofile.MicroProfileMetricsFactory".
```
olim7t: You can declare those two packages as additional defaults, so that all of our built-in implementations work with the unqualified name.

See `DefaultDriverContext.buildMetricsFactory`, the last parameter of `Reflection.buildFromConfig` is a vararg.

olim7t (approved):
LGTM, see minor remarks inline.

We should also revisit the manual now that subclassing the context is no longer necessary, and maybe move it to the regular manual (as opposed to developer docs).
Also cover the fact that `Session.getMetrics()` only works with Dropwizard.

core/src/main/java/com/datastax/oss/driver/internal/core/context/InternalDriverContext.java
adutra: :-) You are right, let's keep the singular form then: `MetricRegistry`. Sorry for initiating this storm in a teacup.

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerMetricsFactory.java
```diff
@@ -87,9 +91,28 @@ public MicrometerMetricsFactory(
...
+        this.registry = (MeterRegistry) possibleMetricRegistry;
+        this.sessionUpdater =
+            new MicrometerSessionMetricUpdater(enabledSessionMetrics, this.registry, this.context);
+      } else {
+        // Metrics are enabled, but the registry object is not an expected type
+        throw new IllegalArgumentException(
+            "Unexpected Metrics registry object. Expected registry object to be of type '"
+                + MeterRegistry.class.getCanonicalName()
+                + "', but was '"
+                + possibleMetricRegistry.getClass().getCanonicalName()
```
adutra: Prefer `getName()` to `getCanonicalName()` as the latter would return null for anonymous inner classes (which could very well be the case for the object passed to the factory).

core/src/main/java/com/datastax/oss/driver/internal/core/context/InternalDriverContext.java
emerkle826: Will do!

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
emerkle826: In general, should the MetricsFactory implementations try to create a Registry object, even if one isn't set on the Context? That is what DropWizardMetricsFactory does now. I can do it for Micrometer as well. But unless we add a compile time dependency on SmallRye, I can't spin up a concrete instance for MircoProfile.

The reason I ask is for consistency. We can add SmallRye as a dependency, and then each of the Factory classes can use a registry object if provided, or use a default implementation if not. Or we could require an appropriate registry object to be set on the context. I'm thinking the former approach may be best as it will capture metrics if an application forgets to set a registry object.

adutra: > But unless we add a compile time dependency on SmallRye, I can't spin up a concrete instance for MircoProfile.

Maybe use reflection for that?

emerkle826: Looking at this a bit more, it's going to be difficult to really test Micrometer and MicroProfile metrics if an explicit Registry object is not provided. The Factory classes currently do _not_ implement `getMetrics()`, only Dropwizard does. So if applications setup Micrometer or MicroProfile and do not also specify the metric Registry, they won't have a way to access driver metrics collected.

For Dropwizard, even if an application doe not provide a Registry, the application can still access the metrics from `getMterics()`.

@adutra @olim7t Should I add an accessor method to the MetricsFactory, or some other way to access the metrics when an explicit registry is not specified (for Micrometer and MicroProfile)? Or should we rstick with requiring an external Registry object to be provided in these cases, and only allow Dropwizard to not specify a Registry object?

adutra: > The Factory classes currently do not implement getMetrics()

Implementing `getMetrics()` may be an interesting addition that stems quite naturally from the general trend of opening up to different metrics implementations. However I don't think that it's required to be done within this PR. It can safely be implemented later as a follow-up ticket. Our current work on Spring Boot does not require that either, so I'd suggest that we keep this PR focused on providing an external registry.

> Should I add an accessor method to the MetricsFactory, or some other way to access the metrics when an explicit registry is not specified (for Micrometer and MicroProfile)? Or should we stick with requiring an external Registry object to be provided in these cases, and only allow Dropwizard to not specify a Registry object?

My suggestions are:

1. For consistency, allow all `MetricsFactory` implementations to use the external registry, if provided; if not, they should either instantiate one on the fly, or error out if the external registry is required.
    a. For microprofile, you _might_ get away by instantiating the SmallRye registry by reflection; but throwing an error is fine too imo.
3. For integration tests, always pass an external registry; this would hopefully make verifications of the registry's contents straightforward, even in the absence of a proper `getMetrics()` implementation.

adutra: What I have in mind for MicroProfile:

```
      Object possibleMetricRegistry = context.getMetricsRegistry();
      if (possibleMetricRegistry == null) {
        // metrics are enabled, but a metrics registry was not supplied to the context;
        // in this case, attempt to instantiate SmallRye's implementation
        Class<?> smallRyeRegistryClass =
            Reflection.loadClass(
                context.getClassLoader(), "io.smallrye.metrics.MetricsRegistryImpl");
        if (smallRyeRegistryClass != null) {
          try {
            possibleMetricRegistry = smallRyeRegistryClass.newInstance();
          } catch (Exception ignored) {
          }
          if (possibleMetricRegistry == null) {
                "Metrics registry object is NULL. Expected registry object to be of type '"
                    + MetricRegistry.class.getCanonicalName()
                    + "'");
          }
        }
      }
```

The code above could be simplified if we move this instantiation logic to `Reflection`. Actually, the logic is already there in `buildFromConfig` but we would need to extract a `buildFromClassName` method out of it.

adutra: And Micrometer:

```
      // try to get the metric registry from the context
      Object possibleMetricRegistry = context.getMetricsRegistry();
      if (possibleMetricRegistry == null) {
        // metrics are enabled, but a metrics registry was not supplied to the context;
        // use global registry in this case.
        possibleMetricRegistry = io.micrometer.core.instrument.Metrics.globalRegistry;
      }
```

emerkle826: The only problem with this approach is access to the Registry if it's not explicitly set on the context. Metrics will be collected, but the reference to the Registry is private inside the Factory. And since there is no `getRegistry()` method on the Factory, client code wouldn't have any way to get at the Metrics being stored.

It's different for Dropwizard, because we have `getMetrics()` on MetricsFactory. So client code doesn't need access to the Registry to get at the stored metrics.

I only realized this when writing the tests for it as I had implemented almost exactly what you suggest, but I wasn't able to get the test to verify the metrics.

emerkle826: I've implemented the suggestion to error out for Micrometer and MicroProfile if no external registry is provided. For Dropwizard, I create a registry to keep the behavior the same. See this [commit](https://github.com/datastax/java-driver/pull/1494/commits/247aa46145c5f7afc88b4ab73948aa9ef32dca18)

adutra: > The only problem with this approach is access to the Registry if it's not explicitly set on the context. Metrics will be collected, but the reference to the Registry is private inside the Factory. And since there is no `getRegistry()` method on the Factory, client code wouldn't have any way to get at the Metrics being stored.

You are right, sorry for not anticipating that. 

That said, unless I'm mistaken in Micrometer the default registry is global, and hence accessible even without `getMetrics()`. So at least for Micrometer I think it's safe to default to the global registry.

For MicroProfile, I'd suggest that we error out as you did already; and if/when `getMetrics()` is implemented for this framework, we could revisit the logic and create a SmallRye registry on the fly.

emerkle826: Good point about Micrometer. I'll change it as you suggest.

emerkle826:
> We should also revisit the manual now that subclassing the context is no longer necessary, and maybe move it to the regular manual (as opposed to developer docs).
> Also cover the fact that `Session.getMetrics()` only works with Dropwizard.

My last commit has moved the metrics docs from the developer page to the Metrics section in Core.

manual/core/metrics/README.md
```diff
@@ -52,10 +52,81 @@ If you specify a metric that doesn't exist, it will be ignored and a warning wil
...
 
-The Dropwizard `MetricRegistry` is exposed via `session.getMetrics()`. You can retrieve it and
-configure a `Reporter` to send the metrics to a monitoring tool.
+The Dropwizard `MetricRegistry` is exposed via `session.getMetrics().getRegistry()`. You can retrieve
+it and configure a `Reporter` to send the metrics to a monitoring tool.
+
+**NOTE:** At this time, `session.getMetrics()` is not available when using Micrometer or MicroProfile
+metrics. If you wish to use either of those metrics frameworks, it is recommended to provide a
+Registry implementation to the driver as described in the [Metric Registry section](#metric-registry),
+and follow best practices for exporting that registry to your desired reporting framework.
```
adutra: Very good docs.

adutra (approved):

olim7t:
I know this is from the other PR, and I don't think we discussed it at the time, but I wonder if throwing from `MicroProfileMetricsFactory.getMetrics()` is the right thing to do. This forces any existing caller to now catch the exception. If someone was using `Metrics`, it's because they were accessing the Dropwizard metrics programmatically. If those metrics are not available, it makes sense to return an empty optional. The fact that some other metrics are available via another path is not really relevant.

core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
olim7t: Sorry for not replying earlier, I agree with what you settled on.

> Implementing `getMetrics()` may be an interesting addition that stems quite naturally from the general trend of opening up to different metrics implementations.

As a side note, this is impossible with the current interface because it directly references DropWizard types. But as Alexandre noted, this interface has become less important now that we have the ability to inject an external registry. Its only real added value is that it saves you the hassle of constructing metric paths (what `buildFullName` does in the metric updater). We could always expose that as a standalone utility, all the factories do it the same way (edit: I created [JAVA-2872](https://datastax-oss.atlassian.net/browse/JAVA-2872) for that).

olim7t:
👍  This looks great.
I'm still not keen on the principle, but I agree that we've reached a good solution that is not too invasive. In particular I like how the injectable registry addresses the obsolescence of `Session.getMetrics()`.
I ended up switching to empty optionals instead of throwing exceptions per my previous comment. We still have time to revisit that if there are other opinions.

PR #1495, opened 2020-08-24
JAVA-2867 Fixed protocol substitutions

JiriOndrusek:
Fixes https://datastax-oss.atlassian.net/jira/software/c/projects/JAVA/issues/JAVA-2867

During investigation of camel-quarkus extension for cassandraQl, I was facing issue, which is already reported as https://datastax-oss.atlassian.net/jira/software/c/projects/JAVA/issues/JAVA-2867
This is simple fix of the issue. 

It would be nice to have some JUnit validating this quarkus stuff. But introducing this is beyond my knowledge of datastax /java-driver..


tomekl007:
@JiriOndrusek Thank You for your contribution! 
I did add unit tests that are asserting that such a situation will not occur in the future. 
Let's wait for a review of other team members.

JiriOndrusek:
@tomekl007 Thanks for the test!

core/src/test/java/com/datastax/oss/driver/internal/core/protocol/CompressSubstitutionsTest.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import org.junit.Test;
+
+public class CompressSubstitutionsTest {
+
+  @Test
+  public void Lz4SubstitutionShouldSubstituteAllProtectedMethodsFromLz4Compressor() {
```
adutra: Instead of this, wouldn't it be simpler to make `Lz4Substitution` and `SnappySubstitution` implement `Compressor<ByteBuf>`?

tomekl007: I was thinking about it but I assumed that not implementing `Compressor<ByteBuf>` was done on purpose and has some valid reason, @absurdfarce  can we do it without problems?

adutra: Hmm I think you meant @absurdfarce.

absurdfarce: This code came into the Java driver with JAVA-2803, and while I didn't see anything precisely on point in the record If I recall correctly the main issue here is that Lz4Substitution shouldn't be thought of as a Compressor itself.  It's just a substitution for certain methods that apply in certain circumstances, in this case when the lz4 libs aren't present.  There's no real cause for this class to implement the algorithm() method, for example, since whatever class it's used as a substitution for will already handle that.

Also worth noting: implementing the interface (arguably) provides a false security.  There's nothing stopping Lz4Compressor from implementing some new interface that's similar to (but not the same as) the hierarchy built up by Compressor<ByteBuf> while the substitution remains unchanged.  In that case everything would continue to type check but the substitution almost certainly wouldn't behave as expected.  That fact alone argues for automated testing of some kind to catch this.

Finally, I'd note that if we want to pursue this path (and I'm not fundamentally opposed to it) then Compressor<ByteBuf> seems like the wrong interface.  Note that SnappySubstitution and Lz4Substitution both have the same methods with the same signatures.... it seems like there's an interface begging to be born there.  I also note that the same interface could be applied to ByteBufCompressor to handle the declared abstract methods in that class as well.  Note, however, that pulling out such an interface would still be subject to the implementation  divergence mentioned above.

To sum up: I don't think of this class as a compressor itself but as a provider of replacement methods for a compressor.  We could readily pull out an interface to define those methods for this replacement (and the Snappy one as well) but without better testing we're still vulnerable to impls diverging.

tomekl007: So I suggest keeping the test that I created in this PR.

adutra: > There's no real cause for this class to implement the algorithm() method

But it doesn't hurt to do so, does it? It would simply return "lz4".

> There's nothing stopping Lz4Compressor from implementing some new interface

Granted, although highly unlikely. But I get your point: some methods in `ByteBufCompressor` are not present in `Compressor` so implementing `Compressor` is not enough. OK.

In this case can we make `Lz4Substitution` extend `ByteBufCompressor`? If that's possible I sill think that for the sake of simplicity that would be enough for now, given that these classes are not expected to change anymore... until protocol v6 :-).

absurdfarce: The methods in question are marked as abstract in ByteBufCompressor so I'd still argue it's worthwhile to pull those out into an interface which everyone should implement.  There's no reason for either of the substitutions to worry about the other implementation details in ByteBufCompressor; they just need to implement the methods in question.  And I'd argue it's a clearer expression of intent to leave them focused on just that.

To make this more concrete I pulled out a simple version of what this interface might look like and put a PR up against @JiriOndrusek 's fork.  I'd argue this PR (https://github.com/JiriOndrusek/java-driver/pull/1) is a cleaner expression of the intent we're after here.

adutra: @olim7t what would be your opinion? @absurdfarce's solution is elegant; but we need this ticket in 4.9 so maybe we could just merge the PR as is and leave this improvement for a follow-up PR.

olim7t: > There's nothing stopping Lz4Compressor from implementing some new interface

True, but that seems very unlikely. The rest of the driver interacts with `Compressor<ByteBuf>` as a completely opaque abstraction. If Lz4Compressor implements a new interface, it means that someone, somewhere would use that interface, so it would probably surface through the parent types as well.
This was the case for the ticket that caused this regression (JAVA-2773), the changes propagated all the way up to `Compressor`.

Also, even if we introduce a `ByteBufCompressOps`, the same thing can happen. If someone completely forgets about substitutions (😳 ), they can still modify the concrete classes in a way that doesn't break at compile time but breaks at substitution time.

I think extending `ByteBufCompressor` is a good middle ground. It doesn't cover every imaginable case, but it would have caught this bug, and most future problems will be of a similar nature given how the type hierarchy is designed. For the rest, we have the unit test.

tomekl007: @olim7t 
I did make Lz4Substitution and SnappySubstitution extend `ByteBufCompressor`. 
regarding:
>For the rest, we have the unit test.

which unit test do you mean? Are you suggesting that we should also keep the `CompressSubstitutionsTest` (https://github.com/datastax/java-driver/pull/1495/files#diff-f7188280f40e3b891f3337cb2e13ba5eR33)?

tomekl007: It turns out that the solution with extends ByteBufCompressor will not work with graal because of the:
```
com.oracle.svm.core.util.UserError$UserException: Annotated class must inherit directly from Object: class com.datastax.oss.driver.internal.core.protocol.Lz4Substitution
```
So we would need to extend Lz4Compressor directly, but this is not possible because the constructor of `Lz4Compressor` is using the lz4 dependency (and we want to make it work when this dependency is not present on the classpath):
```
 Lz4Compressor(String sessionName) {
    if (DependencyCheck.LZ4.isPresent()) {
    ....
```
(same for SnappyCompressor), and the constructor needs to be called from the child class:
```
@Substitute
public Lz4Substitution(DriverContext context) {
    super(context);
}
```
So one feasible solution is what @absurdfarce suggested: extracting a new `ByteBufCompressOps` interface. The other is leaving this code as it is. wdyt?

core/src/test/java/com/datastax/oss/driver/internal/core/protocol/CompressSubstitutionsTest.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import org.junit.Test;
+
+public class CompressSubstitutionsTest {
```
adutra, line 33: OK to keep the test then, but the current test is way too complex, and method names do not comply with our guidelines.

Suggested change:

```
@RunWith(DataProviderRunner.class)
public class CompressSubstitutionsTest {

  private static final Set<String> EXCLUDED_METHOD_NAMES =
      ImmutableSet.of("algorithm", "readUncompressedLength");

  private static final Comparator<Method> METHOD_SIGNATURE_COMPARATOR =
      Comparator.comparing(Method::getName)
          .thenComparing(Method::getReturnType, (ret1, ret2) -> ret1.equals(ret2) ? 0 : -1)
          .thenComparing(
              Method::getParameterTypes,
              (params1, params2) -> Arrays.deepEquals(params1, params2) ? 0 : -1);

  @Test
  @UseDataProvider(value = "substitutionClasses")
  public void should_substitute_compressor_methods(
      Class<?> substitutedClass, Class<?> substitutionClass) {

    Set<Method> methodsToSubstitute =
        Arrays.stream(substitutedClass.getDeclaredMethods())
            .filter(method -> !EXCLUDED_METHOD_NAMES.contains(method.getName()))
            .collect(Collectors.toSet());

    Set<Method> substitutedMethods =
        Arrays.stream(substitutionClass.getDeclaredMethods()).collect(Collectors.toSet());

    assertThat(methodsToSubstitute)
        .usingElementComparator(METHOD_SIGNATURE_COMPARATOR)
        .containsExactlyInAnyOrderElementsOf(substitutedMethods);
  }

  @DataProvider
  public static Object[][] substitutionClasses() {
    return new Object[][] {
      {Lz4Compressor.class, Lz4Substitution.class},
      {SnappyCompressor.class, SnappySubstitution.class}
    };
  }
}
```

adutra (changes requested):

core/src/test/java/com/datastax/oss/driver/internal/core/protocol/CompressSubstitutionsTest.java
tomekl007: done, thx

adutra (approved):

core/src/test/java/com/datastax/oss/driver/internal/core/protocol/CompressSubstitutionsTest.java
adutra: We are running out of time, I suggest that we do not introduce any interface for now, and keep the test.

olim7t:
Sorry to open yet another PR, but I'd like to propose an alternative solution: #1497 

Instead of substituting the classes' contents, take a step back: delete the classes, and substitute the code that was trying to instantiate them. This completely eliminates the need to track changes to the implementations.

olim7t:
Closing in favor of #1497, we've decided to go with the other approach.
Thanks @JiriOndrusek for getting things started.

PR #1496, opened 2020-08-25
JAVA-2870: Optimize memory usage of token map

olim7t:


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/NetworkTopologyReplicationStrategy.java
```diff
@@ -56,21 +56,31 @@
...
+    // The implementation of this method was adapted from
+    // org.apache.cassandra.locator.NetworkTopologyStrategy
+
+    ImmutableMap.Builder<Token, Set<Node>> result = ImmutableMap.builder();
     Map<String, Set<String>> racks = getRacksInDcs(tokenToPrimary.values());
     Map<String, Integer> dcNodeCount = Maps.newHashMapWithExpectedSize(replicationFactors.size());
     Set<String> warnedDcs = Sets.newHashSetWithExpectedSize(replicationFactors.size());
+    // To optimize memory usage, ensure that identical replica sets will be represented by a single
+    // instance. Order matters so the key must be a list.
+    Map<List<Node>, Set<Node>> canonicalReplicaSets = new HashMap<>();
```
olim7t: Contrary to what I said on the ticket, we need a list as the key here. Otherwise sets in different order would get the same canonical representation (LinkedHashSet's hashCode and equals are not order-sensitive).

So I either had to convert `replicas` into a list when I do the lookup, or make it a list to begin with. I chose the latter to avoid yet another copy. This forces me to use the linear `addIfAbsent`, but the impact will be minimal since the cardinalities are low.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/NetworkTopologyReplicationStrategy.java
```diff
@@ -56,21 +56,31 @@
...
+
     // find maximum number of nodes in each DC
     for (Node node : Sets.newHashSet(tokenToPrimary.values())) {
       String dc = node.getDatacenter();
       dcNodeCount.putIfAbsent(dc, 0);
```
adutra: Unrelated, but it's probably more efficient to do:

```
      dcNodeCount.merge(dc, 1, Integer::sum);
```

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/NetworkTopologyReplicationStrategy.java
```diff
@@ -80,8 +90,6 @@
         skippedDcEndpoints.put(dc, new LinkedHashSet<>()); // preserve order
       }
 
-      // Preserve order - primary replica will be first
-      Set<Node> replicas = new LinkedHashSet<>();
       for (int j = 0; j < ring.size() && !allDone(allDcReplicas, dcNodeCount); j++) {
         Node h = tokenToPrimary.get(getTokenWrapping(i + j, ring));
         String dc = h.getDatacenter();
```
adutra, line 88: Again unrelated but it seems that on line 98, we can avoid the integer boxing:

```
        int rf = replicationFactors.get(dc).fullReplicas();
        Set<Node> dcReplicas = allDcReplicas.get(dc);
        if (dcReplicas.size() >= rf) {
          continue;
        }
```

olim7t:
I checked heap usage on a test cluster: 6 nodes in 2 DCs, vnodes = 256, one keyspace with `{'class': 'NetworkTopologyStrategy', 'dc1':3, 'dc2':3}`.

Before the fix, the `KeyspaceTokenMap` instance retains 820,288 bytes. After, it drops down to 565,324. That's a 30% reduction, not a dramatic improvement but still pretty significant.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/NetworkTopologyReplicationStrategy.java
```diff
@@ -56,21 +56,31 @@
...
+    // instance. Order matters so the key must be a list.
+    Map<List<Node>, Set<Node>> canonicalReplicaSets = new HashMap<>();
+
     // find maximum number of nodes in each DC
     for (Node node : Sets.newHashSet(tokenToPrimary.values())) {
       String dc = node.getDatacenter();
       dcNodeCount.putIfAbsent(dc, 0);
       dcNodeCount.put(dc, dcNodeCount.get(dc) + 1);
     }
+    // Preserve order - primary replica will be first
```
adutra: Is this list sorted somehow? If so, couldn't we implement `addIfAbsent` using `Collections.binarySearch()`?

adutra: Also if the operations performed on this list are `add(Node)` and `contains(Node)` mostly (+ `clear()`), maybe a `LinkedList` would perform better since add is O(1).

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationStrategy.java
```diff
@@ -17,11 +17,10 @@
...
 import com.datastax.oss.driver.api.core.metadata.token.Token;
-import com.datastax.oss.driver.shaded.guava.common.collect.SetMultimap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 public interface ReplicationStrategy {
-  SetMultimap<Token, Node> computeReplicasByToken(
-      Map<Token, Node> tokenToPrimary, List<Token> ring);
+  Map<Token, Set<Node>> computeReplicasByToken(Map<Token, Node> tokenToPrimary, List<Token> ring);
```
adutra, line 25: Incidentally, I find the new return type more expressive than the old one.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/NetworkTopologyReplicationStrategy.java
```diff
@@ -139,11 +147,20 @@
...
 
-      result.putAll(ring.get(i), replicas);
+      Set<Node> canonicalReplicaSet =
+          canonicalReplicaSets.computeIfAbsent(replicas, ImmutableSet::copyOf);
+      result.put(ring.get(i), canonicalReplicaSet);
     }
     return result.build();
   }
 
+  private static void addIfAbsent(List<Node> nodes, Node node) {
```
adutra: There is a small `CollectionsUtils` class, maybe you could move this utility method there.

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationStrategy.java
adutra: By the way, in `KeyspaceTokenMap`: shouldn't we also migrate `tokenRangesByNode` to `Map<Node, Set<TokenRange>>`? I'm not a big fan of multimaps.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/EverywhereReplicationStrategy.java
```diff
@@ -17,23 +17,23 @@
...
 public class EverywhereReplicationStrategy implements ReplicationStrategy {
 
   @Override
-  public SetMultimap<Token, Node> computeReplicasByToken(
+  public Map<Token, Set<Node>> computeReplicasByToken(
       Map<Token, Node> tokenToPrimary, List<Token> ring) {
-    ImmutableSetMultimap.Builder<Token, Node> result = ImmutableSetMultimap.builder();
-    Collection<Node> nodes = tokenToPrimary.values();
+    ImmutableMap.Builder<Token, Set<Node>> result = ImmutableMap.builder();
+    Set<Node> allNodes = ImmutableSet.copyOf(tokenToPrimary.values());
```
tomekl007, line 34: I don't know if this is feasible but maybe we could go one level higher and make `Map<Token, Node> tokenToPrimary#values` immutable as well? That way we would not need to do a copy here

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/NetworkTopologyReplicationStrategy.java
```diff
@@ -139,11 +147,20 @@
         }
       }
 
-      result.putAll(ring.get(i), replicas);
+      Set<Node> canonicalReplicaSet =
+          canonicalReplicaSets.computeIfAbsent(replicas, ImmutableSet::copyOf);
+      result.put(ring.get(i), canonicalReplicaSet);
```
tomekl007: Is it thread-safe to split this into two operations? (I guess yes, but wanted to double-check)

adutra: This code does not need to be thread-safe.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/EverywhereReplicationStrategy.java
adutra: But the `Map.values()` method returns `Collection`; here we need `Set`.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/NetworkTopologyReplicationStrategy.java
olim7t: No, the list is not sorted, it's just the primary first and then the replicas in the order they were found while walking the ring.
I don't think `LinkedList` will perform better, it [almost never does](https://errorprone.info/bugpattern/JdkObsolete), and since we're now reusing the array list, we don't even have to pay the allocation cost.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/EverywhereReplicationStrategy.java
olim7t: `tokenToPrimary` is initially built in `DefaultTokenMap.buildTokenToPrimaryAndRing`. It is an immutable map, so its `values()` will be as well.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationStrategy.java
olim7t: A multimap is more convenient in `KeyspaceTokenMap.buildTokenRangesByNode`: unlike in replication strategies, we don't build the sets one at a time. It's essentially an inversion of `replicasByToken`, so as we iterate each token we dispatch it to the "in progress" set of each replica. We would have to use a temporary map of set builders, and then build all the builders in a second pass at the end. Then we could try to also optimize duplicates, although I'm not sure it's common (or even possible?) that two nodes replicate the exact same set of tokens.

But another question is whether we should precompute that information at all. `replicasByToken` is where we really want speed, because it's what the LBP calls for token-aware routing. But `tokenRangesByNode` is only used when the client application calls `TokenMap.getTokenRanges(CqlIdentifier, Node)`. We could trade space for speed and recompute the result on every call, it "only" costs a full traversal of `replicasByToken`, which is O(ring_size * RF). Or maybe initialize `tokenRangesByNode` lazily...

Maybe we can leave that as a follow-up ticket because I need to refresh my memory on why we added this method and how it's typically used in the field. There's some background on [JAVA-312](https://datastax-oss.atlassian.net/browse/JAVA-312) on that.

olim7t:
Another test on a 6-node DSE 6.8.3 cluster, with just the system keyspaces:
- before: `DefaultTokenMap` retains 2,280,200 bytes
- after: 1,281,492

`EverywhereStrategy` (used by `dse_system`, `solr_admin` and `dse_insights`) is the worst contributor. [JAVA-2871](https://datastax-oss.atlassian.net/browse/JAVA-2871) should really help, it's to late to include it in 4.9.0 but I'll add a reminder about keyspace filtering in the release notes.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/ReplicationStrategy.java
olim7t: I created [JAVA-2874](https://datastax-oss.atlassian.net/browse/JAVA-2874).

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/NetworkTopologyReplicationStrategy.java
olim7t: I ended up extracting a `CanonicalNodeSetBuilder` that is reused across the two strategies, it's a nice way to avoid duplication.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/NetworkTopologyReplicationStrategy.java
olim7t: Actually I think that's a regression introduced by [JAVA-2105](https://datastax-oss.atlassian.net/browse/JAVA-2105). The map entry can be absent if the keyspace's replication configuration doesn't mention the DC. We used to skip the DC, but now it's going to NPE on `replicationFactors.get(dc).fullReplicas()`. Will fix.

olim7t: Ah nevermind, that's caught indirectly by the `!allDcReplicas.containsKey(dc)` test two lines above.

PR #1497, opened 2020-08-28
JAVA-2867: Revisit compressor substitutions

olim7t:
Alternative proposal to #1495 

TODO:
- [x] rename `get()` to `newInstance()`
- [x] add javadocs to BuiltInCompressors to link to CompressorSubstitutions


core/src/main/java/com/datastax/oss/driver/internal/core/protocol/BuiltInCompressors.java
```diff
@@ -0,0 +1,40 @@
...
+package com.datastax.oss.driver.internal.core.protocol;
+
+import com.datastax.oss.driver.api.core.config.DefaultDriverOption;
+import com.datastax.oss.driver.api.core.context.DriverContext;
+import com.datastax.oss.protocol.internal.Compressor;
+import io.netty.buffer.ByteBuf;
+
+public class BuiltInCompressors {
+
+  public static Compressor<ByteBuf> get(String name, DriverContext context) {
```
olim7t: The whole thing relies on the fact that this will be the unique entry point to instantiate a compressor. Maybe we could make the constructors package-private to emphasize that.

I think that's reasonable, compressors are internal anyway and clients have no reason to instantiate them directly. And if someone wants to plug a custom compressor, they'd have to extend the context anyway, so we can hope that they see the substitution stuff and adapt it appropriately.

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/CompressorSubstitutions.java
```diff
@@ -0,0 +1,108 @@
...
+      }
+    }
+  }
+
+  public static class Lz4Missing extends Lz4Present {
+    @Override
+    public boolean getAsBoolean() {
+      return !super.getAsBoolean();
+    }
+  }
```
olim7t, line 108: There's no opposite to `onlyWith` 😿

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/BuiltInCompressors.java
absurdfarce: Agreed, I don't see this as a major concern.  As mentioned somebody would have to have a custom DriverContext to even exercise the problem here and in that case it doesn't seem super-surprising that your results might vary if you diverge from the patterns in place.

_Perhaps_ it's worth a comment here to say something to the effect of "BuiltInCompressors exists to allow for easy substitutions on GraalVM in cases where underlying dependent libraries aren't present".  Or, you know, something saying the same thing but saying it better. :feelsbadman:  I don't think it's essential but it might at least help a little bit.

I don't really love the idea of making compressor constructors package-private... that feels more disconnected from the problem we're actually trying to solve here.

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/BuiltInCompressors.java
```diff
@@ -0,0 +1,40 @@
...
+    switch (name.toLowerCase()) {
+      case "lz4":
+        return new Lz4Compressor(context);
```
absurdfarce, line 34: Worth noting: this is immediately going to execute a runtime check of the classpath to validate that the exact same class we just checked for is present.  In a Graal context that check is _completely_ redundant.  Perhaps we should either expose some way to opt out of that check or maybe make the check contingent upon being in a non-Graal context within Lz4Compressor.java?

olim7t: Right. On the other hand it's just a `Class.forName` call, is it that significant at init time?

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/CompressorSubstitutions.java
```diff
@@ -0,0 +1,108 @@
...
+  public static final class DeleteSnappyCompressor {}
+
+  public static class Lz4Present implements BooleanSupplier {
+
+    private static final String LZ4_CLZ_NAME = "net.jpountz.lz4.LZ4Compressor";
+
+    @Override
+    public boolean getAsBoolean() {
+      try {
+        Class.forName(LZ4_CLZ_NAME);
```
adutra, line 95: You could use `DependencyCheck.LZ4.isPresent()` for that, but maybe it's not a good idea to introduce dependencies in a GraalVM substitution code.

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/BuiltInCompressors.java
adutra: Maybe the check done in `DependencyCheck` should be memoized.

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/BuiltInCompressors.java
tomekl007: Isn't it executed only one time during the compilation phase?

tomekl007 (approved):
I've tested this version with the cassandra-quarkus native tests and it looks good. 
I like your version better because we don't need to keep substitutions in sync.

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/BuiltInCompressors.java
absurdfarce: It'll get executed once at driver startup (possibly in native code depending on how the image was built).  To be clear it's a minor thing; it would be nice to not have to execute the same check twice but this is hardly a show-stopper.

absurdfarce (approved):
This solution is the cleanest I've seen so far.  My comments are all fairly minor; there's nothing in there that's anywhere close to a legit reason to hold up moving forward with this.

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/CompressorSubstitutions.java
absurdfarce: I'm not sure it's really that big a problem, though.  If I have this right you wind up with:

```java
  public static class Lz4Present implements BooleanSupplier {

    @Override
    public boolean getAsBoolean() {
      return DependencyCheck.LZ4.isPresent();
    }
  }
```

This seems like a pretty minimal dependency... and probably worth the risk to avoid re-implementing the same thing.

olim7t: Yes, I'd rather keep the substitution code as simple as possible 👍

PR #1498, opened 2020-08-29
JAVA-2827 + JAVA-2859: Upgrade Tinkerpop and remove its unused dependencies

adutra:


core/src/main/java/com/datastax/dse/driver/internal/core/graph/binary/AbstractSimpleGraphBinaryCustomSerializer.java
```diff
@@ -105,11 +103,11 @@ public T readValue(Buffer buffer, GraphBinaryReader context, boolean nullable)
     final int valueLength = buffer.readInt();
 
     if (valueLength <= 0) {
-      throw new SerializationException(String.format("Unexpected value length: %d", valueLength));
+      throw new IOException(String.format("Unexpected value length: %d", valueLength));
```
adutra, line 106: I think this is safe. I explored the call hierarchy of these methods, and no caller catches `SerializationException` nor applies any custom processing to instances of `SerializationException`. Actually the whole Tinkerpop project is moving away from `SerializationException`  it seems: I could only find 2 `catch (SerializationException e)` blocks in the whole codebase. See also [TINKERPOP-2179](https://issues.apache.org/jira/browse/TINKERPOP-2179) for context.

pom.xml
```diff
@@ -155,23 +155,30 @@
         <groupId>org.apache.tinkerpop</groupId>
         <artifactId>gremlin-core</artifactId>
         <version>${tinkerpop.version}</version>
+        <exclusions>
```
adutra, line 158: We cannot exclude javatuples because it is used to serde `Pair` instances. We cannot exclude javapoet either because there is one annotation processor declared in gremlin-core (`GremlinDslProcessor`) that would throw an error and make the project compilation fail if javapoet is absent.

olim7t: I created [TINKERPOP-2411](https://issues.apache.org/jira/browse/TINKERPOP-2411) to suggest moving the Gremlin processor to its own artifact.

olim7t (approved):

PR #1500, opened 2020-09-02
JAVA-2871: Allow keyspace exclusions in the metadata, and exclude system keyspaces by default

olim7t:
The [manual](https://github.com/datastax/java-driver/tree/java2871/manual/core/metadata/schema#filtering) has the best description of the rules. I've left out the weird corner cases, but they are covered by the unit tests.

TODO:
- [x] update manual

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/KeyspaceFilter.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/KeyspaceFilter.java
@@ -1,0 +1,185 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.internal.core.metadata.schema.queries;
+
+import com.datastax.oss.driver.api.core.config.DefaultDriverOption;
+import com.datastax.oss.driver.shaded.guava.common.annotations.VisibleForTesting;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.function.Predicate;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Filters keyspaces during schema metadata queries.
+ *
+ * <p>Depending on the circumstances, we do it either on the server side with a WHERE IN clause that
+ * will be appended to every query, or on the client side with a predicate that will be applied to
+ * every fetched row.
+ */
+public class KeyspaceFilter {
+
+  private static final Logger LOG = LoggerFactory.getLogger(KeyspaceFilter.class);
+
+  private static final Pattern EXACT_INCLUDE = Pattern.compile("\\w+");
+  private static final Pattern EXACT_EXCLUDE = Pattern.compile("!(\\w+)");
+  private static final Pattern REGEX_INCLUDE = Pattern.compile("/(.+)/");
+  private static final Pattern REGEX_EXCLUDE = Pattern.compile("!/(.+)/");
+
+  private final String logPrefix;
+  private final String whereClause;
+  private final Set<String> exactIncludes = new HashSet<>();
+  private final Set<String> exactExcludes = new HashSet<>();
+  private final List<Predicate<String>> regexIncludes = new ArrayList<>();
+  private final List<Predicate<String>> regexExcludes = new ArrayList<>();
+  private final boolean hasNoIncludes;
+
+  private final boolean isDebugEnabled;
+  private final Set<String> loggedKeyspaces;
+
+  public KeyspaceFilter(@NonNull String logPrefix, @NonNull List<String> specs) {
+    this.logPrefix = logPrefix;
+    for (String spec : specs) {
+      spec = spec.trim();
+      Matcher matcher;
+      if (EXACT_INCLUDE.matcher(spec).matches()) {
+        exactIncludes.add(spec);
+      } else if ((matcher = EXACT_EXCLUDE.matcher(spec)).matches()) {
+        exactExcludes.add(matcher.group(1));
+      } else if ((matcher = REGEX_INCLUDE.matcher(spec)).matches()) {
+        compile(matcher.group(1)).map(regexIncludes::add);
+      } else if ((matcher = REGEX_EXCLUDE.matcher(spec)).matches()) {
+        compile(matcher.group(1)).map(regexExcludes::add);
+      } else {
+        LOG.warn(
+            "[{}] Error while parsing {}: invalid element '{}', skipping",
+            logPrefix,
+            DefaultDriverOption.METADATA_SCHEMA_REFRESHED_KEYSPACES.getPath(),
+            spec);
+      }
+    }
+
+    if (!exactIncludes.isEmpty() && regexIncludes.isEmpty()) {
+      // We can filter on the server
+      whereClause = buildWhereClause(exactIncludes);
+      if (!exactExcludes.isEmpty() || !regexExcludes.isEmpty()) {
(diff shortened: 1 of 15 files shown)


manual/core/metadata/schema/README.md
```diff
@@ -132,7 +132,53 @@ You can also limit the metadata to a subset of keyspaces:
 datastax-java-driver.advanced.metadata.schema.refreshed-keyspaces = [ "users", "products" ]
 ```
 
-If the property is absent or the list is empty, it is interpreted as "all keyspaces".
+Each element in the list can be one of the following:
+
+1. An exact name inclusion, for example `"Ks1"`. If the name is case-sensitive, it must appear in
+   its exact case.
+2. An exact name exclusion, for example `"!Ks1"`.
+3. A regex inclusion, enclosed in slashes, for example `"/^Ks.*/"`. The part between the slashes
```
adutra, line 140: From your example it seems that the regex must match the entire input, not portions of it. Maybe it's worth mentioning this here.

manual/core/metadata/schema/README.md
```diff
@@ -132,7 +132,53 @@ You can also limit the metadata to a subset of keyspaces:
...
+* Otherwise, if it matches an exact name exclusion, it is always excluded, regardless of what any
+  regex rule says.
+* Otherwise, if there are regex rules:
+
+  * if they're only inclusions, the keyspace must match at least one of them.
+  * if they're only exclusions, the keyspace must match none of them.
+  * if they're both, the keyspace must match at least one inclusion and none of the
+    exclusions.
+
+For example, given the keyspaces `system`, `ks1`, `ks2`, `data1` and `data2`, here's the outcome of
```
adutra, line 158: The inclusion/exclusion rules are very powerful, yet slightly convoluted. Do we absolutely need to distinguish "exact inclusions" from "regex inclusions"? After all, an exact keyspace name is just a special case of regex that has no wildcards. If we could consider that all the elements are regexes, we could probably simplify the rules and also avoid the usage of the special syntax `/.../` for regexes.

manual/core/metadata/schema/README.md
```diff
@@ -132,7 +132,53 @@ You can also limit the metadata to a subset of keyspaces:
...
+
+
+If an element is malformed, or if its regex has a syntax error, a warning is logged and that single
+element is ignored.
+
+The default configuration (see [reference.conf](../../configuration/reference/)) excludes all
+Cassandra and DSE system keyspaces.
+
+Try to use only exact name inclusions if possible. This allows the driver to filter on the server
+side with a `WHERE IN` clause. If you use any other rule, it has to fetch all system rows and filter
```
adutra, line 181: Oh I see. But still, maybe you can detect if an element is a regex or not, given that 1) keyspace names can only contain alphanumeric characters and underscores and 2) regexes must match the entire keyspace name (iow, `ks1` cannot match `ks12`).

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/KeyspaceFilter.java
```diff
@@ -0,0 +1,65 @@
...
+    if (specs.isEmpty()) {
+      return INCLUDE_ALL;
+    } else {
+      return new RuleBasedKeyspaceFilter(logPrefix, specs);
+    }
+  }
+
+  @VisibleForTesting
+  static KeyspaceFilter newInstance(@NonNull String... specs) {
+    return newInstance("test", Arrays.asList(specs));
```
adutra: Shouldn't we make this a parameter to avoid hard-coding `"test"` here?

core/src/main/resources/reference.conf
```diff
@@ -1785,14 +1785,41 @@ datastax-java-driver {
...
+      # server side with a WHERE IN clause. If you use any other rule, it has to fetch all system
+      # rows and filter on the client side.
+      #
+      # Required: no. The default value excludes all Cassandra and DSE system keyspaces. If the
+      #   option is unset, this is interpreted as "include all keyspaces".
       # Modifiable at runtime: yes, the new value will be used for refreshes issued after the
       #   change.
       # Overridable in a profile: no
-      // refreshed-keyspaces = [ "ks1", "ks2" ]
+      refreshed-keyspaces = [ "!system", "!/^system_.*/", "!/^dse_.*/", "!solr_admin" ]
```
adutra, line 1822: So by default we are not using server-side filtering. I wonder how many users will care about changing the defaults to benefit from server-side filtering? Which brings the question: will server-side filtering be used at all?

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/KeyspaceFilterTest.java
```diff
@@ -0,0 +1,128 @@
...
+    assertThat(apply(filter, KEYSPACES)).isEqualTo(KEYSPACES);
+  }
+
+  @Test
+  public void should_filter_on_server_when_only_exact_rules() {
+    KeyspaceFilter filter =
+        KeyspaceFilter.newInstance("inventory_test", "customers_test", "!system");
+    // Note that exact excludes are redundant in this case: either they match an include and will be
+    // ignored, or they don't and the keyspace is already ignored.
+    // We let it slides, but a warning is logged.
```
adutra:
```suggestion
    // We let it slide, but a warning is logged.
```

manual/core/metadata/schema/README.md
olim7t: Actually no, `Pattern.matcher().matches()` also returns true for partial matches. I included the start symbol `^` because otherwise it will match stuff like `aaaKs`.

So even more justification for documenting it 👍

olim7t: > also returns true for partial matches

Hrmm, in fact no, it doesn't.

manual/core/metadata/schema/README.md
olim7t: I agree that the syntax is a bit convoluted, maybe there is room for improvement there. For example we don't really need the trailing slash.

But from a semantic point of view, exact matches have a different meaning:

* `[ "!/system_.*/", "system.traces" ]`: I think most people would expect this to mean "exclude all system keyspaces, but include `system_traces`.
* on the other hand, if we have two regexes: `[ "/ks.*/", "!/.*2/" ]`. Then it's fair to assume an intersection: all keyspaces that start with `ks`, but do not end with `2`.

If I apply the "two regexes" rule to the first example, I get "all keyspaces named `system_traces`, but that do not start with `system_`". That's the empty set.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/KeyspaceFilter.java
olim7t: Maybe. But hardcoding "test" for tests seems reasonable.

core/src/main/resources/reference.conf
olim7t: > So by default we are not using server-side filtering

Correct. This is still an improvement, 4.9 fetches and parses everything, 4.10 will fetch everything but only parse a subset.

I think we can still expect that filtering to one or two specific keyspaces will be a common occurrence. Adding a WHERE clause is easy and brings a good improvement.

manual/core/metadata/schema/README.md
adutra: I'm still not 100% convinced. If I wanted to "exclude all system keyspaces, but include `system_traces`" I could simply specify `[ "system_traces" ]` right? For your first example `[ "!/system_.*/", "system.traces" ]`, returning an empty set seem legit to me.

olim7t: One thing that is maybe not obvious is that `[ "!/system_.*/", "system.traces" ]` also implies "include everything that doesn't start with system_" (per the third regex rule: if they're only exclusions, the keyspace must match none of them).

If you simply specify `[ "system_traces" ]` you only get system_traces. And this one can't really change because we need to preserve backward compatibility with the 4.9 behavior.

Returning the empty set for `[ "!/system_.*/", "system.traces" ]` seems very wrong to me. If the user mentioned a specific keyspace by name, it's almost certainly because they wanted to include it. That's why exact names win over any other rule.

I think it will take too much time to discuss this on individual examples: if you have a better idea, please suggest a complete set of rules and I'll try to poke holes in it 😉 
I already went over several versions on my own, it will probably be one of them.

Or maybe patterns are overkill. We could simply hard-code the system exclusions when nothing is specified, and keep the current behavior (exact inclusions only) otherwise.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/schema/queries/KeyspaceFilter.java
adutra: I inlined and removed the method then.

manual/core/metadata/schema/README.md
adutra: Added a small note.

manual/core/metadata/schema/README.md
adutra: Since we're running out of time I'm leaving the rules as they are for now. This is not strictly part of the API so I guess we could revisit the rules later on if necessary.

PR #1501, opened 2020-09-03
JAVA-2647: Handle token types in QueryBuilder.literal()

olim7t:


query-builder/src/main/java/com/datastax/oss/driver/api/querybuilder/QueryBuilder.java
```diff
@@ -402,6 +406,15 @@ public static Literal literal(@Nullable Object value) {
    */
   @NonNull
   public static Literal literal(@Nullable Object value, @NonNull CodecRegistry codecRegistry) {
+    if (value instanceof Murmur3Token) {
+      value = ((Murmur3Token) value).getValue();
```
adutra, line 410: Nit: In theory you don't need to call `codecRegistry.codecFor(value)` for token values since you know the codecs to use already.

adutra (approved):

PR #1504, opened 2020-09-18
Add ability to load config directly from typesafe config object

mishamo:


adutra:
Unfortunately, this leaks a TypeSafe type through the driver API, which is forbidden:

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.1.1:javadoc (check-api-leaks) on project java-driver-core: An error has occurred in Javadoc report generation:
[ERROR] Method com.datastax.oss.driver.api.core.config.DriverConfigLoader.fromConfig leaks com.typesafe.config.Config (as parameter 'config')
```

What you are trying to achieve is already doable using the driver's internal API:

```
Config config = ...;
DriverConfigLoader loader = new DefaultDriverConfigLoader(() -> config);
CqlSession session = CqlSession.builder().withConfigLoader(loader).build();
```

mishamo:
Unfortunately the `...` part in your example has to include fallback to the cassandra reference conf. This happens naturally if you're using an `application.conf` and simply using `ConfigFactory.load()`, but if you're loading config from other sources and have a resolved `Config` object, this requires something like the following in the client codebase to work:

```
      config
        .withFallback(ConfigFactory.defaultReference(classOf[CqlSession].getClassLoader))
        .resolve()
        .getConfig("datastax-java-driver")
```

That aside, the API already leaks typesafe config in the example you've provided, just not in that class (DefaultDriverConfigLoader already takes a `Supplier<Config> configSupplier` as a parameter). Maybe it would be more acceptable to put this method into `DefaultDriverConfigLoader` instead of in `DriverConfigLoader`, that would satisfy the exposure requirement, but arguably it would make the method more difficult to discover for users.

adutra:
> Unfortunately the `...` part in your example has to include fallback to the cassandra reference conf.

Yes, to get all the driver defaults. But is that a problem? With your PR, you would have to include that fallback too, wouldn't you?

(If you are looking for ways to get rid completely of TypeSafe Config, we do offer a new API: [OptionsMap](https://docs.datastax.com/en/drivers/java/4.9/com/datastax/oss/driver/api/core/config/OptionsMap.html). Unfortunately, I just discovered that we have a subtle bug to solve: [JAVA-2889](https://datastax-oss.atlassian.net/browse/JAVA-2889) – but you can still try it out).
 
> That aside, the API already leaks typesafe config in the example you've provided, just not in that class (DefaultDriverConfigLoader already takes a `Supplier<Config> configSupplier` as a parameter). 

The driver distinguishes two main packages: `api` and `internal`. Everything in `api` packages is considered public API and must abide by strict backwards compatibility rules; everything in `internal` is considered internal and subject to change without further notice (even if we try to remain backwards compatible if possible). The class `DefaultDriverConfigLoader` is in an internal package, and thus is allowed to reference TypeSafe. 

So far we consider TypeSafe Config as an implementation detail, so it can only appear in internal packages. 

Hope that helps.

mishamo:
With the PR, that fallback happens internally in the driver; without it, the client has to copy code from inside the driver into their client in order to achieve the fallback to the cassandra `reference.conf` themselves.

I appreciate the need for rules based on package and for backward compatibility. As such, would we be happy if I moved the method in this PR into the internal package instead?

EDIT: I've updated the PR with the above change.

adutra:
> With the PR, that fallback happens internally in the driver; without it, the client has to copy code from inside the driver into their client in order to achieve the fallback to the cassandra reference.conf themselves.

I am a bit uncomfortable with this for a few reasons:

1. If you pass an already-constructed `Config` object, how can you know if this object contains the entire config, or just a few portions of it? How can you determine the right way to construct the final config if you don't know how this object was obtained? This makes the API contract for the new `fromConfig` method a bit unclear.
2. If you pass an already-constructed `Config` object, maybe your application is using TypeSafe Config to configure other parts of the system. In this case, it's not impossible that other reference.conf files could be present – but then, loading the reference config using `ConfigFactory.defaultReference(classOf[CqlSession].getClassLoader)` is almost certainly wrong, since the driver's class loader may not be able to locate other reference.conf files. You probably need instead: `ConfigFactory.defaultReference()`.

For these reasons, I'm inclined to think that if a user has already constructed half of their TypeSafe config outside of the driver, then it's their responsibility to build the entire config themselves, then pass it to the driver:

```
Config config = myCustomConfig().withFallback(ConfigFactory.defaultReference()).resolve()
DriverConfigLoader loader = new DefaultDriverConfigLoader(() -> config.getConfig(DEFAULT_ROOT_PATH), false)
CqlSession session = CqlSession.builder().withConfigLoader(loader).build();
```

Would the above work for you?

mishamo:
I think the first point applies to any method you use to pass the config to the config loader; whether it's a file, a string or a typesafe object. With the second point, I believe it's the application's responsibility to pass a relevant typesafe object to the loader, and one that would not break their application when the cassandra reference.conf if picked up. Again, that is the same responsibility the application has when passing a string or a file.

I suppose the point of contention is the value of exposing a convenience method for typesafe config integration vs exposing the `DEFAULT_ROOT_PATH` of the cassandra config root as per above. The suggestion does work for us, but I feel it's less elegant than the driver "handling" the typesafe integration for you. However, if it's definitely a no-go, then we can close this PR.

adutra:
I'm sorry for being on the fence, really. But my gut feeling is that this new method will trick users into all forms of misuse.

For instance, the example in the javadocs you wrote suggest that users should create their Config using `ConfigFactory.load("my-custom-config.conf")`, then pass that to the method.

But `ConfigFactory.load(String)` is already loading the default reference and the default overrides behind the scenes.

So if you do that, the driver would end up reading a config that is equivalent to the following one:

```
    .withFallback(
            .withFallback(ConfigFactory.parseResourcesAnySyntax("my-custom-config.conf"))
            .withFallback(ConfigFactory.defaultReferenceUnresolved())
            .resolve())
    .withFallback(ConfigFactory.defaultReference(CqlSession.class.getClassLoader()))
    .resolve()
    .getConfig(DEFAULT_ROOT_PATH);
```

I guess this wasn't exactly what you hoped to get. Especially the double resolution is concerning, and the mixing of different class loaders.

In fact, the right method to use to load your custom config is probably `ConfigFactory.parseResourcesAnySyntax("my-custom-config.conf")`. But I bet users won't think spontaneously of that one.

For these reasons and the ones we discussed above, if you don't mind, let's close this PR. But again, thank you very much for your effort here, I hope this won't discourage you from contributing again!

PR #1505, opened 2020-10-03
JAVA-2887: Handle composite profiles with more than one key and/or backed by only one profile

adutra:
Composite profiles with more than one key currently throw `ClassCastException`
from `CompositeDriverExecutionProfile.entrySet`.

Composite profiles backed by only one profile currently throw `NullPointerException`
from `CompositeDriverExecutionProfile.entrySet`.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -4,6 +4,7 @@
 
 ### 4.10.0 (in progress)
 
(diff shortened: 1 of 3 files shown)


adutra:
Sorry I had to merge with no reviews because this is becoming a blocker.

PR #1525, opened 2021-01-15
Add availability to pass NettyOptions via ProgrammaticArguments

sokomishalov:
Docs are saying that if you need to customize netty options you should do smth like [that](https://docs.datastax.com/en/developer/java-driver/4.9/manual/developer/common/context/).
Unfortunately, it is impossible to pass these options in case if you have some existing `CqlSessionBuilder` which is already enriched with some logic and data you don't want to lose.
So I think the only way for now.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/session/ProgrammaticArguments.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/session/ProgrammaticArguments.java
@@ -23,6 +23,7 @@ import com.datastax.oss.driver.api.core.ssl.SslEngineFactory;
 import com.datastax.oss.driver.api.core.tracker.RequestTracker;
 import com.datastax.oss.driver.api.core.type.codec.TypeCodec;
 import com.datastax.oss.driver.api.core.type.codec.registry.MutableCodecRegistry;
+import com.datastax.oss.driver.internal.core.context.NettyOptions;
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
 import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableMap;
 import edu.umd.cs.findbugs.annotations.NonNull;
@@ -61,6 +62,7 @@ public class ProgrammaticArguments {
   private final String startupApplicationVersion;
   private final MutableCodecRegistry codecRegistry;
   private final Object metricRegistry;
+  private final NettyOptions nettyOptions;
 
   private ProgrammaticArguments(
       @NonNull List<TypeCodec<?>> typeCodecs,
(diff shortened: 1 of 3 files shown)


adutra:
Hi, thanks for your contribution!

Unfortunately, this change makes the build fail:

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  31.037 s
[INFO] Finished at: 2021-01-21T15:46:30+01:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.2.0:javadoc (check-api-leaks) on project java-driver-core: An error has occurred in Javadoc report generation:
[ERROR] Exit code: 1 - javadoc: warning - No source files for package com.datastax.dse.driver.internal
[ERROR] Method com.datastax.oss.driver.api.core.session.SessionBuilder.withNettyOptions leaks com.datastax.oss.driver.internal.core.context.NettyOptions (as parameter 'nettyOptions')
[ERROR] Method com.datastax.oss.driver.api.core.session.ProgrammaticArguments.getNettyOptions leaks com.datastax.oss.driver.internal.core.context.NettyOptions (as its return type)
[ERROR] Method com.datastax.oss.driver.api.core.session.ProgrammaticArguments.Builder.withNettyOptions leaks com.datastax.oss.driver.internal.core.context.NettyOptions (as parameter 'nettyOptions')
[ERROR]
[ERROR] Found 3 errors
[ERROR]
```

`NettyOptions` in an internal class, and we forbid internal classes leaking through the public API.

When accessing the driver internals, the recommended approach is to subclass existing components. You mentioned having an existing `CqlSessionBuilder` component. Why couldn't you do something as below?

```
class MyCqlSessionBuilder extends CqlSessionBuilder {

  @Override
  protected MyDriverContext buildContext(
      DriverConfigLoader configLoader, ProgrammaticArguments programmaticArguments) {
    return new MyDriverContext(configLoader, programmaticArguments);
  }

}

class MyDriverContext extends DefaultDriverContext {

  public MyDriverContext(
      DriverConfigLoader configLoader, ProgrammaticArguments programmaticArguments) {
    super(configLoader, programmaticArguments);
  }

  @Override
  protected NettyOptions buildNettyOptions() {
    return new MyNettyOptions(this);
  }
}

class MyNettyOptions extends DefaultNettyOptions {

  public MyNettyOptions(MyDriverContext context) {
    super(context);
  }
}
```

adutra:
@sokomishalov have you tried my suggestion above?

sokomishalov:
Hi, thanks for the answer!
How can I instantiate `MyCqlSessionBuilder` from the existing `CqlSessionBuilder` instance without losing its state?

adutra:
> Hi, thanks for the answer!
> How can I instantiate `MyCqlSessionBuilder` from the existing `CqlSessionBuilder` instance without losing its state?

Where is your existing builder coming from? There is no easy way of transferring the state of a builder to another.

sokomishalov:
Yeah, I got it. 
Thanks any way

PR #1529, opened 2021-02-05
[JAVA-2704] Remove protocol v5 beta status, add v6-beta

beobal:


driver-core/src/main/java/com/datastax/driver/core/Connection.java
```diff
@@ -1768,8 +1773,6 @@ protected void initChannel(SocketChannel channel) throws Exception {
    * v5.
    */
   void switchToV5Framing() {
-    assert factory.protocolVersion.compareTo(ProtocolVersion.V5) >= 0;
```
adutra, line 1771: Why remove this?

driver-core/src/main/java/com/datastax/driver/core/ProtocolVersion.java
```diff
@@ -26,13 +26,14 @@
   V2("2.0.0", 2, V1),
   V3("2.1.0", 3, V2),
   V4("2.2.0", 4, V3),
-  V5("3.10.0", 5, V4);
+  V5("3.10.0", 5, V4),
```
adutra: Are we comfortable with V5 reporting 3.10 as its min version? 

E.g. when refreshing the schema, the control connection will assume 3.10-like schema tables if the control host talks V5 but doesn't report any version. This doesn't look accurate to me, but maybe that's not really important after all.

driver-core/src/test/java/com/datastax/driver/core/SegmentBuilderTest.java
```diff
@@ -35,7 +35,7 @@
 public class SegmentBuilderTest {
 
   private static final Message.ProtocolEncoder REQUEST_ENCODER =
-      new Message.ProtocolEncoder(ProtocolVersion.V5);
+      new Message.ProtocolEncoder(ProtocolVersion.V6);
```
adutra: We could have left V5 here imo as it is more important to test the encoding of segments in V5 than in V6. Same in `SegmentToFrameDecoderTest`.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
beobal: Hmm, I have no clue what I was thinking there. I'll revert.

driver-core/src/main/java/com/datastax/driver/core/ProtocolVersion.java
beobal: Sorry, this appears to be a leftover from when I was thinking we could bring v5 out of beta for both the 3.11 and 4.0 lines. I'll make 4.0.0 the min version for V5.

driver-core/src/test/java/com/datastax/driver/core/SegmentBuilderTest.java
beobal: You're right, I'll revert it.

adutra: Thanks! Would you mind reverting `SegmentToFrameDecoderTest` as well?

adutra:
Also, I would require an entry to the changelog. Basically this:

```
### 3.11.0 (in progress)

- [improvement] JAVA-2704: Remove protocol v5 beta status, add v6-beta.

```

beobal:
Thanks @adutra, I've made those additional changes. One thing to note is that I referenced the wrong JIRA in the initial commit - JAVA-2704 is for the 4.x driver. Just something to be aware of at merge time.

adutra (approved):

adutra:
Thanks again @beobal. Indeed the right ticket number is JAVA-2705.

This looks good to go, however our integration tests are failing with:

```
com.datastax.driver.core.exceptions.TransportException: [/127.0.1.1:42535] Error initializing connection: Beta version of the protocol used (5/v5-beta), but USE_BETA flag is unset
```

This is with Cassandra 4.0-beta4.

However my understanding of the situation is that 4.0-beta4 should include both CASSANDRA-15299 and CASSANDRA-16319 (the latter having promoted v5 out of beta).

But the server logs show:

```
INFO  [main] 2021-02-16 15:22:26,914 StorageService.java:728 - Cassandra version: 4.0-beta4
INFO  [main] 2021-02-16 15:22:26,914 StorageService.java:729 - CQL version: 3.4.5
INFO  [main] 2021-02-16 15:22:26,914 StorageService.java:730 - Native protocol supported versions: 3/v3, 4/v4, 5/v5-beta (default: 4/v4)
```

So I'm puzzled... Do you know by chance what we are doing wrong in our tests? Is there a more recent version we should be using instead?

beobal:
Hey @adutra. Yep, the change to promote v5 out of beta is not in 4.0-beta4. The ticket for that is CASSANDRA-14973, and there's a branch you can use for testing [here](https://github.com/beobal/cassandra/commits/14973-trunk)

adutra:
Thanks @beobal for the explanations. I confirmed the tests pass against your branch.

driver-core/src/main/java/com/datastax/driver/core/Connection.java
beobal: @adutra Now I remember why I removed this. It's legal for a client to not supply a specific version, e.g. not call `Cluster.Builder::withProtocolVersion`, in which case `factory.protocolVersion` is null. While v5 was marked beta this wasn't a problem as `Cluster.Builder::allowProtocolVersion` was required, but now this causes an NPE in the driver. In C*, `ProtocolNegotiationTest` has a test case that covers this.

adutra: Thanks @beobal, I pushed a fix directly to 3.x, see 1af7382fe.

PR #1531, opened 2021-03-01
JAVA-2910 repreparePayloads map should use strong referenced values

songdongsheng:


adutra:
Sorry, this is not a proper fix for the issue at hand, and could introduce regressions (using non-weak values could make the cache size grow uncontrollably).

PR #1532, opened 2021-03-03
JAVA-2922: Switch to modern framing format inside a channel handler

adutra:


adutra:
@beobal I can't assign you for review but could you please take a look? Thanks!

beobal:
@adutra will do ASAP

beobal (approved):
LGTM

driver-core/src/main/java/com/datastax/driver/core/FramingFormatHandler.java
```diff
@@ -0,0 +1,78 @@
...
+  private final Connection.Factory factory;
+
+  FramingFormatHandler(Connection.Factory factory) {
+    this.factory = factory;
+  }
+
+  @Override
+  protected void decode(ChannelHandlerContext ctx, Frame frame, List<Object> out) throws Exception {
+    boolean handshakeSuccessful =
+        frame.header.opcode == Type.READY.opcode || frame.header.opcode == Type.AUTHENTICATE.opcode;
```
olim7t, line 41: The protocol allows `OPTIONS` as the first-ever message, so shouldn't we test for `SUPPORTED` here as well? Or maybe v5 does not switch to modern framing for that first `SUPPORTED` response? (I must confess I don't remember, it's been a while)

Alternatively, we could test for `opcode != ERROR` like in the original code.

olim7t (approved):
LGTM, that's an elegant way to address the issue.

driver-core/src/main/java/com/datastax/driver/core/FramingFormatHandler.java
adutra: The specs say that only READY and AUTHENTICATE triggers trigger a framing format change:

> The initial STARTUP message and any OPTIONS messages which precede it are expected to be unframed. Likewise, the responses returned by the server, SUPPORTED in response to OPTIONS and either READY or AUTHENTICATE in response to STARTUP are transmitted unframed. After sending the READY or AUTHENTICATE response to a STARTUP message, the server will begin encoding and decoding all further transmissions according to the protocol version of that STARTUP message.

That's why I figured that detecting those 2 response types would make more sense than detecting an ERROR response.

olim7t: Right, we're good then 👍

PR #1544, opened 2021-04-13
JAVA-2934: Handle empty non-final pages in ReactiveResultSetSubscription

dssysolyatin:
Hi !

I tried to use this driver with scylladb database, because it supports CQL protocol. But code started to hang for some queries:

```
val source = Source.fromPublisher(session.executeReactive(statement))
Await.result(source.runWith(Sink.ignore), scala.concurrent.duration.Duration.Inf))
println("finished")
```

The reason is that scylladb sometime returns a page with no data but with pagingState != null and it leads to code hanging

For instance, if `pages` queue contains three pages:
 - first page *with* data and with `pagingState` != null
 - second page *without* data and `pagingState` != null
 - third page *without* data and `pagingState` = null

Akka stream calls `request(1)` if it needs data  and `request` calls `drain` that sends one row to akka stream subscriber. It works well for pages with data. But if there is page without data and `pagingState` != null the following situation appears:

Akka stream calls `requests(1)` that calls `drain` that calls  `tryNext`. `tryNext` method detects that there is no data anymore in page `1` and it gets second page from `pages` queue. But because of second page has no data, tryNext() will return `null` and no data will be sent to akka stream subscriber. Because no data will be sent to akka subscriber, akka stream will not call `request(1)` anymore and `drain` will not be called from anywhere because all pages are already loaded.

I don't know why scylladb returns pages without data and `pagingState` != null. I will ask scylladb team, but it also would be good to fix this case in driver as well

### DIFF
--- a/core/src/test/java/com/datastax/dse/driver/internal/core/cql/reactive/ReactiveResultSetSubscriptionTest.java
+++ b/core/src/test/java/com/datastax/dse/driver/internal/core/cql/reactive/ReactiveResultSetSubscriptionTest.java
@@ -145,4 +145,34 @@ public class ReactiveResultSetSubscriptionTest {
     assertThat(wasAppliedSubscriber.getElements()).hasSize(1).containsExactly(true);
     assertThat(wasAppliedSubscriber.getError()).isNull();
   }
+
+  @Test
+  public void should_handle_empty_non_final_pages() {
+    CompletableFuture<AsyncResultSet> future1 = new CompletableFuture<>();
+    CompletableFuture<AsyncResultSet> future2 = new CompletableFuture<>();
+    CompletableFuture<AsyncResultSet> future3 = new CompletableFuture<>();
+    MockAsyncResultSet page1 = new MockAsyncResultSet(10, future2);
+    MockAsyncResultSet page2 = new MockAsyncResultSet(0, future3);
+    MockAsyncResultSet page3 = new MockAsyncResultSet(10, null);
+    TestSubscriber<ReactiveRow> mainSubscriber = new TestSubscriber<>(1);
+    TestSubscriber<ColumnDefinitions> colDefsSubscriber = new TestSubscriber<>();
+    TestSubscriber<ExecutionInfo> execInfosSubscriber = new TestSubscriber<>();
+    TestSubscriber<Boolean> wasAppliedSubscriber = new TestSubscriber<>();
+    ReactiveResultSetSubscription<AsyncResultSet> subscription =
+        new ReactiveResultSetSubscription<>(
+            mainSubscriber, colDefsSubscriber, execInfosSubscriber, wasAppliedSubscriber);
+    mainSubscriber.onSubscribe(subscription);
+    subscription.start(() -> future1);
+    future1.complete(page1);
+    future2.complete(page2);
+    // emulate backpressure
+    subscription.request(1);
+    future3.complete(page3);
+    subscription.request(Long.MAX_VALUE);
+    mainSubscriber.awaitTermination();
+    assertThat(mainSubscriber.getError()).isNull();
+    List<Row> expected = new ArrayList<>(page1.currentPage());
+    expected.addAll(page3.currentPage());
+    assertThat(mainSubscriber.getElements()).hasSize(20).extracting("row").isEqualTo(expected);
+  }
 }
--- a/core/src/test/java/com/datastax/dse/driver/internal/core/cql/reactive/TestSubscriber.java
+++ b/core/src/test/java/com/datastax/dse/driver/internal/core/cql/reactive/TestSubscriber.java
@@ -31,8 +31,17 @@ public class TestSubscriber<T> implements Subscriber<T> {
 
   private final List<T> elements = new ArrayList<>();
   private final CountDownLatch latch = new CountDownLatch(1);
+  private final long demand;
   private Subscription subscription;
   private Throwable error;
+
+  public TestSubscriber() {
+    this.demand = Long.MAX_VALUE;
+  }
+
+  public TestSubscriber(long demand) {
+    this.demand = demand;
+  }
 
   @Override
   public void onSubscribe(Subscription s) {
@@ -40,7 +49,7 @@ public class TestSubscriber<T> implements Subscriber<T> {
       fail("already subscribed");
     }
     subscription = s;
-    s.request(Long.MAX_VALUE);
+    subscription.request(demand);
   }
 
   @Override
@@ -71,5 +80,6 @@ public class TestSubscriber<T> implements Subscriber<T> {
 
   public void awaitTermination() {
     Uninterruptibles.awaitUninterruptibly(latch, 1, TimeUnit.MINUTES);
+    if (latch.getCount() > 0) fail("subscriber not terminated");
   }
 }
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -2,6 +2,10 @@
 
 <!-- Note: contrary to 3.x, insert new entries *first* in their section -->
 
+### 4.12.0 (in progress)
+
+- [bug] JAVA-2934: Handle empty non-final pages in ReactiveResultSetSubscription
+
 ### 4.11.0
 
 - [improvement] JAVA-2930: Allow Micrometer to record histograms for timers
--- a/core/src/main/java/com/datastax/dse/driver/internal/core/cql/reactive/ReactiveResultSetSubscription.java
+++ b/core/src/main/java/com/datastax/dse/driver/internal/core/cql/reactive/ReactiveResultSetSubscription.java
@@ -277,13 +277,10 @@ public class ReactiveResultSetSubscription<ResultSetT extends AsyncPagingIterabl
         if (pages.poll() == null) {
           throw new AssertionError("Queue is empty, this should not happen");
         }
-        current = pages.peek();
         // if the next page is readily available,
         // serve its first row now, no need to wait
         // for the next drain.
-        if (current != null && current.hasMoreRows()) {
-          return current.nextRow();
-        }
+        return tryNext();
       }
     }
     // No items available right now.


adutra (changes requested):
Nice and elegant :-)

Could you please:

* Add a unit test for this specific case in `ReactiveResultSetSubscriptionTest`
* Change the commit and PR name to `JAVA-2934: Handle empty non-final pages in ReactiveResultSetSubscription`
* Add an entry to the changelog:

```
4.12.0 (in progress)

- [bug] JAVA-2934: Handle empty non-final pages in ReactiveResultSetSubscription
```

Thanks!

dssysolyatin:
@adutra I fixed PR. Could you re-review my PR ?

core/src/test/java/com/datastax/dse/driver/internal/core/cql/reactive/ReactiveResultSetSubscriptionTest.java
```diff
@@ -145,4 +145,31 @@ public void should_report_error_on_intermediary_page() {
...
+    subscription.start(() -> future1);
+    future1.complete(page1);
+    future2.complete(page2);
+    // emulate backpressure
+    subscription.request(1);
+    future3.complete(page3);
+    subscription.request(Long.MAX_VALUE);
+    mainSubscriber.awaitTermination();
+    assertThat(mainSubscriber.isCompleted()).isTrue();
+  }
```
adutra: It's a minor thing but I prefer to not change `TestSubscriber` and rewrite the test as suggested.

```suggestion
  public void should_handle_empty_non_final_pages() {
    CompletableFuture<AsyncResultSet> future1 = new CompletableFuture<>();
    CompletableFuture<AsyncResultSet> future2 = new CompletableFuture<>();
    CompletableFuture<AsyncResultSet> future3 = new CompletableFuture<>();
    MockAsyncResultSet page1 = new MockAsyncResultSet(1, future2);
    MockAsyncResultSet page2 = new MockAsyncResultSet(0, future3); // empty
    MockAsyncResultSet page3 = new MockAsyncResultSet(1, null);
    TestSubscriber<ReactiveRow> mainSubscriber = new TestSubscriber<>();
    TestSubscriber<ColumnDefinitions> colDefsSubscriber = new TestSubscriber<>();
    TestSubscriber<ExecutionInfo> execInfosSubscriber = new TestSubscriber<>();
    TestSubscriber<Boolean> wasAppliedSubscriber = new TestSubscriber<>();
    ReactiveResultSetSubscription<AsyncResultSet> subscription =
        new ReactiveResultSetSubscription<>(
            mainSubscriber, colDefsSubscriber, execInfosSubscriber, wasAppliedSubscriber);
    mainSubscriber.onSubscribe(subscription);
    subscription.start(() -> future1);
    future1.complete(page1);
    future2.complete(page2);
    // row 1 should be served from page1
    assertThat(mainSubscriber.getElements()).hasSize(1);
    // page1 and page2 should have been discarded, queue should be empty
    assertThat(subscription)
        .extracting("pages.elements")
        .satisfies(queue -> assertThat(((Queue<?>) queue)).isEmpty());
    // row 2 should be served from page3
    future3.complete(page3);
    mainSubscriber.awaitTermination();
    // should have completed normally
    assertThat(mainSubscriber.getError()).isNull();
    // should have received 2 rows
    List<Row> expected = new ArrayList<>(page1.currentPage());
    expected.addAll(page3.currentPage());
    assertThat(mainSubscriber.getElements()).hasSize(2).extracting("row").isEqualTo(expected);
  }
```

core/src/test/java/com/datastax/dse/driver/internal/core/cql/reactive/TestSubscriber.java
```diff
@@ -31,16 +31,26 @@
 
   private final List<T> elements = new ArrayList<>();
   private final CountDownLatch latch = new CountDownLatch(1);
+  private long demand;
```
adutra: Please revert changes in this file, see my other comment + suggestion.

adutra (changes requested):
Looking good, please do the changes I suggested and this should be good to go. Thank you for your time and involvement!

core/src/test/java/com/datastax/dse/driver/internal/core/cql/reactive/ReactiveResultSetSubscriptionTest.java
dssysolyatin: If I rewrite test how you suggested it will pass with previous code. Because `future3.complete(page3);` will call `drain` -> `tryNext()` and code successfully complete.

I tried to rewrite this test without changing `TestSubscriber`, but didn't come up with anything

adutra: I tried my changes before suggesting them. They work as expected, feel free to cherry-pick: 

https://github.com/adutra/java-driver/commit/466e5ddc276576495198be0d5a89a0e2d538c482

dssysolyatin: @adutra Yes, it works. But it works for case
```
current = pages.peek();
        // if the next page is readily available,
        // serve its first row now, no need to wait
        // for the next drain.
        if (current != null && current.hasMoreRows()) {
          return current.nextRow();
        }
```
as well

dssysolyatin: I meant if you change branch to `4.x` branch and try to use your test. It will work, but it should fail

adutra: Ah right, now I understand what you had in mind. Please disregard my previous comments :-)

core/src/test/java/com/datastax/dse/driver/internal/core/cql/reactive/TestSubscriber.java
```diff
@@ -69,6 +80,10 @@ public Throwable getError() {
     return elements;
   }
 
+  public boolean isCompleted() {
```
adutra: I don't like much this new boolean property because it is ambiguous. Some people consider that a flux that terminates with an error is completed. I don't think we need this field imo; but if you absolutely want to expose a boolean property to indicate the termination status, I'd suggest instead to use the word "terminated":

```
  public boolean isTerminated() {
    return latch.getCount() == 0;
  }
``` 

If you want to further distinguish normal vs abnormal termination, you can then inspect `getError() == null`.

core/src/test/java/com/datastax/dse/driver/internal/core/cql/reactive/TestSubscriber.java
adutra:
```suggestion
  private final long demand;
```

core/src/test/java/com/datastax/dse/driver/internal/core/cql/reactive/ReactiveResultSetSubscriptionTest.java
```diff
@@ -145,4 +145,31 @@ public void should_report_error_on_intermediary_page() {
...
+    mainSubscriber.onSubscribe(subscription);
+    subscription.start(() -> future1);
+    future1.complete(page1);
+    future2.complete(page2);
+    // emulate backpressure
+    subscription.request(1);
+    future3.complete(page3);
+    subscription.request(Long.MAX_VALUE);
+    mainSubscriber.awaitTermination();
+    assertThat(mainSubscriber.isCompleted()).isTrue();
```
adutra: Even if you don't change the entire test, at least you could add the following assertions at the end:

```
    assertThat(mainSubscriber.getError()).isNull();
    List<Row> expected = new ArrayList<>(page1.currentPage());
    expected.addAll(page3.currentPage());
    assertThat(mainSubscriber.getElements()).hasSize(...).extracting("row").isEqualTo(expected);
```

core/src/test/java/com/datastax/dse/driver/internal/core/cql/reactive/ReactiveResultSetSubscriptionTest.java
```diff
@@ -145,4 +145,31 @@ public void should_report_error_on_intermediary_page() {
...
   }
+
+  @Test
+  public void should_handle_empty_non_final_pages() {
+    CompletableFuture<AsyncResultSet> future1 = new CompletableFuture<>();
+    CompletableFuture<AsyncResultSet> future2 = new CompletableFuture<>();
+    CompletableFuture<AsyncResultSet> future3 = new CompletableFuture<>();
+    MockAsyncResultSet page1 = new MockAsyncResultSet(10, future2);
+    MockAsyncResultSet page2 = new MockAsyncResultSet(0, future3);
+    MockAsyncResultSet page3 = new MockAsyncResultSet(0, null);
```
adutra: I think it could be interesting to test the case where `page3 = new MockAsyncResultSet(10, null);`. From your discussions with the Scylla devs, it seems that this is something that could happen.

adutra (changes requested):

core/src/test/java/com/datastax/dse/driver/internal/core/cql/reactive/TestSubscriber.java
dssysolyatin: I need it because those two lines don't work for my test case
```
mainSubscriber.awaitTermination();
assertThat(mainSubscriber.getError()).isNull();
```

`mainSubscriber.awaitTermination()` does not throw any exception if after timeout there is still data in publisher.  In this regard, the question, is it correct behaviour that `mainSubscriber.awaitTermination()` does not throw exception after timeout ?

adutra: > is it correct behaviour that mainSubscriber.awaitTermination() does not throw exception after timeout ?

Arguably not :-)
Feel free to change the method to throw throw an assertion error if the timeout fires.

core/src/test/java/com/datastax/dse/driver/internal/core/cql/reactive/ReactiveResultSetSubscriptionTest.java
dssysolyatin: added

core/src/test/java/com/datastax/dse/driver/internal/core/cql/reactive/ReactiveResultSetSubscriptionTest.java
dssysolyatin: Yeap, I changed 0 to 10

core/src/test/java/com/datastax/dse/driver/internal/core/cql/reactive/TestSubscriber.java
dssysolyatin: I deleted isCompleted method and fixed `awaitTermination`

adutra (approved):
Thank you so much @dssysolyatin !

dssysolyatin:
Thank you for reviewing of PR @adutra

adutra:
@dssysolyatin sorry I forgot the legal part :-) I need you to sign our CLA please:

https://cla.datastax.com/

Please ping me when you've signed it. Many thanks!

dssysolyatin:
@adutra No problem. Done ;)

PR #1547, opened 2021-04-29
JAVA-2938: servererrors.OverloadedException message is misleading

chicobento:
OverloadedException is misleading probably due to a copy&paste mistake.

Adding the original server message to the exception, as in 3.x series.

Original thread: https://groups.google.com/a/lists.datastax.com/g/java-driver-user/c/KNPE34iOBgI/m/fWv5zOxNAQAJ

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/servererrors/OverloadedException.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/servererrors/OverloadedException.java
@@ -34,8 +34,8 @@ import edu.umd.cs.findbugs.annotations.Nullable;
  */
 public class OverloadedException extends QueryExecutionException {
 
-  public OverloadedException(@NonNull Node coordinator) {
-    super(coordinator, String.format("%s is bootstrapping", coordinator), null, false);
+  public OverloadedException(@NonNull Node coordinator, @NonNull String message) {
+    super(coordinator, String.format("%s is overloaded: %s", coordinator, message), null, false);
   }
 
   private OverloadedException(
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -2,6 +2,10 @@
 
 <!-- Note: contrary to 3.x, insert new entries *first* in their section -->
 
+### 4.11.2 (in progress)
+
+- [bug] JAVA-2938: OverloadedException message is misleading
+
 ### 4.11.1
 
 - [bug] JAVA-2910: Add a configuration option to support strong values for prepared statements cache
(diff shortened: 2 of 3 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/servererrors/OverloadedException.java
```diff
@@ -34,8 +34,8 @@
  */
 public class OverloadedException extends QueryExecutionException {
 
-  public OverloadedException(@NonNull Node coordinator) {
-    super(coordinator, String.format("%s is bootstrapping", coordinator), null, false);
+  public OverloadedException(@NonNull Node coordinator, @NonNull String message) {
+    super(coordinator, message, null, false);
```
adutra: I think we could do as in 3.x and augment the original message with the coordinator address:

```suggestion
    super(coordinator, String.format("%s was overloaded: %s", coordinator, message), null, false);
```

adutra (changes requested):

adutra (changes requested):
@chicobento looking really good!

Do you mind adding an entry to the changelog as below:

```
4.11.2 (in progress)

- [bug] JAVA-2938: OverloadedException message is misleading

```

Also, we don't have tests covering the change. But I admit that the change is very small so I can live without a test for it.

core/src/main/java/com/datastax/oss/driver/api/core/servererrors/OverloadedException.java
chicobento: Agree. Uploaded a new version of the commit, used "is" instead of "was" to be in-synch with the other error message (bootstraping). But let me know if you prefer was to be in synch with 3.x

chicobento:
> 
> 
> @chicobento looking really good!
> 
> Do you mind adding an entry to the changelog as below:
> 
> ```
> 4.11.2 (in progress)
> 
> - [bug] JAVA-2938: OverloadedException message is misleading
> ```
> 
> Also, we don't have tests covering the change. But I admit that the change is very small so I can live without a test for it.

Yup, I looked for TCs in that sense and I couldn't find any.
I just added the changelog change on the original commit

adutra (approved):
LGTM. Muito obrigado @chicobento 👍

adutra:
@chicobento I committed directly to 4.x with a few modifications to make the changes compatible with the existing API. 

One last thing: could you please sign our CLA?

https://cla.datastax.com/

Thanks!

adutra:
@chicobento sorry to insist: please sign our CLA, otherwise I will have to revert your change. Thanks!

https://cla.datastax.com/

chicobento:
Sorry, Im on vacation and have been disconnected for the last days. Just
signed, hope it's fine now.

Tks,
Francisco

adutra:
Thank you! Closing now. Boa sorte a até a próxima :-)

PR #1555, opened 2021-06-05
A call to  CachingCodecRegistry.codecFor for HeapByteBuffer throws a CodecNotFoundException 

QuentinFAIDIDE:
Please forgive my lack of java proficiency, I'm not a java dev.
It seems like updating our spark job to spark 3 raised some issues with cassandra-driver for scala (see graphsense/graphsense-transformation#38 ).
After some research the error was narrowed down to the `CachingCodedRegistry` and `BlobCodec` classes.
In the first one, in the codecFor function, the primitiveCodec.accept function was sent a JavaType and not a Class type object, leading to the test for ByteBuffer.class.equals(javaClass) to fail which in turn lead to a `CodecNotFoundException` being thrown.
In the BlobCodec class, only ByteBuffer.class.equals(javaClass) was used and it didn't test for child classes.
I think this issue may require a fix.
- Quentin.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/BlobCodec.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/BlobCodec.java
@@ -47,7 +47,7 @@ public class BlobCodec implements TypeCodec<ByteBuffer> {
 
   @Override
   public boolean accepts(@NonNull Class<?> javaClass) {
-    return ByteBuffer.class.equals(javaClass);
+    return ByteBuffer.class.equals(javaClass) || ByteBuffer.class.isAssignableFrom(javaClass);
   }
 
   @Nullable
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
@@ -265,7 +265,7 @@ public abstract class CachingCodecRegistry implements MutableCodecRegistry {
     LOG.trace("[{}] Looking up codec for object {}", logPrefix, value);
 
     for (TypeCodec<?> primitiveCodec : primitiveCodecs) {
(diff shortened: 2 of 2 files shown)


adutra:
Hi @QuentinFAIDIDE I appreciate your involvement but unfortunately your fix is not correct. The `TypeCodec` interface has 3 `accept` methods:

* accepts(GenericType<?> javaType)
* accepts(Class<?> javaClass)
* accepts(Object value)

They have different contracts and usages.

I think your original problem may not reside in the driver, but in a layer above. The error you are seeing is:

```
CodecNotFoundException: Codec not found for requested operation: [BLOB <-> java.nio.HeapByteBuffer]
```

This means that someone is calling `CachingCodecRegistry.codecFor(DataType, Class<?>)` with wrong parameters. 

If you are looking for a BLOB codec, the right call is `CachingCodecRegistry.codecFor(DataType.BLOB, ByteBuffer.class)`. Indeed, there is a default BLOB codec registered in the driver that converts BLOB columns to `ByteBuffer` objects.

But, as explained in the docs, this method _does not accept covariant classes_. So if you try, say, `CachingCodecRegistry.codecFor(DataType.BLOB, HeapByteBuffer.class)` – then the method will throw `CodecNotFoundException` – even if `HeapByteBuffer` is a subtype of `ByteBuffer`.

Please check the code making usage of the CodecRegistry API and verify that all calls are correct.

Also: there are several `codecFor` methods, please read their javadocs carefully. A possible explanation is that the calling code is calling the wrong `codecFor` method.

QuentinFAIDIDE:
Ok, thank you for the fast response, I indeed was not sure where to look for the doc of those classes, and did not check for the accept method overloads or the classes javadocs before making my dirty fix :grimacing: .
I will try when I'll have time to lurk more on the documentation/code and find where the guilty codecFor call lies.
Thank you

adutra:
Hi @QuentinFAIDIDE I'm going to close this as we can't merge it, I hope you are OK with that.

PR #1559, opened 2021-06-28
Implement specific dependency check for Graal substitutions

absurdfarce:
As of JAVA-2917 the usage of substitutions for Graal native image builds was modified to use DependencyCheck to determine whether or not the substitution should be used.  DependencyCheck leverages driver.internal.core.util.Reflection which in turn utilizes a static slf4j Logger to record what it does.  The reliance on the Logger introduced several unexpected build-time initialization errors as documented in JAVA-2940.

This PR sketches a way to resolve that problem.  There's some subtlety here, however.  We certainly want to keep a single representation of the dependencies themselves.  DependencyCheck used to perform that role, but it intertwined that logic with checks for the presence of those dependencies via reflection.  And as discussed above this doesn't work cleanly in all cases.  Ideally we would just implement different methods for checking for these dependencies and then use the appropriate method in each case.

This PR decouples these roles:

* Dependency now serves as the canonical representation for a logical dependency
* There are two methods for checking for dependencies.  The default impl still uses d.i.c.u.Reflection (to benefit from the nice logging it offers).  The Graal-specific version just falls back on Class.forName() directly.  This makes it very similar to the Reflection-based impl in the end, although we do lose out on some of the niceties

Local testing has confirmed that this resolves the errors referenced in JAVA-2940


absurdfarce:
@adutra I wasn't sure if we want this to be against 4.x (for 4.13.0) or 4.11.x (for 4.11.next).  The more I think about it the more it probably makes sense to shift it over to 4.11.x.  Let me know what you think.

adutra:
> @adutra I wasn't sure if we want this to be against 4.x (for 4.13.0) or 4.11.x (for 4.11.next). The more I think about it the more it probably makes sense to shift it over to 4.11.x. Let me know what you think.

I would suggest targeting both as I would like to keep supporting 4.11.x for some time. So I'd recommend setting the base branch to 4.11.x initially, then merge upstream.

core/src/main/java/com/datastax/oss/driver/internal/core/util/DefaultDependencyChecker.java
```diff
@@ -0,0 +1,59 @@
...
+  public static boolean isPresent(Dependency dependency) {
+    try {
+      return CACHE.computeIfAbsent(
+          dependency,
+          (dep) -> {
+            for (String classNameToTest : dependency.classes()) {
+              // Always use the driver class loader, assuming that the driver classes and
+              // the dependency classes are either being loaded by the same class loader,
+              // or – as in OSGi deployments – by two distinct, but compatible class loaders.
+              if (Reflection.loadClass(null, classNameToTest) == null) {
```
adutra, line 48: I wonder: can't we get rid of this class and use only the other one? My thinking is that calling `Reflection.loadClass(null, classNameToTest)` effectively boils down to `Class.forName(classNameToTest)` since the CL is null.

core/src/main/java/com/datastax/oss/driver/internal/core/util/GraalDependencyChecker.java
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * A dependency checker implementation which should be safe to use for build-time checks when
+ * building Graal native images. This class is similar to {@link DefaultDependencyChecker} but
+ * doesn't introduce any external dependencies which might complicate the native image build
+ * process. Expectation is that this will be most prominently used in the various predicate classes
+ * which determine whether or not Graal substitutions should be used.
+ */
+public class GraalDependencyChecker {
+
+  private static ConcurrentHashMap<Dependency, Boolean> CACHE = new ConcurrentHashMap<>();
```
adutra, line 29:
```suggestion
  private static final ConcurrentHashMap<Dependency, Boolean> CACHE = new ConcurrentHashMap<>();
```

core/src/main/java/com/datastax/oss/driver/internal/core/util/GraalDependencyChecker.java
```diff
@@ -0,0 +1,60 @@
...
+          (dep) -> {
+            for (String classNameToTest : dependency.classes()) {
+              // Note that this lands in a pretty similar spot to
+              // Reflection.loadClass() with a null class loader
+              // arg.  Major difference here is that we avoid the
+              // more complex exception handling/logging ops in
+              // that code.
+              try {
+                Class.forName(classNameToTest);
+              } catch (Exception e) {
```
adutra, line 50: I would be more careful here and also catch `LinkageError`.
```suggestion
              } catch (LinkageError | Exception e) {
```

adutra (commented):
I think we should open a Jira ticket for this so that the motivations and fix are clearly documented (and also included in the changelog). I will do that now.

adutra:
https://datastax-oss.atlassian.net/browse/JAVA-2950

Please add an entry to the changelog under a new section `### 4.11.3 (in progress)`. Thanks!

absurdfarce:
Agreed about changing to 4.11.x; I'll update the PR shortly.  I'll also note that the JIRA ticket referenced by @adutra above is JAVA-2950.

absurdfarce:
Bah, I'm gonna open a new PR with the same changes.  If I just switch to 4.11.x on the PR I'll bring in the commits on 4.x since it was branched off of 4.11.x... and we don't want that.  Similar problem (in reverse) if I rebase locally; my local branch would now be _behind_ this PR so it refuses a push.  If I pull this PR back locally we'll bring those commits back in, which we don't want.

Easiest to just open a new PR.

absurdfarce:
Closing in favor of https://github.com/datastax/java-driver/pull/1561.  Outstanding comments will be addressed on the new PR.

PR #1560, opened 2021-07-02
JAVA-2940: Add GraalVM native image build configurations

absurdfarce:
Spun off of work done on JAVA-2940

Developed and tested as a feature branch off of https://github.com/datastax/java-driver/pull/1559.  Moved over to 4.11.x under the assumption we might want to fix this with a 4.11 release as well.  Can always move it back if we decide not to do so.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
@@ -272,98 +272,121 @@ public class ChannelFactory {
       DriverChannelOptions options,
       NodeMetricUpdater nodeMetricUpdater,
       CompletableFuture<DriverChannel> resultFuture) {
-    return new ChannelInitializer<Channel>() {
-      @Override
-      protected void initChannel(Channel channel) {
-        try {
-          DriverExecutionProfile defaultConfig = context.getConfig().getDefaultProfile();
+    return new ChannelFactoryInitializer(
+        endPoint, protocolVersion, options, nodeMetricUpdater, resultFuture);
+  };
 
-          long setKeyspaceTimeoutMillis =
-              defaultConfig
-                  .getDuration(DefaultDriverOption.CONNECTION_SET_KEYSPACE_TIMEOUT)
-                  .toMillis();
-          int maxFrameLength =
-              (int) defaultConfig.getBytes(DefaultDriverOption.PROTOCOL_MAX_FRAME_LENGTH);
-          int maxRequestsPerConnection =
-              defaultConfig.getInt(DefaultDriverOption.CONNECTION_MAX_REQUESTS);
-          int maxOrphanRequests =
-              defaultConfig.getInt(DefaultDriverOption.CONNECTION_MAX_ORPHAN_REQUESTS);
-          if (maxOrphanRequests >= maxRequestsPerConnection) {
-            if (LOGGED_ORPHAN_WARNING.compareAndSet(false, true)) {
-              LOG.warn(
-                  "[{}] Invalid value for {}: {}. It must be lower than {}. "
-                      + "Defaulting to {} (1/4 of max-requests) instead.",
-                  logPrefix,
-                  DefaultDriverOption.CONNECTION_MAX_ORPHAN_REQUESTS.getPath(),
-                  maxOrphanRequests,
-                  DefaultDriverOption.CONNECTION_MAX_REQUESTS.getPath(),
-                  maxRequestsPerConnection / 4);
-            }
-            maxOrphanRequests = maxRequestsPerConnection / 4;
+  class ChannelFactoryInitializer extends ChannelInitializer<Channel> {
+
+    private final EndPoint endPoint;
+    private final ProtocolVersion protocolVersion;
+    private final DriverChannelOptions options;
+    private final NodeMetricUpdater nodeMetricUpdater;
+    private final CompletableFuture<DriverChannel> resultFuture;
+
+    ChannelFactoryInitializer(
+        EndPoint endPoint,
+        ProtocolVersion protocolVersion,
+        DriverChannelOptions options,
+        NodeMetricUpdater nodeMetricUpdater,
+        CompletableFuture<DriverChannel> resultFuture) {
+
+      this.endPoint = endPoint;
+      this.protocolVersion = protocolVersion;
+      this.options = options;
+      this.nodeMetricUpdater = nodeMetricUpdater;
+      this.resultFuture = resultFuture;
+    }
+
+    @Override
+    protected void initChannel(Channel channel) {
+      try {
+        DriverExecutionProfile defaultConfig = context.getConfig().getDefaultProfile();
+
+        long setKeyspaceTimeoutMillis =
+            defaultConfig
+                .getDuration(DefaultDriverOption.CONNECTION_SET_KEYSPACE_TIMEOUT)
+                .toMillis();
+        int maxFrameLength =
+            (int) defaultConfig.getBytes(DefaultDriverOption.PROTOCOL_MAX_FRAME_LENGTH);
+        int maxRequestsPerConnection =
+            defaultConfig.getInt(DefaultDriverOption.CONNECTION_MAX_REQUESTS);
+        int maxOrphanRequests =
+            defaultConfig.getInt(DefaultDriverOption.CONNECTION_MAX_ORPHAN_REQUESTS);
+        if (maxOrphanRequests >= maxRequestsPerConnection) {
+          if (LOGGED_ORPHAN_WARNING.compareAndSet(false, true)) {
+            LOG.warn(
+                "[{}] Invalid value for {}: {}. It must be lower than {}. "
+                    + "Defaulting to {} (1/4 of max-requests) instead.",
+                logPrefix,
+                DefaultDriverOption.CONNECTION_MAX_ORPHAN_REQUESTS.getPath(),
+                maxOrphanRequests,
+                DefaultDriverOption.CONNECTION_MAX_REQUESTS.getPath(),
+                maxRequestsPerConnection / 4);
           }
-
-          InFlightHandler inFlightHandler =
-              new InFlightHandler(
-                  protocolVersion,
-                  new StreamIdGenerator(maxRequestsPerConnection),
-                  maxOrphanRequests,
-                  setKeyspaceTimeoutMillis,
-                  channel.newPromise(),
-                  options.eventCallback,
-                  options.ownerLogPrefix);
-          HeartbeatHandler heartbeatHandler = new HeartbeatHandler(defaultConfig);
-          ProtocolInitHandler initHandler =
-              new ProtocolInitHandler(
-                  context,
-                  protocolVersion,
-                  clusterName,
-                  endPoint,
-                  options,
-                  heartbeatHandler,
-                  productType == null);
-
-          ChannelPipeline pipeline = channel.pipeline();
-          context
-              .getSslHandlerFactory()
-              .map(f -> f.newSslHandler(channel, endPoint))
-              .map(h -> pipeline.addLast(SSL_HANDLER_NAME, h));
-
-          // Only add meter handlers on the pipeline if metrics are enabled.
-          SessionMetricUpdater sessionMetricUpdater =
-              context.getMetricsFactory().getSessionUpdater();
-          if (nodeMetricUpdater.isEnabled(DefaultNodeMetric.BYTES_RECEIVED, null)
-              || sessionMetricUpdater.isEnabled(DefaultSessionMetric.BYTES_RECEIVED, null)) {
-            pipeline.addLast(
-                INBOUND_TRAFFIC_METER_NAME,
-                new InboundTrafficMeter(nodeMetricUpdater, sessionMetricUpdater));
-          }
-
-          if (nodeMetricUpdater.isEnabled(DefaultNodeMetric.BYTES_SENT, null)
-              || sessionMetricUpdater.isEnabled(DefaultSessionMetric.BYTES_SENT, null)) {
-            pipeline.addLast(
-                OUTBOUND_TRAFFIC_METER_NAME,
-                new OutboundTrafficMeter(nodeMetricUpdater, sessionMetricUpdater));
-          }
-
-          pipeline
-              .addLast(
-                  FRAME_TO_BYTES_ENCODER_NAME,
-                  new FrameEncoder(context.getFrameCodec(), maxFrameLength))
-              .addLast(
-                  BYTES_TO_FRAME_DECODER_NAME,
-                  new FrameDecoder(context.getFrameCodec(), maxFrameLength))
-              // Note: HeartbeatHandler is inserted here once init completes
-              .addLast(INFLIGHT_HANDLER_NAME, inFlightHandler)
-              .addLast(INIT_HANDLER_NAME, initHandler);
-
-          context.getNettyOptions().afterChannelInitialized(channel);
-        } catch (Throwable t) {
-          // If the init handler throws an exception, Netty swallows it and closes the channel. We
-          // want to propagate it instead, so fail the outer future (the result of connect()).
-          resultFuture.completeExceptionally(t);
-          throw t;
+          maxOrphanRequests = maxRequestsPerConnection / 4;
         }
+
+        InFlightHandler inFlightHandler =
+            new InFlightHandler(
+                protocolVersion,
+                new StreamIdGenerator(maxRequestsPerConnection),
+                maxOrphanRequests,
+                setKeyspaceTimeoutMillis,
+                channel.newPromise(),
+                options.eventCallback,
+                options.ownerLogPrefix);
+        HeartbeatHandler heartbeatHandler = new HeartbeatHandler(defaultConfig);
+        ProtocolInitHandler initHandler =
+            new ProtocolInitHandler(
+                context,
+                protocolVersion,
+                clusterName,
+                endPoint,
+                options,
+                heartbeatHandler,
+                productType == null);
+
+        ChannelPipeline pipeline = channel.pipeline();
+        context
+            .getSslHandlerFactory()
+            .map(f -> f.newSslHandler(channel, endPoint))
+            .map(h -> pipeline.addLast(SSL_HANDLER_NAME, h));
+
+        // Only add meter handlers on the pipeline if metrics are enabled.
+        SessionMetricUpdater sessionMetricUpdater = context.getMetricsFactory().getSessionUpdater();
+        if (nodeMetricUpdater.isEnabled(DefaultNodeMetric.BYTES_RECEIVED, null)
+            || sessionMetricUpdater.isEnabled(DefaultSessionMetric.BYTES_RECEIVED, null)) {
+          pipeline.addLast(
+              INBOUND_TRAFFIC_METER_NAME,
+              new InboundTrafficMeter(nodeMetricUpdater, sessionMetricUpdater));
+        }
+
+        if (nodeMetricUpdater.isEnabled(DefaultNodeMetric.BYTES_SENT, null)
+            || sessionMetricUpdater.isEnabled(DefaultSessionMetric.BYTES_SENT, null)) {
+          pipeline.addLast(
+              OUTBOUND_TRAFFIC_METER_NAME,
+              new OutboundTrafficMeter(nodeMetricUpdater, sessionMetricUpdater));
+        }
+
+        pipeline
+            .addLast(
+                FRAME_TO_BYTES_ENCODER_NAME,
+                new FrameEncoder(context.getFrameCodec(), maxFrameLength))
+            .addLast(
+                BYTES_TO_FRAME_DECODER_NAME,
+                new FrameDecoder(context.getFrameCodec(), maxFrameLength))
+            // Note: HeartbeatHandler is inserted here once init completes
+            .addLast(INFLIGHT_HANDLER_NAME, inFlightHandler)
+            .addLast(INIT_HANDLER_NAME, initHandler);
+
+        context.getNettyOptions().afterChannelInitialized(channel);
+      } catch (Throwable t) {
+        // If the init handler throws an exception, Netty swallows it and closes the channel. We
+        // want to propagate it instead, so fail the outer future (the result of connect()).
+        resultFuture.completeExceptionally(t);
+        throw t;
       }
-    };
+    }
   }
 }
--- a/core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/native-image.properties
+++ b/core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/native-image.properties
@@ -1,0 +1,7 @@
+Args=-H:IncludeResources=reference\\.conf \
+  -H:IncludeResources=application\\.conf \
+  -H:IncludeResources=application\\.json \
+  -H:IncludeResources=application\\.properties \
+  -H:IncludeResources=.*Driver\\.properties \
+  -H:DynamicProxyConfigurationResources=${.}/proxy.json \
+  -H:ReflectionConfigurationResources=${.}/reflection.json
--- a/core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
+++ b/core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
@@ -1,0 +1,154 @@
+[
+  {
+    "name": "com.datastax.oss.driver.internal.core.loadbalancing.DefaultLoadBalancingPolicy",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext", "java.lang.String" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.loadbalancing.DcInferringLoadBalancingPolicy",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext", "java.lang.String" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.loadbalancing.BasicLoadBalancingPolicy",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext", "java.lang.String" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.connection.ExponentialReconnectionPolicy",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.connection.ConstantReconnectionPolicy",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.retry.DefaultRetryPolicy",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext", "java.lang.String" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.retry.ConsistencyDowngradingRetryPolicy",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext", "java.lang.String" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.specex.NoSpeculativeExecutionPolicy",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext", "java.lang.String" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.specex.ConstantSpeculativeExecutionPolicy",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext", "java.lang.String" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.time.AtomicTimestampGenerator",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.time.ThreadLocalTimestampGenerator",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.time.ServerSideTimestampGenerator",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.tracker.NoopRequestTracker",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.tracker.RequestLogger",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.session.throttling.PassThroughRequestThrottler",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.session.throttling.RateLimitingRequestThrottler",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.session.throttling.ConcurrencyLimitingRequestThrottler",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.metadata.NoopNodeStateListener",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.metadata.schema.NoopSchemaChangeListener",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.addresstranslation.PassThroughAddressTranslator",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.addresstranslation.Ec2MultiRegionAddressTranslator",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.auth.PlainTextAuthProvider",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.dse.driver.internal.core.auth.DseGssApiAuthProvider",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.ssl.DefaultSslEngineFactory",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.metrics.DefaultMetricsFactory",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.metrics.NoopMetricsFactory",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.metrics.DropwizardMetricsFactory",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.metrics.DefaultMetricIdGenerator",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.metrics.TaggingMetricIdGenerator",
+    "methods": [ { "name": "<init>", "parameterTypes": [ "com.datastax.oss.driver.api.core.context.DriverContext" ] } ]
+  },
+  {
+    "name": "io.netty.channel.socket.nio.NioSocketChannel",
+    "methods": [ { "name": "<init>", "parameterTypes": [] } ]
+  },
+  {
+    "name": "io.netty.buffer.AbstractByteBufAllocator",
+    "methods": [ { "name": "toLeakAwareBuffer", "parameterTypes": ["io.netty.buffer.ByteBuf" ] } ]
+  },
+  {
+    "name": "io.netty.util.ReferenceCountUtil",
+    "methods": [ { "name": "touch", "parameterTypes": ["java.lang.Object", "java.lang.Object" ] } ]
+  },
+  {
+    "name" : "io.netty.util.internal.shaded.org.jctools.queues.MpscArrayQueueProducerIndexField",
+    "fields": [ {"name": "producerIndex", "allowUnsafeAccess": true} ]
+  },
+  {
+    "name" : "io.netty.util.internal.shaded.org.jctools.queues.MpscArrayQueueProducerLimitField",
+    "fields": [ {"name": "producerLimit", "allowUnsafeAccess": true} ]
+  },
+  {
+    "name" : "io.netty.util.internal.shaded.org.jctools.queues.MpscArrayQueueConsumerIndexField",
+    "fields": [ {"name": "consumerIndex", "allowUnsafeAccess": true} ]
+  },
+  {
+    "name" : "io.netty.util.internal.shaded.org.jctools.queues.BaseMpscLinkedArrayQueueProducerFields",
+    "fields": [ {"name": "producerIndex", "allowUnsafeAccess": true} ]
+  },
+  {
+    "name" : "io.netty.util.internal.shaded.org.jctools.queues.BaseMpscLinkedArrayQueueColdProducerFields",
+    "fields": [ {"name": "producerLimit", "allowUnsafeAccess": true} ]
+  },
+  {
+    "name" : "io.netty.util.internal.shaded.org.jctools.queues.BaseMpscLinkedArrayQueueConsumerFields",
+    "fields": [ {"name": "consumerIndex", "allowUnsafeAccess": true} ]
+  }
+]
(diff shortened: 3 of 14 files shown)


core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
```diff
@@ -0,0 +1,214 @@
+[
```
absurdfarce, line 1: Contents of this file largely taken from https://github.com/absurdfarce/graal-cassandra-driver/blob/master/reflection.json

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
```diff
@@ -0,0 +1,214 @@
...
+    "allDeclaredMethods": true,
+    "allPublicMethods": true
+  },
+  {
+    "name": "com.datastax.oss.driver.internal.core.metrics.DefaultMetricsFactory",
+    "allDeclaredConstructors": true,
+    "allPublicConstructors": true,
+    "allDeclaredMethods": true,
+    "allPublicMethods": true
+  },
```
absurdfarce, line 93: Entries above largely reflect the driver defaults as of 4.11.  Note that this also includes the Dropwizard -> Default MetricsFactory change introduced with JAVA-2927.

If a user chooses to configure any non-standard versions of these components they'll have to include them in a separate Graal reflection config.

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
```diff
@@ -0,0 +1,214 @@
...
+  {
+    "name": "io.netty.channel.DefaultChannelPipeline$TailContext",
+    "allDeclaredConstructors": true,
+    "allPublicConstructors": true,
+    "allDeclaredMethods": true,
+    "allPublicMethods": true
+  },
```
absurdfarce, line 155: A set of Netty dependencies discovered during earlier testing.  I'm kinda suspicious these may not apply anymore... need to vet this out a bit.

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
```diff
@@ -0,0 +1,214 @@
...
+  {
+    "name": "jnr.ffi.provider.jffi.Provider",
+    "allDeclaredConstructors": true,
+    "allPublicConstructors": true,
+    "allDeclaredMethods": true,
+    "allPublicMethods": true
+  },
```
absurdfarce, line 129: Another set of dependencies, this time around JNR, discovered in earlier testing.  I'm not at all sure these still apply especially since JAVA-2663 (https://github.com/datastax/java-driver/pull/1439 specifically).  Need some more testing here as well.

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
```diff
@@ -0,0 +1,214 @@
...
+    "fields": [
+      {"name": "producerLimit", "allowUnsafeAccess": true}
+    ]
+  },
+  {
+    "name" : "io.netty.util.internal.shaded.org.jctools.queues.BaseMpscLinkedArrayQueueConsumerFields",
+    "fields": [
+      {"name": "consumerIndex", "allowUnsafeAccess": true}
+    ]
+  }
```
absurdfarce, line 153: Shaded jctools config from an actual Netty issue (https://github.com/netty/netty/issues/10376)

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
adutra: I analyzed briefly the reference.conf file for 4.12.0 and found the following interfaces and implementing classes:

```
com.datastax.oss.driver.api.core.loadbalancing.LoadBalancingPolicy
    * com.datastax.oss.driver.internal.core.loadbalancing.DefaultLoadBalancingPolicy
    com.datastax.oss.driver.internal.core.loadbalancing.DcInferringLoadBalancingPolicy
    com.datastax.oss.driver.internal.core.loadbalancing.BasicLoadBalancingPolicy

com.datastax.oss.driver.api.core.connection.ReconnectionPolicy
    * com.datastax.oss.driver.internal.core.connection.ExponentialReconnectionPolicy
    com.datastax.oss.driver.internal.core.connection.ConstantReconnectionPolicy

com.datastax.oss.driver.api.core.retry.RetryPolicy
    * com.datastax.oss.driver.internal.core.retry.DefaultRetryPolicy
    com.datastax.oss.driver.internal.core.retry.ConsistencyDowngradingRetryPolicy

com.datastax.oss.driver.api.core.retry.RetryPolicy
    * com.datastax.oss.driver.internal.core.retry.DefaultRetryPolicy
    com.datastax.oss.driver.internal.core.retry.ConsistencyDowngradingRetryPolicy

com.datastax.oss.driver.api.core.specex.SpeculativeExecutionPolicy
    * com.datastax.oss.driver.internal.core.specex.NoSpeculativeExecutionPolicy
    com.datastax.oss.driver.internal.core.specex.ConstantSpeculativeExecutionPolicy

com.datastax.oss.driver.api.core.time.TimestampGenerator
    * com.datastax.oss.driver.internal.core.time.AtomicTimestampGenerator
    com.datastax.oss.driver.internal.core.time.ThreadLocalTimestampGenerator
    com.datastax.oss.driver.internal.core.time.ServerSideTimestampGenerator

com.datastax.oss.driver.api.core.tracker.RequestTracker
    * com.datastax.oss.driver.internal.core.tracker.NoopRequestTracker
    com.datastax.oss.driver.internal.core.tracker.RequestLogger
    com.datastax.dse.driver.internal.core.tracker.MultiplexingRequestTracker

com.datastax.oss.driver.api.core.session.throttling.RequestThrottler
    * com.datastax.oss.driver.internal.core.session.throttling.PassThroughRequestThrottler
    com.datastax.oss.driver.internal.core.session.throttling.RateLimitingRequestThrottler
    com.datastax.oss.driver.internal.core.session.throttling.ConcurrencyLimitingRequestThrottler

com.datastax.oss.driver.api.core.metadata.NodeStateListener
    * com.datastax.oss.driver.internal.core.metadata.NoopNodeStateListener
    com.datastax.oss.driver.api.core.metadata.SafeInitNodeStateListener

com.datastax.oss.driver.api.core.metadata.schema.SchemaChangeListener
    * com.datastax.oss.driver.internal.core.metadata.schema.NoopSchemaChangeListener

com.datastax.oss.driver.api.core.addresstranslation.AddressTranslator
    * com.datastax.oss.driver.internal.core.addresstranslation.PassThroughAddressTranslator
    com.datastax.oss.driver.internal.core.addresstranslation.Ec2MultiRegionAddressTranslator

com.datastax.oss.driver.api.core.auth.AuthProvider
    com.datastax.oss.driver.internal.core.auth.PlainTextAuthProvider
    com.datastax.oss.driver.internal.core.auth.ProgrammaticPlainTextAuthProvider
    com.datastax.dse.driver.internal.core.auth.DseGssApiAuthProvider
    com.datastax.dse.driver.api.core.auth.ProgrammaticDseGssApiAuthProvider

com.datastax.oss.driver.api.core.ssl.SslEngineFactory
    com.datastax.oss.driver.internal.core.ssl.DefaultSslEngineFactory
    com.datastax.oss.driver.api.core.ssl.ProgrammaticSslEngineFactory
    com.datastax.oss.driver.internal.core.ssl.SniSslEngineFactory

com.datastax.oss.driver.internal.core.metrics.MetricsFactory
    * com.datastax.oss.driver.internal.core.metrics.DefaultMetricsFactory
    com.datastax.oss.driver.internal.core.metrics.NoopMetricsFactory
    com.datastax.oss.driver.internal.core.metrics.DropwizardMetricsFactory
    com.datastax.oss.driver.internal.metrics.microprofile.MicroProfileMetricsFactory
    com.datastax.oss.driver.internal.metrics.micrometer.MicrometerMetricsFactory

com.datastax.oss.driver.internal.core.metrics.MetricIdGenerator
    * com.datastax.oss.driver.internal.core.metrics.DefaultMetricIdGenerator
    com.datastax.oss.driver.internal.core.metrics.TaggingMetricIdGenerator
```

Default implementations are marked with an asterisk.

This list brings the following remarks:

1. You are currently including only the default impls. We have discussed this already offline but haven't came to an agreement: should we include all impls? I understand the risks (larger image, warnings during image build).
2. A few interfaces are missing:
  a. `AuthProvider`
  b. `SslEngineFactory`
  c. `MetricIdGenerator` (introduced recently by JAVA-2872 in 4.11.0)
3. I am not sure about these ones; they are all handlers instantiated inside a `ChannelInitializer` inside `ChannelFactory` (the first one being the anonymous `ChannelInitializer` itself) :
  a. `com.datastax.oss.driver.internal.core.channel.ChannelFactory$1`
  b. `com.datastax.oss.driver.internal.core.channel.InFlightHandler`
  c. `com.datastax.oss.driver.internal.core.channel.ProtocolInitHandler`
  d. `com.datastax.oss.driver.internal.core.protocol.FrameDecoder`
  e. `com.datastax.oss.driver.internal.core.protocol.FrameEncoder`
  f. `com.datastax.oss.driver.internal.core.channel.HeartbeatHandler`
4. I am also not sure about the Netty ones below, and especially worried about these getting out of sync when Netty is upgraded:
  a. `io.netty.channel.socket.nio.NioSocketChannel`
  b. `io.netty.buffer.AbstractByteBufAllocator`
  c. `io.netty.util.ReferenceCountUtil`
  d. `io.netty.channel.DefaultChannelPipeline$HeadContext`
  e. `io.netty.channel.DefaultChannelPipeline$TailContext`
  f. `io.netty.util.internal.shaded.org.jctools.queues.MpscArrayQueueProducerIndexField`
  g. `io.netty.util.internal.shaded.org.jctools.queues.MpscArrayQueueProducerLimitField`
  h. `io.netty.util.internal.shaded.org.jctools.queues.MpscArrayQueueConsumerIndexField`
  i. `io.netty.util.internal.shaded.org.jctools.queues.BaseMpscLinkedArrayQueueProducerFields`
  j. `io.netty.util.internal.shaded.org.jctools.queues.BaseMpscLinkedArrayQueueColdProducerFields`
  k. `io.netty.util.internal.shaded.org.jctools.queues.BaseMpscLinkedArrayQueueConsumerFields`

adutra: Nit: `com.datastax.oss.driver.internal.core.channel.ChannelFactory$1` is a generated anonymous class name that could also get out of sync if we introduce another anonymous class in `ChannelFactory`. It might be a good idea to extract that anonymous class to an inner.

adutra:
I think that since #1559 is now linked to JAVA-2950, we could consider that this PR is the "fix" (as in: remaining tweaks after the `DependencyCheck` issue is solved) for JAVA-2940, wdyt?

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/native-image.properties
```diff
@@ -0,0 +1 @@
+Args = -H:IncludeResources=.*reference\\.conf -H:IncludeResources=.*Driver\\.properties -H:DynamicProxyConfigurationResources=${.}/proxy.json -H:ReflectionConfigurationResources=${.}/reflection.json
```
adutra: Just for my own education on GraalVM tooling: when this file is included in a jar, does it get automatically picked by `native-image`? What happens if a user doesn't want to use the default proxy.json and reflection.json files?

adutra: Also, it just occurred to me that we have a Service implementation in driver-core: `DriverBlockHoundIntegration`. This impl is meant to be discovered via the ServiceLoader API. Therefore I think we need to:

1. Declare that implementation in reflection.json: the FQCN is `com.datastax.oss.driver.internal.core.util.concurrent.DriverBlockHoundIntegration`;
2. Also add `-H:IncludeResources=META-INF/services/*.*` to the arg line, in order to preserve the link between service interface and service impl.

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
```diff
@@ -0,0 +1,192 @@
...
+    "allDeclaredMethods": true,
+    "allPublicMethods": true
+  },
+  {
+    "name": "io.netty.channel.socket.nio.NioSocketChannel",
+    "allDeclaredConstructors": true,
+    "allPublicConstructors": true,
+    "allDeclaredMethods": true,
+    "allPublicMethods": true
+  },
```
absurdfarce, line 127: Provided to Netty impl in the driver via NettyOptions, and since instantiation of driver components is based on reflection won't be discovered unless it's explicitly included here.

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
absurdfarce: I'm still working through the contents of this file but a few notes as I go here.

* NioSocketChannel has been removed.  Originally I thought we had to keep it to support the Helidon case but it looks very much like the problem there is a Helidon issue.  More detail on this at https://github.com/datastax/java-driver/pull/1560/commits/ea5258065cb996b6f92cd222c0c7478a54c89213
* The shaded jctools stuff certainly appears to be required, at least for the version of Netty we're using.  See comments below for the relevant Netty issue.  I agree that maintaining this kind of thing is a pain but I'm not sure what other option we have.
* I'm still testing/spelunking the other Netty classes to see what else (if anything) can be removed from this list
* I very much like the idea of abstracting the ChannelFactory anonymous class into an inner class... I'll try to fold that into other work going in soon.

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/native-image.properties
absurdfarce: Re: usage of this file by native-image... yes, it's discovered as part of the build process.  https://www.graalvm.org/reference-manual/native-image/BuildConfiguration/ has a bit more detail.  Under this scheme there isn't a clean way for the user to avoid the specified reflection + proxy configs... so the defaults will always be included in the built native image even if they select custom impls of some of the components.  In the medium- to long-term this could be ameliorated by switching to a Feature impl.

absurdfarce: Re: the ServiceLoader configs... I haven't explicitly tested this but based on https://github.com/oracle/graal/issues/563 it looks like we should be okay

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
absurdfarce: Update: all the Netty classes that aren't associated with the shaded jctools have been removed based on local testing

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
```diff
@@ -0,0 +1,175 @@
...
+  {
+    "name": "io.netty.channel.socket.nio.NioSocketChannel",
+    "methods": [
+      { "name": "<init>", "parameterTypes": [] }
+    ]
+  },
```
absurdfarce, line 121: Without this exclusion:

```
Exception in thread "main" java.lang.IllegalArgumentException: Class NioSocketChannel does not have a public non-arg constructor                                                  
        at io.netty.channel.ReflectiveChannelFactory.<init>(ReflectiveChannelFactory.java:36)                                                                                     
        at io.netty.bootstrap.AbstractBootstrap.channel(AbstractBootstrap.java:110)                                                                                               
        at com.datastax.oss.driver.internal.core.channel.ChannelFactory.connect(ChannelFactory.java:192)                                                                          
        at com.datastax.oss.driver.internal.core.channel.ChannelFactory.connect(ChannelFactory.java:167)                                                                          
        at com.datastax.oss.driver.internal.core.channel.ChannelFactory.connect(ChannelFactory.java:148)                                                                          
        at com.datastax.oss.driver.internal.core.control.ControlConnection$SingleThreaded.connect(ControlConnection.java:363)                                                     
        at com.datastax.oss.driver.internal.core.control.ControlConnection$SingleThreaded.init(ControlConnection.java:303)                                                        
        at com.datastax.oss.driver.internal.core.control.ControlConnection$SingleThreaded.access$1100(ControlConnection.java:243)                                                 
        at com.datastax.oss.driver.internal.core.control.ControlConnection.lambda$init$0(ControlConnection.java:122)                                                              
        at io.netty.util.concurrent.PromiseTask.runTask(PromiseTask.java:98)                                                                                                      
        at io.netty.util.concurrent.PromiseTask.run(PromiseTask.java:106)                                                                                                         
        at io.netty.channel.DefaultEventLoop.run(DefaultEventLoop.java:54)                                                                                                        
        at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989)                                                                           
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)                                                                                              
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)                                                                                  
        at java.lang.Thread.run(Thread.java:748)                                                                                                                                  
        at com.oracle.svm.core.thread.JavaThreads.threadStartRoutine(JavaThreads.java:553)                                                                                        
        at com.oracle.svm.core.posix.thread.PosixJavaThreads.pthreadStartRoutine(PosixJavaThreads.java:192)                                                                       
Caused by: java.lang.NoSuchMethodException: io.netty.channel.socket.nio.NioSocketChannel.<init>()                                                                                 
        at java.lang.Class.getConstructor0(DynamicHub.java:3082)                                                                                                                  
        at java.lang.Class.getConstructor(DynamicHub.java:1825)                                                                                                                   
        at io.netty.channel.ReflectiveChannelFactory.<init>(ReflectiveChannelFactory.java:34)
        ... 17 more
```

Not abundantly clear why this would be; NioSocketChannel very definitely does have a no-arg constructor.  It's possible there's another Graal substitution at play here, although if that's the case I haven't been able to identify where it lives.

Note that Helidon had to address something similar: see https://github.com/oracle/helidon/pull/1722

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
```diff
@@ -0,0 +1,175 @@
...
+  {
+    "name": "io.netty.buffer.AbstractByteBufAllocator",
+    "methods": [
+      { "name": "toLeakAwareBuffer", "parameterTypes": ["io.netty.buffer.ByteBuf"] }
+    ]
+  },
```
absurdfarce, line 125: Without this exclusion:

```
Exception in thread "main" java.lang.ExceptionInInitializerError
        at com.oracle.svm.core.classinitialization.ClassInitializationInfo.initialize(ClassInitializationInfo.java:315)                                                          
        at java.lang.Class.ensureInitialized(DynamicHub.java:553)
        at com.oracle.svm.core.classinitialization.ClassInitializationInfo.initialize(ClassInitializationInfo.java:260)                                                          
        at io.netty.buffer.ByteBufUtil.<clinit>(ByteBufUtil.java:86)
        at com.oracle.svm.core.classinitialization.ClassInitializationInfo.invokeClassInitializer(ClassInitializationInfo.java:375)                                              
        at com.oracle.svm.core.classinitialization.ClassInitializationInfo.initialize(ClassInitializationInfo.java:295)                                                          
        at io.netty.buffer.ByteBufAllocator.<clinit>(ByteBufAllocator.java:24)
        at com.oracle.svm.core.classinitialization.ClassInitializationInfo.invokeClassInitializer(ClassInitializationInfo.java:375)                                              
        at com.oracle.svm.core.classinitialization.ClassInitializationInfo.initialize(ClassInitializationInfo.java:295)                                                          
        at com.datastax.oss.driver.internal.core.context.DefaultNettyOptions.allocator(DefaultNettyOptions.java:144)                                                             
        at com.datastax.oss.driver.internal.core.channel.ChannelFactory.connect(ChannelFactory.java:193)                                                                         
        at com.datastax.oss.driver.internal.core.channel.ChannelFactory.connect(ChannelFactory.java:167)                                                                         
        at com.datastax.oss.driver.internal.core.channel.ChannelFactory.connect(ChannelFactory.java:148)                                                                         
        at com.datastax.oss.driver.internal.core.control.ControlConnection$SingleThreaded.connect(ControlConnection.java:363)                                                    
        at com.datastax.oss.driver.internal.core.control.ControlConnection$SingleThreaded.init(ControlConnection.java:303)                                                       
        at com.datastax.oss.driver.internal.core.control.ControlConnection$SingleThreaded.access$1100(ControlConnection.java:243)                                                
        at com.datastax.oss.driver.internal.core.control.ControlConnection.lambda$init$0(ControlConnection.java:122)                                                             
        at io.netty.util.concurrent.PromiseTask.runTask(PromiseTask.java:98)
        at io.netty.util.concurrent.PromiseTask.run(PromiseTask.java:106)
        at io.netty.channel.DefaultEventLoop.run(DefaultEventLoop.java:54)
        at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989)                                                                          
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)                                                                                 
        at java.lang.Thread.run(Thread.java:748)
        at com.oracle.svm.core.thread.JavaThreads.threadStartRoutine(JavaThreads.java:553)                                                                                       
        at com.oracle.svm.core.posix.thread.PosixJavaThreads.pthreadStartRoutine(PosixJavaThreads.java:192)                                                                      
Caused by: java.lang.IllegalArgumentException: Can't find '[toLeakAwareBuffer]' in io.netty.buffer.AbstractByteBufAllocator                                                      
        at io.netty.util.ResourceLeakDetector.addExclusions(ResourceLeakDetector.java:576)                                                                                       
        at io.netty.buffer.AbstractByteBufAllocator.<clinit>(AbstractByteBufAllocator.java:36)                                                                                   
        at com.oracle.svm.core.classinitialization.ClassInitializationInfo.invokeClassInitializer(ClassInitializationInfo.java:375)                                              
        at com.oracle.svm.core.classinitialization.ClassInitializationInfo.initialize(ClassInitializationInfo.java:295)                                                          
        ... 25 more
```

Haven't dug into this one much to see if there's anything we can do here.

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
```diff
@@ -0,0 +1,175 @@
...
+  {
+    "name": "io.netty.util.ReferenceCountUtil",
+    "methods": [
+        { "name": "touch", "parameterTypes": ["java.lang.Object", "java.lang.Object"] }
+    ]
+  },
```
absurdfarce, line 129: Without this exclusion:

```
WARNING: An exception was thrown by com.datastax.oss.driver.internal.core.channel.ConnectInitHandler$$Lambda$df350d7c55c0f44e72fb6470156ed35696eba329.operationComplete() [3/1826]
java.lang.ExceptionInInitializerError                                                                                                                                             
        at com.oracle.svm.core.classinitialization.ClassInitializationInfo.initialize(ClassInitializationInfo.java:315)                                                           
        at io.netty.channel.DefaultChannelPipeline.touch(DefaultChannelPipeline.java:116)                                                                                         
        at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:786)
        at io.netty.channel.AbstractChannelHandlerContext.writeAndFlush(AbstractChannelHandlerContext.java:758)
        at io.netty.channel.AbstractChannelHandlerContext.writeAndFlush(AbstractChannelHandlerContext.java:808)
        at io.netty.channel.DefaultChannelPipeline.writeAndFlush(DefaultChannelPipeline.java:1025)
        at io.netty.channel.AbstractChannel.writeAndFlush(AbstractChannel.java:294)
        at com.datastax.oss.driver.internal.core.channel.ChannelHandlerRequest.send(ChannelHandlerRequest.java:75)
        at com.datastax.oss.driver.internal.core.channel.ProtocolInitHandler$InitRequest.send(ProtocolInitHandler.java:193)
        at com.datastax.oss.driver.internal.core.channel.ProtocolInitHandler.onRealConnect(ProtocolInitHandler.java:124)
        at com.datastax.oss.driver.internal.core.channel.ConnectInitHandler.lambda$connect$0(ConnectInitHandler.java:57)
        at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:578)
        at io.netty.util.concurrent.DefaultPromise.notifyListeners0(DefaultPromise.java:571)
        at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:550)
        at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:491)
        at io.netty.util.concurrent.DefaultPromise.setValue0(DefaultPromise.java:616)
        at io.netty.util.concurrent.DefaultPromise.setSuccess0(DefaultPromise.java:605)
        at io.netty.util.concurrent.DefaultPromise.trySuccess(DefaultPromise.java:104)
        at io.netty.channel.DefaultChannelPromise.trySuccess(DefaultChannelPromise.java:84)
        at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.fulfillConnectPromise(AbstractNioChannel.java:300)
        at io.netty.channel.nio.AbstractNioChannel$AbstractNioUnsafe.finishConnect(AbstractNioChannel.java:335)
        at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:707)
        at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:655)
        at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:581)
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:493)
        at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989)
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.lang.Thread.run(Thread.java:748)
        at com.oracle.svm.core.thread.JavaThreads.threadStartRoutine(JavaThreads.java:553)
        at com.oracle.svm.core.posix.thread.PosixJavaThreads.pthreadStartRoutine(PosixJavaThreads.java:192)
Caused by: java.lang.IllegalArgumentException: Can't find '[touch]' in io.netty.util.ReferenceCountUtil
        at io.netty.util.ResourceLeakDetector.addExclusions(ResourceLeakDetector.java:576)
        at io.netty.util.ReferenceCountUtil.<clinit>(ReferenceCountUtil.java:30)
        at com.oracle.svm.core.classinitialization.ClassInitializationInfo.invokeClassInitializer(ClassInitializationInfo.java:375)
        at com.oracle.svm.core.classinitialization.ClassInitializationInfo.initialize(ClassInitializationInfo.java:295)
```

Also haven't dug into the underlying causes for this one yet.

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ChannelFactory.java
```diff
@@ -272,98 +272,121 @@ private void connect(
...
-                  DefaultDriverOption.CONNECTION_MAX_REQUESTS.getPath(),
-                  maxRequestsPerConnection / 4);
-            }
-            maxOrphanRequests = maxRequestsPerConnection / 4;
-          }
+    return new ChannelFactoryInitializer(
+        endPoint, protocolVersion, options, nodeMetricUpdater, resultFuture);
+  };
+
+  class ChannelFactoryInitializer extends ChannelInitializer<Channel> {
```
absurdfarce, line 279: TODO: should be okay to make this a static class

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
absurdfarce: Update: NioSocketChannel and friends are back, although in stripped down form.  Subsequent testing identified an unexpected reflection config which was in play during earlier testing; that config was effectively masking a trio of Netty-related issues from driver code.  These have now been resolved.

absurdfarce: Update: re: the classes from the channel initializer... I'm not sure why these are present either.  I would expect static analysis to come across these pretty readily.  Testing with the current working version of this PR has confirmed that the removal of these entries in a reflection config doesn't have any adverse effects on our test cases.  I suppose this _could be_ due to the refactoring of the ChannelFactory above but I really don't see how... I'd expect similar analysis for both the before and after code.

absurdfarce: Only thing really left here were the three potential new candidates for inclusion: impls for AuthProvider, SslEngineFactory and MetricIdGenerator.  These seem a bit more complicated to me; they all come into play for optional features of the driver but aren't used in the base no-auth/no-ssl case.  I could perhaps see different handling for MetricIdGenerator since we at least have a default defined for that in reference.conf... but that isn't true for AuthProvider and SslEngineFactory.

We already know that we'll have to provide guidance for users who wish to use non-standard components + Graal.  I'm wondering if we shouldn't take a similar line for these three components as well; if you want to use them you have to do some additional manual customizations, specifically create your own local reflection.json file to supplement the one the driver provides (via this PR).

@adutra whaddya think?

adutra: Thanks for the latest changes regarding ChannelFactory and Netty, these areas are looking much better now.

Now back to the question of what classes to include here, only defaults vs all driver built-ins. I am still struggling with 2 fundamental questions:

1. Who is this file going to be useful for?
2. What contents would make it the most useful for most people?

For 1), I can think of two personas: end users, and maintainers of microservice frameworks. Let's see now how to get at 2) for these two cases:

* **As an end user building my native image myself.** 
    * Suppose that only default impls are declared here. If I bootstrap my application using only driver defaults, the native image would build without issues thanks to this file. However if I later on decide to use, say, `RateLimitingRequestThrottler` instead of `PassThroughRequestThrottler` by declaring this in an application.conf file, or using the programmatic config builder, then all of a sudden, the native image build will break. This isn't great in terms of usability because the config change seems unrelated to the build failure. Without proper guidance/documentation, I bet that user would struggle to fix this. (That said, I admit that this user would also need to register their application.conf file if they are using one... I don't think we can do this for them).
    * Suppose now that all built-in impls are declared here. The above config change wouldn't break the native image build (provided that application.conf is registered if used). 
    * I therefore think that for this persona, having all built-in impls declared here is better.
    * Obviously, if that user later on changes `RateLimitingRequestThrottler` for a custom impl of `RequestThrottler`, the native image build will fail again if the custom impl is not registered. I would argue that such a user is a more advanced one, and by the time they start implementing driver interfaces, they should be aware of the implications.
    * Also obviously, the more you declare for reflection, the bigger your native image will be. I think we are trading complexity against disk space here. I'd argue that disk space is cheaper :-)
   
* **As a microservices framework maintainer (Quarkus, Micronaut, Helidon)**. 
    * If only default impls are declared here, but if the framework allows some of these defaults to be replaced (e.g. Quarkus allows the metrics factory to be replaced), then this file is kind of pointless, as the framework will have to provide at least a complement for its contents, if not a complete replacement. This is precisely what Quarkus is doing already, e.g for [metrics](https://github.com/datastax/cassandra-quarkus/blob/main/deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java#L247).
    * Suppose now that all built-in impls are declared here. A config change such as setting the metrics factory to Micrometer for instance, made through the framework config mechanism or through a driver-specific one, wouldn't break the native image build. 
    * So again, I think that for this persona, having all built-in impls declared here is better. The framework maintainers in question wouldn't need to provide any additional Graal customization – unless they want to because they want a more fine-grained control over the native image contents (like we do in Quarkus for metrics – we'll probably keep doing it even after this change is merged).
    * Such microservices frameworks generally discourage users from bypassing their config mechanisms when configuring the driver. E.g. the usage of an application.conf file in Quarkus is possible, although discouraged. If such a file is included in a Quarkus app, the user would need to register it manually. And if such a file contains a custom impl of a driver interface, that would have to be registered too. Again, we are here facing a rather advanced user trying to go beyond what the framework is able to provide. I think we can safely consider this guy as a user that know what he's doing.

So to summarize my POV: including everything we can here would satisfy a broader range of users and require less work from them. This seems especially true for Micronaut and Helidon users.

adutra:
I've been playing with this today and I have a few additional remarks:

## LZ4 Compression

I am trying to validate compression with lz4. It seems that it is not working:

```
java.lang.IllegalStateException: Could not find the LZ4 library on the classpath (the driver declares it as an optional dependency, so you need to declare it explicitly)
        at com.datastax.oss.driver.internal.core.protocol.Lz4Compressor.<init>(Lz4Compressor.java:50)
        at com.datastax.oss.driver.internal.core.protocol.Lz4Compressor.<init>(Lz4Compressor.java:39)
        at com.datastax.oss.driver.internal.core.protocol.BuiltInCompressors.newInstance(BuiltInCompressors.java:44)
```

Indeed, even if `BuiltInCompressors` is properly substituted, LZ4 itself does some more reflection to find out the right classes to instantiate. This isn't working. I had to add a whole bunch of exceptions in reflection.json:

```
  {
    "name" : "net.jpountz.lz4.LZ4Compressor",
    "allDeclaredConstructors": true,
    "allPublicConstructors": true,
    "allDeclaredMethods": true,
    "allPublicMethods": true
  },
  {
    "name" : "net.jpountz.lz4.LZ4JNICompressor",
    "allDeclaredConstructors": true,
    "allPublicConstructors": true,
    "allDeclaredMethods": true,
    "allPublicMethods": true,
    "allPublicFields": true
  },
  {
    "name" : "net.jpountz.lz4.LZ4JavaSafeCompressor",
    "allDeclaredConstructors": true,
    "allPublicConstructors": true,
    "allDeclaredMethods": true,
    "allPublicMethods": true,
    "allPublicFields": true
  },
  {
    "name" : "net.jpountz.lz4.LZ4JavaUnsafeCompressor",
    "allDeclaredConstructors": true,
    "allPublicConstructors": true,
    "allDeclaredMethods": true,
    "allPublicMethods": true,
    "allPublicFields": true
  },
  {
    "name" : "net.jpountz.lz4.LZ4HCJavaSafeCompressor",
    "allDeclaredConstructors": true,
    "allPublicConstructors": true,
    "allDeclaredMethods": true,
    "allPublicMethods": true,
    "allPublicFields": true
  },
  {
    "name" : "net.jpountz.lz4.LZ4HCJavaUnsafeCompressor",
    "allDeclaredConstructors": true,
    "allPublicConstructors": true,
    "allDeclaredMethods": true,
    "allPublicMethods": true,
    "allPublicFields": true
  },
  {
    "name" : "net.jpountz.lz4.LZ4JavaSafeSafeDecompressor",
    "allDeclaredConstructors": true,
    "allPublicConstructors": true,
    "allDeclaredMethods": true,
    "allPublicMethods": true,
    "allPublicFields": true
  },
  {
    "name" : "net.jpountz.lz4.LZ4JavaSafeFastDecompressor",
    "allDeclaredConstructors": true,
    "allPublicConstructors": true,
    "allDeclaredMethods": true,
    "allPublicMethods": true,
    "allPublicFields": true
  },
  {
    "name" : "net.jpountz.lz4.LZ4JavaUnsafeSafeDecompressor",
    "allDeclaredConstructors": true,
    "allPublicConstructors": true,
    "allDeclaredMethods": true,
    "allPublicMethods": true,
    "allPublicFields": true
  },
  {
    "name" : "net.jpountz.lz4.LZ4JavaUnsafeFastDecompressor",
    "allDeclaredConstructors": true,
    "allPublicConstructors": true,
    "allDeclaredMethods": true,
    "allPublicMethods": true,
    "allPublicFields": true
  }
```

We need all of the above unfortunately, including the abstract superclass `LZ4Compressor`  :-(

## TypeVariable

There is something fishy with `TypeVariable`, the proxy config is not working it seems:

```
Error: com.oracle.svm.hosted.substitute.DeletedElementException: Unsupported method java.lang.Class.getConstantPool() is reachable: The declaring class of this element has been substituted, but this element is not present in the substitution class
To diagnose the issue, you can add the option --report-unsupported-elements-at-runtime. The unsupported element is then reported at run time when it is accessed the first time.
Trace: 
        at parsing java.lang.System$2.getConstantPool(System.java:1230)
Call path from entry point to java.lang.System$2.getConstantPool(Class): 
        at java.lang.System$2.getConstantPool(System.java:1230)
        at sun.reflect.annotation.TypeAnnotationParser.parseAllTypeAnnotations(TypeAnnotationParser.java:323)
        at sun.reflect.annotation.TypeAnnotationParser.fetchBounds(TypeAnnotationParser.java:299)
        at sun.reflect.annotation.TypeAnnotationParser.parseAnnotatedBounds(TypeAnnotationParser.java:244)
        at sun.reflect.annotation.TypeAnnotationParser.parseAnnotatedBounds(TypeAnnotationParser.java:237)
        at sun.reflect.generics.reflectiveObjects.TypeVariableImpl.getAnnotatedBounds(TypeVariableImpl.java:241)
        at com.oracle.svm.reflect.TypeVariable_getAnnotatedBounds_88ea462da2b17ea45028db307519aaeeec9a4036_58.invoke(Unknown Source)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at com.datastax.oss.driver.shaded.guava.common.reflect.Types$TypeVariableInvocationHandler.invoke(Types.java:387)
```

I still need to investigate.

## ServiceLoader

I manually validated that this is working by running the following code:

```
    ServiceLoader<BlockHoundIntegration> serviceLoader = ServiceLoader.load(BlockHoundIntegration.class);
    boolean driverBlockHoundIntegrationFound = StreamSupport.stream(serviceLoader.spliterator(),
        false)
        .map(integration -> integration.getClass().getSimpleName())
        .anyMatch(name -> name.contains("DriverBlockHoundIntegration"));
    System.out.println("DriverBlockHoundIntegration registered = " + driverBlockHoundIntegrationFound);
```

adutra:
Also @absurdfarce would you mind rebasing this PR on top of 4.11.x? Otherwise we are going to hit JAVA-2950 during tests.

adutra:
Regarding lz4, what's even worse is that we cannot unconditionally add those LZ4 classes to reflection.json. If LZ4 is not present during image build, the build will fail. I think we need something like an `Lz4Feature` :-(

adutra:
> Regarding lz4, what's even worse is that we cannot unconditionally add those LZ4 classes to reflection.json. If LZ4 is not present during image build, the build will fail. I think we need something like an `Lz4Feature` :-(

Update: in fact the build succeeds, but generates warnings:

```
WARNING: Could not resolve net.jpountz.lz4.LZ4Compressor for reflection configuration. Reason: java.lang.ClassNotFoundException: net.jpountz.lz4.LZ4Compressor.
WARNING: Could not resolve net.jpountz.lz4.LZ4JNICompressor for reflection configuration. Reason: java.lang.ClassNotFoundException: net.jpountz.lz4.LZ4JNICompressor.
WARNING: Could not resolve net.jpountz.lz4.LZ4JavaSafeCompressor for reflection configuration. Reason: java.lang.ClassNotFoundException: net.jpountz.lz4.LZ4JavaSafeCompressor.
WARNING: Could not resolve net.jpountz.lz4.LZ4JavaUnsafeCompressor for reflection configuration. Reason: java.lang.ClassNotFoundException: net.jpountz.lz4.LZ4JavaUnsafeCompressor.
WARNING: Could not resolve net.jpountz.lz4.LZ4HCJavaSafeCompressor for reflection configuration. Reason: java.lang.ClassNotFoundException: net.jpountz.lz4.LZ4HCJavaSafeCompressor.
WARNING: Could not resolve net.jpountz.lz4.LZ4HCJavaUnsafeCompressor for reflection configuration. Reason: java.lang.ClassNotFoundException: net.jpountz.lz4.LZ4HCJavaUnsafeCompressor.
WARNING: Could not resolve net.jpountz.lz4.LZ4JavaSafeSafeDecompressor for reflection configuration. Reason: java.lang.ClassNotFoundException: net.jpountz.lz4.LZ4JavaSafeSafeDecompressor.
WARNING: Could not resolve net.jpountz.lz4.LZ4JavaSafeFastDecompressor for reflection configuration. Reason: java.lang.ClassNotFoundException: net.jpountz.lz4.LZ4JavaSafeFastDecompressor.
WARNING: Could not resolve net.jpountz.lz4.LZ4JavaUnsafeSafeDecompressor for reflection configuration. Reason: java.lang.ClassNotFoundException: net.jpountz.lz4.LZ4JavaUnsafeSafeDecompressor.
WARNING: Could not resolve net.jpountz.lz4.LZ4JavaUnsafeFastDecompressor for reflection configuration. Reason: java.lang.ClassNotFoundException: net.jpountz.lz4.LZ4JavaUnsafeFastDecompressor.
```

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
absurdfarce: I'm sympathetic to the position you lay out @adutra, but I think I still come down on the other side.  I'll stick with your framework because I think it's a useful way of formulating the questions.

My recommended approach is the following:

* Provide a minimal config embedded in the JAR that only includes the default instances in a reflection config
* Any users wishing to include custom component impls (or other config customizations) must supply additional configs to support such work.  We would supply docs to communicate this and provide some guidance on how to do so.

My rationale runs as follows:

* Native image size is a factor for Graal users; I've seen it mentioned as a concern on any number of tickets/conversations around the ecosystem
  * If we include all the classes in our reflection config there's no way for users to _not_ include those classes in the image; subsequent config files can't unselect a class which was previously selected for inclusion.  The only way to undo it would be to edit the config directly and update the config in the JAR.
* The number of "end users" (again using the language of your scenarios) is fairly small.  The intersection of those users with users who wish to leverage custom driver components is either empty or very nearly empty (at least for the moment).  As a result the default configs allow us to address all or nearly all "end user" cases without including the additional classes (and taking on the additional image size)
* Shifting to frameworks, these would have to take additional steps to properly configure the drivers if a users selects a custom feature at a framework level.  As you point out we do this in the Quarkus plugin, but I'd argue that's exactly the right place to do it.  
  * Frameworks already have built-in mechanisms for including additional resources in the native image build
  * There doesn't seem to be a clear argument for _also_ implementing a similar feature at the driver level itself, especially if we grant the argument (outlined above) that the population of "end users" who wish to leverage such a similar feature at a driver level is vanishingly small.
* In the medium-to-longer term we can more thoroughly investigate the idea of adding a feature which will perform the selective loading of dependencies at a driver level
  * But I would like to see additional feedback from the "end users" in your scenario saying this is useful before we add the additional complexity here

In summary, I guess I'm thinking of the following:

* "Always add" the smallest amount we can to maximize functionality while keeping the image size small
* Work with framework authors to make sure the framework case can leverage the infrastructure it already has to do the right job
* Revise our plan if we get feedback from actual users saying our assumptions above are flawed

adutra:
More on the `TypeVariable` error: It seems related to the specific GraalVM version I'm using.

* graalvm-ce-java8-21.0.0.2 : DOES NOT WORK, spits that weird error.
* graalvm-ce-java11-21.2.0 : works
* graalvm-ee-java8-21.2.0: works

I guess we can consider that a bug in that GraalVM version and shovel it under the rug.

core/src/main/resources/META-INF/native-image/com.datastax.oss/java-driver-core/reflection.json
adutra: I'm also very sympathetic to your position @absurdfarce :-) And I could easily let myself convince to adopt your suggestions.

But in this case I think we should go even further: why even provide a minimal reflection config? 

* As you rightly pointed out, "native image size is a factor for Graal users". Which means that even the minimal config could add some clutter to the native image. E.g. if the request throttler is changed, the default one would still be registered for reflection, and occupy some space for nothing.
* "[Frameworks] would have to take additional steps to properly configure the drivers if a users selects a custom feature at a framework level": yes. And that's the core of the question. If we go down your route, how is this minimal config going to be useful to, say, Iván López and his struggle to make Micronaut work seamlessly with the Java driver? He would still have to implement for Micronaut a fair amount of the logic that we put in the Quarkus extension. The goal of this ticket was to give Iván a clean base: without any further customization on his side, the driver would still be guaranteed to work and generate a valid native image – even if not the most optimized one.
* And let's not forget the case where a user of a framework is bypassing its configuration mechanism and configuring the driver directly. This is the case, for instance, of all DSE users using Quarkus currently. How is the framework going to detect the bits of configuration being done outside of its realm? If it cannot detect those bits, how is reflection going to be configured, by the end users themselves?

In summary, I can totally see 2 valid options: register for reflection everything that ships with the driver, or nothing at all. But registering just the defaults does NOT seem to provide enough added-value to justify the inclusion fo these configs in the driver jar.

The "nothing at all" option is actually very appealing. If for nothing else, it has the advantage of letting the user decide what they need to include. This is clearly better than the "include all" option at least under one aspect: once some class is registered for reflection, there is no way to "un-register" it. 

Why don't we go down that route then? In this case the work to do here could be restricted to just:

* Declaring a few resources (reference.conf, Driver.properties);
* Declaring the TypeVariable proxy;
* And documenting the reflection side of things: how to properly declare classes for reflection depending on what is going to be used. But no class would be registered by default.
* Additionally, the documentation could include a giant "reflection.json" config file containing all the impls that ship with the driver. That file could serve as a starting point for users trying to create their own file.

Wdyt?

adutra: For future reference: after feedback from the Micronaut team, we decided to go with the option of providing automatic registration for reflection of all driver built-in impls.

adutra (approved):

PR #1561, opened 2021-07-08
JAVA-2950: Graal-friendly DependencyCheck impl, 411x

absurdfarce:
Copied from https://github.com/datastax/java-driver/pull/1559:

As of JAVA-2917 the usage of substitutions for Graal native image builds was modified to use DependencyCheck to determine whether or not the substitution should be used.  DependencyCheck leverages driver.internal.core.util.Reflection which in turn utilizes a static slf4j Logger to record what it does.  The reliance on the Logger introduced several unexpected build-time initialization errors as documented in JAVA-2940.

This PR sketches a way to resolve that problem.  There's some subtlety here, however.  We certainly want to keep a single representation of the dependencies themselves.  DependencyCheck used to perform that role, but it intertwined that logic with checks for the presence of those dependencies via reflection.  And as discussed above this doesn't work cleanly in all cases.  Ideally we would just implement different methods for checking for these dependencies and then use the appropriate method in each case.

This PR decouples these roles:

* Dependency now serves as the canonical representation for a logical dependency
* There are two methods for checking for dependencies.  The default impl still uses d.i.c.u.Reflection (to benefit from the nice logging it offers).  The Graal-specific version just falls back on Class.forName() directly.  This makes it very similar to the Reflection-based impl in the end, although we do lose out on some of the niceties

Local testing has confirmed that this resolves the errors referenced in JAVA-2940


absurdfarce:
This PR is a 4.11.x-friendly version of https://github.com/datastax/java-driver/pull/1559

core/src/main/java/com/datastax/oss/driver/internal/core/util/DefaultDependencyChecker.java
```diff
@@ -0,0 +1,59 @@
...
+  public static boolean isPresent(Dependency dependency) {
+    try {
+      return CACHE.computeIfAbsent(
+          dependency,
+          (dep) -> {
+            for (String classNameToTest : dependency.classes()) {
+              // Always use the driver class loader, assuming that the driver classes and
+              // the dependency classes are either being loaded by the same class loader,
+              // or – as in OSGi deployments – by two distinct, but compatible class loaders.
+              if (Reflection.loadClass(null, classNameToTest) == null) {
```
absurdfarce, line 48: Preserving comment from @adutra on the original PR:

> I wonder: can't we get rid of this class and use only the other one? My thinking is that calling
> Reflection.loadClass(null, classNameToTest) effectively boils down to Class.forName(classNameToTest) since 
> the CL is null.

This is a fairly significant point and I wanted to make sure it was discussed and addressed fully.

I definitely considered this, but in the end I thought it was worthwhile to keep this usage here.  My argument goes as follows: this checker is used in a number of runtime checks in normal non-Graal operations (the Graal case rules them out via substitutions).  In those cases I'd argue we _want_ to use the Reflection class; yes, the underlying mechanism resolves down to Class.forName() in that case as well but Reflection.loadClass() does a lot more logging about what's going on... and I could see that information being very useful when trying to debug runtime errors.

I _definitely_ see the argument for moving to a single implementation; believe me, I was very tempted.  But in the end I thought preserving the solid logging that is the status quo was more important.  I'm certainly open to argument if you still disagree, however.

adutra: Hmm I think you are right here. Debugging class loading issues is sometimes tricky and having some logs to look at can definitely be a life saver. Let's keep the 2 impls.

changelog/README.md
```diff
@@ -2,6 +2,10 @@
 
 <!-- Note: contrary to 3.x, insert new entries *first* in their section -->
 
+### 4.11.3
```
adutra:
```suggestion
### 4.11.3 (in progress)
```

adutra (approved):

PR #1562, opened 2021-07-13
2490 Sketch of a Graal feature to support automated resource loading based on the Typesafe config

absurdfarce:


core/src/main/java/com/datastax/oss/driver/internal/core/graal/DefaultFeature.java
```diff
@@ -0,0 +1,122 @@
...
+          try {
+            return Optional.of(Class.forName(qualifiedClassName));
+          } catch (ClassNotFoundException cnfe) {
+          }
+        }
+        return Optional.empty();
+      }
+    } catch (LinkageError | Exception e) {
+      return Optional.empty();
+    }
```
absurdfarce, line 80: We wind up re-implementing some of the bits here from Reflection.java but I don't see an obvious way around that.  The config is written to assume the presence of this logic when processing entries (the default for LBP is literally "DefaultLoadBalancingPolicy" under the assumption the package stuff will clean it up at instantiation time) so we're left with some redundancy.

It would be nice to consolidate this into a single impl used by both... but there are other pressing matters to work on at the moment.

core/src/main/java/com/datastax/oss/driver/internal/core/graal/DefaultFeature.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  @Override
+  public void beforeAnalysis(Feature.BeforeAnalysisAccess access) {
+
+    /* Make the Typesafe classes we need to do our work available at build-time */
+    RuntimeClassInitialization.initializeAtBuildTime("com.typesafe.config.impl");
+
+    /* slf4j logger class, used by the parent class of the default LBP (and probably plenty more) */
+    RuntimeClassInitialization.initializeAtBuildTime("org.slf4j.LoggerFactory");
```
absurdfarce, line 113: We don't have to actually instantiate the classes but we do wind up loading them here, so static initializers come into play.  Superclasses must also be resolved, so in this case LoggerFactory is actually used in BasicLoadBalancingPolicy, the superclass of DefaultLoadBalancingPolicy.

Like the Typesafe packages above the goal is to ultimately replace this with something more dynamic; this is just a placeholder to keep things moving for now.

core/src/main/java/com/datastax/oss/driver/internal/core/graal/DefaultFeature.java
```diff
@@ -0,0 +1,122 @@
...
+                RuntimeReflection.register(c);
+              }
+            });
+  }
+
+  @Override
+  public void beforeAnalysis(Feature.BeforeAnalysisAccess access) {
+
+    /* Make the Typesafe classes we need to do our work available at build-time */
+    RuntimeClassInitialization.initializeAtBuildTime("com.typesafe.config.impl");
```
absurdfarce, line 110: We use the Typesafe classes to read the config above and since this feature executes at build time those classes must be available for build-time initialization.

This is just a placeholder for now until we implement a more generic way for dealing with this problem.

core/src/main/java/com/datastax/oss/driver/internal/core/graal/DefaultFeature.java
```diff
@@ -0,0 +1,122 @@
...
+              RuntimeReflection.register(clz);
+              for (Method m : clz.getMethods()) {
+                RuntimeReflection.register(m);
+              }
+              for (Field f : clz.getFields()) {
+                RuntimeReflection.register(f);
+              }
+              for (Constructor c : clz.getConstructors()) {
+                RuntimeReflection.register(c);
+              }
```
absurdfarce, line 102: Note that the inclusion of both constructors here means we don't have to worry about whether the profile or non-profile constructor will be used.  We _do_ still need to add support for profiles when determining what values to return from the config.

core/src/main/java/com/datastax/oss/driver/internal/core/graal/DefaultFeature.java
```diff
@@ -0,0 +1,122 @@
...
+
+  public Config buildConfig() {
+
+    /* Copied from DefaultDriverConfigLoader, including the DEFAULT_ROOT_PATH value there as the literal
+    "datastax-java-driver" below.  Can't reference the class directly here since that will introduce
+    a requirement to specify it as a build-time dependency (as well as the slf4j logger it creates
+    via static init */
+    return ConfigFactory.defaultOverrides()
+        .withFallback(ConfigFactory.defaultApplication())
+        .withFallback(ConfigFactory.defaultReference(CqlSession.class.getClassLoader()))
```
adutra, line 47: I _think_ it would be safer to use `access.getApplicationClassLoader()` as the right CL to locate both application and reference files.

The reason why we use `CqlSession.class.getClassLoader()` has to do with OSGi class loading issues. Here, we are at build-time, trying to generate a native image; and I bet that the only CL available is `access.getApplicationClassLoader()`.

core/src/main/java/com/datastax/oss/driver/internal/core/graal/DefaultFeature.java
adutra: Gotcha. In fact, this class should _probably_ start by detecting whether TypeSafe is available or not. If it is, then it should go on and load the config; if it is not, it could do two things IMO:

1. Nothing, and return immediately.
2. Resort to a best-effort thing and register for reflection a predefined set of built-in impls (the ones being discussed [here](https://github.com/datastax/java-driver/pull/1560/files#r665605797)).

Wdyt?

core/src/main/java/com/datastax/oss/driver/internal/core/graal/DefaultFeature.java
```diff
@@ -0,0 +1,122 @@
...
+        .resolve()
+        .getConfig("datastax-java-driver");
+  }
+
+  public Optional<String> tryString(Config config, DefaultDriverOption option) {
```
adutra, line 52: JAVA-2951 will most likely introduce options that are _lists of classes_. I suggest that we start looking for lists of strings here as well.

core/src/main/java/com/datastax/oss/driver/internal/core/graal/DefaultFeature.java
adutra: Ouch. I naively thought that once you register some class for reflection, imported classes would be automatically detected.

adutra:
Closing for now, keeping the branch for future reference.

PR #1567, opened 2021-08-11
Updated log message to inform about incorrect CL for read repair (3.x)

AnkitBarsainyaPSL:
During read repair failure, incorrect CL is being generated which needs correction.

During read repair failure, incorrect CL is being generated. To rectify this we're adding additional clarification to not refer to CL logged in the error message in case of Read Repair Failure.


adutra (approved):

adutra:
Heads up that you need to format the file with `mvn fmt:format`.

See https://github.com/datastax/java-driver/blob/3.x/CONTRIBUTING.md.

PR #1568, opened 2021-08-11
Updated log message to inform about incorrect CL for read repair (3.x)

AnkitBarsainyaPSL:
What was the issue?
During read repair failure, incorrect CL is being generated which needs correction.

What was fixed?
During read repair failure, incorrect CL is being generated. To rectify this we're adding additional clarification to not refer to CL logged in the error message in case of Read Repair Failure.

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/exceptions/ReadTimeoutException.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/exceptions/ReadTimeoutException.java
@@ -42,7 +42,8 @@ public class ReadTimeoutException extends QueryConsistencyException {
     super(
         endPoint,
         String.format(
-            "Cassandra timeout during read query at consistency %s (%s)",
+            "Cassandra timeout during read query at consistency %s (%s). "
+                + "In case this was generated during read repair, the consistency level is not representative of the actual consistency.",
             consistency, formatDetails(received, required, dataPresent)),
         consistency,
         received,


adutra (approved):

driver-core/src/main/java/com/datastax/driver/core/exceptions/ReadTimeoutException.java
```diff
@@ -42,8 +42,11 @@ public ReadTimeoutException(
     super(
         endPoint,
         String.format(
+            "Cassandra timeout during read query at consistency %s (%s). "
+                + "In case this was generated during read repair, the consistency level is not representative of the actual consistency.",
             "Cassandra timeout during read query at consistency %s (%s)",
```
justinchuch: this is a redundant line. I believe this line is supposed to be replaced by the new changes.

justinchuch (changes requested):

driver-core/src/main/java/com/datastax/driver/core/exceptions/ReadTimeoutException.java
```diff
@@ -42,8 +42,11 @@ public ReadTimeoutException(
...
-            consistency, formatDetails(received, required, dataPresent)),
+            consistency,
```
berndocklin: another comment is that this is an unnecessary formatting change not related to the problem we want to fix. please remove.

driver-core/src/main/java/com/datastax/driver/core/exceptions/ReadTimeoutException.java
AnkitBarsainyaPSL: Thanks for pointing out. Probably skipped while switching/raising multiple PR on this. Taken care now.

driver-core/src/main/java/com/datastax/driver/core/exceptions/ReadTimeoutException.java
AnkitBarsainyaPSL: Thanks. Taken care of now.

adutra (approved):

justinchuch (approved):

justinchuch:
maybe squash the commits into one after approval?

PR #1570, opened 2021-08-15
JAVA-2959 Descriptive error when a Node has no available connections

akhaku:
For cases in which there are no connections available to
send requests to a Node in the query plan, collect the error
rather than silently skipping over the node. The error will
be thrown as part of an AllNodesFailedException if all nodes
fail.
This can happen when we've saturated the max in-flight
requests across all nodes or when the request is directed to
a particular node and it has no connections available (or
all its connections are saturated).
Note that in the latter case we used to throw a
NoNodeAvailableException but we now throw
AllNodesFailedException.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -2,6 +2,10 @@
 
 <!-- Note: contrary to 3.x, insert new entries *first* in their section -->
 
+### 4.14.0 (in progress)
+
+- [improvement] JAVA-2959: Don't throw NoNodeAvailableException when all connections busy
+
 ### 4.13.0
 
 - [improvement] JAVA-2940: Add GraalVM native image build configurations
--- a/core/src/main/java/com/datastax/oss/driver/api/core/NodeUnavailableException.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/NodeUnavailableException.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core;
+
+import com.datastax.oss.driver.api.core.metadata.Node;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Objects;
+
+/**
+ * Indicates that a {@link Node} was selected in a query plan, but it had no connection available.
+ *
+ * <p>A common reason to encounter this error is when the configured number of connections per node
+ * and requests per connection is not high enough to absorb the overall request rate. This can be
+ * mitigated by tuning the following options:
+ *
+ * <ul>
+ *   <li>{@code advanced.connection.pool.local.size};
+ *   <li>{@code advanced.connection.pool.remote.size};
+ *   <li>{@code advanced.connection.max-requests-per-connection}.
+ * </ul>
+ *
+ * See {@code reference.conf} for more details.
+ *
+ * <p>Another possibility is when you are trying to direct a request {@linkplain
+ * com.datastax.oss.driver.api.core.cql.Statement#setNode(Node) to a particular node}, but that node
+ * has no connections available.
+ */
+public class NodeUnavailableException extends DriverException {
+
+  private final Node node;
+
+  public NodeUnavailableException(Node node) {
+    super("No connection was available to " + node, null, null, true);
+    this.node = Objects.requireNonNull(node);
+  }
+
+  @NonNull
+  public Node getNode() {
+    return node;
+  }
+
+  @Override
+  @NonNull
+  public DriverException copy() {
+    return new NodeUnavailableException(node);
+  }
+}
--- a/core/src/test/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerTest.java
+++ b/core/src/test/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerTest.java
@@ -21,9 +21,11 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import com.datastax.oss.driver.api.core.AllNodesFailedException;
 import com.datastax.oss.driver.api.core.CqlIdentifier;
 import com.datastax.oss.driver.api.core.DriverTimeoutException;
 import com.datastax.oss.driver.api.core.NoNodeAvailableException;
+import com.datastax.oss.driver.api.core.NodeUnavailableException;
 import com.datastax.oss.driver.api.core.config.DefaultDriverOption;
 import com.datastax.oss.driver.api.core.cql.AsyncResultSet;
 import com.datastax.oss.driver.api.core.cql.BoundStatement;
@@ -32,6 +34,7 @@ import com.datastax.oss.driver.api.core.cql.ExecutionInfo;
 import com.datastax.oss.driver.api.core.cql.PreparedStatement;
 import com.datastax.oss.driver.api.core.cql.Row;
 import com.datastax.oss.driver.api.core.cql.Statement;
+import com.datastax.oss.driver.api.core.metadata.Node;
 import com.datastax.oss.driver.internal.core.session.RepreparePayload;
 import com.datastax.oss.driver.internal.core.util.concurrent.CapturingTimer.CapturedTimeout;
 import com.datastax.oss.protocol.internal.request.Prepare;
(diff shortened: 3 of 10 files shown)


akhaku:
Still needs tests for `CqlPrepareHandler` and `GraphRequestHandler`, but I wanted to get this up for feedback first. This pretty closely tracks Alex's guidance in the linked ticket but there's a technically breaking change i.e. a different exception type thrown in one case. The changes in ContinuousCqlRequestHandlerNodeTargetingTest illustrate the change in exception type.

core/src/test/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerTest.java
```diff
@@ -105,6 +110,28 @@ public void should_fail_if_no_node_available() {
...
+      CompletionStage<AsyncResultSet> resultSetFuture =
+              new CqlRequestHandler(
+                      UNDEFINED_IDEMPOTENCE_STATEMENT,
+                      harness.getSession(),
+                      harness.getContext(),
+                      "test")
+                      .handle();
+      assertThatStage(resultSetFuture)
+              .isFailed(error -> {
+                assertThat(error).isInstanceOf(AllNodesFailedException.class);
```
akhaku, line 128: Clearly haven't worked with assertj before, I'd love feedback on how to make this block more readable!

core/src/main/java/com/datastax/oss/driver/api/core/NodeUnavailableException.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Objects;
+
+/**
+ * Indicates that a Node encountered as part of processing a query plan has no connections
+ * available.
+ *
+ * <p>This might happen under heavy load. The driver will automatically try the next node in the
+ * query plan. Therefore, the only way that the client can observe this exception is as part of a
+ * {@link AllNodesFailedException}.
+ * <p>The default load balancer will attempt to evenly distribute requests across nodes, so a common
```
adutra: You need to run `mvn fmt:format` on your branch, I see that a few classes are not compliant.

core/src/main/java/com/datastax/oss/driver/api/core/NodeUnavailableException.java
```diff
@@ -0,0 +1,54 @@
...
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core;
+
+import com.datastax.oss.driver.api.core.metadata.Node;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.util.Objects;
+
+/**
+ * Indicates that a Node encountered as part of processing a query plan has no connections
```
adutra: Just trying to improve the style a bit:

```suggestion
 * Indicates that a {@link Node} was selected in a query plan, but it had no connection available.
```

core/src/main/java/com/datastax/oss/driver/api/core/NodeUnavailableException.java
adutra: That's a good explanation, but I wouldn't mention "load balancer" and "evenly distribute" here. I would rephrase this paragraph and the following one as follows:

```
 * <p>A common reason to encounter this error is when the configured number of connections per node
 * and requests per connection is not high enough to absorb the overall request rate. This can be
 * mitigated by tuning the following options:
 *
 * <ul>
 *   <li>{@code advanced.connection.pool.local.size};
 *   <li>{@code advanced.connection.pool.remote.size};
 *   <li>{@code advanced.connection.max-requests-per-connection}.
 * </ul>
 *
 * See {@code reference.conf} for more details.
 *
 * <p>Another possibility is when you are trying to direct a request {@linkplain
 * com.datastax.oss.driver.api.core.cql.Statement#setNode(Node) to a particular node}, but that node
 * has no connections available.
```

core/src/test/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerTest.java
adutra: That's very well written, congrats :-)

Just as a suggestion, how about we include 2 nodes in this test? Something like:

```
@Test
public void should_fail_if_nodes_unavailable() {
  RequestHandlerTestHarness.Builder harnessBuilder = RequestHandlerTestHarness.builder();
  try (RequestHandlerTestHarness harness =
      harnessBuilder.withEmptyPool(node1).withEmptyPool(node2).build()) {
    CompletionStage<AsyncResultSet> resultSetFuture =
        new CqlRequestHandler(
                UNDEFINED_IDEMPOTENCE_STATEMENT,
                harness.getSession(),
                harness.getContext(),
                "test")
            .handle();
    assertThatStage(resultSetFuture)
        .isFailed(
            error -> {
              assertThat(error).isInstanceOf(AllNodesFailedException.class);
              Map<Node, List<Throwable>> allErrors =
                  ((AllNodesFailedException) error).getAllErrors();
              assertThat(allErrors).hasSize(2);
              assertThat(allErrors)
                  .hasEntrySatisfying(
                      node1,
                      nodeErrors ->
                          assertThat(nodeErrors)
                              .singleElement()
                              .isInstanceOf(NodeUnavailableException.class));
              assertThat(allErrors)
                  .hasEntrySatisfying(
                      node2,
                      nodeErrors ->
                          assertThat(nodeErrors)
                              .singleElement()
                              .isInstanceOf(NodeUnavailableException.class));
            });
  }
}
```

adutra (changes requested):
Impressive work, thank you!

Please format your code with `mvn fmt:format` and apply my suggestions if you agree with them.

Also, I would need an entry in the changelog, something like:

```
### 4.14.0 (in progress)

- [improvement] JAVA-2959: Don't throw NoNodeAvailableException when all connections busy
```

adutra:
> Still needs tests for `CqlPrepareHandler` and `GraphRequestHandler`, but I wanted to get this up for feedback first. This pretty closely tracks Alex's guidance in the linked ticket but there's a technically breaking change i.e. a different exception type thrown in one case. The changes in ContinuousCqlRequestHandlerNodeTargetingTest illustrate the change in exception type.

Tests for `CqlPrepareHandler` and `GraphRequestHandler` would be absolutely awesome, but don't spend too much time on them, basically the same test you wrote in `CqlRequestHandlerTest` would be more than enough.

And yes, I agree that this PR introduces a behavioral change. We might want to add a note in the upgrade guide as well and warn users that if they were expecting a `NoNodeAvailableException` in certain situations, they now will get an `AllNodesFailedException`.

core/src/test/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandlerTest.java
akhaku: thanks, done

core/src/main/java/com/datastax/oss/driver/api/core/NodeUnavailableException.java
akhaku: done and done

core/src/main/java/com/datastax/oss/driver/api/core/NodeUnavailableException.java
akhaku: done, thanks

akhaku:
Added a note to the upgrade guide as you suggested, ran `mvn fmt:format` and added a test to `CqlPrepareHandlerTest`. I started looking at adding a test to `GraphRequestHandlerTest` for completeness but it didn't look as easy as the other ones so I opted to take your advice and punt on it.

Thank you very much for the quick review! Sorry it was a busy week.

changelog/README.md
```diff
@@ -2,6 +2,10 @@
 
 <!-- Note: contrary to 3.x, insert new entries *first* in their section -->
 
+### 4.14.0
```
adutra:
```suggestion
### 4.14.0 (in progress)
```

upgrade_guide/README.md
```diff
@@ -1,5 +1,15 @@
...
 
+### 4.14.0
+
+#### AllNodesFailedException instead of NoNodeAvailableException in certain cases
+
+[JAVA-2959](https://datastax-oss.atlassian.net/browse/JAVA-2959) changed the behavior for when a
+request is directed to a particular node (via `Statement#setNode(Node)`) and all connections to that
+node are busy. Previously you would get back a `NoNodeAvailableException` but you will now get back
+an `AllNodesFailedException` where the `getAllErrors` map contains a `NodeUnavailableException`
+for that node.
```
adutra: `Statement#setNode(Node)` is not the only way by which you could observe the new behavior. I suggest a more generic rephrasing:

```suggestion
[JAVA-2959](https://datastax-oss.atlassian.net/browse/JAVA-2959) changed the behavior for when a
request cannot be executed because all nodes tried were busy. Previously you would get back a 
`NoNodeAvailableException` but you will now get back an `AllNodesFailedException` where the 
`getAllErrors` map contains a `NodeUnavailableException` for that node.
```

adutra (changes requested):
Thank you @akhaku we are almost there :-)

changelog/README.md
akhaku: done

upgrade_guide/README.md
akhaku: done

adutra (approved):
Awesome work, thank you!

akhaku:
Thank you! Please merge when ready - it looks like it's waiting for a branch check of some sort, perhaps that won't happen since this PR is from a fork.

adutra:
@akhaku would you mind signing our [CLA](https://cla.datastax.com/) if you haven't done so already? Thank you!

akhaku:
Gotcha, all done now

PR #1573, opened 2021-09-27
JAVA-2967: Support native_transport_(address|port) + native_transport_port_ssl for DSE 6.8

absurdfarce:


absurdfarce:
As of this writing this is still an in-progress PR.  The basic fix contained within seems to resolve the issue in my local repro case but no tests have been added yet for the new functionality.

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
```diff
@@ -956,7 +975,11 @@ private boolean isValidPeer(Row peerRow, boolean logIfInvalid) {
               && !peerRow.isNull("native_port");
     } else {
       isValid &=
-          peerRow.getColumnDefinitions().contains("rpc_address") && !peerRow.isNull("rpc_address");
+          (peerRow.getColumnDefinitions().contains("rpc_address") && !peerRow.isNull("rpc_address"))
+              || (peerRow.getColumnDefinitions().contains("native_transport_address")
+                  && peerRow.getColumnDefinitions().contains("native_transport_port")
+                  && !peerRow.isNull("native_transport_address")
+                  && !peerRow.isNull("native_transport_port"));
```
adutra, line 982: How about a row that has `native_transport_port_ssl` but not `native_transport_port`? Can this even happen?

absurdfarce: native_transport_port has a default value of 9042 so even if nothing is specified for it in cassandra.yaml the corresponding system.peers row will still have a value.  I don't think there is a way to create a row here that doesn't have a native_transport_port entry so this logic should hold up.  I'll take a look at DSE code in a bit to confirm.

I also just fired up my local 3 node DSE 6.8.13 cluster (with no value specified for native_transport_port on any node) to confirm... and it does look like the value is populated as expected.

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
```diff
@@ -611,6 +611,17 @@ private static void updateInfo(
...
+      // DSE 6.8 introduced native_transport_address and native_transport_port for the
+      // listen address.  Also included is native_transport_port_ssl (in case users
+      // want to setup a different port for SSL and non-SSL conns).
+      InetAddress nativeAddress = row.getInet("native_transport_address");
+      int nativePort = row.getInt("native_transport_port");
+      if (cluster.getCluster().getConfiguration().getProtocolOptions().getSSLOptions() != null
+          && !row.isNull("native_transport_port_ssl")) {
+        nativePort = row.getInt("native_transport_port_ssl");
+      }
+      broadcastRpcAddress = new InetSocketAddress(nativeAddress, nativePort);
```
absurdfarce, line 624: Location of this block (and similar logic in the other blocks) was very deliberate.  We can't go underneath the base case handler of rpc_address + default port but it does seem like this specific case will be less likely than the native_address/port case above us.

driver-core/src/test/java/com/datastax/driver/core/exceptions/ExceptionsTest.java
```diff
@@ -284,7 +284,7 @@ public void should_create_proper_read_timeout_exception() {
     ReadTimeoutException e = new ReadTimeoutException(endPoint1, LOCAL_QUORUM, 2, 3, true);
     assertThat(e.getMessage())
         .isEqualTo(
-            "Cassandra timeout during read query at consistency LOCAL_QUORUM (3 responses were required but only 2 replica responded)");
+            "Cassandra timeout during read query at consistency LOCAL_QUORUM (3 responses were required but only 2 replica responded). In case this was generated during read repair, the consistency level is not representative of the actual consistency.");
```
absurdfarce: This and the similar change below address some minor fallout from https://github.com/datastax/java-driver/pull/1571

driver-core/src/test/java/com/datastax/driver/core/ControlConnectionTest.java
```diff
@@ -560,6 +571,292 @@ public void should_connect_when_peers_v2_table_not_present() {
...
+                }
+              });
+      assertThat(hostEndPoints).contains(state.getExpectedEndPoint(scassandras));
+    } finally {
+      if (cluster != null) cluster.close();
+      scassandras.stop();
+    }
+  }
+
+  static class PeerRowState {
```
absurdfarce, line 729: State object + builder is a bit more convoluted than I'd like but it has the effect of nicely isolating changes required for an individual test.  Without this class I found myself passing around multiple config options (all of which were either related or similar to each other) in order to setup a given test.  Introducing this class had the effect of making the intent of a single test case above very clear.

driver-core/src/test/java/com/datastax/driver/core/ControlConnectionTest.java
```diff
@@ -560,6 +571,292 @@ public void should_connect_when_peers_v2_table_not_present() {
...
+    runPeerTest(state);
+  }
+
+  /**
+   * If both native_transport_port and native_transport_port_ssl are present we expect the SSL
+   * version to win. Note that we don't make this conditional on SSL actually being configured.
+   */
+  @Test(groups = "short")
+  @CCMConfig(createCcm = false)
+  @Ignore("Requires SSL support in scassandra")
```
absurdfarce, line 619: Really no good way to test this without SSL support for scassandra.  https://github.com/scassandra/scassandra-server/issues/65 is of relevance here.

A minor addendum; this might be obvious but any change here would also require tweaks to runPeerTest() in order to allow it to (optionally) enable SSL based on the contents of the PeerRowState instance.

driver-core/src/test/java/com/datastax/driver/core/ControlConnectionTest.java
```diff
@@ -560,6 +571,292 @@ public void should_connect_when_peers_v2_table_not_present() {
...
+   */
+  @Test(groups = "short")
+  @CCMConfig(createCcm = false)
+  public void should_extract_hosts_using_rpc_address_from_peers() throws UnknownHostException {
+
+    InetAddress expectedAddress = InetAddress.getByName("4.3.2.1");
+    PeerRowState state =
+        PeerRowState.builder()
+            .peers("rpc_address", expectedAddress)
+            /* DefaultEndPointFactory isn't happy if we don't have both a value for
```
adutra:
```suggestion
            /* DefaultEndPointFactory isn't happy if we don't have a value for
```

driver-core/src/test/java/com/datastax/driver/core/exceptions/ExceptionsTest.java
adutra: Ouch, and I approved that PR :-D.
Alternatively, you could use `assertThat(...).contains(...)` to keep the code concise.

driver-core/src/test/java/com/datastax/driver/core/ControlConnectionTest.java
```diff
@@ -560,6 +571,292 @@ public void should_connect_when_peers_v2_table_not_present() {
...
+            .peers("native_transport_port", expectedPort)
+            .expectedAddress(expectedAddress)
+            .expectedPort(expectedPort)
+            .build();
+    runPeerTest(state);
+  }
+
+  /**
+   * If both native_transport_port and native_transport_port_ssl are present we expect the SSL
+   * version to win. Note that we don't make this conditional on SSL actually being configured.
```
adutra: > Note that we don't make this conditional on SSL actually being configured

Are we sure? I see SSL being tested in a few places with the expression:

```
cluster.getCluster().getConfiguration().getProtocolOptions().getSSLOptions() != null
```

driver-core/src/test/java/com/datastax/driver/core/ControlConnectionTest.java
```diff
@@ -560,6 +571,292 @@ public void should_connect_when_peers_v2_table_not_present() {
...
+    static class Builder {
+
+      private ImmutableMap.Builder<String, Object> peers = this.basePeerRow();
+      private ImmutableMap.Builder<String, Object> peersV2 = this.basePeerRow();
+      private ImmutableMap.Builder<String, Object> local = this.basePeerRow();
+
+      private InetAddress expectedAddress;
+      private Optional<Integer> expectedPort = Optional.absent();
+
+      private BitSet shouldPrime = new BitSet(3);
```
adutra: What is the third bit used for? If only 2 are required, maybe you could simplify and use 2 booleans?

driver-core/src/test/java/com/datastax/driver/core/ControlConnectionTest.java
adutra: Very clean design imo :-)

adutra (changes requested):
Looking really good!

driver-core/src/test/java/com/datastax/driver/core/ControlConnectionTest.java
absurdfarce: Agreed.  I refreshed my recollection after writing that comment and just forgot to go clean it up.  I'll clean this up so that it makes sense.

driver-core/src/test/java/com/datastax/driver/core/ControlConnectionTest.java
absurdfarce: This is also a holdover that prolly doesn't make sense anymore.  The third bit originally referred to system.local, but as discussed above we have to _always_ prime system.local now.

It also probably does make sense to just switch to booleans now.  With three vals it seemed worth encapsulating everything in a single data structure... but by the time you add in static flags for each name the benefit is somewhat lost.

driver-core/src/test/java/com/datastax/driver/core/exceptions/ExceptionsTest.java
absurdfarce: Hmmmm... that might actually be desirable anyway.  It makes the test a purer expression of the thing we're actually testing for rather than worrying about the total structure of the log message.  I think I like that.

absurdfarce:
Thanks for the very complete review @adutra !  I _believe_ I've addressed all the points you raised so this should be ready for another pass.  If I missed something then by all means let me know.

Thanks again!

adutra (approved):

PR #1576, opened 2021-11-02
BasicLoadBalancingPolicy remote compute nodes check all liveNodes

Schnikonos:
BasicLoadBalancingPolicy remote compute nodes minor fix : don't presume local dc nodes are up and among the liveNodes that where found (can happen if the local dc wasn't up when the application started)

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/BasicLoadBalancingPolicy.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/BasicLoadBalancingPolicy.java
@@ -42,6 +42,7 @@ import com.datastax.oss.driver.internal.core.util.collection.CompositeQueryPlan;
 import com.datastax.oss.driver.internal.core.util.collection.LazyQueryPlan;
 import com.datastax.oss.driver.internal.core.util.collection.QueryPlan;
 import com.datastax.oss.driver.internal.core.util.collection.SimpleQueryPlan;
+import com.datastax.oss.driver.shaded.guava.common.base.Predicates;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import java.nio.ByteBuffer;
@@ -54,6 +55,8 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.IntUnaryOperator;
+import java.util.stream.Collectors;
+
 import net.jcip.annotations.ThreadSafe;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


absurdfarce:
Thanks for the contribution @Schnikonos !  Unfortunately I've been swamped with other work recently but I'm trying to make my way back to PRs for the Java driver (including this one!) soon.

Schnikonos:
Hello @absurdfarce 

No pb, thx for the feedback :)

I also followed the link you provided and signed the license agreement.

Have a nice day !

jdonenine:
Hey @Schnikonos thanks again for the contribution and for taking care of the CLA.  I've created this issue here to for tracking purposes as we hope to start reviewing your PR soon:  https://datastax-oss.atlassian.net/browse/JAVA-2987

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/BasicLoadBalancingPolicy.java
```diff
@@ -323,10 +323,7 @@ protected NodeDistanceEvaluator createNodeDistanceEvaluator(
           @Override
           protected Object[] computeNodes() {
             Object[] dcs = liveNodes.dcs().toArray();
-            if (dcs.length <= 1) {
-              return EMPTY_NODES;
-            }
-            Object[] remoteNodes = new Object[(dcs.length - 1) * maxNodesPerRemoteDc];
+            Object[] remoteNodes = new Object[dcs.length * maxNodesPerRemoteDc];
```
absurdfarce: Hey @Schnikonos , again my apologies for the delay in getting back to you on this.

This does fix the problem, but unfortunately we lose the benefit of the early return.  We're now relying on the !dc.equals(localDc) check in the for loop below before we can exit in the no-remote DC case.  Put differently we have to loop through the list of DCs twice in order to exit, once when constructing the array of DC names and again when looping through those DCs to add nodes.

I'm wondering if we can't improve on that by looping through DCs once before we get to the early return:

```java
         @Override
          protected Object[] computeNodes() {
            Set<String> remoteDcs =
                liveNodes.dcs().stream()
                    .filter(Predicates.not(Predicates.equalTo(localDc)))
                    .collect(Collectors.toSet());
            if (remoteDcs.isEmpty()) {
              return EMPTY_NODES;
            }
            Object[] remoteNodes = new Object[remoteDcs.size() * maxNodesPerRemoteDc];
            int remoteNodesLength = 0;
            for (String dc : remoteDcs) {
              Object[] remoteNodesInDc = liveNodes.dc(dc).toArray();
...
```

This works, but if we've come this far why not embrace the streaming API more completely?

```java
          @Override
          protected Object[] computeNodes() {
            Set<String> dcs = liveNodes.dcs();
            Object[] remoteNodes =
                dcs.stream()
                    .filter(Predicates.not(Predicates.equalTo(localDc)))
                    .flatMap(dc -> liveNodes.dc(dc).stream().limit(maxNodesPerRemoteDc))
                    .collect(Collectors.toList())
                    .toArray();

            int remoteNodesLength = remoteNodes.length;
            if (remoteNodesLength == 0) {
              return EMPTY_NODES;
            }
            shuffleHead(remoteNodes, remoteNodesLength);
            return remoteNodes;
          }
```

This feels like it might be the cleanest choice as it avoids the creation of the intermediate objects while also simplifying the logic considerably.

What do you think?

Schnikonos: Hey @absurdfarce 
Indeed I like your last suggestion a lot -> I also think that it makes the code far cleaner. 
I updated my PR with it (thx for the proposition !)

absurdfarce:
Excellent, thanks @Schnikonos!  This looks pretty good to me so I'm gonna call this done.  Thanks for raising the issue _and_ providing a PR to fix it!

PR #1577, opened 2021-12-15
Upgrade ESRI geometry to 2.x

absurdfarce:
An experiment in upgrading ESRI geometry to 2.x.  Newer versions removed the dependency on an older version of Jackson which in turn would allow us to remove that same dependency from the Java driver.

Additional context at https://github.com/datastax/java-driver/pull/1575#issuecomment-995061242

### DIFF
--- a/core/src/test/java/com/datastax/dse/driver/internal/core/data/geometry/DefaultLineStringTest.java
+++ b/core/src/test/java/com/datastax/dse/driver/internal/core/data/geometry/DefaultLineStringTest.java
@@ -34,7 +34,8 @@ public class DefaultLineStringTest {
 
   private final String wkt = "LINESTRING (30 10, 10 30, 40 40)";
 
-  private final String json =
+  /* A JSON representation that was explicitly tested for in ESRI < 2.x impls */
+  private final String legacyJson =
       "{\"type\":\"LineString\",\"coordinates\":[[30.0,10.0],[10.0,30.0],[40.0,40.0]]}";
 
   @Test
@@ -94,13 +95,17 @@ public class DefaultLineStringTest {
   }
 
   @Test
-  public void should_parse_valid_geo_json() {
-    assertThat(LineString.fromGeoJson(json)).isEqualTo(lineString);
+  public void should_create_and_parse_geo_json() {
+
+    String geoJsonStr = lineString.asGeoJson();
+    assertThat(JSONUtils.isValidJsonMap(geoJsonStr)).isTrue();
+    LineString newLineString = LineString.fromGeoJson(geoJsonStr);
+    assertThat(newLineString).isEqualTo(lineString);
   }
 
   @Test
-  public void should_convert_to_geo_json() {
-    assertThat(lineString.asGeoJson()).isEqualTo(json);
+  public void should_parse_legacy_geo_json() {
+    assertThat(LineString.fromGeoJson(legacyJson)).isEqualTo(lineString);
   }
 
   @Test
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,6 @@
     <json.version>20210307</json.version>
     <jackson.version>2.12.2</jackson.version>
     <jackson-databind.version>2.12.2</jackson-databind.version>
-    <legacy-jackson.version>1.9.12</legacy-jackson.version>
     <!-- optional dependencies -->
     <snappy.version>1.1.7.3</snappy.version>
     <lz4.version>1.7.1</lz4.version>
--- a/pom.xml
+++ b/pom.xml
@@ -48,7 +48,7 @@
     <hdrhistogram.version>2.1.12</hdrhistogram.version>
     <metrics.version>4.1.18</metrics.version>
     <netty.version>4.1.60.Final</netty.version>
-    <esri.version>1.2.1</esri.version>
+    <esri.version>2.2.4</esri.version>
     <!--
     When upgrading TinkerPop please upgrade the version matrix in
     manual/core/integration/README.md
--- a/core/src/test/java/com/datastax/dse/driver/internal/core/data/geometry/DefaultPointTest.java
+++ b/core/src/test/java/com/datastax/dse/driver/internal/core/data/geometry/DefaultPointTest.java
@@ -30,7 +30,8 @@ public class DefaultPointTest {
 
   private final String wkt = "POINT (1.1 2.2)";
 
-  private final String json = "{\"type\":\"Point\",\"coordinates\":[1.1,2.2]}";
+  /* A JSON representation that was explicitly tested for in ESRI < 2.x impls */
+  private final String legacyJson = "{\"type\":\"Point\",\"coordinates\":[1.1,2.2]}";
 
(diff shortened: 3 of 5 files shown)


core/src/test/java/com/datastax/dse/driver/internal/core/data/geometry/DefaultLineStringTest.java
```diff
@@ -94,13 +95,17 @@ public void should_load_from_well_known_binary() {
...
 
   @Test
-  public void should_parse_valid_geo_json() {
-    assertThat(LineString.fromGeoJson(json)).isEqualTo(lineString);
+  public void should_create_and_parse_geo_json() {
+
+    String geoJsonStr = lineString.asGeoJson();
+    assertThat(JSONUtils.isValidJsonMap(geoJsonStr)).isTrue();
+    LineString newLineString = LineString.fromGeoJson(geoJsonStr);
+    assertThat(newLineString).isEqualTo(lineString);
```
absurdfarce, line 103: Not precisely what it says on the tin anymore.  We're validating the following:

1. We can generate JSON through a method called "asGeoJson"
2. The JSON so generated is valid
3. We can use this JSON to construct a new geometry type that matches up to the original

This doesn't strictly validate that we're generating GeoJSON but it _does_ validate that we're generating something that the ESRI datatypes can work with.  Seems like that's what we're really interested in here; we're then deferring the question of whether that JSON matches up to the GeoJSON spec to the ESRI impl.

I'm very open to the idea that renaming this method to reflect these changed expectations might be a good idea.

core/src/test/java/com/datastax/dse/driver/internal/core/data/geometry/DefaultLineStringTest.java
```diff
@@ -94,13 +95,17 @@ public void should_load_from_well_known_binary() {
...
   }
 
   @Test
-  public void should_convert_to_geo_json() {
-    assertThat(lineString.asGeoJson()).isEqualTo(json);
+  public void should_parse_legacy_geo_json() {
+    assertThat(LineString.fromGeoJson(legacyJson)).isEqualTo(lineString);
```
absurdfarce, line 108: I wanted to also make sure we could validate any serialized strings of the old format that might be out there somewhere.

pom.xml
```diff
@@ -59,7 +59,6 @@
     <json.version>20210307</json.version>
     <jackson.version>2.12.2</jackson.version>
     <jackson-databind.version>2.12.2</jackson-databind.version>
-    <legacy-jackson.version>1.9.12</legacy-jackson.version>
```
absurdfarce, line 62: I couldn't find any other consumes of this prop and the version mentioned here matches the Jackson version included by the old ESRI implementation exactly.

absurdfarce:
@adutra I'd love to get your feedback on this.  Removing the legacy Jackson impl from the list of driver dependencies would be great but I want to make sure there isn't anything else I need to account for here.

adutra:
@absurdfarce this idea was brought up when we released driver 4.0. We decided to not upgrade ESRI because we thought it would be more prudent to use the exact same version of ESRI that is used server-side, in order to avoid discrepancies in how geo types are serialized. As far as I know, all versions of DSE use ESRI 1.

So I'm a bit on the fence about this upgrade now. Have you at least run the Geo types integration tests, including the Graph ones? Are they passing?

absurdfarce:
@adutra I hadn't run the integration tests against this yet.  I guess that's what I was after; what results would reasonably make me confident that this change is safe?

LineStringIT and PolygonIT were fine against DSE 6.8.14 but PointIT did fail on the map insert:

```
  <testcase name="should_insert_as_map_keys" classname="com.datastax.dse.driver.api.core.data.geometry.PointIT" time="0.007">
    <error message="Multiple entries with same key: POINT (1.7976931348623157e+308 4.9000000000000000e-324)=4 and POINT (0 0)=1" type="java.lang.IllegalArgumentException">java.la
ng.IllegalArgumentException: Multiple entries with same key: POINT (1.7976931348623157e+308 4.9000000000000000e-324)=4 and POINT (0 0)=1
</error>
  </testcase>
```

Underlying problem there is the curious behaviour that this test fails with ESRI 2.x:

```
  @Test
  public void should_equal_only_itself() {
    assertThat(Point.fromCoordinates(0, 0))
            .isNotEqualTo(Point.fromCoordinates(Double.MAX_VALUE, Double.MIN_VALUE));
  }
```

Interestingly that's true even if I change the impl of DefaultGeometry.equals() to use the operator syntax from ESRI 2.x:

```
  @Override
    if (this == o) {
      return true;
    }
    if (!(o instanceof DefaultGeometry)) {
      return false;
    }
    DefaultGeometry that = (DefaultGeometry) o;
    return OperatorEquals.local().execute(this.getEsriGeometry(),that.getEsriGeometry(),SPATIAL_REFERENCE_4326,null);
  }
```

This is not exactly filling me with confidence.

absurdfarce:
The PointIT failure definitely has me worried.  My inclination is to walk away from the ESRI upgrade idea in general.

This originally came up in the context of a [PR](https://github.com/datastax/java-driver/pull/1575) which made ESRI an optional dependency.  This PR came about as an attempt to sidestep that issue by upgrading to a version of ESRI which doesn't have the problems of the earlier version, but clearly that doesn't look terribly viable.  The other option was to make ESRI optional.  My concern about doing that stemmed from historical decisions regarding splitting the DSE module into submodules.  That's really an orthogonal concern to the question of whether a given dependency is optional or not, though, and the recent conversion of Tinkerpop seems to provide a pretty clear example for this approach.

All in all I'm swinging back towards the idea of just making ESRI an optional dependency (along the lines of what was done with Tinkerpop).  @adutra any objection to such an approach?

adutra:
@absurdfarce in all honesty, the PointIT failure looks like an edge case. But still, even if all the ITs pass we still can't really guarantee that every possible geo type instance is going to be persisted in exactly the same form as before – and if that's not the case, it could be really bad for existing applications.

We had another similar case in the past: `timestamp` type instances and CQL literals in string form are still deserialized server-side using the old venerable `java.util.Date` type (and that's never going to change, in fact). In driver 4.0, we decided to replace everything with `java.time.Instant`, even knowing that especially CQL literals in string form could suffer from subtle parsing discrepancies. We covered the whole thing with a lot of tests and in the end the change was deemed feasible – and nobody complained so far.

With geo types though, I am more on the fence: it is harder to consider that we have enough test coverage to go ahead and do the change. 

In summary I would stay prudent and keep the current ESRI version. However I +1 the idea of just making ESRI an optional dependency just as we did with Tinkerpop.

absurdfarce:
Yeah, I wasn't super-clear in my original comment @adutra.  My concern was not with the specifics of the test failure... it was more that something like that would fail after conversion at all.  That in turn leads to questions about rigorous backwards compatibility between the major versions... but at that point I'm basically repeating what you said.

Agreed that this is not the way forward.  Closing this out and we'll move forward with the idea of making it an optional dependency.

PR #1578, opened 2022-01-05
Fossa scanning workflow (4.x)

jdonenine:
Add a new GitHub action to trigger Fossa scans using a provided build.  This will always scan the latest content of the 4.x branch.

Successful test run of the action in my fork here for reference:  https://github.com/jdonenine/java-driver/actions/runs/1658537156

There will be a similar PR for the 3.x branch as well.


absurdfarce:
Belief is that this will actually cover 4.x and 3.x but we'll confirm after the next commit on 3.x

absurdfarce (approved):

PR #1579, opened 2022-01-05
Fossa scanning workflow (3.x)

jdonenine:
Add a new GitHub action to trigger Fossa scans using a provided build. This will always scan the latest content of the 3.x branch.

Successful test run of the action in my fork here for reference:  https://github.com/jdonenine/java-driver/actions/runs/1658560860

### DIFF
--- a/.github/workflows/dep-lic-scan.yaml
+++ b/.github/workflows/dep-lic-scan.yaml
@@ -1,0 +1,23 @@
+name: Dependency and License Scan
+on:
+  push:
+    branches:
+      - '4.x'
+      - '3.x'
+    paths-ignore:
+      - 'manual/**'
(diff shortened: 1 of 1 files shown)


absurdfarce:
Holding on this for now.  Hope is that we actually cover 3.x with https://github.com/datastax/java-driver/pull/1578 since 4.x is the default branch and (according to docs anyway) having this config on the default branch is enough.  If that doesn't work we can always come back and merge this as well, although if we do so we'll prolly want to make the version on each branch branch-specific.

absurdfarce:
Based on conversation with @jdonenine believe we can get this working on 3.x with just this PR.

absurdfarce (approved):

PR #1580, opened 2022-01-07
JAVA-2982: Follow-up to recent PR making ESRI an optional dependency

absurdfarce:
Follow up to recent PR (https://github.com/datastax/java-driver/pull/1575) making ESRI an optional dependency.

I've empirically validated that as of this change the following statements are true:

* The core JAR contains no classes from the ESRI jar (expected... this was true before and shouldn't have changed in this effort)
* The shaded JAR contains no classes from the ESRI jar (new)
    * After https://github.com/datastax/java-driver/pull/1580/commits/953c72f428c70a6c7450d94843fc33b1c7cd3cf1 no classes from ESRI's transitive dependencies are included in the shaded JAR
* Both the core and shaded JAR now declare an optional dependency on ESRI (new)
    * ESRI's transitive dependencies are also optional as well
* All unit and integration tests still pass

### DIFF
--- a/core-shaded/pom.xml
+++ b/core-shaded/pom.xml
@@ -146,8 +146,6 @@
                   -->
                   <include>com.datastax.oss:java-driver-core</include>
                   <include>io.netty:*</include>
-                  <include>org.json:*</include>
-                  <include>org.codehaus.jackson:*</include>
                   <include>com.fasterxml.jackson.core:*</include>
                 </includes>
               </artifactSet>
@@ -159,14 +157,6 @@
                 <relocation>
                   <pattern>io.netty</pattern>
                   <shadedPattern>com.datastax.oss.driver.shaded.netty</shadedPattern>
-                </relocation>
-                <relocation>
-                  <pattern>org.json</pattern>
-                  <shadedPattern>com.datastax.oss.driver.shaded.json</shadedPattern>
-                </relocation>
-                <relocation>
-                  <pattern>org.codehaus.jackson</pattern>
-                  <shadedPattern>com.datastax.oss.driver.shaded.codehaus.jackson</shadedPattern>
                 </relocation>
                 <relocation>
                   <pattern>com.fasterxml.jackson</pattern>
@@ -185,18 +175,6 @@
                 </filter>
                 <filter>
                   <artifact>io.netty:*</artifact>
-                  <excludes>
-                    <exclude>META-INF/**</exclude>
-                  </excludes>
-                </filter>
-                <filter>
-                  <artifact>org.json:*</artifact>
-                  <excludes>
-                    <exclude>META-INF/**</exclude>
-                  </excludes>
-                </filter>
-                <filter>
-                  <artifact>org.codehaus.jackson:*</artifact>
                   <excludes>
                     <exclude>META-INF/**</exclude>
                   </excludes>
--- a/core-shaded/pom.xml
+++ b/core-shaded/pom.xml
@@ -312,7 +312,7 @@
                   <!--
                   1) Don't import packages shaded in the driver bundle. Note that shaded-guava lives
                   in its own bundle, so we must explicitly *not* mention it here.
-                  -->!com.datastax.oss.driver.shaded.netty.*, !com.datastax.oss.driver.shaded.json.*, !com.datastax.oss.driver.shaded.codehaus.jackson.*, !com.datastax.oss.driver.shaded.fasterxml.jackson.*,
+                  -->!com.datastax.oss.driver.shaded.netty.*, !com.datastax.oss.driver.shaded.fasterxml.jackson.*,
                   <!--
                   2) Don't include the packages below because they contain annotations only and are
                   not required at runtime.
@@ -338,7 +338,7 @@
                 1) The driver's packages (API and internal);
                 2) All shaded packages, except Guava which resides in a separate bundle.
                 -->
-                <Export-Package>com.datastax.oss.driver.api.core.*, com.datastax.oss.driver.internal.core.*, com.datastax.dse.driver.api.core.*, com.datastax.dse.driver.internal.core.*, com.datastax.oss.driver.shaded.netty.*, com.datastax.oss.driver.shaded.json.*, com.datastax.oss.driver.shaded.codehaus.jackson.*, com.datastax.oss.driver.shaded.fasterxml.jackson.*,</Export-Package>
+                <Export-Package>com.datastax.oss.driver.api.core.*, com.datastax.oss.driver.internal.core.*, com.datastax.dse.driver.api.core.*, com.datastax.dse.driver.internal.core.*, com.datastax.oss.driver.shaded.netty.*, com.datastax.oss.driver.shaded.fasterxml.jackson.*,</Export-Package>
               </instructions>
               <rebuildBundle>true</rebuildBundle>
             </configuration>
--- a/core-shaded/pom.xml
+++ b/core-shaded/pom.xml
@@ -146,7 +146,6 @@
                   -->
                   <include>com.datastax.oss:java-driver-core</include>
                   <include>io.netty:*</include>
-                  <include>com.esri.geometry:*</include>
                   <include>org.json:*</include>
                   <include>org.codehaus.jackson:*</include>
                   <include>com.fasterxml.jackson.core:*</include>
@@ -160,10 +159,6 @@
                 <relocation>
                   <pattern>io.netty</pattern>
                   <shadedPattern>com.datastax.oss.driver.shaded.netty</shadedPattern>
-                </relocation>
-                <relocation>
-                  <pattern>com.esri</pattern>
-                  <shadedPattern>com.datastax.oss.driver.shaded.esri</shadedPattern>
                 </relocation>
                 <relocation>
                   <pattern>org.json</pattern>
@@ -190,12 +185,6 @@
                 </filter>
                 <filter>
                   <artifact>io.netty:*</artifact>
-                  <excludes>
-                    <exclude>META-INF/**</exclude>
-                  </excludes>
-                </filter>
-                <filter>
-                  <artifact>com.esri.geometry:*</artifact>
                   <excludes>
                     <exclude>META-INF/**</exclude>
                   </excludes>
@@ -311,6 +300,11 @@
                   <artifactId>jctools-core</artifactId>
                   <version>2.1.2</version>
                 </additionalDependency>
+                <additionalDependency>
+                  <groupId>com.esri.geometry</groupId>
+                  <artifactId>esri-geometry-api</artifactId>
+                  <version>1.2.1</version>
+                </additionalDependency>
               </additionalDependencies>
             </configuration>
           </execution>
@@ -340,7 +334,7 @@
                   <!--
                   1) Don't import packages shaded in the driver bundle. Note that shaded-guava lives
                   in its own bundle, so we must explicitly *not* mention it here.
-                  -->!com.datastax.oss.driver.shaded.netty.*, !com.datastax.oss.driver.shaded.esri.*, !com.datastax.oss.driver.shaded.json.*, !com.datastax.oss.driver.shaded.codehaus.jackson.*, !com.datastax.oss.driver.shaded.fasterxml.jackson.*,
+                  -->!com.datastax.oss.driver.shaded.netty.*, !com.datastax.oss.driver.shaded.json.*, !com.datastax.oss.driver.shaded.codehaus.jackson.*, !com.datastax.oss.driver.shaded.fasterxml.jackson.*,
                   <!--
                   2) Don't include the packages below because they contain annotations only and are
                   not required at runtime.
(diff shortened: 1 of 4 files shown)


core-shaded/pom.xml
```diff
@@ -311,6 +300,11 @@
                   <artifactId>jctools-core</artifactId>
                   <version>2.1.2</version>
                 </additionalDependency>
+                <additionalDependency>
+                  <groupId>com.esri.geometry</groupId>
+                  <artifactId>esri-geometry-api</artifactId>
+                  <version>1.2.1</version>
+                </additionalDependency>
```
absurdfarce, line 285: javadoc generation was throwing warnings about missing classes without this declaration in place

manual/core/integration/README.md
```diff
@@ -479,8 +479,9 @@ don't use any of the above features, you can safely exclude the dependency:
 Our [geospatial types](../dse/geotypes/) implementation is based on the [Esri Geometry
 API](https://github.com/Esri/geometry-api-java).
 
-Esri is declared as a required dependency, but the driver can operate normally without it. If you
-don't use geospatial types anywhere in your application, you can exclude the dependency:
+For driver versions >= 4.4.0 and < 4.14.0 Esri is declared as a required dependency,
```
absurdfarce, line 482: The TinkerPop example explicitly referred to an inclusive range, but in this case I wanted the range to be inclusive on the low end but exclusive on the high end (in order to explicitly convey the idea that all drivers before the named version used a different behaviour).  Since this changed the semantics slightly I thought that perhaps identifying versions via the >= and < syntax might be more explicit.

I'm open to other ideas of presenting this information if this isn't clear.

upgrade_guide/README.md
```diff
@@ -9,6 +9,26 @@ request cannot be executed because all nodes tried were busy. Previously you wou
...
+```xml
+<dependency>
+  <groupId>com.esri.geometry</groupId>
+  <artifactId>esri-geometry-api</artifactId>
+  <version>${esri.version}</version>
+</dependency>
+```
+
+See the [integration](../manual/core/integration/#esri) section in the manual for more details.
+
```
absurdfarce, line 31: Very much based on the example used when TinkerPop was made optional.

I've explicitly added this in the 4.14.0 section under the assumption that changing a dependency from required to optional breaks backwards compat and thus requires at least a minor release.

core-shaded/pom.xml
```diff
@@ -146,7 +146,6 @@
                   -->
                   <include>com.datastax.oss:java-driver-core</include>
                   <include>io.netty:*</include>
-                  <include>com.esri.geometry:*</include>
```
adutra, line 149: I'm a bit hesitant about this. Making a dependency optional and shading it are different concerns.

Initially, the decision to shade a dependency was taken by looking at the likelihood of a version conflict for a given dependency. For Esri, we deemed it high because the driver requires a very old Esri version. 

I think we should keep shading Esri even if it is optional, wdyt? This means that in the shaded jar Esri would be present and shaded, just like it is now. I can't see a situation where this could be a problem for the end user.

manual/core/integration/README.md
```diff
@@ -496,6 +497,18 @@ don't use geospatial types anywhere in your application, you can exclude the dep
 </dependency>
 ```
 
+Starting with driver 4.14.0 Esri has been changed to an optional dependency.  You no longer have to
```
adutra, line 500: Now that we are doing this for Esri, I wonder why we wouldn't do the same for Jackson. Do you mind opening a ticket for that, if it doesn't exist yet?

manual/core/integration/README.md
```diff
@@ -496,6 +497,18 @@ don't use geospatial types anywhere in your application, you can exclude the dep
...
+explicitly exclude the dependency if it's not used, but if you do wish to make use of the Esri
+library you must now explicitly specify it as a dependency :
+
+```xml
+<dependency>
+  <groupId>com.esri.geometry</groupId>
+  <artifactId>esri-geometry-api</artifactId>
+  <version>${esri.version}</version>
```
adutra, line 508: We could add a note here about the fact that the driver requires Esri 1.2.x, wdyt? This could be useful for developers since it's such an old version.

upgrade_guide/README.md
```diff
@@ -9,6 +9,26 @@ request cannot be executed because all nodes tried were busy. Previously you wou
 `NoNodeAvailableException` but you will now get back an `AllNodesFailedException` where the
 `getAllErrors` map contains a `NodeUnavailableException` for that node.
 
+#### Esri Geometry dependency now optional
```
adutra, line 12: Now that we are doing this for Esri, I wonder why we wouldn't do the same for Jackson and Reactive Streams. Indeed, these 2 dependencies are handled the same way as Esri used to be handled: declared as mandatory in pom.xml, but made optional by the magic of reflection. 

Do you mind opening a ticket for that, if it doesn't exist yet?

adutra (changes requested):

manual/core/integration/README.md
absurdfarce: This isn't quite as clear cut for me.  Jackson is directly used at several spots within the driver:

* Insights code
* Graph code
* Cloud config code
* JsonCodec

The first is a no-op now that Insights is defunct; if anything I'd argue the driving code on our side should be removed, but that's a distraction from our current conversation.  The second usage above fits nicely into the model we're establishing here: dependencies of DSE features should be made optional.  If we stopped there I'd be more inclined to agree.

It's the last two that give me pause.  I don't know that we want to require users to manually add a dep when using the driver against Astra.  I'd make a similar argument for JsonCodec, although I readily agree it's less significant than the Astra use case.

Worth noting: the user who created the original PR to make ESRI optional did so because of the old Jackson version that came along with it.  Just making ESRI optional resolves that by using the more recent version we explicitly include.

Whaddya think?

manual/core/integration/README.md
absurdfarce: I completely agree with this.  My original version of these docs spelled out the version of ESRI to include but I shifted over the the prop-based model in use here to avoid customer trying to use a newer version.  That said, I readily agree that an additional warning here about the older version would be worthwhile... I'll add that in.

upgrade_guide/README.md
absurdfarce: See above re: my hesitation about doing so with Jackson but the reactive streams dependency would seem to make a lot of sense on this front.  I'll file a ticket for that.

absurdfarce: Ticket is https://datastax-oss.atlassian.net/browse/JAVA-2981

manual/core/integration/README.md
absurdfarce: Added the trailing note here re: the use of an older ESRI version.  Does that address your concerns here?

core-shaded/pom.xml
absurdfarce: I see your point here... I'm just not sure I agree.

To be clear, the case you're describing is fairly precise.  It would have to be something like the following:

* The user's app uses ESRI 2.x for some other functionality in their app
* The user's app also uses the (shaded) Java driver
* The user runs this app against DSE
* The user wants to leverage the geometric type support in DSE

This user will (possibly, even likely) experience a class conflict.  But the cost for supporting this (quite limited case) is that _every user of the shaded JAR_ has a larger JAR on their classpath.  Initially it seemed like optimizing for the broad case made sense rather than doing so for the extremely precise case but I'm certainly open to argument.

adutra: > every user of the shaded JAR has a larger JAR on their classpath

Well the shaded jar already has all of Netty in it, so adding Esri on top of it won't really make a noticeable size difference :-)

But OK, I don't mind if we stop shading Esri.

manual/core/integration/README.md
adutra: JsonCodec is optional, the driver doesn't register it by default, so I still think it falls under the "could-be-optional" category. Insights and Graph too.

It's only Astra that bugs me. 

After some more thought, OK to keep it mandatory. I'd make a case for creating a ticket to remove the dependency to Jackson in `CloudConfigFactory` first, and only then make Jackson optional.

> the user who created the original PR to make ESRI optional did so because of the old Jackson version that came along with it.

Indeed, Esri comes with two annoying dependencies because they are old and popular at the same time:

* org.json:json:20210307
* org.codehaus.jackson:jackson-core-asl:1.9.12

2 additional questions:

* Are we sure that the above dependencies are now absent from the shaded jar as well?
* Have you seen this? https://github.com/spring-projects/spring-boot/issues/29387#issuecomment-1015332594 In short: `org.json:json-api` was promoted to direct dependency for obscure OSGi-related reasons. Please check what's the current status of this dependency in the core pom.xml, in the shaded jar and in the OSGi tests.

manual/core/integration/README.md
```diff
@@ -496,6 +497,22 @@ don't use geospatial types anywhere in your application, you can exclude the dep
...
+</dependency>
+```
+
+The dependency specification above will introduce a dependency on version 1.2.1 of the Esri
```
adutra: Small suggestion to better link this to the above snippet:

> In the dependency specification above, `esri.version` should be any version in the range `[1.2,2.0)`, for example 1.2.1. While this version is now old, it is the only version that is guaranteed to be fully compatible with DSE.

absurdfarce: Oh, I like that _much_ better.  I tried to re-phrase what I had a couple of times but I never really liked what I came up with... I'll switch to your language instead.

absurdfarce: I basically used the language you provided with a (hopefully small and non-obtrusive) tweak to avoid specifying the version range with the [) syntax; I was worried it might be confusing to some users.  Take a look and let me know if you're okay with that usage.

manual/core/integration/README.md
adutra: We are good thanks!

manual/core/integration/README.md
adutra: That's excellent, thanks!

adutra:
@absurdfarce we are almost good to merge this but I wanted first your input on this:

spring-projects/spring-boot#29387

What should we do with `org.json:json-api`? I guess it should be made optional, or even removed completely from our pom (if we can keep the OSGi tests passing).

absurdfarce:
@adutra I looked at the Spring Boot issue briefly but I haven't really set out to address it specifically yet.  We're still getting the transitive dependencies from the Esri lib (the JSON JAR and the older version of Jackson) included in the shaded JAR so I wanted to get that sorted out before considering the Spring issue (since they looked to be related).

More on that shortly (I hope).

manual/core/integration/README.md
absurdfarce: I naively thought that these would be removed by my earlier changes.  As I look back on it I'm not sure why I thought that but I did. :(  Regardless, this should be fixed now in the most recent push.

I'll look into the spring-boot issue as well.

absurdfarce: I naively thought that these would be removed by my earlier changes.  As I look back on it I'm not sure why I thought that but I did. :(  Regardless, this should be fixed now in the most recent push.

I'll look into the spring-boot issue as well.

absurdfarce: I can also confirm that making ESRI optional has the effect of making it's transitive dependencies optional.  That means that org.json:json is now an optional dependency for core and core-shaded.  I'll add a comment onto that Spring ticket confirming that fact.

absurdfarce:
Okay, I _believe_ all the relevant issues are sorted out now.  Would you mind taking another look @adutra to confirm that we're good?

core-shaded/pom.xml
```diff
@@ -340,7 +312,7 @@
                   <!--
                   1) Don't import packages shaded in the driver bundle. Note that shaded-guava lives
                   in its own bundle, so we must explicitly *not* mention it here.
-                  -->!com.datastax.oss.driver.shaded.netty.*, !com.datastax.oss.driver.shaded.esri.*, !com.datastax.oss.driver.shaded.json.*, !com.datastax.oss.driver.shaded.codehaus.jackson.*, !com.datastax.oss.driver.shaded.fasterxml.jackson.*,
+                  -->!com.datastax.oss.driver.shaded.netty.*, !com.datastax.oss.driver.shaded.json.*, !com.datastax.oss.driver.shaded.codehaus.jackson.*, !com.datastax.oss.driver.shaded.fasterxml.jackson.*,
```
adutra: I just noticed: we are still mentioning here two shaded packages that do not exist anymore:

* com.datastax.oss.driver.shaded.json
* com.datastax.oss.driver.shaded.codehaus.jackson

core-shaded/pom.xml
```diff
@@ -366,7 +338,7 @@
                 1) The driver's packages (API and internal);
                 2) All shaded packages, except Guava which resides in a separate bundle.
                 -->
-                <Export-Package>com.datastax.oss.driver.api.core.*, com.datastax.oss.driver.internal.core.*, com.datastax.dse.driver.api.core.*, com.datastax.dse.driver.internal.core.*, com.datastax.oss.driver.shaded.netty.*, com.datastax.oss.driver.shaded.esri.*, com.datastax.oss.driver.shaded.json.*, com.datastax.oss.driver.shaded.codehaus.jackson.*, com.datastax.oss.driver.shaded.fasterxml.jackson.*,</Export-Package>
+                <Export-Package>com.datastax.oss.driver.api.core.*, com.datastax.oss.driver.internal.core.*, com.datastax.dse.driver.api.core.*, com.datastax.dse.driver.internal.core.*, com.datastax.oss.driver.shaded.netty.*, com.datastax.oss.driver.shaded.json.*, com.datastax.oss.driver.shaded.codehaus.jackson.*, com.datastax.oss.driver.shaded.fasterxml.jackson.*,</Export-Package>
```
adutra: Same here.

adutra (changes requested):
I manually verified that the OSGi tests pass against DSE with geotypes, so we are good wrt to the org.json dependency there. I also manually confirmed that org.json is not present neither in driver-core nor driver-core-shaded, so we are good there too.

However I just spotted two bnd directives in driver-core-shaded that should be removed since they reference packages that are not shaded anymore. Please have a look.

core-shaded/pom.xml
absurdfarce: Ah, good catch, you're absolutely right!  I'll clean this up (and the one below) shortly.

adutra (approved):
And we are done \o/ :shipit:

PR #1581, opened 2022-01-12
chore(pom): update netty to 4.1.68

actionless:
Version 4.1.60 have known security vulnerabilities, which are fixed in 4.1.68:

 - `CVE-2021-37136`
 - `CVE-2021-37137`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -47,7 +47,7 @@
     <config.version>1.4.1</config.version>
     <hdrhistogram.version>2.1.12</hdrhistogram.version>
     <metrics.version>4.1.18</metrics.version>
-    <netty.version>4.1.60.Final</netty.version>
+    <netty.version>4.1.68.Final</netty.version>
     <esri.version>1.2.1</esri.version>
     <!--
     When upgrading TinkerPop please upgrade the version matrix in


absurdfarce:
Thanks for the contribution @actionless!  I'm working to try to get a Java driver release out which includes fixes for a few other issues but looking into upgrading Netty was the next item on my list after that release.  In the past we've seen significant performance changes from even patch upgrades to Netty so we'll have to spend some time vetting this upgrade.  But getting those CVEs addressed will be very nice indeed.

actionless:
thanks for the quick review @absurdfarce 

i need to check that (and will get back to you on Monday)

just to clarify one of possible options, could i submit patch as Public Domain license (ie disown it) to avoid signing any papers?

absurdfarce:
Hey @actionless thanks again for the PR!  We're working on a release right now to address some other issues (unrelated to Netty) but we do hope to include an upgrade to Netty in a near-future release.  [JAVA-2977](https://datastax-oss.atlassian.net/browse/JAVA-2977) has more detail on this effort but unfortunately I just don't think we're ready to jump on it yet.  Netty upgrades always involve a lot of testing to make sure we don't see regressions (something we have seen in the past) so there will likely be a lot of work around this besides the dependency upgrade.

A change like this will be included once we start work on JAVA-2977, but it's not something I want to do in isolation right now.

Thanks again!

PR #1585, opened 2022-02-02
JAVA-2976 Protocol v5 error codes CAS_WRITE_UNKNOWN, CDC_WRITE_FAILURE not supported (3.x)

absurdfarce:


absurdfarce:
Code is a fairly cookie-cutter adaptation of what was already present for the other error messages.  Build and existing tests pass straightaway.

Major problem I have here is the lack of testing.  There's no explicit testing around our ability to construct exceptions for these error codes when we get server responses.  That's true for the existing error codes just as much as it is for the ones I've added.  I can accept an argument that says "file a ticket to enhance testing in this area later" but that still leaves this PR without good test coverage.

driver-core/src/main/java/com/datastax/driver/core/exceptions/CDCWriteException.java
```diff
@@ -0,0 +1,61 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.driver.core.exceptions;
+
+import com.datastax.driver.core.EndPoint;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+
+/** An attempt was made to write to a commitlog segment which doesn't support CDC mutations * */
```
adutra: Judging from the verbiage chosen server-side, it seems it has more to do with the CDC log folder being full:

```
String logMsg = String.format("Rejecting mutation to keyspace %s. Free up space in %s by processing CDC logs.",
                mutation.getKeyspaceName(), DatabaseDescriptor.getCDCLogLocation());
```

adutra (approved):
LGTM – and for the record, I agree that we lack of tests for this kind of thing, but there is no easy way around that in driver 3.

driver-core/src/main/java/com/datastax/driver/core/exceptions/CDCWriteException.java
absurdfarce: My original comment was based on https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/db/commitlog/CommitLogSegmentManagerCDC.java#L130-L138.  As I look at the entirety of the comments + source there it is... quite confusing.  It seems like the comment in question literally contradicts itself.

I might just change this to some more generic language.

absurdfarce: Fixed

adutra (approved):

PR #1586, opened 2022-02-03
JAVA-2976 Protocol v5 error codes CAS_WRITE_UNKNOWN, CDC_WRITE_FAILURE not supported (4.x)

absurdfarce:
4.x equivalent to https://github.com/datastax/java-driver/pull/1585.  Relies on https://github.com/datastax/native-protocol/pull/42 to build and function correctly.

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -71,7 +71,7 @@
       <dependency>
         <groupId>com.datastax.oss</groupId>
         <artifactId>native-protocol</artifactId>
-        <version>1.5.1-SNAPSHOT</version>
+        <version>1.5.1</version>
       </dependency>
       <dependency>
         <groupId>com.datastax.oss</groupId>
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -71,7 +71,7 @@
       <dependency>
         <groupId>com.datastax.oss</groupId>
         <artifactId>native-protocol</artifactId>
-        <version>1.5.0</version>
+        <version>1.5.1-SNAPSHOT</version>
       </dependency>
       <dependency>
         <groupId>com.datastax.oss</groupId>
--- a/core/src/main/java/com/datastax/oss/driver/api/core/servererrors/CASWriteUnknownException.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/servererrors/CASWriteUnknownException.java
@@ -25,7 +25,7 @@ import com.datastax.oss.driver.api.core.session.Request;
 import edu.umd.cs.findbugs.annotations.NonNull;
 
 /**
- * The result of a CAS operation is in an unknown state
+ * The result of a CAS operation is in an unknown state.
  *
  * <p>This exception is processed by {@link RetryPolicy#onErrorResponseVerdict(Request,
  * CoordinatorException, int)} , which will decide if it is rethrown directly to the client or if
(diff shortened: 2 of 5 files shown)


bom/pom.xml
```diff
@@ -71,7 +71,7 @@
       <dependency>
         <groupId>com.datastax.oss</groupId>
         <artifactId>native-protocol</artifactId>
-        <version>1.5.0</version>
+        <version>1.5.1-SNAPSHOT</version>
```
absurdfarce: Just a placeholder for now to use a locally-built version of native-protocol that includes https://github.com/datastax/native-protocol/pull/42.  To be replaced by an actual version once we release a new version of native-protocol.

core/src/main/java/com/datastax/oss/driver/api/core/servererrors/CASWriteUnknownException.java
```diff
@@ -0,0 +1,84 @@
...
+import com.datastax.oss.driver.api.core.cql.ExecutionInfo;
+import com.datastax.oss.driver.api.core.metadata.Node;
+import com.datastax.oss.driver.api.core.retry.RetryPolicy;
+import com.datastax.oss.driver.api.core.session.Request;
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+/**
+ * The result of a CAS operation is in an unknown state
+ *
+ * <p>This exception is processed by {@link RetryPolicy#onReadTimeoutVerdict(Request,
```
adutra: This is not accurate, it is processed by `onErrorResponseVerdict`.

adutra (changes requested):
I thought it would be easier to add tests in 4.x than in 3.x, but looking at existing tests, we seem to have a grand total of zero unit tests for specific error codes. I guess we can live we that since the codes were tested in native-protocol. It's just the translation from error code to exception that is not covered, but that's minor.

adutra:
FYI I ran a build on our internal CI off of this branch, it was clean.

core/src/main/java/com/datastax/oss/driver/api/core/servererrors/CASWriteUnknownException.java
absurdfarce: Ah, another good catch... and another copy/paste error.  Fixing.

core/src/main/java/com/datastax/oss/driver/api/core/servererrors/CASWriteUnknownException.java
```diff
@@ -0,0 +1,84 @@
...
+import com.datastax.oss.driver.api.core.ConsistencyLevel;
+import com.datastax.oss.driver.api.core.DriverException;
+import com.datastax.oss.driver.api.core.cql.ExecutionInfo;
+import com.datastax.oss.driver.api.core.metadata.Node;
+import com.datastax.oss.driver.api.core.retry.RetryPolicy;
+import com.datastax.oss.driver.api.core.session.Request;
+import edu.umd.cs.findbugs.annotations.NonNull;
+
+/**
+ * The result of a CAS operation is in an unknown state
```
adutra:
```suggestion
 * The result of a CAS operation is in an unknown state.
```

adutra (approved):

bom/pom.xml
absurdfarce: native-protocol 1.5.1 has now been released

PR #1593, opened 2022-04-15
JAVA-2984 Upgrade jackson-databind to latest 2.7.x (3.x)

absurdfarce:


pom.xml
```diff
@@ -59,7 +59,7 @@
         <hdr.version>2.1.10</hdr.version>
         <jackson.version>2.8.11</jackson.version>
         <!-- jackson-databind 2.7.x is the last to support java 6 -->
-        <jackson-databind.version>2.7.9.3</jackson-databind.version>
+        <jackson-databind.version>2.7.9.7</jackson-databind.version>
```
absurdfarce, line 62: This is [discussed more extensively on the ticket](https://datastax-oss.atlassian.net/browse/JAVA-2984?focusedCommentId=54811) but the basic idea here is as follows:

* Upgrade databind to max 2.7.9.x version.  This doesn't resolve some of vulnerabilities for jackson-databind but moving beyond 2.7.x risks losing good Java 1.6 support
  * Most of the exploits require some pretty specific setup anyway
* jackson-annotation and jackson-core maintain Java 1.6 support all the way up through the current release so we _could_ go to a more recent version here.  But there's no clear reason to do so and I'm a little worried about getting too much distance between jackson-core and jackson-databind.

adutra (approved):

PR #1594, opened 2022-04-15
JAVA-3008 Upgrade to Netty 4.1.75 to address CVEs (3.x)

absurdfarce:


pom.xml
```diff
@@ -51,8 +51,8 @@
         <slf4j.version>1.7.25</slf4j.version>
         <slf4j-log4j12.version>1.7.25</slf4j-log4j12.version>
         <guava.version>19.0</guava.version>
-        <netty.version>4.0.56.Final</netty.version>
-        <netty-tcnative.version>2.0.7.Final</netty-tcnative.version>
+        <netty.version>4.1.75.Final</netty.version>
+        <netty-tcnative.version>2.0.50.Final</netty-tcnative.version>
```
absurdfarce, line 55: Upgraded to match [what's in the 4.1.75 pom.xml](https://github.com/netty/netty/blob/netty-4.1.75.Final/pom.xml#L545) based on [this suggestion](https://github.com/netty/netty-tcnative/issues/366) from Norman

pom.xml
```diff
@@ -51,8 +51,8 @@
         <slf4j.version>1.7.25</slf4j.version>
         <slf4j-log4j12.version>1.7.25</slf4j-log4j12.version>
         <guava.version>19.0</guava.version>
-        <netty.version>4.0.56.Final</netty.version>
-        <netty-tcnative.version>2.0.7.Final</netty-tcnative.version>
+        <netty.version>4.1.75.Final</netty.version>
```
absurdfarce, line 54: This is a _huge_ upgrade in Netty version but it may not be as bad as it appears.  Going to this version is necessary to address all the CVEs; [CVE-2021-21409](https://www.cve.org/CVERecord?id=CVE-2021-21409) and [CVE-2021-37136](https://www.cve.org/CVERecord?id=CVE-2021-37136), for example, are only resolved by upgrading to a recent 4.1.x Netty version.

The argument for an upgrade this big is two-fold.  First, it addresses all the CVEs.  Second, it puts the 3.x and 4.x releases in sync re: Netty versions.  This should at least give us semi-predictable behaviours between the two versions (excluding varying usages between the two drivers).

That said, I have considerable sympathy for such a major change on the prior version of Java driver development.  We should at least consider whether this exposes us to too much risk just to address the CVEs; an alternate answer might be "you have to upgrade to 4.x".

pom.xml
adutra: Excellent, this is indeed a common source of issues when upgrading Netty.

adutra (approved):

absurdfarce:
Just confirmed Netty 4.x will work on Java 1.6 (the minimum version supported by 3.x).  See [here](https://github.com/netty/netty/wiki/Requirements-for-4.x) for more detail.

PR #1597, opened 2022-05-25
JAVA-3022 implementation of private link address translator

smiklosovic:
We are open to rename this translator to match more generic use case. See JAVA-3022.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/addresstranslation/FixedHostNameAddressTranslator.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/addresstranslation/FixedHostNameAddressTranslator.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.internal.core.addresstranslation;
+
+import com.datastax.oss.driver.api.core.addresstranslation.AddressTranslator;
+import com.datastax.oss.driver.api.core.config.DriverOption;
+import com.datastax.oss.driver.api.core.context.DriverContext;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.net.InetSocketAddress;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This translator always returns same hostname, no matter what IP address a node has but still
+ * using its native transport port.
+ *
+ * <p>The translator can be used for scenarios when all nodes are behind some kind of proxy, and it
+ * is not tailored for one concrete use case. One can use this, for example, for AWS PrivateLink as
+ * all nodes would be exposed to consumer - behind one hostname pointing to AWS Endpoint.
+ */
+public class FixedHostNameAddressTranslator implements AddressTranslator {
+
+  private static final Logger LOG = LoggerFactory.getLogger(FixedHostNameAddressTranslator.class);
+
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/addresstranslation/PrivateLinkAddressTranslator.java
```diff
@@ -0,0 +1,50 @@
...
+  public PrivateLinkAddressTranslator(@NonNull DriverContext context) {
+    DriverExecutionProfile config = context.getConfig().getDefaultProfile();
+    advertisedHostname = config.getString(ADVERTISED_HOSTNAME_OPTION);
+  }
+
+  @NonNull
+  @Override
+  public InetSocketAddress translate(@NonNull InetSocketAddress address) {
+    final int port = address.getPort();
+    LOG.debug(String.format("translating %s:%d to %s:%d", address, port, advertisedHostname, port));
```
adutra: You should avoid `String.format`:
```suggestion
    LOG.debug("Translating {}:{} to {}:{}", address, port, advertisedHostname, port));
```

core/src/main/java/com/datastax/oss/driver/internal/core/addresstranslation/PrivateLinkAddressTranslator.java
```diff
@@ -0,0 +1,50 @@
...
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.net.InetSocketAddress;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PrivateLinkAddressTranslator implements AddressTranslator {
+
+  private static final Logger LOG = LoggerFactory.getLogger(PrivateLinkAddressTranslator.class);
+
+  public static final DriverOption ADVERTISED_HOSTNAME_OPTION =
```
adutra: Since this is going to be part of the core driver, you should rather add this new option to `DefaultDriverOption`.

Then you should also declare a new `TypedDriverOption` so that folks using programmatic configuration can use the new option too.

core/src/main/java/com/datastax/oss/driver/internal/core/addresstranslation/PrivateLinkAddressTranslator.java
```diff
@@ -0,0 +1,50 @@
...
+import com.datastax.oss.driver.api.core.addresstranslation.AddressTranslator;
+import com.datastax.oss.driver.api.core.config.DriverExecutionProfile;
+import com.datastax.oss.driver.api.core.config.DriverOption;
+import com.datastax.oss.driver.api.core.context.DriverContext;
+import edu.umd.cs.findbugs.annotations.NonNull;
+import java.net.InetSocketAddress;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PrivateLinkAddressTranslator implements AddressTranslator {
```
adutra: As discussed in the Jira ticket, you may want to give this a more generic name. Also, a bit of javadoc explaining why it is useful would be nice.

adutra (changes requested):
Thanks for the PR!

If this lands in the driver I'd also like to see reference.conf amended. Please modify the `advanced.address-translator` section and:

* Mention the new translator there;
* Add the new `advertised-hostname` option there (it should be commented out by default).

smiklosovic:
@adutra I fixed what I could, thanks for the review. This build will fail anyway on `MapBasedDriverConfigLoaderTest.should_fill_default_profile_like_reference_file:74` because I commented that property in reference config but it expects it is not commented? I am a little bit confused here. Would you please help me here?

adutra (dismissed):
I fixed your build issues, hope you don't mind. I will let @absurdfarce have a look and merge since he's the real boss ™️  now :-)

adutra:
> This build will fail anyway on MapBasedDriverConfigLoaderTest.should_fill_default_profile_like_reference_file:74 because I commented that property in reference config but it expects it is not commented? I am a little bit confused here. Would you please help me here?

This test is a special one that verifies that the programmatic configuration API would produce the exact same configuration as the file-based configuration API.

Since the option in reference.conf is commented out, it should not appear either in `OptionsMap.driverDefaults()`.

smiklosovic:
Hi @adutra , would you mind to fix the build once again? I am not sure what's wrong but "mvn clean install" fails again. I think I just force pushed something by accident and your changes are not there anymore.

smiklosovic:
Ok it seems I managed to build it.

PR #1598, opened 2022-05-25
JAVA-2995 make CodecNotFoundException extend DriverException

smiklosovic:


smiklosovic:
@adutra would you mind to take a look? I just found this low-hanger as I was creating 3022 :D

core/src/main/java/com/datastax/oss/driver/api/core/type/codec/CodecNotFoundException.java
```diff
@@ -62,4 +63,11 @@ public DataType getCqlType() {
   public GenericType<?> getJavaType() {
     return javaType;
   }
+
+  @NonNull
+  @Override
+  public DriverException copy() {
+    return new CodecNotFoundException(
+        getMessage(), getCause(), getCqlType(), getJavaType());
```
adutra: The build is failing due to wrong format:

```suggestion
return new CodecNotFoundException(getMessage(), getCause(), getCqlType(), getJavaType());
```

adutra (commented):
As discussed in the Jira ticket, I'd like to see a note in the upgrade guide because this is in theory a breaking change, which is not allowed in a minor version.

smiklosovic:
@adutra  please run the build again, I addressed your comments.

adutra:
I pushed a few fixes for the upgrade guide. I think this is looking good but I'm still torn on the idea of including this breaking change in a minor. I will leave this open for now and wait for others to chime in, maybe @absurdfarce ?

adutra (approved):

absurdfarce:
I too have been going back and forth on the question of whether we roll this out in anything less than a major release, and I think ultimately I'd argue that we shouldn't.  While the risk of surprising behaviour for this specific change may be fairly small it certainly is a possibility for some users... and I'd rather be very clear about that from a semver perspective.

absurdfarce:
After sleeping on this I'm reversing myself; I'm going to put this into 4.15.0 with a note in the upgrade guide.  It's very consistent with how we've handled similar issues in the past, and going to 5.0 for this change seems like it would create more confusion (to say nothing of false expectations of major new features/improvements).

PR #1599, opened 2022-05-31
JAVA-3021 Update docs to use withPartitionKey instead of withPrimaryKey

absurdfarce:


ErickRamirezAU (approved):
Looks good to me. Thanks for the quick turnaround. 🍻

PR #1604, opened 2022-07-04
JAVA-3005 Refresh entire node list when new node added

akhaku:
Empirically, Cassandra does not appear to send
TopologyChangeType.REMOVED_NODE when a node in the ring
crashes hard and is eventually replaced, even though the
bad node is removed from the system tables. This results in
the driver holding on to that node until a restart.
The 3.x driver refreshes the entire node list when a node is
added, so when a new node comes up to replace the crashed
node, the crashed node is removed from the driver's state.
With this change we port that behavior to the 4.x driver and
refresh the entire node list when a new node is added.
This also fixes JAVA-3002.

### DIFF
--- a/changelog/README.md
+++ b/changelog/README.md
@@ -2,6 +2,10 @@
 
 <!-- Note: contrary to 3.x, insert new entries *first* in their section -->
 
+### 4.14.2 (in progress)
+
+- [bug] JAVA-3002 JAVA-3005: Refresh entire node list when a new node is added
+
 ### 4.14.1
 
 - [improvement] JAVA-3013: Upgrade dependencies to address CVEs and other security issues, 4.14.1 edition
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeStateManager.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeStateManager.java
@@ -185,10 +185,10 @@ public class NodeStateManager implements AsyncAutoCloseable {
             }
           } else {
             LOG.debug(
-                "[{}] Received UP event for unknown node {}, adding it",
+                "[{}] Received UP event for unknown node {}, refreshing node list",
                 logPrefix,
                 event.broadcastRpcAddress);
-            metadataManager.addNode(event.broadcastRpcAddress);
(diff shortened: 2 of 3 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeStateManager.java
```diff
@@ -185,10 +185,10 @@ private void onDebouncedTopologyEvent(TopologyEvent event) {
             }
           } else {
             LOG.debug(
-                "[{}] Received UP event for unknown node {}, adding it",
+                "[{}] Received UP event for unknown node {}, refreshing node list",
                 logPrefix,
                 event.broadcastRpcAddress);
-            metadataManager.addNode(event.broadcastRpcAddress);
+            metadataManager.refreshNodes();
```
akhaku, line 191: Note that `refreshNodes` is an async operation and we aren't waiting for the result. That's probably fine though, we're in an async code path anyway (debounced handling of topology events).

changelog/README.md
```diff
@@ -2,6 +2,10 @@
 
 <!-- Note: contrary to 3.x, insert new entries *first* in their section -->
 
+### 4.14.2 (in progress)
+
+- [bug] JAVA-3002 JAVA-3005: Refresh entire node list when a new node is added
```
akhaku, line 7: we could also just reference one of these tickets and close the other one as a duplicate, let me know

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeStateManager.java
adutra: Indeed, but `addNode` is also async. That's fine.

akhaku:
cc @absurdfarce

absurdfarce:
No worries @akhaku , it is in the process of being reviewed right now :)

absurdfarce (approved):
Changes make perfect sense given the underlying discussion of the issue on JAVA-3002 and JAVA-3005.  Re-ran test scenario described in JAVA-3002 and confirmed that behaviour with this change matches up to exactly what I would expect.

PR #1606, opened 2022-08-18
JAVA-3041 Guava session - cloud bundle support. replaced deprecated constructors

smatvienko-tb:
# Guava session - cloud bundle support

Hey from [ThingsBoard](https://thingsboard.io/) team!
I want to share my experience connecting [Datastax Astra DB cloud with GuavaSession using Thingsboard](https://github.com/thingsboard/thingsboard/pull/7110). 

The old GuavaSession implementation does not connect with cloud Astra DB.
This is because deprecated constructors were in place.
The paper that describes how to implement GuavaSession instead CqlSession is here:
https://docs.datastax.com/en/developer/java-driver/4.4/manual/developer/request_execution/

Here are the screenshots before I tried to connect with the old version:

image

```
com.datastax.oss.driver.api.core.AllNodesFailedException: Could not reach any contact point, make sure you've provided valid addresses (showing first 3 nodes, use getAllErrors() for more): Node(endPoint=8ca57896-18c4-4a61-bb26-e2fc05985081-europe-west1.db.astra.datastax.com:29042:a837fe38-9fd5-49a6-958e-7f7ca71ae890, hostId=null, hashCode=30c14db4): [com.datastax.oss.driver.api.core.connection.ConnectionInitException: [s0|control|id: 0x70374974, L:/10.8.0.66:19408 - R:8ca57896-18c4-4a61-bb26-e2fc05985081-europe-west1.db.astra.datastax.com/34.77.135.39:29042] Protocol initialization request, step 1 (OPTIONS): unexpected failure (com.datastax.oss.driver.api.core.connection.ClosedConnectionException: Lost connection to remote peer)], Node(endPoint=8ca57896-18c4-4a61-bb26-e2fc05985081-europe-west1.db.astra.datastax.com:29042:ba266e06-dfa4-4171-88cb-75758881bcba, hostId=null, hashCode=4ea6381e): [com.datastax.oss.driver.api.core.connection.ConnectionInitException: [s0|control|id: 0xd498c0a3, L:/10.8.0.66:12650 - R:8ca57896-18c4-4a61-bb26-e2fc05985081-europe-west1.db.astra.datastax.com/34.78.6.86:29042] Protocol initialization request, step 1 (OPTIONS): unexpected failure (com.datastax.oss.driver.api.core.connection.ClosedConnectionException: Lost connection to remote peer)], Node(endPoint=8ca57896-18c4-4a61-bb26-e2fc05985081-europe-west1.db.astra.datastax.com:29042:b16cc260-ee81-4017-b6cf-6b793a5b9935, hostId=null, hashCode=61928b01): [com.datastax.oss.driver.api.core.connection.ConnectionInitException: [s0|control|id: 0x0509e0f8, L:/10.8.0.66:39028 - R:8ca57896-18c4-4a61-bb26-e2fc05985081-europe-west1.db.astra.datastax.com/34.140.171.129:29042] Protocol initialization request, step 1 (OPTIONS): unexpected failure (com.datastax.oss.driver.api.core.connection.ClosedConnectionException: Lost connection to remote peer)]

	at com.datastax.oss.driver.api.core.AllNodesFailedException.copy(AllNodesFailedException.java:141)
	at com.datastax.oss.driver.internal.core.util.concurrent.CompletableFutures.getUninterruptibly(CompletableFutures.java:149)
	at com.datastax.oss.driver.api.core.session.SessionBuilder.build(SessionBuilder.java:835)
	at com.datastax.oss.driver.core.session.GuavaToAstraConnectTest.astraConnectWithGuava(GuavaToAstraConnectTest.java:59)
...	
	Suppressed: com.datastax.oss.driver.api.core.connection.ConnectionInitException: [s0|control|id: 0x70374974, L:/10.8.0.66:19408 - R:8ca57896-18c4-4a61-bb26-e2fc05985081-europe-west1.db.astra.datastax.com/34.77.135.39:29042] Protocol initialization request, step 1 (OPTIONS): unexpected failure (com.datastax.oss.driver.api.core.connection.ClosedConnectionException: Lost connection to remote peer)
		at com.datastax.oss.driver.internal.core.channel.ProtocolInitHandler$InitRequest.fail(ProtocolInitHandler.java:356)
		at com.datastax.oss.driver.internal.core.channel.ChannelHandlerRequest.onFailure(ChannelHandlerRequest.java:104)
		at com.datastax.oss.driver.internal.core.channel.InFlightHandler.fail(InFlightHandler.java:381)
		at com.datastax.oss.driver.internal.core.channel.InFlightHandler.abortAllInFlight(InFlightHandler.java:371)
		at com.datastax.oss.driver.internal.core.channel.InFlightHandler.abortAllInFlight(InFlightHandler.java:353)
		at com.datastax.oss.driver.internal.core.channel.InFlightHandler.channelInactive(InFlightHandler.java:331)
...
	Caused by: com.datastax.oss.driver.api.core.connection.ClosedConnectionException: Lost connection to remote peer
```

Now it works like a charm.

image

Here is the code example of how I tested the Astra cloud connection with CqlSession and guavaSession

```java
package com.datastax.oss.driver.core.session;

import com.datastax.oss.driver.api.core.cql.AsyncResultSet;
import com.datastax.oss.driver.api.core.cql.ResultSet;
import com.datastax.oss.driver.api.core.cql.Row;
import com.datastax.oss.driver.example.guava.api.GuavaSession;
import com.datastax.oss.driver.example.guava.api.GuavaSessionUtils;
import org.junit.Test;

import java.nio.file.Path;
import java.nio.file.Paths;
import java.util.concurrent.ExecutionException;

public class GuavaToAstraConnectTest {

  static final Path CLOUD_CONFIG_PATH = Paths.get("/home/your_user/projects/astra/secure-connect-thingsboard.zip");
  static final String CLIENT_ID = "tQPHMztQPHMztQPHMztQPHMz";
  static final String SECRET_ID = "5ejkcF9F.j6f64j71Sr.tiRe0Fsq25ejkcF9F.j6f64j71Sr.tiRe0Fsq25ejkcF9F.j6f64j71Sr.tiRe0Fsq2";

  @Test
  public void astraConnectCqlSession() {
    try (CqlSession session = CqlSession.builder()
            .withCloudSecureConnectBundle(CLOUD_CONFIG_PATH)
            .withAuthCredentials(CLIENT_ID, SECRET_ID)
            .build()) {
      ResultSet rs = session.execute("select release_version from system.local");
      Row row = rs.one();
      
      if (row != null) {
        System.out.println(row.getString("release_version"));
      } else {
        System.out.println("An error occurred.");
      }
    }
  }

  @Test
  public void astraConnectWithGuava() throws ExecutionException, InterruptedException {
    try (GuavaSession session = GuavaSessionUtils.builder()
            .withCloudSecureConnectBundle(CLOUD_CONFIG_PATH)
            .withAuthCredentials(CLIENT_ID, SECRET_ID)
            .build()) {
      AsyncResultSet rs = session.executeAsync("select release_version from system.local").get();
      Row row = rs.one();
      if (row != null) {
        System.out.println(row.getString("release_version"));
      } else {
        System.out.println("An error occurred.");
      }
    }
  }
}
```

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/example/guava/api/GuavaSessionBuilder.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/example/guava/api/GuavaSessionBuilder.java
@@ -18,40 +18,17 @@ package com.datastax.oss.driver.example.guava.api;
 import com.datastax.oss.driver.api.core.CqlSession;
 import com.datastax.oss.driver.api.core.config.DriverConfigLoader;
 import com.datastax.oss.driver.api.core.context.DriverContext;
-import com.datastax.oss.driver.api.core.metadata.Node;
-import com.datastax.oss.driver.api.core.metadata.NodeStateListener;
-import com.datastax.oss.driver.api.core.metadata.schema.SchemaChangeListener;
+import com.datastax.oss.driver.api.core.session.ProgrammaticArguments;
 import com.datastax.oss.driver.api.core.session.SessionBuilder;
-import com.datastax.oss.driver.api.core.tracker.RequestTracker;
-import com.datastax.oss.driver.api.core.type.codec.TypeCodec;
 import com.datastax.oss.driver.example.guava.internal.DefaultGuavaSession;
 import com.datastax.oss.driver.example.guava.internal.GuavaDriverContext;
 import edu.umd.cs.findbugs.annotations.NonNull;
-import java.util.List;
-import java.util.Map;
-import java.util.function.Predicate;
 
 public class GuavaSessionBuilder extends SessionBuilder<GuavaSessionBuilder, GuavaSession> {
 
   @Override
-  protected DriverContext buildContext(
-      DriverConfigLoader configLoader,
-      List<TypeCodec<?>> typeCodecs,
-      NodeStateListener nodeStateListener,
-      SchemaChangeListener schemaChangeListener,
-      RequestTracker requestTracker,
-      Map<String, String> localDatacenters,
-      Map<String, Predicate<Node>> nodeFilters,
-      ClassLoader classLoader) {
-    return new GuavaDriverContext(
-        configLoader,
-        typeCodecs,
-        nodeStateListener,
-        schemaChangeListener,
-        requestTracker,
-        localDatacenters,
-        nodeFilters,
-        classLoader);
+  protected DriverContext buildContext(DriverConfigLoader configLoader, ProgrammaticArguments programmaticArguments) {
+    return new GuavaDriverContext(configLoader, programmaticArguments);
   }
 
   @Override
--- a/integration-tests/src/test/java/com/datastax/oss/driver/example/guava/api/GuavaSessionBuilder.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/example/guava/api/GuavaSessionBuilder.java
@@ -27,7 +27,8 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 public class GuavaSessionBuilder extends SessionBuilder<GuavaSessionBuilder, GuavaSession> {
 
   @Override
-  protected DriverContext buildContext(DriverConfigLoader configLoader, ProgrammaticArguments programmaticArguments) {
+  protected DriverContext buildContext(
+      DriverConfigLoader configLoader, ProgrammaticArguments programmaticArguments) {
     return new GuavaDriverContext(configLoader, programmaticArguments);
   }
 
--- a/integration-tests/src/test/java/com/datastax/oss/driver/example/guava/internal/GuavaDriverContext.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/example/guava/internal/GuavaDriverContext.java
@@ -18,11 +18,7 @@ package com.datastax.oss.driver.example.guava.internal;
 import com.datastax.oss.driver.api.core.config.DriverConfigLoader;
 import com.datastax.oss.driver.api.core.cql.PrepareRequest;
 import com.datastax.oss.driver.api.core.cql.Statement;
-import com.datastax.oss.driver.api.core.metadata.Node;
-import com.datastax.oss.driver.api.core.metadata.NodeStateListener;
-import com.datastax.oss.driver.api.core.metadata.schema.SchemaChangeListener;
-import com.datastax.oss.driver.api.core.tracker.RequestTracker;
-import com.datastax.oss.driver.api.core.type.codec.TypeCodec;
+import com.datastax.oss.driver.api.core.session.ProgrammaticArguments;
 import com.datastax.oss.driver.example.guava.api.GuavaSession;
 import com.datastax.oss.driver.internal.core.context.DefaultDriverContext;
 import com.datastax.oss.driver.internal.core.cql.CqlPrepareAsyncProcessor;
@@ -30,9 +26,6 @@ import com.datastax.oss.driver.internal.core.cql.CqlPrepareSyncProcessor;
 import com.datastax.oss.driver.internal.core.cql.CqlRequestAsyncProcessor;
 import com.datastax.oss.driver.internal.core.cql.CqlRequestSyncProcessor;
 import com.datastax.oss.driver.internal.core.session.RequestProcessorRegistry;
-import java.util.List;
-import java.util.Map;
-import java.util.function.Predicate;
 
 /**
  * A Custom {@link DefaultDriverContext} that overrides {@link #getRequestProcessorRegistry()} to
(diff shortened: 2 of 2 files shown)


ErickRamirezAU:
Hello, @smatvienko-tb. Mica brought this to our attention. Thanks for submitting a PR. And thanks for completing the contributor license agreement (CLA).

Just wondering if you've logged a jira for this. I tried to do a search but couldn't find one. You can [log a ticket here](https://datastax-oss.atlassian.net/browse/JAVA) and link the PR.

We're a little behind but I'll get someone to do a review and let you know. Cheers! 🍻

smatvienko-tb:
Hi, @ErickRamirezAU!
Thanks for the feedback!

Here is the new Jira ticket for this PR: https://datastax-oss.atlassian.net/browse/JAVA-3041

Stay tuned!

ErickRamirezAU:
> Here is the new Jira ticket for this PR: https://datastax-oss.atlassian.net/browse/JAVA-3041

Thanks! 👍

absurdfarce:
So what appears to be happening here is that the necessary config vals for Astra aren't passed through by the old constructor-based mechanism.  [SessionBuilder.buildContext()](https://github.com/datastax/java-driver/blob/4.14.1/core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java#L945-L969) will prefer to use the constructor option if it returns a non-null value; this allows subclasses to define their own context impls.  Problem is that the constructor option doesn't allow for every option to be passed; it looks to have been deprecated before some were added, most relevant to our case before the Astra config values were updated.  And the [calling code](https://github.com/datastax/java-driver/blob/4.14.1/core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java#L903-L906) in this case builds up most of it's data in programmatic arguments before calling buildContext().  So if buildContext() doesn't have _every_ value to be preserved passed as an argument it won't be able to create a ProgrammaticBuilder instance that includes them.  Upshot: any values that aren't passed as args to buildContext() are basically left out.

To prove this out try the following change:

```diff
diff --git a/core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java b/core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
index 966372fa2..a6eaa0c99 100644
--- a/core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/session/SessionBuilder.java
@@ -60,6 +60,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.Callable;
@@ -900,8 +901,10 @@ public abstract class SessionBuilder<SelfT extends SessionBuilder, SessionT> {
             CqlIdentifier.fromCql(defaultConfig.getString(DefaultDriverOption.SESSION_KEYSPACE));
       }
 
+      DriverContext ctx = buildContext(configLoader, programmaticArguments);
+      Optional<SslEngineFactory> sslFactory = ctx.getSslEngineFactory();
       return DefaultSession.init(
-          (InternalDriverContext) buildContext(configLoader, programmaticArguments),
+              (InternalDriverContext) ctx,
           contactPoints,
           keyspace);
``` 

Take a look at sslFactory in a debugger.  When using the current 4.14.1 impl this value is an empty option; the SSL factory configs aren't preserved in the args passed to GuavaSessionBuilder.buildSession().  Applying the changes in this PR you'll see that the sslFactory is now a non-empty value.  This also explains the observed error; because the SSL configs aren't included Astra hangs up before we even get a chance to pass in an OPTIONS message.

Note that it's the combination of the Guava session with Astra that brings out the problem here.  The example in the docs should work fine if Astra isn't involved.

Need to think a bit about the best way to address this problem.  It's probably okay to make the change contained in this PR but at a minimum we'll also want to update the docs at the same time.

smatvienko-tb:
You are right. If the Guava uses the old super constructor and returns non-null, the new features are not reachable.
image
image

smatvienko-tb:
The documentation seems to be fine. It does not expose the details and refers directly to the git repo. Probably updating the repo is just fine for this page: https://docs.datastax.com/en/developer/java-driver/4.4/manual/developer/request_execution/
image

Connecting to the Astra example is welcomed. 
We found that cloud Keyspace has to be created manually, and install scripts have to not create keyspace for the cloud. 
It is ok, but not clear when we have a first experience with the cloud.
And for the cloud we should not use the local DC, keyspace, URL, and SSL parameters that come from the cloud. 
It is also not clear on the first steps after using the local Cassandra.

Probably, the better compatibility with the install scripts for local Cassandra will be accepting "CREATE KEYSPACE IF NOT EXISTS" and report success if the keyspace already exists.

absurdfarce:
Apologies @smatvienko-tb , I've been off working on other projects for the past little while. :(

It's the docs you cite above that originally had me concerned.  Note the ambiguity of the "buildContext(...)" call in the sample code you cited:

```java
public class GuavaSessionBuilder extends SessionBuilder<GuavaSessionBuilder, GuavaSession> {

  @Override
  protected DriverContext buildContext( ... ) { ... }

  @Override
  protected GuavaSession wrap(CqlSession defaultSession) {
    return new DefaultGuavaSession(defaultSession);
  }
```

My concern was that this code wasn't clearly steering users towards the ProgrammaticArguments version introduced in [JAVA-2315](https://datastax-oss.atlassian.net/browse/JAVA-2315).  The more I look at it there more I'm inclined to agree with you, though; I think the docs are probably fine.  They point users at the buildContext() method generally, and from there it should be fairly clear which method should be overridden... especially if our sample code now uses the correct method (which it does with your changes in this PR).

I may still rethink this and change the docs later to show an explicit override of buildContext(ProgrammaticArguments) but it's probably okay for now.

absurdfarce:
Thanks for the contribution (and for the dialogue) @smatvienko-tb !

PR #1605, opened 2022-07-08
Update branding

msmygit:

PR #1608, opened 2022-09-30
Standardize on DataStax vs Datastax since that appears to be more common

akhaku:
prior to this change:
```
akhaku@host:~/datastax-java-driver>(4.x) git grep Datastax | wc -l
      27
akhaku@host:~/datastax-java-driver>(4.x) git grep DataStax | wc -l
    1999
```


akhaku:
Looks like #1605 is similar

absurdfarce:
Thanks @akhaku.  An increase in consistency is always welcome!

PR #1609, opened 2022-10-01
JAVA-3033 - Consider handling IOException in a manner consistent with other exceptions in CloudConfigFactory.fetchProxyMetadata()

tongjixianing:


tongjixianing:
This PR originally is about handling the IOE from HttpsURLConnection.getInputStream() call in fetchProxyMetaData method which was triggered by 401 request error. During the Unit test, we also found it also needs to handle the FileNotFoundException triggered by 404 error code when the metadata request URL is not correct or the service cannot be located.

hhughes:
Taking over this PR so we can close out this feature branch.
- rebased with latest 4.x
- moved FNF handling to earlier catch block so it doesn't get grabbed by the clause catching it's superclass IOE
- Use exact exception message matching in the unit test to make sure that the correct ISEs are thrown
- Added test for 502 bad gateway which was the error I got from astra when fiddling with the `/metadata` path
- Removed wildcard exceptions

core/src/main/java/com/datastax/oss/driver/internal/core/config/cloud/CloudConfigFactory.java
```diff
@@ -228,8 +229,18 @@ protected BufferedReader fetchProxyMetadata(
...
       connection.setRequestProperty("host", "localhost");
-      return new BufferedReader(
-          new InputStreamReader(connection.getInputStream(), StandardCharsets.UTF_8));
+      try {
+        return new BufferedReader(
+            new InputStreamReader(connection.getInputStream(), StandardCharsets.UTF_8));
+      } catch (FileNotFoundException e) {
+        throw new IllegalStateException(
+            "Metadata service request path not found. Please make sure the metadata service url is correct",
+            e);
```
absurdfarce, line 238: This change relies on undocumented behaviour (HttpsURLConnection throwing FNFE in this case).  We should either find a more conventional way to note this case or else remove the special logging for this specific code path.

core/src/main/java/com/datastax/oss/driver/internal/core/config/cloud/CloudConfigFactory.java
```diff
@@ -228,8 +229,18 @@ protected BufferedReader fetchProxyMetadata(
...
+      } catch (IOException ioe) {
+        throw new IllegalStateException(
+            "Unable to read data from cloud metadata service. Please make sure your cluster is not parked or terminated",
+            ioe);
+      }
```
absurdfarce, line 243: I'd like for more targeted catch blocks here.  As it is the exceptions indicated here could be thrown by either HttpsURLConneciton.getInputStream(), the InputStreamReader constructor or the BufferedReader constructor.

PR #1610, opened 2022-10-07
Add onAuthenticationFailure callback to Authenticator

iyogi:
Add `onAuthenticationFailure()` callback to the Authenticator lifecycle methods so that authenticator instances are made aware of authentication failures as well - which would truly complete the lifecycle callbacks.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/auth/Authenticator.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/auth/Authenticator.java
@@ -39,6 +39,9 @@ import java.util.concurrent.CompletionStage;
  *   <li>When the server indicates that authentication is successful, the {@link
  *       #onAuthenticationSuccess} method will be called with the last information that the server
  *       may optionally have sent.
+ *   <li>When the server indicates that authentication has failed, the {@link
+ *       #onAuthenticationFailure} method will be called with {@link AuthenticationException}
+ *       encapsulating the server error message
  * </ol>
  *
  * The exact nature of the negotiation between client and server is specific to the authentication
@@ -93,4 +96,11 @@ public interface Authenticator {
    */
   @NonNull
   CompletionStage<Void> onAuthenticationSuccess(@Nullable ByteBuffer token);
+
+  /**
+   * Called when authentication fails.
+   *
+   * @param exception contains authentication error message
+   */
+  default void onAuthenticationFailure(AuthenticationException exception) {}
 }
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/channel/ProtocolInitHandler.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/channel/ProtocolInitHandler.java
@@ -280,12 +280,14 @@ class ProtocolInitHandler extends ConnectInitHandler {
         } else if (step == Step.AUTH_RESPONSE
             && response instanceof Error
             && ((Error) response).code == ProtocolConstants.ErrorCode.AUTH_ERROR) {
-          fail(
+          AuthenticationException cause =
               new AuthenticationException(
                   endPoint,
                   String.format(
                       "server replied with '%s' to AuthResponse request",
-                      ((Error) response).message)));
+                      ((Error) response).message));
+          authenticator.onAuthenticationFailure(cause);
+          fail(cause);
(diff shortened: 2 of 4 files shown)


absurdfarce:
Thanks for the PR @iyogi!

iyogi:
> Thanks for the PR @iyogi!
> 
> Have you signed the Contributor License Agreement for contributions to DataStax open source projects? If not you can find it at https://cla.datastax.com/. Thanks!

Yes, have just signed it.

core/src/main/java/com/datastax/oss/driver/api/core/auth/Authenticator.java
```diff
@@ -93,4 +96,11 @@ public interface Authenticator {
    */
   @NonNull
   CompletionStage<Void> onAuthenticationSuccess(@Nullable ByteBuffer token);
+
+  /**
+   * Called when authentication fails.
+   *
+   * @param exception contains authentication error message
+   */
+  default void onAuthenticationFailure(AuthenticationException exception) {}
```
absurdfarce, line 105: This should return a CompletionStage<Void> in order to better fit in with the other methods on this interface.  I agree with making it a default method (since that avoids breaking existing impls) but the default should return something like `new CompletableFuture().complete(null)`

core/src/main/java/com/datastax/oss/driver/internal/core/channel/ProtocolInitHandler.java
```diff
@@ -280,12 +280,14 @@ void onResponse(Message response) {
...
-          fail(
+          AuthenticationException cause =
               new AuthenticationException(
                   endPoint,
                   String.format(
                       "server replied with '%s' to AuthResponse request",
-                      ((Error) response).message)));
+                      ((Error) response).message));
+          authenticator.onAuthenticationFailure(cause);
+          fail(cause);
```
absurdfarce, line 290: Keep in mind that Authenticator.onAuthenticationFailure() could itself throw an exception.  We should handle that case gracefully here and throw a different AuthenticationException if that occurs.  That's the point of this code pattern elsewhere within ProtocolInitHandler:

```java
          authenticator
              .onAuthenticationSuccess(token)
              .whenCompleteAsync(
                  (ignored, error) -> {
                    if (error != null) {
                      fail(
                          new AuthenticationException(
                              endPoint,
                              String.format(
                                  "Authenticator.onAuthenticationSuccess(): stage completed exceptionally (%s)",
                                  error),
                              error));
                    } else {
                      step = Step.GET_CLUSTER_NAME;
                      send();
                    }
                  },
                  channel.eventLoop())
              .exceptionally(UncaughtExceptions::log);
```

We should probably do something similar here in order to clearly distinguish between these two cases.  Note that the API change re: making onAuthenticationFailure() return a CompletionStage facilitates such a change.

core/src/test/java/com/datastax/oss/driver/internal/core/channel/ProtocolInitHandlerTest.java
```diff
@@ -409,6 +409,49 @@ public void should_fail_to_initialize_if_server_sends_auth_error() throws Throwa
...
+    writeInboundFrame(
+        requestFrame, new Error(ProtocolConstants.ErrorCode.AUTH_ERROR, "mock error"));
+
+    assertThat(connectFuture)
+        .isFailed(e -> assertThat(e).isInstanceOf(AuthenticationException.class));
+
+    // verify that onAuthenticationFailure callback was invoked
+    assertThat(authenticator.authFailure).isTrue();
+  }
+
```
absurdfarce, line 454: This test winds up being almost identical to should_fail_to_initialize_if_server_sends_auth_error().  I guess for me it's probably cleaner to add a test for this functionality into the existing test (by adding an impl of onAuthenticationFailure() to the mock) and changing it's name to something that indicates we're validating all behaviours in the AuthError case rather than just the failure to initialize.

Does that make sense?

absurdfarce (changes requested):
Overall I like this change quite a bit.  The enhancement of the symmetry of Authenticator is probably worth it by itself.  I do think we need to make the impl more consistent with the rest of Authenticator but hopefully that shouldn't be an enormous change.

Thanks again @iyogi !

PR #1614, opened 2022-10-31
Allow the use of CODECs with primitive getters

chibenwa:
From what I observe a typical Cassandra application spends a few percent
CPU time computing CODEC. This computation can typically be avoided
by passing manually the codec that matches the data types.

In this draft pull request, I propose to allow similar optimisation for primitive getX methods, 
withoutpaying the price of unboxing X.class.

I did the change for `Long` and `Boolean` so far.

If this is consensual enough, I will carry other the change to the other types (Float, Double, Short, Int, Byte) and take care of this change for setter as well to get a well rounded API.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/data/GettableById.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/data/GettableById.java
@@ -197,6 +197,29 @@ public interface GettableById extends GettableByIndex, AccessibleById {
   }
 
   /**
+   * Returns the value for the first occurrence of {@code id} as a Java primitive boolean
+   * using the provided {@code codec}.
+   *
+   * <p>By default, this works with CQL type {@code boolean}.
+   *
+   * <p>Note that, due to its signature, this method cannot return {@code null}. If the CQL value is
+   * {@code NULL}, it will return {@code false}. If this doesn't work for you, either call {@link
+   * #isNull(CqlIdentifier)} before calling this method, or use {@code get(id, Boolean.class)}
+   * instead.
+   *
+   * <p>If an identifier appears multiple times, this can only be used to access the first value.
+   * For the other ones, use positional getters.
+   *
+   * <p>If you want to avoid the overhead of building a {@code CqlIdentifier}, use the variant of
+   * this method that takes a string argument.
+   *
+   * @throws IllegalArgumentException if the id is invalid.
+   */
+  default boolean getBoolean(@NonNull CqlIdentifier id, TypeCodec<Boolean> codec) {
+    return getBoolean(firstIndexOf(id), codec);
+  }
+
+  /**
    * @deprecated this method only exists to ease the transition from driver 3, it is an alias for
    *     {@link #getBoolean(CqlIdentifier)}.
    */
--- a/core/src/main/java/com/datastax/oss/driver/api/core/data/GettableById.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/data/GettableById.java
@@ -314,6 +314,27 @@ public interface GettableById extends GettableByIndex, AccessibleById {
   }
 
   /**
+   * Returns the value for the first occurrence of {@code id} as a Java primitive long using the given {@code codec}.
+   *
+   * <p>By default, this works with CQL types {@code bigint} and {@code counter}.
+   *
+   * <p>Note that, due to its signature, this method cannot return {@code null}. If the CQL value is
+   * {@code NULL}, it will return {@code 0}. If this doesn't work for you, either call {@link
+   * #isNull(CqlIdentifier)} before calling this method, or use {@code get(id, codec)} instead.
+   *
+   * <p>If an identifier appears multiple times, this can only be used to access the first value.
+   * For the other ones, use positional getters.
+   *
+   * <p>If you want to avoid the overhead of building a {@code CqlIdentifier}, use the variant of
+   * this method that takes a string argument.
+   *
+   * @throws IllegalArgumentException if the id is invalid.
+   */
+  default long getLong(@NonNull CqlIdentifier id, TypeCodec<Long> codec) {
+    return getLong(firstIndexOf(id), codec);
+  }
+
+  /**
    * Returns the value for the first occurrence of {@code id} as a Java primitive short.
    *
    * <p>By default, this works with CQL type {@code smallint}.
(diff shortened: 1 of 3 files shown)


absurdfarce:
Thanks for the PR (and the other one!) @chibenwa!

chibenwa:
Signed as shown here https://github.com/datastax/java-driver/pull/656

chibenwa:
Should I conclude there is interest in this work (typeCodecs for primitives) and that this is worth generalizing this to other getters / setters?

absurdfarce:
Apologies @chibenwa , the CLA signing is a very necessary first step.  I had planned on reviewing your PR once we knew we were good with the CLA.

I have since taken a look at your PR and honestly I'm a little bit torn.  In general I like your idea and my inclination is to proceed with this work, but I am just a bit hesitant at the prospect of adding a host of new methods to the GettableBy* interfaces.  As it stands we're talking about adding only seven new methods, but note that we're still doing an instanceof check and a cast from `TypeCodec<Boolean>` to `PrimitiveBooleanCodec`.  We can avoid even that with an extra set of methods, something along the lines of the following:

```patch
diff --git a/core/src/main/java/com/datastax/oss/driver/api/core/data/GettableById.java b/core/src/main/java/com/datastax/oss/driver/api/core/data/GettableById.java
index 2aad16d29..591c81c03 100644
--- a/core/src/main/java/com/datastax/oss/driver/api/core/data/GettableById.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/data/GettableById.java
@@ -18,6 +18,7 @@ package com.datastax.oss.driver.api.core.data;
 import com.datastax.oss.driver.api.core.CqlIdentifier;
 import com.datastax.oss.driver.api.core.metadata.token.Token;
 import com.datastax.oss.driver.api.core.type.codec.CodecNotFoundException;
+import com.datastax.oss.driver.api.core.type.codec.PrimitiveBooleanCodec;
 import com.datastax.oss.driver.api.core.type.codec.TypeCodec;
 import com.datastax.oss.driver.api.core.type.reflect.GenericType;
@@ -219,6 +220,10 @@ public interface GettableById extends GettableByIndex, AccessibleById {
     return getBoolean(firstIndexOf(id), codec);
   }
 
+  default boolean getBoolean(@NonNull CqlIdentifier id, PrimitiveBooleanCodec codec) {
+    return getBoolean(firstIndexOf(id), codec);
+  }
+
   /**
    * @deprecated this method only exists to ease the transition from driver 3, it is an alias for
    *     {@link #getBoolean(CqlIdentifier)}.
diff --git a/core/src/main/java/com/datastax/oss/driver/api/core/data/GettableByIndex.java b/core/src/main/java/com/datastax/oss/driver/api/core/data/GettableByIndex.java
index e861cc4f6..29d8ee912 100644
--- a/core/src/main/java/com/datastax/oss/driver/api/core/data/GettableByIndex.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/data/GettableByIndex.java
@@ -183,13 +183,17 @@ public interface GettableByIndex extends AccessibleByIndex {
    */
   default boolean getBoolean(int i, TypeCodec<Boolean> codec) {
     if (codec instanceof PrimitiveBooleanCodec) {
-      return ((PrimitiveBooleanCodec) codec).decodePrimitive(getBytesUnsafe(i), protocolVersion());
+      return getBoolean(i, (PrimitiveBooleanCodec) codec);
     } else {
       Boolean value = get(i, codec);
       return value == null ? false : value;
     }
   }
 
+  default boolean getBoolean(int i, PrimitiveBooleanCodec codec) {
+      return codec.decodePrimitive(getBytesUnsafe(i), protocolVersion());
+  }
+
   /**
    * @deprecated this method only exists to ease the transition from driver 3, it is an alias for
    *     {@link #getBoolean(int)}.
```

This further optimizes the process but now has us adding 14 methods to the interface rather than 7.

It's also worth noting that there _is_ a way for performance-conscious applications to do something like this now.  Rather than tweak the API to provide multiple access points for different types of codecs you can also get the relevant bytes and handle the decode manually on the back end.  It's not ideal (and not something that may be desirable to expose as a general pattern) but it is at least a viable solution:

```java
    public static void main(String[] args) {

        InetAddress localhost = null;
        try { localhost = InetAddress.getByName("127.0.0.1"); }
        catch (UnknownHostException uhe) {

            System.out.println("Couldn't find localhost!");
        }

        try (CqlSession session = CqlSession
                .builder()
                .addContactPoint(new InetSocketAddress(localhost, 9042))
                .withLocalDatacenter("datacenter1")
                .build()) {
            ResultSet rs = session.execute("select count(*) from system_schema.keyspaces");
            Row row = rs.one();
            long count =
                    TypeCodecs.BIGINT.decodePrimitive(
                            row.getBytesUnsafe(row.firstIndexOf("count")),
                            session.getContext().getProtocolVersion());
            System.out.println(String.format("Answer: %s keyspaces", count));
        }
    }
```

One could credibly argue that this is a better solution for the (possibly small?) amount of apps that care about performance to this degree.  The counter-argument, of course, is that it's not like we're talking about re-writing the GetBy* interfaces... we're just talking about expanding them.

As mentioned above my _inclination_ is to just add the new methods and move on but it seemed worth at least talking about.  @adutra do you have any opinions here?

chibenwa:
> Apologies @chibenwa , the CLA signing is a very necessary first step. I had planned on reviewing your PR once we knew we were good with the CLA. 

As said, it is already processed:

Screenshot from 2022-11-07 09-34-17

> I am just a bit hesitant at the prospect of adding a host of new methods to the GettableBy* interfaces.

I agree this PR have limited value and brings significant complexity. An alternative exists, and as such proposes to close it. (thanks for the reply). IMO the syntactic sugar is not worth the API pollution. + soon that gonna be a non-problem with vallhala...

> One could credibly argue that this is a better solution for the (possibly small?) amount of apps that care about performance to this degree. 

Maybe it is worth developing a bit further my use case. I am working on a Cassandra based email server. IMAP servers allows addressing emails by their position in a mailbox, and as such upon selects we need to load the following mapping: `identifier -> position`. Given that some users have millions of emails in their INBOX (yep) then these small gains adds up quickly and represents ~10% of CPU time (measured with flame graphs) on a production set up...

Here is a screenshot of a flame graph of an IMAP FETCH on messages metadata (yep some bad email clients use that for re-synchronization...)

Screenshot from 2022-11-07 09-21-12

 - `CachingCodecRgistry::codecFor` : 19%
 - `DefaultRow::protocolVersion`: 6.5%
 - `DefaultRow::codecRegistry`: 4.5%
 - `DefaultRow::firstIndexOf`: 50.2% CF https://github.com/datastax/java-driver/pull/1615 (that's where big gains are, without API changes, without downsides, and for all applications <3 )
 
Total: 80% CPU gains on row decoding...

chibenwa:
As said in last week comment the changeset I proposed is bitter sweat and the alternatives of `row.getByteUnsafe + manual codec` usage is likely enough for thiose that might want the last bits of performance while waiting for project valhala...

As such, I am closing this issue.

absurdfarce:
Makes sense @chibenwa.  Thanks for the obvious effort you put into this on both the analysis and implementation front!

PR #1616, opened 2022-11-08
Optimizations for TypeSafeDriverConfig

chibenwa:
Small but easy gains...

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfig.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfig.java
@@ -139,11 +139,12 @@ public class TypesafeDriverConfig implements DriverConfig {
   @NonNull
   @Override
   public DriverExecutionProfile getProfile(@NonNull String profileName) {
+    TypesafeDriverExecutionProfile.Base profile = profiles.get(profileName);
     Preconditions.checkArgument(
-        profiles.containsKey(profileName),
+        profile != null,
         "Unknown profile '%s'. Check your configuration.",
         profileName);
-    return profiles.get(profileName);
+    return profile;
   }
 
   @NonNull
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfig.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfig.java
@@ -32,6 +32,7 @@ import com.typesafe.config.ConfigValueFactory;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.net.URL;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import net.jcip.annotations.ThreadSafe;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfig.java
```diff
@@ -136,14 +139,20 @@ private Map<String, Config> extractProfiles(Config sourceConfig) {
...
   @Override
   public DriverExecutionProfile getProfile(@NonNull String profileName) {
+    TypesafeDriverExecutionProfile.Base profile = profiles.get(profileName);
     Preconditions.checkArgument(
-        profiles.containsKey(profileName),
+        profile != null,
         "Unknown profile '%s'. Check your configuration.",
         profileName);
-    return profiles.get(profileName);
+    return profile;
```
absurdfarce: Use of Preconditions.checkArgument() seems a bit strange now given that we already got the key above but such is the weirdness of dealing with null.  This answer doesn't seem better due to the intermediate Optional object:

```java
    TypesafeDriverExecutionProfile.Base profile = profiles.get(profileName);
    return Optional.ofNullable(profile).orElseThrow(() -> new IllegalArgumentException(String.format("Unknown profile '%s'. Check your configuration.", profileName)));
```

My kingdom for a basic Map.getOrThrow()

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfig.java
```diff
@@ -136,14 +139,20 @@ private Map<String, Config> extractProfiles(Config sourceConfig) {
...
 
+  @Override
+  public DriverExecutionProfile getDefaultProfile() {
+    return defaultProfile;
+  }
+
   @NonNull
   @Override
   public DriverExecutionProfile getProfile(@NonNull String profileName) {
+    TypesafeDriverExecutionProfile.Base profile = profiles.get(profileName);
```
absurdfarce: If we keep defaultProfile as a member we already have the default case covered here.  Maybe before this we should add a check for profileName.equals(DriverExecutionProfile.DEFAULT_NAME) and return this.defaultProfile if it does?

core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/TypesafeDriverConfig.java
```diff
@@ -62,6 +64,7 @@ public TypesafeDriverConfig(Config config) {
           new TypesafeDriverExecutionProfile.Base(entry.getKey(), entry.getValue()));
     }
     this.profiles = builder.build();
+    this.defaultProfile = profiles.get(DriverExecutionProfile.DEFAULT_NAME);
```
absurdfarce, line 68: Making this stateful doesn't impact config reloading.  Makes sense given the refresh() op in TypesafeDriverConfig.reload() but it seemed worth validating in an explicit test:

```java
  @Test
  public void should_update_default_profile_on_reload() {
    TypesafeDriverConfig config = parse("int1 = 42\n profiles { profile1 { int1 = 43 } }");
    assertThat(config.getDefaultProfile().getInt(MockOptions.INT1)).isEqualTo(42);
    config.reload(ConfigFactory.parseString("int1 = 44\n profiles { profile1 { int1 = 45 } }"));
    assertThat(config.getDefaultProfile().getInt(MockOptions.INT1)).isEqualTo(44);
  }
```

Might be a good idea to add something like this regardless.

absurdfarce (approved):
Looks pretty good to me.  A couple fairly small things I'd like to get @chibenwa 's feedback on but none of it is essential; I'd be fine putting this in as it stands.

absurdfarce:
Looks good!  Thanks (again) @chibenwa!

PR #1617, opened 2022-11-08
Improve ByteBufPrimitiveCodec readBytes

chibenwa:
-> Remove a needless call to `ByteBuff::hasArray`
 
  -> Removes needless slicing

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufPrimitiveCodec.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufPrimitiveCodec.java
@@ -115,7 +115,9 @@ public class ByteBufPrimitiveCodec implements PrimitiveCodec<ByteBuf> {
     int length = readInt(source);
     if (length < 0) return null;
     ByteBuf slice = source.readSlice(length);
-    return ByteBuffer.wrap(readRawBytes(slice));
+    byte[] bytes = new byte[slice.readableBytes()];
+    slice.readBytes(bytes);
+    return ByteBuffer.wrap(bytes);
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufPrimitiveCodec.java
```diff
@@ -220,22 +222,6 @@ public void writeShortBytes(byte[] bytes, ByteBuf dest) {
...
 
-  // Reads *all* readable bytes from a buffer and return them.
-  // If the buffer is backed by an array, this will return the underlying array directly, without
-  // copy.
-  private static byte[] readRawBytes(ByteBuf buffer) {
-    if (buffer.hasArray() && buffer.readableBytes() == buffer.array().length) {
-      // Move the readerIndex just so we consistently consume the input
-      buffer.readerIndex(buffer.writerIndex());
-      return buffer.array();
-    }
```
absurdfarce, line 231: @chibenwa I'm not sure I understand.  Is there a reason why we don't want to use the underlying byte array directly if we can get access to it?

chibenwa: Because the conditions for reusing the array are never fullfilled.

The buffer layout is `4 bytes size, byyyyyyyyyyyyytes` so we would expect the underlying array to match (best case) this layout.

However reading the size means that the leading 4 bytes are read first, and then the payload never matches the underlying array.

Checking if this optimization is doable is impactful enough to be removing it.

Clearer?

absurdfarce: Apologies @chibenwa, I've been out for the holiday here... just getting back to this.

Yeah, your explanation makes sense, thanks for the additional context.  I'm reluctant to give up the re-use of the backing array, though, so I put together an [alternate implementation](https://github.com/datastax/java-driver/pull/1621) which preserves that feature.  What do you think of something in that vein?

chibenwa:
I added some tests to better specify the expected behaviour of `readBytes` and removed the needless test.

core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufPrimitiveCodec.java
absurdfarce: The alternate proposal never went anywhere so we're moving forward with the original proposal from @chibenwa

PR #1619, opened 2022-11-16
JAVA-3030 Use configuration to determine batch-type instead of settin…

Maxwell-Guo:
JAVA-3030 Use configuration to determine batch-type instead of setting on a per-batch transaction basis

### DIFF
--- a/core/src/main/java/com/datastax/dse/driver/api/core/config/DseDriverOption.java
+++ b/core/src/main/java/com/datastax/dse/driver/api/core/config/DseDriverOption.java
@@ -288,6 +288,13 @@ public enum DseDriverOption implements DriverOption {
    * <p>Value-type: {@link java.time.Duration Duration}
    */
   METRICS_NODE_GRAPH_MESSAGES_SLO("advanced.metrics.node.graph-messages.slo"),
+
+  /**
+   * The batch type that is set in the configuration file
+   *
+   * <p>Value-type: {@link String}
+   */
+  BATCH_TYPE_CONFIGURATION("advanced.batch-type-configuration"),
   ;
 
   private final String path;
--- a/core/src/main/java/com/datastax/oss/driver/api/core/cql/BatchStatement.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/cql/BatchStatement.java
@@ -41,6 +41,94 @@ import java.util.Collections;
 public interface BatchStatement extends Statement<BatchStatement>, Iterable<BatchableStatement<?>> {
 
   /**
+   * Creates an instance of the default implementation with null batch type.
+   *
+   * <p>Note that the returned object is <b>immutable</b>.
+   */
+  static BatchStatement newInstance() {
+    return new DefaultBatchStatement(
+        BatchType.UNKNOW,
+        new ArrayList<>(),
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
+        Collections.emptyMap(),
+        null,
+        false,
+        Statement.NO_DEFAULT_TIMESTAMP,
+        null,
+        Integer.MIN_VALUE,
+        null,
+        null,
+        null,
+        null,
+        Statement.NO_NOW_IN_SECONDS);
+  }
+
+  /**
+   * Creates an instance of the default implementation for the null batch type, containing the given
+   * statements.
+   *
+   * <p>Note that the returned object is <b>immutable</b>.
+   */
+  @NonNull
+  static BatchStatement newInstance(@NonNull Iterable<BatchableStatement<?>> statements) {
+    return new DefaultBatchStatement(
+        BatchType.UNKNOW,
+        ImmutableList.copyOf(statements),
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
+        Collections.emptyMap(),
+        null,
+        false,
+        Statement.NO_DEFAULT_TIMESTAMP,
+        null,
+        Integer.MIN_VALUE,
+        null,
+        null,
+        null,
+        null,
+        Statement.NO_NOW_IN_SECONDS);
+  }
+
+  /**
+   * Creates an instance of the default implementation for the null batch type, containing the given
+   * statements.
+   *
+   * <p>Note that the returned object is <b>immutable</b>.
+   */
+  @NonNull
+  static BatchStatement newInstance(@NonNull BatchableStatement<?>... statements) {
+    return new DefaultBatchStatement(
+        BatchType.UNKNOW,
+        ImmutableList.copyOf(statements),
+        null,
+        null,
+        null,
+        null,
+        null,
+        null,
+        Collections.emptyMap(),
+        null,
+        false,
+        Statement.NO_DEFAULT_TIMESTAMP,
+        null,
+        Integer.MIN_VALUE,
+        null,
+        null,
+        null,
+        null,
+        Statement.NO_NOW_IN_SECONDS);
+  }
+
+  /**
    * Creates an instance of the default implementation for the given batch type.
    *
    * <p>Note that the returned object is <b>immutable</b>.
@@ -129,6 +217,15 @@ public interface BatchStatement extends Statement<BatchStatement>, Iterable<Batc
         null,
         null,
         Statement.NO_NOW_IN_SECONDS);
+  }
+
+  /**
+   * Returns a builder to create an instance of the default implementation.
+   *
+   * <p>Note that this builder is mutable and not thread-safe.
+   */
+  static BatchStatementBuilder builder() {
+    return new BatchStatementBuilder(BatchType.UNKNOW);
   }
 
   /**
--- a/core/src/main/java/com/datastax/oss/driver/api/core/cql/DefaultBatchType.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/cql/DefaultBatchType.java
@@ -37,6 +37,11 @@ public enum DefaultBatchType implements BatchType {
    * it can only contain these.
    */
   COUNTER(ProtocolConstants.BatchType.COUNTER),
+
+  /** UNKNOW BATCH TYPE when user do not set at BatchStatement builder */
+  // TODO should we set the number 3 be a constant at some where or ProtocolConstants.BatchType add
+  // a unknow type ?
+  UNKNOW((byte) 3),
   ;
   // Note that, for the sake of convenience, we also expose shortcuts to these constants on the
   // BatchType interface. If you add a new enum constant, remember to update the interface as
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/BatchTypeRegistry.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/BatchTypeRegistry.java
@@ -1,0 +1,25 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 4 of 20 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/cql/DefaultBatchType.java
```diff
@@ -37,6 +37,11 @@ public enum DefaultBatchType implements BatchType {
    * it can only contain these.
    */
   COUNTER(ProtocolConstants.BatchType.COUNTER),
+
+  /** UNKNOW BATCH TYPE when user do not set at BatchStatement builder */
+  // TODO should we set the number 3 be a constant at some where or ProtocolConstants.BatchType add
+  // a unknow type ?
+  UNKNOW((byte) 3),
```
absurdfarce, line 44: Items in the DefaultBatchType represent the set of batch types supported by the driver by default... which should be equal to what's defined in the v5 protocol spec. If you wanted to define a custom batch type here for some reason then this should be done in com.datastax.oss.driver.api.core.cql.BatchType only.

That said, it's far from clear to me that we need to do such a thing here.

core/src/main/java/com/datastax/oss/driver/internal/core/BatchTypeRegistry.java
```diff
@@ -0,0 +1,25 @@
...
+package com.datastax.oss.driver.internal.core;
+
+import com.datastax.oss.driver.api.core.cql.BatchType;
+
+public interface BatchTypeRegistry {
+  BatchType fromName(String name);
+
+  /** @return all the values known to this driver instance. */
+  Iterable<BatchType> getValues();
+}
```
absurdfarce, line 25: This is the kind of change that sets off warning bells for me. There needs to be a pretty high threshold for adding new items to the DriverContext and it's far from clear to me that the need for a registry for batch types meets that threshold.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -533,6 +535,33 @@ public static CoordinatorException toThrowable(
...
+          throw new IllegalArgumentException(
+              "IllegalArgumentException occurs for batch type option "
+                  + DseDriverOption.BATCH_TYPE_CONFIGURATION.getPath()
+                  + "'s value in configuration file is illegal when use do not set BATCH TYPE at BatchStatement build stage. ",
+              throwable);
+        }
+        return batchStatement.setBatchType(batchType);
+      }
+    }
+    return statement;
```
absurdfarce, line 562: Probably makes sense to switch the order around here; if the statement isn't a batch statement just return, otherwise do the logic above. In most cases batch statements will likely not be the majority of statements executed so exiting quickly for non-batches seems desirable.

integration-tests/src/test/resources/application.conf
```diff
@@ -17,6 +17,7 @@ datastax-java-driver {
       init-query-timeout = 5 seconds
       set-keyspace-timeout = 5 seconds
     }
+    batch-type-configuration = UNLOGGED
```
absurdfarce, line 20: This covers the config for integration testing only. We also should have an entry (ideally with documentation) in reference.conf as well.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
```diff
@@ -138,11 +138,10 @@ protected CqlRequestHandler(
     this.startTimeNanos = System.nanoTime();
     this.logPrefix = sessionLogPrefix + "|" + this.hashCode();
     LOG.trace("[{}] Creating new handler for request {}", logPrefix, statement);
-
-    this.initialStatement = statement;
     this.session = session;
     this.keyspace = session.getKeyspace().orElse(null);
     this.context = context;
+    this.initialStatement = Conversions.resolveStatement(statement, context);
```
absurdfarce, line 144: I agree that this is _one_ way to address the problem: before you send a batch statement make sure that you've got a batch type.  But this has a couple of very unpleasant consequences:

* You _have_ to introduce something like an "unknown" batch type to represent the state of batches which haven't had anything specified yet because they're going to use the default at execution time.
* In driver 4.x statements are immutable, so every time you have to use the default batch type you're creating a copy of the input BatchStatement object with exactly one field changed.

The much simpler answer is to enforce the presence of a batch type _at creation time_.  The builder interface at BatchStatement.builder() already requires a non-null BatchType so all we need to worry about are the BatchStatement.newInstance() types.  In that case it seems like we can make a modified version which accepts a non-null DriverContext as a param instead of a BatchType.  We can then use that DriverContext to do a lookup for the configured default batch type.  So long as our config defines a default (which it certainly should) we'll always generate a BatchStatement with a non-null batch type using that method.

core/src/main/java/com/datastax/dse/driver/api/core/config/DseDriverOption.java
```diff
@@ -288,6 +288,13 @@ public enum DseDriverOption implements DriverOption {
    * <p>Value-type: {@link java.time.Duration Duration}
    */
   METRICS_NODE_GRAPH_MESSAGES_SLO("advanced.metrics.node.graph-messages.slo"),
+
+  /**
+   * The batch type that is set in the configuration file
+   *
+   * <p>Value-type: {@link String}
+   */
+  BATCH_TYPE_CONFIGURATION("advanced.batch-type-configuration"),
```
absurdfarce, line 297: This enum represents the set of config values for options specific to DSE.  We want this config to be available for Cassandra users as well so it really should be in com.datastax.oss.driver.api.core.config.DefaultDriverOption instead.

core/src/main/java/com/datastax/oss/driver/api/core/cql/BatchStatement.java
```diff
@@ -131,6 +219,15 @@ static BatchStatement newInstance(
...
   }
 
+  /**
+   * Returns a builder to create an instance of the default implementation.
+   *
+   * <p>Note that this builder is mutable and not thread-safe.
+   */
+  static BatchStatementBuilder builder() {
+    return new BatchStatementBuilder(BatchType.UNKNOW);
+  }
```
absurdfarce, line 229: I don't think we want to do this at all.  I mentioned this in comments below but the current builder interface on this class already requires users to provide a non-null BatchType in order to get a builder.  That seems like exactly the kind of thing we want to encourage.  We should only worry about fixing the newInstance() methods above, and I provided some feedback on a way to do that that doesn't require the introduction of an UNKNOWN batch type.

absurdfarce (changes requested):
@Maxwell-Guo This is really good work!  There are some things I think we need to change (see my comments for more detail) but please don't let this discourage you; I really like what you're working on here.  The quantity and nature of your tests alone is a _great_ sign!

There are three areas of change I'd like to see from what you have here:

1. Make sure the relevant config for default batch type is specified in reference.conf, ideally with some comments explaining what it is and how to use it.  There are other examples in some of the other entries in reference.conf
2. Remove the UNKNOWN batch type and make sure that we always have a batch type specified at BatchStatement construction type.  This also enables the removal of the null batch type check in CqlRequestHandler.
3. Remove the batch type registry all together

If you have any questions about any of this please don't hesitate to ask!

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
Maxwell-Guo: +1 on this

integration-tests/src/test/resources/application.conf
Maxwell-Guo: for i'm not familiar with the codebase , so the first patch may be just a test  with a begining. I will modify this in reference.conf with some detail documention.

core/src/main/java/com/datastax/dse/driver/api/core/config/DseDriverOption.java
Maxwell-Guo: at first , I just put this enum variable in DefaultDriverOption for I think it should also be used by c* users, But after I read the jira agagin I found out that The description of the jira is "FedEx (David Twynam) would like the ability to set batch-type application-wide ", I think FedEx may be one of datastax's customers ，then I change the  BATCH_TYPE_CONFIGURATION to DseDriverOption , I will change it back to DefaultDriverOption.

core/src/main/java/com/datastax/oss/driver/api/core/cql/DefaultBatchType.java
Maxwell-Guo: I have also define a BatchType in com.datastax.oss.driver.api.core.cql.BatchType and the type is named UNKNOW and the value of https://github.com/Maxwell-Guo/java-driver/blob/4.x-JAVA-3030/core/src/main/java/com/datastax/oss/driver/api/core/cql/BatchType.java#L31

The reason that an UNKNOW BatchType is defined is  that after I read the jira and the example of how BatchStatement is used for the users in the codebase's tests , I found there is three ways :
1. BatchStatement.builder with  the Batch  Type setted;
2. BatchStatement.newInstance with the Batch Type setted ;
3. Just set the cql string for SimpleStatement ,this way if batch type is not write in cql string the c* will use LOGGED type;

So I add a BatchStatement.builder which no BatchType is needed for the input parameter, and then a flag named UNKNOW batch type is setted for this kind of BatchStatement, then I can use this flag to identify whether the user actively sets the batchtype. what I want to do is that when user do not set the batch type on batch invocation ,the configured batch type will be used ; when user set the batch type on batch invocation ,the user setted type is used.

In fact, at the beginning of this jira, we should discuss the how user use the function that the jira will support. Or is the useage 
(https://github.com/Maxwell-Guo/java-driver/blob/4.x-JAVA-3030/integration-tests/src/test/java/com/datastax/oss/driver/core/cql/BatchStatementUseConfigIT.java#L82  defained) that user need ?

core/src/main/java/com/datastax/oss/driver/internal/core/BatchTypeRegistry.java
Maxwell-Guo: +1

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
Maxwell-Guo: Sorry for Not familiar with the codebase , and I have got this one question here ;do you mean  something like : BatchStatement.newInstance(DriverContext context)? It seems  DriverContext can ony be get from the variable of  session, so what kind of usage of the interface can user  use ? That is what  I am asked at the beigining , is https://github.com/Maxwell-Guo/java-driver/blob/4.x-JAVA-3030/integration-tests/src/test/java/com/datastax/oss/driver/core/cql/BatchStatementUseConfigIT.java#L82  people need ? or something eles.

core/src/main/java/com/datastax/oss/driver/api/core/cql/BatchStatement.java
Maxwell-Guo: Thanks , I aggree that introduce an UNKNOW type is not a good idea.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
absurdfarce: Yes, BatchStatement.newInstance(DriverContext ctx) is exactly what I had in mind.  Keep in mind that the user will have to establish a Session of some kind in order to execute any of these statements.  The ["Quick start" example](https://github.com/datastax/java-driver/tree/4.x/manual/core#quick-start) in the manual demonstrates a really simple example.  For batch statements the user would be creating a BatchStatement object and then adding other statement objects to it but you can see how the general API works here.

The upshot is that being able to extract a DriverContext from the Session shouldn't be a problem when the user goes about creating a BatchStatement instance.

core/src/main/java/com/datastax/oss/driver/api/core/cql/DefaultBatchType.java
absurdfarce: As we discuss below I think you'll find it much easier to just enforce a batch type at BatchStatement creation... you just need some way to allow users to say "I want to use the batch type specified in my config".  That should address this issue (unless I'm missing something else).

PR #1620, opened 2022-11-21
Don't return empty routing key when partition key is unbound

akhaku:
DefaultBoundStatement#getRoutingKey has logic to infer the routing key when no one has explicitly called setRoutingKey or otherwise set the routing key on the statement. It however doesn't check for cases where nothing has been bound yet on the statement.
This causes more problems if the user decides to get a BoundStatementBuilder from the PreparedStatement, set some fields on it, and then copy it by constructing new BoundStatementBuilder objects with the BoundStatement as a parameter, since the empty ByteBuffer gets copied to all bound statements, resulting in all requests being targeted to the same Cassandra node in a token-aware load balancing policy.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultBoundStatement.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultBoundStatement.java
@@ -358,7 +358,8 @@ public class DefaultBoundStatement implements BoundStatement {
       if (indices.isEmpty()) {
         return null;
       } else if (indices.size() == 1) {
-        return getBytesUnsafe(indices.get(0));
+        int index = indices.get(0);
+        return isSet(index) ? getBytesUnsafe(index) : null;
       } else {
         ByteBuffer[] components = new ByteBuffer[indices.size()];
         for (int i = 0; i < components.length; i++) {
--- a/integration-tests/src/test/java/com/datastax/oss/driver/core/cql/PreparedStatementIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/core/cql/PreparedStatementIT.java
@@ -475,6 +475,25 @@ public class PreparedStatementIT {
     assertThat(tokenFactory.hash(boundStatement.getRoutingKey())).isEqualTo(expectedToken);
   }
 
+  @Test
+  public void should_return_null_routing_information_when_single_partition_key_is_unbound() {
+    should_return_null_routing_information_when_single_partition_key_is_unbound(
+        "SELECT a FROM prepared_statement_test WHERE a = ?");
+    should_return_null_routing_information_when_single_partition_key_is_unbound(
+        "INSERT INTO prepared_statement_test (a) VALUES (?)");
+    should_return_null_routing_information_when_single_partition_key_is_unbound(
+        "UPDATE prepared_statement_test SET b = 1 WHERE a = ?");
+    should_return_null_routing_information_when_single_partition_key_is_unbound(
+        "DELETE FROM prepared_statement_test WHERE a = ?");
+  }
+
+  private void should_return_null_routing_information_when_single_partition_key_is_unbound(
+      String queryString) {
+    CqlSession session = sessionRule.session();
+    BoundStatement boundStatement = session.prepare(queryString).bind();
+    assertThat(boundStatement.getRoutingKey()).isNull();
+  }
+
   private static Iterable<Row> firstPageOf(CompletionStage<AsyncResultSet> stage) {
     return CompletableFutures.getUninterruptibly(stage).currentPage();
   }


core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
```diff
@@ -99,4 +127,72 @@ public void should_match_set_routing_key_vararg() {
...
+    BoundStatement copy = new BoundStatementBuilder(boundStatement).build();
+    // we should not have a routing key set
+    assertThat(copy.getRoutingKey()).isNull();
+    copy = copy.set(0, 1, Integer.class).set(1, 2, Integer.class);
+    // we should have a non-empty routing key now
+    assertThat(copy.getRoutingKey()).isNotNull();
+    assertThat(copy.getRoutingKey().hasRemaining()).isTrue();
+  }
+
+  // copied from RequestLogFormatterTest, we should move somewhere to share b/w tests
```
akhaku: hacked the test together by copying this code from `RequestLogFormatterTest`, I'm looking to you folks for advice on where this code should actually live

core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultBoundStatement.java
```diff
@@ -358,7 +358,7 @@ public ByteBuffer getRoutingKey() {
       if (indices.isEmpty()) {
         return null;
       } else if (indices.size() == 1) {
-        return getBytesUnsafe(indices.get(0));
+        return isSet(0) ? getBytesUnsafe(indices.get(0)) : null;
```
akhaku: probably worth noting that the multi-column partition key index code branch has already fixed this bug at line 367 below

akhaku:
More related arguments we could make since we can re-bind the partition key:
1. we should never copy over `getRoutingKey` while copying the `template` at StatementBuilder:71
2. OR we should copy over the raw instance variable, not the computed value
3. OR we should only generate the routing key when evaluating the statement in the load balancing policy, if `Request#getRoutingKey` returns null at BasicLoadBalancingPolicy:290

Between all of those, the simplest solution is probably 1. The second would be nice too but there's no easy way to get the instance variable (save casting to a DefaultBoundStatement and perhaps exposing another getter), and I can't think of a non-brittle way to do the third one.

Worth nothing that as far as I can tell, `DefaultBoundStatement#getRoutingKeyspace` doesn't have this problem because the field that's used to calculate its value (the prepared statement) can't change.

In any case, probably a different PR for that follow-up change?

core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultBoundStatement.java
absurdfarce: I don't think you can use "isSet(0)" here.  The first index could be somewhere else depending on the query... pretty sure this needs to be:

```java
      } else if (indices.size() == 1) {
        int index = indices.get(0);
        return isSet(index) ? getBytesUnsafe(index) : null;
      } else {
```

core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
absurdfarce: This functionality is probably something that's more easily validated in an integration test.  PreparedStatementIT already does some validation of the ability to compute routing information when the partition key is bound; a test for this functionality should slot in next to it pretty nicely.

I was doing some testing with something like the following:

```java
  @Test
  public void should_return_null_routing_information_when_single_partition_key_is_unbound() {
    should_return_null_routing_information_when_single_partition_key_is_unbound(
            "SELECT a FROM prepared_statement_test WHERE a = ?");
    should_return_null_routing_information_when_single_partition_key_is_unbound(
            "INSERT INTO prepared_statement_test (a) VALUES (?)");
    should_return_null_routing_information_when_single_partition_key_is_unbound(
            "UPDATE prepared_statement_test SET b = 1 WHERE a = ?");
    should_return_null_routing_information_when_single_partition_key_is_unbound(
            "DELETE FROM prepared_statement_test WHERE a = ?");
  }

  private void should_return_null_routing_information_when_single_partition_key_is_unbound(String queryString) {

    CqlSession session = sessionRule.session();
    BoundStatement boundStatement = session.prepare(queryString).bind();
    assertThat(boundStatement.getRoutingKey()).isNull();
  }
```

This test fails with the existing 4.x code but passes with the changes referenced above.

absurdfarce:
Thanks for this @akhaku , and my apologies for taking so long to get back to you... holidays and all that!  This change looks pretty good; I had a few comments on your current implementation but I don't think any of them are major.  I haven't yet given a great deal of thought to your points about handling the cases around the builder... I want to think about that a bit more.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultBoundStatement.java
akhaku: ah right, sorry I missed that, thanks

core/src/test/java/com/datastax/oss/driver/api/core/cql/StatementBuilderTest.java
akhaku: thank you, much cleaner than my hacked-together test!

akhaku: And for simplicity I'm going to go with exactly what you have here and not worry about the extra copying tests

akhaku:
No worries, thanks for looking at this! The other comments are much less urgent, I think this is a somewhat serious bug (if you happen to hit it) and so good to get this fixed before worrying about my other comments about cleanup.

akhaku:
Gentle nudge @absurdfarce :)

absurdfarce:
Hey @akhaku, apologies for the delay in getting back to you!

I'm inclined to agree with your analysis above; I think we're probably okay to merge this change as it stands and address the questions copying values in another PR.  With that in mind I think this PR is ready to go as it stands so I'll try to get it merged later today and create a follow-up ticket for the other questions.

Thanks for all your work on this!

absurdfarce:
Hey @akhaku, regarding the follow-up ticket... I think I need to understand the case in question a bit better.

You [mentioned above](https://github.com/datastax/java-driver/pull/1620#issuecomment-1322497210) that we can "re-bind the partition key" for the DefaultPreparedStatement referenced by a given DefaultBoundStatement.  For my own understanding... is the concern here that the List returned by getPartitionKeyIndices() is mutable?  Or are you referring to another operation that I'm just not thinking of at the moment?

I'm also a bit confused by your (correct) point that DefaultBoundStatement.getRoutingKeyspace() uses a field (preparedStatement) in DefaultBoundStatement which cannot change.  What confuses me is that getRoutingKey() references the same field if a fixed routing key wasn't supplied when the instance was constructed.  So the only difference there is a difference in return value (ColumnDefinitions via getVariableDefinitions() in the routing keyspace case, a List of Integers via getPartitionKeyIndices in the routing key case).... which returns me to my earlier question about mutability.

All of _that_ said, the second option on your list ("we should copy over the raw instance variable, not the computed value") seems most desirable (at least given my current understanding).  If the user has explicitly specified a routing key preserve it, otherwise allow it to be computed as it normally would.

akhaku:
Just a rebase.
It's been a while since I looked at this, so based on my fuzzy recollection of what I was talking about:

Regarding rebinding the partition key: you can use `SettableByName` or `SettableById` to reset the partition key after it's already set. If we had the old routing key cached, then after changing the bound parameter, the cached routing key is no longer accurate, we need to recompute the routing key based on the new bound partition key.

That also sort of answers your question about `DefaultBoundStatement#getRoutingKeyspace` - while we can change the underlying information that's used to calculate the routing key (the bound parameter for the partition key), we cannot change the parameter that is used to figure out the routing keyspace.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/DefaultBoundStatement.java
tolbertam: this exactly ^.    On initial look, I came to the same conclusion and felt that you could even just delete this branch and fall into the else and the behavior matches this change exactly, but I think this branch sort of exists as an optimization (avoids creating an array), so :+1: to accept your change as it is.

tolbertam (approved):
:+1: looks good to me;  The behavior appears to make this consistent with > 1 partition key.  If any partition key value is unset, this returns null, so seems correct to also return null in this case.

absurdfarce:
This has been sitting for far too long and the core change here is a good one.  We can and should continue any discussion around the handling of getRoutingKey() in DefaultBoundStatement but I want to make sure the core change of this PR gets into 4.19.0.

absurdfarce (approved):

PR #1621, opened 2022-12-01
Alternate idea to PR 1617

absurdfarce:
Motivated by conversations with @chibenwa on https://github.com/datastax/java-driver/pull/1617.  My hope was that we could implement the fix he discussed while retaining the ability to re-use the backing array from the ByteBuf whenever possible.

### DIFF
--- a/core/src/main/java/com/datastax/dse/driver/internal/core/graph/ByteBufUtil.java
+++ b/core/src/main/java/com/datastax/dse/driver/internal/core/graph/ByteBufUtil.java
@@ -21,14 +21,19 @@ import java.nio.ByteBuffer;
 
 public class ByteBufUtil {
 
+  public static byte[] copyBytes(ByteBuf buffer, int length) {
+
+    final byte[] bytes = new byte[length];
+    buffer.getBytes(buffer.readerIndex(), bytes, 0, length);
+    return bytes;
+  }
+
   // Does not move the reader index of the ByteBuf parameter
   public static ByteBuffer toByteBuffer(ByteBuf buffer) {
     if (buffer.isDirect()) {
       return buffer.nioBuffer();
     }
-    final byte[] bytes = new byte[buffer.readableBytes()];
-    buffer.getBytes(buffer.readerIndex(), bytes);
-    return ByteBuffer.wrap(bytes);
+    return ByteBuffer.wrap(copyBytes(buffer, buffer.readableBytes()));
   }
 
   static ByteBuf toByteBuf(ByteBuffer buffer) {
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufPrimitiveCodec.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/protocol/ByteBufPrimitiveCodec.java
@@ -15,6 +15,7 @@
  */
 package com.datastax.oss.driver.internal.core.protocol;
 
+import com.datastax.dse.driver.internal.core.graph.ByteBufUtil;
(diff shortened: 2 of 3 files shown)


absurdfarce:
@adutra (or even @olim7t if you have the time) I'd love to get your input on this.  @chibenwa pointed out in his [earlier PR](https://github.com/datastax/java-driver/pull/1617) that the existing impl never enters the logic to re-use the backing array when converting from Netty's ByteBuf to the JDK's ByteBuffer.  My goal with this PR was to preserve that logic to the largest extent possible.

Want to make sure I didn't break anything here :)

olim7t:
> @chibenwa pointed out in his [earlier PR](https://github.com/datastax/java-driver/pull/1617) that the existing impl never enters the logic to re-use the backing array

Agreed, this was a bad assumption in the original code.

In hindsight, sharing the underlying array might be a bad idea anyway: if `source` is a pooled `ByteBuf`,  it might eventually get recycled and overwritten.

I would just remove the unused `if` (as proposed in the other PR), and then inline `readRawBytes`:
```java
  @Override
  public ByteBuffer readBytes(ByteBuf source) {
    int length = readInt(source);
    if (length < 0) return null;
    byte[] bytes = new byte[length];
    source.readBytes(bytes);
    return ByteBuffer.wrap(bytes);
  }
```
And maybe add a length check to mirror `readShortBytes`.

chibenwa:
@olim7t that's exactly what I ended up doing in https://github.com/datastax/java-driver/pull/1617 ;-)

> And maybe add a length check to mirror readShortBytes.

That's not even needed `source.readBytes(bytes);` already does this.

chibenwa:
You don't support extra data in this proposal:

```
  @Test
  public void should_read_bytes_when_extra_data() {
    ByteBuf source =
        ByteBufs.wrap(
            // length (as an int)
            0x00,
            0x00,
            0x00,
            0x04,
            // contents
            0xca,
            0xfe,
            0xba,
            0xbe,
            0xde,
            0xda,
            0xdd);
    ByteBuffer bytes = codec.readBytes(source);
    assertThat(Bytes.toHexString(bytes)).isEqualTo("0xcafebabe");
  }
```

fails when the "optimisation" kicks in.

absurdfarce:
Man, I can't make threads on general comments?  That makes me sad.

Good point on the test failure @chibenwa, although the test you have is actually exercising the non-optimized route.  ByteBufs.wrap() will create a pooled byte buffer (or at least it does for me) which has a large initial size, so because that array size is so much larger than the available readable bytes the logic to re-use the underlying buffer never kicks in.  That said, the old logic did fail in that case; the impl I just pushed addresses it in both cases (while also being a lot simpler).

All of that said I'm still very much on the fence about this whole approach.  As a practical matter the optimization only kicks in if we're dealing with unpooled buffers that wrap an existing byte array.  In the pooled case the underlying byte buffer will have some (fairly large) initial size which can expand as new data is written in.  That has two effects of relevance here:

* The concern about pooled buffers being re-used/overwritten doesn't seem to apply here (since these buffers will almost never be subject to this optimization).  We could make this more explicit with type checks if needed but at a minimum this seems to take the concern about data changing underneath you off the table.
* I'm not sure how much gain we get if we're only optimizing unpooled byte buffers.

@olim7t does any of the above change your thinking at all?  I really do like the zero-copy idea where we can leverage it but if we're only optimizing for a subset of cases it seems like that win is probably not worth it.  And if that's the case I agree that @chibenwa 's original PR is the way to go.

absurdfarce:
I don't think this idea is going anywhere so I'm going to just merge the original PR from @chibenwa

PR #1623, opened 2023-01-22
Limit calls to Conversions.resolveExecutionProfile

chibenwa:
Those repeated calls account for a non-negligible portion of my application
CPU (0.6%) and can definitly be a final field so that it gets resolved only
once per CqlRequestHandler.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
@@ -535,29 +535,59 @@ public class Conversions {
     }
   }
 
+  /** Use {@link #resolveIdempotence(Request, DriverExecutionProfile)} instead. */
+  @Deprecated
   public static boolean resolveIdempotence(Request request, InternalDriverContext context) {
+    return resolveIdempotence(request, resolveExecutionProfile(request, context));
+  }
+
+  public static boolean resolveIdempotence(
+      Request request, DriverExecutionProfile executionProfile) {
     Boolean requestIsIdempotent = request.isIdempotent();
-    DriverExecutionProfile executionProfile = resolveExecutionProfile(request, context);
     return (requestIsIdempotent == null)
         ? executionProfile.getBoolean(DefaultDriverOption.REQUEST_DEFAULT_IDEMPOTENCE)
         : requestIsIdempotent;
   }
 
+  /** Use {@link #resolveRequestTimeout(Request, DriverExecutionProfile)} instead. */
+  @Deprecated
   public static Duration resolveRequestTimeout(Request request, InternalDriverContext context) {
-    DriverExecutionProfile executionProfile = resolveExecutionProfile(request, context);
-    return request.getTimeout() != null
-        ? request.getTimeout()
+    return resolveRequestTimeout(request, resolveExecutionProfile(request, context));
+  }
+
+  public static Duration resolveRequestTimeout(
+      Request request, DriverExecutionProfile executionProfile) {
+    Duration timeout = request.getTimeout();
+    return timeout != null
+        ? timeout
         : executionProfile.getDuration(DefaultDriverOption.REQUEST_TIMEOUT);
   }
 
+  /** Use {@link #resolveRetryPolicy(InternalDriverContext, DriverExecutionProfile)} instead. */
+  @Deprecated
   public static RetryPolicy resolveRetryPolicy(Request request, InternalDriverContext context) {
     DriverExecutionProfile executionProfile = resolveExecutionProfile(request, context);
     return context.getRetryPolicy(executionProfile.getName());
   }
 
+  public static RetryPolicy resolveRetryPolicy(
+      InternalDriverContext context, DriverExecutionProfile executionProfile) {
+    return context.getRetryPolicy(executionProfile.getName());
+  }
+
+  /**
+   * Use {@link #resolveSpeculativeExecutionPolicy(InternalDriverContext, DriverExecutionProfile)}
+   * instead.
+   */
+  @Deprecated
   public static SpeculativeExecutionPolicy resolveSpeculativeExecutionPolicy(
       Request request, InternalDriverContext context) {
     DriverExecutionProfile executionProfile = resolveExecutionProfile(request, context);
+    return context.getSpeculativeExecutionPolicy(executionProfile.getName());
+  }
+
+  public static SpeculativeExecutionPolicy resolveSpeculativeExecutionPolicy(
+      InternalDriverContext context, DriverExecutionProfile executionProfile) {
     return context.getSpeculativeExecutionPolicy(executionProfile.getName());
   }
 }
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
@@ -126,6 +126,7 @@ public class CqlRequestHandler implements Throttled {
   private final RequestThrottler throttler;
   private final RequestTracker requestTracker;
   private final SessionMetricUpdater sessionMetricUpdater;
+  private final DriverExecutionProfile executionProfile;
 
   // The errors on the nodes that were already tried (lazily initialized on the first error).
   // We don't use a map because nodes can appear multiple times.
@@ -167,7 +168,8 @@ public class CqlRequestHandler implements Throttled {
     this.sessionMetricUpdater = session.getMetricUpdater();
 
     this.timer = context.getNettyOptions().getTimer();
-    Duration timeout = Conversions.resolveRequestTimeout(statement, context);
+    this.executionProfile = Conversions.resolveExecutionProfile(initialStatement, context);
+    Duration timeout = Conversions.resolveRequestTimeout(statement, executionProfile);
     this.scheduledTimeout = scheduleTimeout(timeout);
 
     this.throttler = context.getRequestThrottler();
@@ -176,8 +178,6 @@ public class CqlRequestHandler implements Throttled {
 
   @Override
   public void onThrottleReady(boolean wasDelayed) {
-    DriverExecutionProfile executionProfile =
-        Conversions.resolveExecutionProfile(initialStatement, context);
     if (wasDelayed
         // avoid call to nanoTime() if metric is disabled:
         && sessionMetricUpdater.isEnabled(
@@ -276,8 +276,6 @@ public class CqlRequestHandler implements Throttled {
               retryCount,
               scheduleNextExecution,
               logPrefix);
-      DriverExecutionProfile executionProfile =
-          Conversions.resolveExecutionProfile(statement, context);
       Message message = Conversions.toMessage(statement, executionProfile, context);
       channel
           .write(message, statement.isTracing(), statement.getCustomPayload(), nodeResponseCallback)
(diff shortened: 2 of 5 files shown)


lukasz-antoniak:
Do you mind closing https://github.com/apache/cassandra-java-driver/pull/1622, as from what I see, this PR contains those changes?

chibenwa:
:+1:

lukasz-antoniak:
I think you closed the wrong PR. This one is mostly fine :).

chibenwa:
I forced push this PR in order to remove the now merged commit.

Sorry for the confusion.

chibenwa:
For the record https://github.com/apache/cassandra-java-driver/pull/1622 being merge the remaining resolveExecutionProfile only accounts for .07% of Apache James CPU footprint.

According to me it would fall under the optimization ratio... 

Shall I close this PR?

lukasz-antoniak:
Any optimisation is fine. The changes you proposed make code cleaner and the initialisation of execution profile easier to follow. Let us just address one comment that I had regarding the reference in callback class.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -543,21 +543,44 @@ public static boolean resolveIdempotence(Request request, InternalDriverContext
         : requestIsIdempotent;
   }
 
+  public static boolean resolveIdempotence(Request request, DriverExecutionProfile executionProfile) {
+    Boolean requestIsIdempotent = request.isIdempotent();
+    return (requestIsIdempotent == null)
+        ? executionProfile.getBoolean(DefaultDriverOption.REQUEST_DEFAULT_IDEMPOTENCE)
+        : requestIsIdempotent;
+  }
+
```
absurdfarce, line 551: This approach provides a nice optimization for callers which have already resolved the DriverExecutionProfile but it does lead to a lot of code duplication.  Seems like if we pursue this path `resolveIdempotence(Request, InternalDriverContext)` should be replaced with:

```java
 public static boolean resolveIdempotence(Request request, InternalDriverContext context) {
    return resolveIdempotence(request, resolveExecutionProfile(request, context));
  }
```

Similar argument for resolveRequestTimeout() as well.

absurdfarce: It's worth noting that the implication of the change referenced above (and indeed this entire PR really) is that a user of this API will only see the benefit of reduced resolution of ExecutionProfiles if they use the new API; anyone using the old method `(resolveIdempotence(Request, InternalDriverContext)` will still see just as many resolution operations as before.

Seems like at a minimum this older method should be marked as deprecated (in favor of the new one) but it doesn't seem like a huge leap from there to say if we're going to go down this path let's just remove the old method and make all callers responsible for managing DriverExecutionProfile as state.  This will require some additional work; based on a quick check it looks like the DSE-specific request handlers (such as `com.datastax.dse.driver.internal.core.cql.continuous.ContinuousRequestHandlerBase` and `com.datastax.dse.driver.internal.core.graph.GraphRequestHandler`) also use resolveIdempotence().  These calls would also need to be refactored if we choose to remove the old method calls.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
```diff
@@ -167,7 +168,8 @@ protected CqlRequestHandler(
     this.sessionMetricUpdater = session.getMetricUpdater();
 
     this.timer = context.getNettyOptions().getTimer();
-    Duration timeout = Conversions.resolveRequestTimeout(statement, context);
+    this.executionProfile = Conversions.resolveExecutionProfile(initialStatement, context);
+    Duration timeout = Conversions.resolveRequestTimeout(statement, executionProfile);
```
absurdfarce, line 172: This change does introduce a change in behaviour that I think is worth calling out and talking about.  I don't think it amounts to a substantial change so I suspect it'll be just fine... but it did seem worth discussing.

The current impl of 4.x resolves execution profiles in real time.  Indeed that's kind of the point of this entire change; we want to do this resolution a lot less.  The effect of resolving these profiles in real time means that any changes we might make to the configuration (say via the driver's support for [manual config reloading](https://github.com/apache/cassandra-java-driver/tree/4.x/manual/core/configuration#manual-reloading)) between sending the message (at the creation of CqlRequestHandler) and receipt of the response (via the callback) will be reflected in the callback handling.  That is no longer the case with this change; since we resolve a single ExecutionProfile at CqlRequestHandler creation and call it directly in the callback any configuration changes won't be reflected in the behaviour of the callback.

To be very clear: I'm not saying this is a problem.  There's a decent argument to be made that an entire request/response cycle should have a constant config in play, something which wasn't present before and is now.  So there's a non-trivial argument that this approach is actually _better_.  Either way it is a _difference_, however, so it seemed worthwhile spending a moment on it.

absurdfarce (changes requested):
Nice work @chibenwa; overall I like where this is headed.  I do think we need to sort out a few things w.r.t. code duplication and consistent use of the new methods throughout the driver code but neither of those should be enormous changes in light of what's already been done here.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
chibenwa: > That is no longer the case with this change

Sure. I'm expecting request to be short lived (5 second ish) and assumed that we only needed the configuration to eventually be propagated.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
chibenwa: Ok I proposed two fixups for implementing those suggested changes.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -535,18 +535,33 @@ public static CoordinatorException toThrowable(
     }
   }
```
tolbertam, line 537: Tiny nit: move `@Deprecated` below the javadoc

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -535,18 +535,33 @@ public static CoordinatorException toThrowable(
...
 
+  @Deprecated
+  /**
+   * Use {@link #resolveRetryPolicy(Request, DriverExecutionProfile)} instead.
```
tolbertam:
```suggestion
   * Use {@link #resolveIdempotence(Request, DriverExecutionProfile)} instead.
```

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -555,9 +570,18 @@ public static RetryPolicy resolveRetryPolicy(Request request, InternalDriverCont
     return context.getRetryPolicy(executionProfile.getName());
   }
 
+  public static RetryPolicy resolveRetryPolicy(Request request, InternalDriverContext context, DriverExecutionProfile executionProfile) {
```
tolbertam: couple nits

Request is an unused parameter here, so probably can be removed.

It looks like the existing method `resolveRetryPolicy` is still being used in places where we already have an `ExecutionPolicy`, would be good to update those and mark this method as `@Deprecated` as well.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
tolbertam: Since resolving an `DriverExecutionProfile` from an `InternalDriverContext` is expensive enough,  I agree with marking the existing methods with `Deprecated` and maybe even removing them in the next minor version (as the project does not ensure compatibility between versions ([reference](https://github.com/apache/cassandra-java-driver/tree/4.x/manual/api_conventions#api-conventions)), but I think it may just be best to keep these methods deprecated.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -555,9 +570,18 @@ public static RetryPolicy resolveRetryPolicy(Request request, InternalDriverCont
...
+  }
+
   public static SpeculativeExecutionPolicy resolveSpeculativeExecutionPolicy(
       Request request, InternalDriverContext context) {
     DriverExecutionProfile executionProfile = resolveExecutionProfile(request, context);
     return context.getSpeculativeExecutionPolicy(executionProfile.getName());
   }
+
+  public static SpeculativeExecutionPolicy resolveSpeculativeExecutionPolicy(
+      Request request, InternalDriverContext context, DriverExecutionProfile executionProfile) {
```
tolbertam: Like `resolveRetryPolicy` (request unused, and existing `resolveSpeculativeExecutionPolicy` is still being used, can be deprecated and callers can call this instead).

core/src/main/java/com/datastax/oss/driver/internal/core/cql/Conversions.java
```diff
@@ -555,9 +570,18 @@ public static RetryPolicy resolveRetryPolicy(Request request, InternalDriverCont
     return context.getRetryPolicy(executionProfile.getName());
   }
 
+  public static RetryPolicy resolveRetryPolicy(Request request, InternalDriverContext context, DriverExecutionProfile executionProfile) {
+    return context.getRetryPolicy(executionProfile.getName());
+  }
+
   public static SpeculativeExecutionPolicy resolveSpeculativeExecutionPolicy(
```
tolbertam, line 583: should mark this as deprecated too.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlRequestHandler.java
tolbertam: > To be very clear: I'm not saying this is a problem. There's a decent argument to be made that an entire request/response cycle should have a constant config in play, something which wasn't present before and is now. So there's a non-trivial argument that this approach is actually better. Either way it is a difference, however, so it seemed worthwhile spending a moment on it.

:+1:, I think it is good to call out, and that this change is more desirable (on the fly changes should only apply to subsequent requests, not in flight ones), but in any case I don't think it good to rely on the behavior working either way.

tolbertam (changes requested):
This looks good to me, thanks @chibenwa!  Just a few tiny nits in the `Conversions` overloaded methods, with those cleaned up I'd be eager to add my :+1:.

chibenwa:
Hi!

Thanks for the helpfull reviews.

I implemented the requested changes in the latest fixup.

Best regards,

Benoit

tolbertam (approved):
Changes look great, thanks @chibenwa !

absurdfarce:
Agree with @tolbertam; I'm pretty pleased with where this landed.  I just kicked off a Jenkins build to make sure there aren't any unexpected test regressions or anything but unless we see some trouble in those results I think we're there.

I'll post the results here when the build finishes.

absurdfarce:
Bah, build failed due to a code formatting issue.  @chibenwa can you run the following:

```
mvn com.coveo:fmt-maven-plugin:format
```

and commit the result to this branch?

absurdfarce:
Awesome, thanks @chibenwa!  New Jenkins build kicked off, will post results for this one upon completion.

absurdfarce:
Jenkins build finished with only a single error which is pretty clearly a one-off failure (only failed on a single Java version and consists of one missing value in a set of a million).  Might need to look at that just a bit more to see if we can repro but this isn't something we've seen often... and there's no reason to believe it's connected to anything in this change.

```
Per-Commit / Matrix - SERVER_VERSION = 'dse-6.8.30', JABBA_VERSION = 'openjdk@1.17' / Execute-Tests / com.datastax.oss.driver.api.core.uuid.UuidsTest.should_generate_unique_random_uuids_Random
...
Error Message
Expected size: 1000000 but was: 999999 in:
[e4182c8a-d3f8-463e-91c8-15320dc8d466,
    0ff0a789-10e8-4060-a43c-6a99bc448a19,
    9c5c8d25-86c9-463d-9605-9f9ceed036ce,
    4e77a939-5366-4290-899a-f5ce18ab9248,
    ac73d701-6ed8-4451-8539-249a0f92ffd9,
```

absurdfarce:
Given the results above I'm adding my approval to @tolbertam's.  @chibenwa can you consolidate these changes down to a single commit (as discussed in the last PR) so that we can get this merged?

Thanks!

absurdfarce (approved):

chibenwa:
Here you are

absurdfarce:
This looks great, thanks for all your work on this @chibenwa!

PR #1624, opened 2023-02-15
Updated esri-geometry-api

tobiasstadler:


tobiasstadler:
@absurdfarce Would you please have a look?

absurdfarce:
Greetings @tobiasstadler, thanks for the pull request!

For version 4.14.0 of the Java driver we changed the ESRI dependency to be optional.  [JAVA-2982](https://datastax-oss.atlassian.net/browse/JAVA-2982) has more information on this change.  We updated our [upgrade guide](https://github.com/datastax/java-driver/tree/4.x/upgrade_guide#esri-geometry-dependency-now-optional) as part of that change as well as the section of our [integration documentation](https://github.com/datastax/java-driver/tree/4.x/manual/core/integration#esri) that covers ESRI.  As these docs explain you can configure your application using the Java driver to use any version of ESRI you wish but we strongly recommend using something in the 1.2.x line, ideally 1.2.1.  We [considered upgrading ESRI](https://github.com/datastax/java-driver/pull/1577) as part of the work on JAVA-2982 but ultimately decided not to; the PR has more detail on this point.

The argument for using ESRI 1.2.1 is that that is the version currently in use by DSE, so using the same version on the client side guarantees that we won't experience any compatibility issues.

[JAVA-2982]: https://datastax-oss.atlassian.net/browse/JAVA-2982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1626, opened 2023-02-21
Fix maven https issue (3.x)

joao-r-reis:


joao-r-reis:
Fixed by https://github.com/datastax/java-driver/pull/1701 instead

PR #1629, opened 2023-03-31
JAVA-3042 Add support for ad-hoc Java 17 builds

absurdfarce:
Need to cover both Jenkins + TravisCI here

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,16 @@
 language: java
-dist: jammy
+dist: trusty
 sudo: false
+# see https://sormuras.github.io/blog/2018-03-20-jdk-matrix.html
 matrix:
   include:
     # 8
     - env: JDK='OpenJDK 8'
       jdk: openjdk8
-      before_script: jdk_switcher use openjdk8
-    # OpenJDK 11 & 17 are both installed by default for jammy images
     # 11
     - env: JDK='OpenJDK 11'
-      before_script: jdk_switcher use openjdk11
-    # 17
-    - env: JDK='OpenJDK 17'
-      before_script: jdk_switcher use openjdk17
+      # switch to JDK 11 before running tests
+      before_script: . $TRAVIS_BUILD_DIR/ci/install-jdk.sh -F 11 -L GPL
 before_install:
   # Require JDK8 for compiling
   - jdk_switcher use openjdk8
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,19 @@
 language: java
-dist: trusty
+dist: jammy
 sudo: false
-# see https://sormuras.github.io/blog/2018-03-20-jdk-matrix.html
 matrix:
   include:
     # 8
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,16 +1,19 @@
...
+      before_script: jdk_switcher use openjdk8
+    # OpenJDK 11 & 17 are both installed by default for jammy images
     # 11
     - env: JDK='OpenJDK 11'
-      # switch to JDK 11 before running tests
-      before_script: . $TRAVIS_BUILD_DIR/ci/install-jdk.sh -F 11 -L GPL
+      before_script: jdk_switcher use openjdk11
+    # 17
+    - env: JDK='OpenJDK 17'
+      before_script: jdk_switcher use openjdk17
```
absurdfarce: These changes are failing; apparently jdk_switcher isn't available on the Jammy image used by Travis.

We've got another problem to sort out; note that we explicitly use Java8 for building the JARs regardless of which JVM we use for testing... but Java8 isn't installed by default on Jammy images.  So whatever solution we find for switching JVMs must support the ability to switch to Java8 for compilation (unless we want to change that too).

absurdfarce:
After thinking about this some more and talking with some other folks about it I'm inclined to leave Travis alone for now.  Getting to a Travis config which supports Java8, 11 and 17 seems non-obvious and we need Travis to support _at least_ Java8 so that we don't get any code in which won't work with the current minimum JDK version.  To meet that goal the safest approach seems to be to leave Travis as-is (thus giving us Java8 testing) and just update Jenkins to support Java17.

absurdfarce:
Note: full support for Java 17 on Jenkins requires an update to the runner AMI + a new build + update to the AMI ID in Jenkins config.  The update to the runner has already been handled in another PR.

joao-r-reis (commented):
Currently, the per-commit build only runs with java 8, should we have a matrix for per commit with multiple java versions?

You only changed the ADHOC build here, if you want to have nightly/weekly builds running with 17 then you have to change the cron part of the file

absurdfarce:
Understood @joao-r-reis .  For now my plan was to change the ad-hoc builds only; thinking was that we probably have some work to do to get that building + passing all tests with Java17.  Once we get to that spot _then_ we can add Java17 to one or more of the scheduled builds (likely as part of a broader re-consideration of which Java versions we use for scheduled builds across the board).

This is all assuming we have some work to do to get compilation + passing tests working.  If that's wrong and everything Just Works :tm: then we'll skip the first part and move to tweaking the scheduled builds.

joao-r-reis:
Hmm do you want to limit the scope of this PR to just adhoc builds then? If so I'd recommend changing the PR title so it's clearer.

absurdfarce:
Yeah, that very much was my intent @joao-r-reis... and I agree the title should make that clearer.  Title of this PR has been updated accordingly.

joao-r-reis (approved):

PR #1638, opened 2023-05-22
JAVA-3058 Clear prepared statement cache on UDT type change event

absurdfarce:


core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
```diff
@@ -15,34 +15,78 @@
...
     implements RequestProcessor<PrepareRequest, CompletionStage<PreparedStatement>> {
 
+  private static final Logger LOG = LoggerFactory.getLogger(CqlPrepareAsyncProcessor.class);
+
   protected final Cache<PrepareRequest, CompletableFuture<PreparedStatement>> cache;
 
-  public CqlPrepareAsyncProcessor() {
-    this(CacheBuilder.newBuilder().weakValues().build());
+  public CqlPrepareAsyncProcessor(DefaultDriverContext context) {
+    this(CacheBuilder.newBuilder().weakValues().build(), context);
```
emerkle826: I understand this is an _internal_ class, but perhaps we should:
1) leave the no-arg constructor and simply add this one that takes a DriverContext
1) call the protected constructor with a `null` for the DriverContext when the no-arg constructor is used
1) Check that the DriverContext is not null in the protected constructor before registering the type change handler
I know that makes it less simple, but it prevents compiler issues in case anyone is doing custom things in their own code with this Processor.

Second comment here. Is it possible we want the DriverContext interface rather than the concrete DefaultDriverContext here? Again, I'm not sure if anyone is digging into the driver to this level, but if `getEventBus()` is defined on the interface, then we can loosen it up and not only allow the DefaultDriverContext.

emerkle826 (approved):
This looks ok to me, I had a couple of comments that don't necessarily require code changes to resolve. If we aren't worried about removing the public no-arg constructor, and DefaultDriverContext is required to register the type change handler, then my comments are moot.

emerkle826:
@absurdfarce Not sure how much of a hassle it would be, but maybe add a test of some sort to verify the cache is cleared on a UDT change.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
```diff
@@ -15,34 +15,78 @@
...
+    EventExecutor adminExecutor = context.getNettyOptions().adminEventExecutorGroup().next();
+    context
+        .getEventBus()
+        .register(TypeChangeEvent.class, RunOrSchedule.on(adminExecutor, this::onTypeChanged));
+  }
+
+  private static boolean typeMatches(UserDefinedType oldType, ColumnDefinition def) {
+    if (def.getType().getProtocolCode() != ProtocolConstants.DataType.UDT) return false;
+    UserDefinedType defType = (UserDefinedType) def.getType();
+    return defType.equals(oldType);
```
olim7t: Don't we need to recurse into complex types as well? Collection of UDTs, UDT nested into another UDT/tuple, etc.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
absurdfarce: Unfortunately getEventBus() isn't part of DriverContext; that's entirely an implementation-dependent thing.  But I do like the idea of maintaining the no-arg constructor to avoid breaking any existing customizations.  I think I'll fold that in... good suggestion!

absurdfarce: Constructor re-work can be found in [this commit](https://github.com/datastax/java-driver/pull/1638/commits/ceb2882dea1ac586fb55f8bb0a6cec84681daa85).  Seems to have worked out reasonably cleanly.

absurdfarce:
Adding the integration test wasn't too bad @emerkle826.  [This commit](https://github.com/datastax/java-driver/pull/1638/commits/fff2af687695228a512b22a59a27c6732566c218) has it, and it includes checks for affected UDTs in the types of either the result set types or the variable definition types.  As expected this fails with some variant of "2 != 1" if the type change handler isn't in place but passes without issue if it is.

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
absurdfarce: Yes... yes, it does.  You're 💯 correct @olim7t .  I'll expand the check in the type change handler and add some additional cases to the integration test.  I'm glad you caught this!

absurdfarce:
@olim7t New code added to address your point about nested/container types.  Assuming that change is good I _believe_ this addresses everything that was outstanding.

I'd like to get this change included in the release going out today so if you guys see anything else let me know... otherwise this change will be going out today.

Thanks again!

core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
```diff
@@ -18,31 +18,112 @@
...
+      try {
+        PreparedStatement stmt = entry.getValue().get();
+        if (Iterables.any(
+                stmt.getResultSetDefinitions(), (def) -> typeMatches(event.oldType, def.getType()))
+            || Iterables.any(
+                stmt.getVariableDefinitions(),
+                (def) -> typeMatches(event.oldType, def.getType()))) {
+
+          this.cache.invalidate(entry.getKey());
+          this.cache.cleanUp();
```
absurdfarce, line 121: Make sure we see the invalidation as quickly as possible.  It's a heavier-weight operation but since changes to this cache should be relatively infrequent it seems like a reasonable compromise.

absurdfarce:
Jenkins failures are caused by some known flakey tests + a few failures around the PreparedStatementIT changes contained in this PR.  Those failures appear to be timing related; we've got asynchronous type change updates + async cache eviction as a result so it's hard to get the timing right.  Multiple local runs of these tests show no failures at all so it very much looks like this is driven by the async behaviour.  I'm content to try and refine this test after release.

absurdfarce:
To be addressed in [JAVA-3062](https://datastax-oss.atlassian.net/browse/JAVA-3062)

[JAVA-3062]: https://datastax-oss.atlassian.net/browse/JAVA-3062?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1639, opened 2023-05-23
JAVA-3060 Add vector type, codec + support for parsing CQL type

absurdfarce:


core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/CqlVectorCodecTest.java
```diff
@@ -0,0 +1,110 @@
...
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import com.datastax.oss.driver.api.core.data.CqlVector;
+import com.datastax.oss.driver.api.core.type.CqlVectorType;
+import com.datastax.oss.driver.api.core.type.DataTypes;
+import com.datastax.oss.driver.api.core.type.codec.TypeCodecs;
+import com.datastax.oss.driver.api.core.type.reflect.GenericType;
+import org.junit.Test;
+
+public class CqlVectorCodecTest extends CodecTestBase<CqlVector> {
```
msmygit: #nit - should this also check to make sure `null` cannot be accepted [per these discussions](https://lists.apache.org/list?dev@cassandra.apache.org:lte=1M:vector)?

absurdfarce: We have checks around the [encode](https://github.com/datastax/java-driver/blob/cep-vsearch/core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/CqlVectorCodecTest.java#L43) and [decode](https://github.com/datastax/java-driver/blob/cep-vsearch/core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/CqlVectorCodecTest.java#L50) functions for null values.  Were you thinking of something else here?

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/CqlVectorCodec.java
```diff
@@ -0,0 +1,105 @@
...
+      throw new IllegalArgumentException("Expected CqlVector to consist of a multiple of 4 bytes");
+    int valuesCnt = length / 4;
+    float[] values = new float[valuesCnt];
+    for (int i = 0; i < valuesCnt; ++i) {
+      values[i] = bytes.getFloat();
+    }
+    /* Restore the input ByteBuffer to its original state */
+    bytes.rewind();
+    return new CqlVector(values);
+  }
```
absurdfarce, line 116: Note that encode/decode above represent something of a compromise.  There's no correlation between the number of floats we try to read or write and the number that are actually expected according to the type.  We do extract the expected number of dimensions for a given vector and we do store it in a CqlVectorType instance but that information isn't available to the encode/decode impls above.  This isn't a huge surprise; in the past parameterized types have been used exclusively to indicate subtypes, not meta-information about a type.

The really robust answer is probably to expand this API so that the underlying type is provided to encode() and decode().  In most cases that type will be ignored, but in this case we'd use it to see how many floats we should expect to read and write.  In the interest of time I avoided this larger-scale refactoring and settled on the compromise above.  I'm open to a follow-up ticket which would do things the right way.

core/src/main/java/com/datastax/oss/driver/api/core/type/DataTypes.java
```diff
@@ -51,13 +53,24 @@ public class DataTypes {
   public static final DataType TINYINT = new PrimitiveType(ProtocolConstants.DataType.TINYINT);
   public static final DataType DURATION = new PrimitiveType(ProtocolConstants.DataType.DURATION);
 
+  /* Vector types are currently expressed as a custom type but the class name in that custom type
+   * if followed by the number of dimensions for this specific type of vector enclosed by parentheses.
+   * A vector with M dimensions is of a different type than a vector with N dimensions (M != N) so
+   * we include the number of dimensions as part of the type.  */
+  private static final Pattern VECTOR_PATTERN =
+      Pattern.compile(CqlVectorType.CQLVECTOR_CLASS_NAME + "(\\d+)");
```
absurdfarce: I don't _love_ the idea of a regex compare here but it's not as bad as it appears at first blush.  We're only doing this for custom types so hopefully it should be a pretty rare event even for users of the vector type.  More importantly, most of the regex in question is a long literal string so eval'ing it as a regex shouldn't be vastly less performant than a base string compare.  Since it gives us the ability to extract the dimensions easily it seemed like a reasonable compromise, although we certainly can implement the extraction logic manually if need be.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/CqlVectorCodec.java
absurdfarce: The above statement is no longer true with the changes made to support CEP-7 syntax.

core/src/main/java/com/datastax/oss/driver/api/core/type/DataTypes.java
absurdfarce: The regex is no longer used; we've addressed this issue via basic string ops.

core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+package com.datastax.oss.driver.api.core.data;
+
+import com.datastax.oss.driver.shaded.guava.common.base.Joiner;
+import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
+import com.datastax.oss.driver.shaded.guava.common.collect.Iterators;
+import java.util.Arrays;
+
+/** An n-dimensional vector defined in CQL */
+public class CqlVector<T> {
```
jshook, line 24: I'd like to make this a simple value type (immutable, final, ...) and javadoc it as such.

core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
```diff
@@ -0,0 +1,91 @@
...
+
+  private final ImmutableList<T> values;
```
jshook, line 26: There is no strong reason I can see here for hoisting Guava in, since immutability can be offered with standard lib conventions in practice. (You're not going to get through a dev cycle treating it differently)

core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
```diff
@@ -0,0 +1,91 @@
...
+
+  private CqlVector(ImmutableList<T> values) {
```
jshook, line 28: Instead of this, why not
```
public CqlVector(List<T> values) { ...
```
and
```
public CqlVector(float... components) { ...
```
[update: I realized the varargs signatures were distinct enough and changed this ^^]
Not sure if can easily combine varargs and type params...

core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
```diff
@@ -0,0 +1,91 @@
...
+    this.values = values;
+  }
+
+  public static Builder builder() {
```
jshook, line 32: Requiring the builder here for such a simple container type causes contorted patterns in the caller's code. Without requiring this, usage patterns can be way easier.

core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
```diff
@@ -0,0 +1,91 @@
...
+    return new Builder();
+  }
+
+  public Iterable<T> getValues() {
```
jshook, line 36: This could be implemented on the value type directly for simplicity, as
```
public class CqlVector<T> implements Iterable<T> {
```

core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
```diff
@@ -0,0 +1,91 @@
...
+    return values;
```
jshook, line 37: Requiring the use of the iterator pattern here for simple collections types is also making the user write more code that is unnecessary. There should be a simple collection type that can be accessed without all the machinations.

jshook (changes requested):
I put some opinionated questions/requests in. Much of these come from the need to be able to create many vectors efficiently. By hiding the APIs which allow for direct injection of pre-computed structures, this will not only be slower, it will be harder to use. Happy to submit a patch on top of this with these changes if you like.

core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
```diff
@@ -0,0 +1,91 @@
...
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.api.core.data;
+
+import com.datastax.oss.driver.shaded.guava.common.base.Joiner;
+import com.datastax.oss.driver.shaded.guava.common.collect.ImmutableList;
+import com.datastax.oss.driver.shaded.guava.common.collect.Iterators;
+import java.util.Arrays;
+
+/** An n-dimensional vector defined in CQL */
```
JeremiahDJordan, line 23: For the server type "n" is part of the type definition.  Should it be on the client as well?  Vector<float, 2> and Vector<float, 3> are different types.
You need to ensure that someone is not trying to bind a Vector<float, 3> into a Vector<float, 2>

absurdfarce: This is only the value type.  CqlVectorType actually defines the specific type of this vector, and there we include both subtype and dimension.  Also note that this is the type returned by CqlVectorCodec.getCqlType(), which in turn means that it's used as the comparison to determine whether or not a given DataType can be handled by the codec.  As a consequence the codec will reject any type which doesn't match it specifically.  You can see this tested in CqlVectorCodecTest.should_accept_vector_type_correct_dimension_only().

absurdfarce:
Note: most of the comments here will be addressed in [JAVA-3061](https://datastax-oss.atlassian.net/browse/JAVA-3061)

[JAVA-3061]: https://datastax-oss.atlassian.net/browse/JAVA-3061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1642, opened 2023-06-07
JAVA-3042: Support testing against Java17

hhughes:
Various changes to support running tests with jdk17 Todo: add explanation of changes and dep updgrades

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,6 +18,24 @@ def initializeEnvironment() {
 
   env.MAVEN_HOME = "${env.HOME}/.mvn/apache-maven-3.3.9"
   env.PATH = "${env.MAVEN_HOME}/bin:${env.PATH}"
+
+  /*
+  * As of JAVA-3042 JAVA_HOME is always set to JDK8 and this is currently necessary for mvn compile and DSE Search/Graph.
+  * To facilitate testing with JDK11/17 we feed the appropriate JAVA_HOME into the maven build via commandline.
+  *
+  * Maven command-line flags:
+  * - -DtestJavaHome=/path/to/java/home: overrides JAVA_HOME for surefire/failsafe tests, defaults to environment JAVA_HOME.
+  * - -Ptest-jdk-N: enables profile for running tests with a specific JDK version (substitute N for 8/11/17).
+  *
+  * Note test-jdk-N is also automatically loaded based off JAVA_HOME SDK version so testing with an older SDK is not supported.
+  *
+  * Environment variables:
+  * - JAVA_HOME: Path to JDK used for mvn (all steps except surefire/failsafe), Cassandra, DSE.
+  * - JAVA8_HOME: Path to JDK8 used for Cassandra/DSE if ccm determines JAVA_HOME is not compatible with the chosen backend.
+  * - TEST_JAVA_HOME: PATH to JDK used for surefire/failsafe testing.
+  * - TEST_JAVA_VERSION: TEST_JAVA_HOME SDK version number [8/11/17], used to configure test-jdk-N profile in maven (see above)
+  */
+
   env.JAVA_HOME = sh(label: 'Get JAVA_HOME',script: '''#!/bin/bash -le
     . ${JABBA_SHELL}
     jabba which ${JABBA_VERSION}''', returnStdout: true).trim()
@@ -25,9 +43,15 @@ def initializeEnvironment() {
     . ${JABBA_SHELL}
     jabba which 1.8''', returnStdout: true).trim()
 
+  env.TEST_JAVA_HOME = sh(label: 'Get TEST_JAVA_HOME',script: '''#!/bin/bash -le
+    . ${JABBA_SHELL}
+    jabba which ${JABBA_VERSION}''', returnStdout: true).trim()
+  env.TEST_JAVA_VERSION = sh(label: 'Get TEST_JAVA_VERSION',script: '''#!/bin/bash -le
+    echo "${JABBA_VERSION##*.}"''', returnStdout: true).trim()
+
   sh label: 'Download Apache CassandraⓇ or DataStax Enterprise',script: '''#!/bin/bash -le
     . ${JABBA_SHELL}
-    jabba use ${JABBA_VERSION}
+    jabba use 1.8
     . ${CCM_ENVIRONMENT_SHELL} ${SERVER_VERSION}
   '''
 
@@ -53,7 +77,7 @@ ENVIRONMENT_EOF
     set +o allexport
 
     . ${JABBA_SHELL}
-    jabba use ${JABBA_VERSION}
+    jabba use 1.8
 
     java -version
     mvn -v
@@ -80,7 +104,7 @@ def executeTests() {
     set +o allexport
 
     . ${JABBA_SHELL}
-    jabba use ${JABBA_VERSION}
+    jabba use 1.8
 
     if [ "${JABBA_VERSION}" != "1.8" ]; then
       SKIP_JAVADOCS=true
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -61,7 +61,7 @@
     <jackson-databind.version>2.13.2.2</jackson-databind.version>
     <legacy-jackson.version>1.9.12</legacy-jackson.version>
     <!-- optional dependencies -->
-    <snappy.version>1.1.7.3</snappy.version>
+    <snappy.version>1.1.8.4</snappy.version>
```
hhughes: note to myself to remove this dependency update since it's unrelated to jdk17

pom.xml
```diff
@@ -714,6 +715,7 @@ limitations under the License.]]></inlineHeader>
             </property>
           </properties>
           <skip>${skipUnitTests}</skip>
+          <argLine>--add-opens java.base/jdk.internal.util.random=ALL-UNNAMED</argLine>
```
hhughes: this line will cause jdk8 build to fail :(

Jenkinsfile
```diff
@@ -25,9 +25,16 @@ def initializeEnvironment() {
...
+    . ${JABBA_SHELL}
+    jabba which ${JABBA_VERSION}''', returnStdout: true).trim()
+  env.TEST_JAVA_VERSION = sh(label: 'Get TEST_JAVA_HOME',script: '''#!/bin/bash -le
+    echo "${JABBA_VERSION##*.}"''', returnStdout: true).trim()
+
+
   sh label: 'Download Apache CassandraⓇ or DataStax Enterprise',script: '''#!/bin/bash -le
     . ${JABBA_SHELL}
-    jabba use ${JABBA_VERSION}
+    jabba use 1.8
```
weideng1, line 54: Shall we have a comment explaining the difference between JABBA_VERSION, TEST_JAVA_VERSION, etal?

integration-tests/src/test/java/com/datastax/dse/driver/api/core/cql/continuous/ContinuousPagingITBase.java
```diff
@@ -61,7 +62,9 @@ protected static void initialize(CqlSession session, DriverExecutionProfile slow
     }
     int count = 0;
     for (int i = 0; i < 200; i++) {
-      BatchStatement batch = BatchStatement.newInstance(DefaultBatchType.UNLOGGED);
+      BatchStatement batch =
+          BatchStatement.newInstance(DefaultBatchType.UNLOGGED)
+              .setConsistencyLevel(ConsistencyLevel.ALL);
```
weideng1: Which test failure this CL_ALL is trying to address?

hhughes: This was an experiement to attempt to fix flakiness where the queries in these tests fail to find all the items added in the test set-up. It wasn't supposed to be included in this PR so should be removed, thanks for catching this.

Jenkinsfile
hhughes: Done. I've also filed JAVA-3064 to tidy up how jabba is used in Jenkinsfile since after this change we no longer need the all the `JAVA_HOME` switching stuff in `buildDriver` and `executeTests`

weideng1 (approved):
nicely done! Not only this added automated java17 test (and fixed all compatibility issues), it also expanded our per-commit test combinations so we can quickly catch test failures in higher-than-java8 environments without resorting to manually triggered tests.

Jenkinsfile
```diff
@@ -18,16 +18,40 @@ def initializeEnvironment() {
...
+  env.TEST_JAVA_HOME = sh(label: 'Get TEST_JAVA_HOME',script: '''#!/bin/bash -le
+    . ${JABBA_SHELL}
+    jabba which ${JABBA_VERSION}''', returnStdout: true).trim()
+  env.TEST_JAVA_VERSION = sh(label: 'Get TEST_JAVA_HOME',script: '''#!/bin/bash -le
+    echo "${JABBA_VERSION##*.}"''', returnStdout: true).trim()
+
   sh label: 'Download Apache CassandraⓇ or DataStax Enterprise',script: '''#!/bin/bash -le
     . ${JABBA_SHELL}
-    jabba use ${JABBA_VERSION}
+    jabba use 1.8
```
absurdfarce, line 54: Worth noting that this is the setup for the attempt to install the requested C* server version via ccm_environment.  This operation assumes that a Jabba install with the requested version is available (and uses that as JAVA_HOME for the C* invocation).  See [here](https://github.com/riptano/openstack-jenkins-drivers/blob/master/shared/scripts/ccm_environment.sh#L273) and [here](https://github.com/riptano/openstack-jenkins-drivers/blob/master/shared/scripts/ccm_environment.sh#L49-L54) and [here](https://github.com/riptano/openstack-jenkins-drivers/blob/master/shared/scripts/ccm_environment.sh#L278-L281) for more on that point.

Bottom line: it doesn't matter what JDK is setup here... we'll wind up using Java8 for the ccm invocation regardless.

Jenkinsfile
```diff
@@ -18,16 +18,40 @@ def initializeEnvironment() {
...
     . ${JABBA_SHELL}
     jabba which ${JABBA_VERSION}''', returnStdout: true).trim()
   env.JAVA8_HOME = sh(label: 'Get JAVA8_HOME',script: '''#!/bin/bash -le
     . ${JABBA_SHELL}
     jabba which 1.8''', returnStdout: true).trim()
 
+  env.TEST_JAVA_HOME = sh(label: 'Get TEST_JAVA_HOME',script: '''#!/bin/bash -le
+    . ${JABBA_SHELL}
+    jabba which ${JABBA_VERSION}''', returnStdout: true).trim()
+  env.TEST_JAVA_VERSION = sh(label: 'Get TEST_JAVA_HOME',script: '''#!/bin/bash -le
```
absurdfarce: Presumably this should be 'Get TEST_JAVA_VERSION'... ?

pom.xml
```diff
@@ -936,6 +936,69 @@ height="0" width="0" style="display:none;visibility:hidden"></iframe></noscript>
...
+      <activation>
+        <property>
+          <name>!testJavaHome</name>
+        </property>
+      </activation>
+      <properties>
+        <testing.jvm>${env.JAVA_HOME}</testing.jvm>
+      </properties>
+    </profile>
+    <!-- set -DtestJdkHome=/path/to/jdk/home to use a different JDK for surefire/failsafe -->
```
absurdfarce: Believe this should be "testJavaHome" rather than "testJdkHome" here, right?

pom.xml
```diff
@@ -533,12 +533,12 @@
...
           <groupId>org.jacoco</groupId>
           <artifactId>jacoco-maven-plugin</artifactId>
-          <version>0.8.5</version>
+          <version>0.8.10</version>
         </plugin>
         <plugin>
           <groupId>org.apache.felix</groupId>
           <artifactId>maven-bundle-plugin</artifactId>
-          <version>4.2.1</version>
+          <version>5.1.1</version>
```
absurdfarce, line 541: Does this change have any implications on the minimum supported Maven version?

absurdfarce (changes requested):
A few minor changes and a question on Maven versions but otherwise this looks very much like what we've been discussing to date.

Jenkinsfile
hhughes: Makes sense. Here I'm really making sure that jabba is not accidentally loading some other version of JDK since we now always want JAVA_HOME to be 1.8 during the jenkins run. I'd like to leave them all in for now to reduce the complexity of this change and I've filed [JAVA-3064](https://datastax-oss.atlassian.net/browse/JAVA-3064) to clean up all these extra usages of jabba in this script.

pom.xml
hhughes: This is the changelog https://github.com/apache/felix-dev/blob/master/tools/maven-bundle-plugin/changelog.txt#L59

There's nothing calling out bumping maven version or other dependencies

hhughes:
Closing and reopening from a branch

PR #1644, opened 2023-06-12
JAVA-3073 Memoize ByteOrderedToken hashCode to avoid expensive computation

schlosna:
While profiling a service using java-driver 3.x, in JFR output I noticed significant CPU cycles and memory allocations corresponding to computing `TokenRange::hashCode()`. Note that this is likely due to use of byte ordered partitioner for this Cassandra instance as the ByteOrderToken must traverse the ByteBuffer each time, and `TokenRangeBase` (in 4.x, `TokenRange` in 3.x) is using `Objects.hash(start, end)` that allocates an array for var args for each invocation. 

image

image

image

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/TokenRangeBase.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/TokenRangeBase.java
@@ -26,7 +26,6 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import java.math.BigInteger;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Objects;
 import net.jcip.annotations.Immutable;
 
 @Immutable
@@ -277,7 +276,7 @@ public abstract class TokenRangeBase implements TokenRange {
 
   @Override
   public int hashCode() {
-    return Objects.hash(start, end);
+    return (31 + start.hashCode()) * 31 + end.hashCode();
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/TokenRangeBase.java
```diff
@@ -275,7 +274,7 @@ public boolean equals(Object other) {
 
   @Override
   public int hashCode() {
-    return Objects.hash(start, end);
```
schlosna: `Objects.hash(Object...)` allocates an array for each invocation

mswintermeyer (approved):

schlosna:
Filed https://datastax-oss.atlassian.net/browse/JAVA-3073

absurdfarce:
Thanks for the PR @schlosna , and for filing the JIRA ticket!  Apologies for not digging into this yet; I've been completely swamped with other work but both the analysis and the PR look very thorough so I'm hoping to dig into it soon!

schlosna:
Thanks for taking a look @absurdfarce ! I will review and get back on the CLA.

One area where this issue pops up is when performing bulk DDL schema operations as the client is trying to stay in sync with C*. A workaround may be using the debouncing and schema disable/enable to avoid threshing the client's view of C* schemas.

* https://docs.datastax.com/en/developer/java-driver/4.14/manual/core/performance/#debouncing
* https://docs.datastax.com/en/developer/java-driver/4.14/manual/core/metadata/schema/#enabling-disabling

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/token/TokenRangeBase.java
```diff
@@ -275,7 +274,7 @@ public boolean equals(Object other) {
...
+    return 31 * start.hashCode() + end.hashCode();
```
SiyaoIsHiding: According to [List.hashCode()](https://docs.oracle.com/javase/8/docs/api/java/util/List.html#hashCode--)

```
     int hashCode = 1;
     for (E e : list)
         hashCode = 31*hashCode + (e==null ? 0 : e.hashCode());
```
It seems to me that it should be ` (31 + start.hashCode()) * 31 + end.hashCode()`.

PR #1645, opened 2023-06-12
JAVA-3062: Figure out a better solution for PreparedStatementIT tests…

hhughes:
… around JAVA-3058

- Make PreparedStatementIT resistant to JVM GC clearing items from prepared statement cache mid-test
- Add tests around prepared statement cache garbage collection

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/core/cql/PreparedStatementCachingIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/core/cql/PreparedStatementCachingIT.java
@@ -27,10 +27,10 @@ import com.datastax.oss.driver.api.core.cql.PreparedStatement;
 import com.datastax.oss.driver.api.core.metrics.DefaultSessionMetric;
 import com.datastax.oss.driver.api.core.session.ProgrammaticArguments;
 import com.datastax.oss.driver.api.core.session.SessionBuilder;
-import com.datastax.oss.driver.api.testinfra.ccm.CcmRule;
+import com.datastax.oss.driver.api.testinfra.ccm.CustomCcmRule;
 import com.datastax.oss.driver.api.testinfra.session.SessionRule;
 import com.datastax.oss.driver.api.testinfra.session.SessionUtils;
-import com.datastax.oss.driver.categories.ParallelizableTests;
+import com.datastax.oss.driver.categories.IsolatedTests;
 import com.datastax.oss.driver.internal.core.context.DefaultDriverContext;
 import com.datastax.oss.driver.internal.core.cql.CqlPrepareAsyncProcessor;
 import com.datastax.oss.driver.internal.core.cql.CqlPrepareSyncProcessor;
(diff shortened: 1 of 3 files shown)


integration-tests/src/test/java/com/datastax/oss/driver/core/cql/PreparedStatementIT.java
```diff
@@ -392,10 +391,15 @@ public void should_create_separate_instances_for_differently_formatted_queries()
...
+          session.prepareAsync("SELECT * FROM prepared_statement_test WHERE a = ?");
+      CompletionStage<PreparedStatement> preparedStatement2Future =
+          session.prepareAsync("select * from prepared_statement_test where a = ?");
+
       PreparedStatement preparedStatement1 =
-          session.prepare("SELECT * FROM prepared_statement_test WHERE a = ?");
+          CompletableFutures.getUninterruptibly(preparedStatement1Future);
       PreparedStatement preparedStatement2 =
-          session.prepare("select * from prepared_statement_test where a = ?");
+          CompletableFutures.getUninterruptibly(preparedStatement2Future);
```
absurdfarce, line 402: So I'm a little confused about this.

The original test(s) include an initial check to make sure the cache is empty at first, followed by checks to make sure reprepare either returns the same instance (if it's the same text) or something different (if the text differs slightly).  We then validate the cache size to see if it matches expectations.

I'm a bit confused as to what the point of the cache size checks are.  The whole point of this test is to confirm whether we get back the same instance for the same text (or a different instance for slightly different text).  Why is the cache size relevant there?  Feels like that's an implementation detail.  If the driver returns the same instance for the same query this test should pass; whether it uses a cache or not is irrelevant.

I understand you didn't write this test and that you're only modifying what was there to avoid the discussed GC issue.  But as I mentioned when we first talked about this I don't _love_ the preservation of the CompletionStages above to short-circuit the weak ref cache vals.  That led me to wonder whether we really need them here... which in turn led me to wonder whether the cache size checks really make sense here.

integration-tests/src/test/java/com/datastax/oss/driver/core/cql/PreparedStatementCachingIT.java
```diff
@@ -0,0 +1,401 @@
...
+
+    @Override
+    protected RequestProcessorRegistry buildRequestProcessorRegistry() {
+      List<RequestProcessor<?, ?>> processors =
+          BuiltInRequestProcessors.createDefaultProcessors(this);
+      processors.removeIf((processor) -> processor instanceof CqlPrepareAsyncProcessor);
+      processors.removeIf((processor) -> processor instanceof CqlPrepareSyncProcessor);
+      CqlPrepareAsyncProcessor asyncProcessor = new TestCqlPrepareAsyncProcessor(Optional.of(this));
+      processors.add(2, asyncProcessor);
+      processors.add(3, new CqlPrepareSyncProcessor(asyncProcessor));
```
absurdfarce, line 157: A question: do we need some comments here explaining why we're doing these manipulations of the RequestProcessorRegistry?  Perhaps it's obvious but it seemed at least worth asking about...

absurdfarce (approved):
I'm basically good with what's here.  A few questions but I don't know that either of them are essential.

integration-tests/src/test/java/com/datastax/oss/driver/core/cql/PreparedStatementIT.java
absurdfarce: If we're going to keep the CompletionStages around we should at least add a comment saying this represents an explicit design choice and explain why we're doing it.  Said ticket should probably reference the JIRA ticket as well.

hhughes:
Closing and re-opening from a branch because we snyk won't pass from a fork

PR #1650, opened 2023-06-16
JAVA-3065: PreparedStatementIT#should_fail_fast_if_id_changes_on_repr…

hhughes:
…epare fails with recent C*/DSE versions

PreparedStatementIT.java
- adds logic to should_fail_fast_if_id_changes_on_reprepare to check for an exception/no-exception depending on whether CASSANDRA-15252 should be fixed

CcmBridge.java
- set ccm.version 4.0.0->4.0.10 to use the same 4.0 cassandra version as Jenkins

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/core/cql/PreparedStatementIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/core/cql/PreparedStatementIT.java
@@ -16,7 +16,7 @@
 package com.datastax.oss.driver.core.cql;
 
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.assertj.core.api.Assertions.assertThatCode;
 import static org.assertj.core.api.Assertions.catchThrowable;
 
 import com.codahale.metrics.Gauge;
@@ -36,6 +36,8 @@ import com.datastax.oss.driver.api.core.servererrors.InvalidQueryException;
 import com.datastax.oss.driver.api.core.type.DataTypes;
 import com.datastax.oss.driver.api.testinfra.CassandraRequirement;
 import com.datastax.oss.driver.api.testinfra.ccm.CcmRule;
+import com.datastax.oss.driver.api.testinfra.requirement.BackendRequirement;
+import com.datastax.oss.driver.api.testinfra.requirement.BackendType;
 import com.datastax.oss.driver.api.testinfra.session.SessionRule;
 import com.datastax.oss.driver.api.testinfra.session.SessionUtils;
 import com.datastax.oss.driver.categories.ParallelizableTests;
@@ -54,6 +56,7 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.function.Consumer;
 import junit.framework.TestCase;
+import org.assertj.core.api.AbstractThrowableAssert;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
@@ -425,13 +428,11 @@ public class PreparedStatementIT {
   }
 
   /**
-   * This test relies on CASSANDRA-15252 to reproduce the error condition. If the bug gets fixed in
-   * Cassandra, we'll need to add a version restriction.
+   * This method reproduces CASSANDRA-15252 which is fixed in 3.0.26/3.11.12/4.0.2.
    *
    * @see <a href="https://issues.apache.org/jira/browse/CASSANDRA-15252">CASSANDRA-15252</a>
    */
-  @Test
-  public void should_fail_fast_if_id_changes_on_reprepare() {
+  private AbstractThrowableAssert<?, ? extends Throwable> assertableReprepareAfterIdChange() {
     try (CqlSession session = SessionUtils.newSession(ccmRule)) {
       PreparedStatement preparedStatement =
           session.prepare(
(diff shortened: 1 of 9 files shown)


test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
```diff
@@ -52,7 +52,7 @@ public class CcmBridge implements AutoCloseable {
   private static final Logger LOG = LoggerFactory.getLogger(CcmBridge.class);
 
   public static final Version VERSION =
-      Objects.requireNonNull(Version.parse(System.getProperty("ccm.version", "4.0.0")));
+      Objects.requireNonNull(Version.parse(System.getProperty("ccm.version", "4.0.10")));
```
hhughes: Bringing default Cassandra version used in testing to be same 4.0 one used by Jenkins.

This is why this test passed locally but not in 4.0/3.11 in Jenkins.

Happy to remove if this change is too disruptive.

absurdfarce: Since we're talking about a default here I'd prefer to avoid changing it for now.  Worth noting that on Jenkins at least this value is always provided [by the runner](https://github.com/riptano/openstack-jenkins-drivers/blob/master/shared/scripts/ccm_environment.sh#L260) by way of the [Jenkinsfile](https://github.com/datastax/java-driver/blob/4.x/Jenkinsfile#LL101C1-L101C1).  So for our purposes it will always be set to the expected version based on Jenkins args.

integration-tests/src/test/java/com/datastax/oss/driver/core/cql/PreparedStatementIT.java
```diff
@@ -444,9 +447,32 @@ public void should_fail_fast_if_id_changes_on_reprepare() {
...
+
+      ThrowableAssert.ThrowingCallable errorTrigger =
+          () -> session.execute(preparedStatement.bind(1));
+      if (isCassandra15252Fixed) {
+        assertThatCode(errorTrigger).doesNotThrowAnyException();
+      } else {
+        assertThatThrownBy(errorTrigger)
+            .isInstanceOf(IllegalStateException.class)
+            .hasMessageContaining("ID mismatch while trying to reprepare");
+      }
```
absurdfarce: Within the integration test framework this kind of thing is normally handled via the [CassandraRequirement annotation](https://github.com/datastax/java-driver/blob/4.x/test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/CassandraRequirement.java).  You can find examples in [other test methods of PreparedStatementIT](https://github.com/datastax/java-driver/blob/4.x/integration-tests/src/test/java/com/datastax/oss/driver/core/cql/PreparedStatementIT.java#L154) (as well as other integration tests).  Enforcement of these annotations can be found in [BaseCcmRule](https://github.com/datastax/java-driver/blob/4.x/test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/BaseCcmRule.java#L76-L132).

I'd very strongly prefer not to introduce multiple methods for supporting this kind of behaviour unless it's absolutely necessary, but it's worth pointing out that the current impl does have a limitation of relevance for us here.  To really do this right you'd want to be able to (a) attach multiple annotations to a given test method, one for each major.minor version, with a min or max version specified as appropriate and (b) have BaseCcmRule evaluate _all_ annotation on the test method and enforce all of them.  After a quick review of the code I don't think that change is actually that bad, so maybe this is an opportunity to add a new feature (support for multiple annotations) in order to avoid having to introduce another way to handle version constraints.

absurdfarce (changes requested):
Looks good, but I'm a little concerned about the verbosity of the version checks in code and the fact that it's handled directly in the test method rather than in some abstracted way.  I mentioned how we do this kind of thing in tests now.  It might be a good idea to see if we can add the functionality we need here to the existing implementation; that gives us a general feature which can be re-used in multiple contexts.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
hhughes: Ack, removed

integration-tests/src/test/java/com/datastax/oss/driver/core/cql/PreparedStatementIT.java
hhughes: Added new annotation: `BackendRequirement` which supports annotation repetition and am now using this with `should_fail_fast_if_id_changes_on_reprepare`. I've left the previous anotations in for now because I think there is some more tidying up, e.g. for tests which have been copied to support multiple c*/dse requirement, I've filed JAVA-3069 to track this work.

hhughes: Wrong version/backend error in intellij test output looks like:
```
org.junit.AssumptionViolatedException: Test requires one of:
  - C* 3.0.26 or greater, but less than 3.11.0
  - C* 3.11.12 or greater, but less than 4.0.0
  - C* 4.0.2 or greater
but configuration is C* 4.0.0.
```

integration-tests/src/test/java/com/datastax/oss/driver/core/cql/PreparedStatementIT.java
```diff
@@ -444,12 +445,38 @@ public void should_fail_fast_if_id_changes_on_reprepare() {
       executeDdl("DROP TABLE prepared_statement_test");
       executeDdl("CREATE TABLE prepared_statement_test (a int PRIMARY KEY, b int, c int)");
 
-      assertThatThrownBy(() -> session.execute(preparedStatement.bind(1)))
-          .isInstanceOf(IllegalStateException.class)
-          .hasMessageContaining("ID mismatch while trying to reprepare");
+      return assertThatCode(() -> session.execute(preparedStatement.bind(1)));
```
absurdfarce, line 448: Slick, I like this... 👍

integration-tests/src/test/java/com/datastax/oss/driver/core/cql/PreparedStatementIT.java
```diff
@@ -444,12 +445,38 @@ public void should_fail_fast_if_id_changes_on_reprepare() {
...
     }
   }
 
+  @BackendRequirement(type = BackendType.DSE)
```
absurdfarce: DSE is actually a different code base; it has to explicitly bring changes from the OSS C* project along.  I'm wondering if DSE has even brought the relevant change over to its code base (let alone released it yet).

It doesn't _look like_ that's happened yet (although I haven't made an exhaustive search).  Perhaps add a comment here saying that we'll need to add version boundaries to this guy as well once DSE brings the change over.

absurdfarce (approved):
Very nice work on this; the improved annotation is _significantly_ more expressive without compromising readability.  A _very_ minor point re: DSE vs. Cassandra version but it's not a huge thing at all.

I'm very much looking forward to getting this one in!

integration-tests/src/test/java/com/datastax/oss/driver/core/cql/PreparedStatementIT.java
hhughes: Yes it's currently `4.0.0`: https://github.com/riptano/bdp/blob/6.8-dev/build.properties#L31, updated the description and added a comment

hhughes:
Closing and reopening from a branch

PR #1651, opened 2023-06-17
JAVA-3068: Use fully qualified table CQL in should_not_allow_unset_va…

hhughes:
…lue_when_protocol_less_than_v4 to work around intermittent server error

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/core/cql/BoundStatementCcmIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/core/cql/BoundStatementCcmIT.java
@@ -16,6 +16,7 @@
 package com.datastax.oss.driver.core.cql;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.assertj.core.api.Assumptions.assumeThat;
 
 import com.datastax.oss.driver.api.core.ConsistencyLevel;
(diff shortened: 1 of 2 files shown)


integration-tests/src/test/java/com/datastax/oss/driver/core/cql/BoundStatementCcmIT.java
```diff
@@ -126,19 +127,23 @@ public void setupSchema() {
...
+          v3Session.prepare(
+              String.format("INSERT INTO %s.test2 (k, v0) values (?, ?)", sessionRule.keyspace()));
 
       BoundStatement boundStatement =
           prepared.boundStatementBuilder().setString(0, name.getMethodName()).unset(1).build();
 
-      v3Session.execute(boundStatement);
+      assertThatThrownBy(() -> v3Session.execute(boundStatement))
+          .isInstanceOf(IllegalStateException.class)
+          .hasMessageContaining("Unset value at index");
```
absurdfarce, line 148: 🎉

integration-tests/src/test/java/com/datastax/oss/driver/core/cql/BoundStatementCcmIT.java
```diff
@@ -126,19 +127,23 @@ public void setupSchema() {
...
     DriverConfigLoader loader =
         SessionUtils.configLoaderBuilder()
             .withString(DefaultDriverOption.PROTOCOL_VERSION, "V3")
             .build();
-    try (CqlSession v3Session = SessionUtils.newSession(ccmRule, sessionRule.keyspace(), loader)) {
-      PreparedStatement prepared = v3Session.prepare("INSERT INTO test2 (k, v0) values (?, ?)");
+    try (CqlSession v3Session = SessionUtils.newSession(ccmRule, loader)) {
+      PreparedStatement prepared =
+          v3Session.prepare(
+              String.format("INSERT INTO %s.test2 (k, v0) values (?, ?)", sessionRule.keyspace()));
```
absurdfarce, line 141: Looked at in isolation it's a bit unclear why this change fixes the problem.  Perhaps add a comment with additional details here, most importantly the error message that you pulled out of Cassandra via the debugger (and optionally where it came from as well).  Might be useful if we find ourselves wanting to add a version bound to this test later on.

absurdfarce (approved):
Really nice work on this one too!  Here as well we might want to add some additional detail about what was going on here (and how you identified the underlying issue) since the code change doesn't really lend itself to a lot of description... but it's certainly not required.

integration-tests/src/test/java/com/datastax/oss/driver/core/cql/BoundStatementCcmIT.java
absurdfarce: I suppose it is basically spelled out in the JIRA ticket so if you just want to leave it there I'm good with that.  Maybe a pointer here saying something like "see JAVA-3068 for relevant details about the structure of this test" might not be the worst idea... but given that the ticket spells it out it's certainly not necessary to do a whole lot more here.

hhughes:
Closing and reopening from a branch so that snyk can run

PR #1655, opened 2023-06-22
JAVA-3050: Upgrade Netty 4.1.94 to address recent CVEs

hhughes:
osgi-tests/BundleOptions.java
- since netty-4.1.78, netty-handler additionally depends on netty-transport-native-unix-common so we need to pull that in when configuring pax exam

### DIFF
--- a/osgi-tests/src/test/java/com/datastax/oss/driver/internal/osgi/support/BundleOptions.java
+++ b/osgi-tests/src/test/java/com/datastax/oss/driver/internal/osgi/support/BundleOptions.java
@@ -91,6 +91,7 @@ public class BundleOptions {
             mavenBundle("io.netty", "netty-codec").versionAsInProject(),
             mavenBundle("io.netty", "netty-common").versionAsInProject(),
             mavenBundle("io.netty", "netty-transport").versionAsInProject(),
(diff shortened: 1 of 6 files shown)


osgi-tests/src/test/java/com/datastax/oss/driver/internal/osgi/support/BundleOptions.java
```diff
@@ -91,6 +91,7 @@ public static CompositeOption nettyBundles() {
             mavenBundle("io.netty", "netty-codec").versionAsInProject(),
             mavenBundle("io.netty", "netty-common").versionAsInProject(),
             mavenBundle("io.netty", "netty-transport").versionAsInProject(),
+            mavenBundle("io.netty", "netty-transport-native-unix-common").versionAsInProject(),
```
hhughes: Netty added `netty-transport-native-unix-common` as a dependency for `netty-handler` in 4.1.78-Final, see:
https://github.com/netty/netty/commit/bc085b06e3153bc975353671151c0c90e4265baa#diff-c8414afca33f9431351e310ca1203eff4c16093b1a0a2fc587f3b302de217e4aR60

hhughes:
This change was cherry-picked in so I'm closing this PR

PR #1663, opened 2023-06-27
JAVA-3042: Support testing against Java17

hhughes:
Add property 'testJavaHome' to specify a different JDK for surefire/failsafe to run tests to facilitate testing with different JDKs.

pom.xml:
- Add '--add-opens java.base/jdk.internal.util.random=ALL-UNNAMED' as maven-surefire-plugin argLine to support deep reflection for mockito, only loaded for JDK17

Dependency updates:
- jacoco-maven-plugin -> 0.8.10, resolves "Error while instrumenting path/to/class" with JDK17
- maven-bundle-plugin -> 5.1.1, resolves java.util.ConcurrentModificationException [FELIX-6259] with JDK17
- blockhound-junit-platform -> 1.0.8.RELEASE, earlier version did not pick up -XX:+AllowRedefinitionToAddDeleteMethods properly

Jenkinsfile:
- Add matrix axis for JABBER_VERSION for each of JDK8, JDK11, JDK17
- Always run maven with JDK8, use testJavaHome to set JDK version for testing

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,6 +18,24 @@ def initializeEnvironment() {
 
   env.MAVEN_HOME = "${env.HOME}/.mvn/apache-maven-3.3.9"
   env.PATH = "${env.MAVEN_HOME}/bin:${env.PATH}"
+
+  /*
+  * As of JAVA-3042 JAVA_HOME is always set to JDK8 and this is currently necessary for mvn compile and DSE Search/Graph.
+  * To facilitate testing with JDK11/17 we feed the appropriate JAVA_HOME into the maven build via commandline.
+  *
+  * Maven command-line flags:
+  * - -DtestJavaHome=/path/to/java/home: overrides JAVA_HOME for surefire/failsafe tests, defaults to environment JAVA_HOME.
+  * - -Ptest-jdk-N: enables profile for running tests with a specific JDK version (substitute N for 8/11/17).
+  *
+  * Note test-jdk-N is also automatically loaded based off JAVA_HOME SDK version so testing with an older SDK is not supported.
+  *
+  * Environment variables:
+  * - JAVA_HOME: Path to JDK used for mvn (all steps except surefire/failsafe), Cassandra, DSE.
+  * - JAVA8_HOME: Path to JDK8 used for Cassandra/DSE if ccm determines JAVA_HOME is not compatible with the chosen backend.
+  * - TEST_JAVA_HOME: PATH to JDK used for surefire/failsafe testing.
+  * - TEST_JAVA_VERSION: TEST_JAVA_HOME SDK version number [8/11/17], used to configure test-jdk-N profile in maven (see above)
+  */
+
   env.JAVA_HOME = sh(label: 'Get JAVA_HOME',script: '''#!/bin/bash -le
     . ${JABBA_SHELL}
     jabba which ${JABBA_VERSION}''', returnStdout: true).trim()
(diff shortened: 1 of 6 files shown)


hhughes:
Reopening from #1642

Jenkinsfile
```diff
@@ -431,7 +459,7 @@ pipeline {
           }
           stage('Build-Driver') {
             steps {
-              buildDriver(env.JABBA_VERSION)
+              buildDriver('default')
```
absurdfarce, line 462: Note: this relies on configuring Java to support some notion of a default (general Jabba usage notes [here](https://github.com/shyiko/jabba#usage)).  In our case this is configured via the install scripts for the Jenkins runners.

Jenkinsfile
```diff
@@ -403,15 +429,17 @@ pipeline {
                    '4.0',      // Development Apache CassandraⓇ
                    'dse-6.8.30' // Current DataStax Enterprise
           }
+          axis {
+            name 'JABBA_VERSION'
+            values '1.8',           // jdk8
+                   'openjdk@1.11',  // jdk11
+                   'openjdk@1.17'   // jdk17
+          }
```
absurdfarce, line 437: Not a thing to fix for this PR but something to think about in the future: JABBA_VERSION is an increasingly less-than-useful name for what this var does.  I just went to look at the env vars from a Jenkins run with Java17 + some C* instance and I get the following:

```
[2023-06-27T16:53:23.719Z] JABBA_HOME=/usr/lib/jabba
[2023-06-27T16:53:23.719Z] JABBA_SHELL=/usr/lib/jabba/jabba.sh
[2023-06-27T16:53:23.719Z] JABBA_VERSION=openjdk@1.17
[2023-06-27T16:53:23.719Z] JAVA11_HOME=/usr/lib/jabba/jdk/openjdk@1.11
[2023-06-27T16:53:23.719Z] JAVA8_HOME=/usr/lib/jabba/jdk/1.8
[2023-06-27T16:53:23.719Z] JAVA_HOME=/usr/lib/jabba/jdk/1.8
[2023-06-27T16:53:23.719Z] JAVA_HOME_BEFORE_JABBA=/usr/lib/jabba/jdk/openjdk@1.17.0
[2023-06-27T16:53:23.719Z] JDK=1.8
...
[2023-06-27T16:53:23.719Z] TEST_JAVA_HOME=/usr/lib/jabba/jdk/openjdk@1.17.0
[2023-06-27T16:53:23.719Z] TEST_JAVA_VERSION=17
...
[2023-06-27T16:53:23.719Z] Maven home: /home/jenkins/.mvn/apache-maven-3.3.9
[2023-06-27T16:53:23.719Z] Java version: 1.8.0_241, vendor: Oracle Corporation
[2023-06-27T16:53:23.719Z] Java home: /usr/lib/jabba/jdk/1.8/jre
[2023-06-27T16:53:23.719Z] Default locale: en, platform encoding: UTF-8
[2023-06-27T16:53:23.719Z] OS name: "linux", version: "5.4.0-1099-aws", arch: "amd64", family: "unix"
```

I want to be very clear: the output in question stems from the [initialize environment](https://github.com/datastax/java-driver/blob/4.x/Jenkinsfile#L49-L61) op.  More importantly all of it is correct; I'm not suggesting there's an error of any kind here.  I'm just saying it took me a minute to figure out why all those vars had those values... and I do kinda wonder if JABBA_VERSION should perhaps be replaced by TEST_JDK_VERSION or something more descriptive.

Again, absolutely _not_ something that needs to be resolved right now.

Jenkinsfile
```diff
@@ -53,7 +77,7 @@ ENVIRONMENT_EOF
     set +o allexport
 
     . ${JABBA_SHELL}
-    jabba use ${JABBA_VERSION}
+    jabba use 1.8
```
absurdfarce, line 80: This op is setting the Java version for what we eventually use to print Java and Maven versions.  Since build version is fixed now at 1.8 I'm not sure we care as much about printing the Java version we use to build things with... although as I think about it we may very well be making _that_ configurable before long to so it really doesn't make sense to take it out either.

But it seems like at a minimum we should print JVM info for the JVM we'll be using for building and for testing, something like:

```
jabba use 1.8 // eventually to be made configurable
echo "Java version for build"
java -version
jabba use $JABBA_VERSION
echo "Java version for tests"
java -version
jabba use 1.8
mvn -v
```

Again, also not a thing we need to fix now... more a marker for future improvements.

absurdfarce (approved):
Appears to do what it says on the tin.  It'll be really nice to get regular testing with Java17 in place!

absurdfarce:
One general thing I observed while looking into this: our integration test framework doesn't appear to display the Java version it's using when running tests (or if it did I couldn't find them).  We have one, maybe more spots where _Maven_ or _the Jenkinsfile_ displays Java version but there could be mistakes there or configs could be misapplied.  It's probably useful for the test framework to do something similar in order to be very explicit about what JVM it's running on.  Not with every test, certainly, but at least one time... maybe something similar to what we do with ccm versioning in CcmBridge.

PR #1643, opened 2023-06-12
Automatically register the codecs for CqlVectorType in CachingCodecRegistry

eolivelli:

PR #1666, opened 2023-06-28
JAVA-3061 Re-introduce an improved CqlVector, add support for accessing vectors directly as float arrays

absurdfarce:
Fixes in the same vein as #1643 adapted to play nicely with recent changes to the Java types representing CQL vectors.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
@@ -22,7 +22,6 @@ import com.datastax.oss.driver.shaded.guava.common.base.Splitter;
 import com.datastax.oss.driver.shaded.guava.common.collect.Iterables;
 import com.datastax.oss.driver.shaded.guava.common.collect.Streams;
 import edu.umd.cs.findbugs.annotations.NonNull;
-
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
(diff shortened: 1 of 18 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
```diff
@@ -0,0 +1,161 @@
...
+        this.list = list;
+    }
+
+    /**
+     * Retrieve the value at the specified index.  Modelled after {@link List#get(int)}
+     *
+     * @param idx the index to retrieve
+     * @return the value at the specified index
+     */
+    public T get(int idx) { return list.get(idx); }
```
ShaunakDas88: be mindful of `IndexOutofBoundsException`

core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
```diff
@@ -0,0 +1,161 @@
...
+
+    /**
+     * Update the value at the specified index.  Modelled after {@link List#set(int, Object)}
+
+     * @param idx the index to set
+     * @param val the new value for the specified index
+     * @return the old value for the specified index
+     */
+    public T set(int idx, T val) { return list.set(idx,val); }
```
ShaunakDas88: same concern as `get()`

core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
```diff
@@ -0,0 +1,181 @@
...
+   */
+  public static <V extends Number> CqlVector<V> newInstance(V... vals) {
+
+    // Note that Array.asList() guarantees the return of an array which implements RandomAccess
+    return new CqlVector(Arrays.asList(vals));
+  }
+
+  /**
+   * Create a new CqlVector that "wraps" an existing ArrayList. Modifications to the passed
+   * ArrayList will also be reflected in the returned CqlVector.
```
hhughes, line 63: It would be good to have some tests around this behavior, along with similar tests that modifying the vals var-args array in the other constructor is not reflected in the CqlVector

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/VectorCodecTest.java
```diff
@@ -49,21 +51,21 @@ public void should_encode() {
...
   }
 
   @Test
   public void should_encode_with_too_many_elements() {
-    List<Float> doubleVector = Lists.newArrayList(VECTOR);
-    doubleVector.addAll(VECTOR);
-    assertThat(encode(doubleVector)).isEqualTo(VECTOR_HEX_STRING);
+    Float[] doubledVectorContents = Arrays.copyOf(VECTOR_ARGS, VECTOR_ARGS.length * 2);
+    System.arraycopy(VECTOR_ARGS, 0, doubledVectorContents, VECTOR_ARGS.length, VECTOR_ARGS.length);
+    assertThat(encode(CqlVector.newInstance(doubledVectorContents))).isEqualTo(VECTOR_HEX_STRING);
```
hhughes, line 68: Why does encode not fail when encoding a 4-length vector with a 2-length codec? Is there a use-case for this?

absurdfarce:
Okay, this should largely be ready for validation from everybody involved.  I still need an explicit test for CqlVector itself; that's next, but my main goal was to get the API in front of everyone so that we could get buy-off on the general idea.

As of the most recent round of commits this should build a usable artifact.  @eolivelli I'm most interested in feedback from you confirming that this will address the use case you were addressing with the change in #1643.  There _should_ be equivalent changes to what you had there in this PR but I'd like explicit confirmation from you before we proceed with this thing.

core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
absurdfarce: Agreed.  There's a general need for a CqlVectorTest to validate behaviours for the value type itself.  That's next on my list.

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/VectorCodecTest.java
absurdfarce: Short version is that the codec loops through the expected number of dimension, pulling off a ByteBuffer for the next element and attempting to encode that.  If there are additional bytes after that they're just silently ignored.

There is no real use case for this, and it shouldn't happen in regular operation (barring some weird server error).  I included it in the test here in an attempt to document the differing behaviour between getting too little data vs. too much data.

hhughes: Should we have the codec enforce exact length on the input vector? I spent some time puzzling why encoded bytes length-4 vector was the same as for the length-2 version. I can see this leading to unexpected behavior if lengths get mixed up in customer code

core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
absurdfarce: This has been addressed by the newly-added CqlVectorTest

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/VectorCodecTest.java
absurdfarce: There's not a clean answer here.  This isn't the kind of thing that customers will manipulate manually.  Normally we'll get a type from the server saying something like "this field is a vector of subtype float with 10 dimensions" and we'll create a codec for that in response.  If the server then sends garbage data for fields with that type we have a problem but the user isn't actually involved in that interaction.

I can see this both ways.  It's not unreasonable to say "you told me you were going to send 10 dimensions and you sent 20, that's an error".  It's also not unreasonable to say "you told me you were sending 10 dimensions and you sent 20... well, at least I can do my job (and ignore the rest)".

It's worth noting that ListCodec has a similar problem.  In the case of a list the number of elements is passed along in an int before the actual elements appear, but (hypothetically) that number could be off and the server could send more data.  ListCodec doesn't care; it tries to decode the number of elements it believes are in the list and returns that.  There's no validation that I can see saying "did you read everything?".  If any such validation _does_ exist it would have to be higher up in the stack... which means it would apply to CqlVector as well.

core/src/test/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistryTestDataProviders.java
```diff
@@ -285,6 +286,55 @@ public static Object[][] collectionsWithCqlAndJavaTypes()
         ImmutableMap.of(
             ImmutableMap.of(udtValue, udtValue), ImmutableMap.of(tupleValue, tupleValue))
       },
+      // vexctors
```
hhughes: typo?

absurdfarce: Very much so.... fixing.

absurdfarce: Fixed.

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
```diff
@@ -371,6 +372,23 @@ protected GenericType<?> inspectType(@NonNull Object value, @Nullable DataType c
             inspectType(firstValue, cqlType == null ? null : ((MapType) cqlType).getValueType());
         return GenericType.mapOf(keyType, valueType);
       }
+    } else if (value instanceof CqlVector) {
+      CqlVector<?> set = (CqlVector<?>) value;
```
hhughes: nit: should this variable be `vector`?

hhughes (approved):
Added a few nit-pick comments, lgtm

core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CachingCodecRegistry.java
absurdfarce: Sure, that prolly makes more sense.  Flippin' copy/paste... 😦

absurdfarce: Fixed.

jshook (approved):
This looks easy to use and efficient!

msmygit (approved):

jeffbanks (approved):
LGTM. From perspective of flexibility for usage, I like the operations provided by `CqlVector`.

core/src/main/java/com/datastax/oss/driver/api/core/type/reflect/GenericType.java
```diff
@@ -15,10 +15,7 @@
  */
 package com.datastax.oss.driver.api.core.type.reflect;
 
-import com.datastax.oss.driver.api.core.data.CqlDuration;
-import com.datastax.oss.driver.api.core.data.GettableByIndex;
-import com.datastax.oss.driver.api.core.data.TupleValue;
-import com.datastax.oss.driver.api.core.data.UdtValue;
+import com.datastax.oss.driver.api.core.data.*;
```
hhughes: Nit: remove wildcard import

absurdfarce: On it.  Thanks for nothing IntelliJ... :growl:

hhughes (approved):
LGTM

absurdfarce:
Excellent, thanks @hhughes!

Given all of the feedback above and the various conversations in various other venues I'm going to call this good and merge this change now.  There are several downstream efforts which depend on a released driver that includes these changes and finalizing this and getting it merged is necessary to make that happen.  A huge thanks to _everybody_ who was involved in this process; thanks for making the Java driver's support for vectors even better!

PR #1670, opened 2023-06-30
JAVA-3077: ListenersIT intermittently failing with: Wanted but not in…

hhughes:
…voked: schemaListener1.onSessionReady

ListenersIT.java:
- Add comment explaining what methods are called from which thread
- add 500ms timeout to mockito verify to wait for slow callbacks
- add latches to wait for methods on custom listener subclasses

### DIFF
--- a/integration-tests/src/test/java/com/datastax/oss/driver/core/session/ListenersIT.java
+++ b/integration-tests/src/test/java/com/datastax/oss/driver/core/session/ListenersIT.java
@@ -17,6 +17,7 @@ package com.datastax.oss.driver.core.session;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.inOrder;
+import static org.mockito.Mockito.timeout;
 import static org.mockito.Mockito.verify;
 
 import com.datastax.oss.driver.api.core.CqlSession;
@@ -33,9 +34,12 @@ import com.datastax.oss.driver.api.core.tracker.RequestTracker;
 import com.datastax.oss.driver.api.testinfra.session.SessionUtils;
 import com.datastax.oss.driver.api.testinfra.simulacron.SimulacronRule;
 import com.datastax.oss.driver.categories.ParallelizableTests;
+import com.datastax.oss.driver.shaded.guava.common.util.concurrent.Uninterruptibles;
 import com.datastax.oss.simulacron.common.cluster.ClusterSpec;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import java.util.Collections;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import org.junit.ClassRule;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
@@ -90,6 +94,9 @@ public class ListenersIT {
                         .build())
                 .build()) {
 
+      // These NodeStateListeners are wrapped with SafeInitNodeStateListener which delays #onUp
+      // callbacks until #onSessionReady is called, these will all happen during session
+      // initialization
       InOrder inOrder1 = inOrder(nodeListener1);
       inOrder1.verify(nodeListener1).onSessionReady(session);
       inOrder1.verify(nodeListener1).onUp(nodeCaptor1.capture());
@@ -104,20 +111,29 @@ public class ListenersIT {
       assertThat(nodeCaptor2.getValue().getEndPoint())
           .isEqualTo(SIMULACRON_RULE.getContactPoints().iterator().next());
 
-      verify(schemaListener1).onSessionReady(session);
-      verify(schemaListener2).onSessionReady(session);
+      // SchemaChangeListener#onSessionReady is called asynchronously from AdminExecutor so we may
+      // have to wait a little
+      verify(schemaListener1, timeout(500).times(1)).onSessionReady(session);
+      verify(schemaListener2, timeout(500).times(1)).onSessionReady(session);
 
+      // Request tracker #onSessionReady is called synchronously during session initialization
       verify(requestTracker1).onSessionReady(session);
       verify(requestTracker2).onSessionReady(session);
 
       assertThat(MyNodeStateListener.onSessionReadyCalled).isTrue();
       assertThat(MyNodeStateListener.onUpCalled).isTrue();
 
-      assertThat(MySchemaChangeListener.onSessionReadyCalled).isTrue();
+      // SchemaChangeListener#onSessionReady is called asynchronously from AdminExecutor so we may
+      // have to wait a little
+      assertThat(
+              Uninterruptibles.awaitUninterruptibly(
+                  MySchemaChangeListener.onSessionReadyLatch, 500, TimeUnit.MILLISECONDS))
+          .isTrue();
 
       assertThat(MyRequestTracker.onSessionReadyCalled).isTrue();
     }
 
+    // CqlSession#close waits for all listener close methods to be called
     verify(nodeListener1).close();
     verify(nodeListener2).close();
 
@@ -163,14 +179,14 @@ public class ListenersIT {
 
   public static class MySchemaChangeListener extends SchemaChangeListenerBase {
 
-    private static volatile boolean onSessionReadyCalled = false;
+    private static CountDownLatch onSessionReadyLatch = new CountDownLatch(1);
     private static volatile boolean closeCalled = false;
 
     public MySchemaChangeListener(@SuppressWarnings("unused") DriverContext ignored) {}
 
     @Override
     public void onSessionReady(@NonNull Session session) {
-      onSessionReadyCalled = true;
+      onSessionReadyLatch.countDown();
     }
 
     @Override


integration-tests/src/test/java/com/datastax/oss/driver/core/session/ListenersIT.java
```diff
@@ -90,71 +94,102 @@ public void should_inject_session_in_listeners() throws Exception {
...
+      // - RequestTracker#onSessionReady
+      // AdminExecutor thread:
+      // - NodeStateListener#onUp
+      // - SchemaChangeListener#onSessionReady
+      // - *#close
       InOrder inOrder1 = inOrder(nodeListener1);
-      inOrder1.verify(nodeListener1).onSessionReady(session);
-      inOrder1.verify(nodeListener1).onUp(nodeCaptor1.capture());
+      inOrder1.verify(nodeListener1, timeout(500).times(1)).onSessionReady(session);
+      inOrder1.verify(nodeListener1, timeout(500).times(1)).onUp(nodeCaptor1.capture());
```
absurdfarce: The timeout argument shouldn't be necessary for NodeStateListener.onUp() here.  The test is built so that all NSLs are actually [SafeInitNodeStateListeners](https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/api/core/metadata/SafeInitNodeStateListener.java) which handle the possible out-of-order calls of on*() in relation to onSessionReady().  As a result any on*() method will get replayed once the session is available.

integration-tests/src/test/java/com/datastax/oss/driver/core/session/ListenersIT.java
```diff
@@ -90,71 +94,102 @@ public void should_inject_session_in_listeners() throws Exception {
...
+      // - NodeStateListener#onSessionReady
+      // - RequestTracker#onSessionReady
+      // AdminExecutor thread:
+      // - NodeStateListener#onUp
+      // - SchemaChangeListener#onSessionReady
+      // - *#close
       InOrder inOrder1 = inOrder(nodeListener1);
-      inOrder1.verify(nodeListener1).onSessionReady(session);
-      inOrder1.verify(nodeListener1).onUp(nodeCaptor1.capture());
+      inOrder1.verify(nodeListener1, timeout(500).times(1)).onSessionReady(session);
```
absurdfarce: We shouldn't need a timeout for this call, right?  I haven't validated against the code but won't this be called before session initialization is completed?

integration-tests/src/test/java/com/datastax/oss/driver/core/session/ListenersIT.java
```diff
@@ -90,71 +94,102 @@ public void should_inject_session_in_listeners() throws Exception {
...
-      assertThat(MyNodeStateListener.onSessionReadyCalled).isTrue();
-      assertThat(MyNodeStateListener.onUpCalled).isTrue();
-
-      assertThat(MySchemaChangeListener.onSessionReadyCalled).isTrue();
-
-      assertThat(MyRequestTracker.onSessionReadyCalled).isTrue();
+      verify(schemaListener1, timeout(500).times(1)).onSessionReady(session);
+      verify(schemaListener2, timeout(500).times(1)).onSessionReady(session);
+
+      verify(requestTracker1, timeout(500).times(1)).onSessionReady(session);
```
absurdfarce: RequestTracker only has the single callback method (so no interleaving between threads) and is called by the thread handling session initialization.  Like my comment above: do we need the timeout here or are we guaranteed that this will always be called before the session is returned?

integration-tests/src/test/java/com/datastax/oss/driver/core/session/ListenersIT.java
```diff
@@ -90,71 +94,102 @@ public void should_inject_session_in_listeners() throws Exception {
...
+
+      assertThat(
+              Uninterruptibles.awaitUninterruptibly(
+                  MySchemaChangeListener.onSessionReadyLatch, 500, TimeUnit.MILLISECONDS))
+          .isTrue();
+
+      assertThat(
+              Uninterruptibles.awaitUninterruptibly(
+                  MyRequestTracker.onSessionReadyLatch, 500, TimeUnit.MILLISECONDS))
+          .isTrue();
```
absurdfarce: Similar argument for the instances declared via config.  Since MyNodeStateListener already extends SafeInitNodeStateListener we should be okay here (as long as onSessionReady() is checked before onUp() for NSL ops), right?

Similar arguments for the other listeners.

integration-tests/src/test/java/com/datastax/oss/driver/core/session/ListenersIT.java
```diff
@@ -90,71 +94,102 @@ public void should_inject_session_in_listeners() throws Exception {
...
+                MyNodeStateListener.closeLatch, 500, TimeUnit.MILLISECONDS))
+        .isTrue();
+    assertThat(
+            Uninterruptibles.awaitUninterruptibly(
+                MySchemaChangeListener.closeLatch, 500, TimeUnit.MILLISECONDS))
+        .isTrue();
+    assertThat(
+            Uninterruptibles.awaitUninterruptibly(
+                MyRequestTracker.closeLatch, 500, TimeUnit.MILLISECONDS))
+        .isTrue();
```
absurdfarce: Do we think the close() methods are a problem yet?  This seems like a lot of code to address that case.  I guess I'm wondering if a single timeout check (or even a CountDownLatch perhaps) would do the job here.

I say all that having not actually looked at the relevant code yet.

integration-tests/src/test/java/com/datastax/oss/driver/core/session/ListenersIT.java
```diff
@@ -90,71 +94,102 @@ public void should_inject_session_in_listeners() throws Exception {
...
-      verify(requestTracker1).onSessionReady(session);
-      verify(requestTracker2).onSessionReady(session);
-
-      assertThat(MyNodeStateListener.onSessionReadyCalled).isTrue();
-      assertThat(MyNodeStateListener.onUpCalled).isTrue();
-
-      assertThat(MySchemaChangeListener.onSessionReadyCalled).isTrue();
-
-      assertThat(MyRequestTracker.onSessionReadyCalled).isTrue();
+      verify(schemaListener1, timeout(500).times(1)).onSessionReady(session);
```
absurdfarce, line 116: This seems like the heart of the problem; an event that's called from a thread totally unrelated to the thread governing where we're at in this code.  Seems like this is the truly essential change we need here.

absurdfarce (commented):
I think the changes here will resolve the issue identified by the ticket.  But I'm afraid we're doing quite a bit more as well, not all of which is necessary given the current structure of the driver code.  That isn't necessarily a huge problem, but in this case I'd argue that some of these changes make the test a modest amount harder to read, which in turn obfuscates it's intent a bit.  I'd prefer to avoid that if we possibly can.

Full disclosure: I haven't confirmed all of my assumptions against the code so it's quite possible I'm misunderstanding something.

integration-tests/src/test/java/com/datastax/oss/driver/core/session/ListenersIT.java
absurdfarce: As I think about this I don't think the analysis above is right.  SafeInitNodeStateListener will make sure we don't lose any event calls but it won't handle the concern about whether another thread has called this method when we aren't doing anything to ensure synchronization with that thread.

There is a big part of me that wishes we could just explicitly block until the admin thread has had a chance to notify all its listeners.  I'm not sure what such an implementation would look like (and we wouldn't want to connect it too strongly to driver internals) but it feels like we're dancing around that kind of solution here without actually embracing it.

hhughes: NodeStateListener#onSessionReady is called before the session's initFuture is completed so you right that both #onSessionReady and #onUp (called in SafeInitNodeStateListener#onSessionReady) will have been called by the time we get to this point in the test

integration-tests/src/test/java/com/datastax/oss/driver/core/session/ListenersIT.java
hhughes: close for all listeners is called before the session's closeFuture is completed so you are right and there were no issues with how close was being checked previously in the test, I'll revert those changes

integration-tests/src/test/java/com/datastax/oss/driver/core/session/ListenersIT.java
hhughes: Thank you for the review! You are correct, only the schema listener's callbacks are not being waited for by other mechanisms (session init/close) so these should be the only ones that need to change. I think we also need to keep the code that waits for MySchemaChangeListener#onSessionReady because the bg thread may not have gone through all the schema listeners, even if we verfied one was called earlier. I'll revert all the unnecessary synchronization/waiting and update the comments with my learnings here.

absurdfarce (approved):
Nice, I like where this landed!

PR #1195, opened 2019-02-21
Avoid full cluster scans in Connection.checkClusterName

haaawk:
Select without a partition key results in full cluster
scan that can hurt the performance of the cluster.
There is an easy solution to the problem by just
using a 'local' partition key for the query.

Signed-off-by: Piotr Jastrzebski <piotr@scylladb.com>

PR #1671, opened 2023-07-03
Fix LatencyAwarePolicy scale docstring

wprzytula:
A parenthesis was too far to the right.
> I'm contributing this code on behalf of my employer ScyllaDB. We have signed Contributor License Agreement in the past when we were contributing fixes to Java driver. For more context see https://github.com/datastax/java-driver/pull/1195

### DIFF
--- a/driver-core/src/main/java/com/datastax/driver/core/policies/LatencyAwarePolicy.java
+++ b/driver-core/src/main/java/com/datastax/driver/core/policies/LatencyAwarePolicy.java
@@ -709,7 +709,7 @@ public class LatencyAwarePolicy implements ChainableLoadBalancingPolicy {
      *
      * <pre>{@code d = (t - t') / scale
      * alpha = 1 - (ln(d+1) / d)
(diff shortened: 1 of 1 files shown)


absurdfarce:
Thanks for the contribution @wprzytula !

For the record: relevant code is [here](https://github.com/datastax/java-driver/blob/a9f13f468f73a76f69cfb267a406a3c2dd8199d9/driver-core/src/main/java/com/datastax/driver/core/policies/LatencyAwarePolicy.java#L624-L631).  Comparing this to the description in the Javadoc:

```
     * <pre>{@code d = (t - t') / scale
     * alpha = 1 - (ln(d+1) / d)
     * avg = alpha * l + (1 - alpha * prev)}</pre>
```

it's clear that d == scaledDelay and alpha == (1 - prevWeight).  Since we're then adding only prevWeight when we compute newAverage it's pretty clear that should be (1 - alpha) rather than (1 - alpha * prev).

absurdfarce (approved):

PR #1676, opened 2023-07-11
JAVA-3070: Make CqlVector and CqlDuration serializable

hhughes:


core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
```diff
@@ -90,7 +96,9 @@ public static <V extends Number> CqlVector<V> from(
     return new CqlVector(vals);
   }
 
-  private final List<T> list;
+  // transient because SerializationProxy is responsible for serializing/deserializing vector
+  // elements.
+  private final transient List<T> list;
```
hhughes: Todo: does this really need to be transient?

absurdfarce (approved):
Reviewed in a call with @hhughes.  Changes seem pretty robust; I very much like the fact that we aren't relying on both sides having the concrete List implementation for CqlVectors built from some random List impl.

core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
hhughes: no :)

PR #1678, opened 2023-07-11
JAVA-3080: Support PEM files for SSL

yukim:


yukim:
Rebased and updated the license header.

Note that the current implementation does not implement hot reloading of PEM files like CASSANDRA-19180 does for keystore file.

PR #1679, opened 2023-07-12
JAVA-3084: Add integration test coverage for ExtraTypeCodecs

hhughes:


integration-tests/src/test/java/com/datastax/oss/driver/core/type/codec/ExtraTypeCodecsIT.java
```diff
@@ -0,0 +1,301 @@
...
+                      SimpleStatement.builder(
+                              String.format(
+                                  "SELECT %s FROM extra_type_codecs_it WHERE key = ?", field.name))
+                          .addPositionalValue(key)
+                          .build())
+                  .one()
+                  .get(field.name, codec))
+          .isEqualTo(value);
+    }
+  }
```
absurdfarce, line 141: It's not essential, but opening a new codec for each codec we want to test is probably more than we need to do here.  We can apply codecs at read time directly to a row and if we use prepared statements we can also supply a codec when we generate a BoundStatement from a PreparedStatement.  Using that approach we can do everything through a single session:

```java
  private <T> void insertAndRead(TableField field, T value, TypeCodec<T> codec) {
    CqlSession session = SESSION_RULE.session();
    PreparedStatement psi =
        session.prepare(
            String.format("INSERT INTO extra_type_codecs_it (key, %s) VALUES (?, ?)", field.name));
    // write value under new key using provided codec
    UUID key = UUID.randomUUID();
    session.execute(psi.boundStatementBuilder().setUuid(0, key).set(1, value, codec).build());

    PreparedStatement pss =
        session.prepare(
            String.format("SELECT %s FROM extra_type_codecs_it WHERE key = ?", field.name));
    // read value using provided codec
    assertThat(
            session
                .execute(pss.boundStatementBuilder().setUuid(0, key).build())
                .one()
                .get(field.name, codec))
        .isEqualTo(value);
  }
```

The above passes for me locally against C* 4.0.0.

This is not a particularly efficient way to use prepared statements, but it's probably somewhat better than having to open and close a session for each type.  As I mentioned above it's not essential (although it probably is a bit faster); there's nothing wrong with what you have now.

absurdfarce (approved):
I like the overall approach taken here; this is a solid implementation and a clever way to handle the various types of codecs involved.  Left a comment on how to do this without having to open a session for each codec but it's not strictly required.

Nice work!

integration-tests/src/test/java/com/datastax/oss/driver/core/type/codec/ExtraTypeCodecsIT.java
hhughes: Thanks. Updated `insertAndRead` to use single session with bound statements per your suggestion. If the tests pass I'll merge this one :)

hhughes:
Test failures unrelated - merging

PR #1680, opened 2023-07-12
JAVA-3089: Forbid wildcard imports

hhughes:


pom.xml
```diff
@@ -588,6 +588,7 @@
             <compilerArg>-Xep:FutureReturnValueIgnored:OFF</compilerArg>
             <compilerArg>-Xep:PreferJavaTimeOverload:OFF</compilerArg>
             <compilerArg>-Xep:AnnotateFormatMethod:OFF</compilerArg>
+            <compilerArg>-Xep:WildcardImport:WARN</compilerArg>
```
absurdfarce, line 591: Does a "WARN" result actually fail the build or do we need to go to "ERROR" for that?

hhughes: It does - we're configuring `maven-compiler-plugin` with `<failOnWarning>true</failOnWarning>` (and this is how I found the violations fixed along with this change :) )

absurdfarce (approved):

hhughes:
Test failure unrelated

PR #1684, opened 2023-07-17
Compile 4.x driver with java11

hhughes:
… under the same pom.xml)

Update to api-plumber-doclet 2.0 (todo: remove dependence on SNAPSHOT build)


hhughes:
Moving branch to fork, see #1692 1692

PR #1686, opened 2023-07-19
JAVA-3094: Fix decrement query error

KSH-code:
If decrement with a negative value, then the query is made like `COLUMN = COLUMN--value`. so it can occur error because, double dashes (--) is used for comment.


absurdfarce:
Thanks for the PR @KSH-code!

KSH-code:
@absurdfarce Yes I have signed right now
Thanks!

KSH-code:
@absurdfarce @hhughes Hi, can you guys please check this pr if it will be merged or closed?

hhughes:
I'm curious what is the use case for a negative increment/decrement?

Additionally [this comment](https://github.com/apache/cassandra/blob/trunk/src/antlr/Parser.g#L1689) in the cql parser code makes me think that increment/decrement by a signed value is probably not supported/expected in variable assignment clauses.

If you agree with the above, perhaps a better fix here would be to throw an invalid argument somewhere in this code path when incrementing/decrementing with terms which are not positive integers or placeholders? Let me know what you think.

KSH-code:
The logic which occur error is simple
```
QueryBuilder.update(KEYSPACE, TABLE)
            .decrement(COLUMN_SCORE, QueryBuilder.literal(delta))
            .build()
```
if delta is a negative value, it represents such as `COLUMN = COLUMN--delta`.

I think the function manipulates a value is better than throwing error now.
The reason why it is better is because,
First, it makes breaking changes.
Second, increment function cannot allow negative values if decrement function does not allow negative value as well.

KSH-code:
@absurdfarce @hhughes I do not use Scylla now, so I have no time to improve this task now.
I am going to close this now.

KSH-code:
Can anybody check this again?

PR #1687, opened 2023-07-24
JAVA-3098: Improve ability to trace active connections across ControlConnection, BasicLoadBalancingPolicy and NodeRefresh classes

hhughes:


core/src/main/java/com/datastax/oss/driver/internal/core/control/ControlConnection.java
```diff
@@ -399,39 +415,44 @@ private void connect(
...
                       DriverChannel previousChannel = ControlConnection.this.channel;
                       ControlConnection.this.channel = channel;
                       if (previousChannel != null) {
                         // We were reconnecting: make sure previous channel gets closed (it may
                         // still be open if reconnection was forced)
                         LOG.debug(
-                            "[{}] Forcefully closing previous channel {}", logPrefix, channel);
+                            "[{}] Forcefully closing previous channel {}",
+                            logPrefix,
+                            previousChannel);
```
hhughes, line 455: Note: I think this was supposed to be `previousChannel` so fixing that at the same time

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/BasicLoadBalancingPolicy.java
```diff
@@ -422,4 +436,25 @@ protected NodeDistance computeNodeDistance(@NonNull Node node) {
...
+        logPrefix,
+        liveNodes.dcs().stream()
+            .map(
+                dc ->
+                    dc
+                        + ": "
+                        + liveNodes.dc(dc).stream()
+                            .map(Objects::toString)
+                            .collect(Collectors.joining(", ")))
+            .collect(Collectors.joining(", ")));
```
hhughes, line 458: let me know if theres a saner way to format these than to put everything into a single line and let the maven plugin figure it out 😅

hhughes:
Moved to PR from fork see #1711

PR #1689, opened 2023-07-26
add support for publishing percentile time series for the histogram m…

nparaddi-walmart:
…etrics cql-requests, cql-messages and throttling delay.

Motivation:

Histogram metrics is generating too many metrics overloading the promethous servers. if application has 500 Vms and 1000 cassandra nodes, The histogram metrics generates 100*500*1000 = 50,000,000 time series every 30 seconds. This is just too much metrics.  Let us say we can generate percentile 95 timeseries for for every cassandra nodes, then we only have 1*500 = 500 metrics and in applciation side, we can ignore the _bucket time series. This way there will be very less metrics.

Modifications:
add configurable pre-defined percentiles to Micrometer Timer.Builder.publishPercentiles. This change is being added to cql-requests, cql-messages and throttling delay.
Also this PR adds configuration option which switches aggregable histogram generation on/off for all metric flavors [default=on]

Result:
Based on the configuration, we will see additonal quantile time series for cql-requests, cql-messages and throttling delay histogram metrics.
Also there will be configuration option available to disable aggregable histogram generation if not needed


absurdfarce:
Thanks for the PR @nparaddi-walmart !

nparaddi-walmart:
i already signed the CLA

Username

nparaddi-walmart
Name

Nagappa Paraddi
Email

nagappa.paraddi@walmart.com
Company

Walmart Associates Inc
Status

Completed

nparaddi-walmart:
@absurdfarce you may be already aware. Just FYI. Snyk is failed with
"`snyk` requires an authenticated account. Please run `snyk auth` and try again."

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerNodeMetricUpdater.java
```diff
@@ -97,7 +97,7 @@ protected void cancelMetricsExpirationTimeout() {
   protected Timer.Builder configureTimer(Timer.Builder builder, NodeMetric metric, MetricId id) {
     DriverExecutionProfile profile = context.getConfig().getDefaultProfile();
     if (metric == DefaultNodeMetric.CQL_MESSAGES) {
-      return builder
+      builder
           .publishPercentileHistogram()
```
absurdfarce: I feel like we're only getting at half the solution if we leave this method call in unmodified.

I agree with your point that you can avoid the _bucket vals if you specify something in publish-percentiles but in that case you're (a) asking Micrometer to collect data for a histogram that you know up front you'll never use and (b) having to build logic in to select only certain tags for scraping.  Neither of those seem great.

You can address the second point by changing the export process to remove the histogram data... say by pruning out any line that contains the literal string "_bucket".  But (c) you're now introducing a lot of string processing every time you scrape metrics and (d) you haven't really addressed point (a) above; there's really no reason for Micrometer to collect this data if you aren't going to use it.

So really I think you're asking for _two_ things here:

1. The ability to specify percentiles for timers (and to have those percentiles published)
2. The ability to shut off the percentile histogram if desired

Your PR implements the first, but it sure seems like you should have the second as well to really solve this problem.

core/src/main/resources/reference.conf
```diff
@@ -1578,6 +1578,13 @@ datastax-java-driver {
         # Valid for: Micrometer.
         // slo = [ 100 milliseconds, 500 milliseconds, 1 second ]
         
+        # An optional list of percentiles to be published by Micrometer.
+        #
+        # If defined, the histogram is guaranteed to contain these percentiles alongside other
+        # buckets used to generate aggregable percentile approximations.
```
absurdfarce: We should probably be very explicit here in saying that these percentiles can **_NOT_** be aggregated across multiple nodes in the way that percentile histograms can.

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerNodeMetricUpdater.java
```diff
@@ -111,8 +111,17 @@ protected Timer.Builder configureTimer(Timer.Builder builder, NodeMetric metric,
...
           .percentilePrecision(
               profile.getInt(DefaultDriverOption.METRICS_NODE_CQL_MESSAGES_DIGITS));
+
+      if (profile.isDefined(DefaultDriverOption.METRICS_NODE_CQL_MESSAGES_PUBLISH_PERCENTILES)) {
+        builder.publishPercentiles(
+            profile.getDoubleList(DefaultDriverOption.METRICS_NODE_CQL_MESSAGES_PUBLISH_PERCENTILES)
+                .stream()
+                .mapToDouble(Double::doubleValue)
+                .toArray());
+      }
```
absurdfarce: There's already a lot of repetition in this file but let's start to minimize it by abstracting this out to a helper method:

```java
private double[] toDoubleArray(List<Double> doubleList) {                                          
  return doubleList.stream().mapToDouble(Double::doubleValue).toArray();               
} 
```

I would've thought we could have avoided some of the intermediate object creation here by just calling publishPercentiles() multiple times on the builder:

```java
       if (profile.isDefined(DseDriverOption.METRICS_SESSION_GRAPH_REQUESTS_PUBLISH_PERCENTILES)) {
         profile          
           .getDoubleList(DseDriverOption.METRICS_SESSION_GRAPH_REQUESTS_PUBLISH_PERCENTILES)
           .forEach(builder::publishPercentiles);
       } 
```

But weirdly that didn't work.  With this change the unit tests fail to pass, claiming that this assert:

```java
    assertThat(snapshot.percentileValues()).hasSize(3);
```

is only finding one item rather than three... which suggests that percentiles aren't just added together.

absurdfarce (changes requested):
Well done @nparaddi-walmart ... this is a good, solid PR!

A few smaller things, but the main issue to be addressed here is whether it makes sense to compute the percentile histogram if we know for sure it's going to be ignored.

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerNodeMetricUpdater.java
nparaddi-walmart: Regarding second problem. we can always whitelist what kind of metrics we allow. That is what we are doing in Walmart.  so we already have solution for second. So also this is configurable and adding new 3 metrics series.

nparaddi-walmart: also am ok to remove publishPercentileHistogram if percentiles are defined. let me know if that is the path we want to go.

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerNodeMetricUpdater.java
nparaddi-walmart: Thank you. added helper method. let me know anything else needed

core/src/main/resources/reference.conf
nparaddi-walmart: thanks. added comment

core/src/main/java/com/datastax/oss/driver/api/core/config/TypedDriverOption.java
```diff
@@ -409,6 +409,12 @@ public String toString() {
       new TypedDriverOption<>(
           DefaultDriverOption.METRICS_SESSION_CQL_REQUESTS_SLO,
           GenericType.listOf(GenericType.DURATION));
+  /** Optional pre-defined percentile of cql resquests to publish, as a list of percentiles . */
```
hhughes: nit: `resquests` is mis-spelled

core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOption.java
```diff
@@ -939,6 +939,30 @@ public enum DefaultDriverOption implements DriverOption {
...
+   * List of percentiles in double to be published by the node cql-messages metric. 95th percentile
+   * to represented as 0.95.
+   *
+   * <p>Value type: {@link java.util.List List}&#60;{@link Double}&#62;
+   */
+  METRICS_NODE_CQL_MESSAGES_PUBLISH_PERCENTILES(
+      "advanced.metrics.node.cql-messages.publish-percentiles"),
+  /**
+   * List of percentiles in double to be published by the throttling delay metric. 95th percentile
+   * to represented as 0.95.
```
hhughes: Can you rework the javadoc descriptions for enums in this class and in `DseDriverOption.java`? They are a little hard to understand. For example I don't think we need to repeat that the value type as this is stated in the `Value type:` section of the description.

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerNodeMetricUpdater.java
hhughes: I agree with @absurdfarce - if you're trying to turn off the metrics produced by `publishPercentileHistogram` because it's producing too many values there will be others in the same situation too. As pointed out earlier there is an efficiency benefit of not having those produced in the first place if they are not needed, and additionally I think there is a usability benefit of having that configured directly in the driver, rather than directing users to another configuration system to strip out these the metrics they don't want.

I also wonder that since micrometer supports publishing both agregable histogram and local histogram metrics simultaneously that having the driver configure only one or the other might be too restrictive. Instead, perhaps you could add a new config entry for toggling `publishPercentileHistogram` in each scenario, maybe named something along the lines of "publish aggregable histogram" to contrast with the `publishPercentiles` settings which could be re-named "publish local percentiles"?

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerMetricUpdater.java
```diff
@@ -159,4 +160,8 @@ protected DistributionSummary.Builder configureDistributionSummary(
       DistributionSummary.Builder builder, MetricT metric, MetricId id) {
     return builder.publishPercentileHistogram();
   }
+
+  protected double[] toDoubleArray(List<Double> doubleList) {
```
hhughes: nit: this could be static

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerNodeMetricUpdater.java
hhughes: optional: it would be nice not to have to repeat the driver option name in the conditional and in the getter, particularly with these long similar names, consider breaking the repeated pattern of `if(isDefined(option)) { ... profile.getDoubleList(option) ...` into a method

metrics/micrometer/src/test/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerNodeMetricUpdaterTest.java
```diff
@@ -190,6 +195,60 @@ public void should_create_timer(
     assertThat(timer.count()).isEqualTo(10);
     HistogramSnapshot snapshot = timer.takeSnapshot();
     assertThat(snapshot.histogramCounts()).hasSize(2);
+    assertThat(snapshot.percentileValues()).hasSize(3);
```
hhughes, line 196: nit: it would provide better coverage if you could assert the exact set of percentiles you expect here instead of just the number of items (i.e. `[0.75, 0.95, 0.99]`) - same feedback to the other test in `MicrometerSessionMetricUpdaterTest`

metrics/micrometer/src/test/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerNodeMetricUpdaterTest.java
```diff
@@ -174,6 +175,10 @@ public void should_create_timer(
     when(profile.isDefined(sla)).thenReturn(true);
     when(profile.getDurationList(sla))
         .thenReturn(Arrays.asList(Duration.ofMillis(100), Duration.ofMillis(500)));
+    when(profile.isDefined(percentiles)).thenReturn(true);
+    when(profile.getDoubleList(percentiles))
+        .thenReturn(
+            Arrays.asList(Double.valueOf(0.75), Double.valueOf(0.95), Double.valueOf(0.99)));
```
hhughes: nit: unneccessary boxing of doubles (and again in the other tests)

hhughes (changes requested):
Hi @nparaddi-walmart, thanks for your patience on this one! Biggest concern here is we'd like to provide users a way to turn off the large-volumed metrics in addition to providing config to add new ones. I've also added a few other pieces of optional feedback. Please hit me back if you have any concerns! Thanks

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerNodeMetricUpdater.java
nparaddi-walmart: Yes. because too many metrics, none of the walmart teams are enabled the metrics.
As i understand correctly, you would like two config one for "publish local percentiles" and another one for ""publish aggregable histogram"
i am ok to that.. correct me if I am wrong.

nparaddi-walmart: Another general clarification i would like to see that is this going to be merged before handing to over apache foundation?

hhughes: After some more thought and discussion on our side I think we've landed on one new configuration option which switches aggregable histogram generation on/off for all metric flavors [default=on], in addition to the per-metric publishPercentile settings you have already pulled together in this PR. If you could add that to this PR and tackle some of the other newly added feedback around doc strings then I think we'll be good to go here :)

core/src/main/java/com/datastax/oss/driver/api/core/config/DefaultDriverOption.java
nparaddi-walmart: updated the java doc

core/src/main/java/com/datastax/oss/driver/api/core/config/TypedDriverOption.java
nparaddi-walmart: fixed it

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerMetricUpdater.java
nparaddi-walmart: moved to static

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerNodeMetricUpdater.java
nparaddi-walmart: created separate static method

metrics/micrometer/src/test/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerNodeMetricUpdaterTest.java
nparaddi-walmart: removed boxing

metrics/micrometer/src/test/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerNodeMetricUpdaterTest.java
nparaddi-walmart: added the validation

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerNodeMetricUpdater.java
nparaddi-walmart: Thank you. added addtional configuration option for aggregable histogram generation

nparaddi-walmart: Note could not add junit cases for this new configuration as micrometer SimpleMeterRegistry does not generate _bucket metrics . I have changes ready to add junit test cases but it requires PrometheusMeterRegistry( prometheus dependency). I can add those changes this PR if needed

nparaddi-walmart: i have this junit test case id different branch. It needs micrometer-registry-prometheus dependecy. I can pull junit cases in this PR if thats ok
https://github.com/nparaddi-walmart/java-driver/blob/f5a9b39ff37831a61b8dcd0f4ba5e871fa2ccf8c/metrics/micrometer/src/test/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerNodeMetricUpdaterTest.java#L294

hhughes (approved):
Thank you for the changes! LGTM

hhughes:
One unrelated test-failure (ChannelPoolResizeTest#should_grow_during_reconnection / JAVA-3113)

PR #1694, opened 2023-07-28
JAVA-3100: Update jackson-databind to 2.13.4.1 and jackson-jaxrs-json-provider to 2.13.4 to address recent CVEs

hhughes:
Additional:
- Remove unused maven property legacy-jackson.version
- Remove unneccesary explicit dependencies on jackson-core and jackson-jaxrs-base

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,9 +57,8 @@
     <slf4j.version>1.7.26</slf4j.version>
     <reactive-streams.version>1.0.3</reactive-streams.version>
     <json.version>20230227</json.version>
-    <jackson.version>2.13.2</jackson.version>
-    <jackson-databind.version>2.13.2.2</jackson-databind.version>
-    <legacy-jackson.version>1.9.12</legacy-jackson.version>
+    <jackson.version>2.13.4</jackson.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -327,11 +326,6 @@
         <artifactId>javax.annotation-api</artifactId>
         <version>1.3.2</version>
       </dependency>
-      <dependency>
-        <groupId>com.fasterxml.jackson.core</groupId>
-        <artifactId>jackson-core</artifactId>
-        <version>${jackson.version}</version>
-      </dependency>
```
absurdfarce: I'm not 💯 sure I'd actually call this a bug but this is a pretty significant change in behaviour.  We're driving jackson-core versions by whatever jackson-databind indicates, which means we don't have the ability to update jackson-core directly.  That's _probably_ okay, but it is at least worth thinking about.

absurdfarce (approved):
I'm okay with this, although I'm not convinced that removing explicit versioning management for jackson-core is a good idea.  It's something we can always go back to in a future release if necessary, though, so it's not the end of the world either.

PR #1698, opened 2023-07-29
JAVA-3078: Provide support for building with Java 11 and Java 17

hhughes:
Details:
- Update to api-plumber-doclet 2.0.0
- Add org.jetbrains:annotations as additionalDependency for core-shaded
- Refactor table summary tag as caption element to conform to HTML5
- Refactor &lt;h3> usage to &lt;h2> to conform to heading tag ordering for HTML5
- Explicitly define annotationProcessorPaths for mapper-processor, slf4j-nop and gremlin-core
- [java11+] Upgrade errorprone to 2.19.1
- [java11+] Update custom javadoc leaks tag to full string: leaks-private-api
- Add build-java<span>-</span>8 profile for compiling with java 8
- Add build-java<span>-</span>11 profile for compiling with java 11
- [java11+] Exclude many new error prone checks enabled by default in the new version (see JAVA-3102)
- [java11+] Use release=8 instead of source/target=1.8 to automatically select correct bootstrap path
- [java11+] Use fork=true with maven-compiler-plugin to pick up compilerArgs
- [java11+] Load error-prone plugin using annotationProcessorPaths, per documentation guidance
- Refactor Jenkinsfile to compile project with the Java version selected in the pipeline matrix
- Remove test-jdk profiles and surefire/failsafe JVM overrides added in JAVA-3042
- Update org.apache.felix.framework to 7.0.1 to support java17 without forking new JVM (FELIX-6287)
- Update commons-configuration2 in OSGi BundleOptions to 2.9.0 for java11
- [java11+] Set JAVA_HOME=JAVA8_HOME in CcmBridge if using DSE which only supports java8

### DIFF
--- a/test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
+++ b/test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
@@ -44,6 +44,7 @@ import org.apache.commons.exec.ExecuteWatchdog;
 import org.apache.commons.exec.Executor;
 import org.apache.commons.exec.LogOutputStream;
 import org.apache.commons.exec.PumpStreamHandler;
+import org.apache.commons.exec.environment.EnvironmentUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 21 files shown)


test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
```diff
@@ -378,7 +379,21 @@ protected void processLine(String line, int logLevel) {
       executor.setStreamHandler(streamHandler);
       executor.setWatchdog(watchDog);
 
-      int retValue = executor.execute(cli);
+      // under java11+, ccm selects wrong java version for dse
+      // force it to use java8 by mapping JAVA8_HOME to JAVA_HOME
+      Map<String, String> ccmEnv = null;
+      String javaVersion = System.getProperty("java.version");
+      if (DSE_ENABLEMENT && !javaVersion.startsWith("1.8")) {
```
hhughes: Filed [DSP-23501](https://datastax.jira.com/browse/DSP-23501) to track fixing this on the ccm side

hhughes: Actually for theres a parameter `--jvm_version=8` available on `ccm start` with the version of ccm used with DSE. Once https://github.com/riptano/ccm-private/pull/79 get's merged we can remove this jvm-swapping code and use that approach instead.

hhughes:
```
com.datastax.dse.driver.api.core.graph.ClassicGraphGeoSearchIndexIT
com.datastax.dse.driver.api.core.graph.ClassicGraphTextSearchIndexIT
com.datastax.dse.driver.api.core.graph.CoreGraphGeoSearchIndexIT
com.datastax.dse.driver.api.core.graph.CoreGraphTextSearchIndexIT
```
A few DSE integration tests failing with jdk17 due to https://datastax.jira.com/browse/DSP-23506

hhughes:
Reopened from fork. See #1712

PR #1700, opened 2023-08-08
JAVA-3103: Update shaded-guava/guava to 32.0.1

hhughes:
Depends on https://github.com/datastax/java-driver-shaded-guava/pull/11 + new version of shaded-guava being published


hhughes:
Move to PR from fork: #1710

PR #1704, opened 2023-08-15
JAVA-3108 Support running integration test suite against a remote server via cql-proxy

absurdfarce:
Converting code from an [old local branch](https://github.com/datastax/java-driver/tree/integration_tests_with_cql_proxy_bmcguire) into something which can actually be merged


absurdfarce:
This PR has been replaced by https://github.com/datastax/java-driver/pull/1717

PR #1707, opened 2023-08-17
use commit hash instead of doc-specific tag for 3.6 docs ref

hhughes:


hhughes:
Confirmed I am able to build docs locally with this change, and that the 3.6 docs look reasonable (attaching screenshot)
Screenshot 2023-08-17 at 10 13 32 AM

absurdfarce (approved):
Agreed that this is a logical solution.  I'm not thrilled with adding commit hashes to docs.yaml but cases like this seems like exactly the kind of situation where you might need to do it.

PR #1717, opened 2023-08-22
JAVA-3108 Support running integration test suite against a remote server via cql-proxy

absurdfarce:
Converting code from a [local PR](https://github.com/datastax/java-driver/pull/1704) into a fork-based PR


hhughes (changes requested):
Testing review-slack integration

PR #1718, opened 2023-08-23
JAVA-3115: SimpleStrategy#equals should compare replicationFactor using ReplicationFactor#equals

hhughes:


absurdfarce (approved):
Certainly appears to have been an oversight during refactoring.  The additional tests are quite nice as well!

PR #1719, opened 2023-08-23
JAVA-3116: update surefire/failsafe to 3.0.0 to fix issue running tests with specified jvm

hhughes:
Additionally:
- Set --jvm_version=8 when running dse 6.8.19+ with graph workloads (DSP-23501)
- Update commons-configuration2 to 2.9.0 + deps in BundleOptions to support java17
- Update felix framework version to 7.0.1 for java17 (FELIX-6287)
- Pick up newer bndlib for ArrayIndexOutOfBounds error printed with OsgiGraphIT (bndtools/bnd issue#3405)
  - Update pax-url-wrap to 2.6.4 (and bring pax-url-reference up to the same version)
  - Force newer tinybundles version 3.0.0 (default 2.1.1 version required older bndlib)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,8 +68,9 @@
     <junit.version>4.13.2</junit.version>
     <logback.version>1.2.3</logback.version>
     <osgi.version>6.0.0</osgi.version>
-    <felix.version>6.0.3</felix.version>
+    <felix.version>7.0.1</felix.version>
     <pax-exam.version>4.13.4</pax-exam.version>
+    <pax-url.version>2.6.4</pax-url.version>
     <simulacron.version>0.11.0</simulacron.version>
     <jsr353-api.version>1.1.4</jsr353-api.version>
     <jersey.version>2.31</jersey.version>
@@ -79,7 +80,7 @@
     <rxjava.version>2.2.2</rxjava.version>
     <awaitility.version>4.0.3</awaitility.version>
     <apacheds.version>2.0.0-M19</apacheds.version>
-    <surefire.version>2.22.2</surefire.version>
+    <surefire.version>3.0.0</surefire.version>
     <graalapi.version>22.0.0.2</graalapi.version>
     <skipTests>false</skipTests>
     <skipUnitTests>${skipTests}</skipUnitTests>
@@ -269,12 +270,17 @@
       <dependency>
         <groupId>org.ops4j.pax.url</groupId>
         <artifactId>pax-url-wrap</artifactId>
-        <version>2.6.3</version>
+        <version>${pax-url.version}</version>
       </dependency>
       <dependency>
         <groupId>org.ops4j.pax.url</groupId>
         <artifactId>pax-url-reference</artifactId>
-        <version>2.6.2</version>
+        <version>${pax-url.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.ops4j.pax.tinybundles</groupId>
+        <artifactId>tinybundles</artifactId>
+        <version>3.0.0</version>
       </dependency>
       <dependency>
         <groupId>org.glassfish</groupId>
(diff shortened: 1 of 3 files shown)


hhughes:
Test failures appear to caused by flaky `ccm start` in `DefaultLoadBalancingPolicyIT` + [JAVA-3117](https://datastax-oss.atlassian.net/browse/JAVA-3117)

[JAVA-3117]: https://datastax-oss.atlassian.net/browse/JAVA-3117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

pom.xml
```diff
@@ -269,12 +270,17 @@
...
       <dependency>
         <groupId>org.ops4j.pax.url</groupId>
         <artifactId>pax-url-reference</artifactId>
-        <version>2.6.2</version>
+        <version>${pax-url.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.ops4j.pax.tinybundles</groupId>
+        <artifactId>tinybundles</artifactId>
+        <version>3.0.0</version>
```
absurdfarce, line 283: Just to make sure my understanding is correct: all the version upgrades (both in this pom as well as in BundleOptions.java) are focused only on upgrading deps for the standalone OSGi tests, right?  We're not updating anything here that would be visible to users of the Java driver (whether they use OSGi or not), correct?

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
```diff
@@ -296,7 +300,10 @@ public void resume(int n) {
   }
 
   public void start(int n) {
-    execute("node" + n, "start");
+    List<String> cmdAndArgs = Lists.newArrayList("node" + n, "start");
+    overrideJvmVersionForDseWorkloads()
+        .ifPresent(jvmVersion -> cmdAndArgs.add(String.format("--jvm_version=%d", jvmVersion)));
+    execute(cmdAndArgs.toArray(new String[0]));
```
absurdfarce, line 306: I don't _love_ the duplication of the --jvm_version handling here across two instances of start() but fixing it is harder than one might like.  There's a dependency not only on class member fields (in the form of jvmArgs) but also method call args (in the form of  n in the node case) that make abstracting it kinda ugly.  It's prolly okay to go with for now; we can always improve it later if somebody hits upon a good idea for doing so.

absurdfarce (approved):
A few small notes, neither of them nearly enough to request changes here.  Unless my understanding of the goal here is considerably different than what it actually is this should work quite well.

pom.xml
hhughes:
```
% mvn dependency:tree -Dincludes=org.ops4j.pax.tinybundles:tinybundles
[INFO] --- dependency:3.1.2:tree (default-cli) @ java-driver-osgi-tests ---
[INFO] com.datastax.oss:java-driver-osgi-tests:jar:4.17.1-SNAPSHOT
[INFO] \- org.ops4j.pax.exam:pax-exam-junit4:jar:4.13.4:test
[INFO]    \- org.ops4j.pax.exam:pax-exam-spi:jar:4.13.4:test
[INFO]       \- org.ops4j.pax.tinybundles:tinybundles:jar:3.0.0:test
[INFO] 
```

`tinybundles` is only used by osgi-tests under the `test` scope

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
hhughes: Agreed. This is just test code so we can tidy it up later if needed.

PR #1722, opened 2023-08-24
JAVA-3114 Attempting to fix OSGi test issues after shading was implemented

absurdfarce:


absurdfarce:
This is no longer a going concern, it's been superseded by more complete work from @hhughes in https://github.com/datastax/java-driver/pull/1724

PR #1723, opened 2023-08-25
Fix hyperlink on the https://docs.datastax.com/en/developer/java-driv…

msmygit:
…er/latest/manual/mapper/mapper/#mapper-builder website


hhughes (approved):
LGTM

PR #1725, opened 2023-08-31
JAVA-3055 CqlPrepareAsyncProcessor must handle cancellations of the r…

tomas-bilka:
[JAVA-3055](https://datastax-oss.atlassian.net/browse/JAVA-3055)
CqlPrepareAsyncProcessor must handle cancellations of the returned Future

[JAVA-3055]: https://datastax-oss.atlassian.net/browse/JAVA-3055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/cql/CqlPrepareAsyncProcessor.java
@@ -151,7 +151,7 @@ public class CqlPrepareAsyncProcessor
                     if (error != null) {
                       mine.completeExceptionally(error);
                       cache.invalidate(request); // Make sure failure isn't cached indefinitely
-                    } else if (mine.isCancelled()){
+                    } else if (mine.isCancelled()) {
                       cache.invalidate(request);
                       process(request, session, context, sessionLogPrefix);
                     } else {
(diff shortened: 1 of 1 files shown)


hhughes:
Hi @tomas-bilka thank you for the PR.

tomas-bilka:
I just signed the CLA.

lucboutier:
I made another PR for 3055 some time ago. Then had a review from a colleague for a better and simpler implementation which I believe is quite nice.

See
https://github.com/apache/cassandra-java-driver/pull/1757

mcoolive:
Any progress?

aerain:
I think this approach doesn't seem to be able to prevent the underlying problem.

There is edge case
When two duplicate prepare requests A and B are received (assuming A arrived slightly earlier), the cancellation of one request propagates to the other.

1. Request A
2. Cache loaded with request A. Asynchronous prepare call to Cassandra (referred to as 'a')
3. Request B
4. B waits for a response from the cache based on request A
5. cancel() in request A
6. B receives the cancel() from request A and propagates a CancellationException
7. Response from 'a' completes. Upon receiving the cancellation of A, cache invalidation proceeds at this time

In my opinion, lucboutier's PR looks better to me.

absurdfarce:
I agree the defensive copy approach is probably a more robust way to fix this issue.  Would be nice if we could more clearly disentangle the actors involved as well.

absurdfarce:
Closing in favor of #1757

PR #1726, opened 2023-09-01
WIP: address transport errors seen during astra coordinator deployment

hhughes:


driver-core/src/main/java/com/datastax/driver/core/SessionManager.java
```diff
@@ -521,7 +521,7 @@ void onDown(Host host) throws InterruptedException, ExecutionException {
     // not necessary
     // since updateCreatedPools should take care of it. But better protect against non well behaving
     // policies.
-    removePool(host).force().get();
+    removePool(host).get();
```
hhughes, line 524: Wait for inflight requests to complete before closing connections rather that forcing them to abort immediately. The final response should end up closing the connection once it's been processed and if any inflight operations time out, the connection will be force closed automatically by `ConnectionReaper` after 2 * `readTimeoutMillis`.

It's worth noting in my testing where I was redeploying the coordinator by the removing the special label it looked like the operator only waits 5-7 seconds after issuing the `REMOVED_NODE` before closing connections on the coordinator. Since the default for `readTimeoutMillis` is 12 seconds (so client waits a total of 24 seconds before forcing them closed) we may need to bump up the amount of time the operator allows for clients to gracefully close their connections.

driver-core/src/main/java/com/datastax/driver/core/ControlConnection.java
```diff
@@ -617,10 +617,6 @@ private static void updateInfo(
       // want to setup a different port for SSL and non-SSL conns).
       InetAddress nativeAddress = row.getInet("native_transport_address");
       int nativePort = row.getInt("native_transport_port");
-      if (cluster.getCluster().getConfiguration().getProtocolOptions().getSSLOptions() != null
-          && !row.isNull("native_transport_port_ssl")) {
-        nativePort = row.getInt("native_transport_port_ssl");
-      }
```
hhughes, line 623: This code added for JAVA-2967 ends up overwriting the `native_transport_port` for the initial control connection before it gets stored in Metadata#hosts. When we later receive a `REMOVE_NODE` event for this host we are unable to find the entry to remove because the broadcast-rpc port does not match. This results in the driver skipping the graceful connection closing code and continuing to use the connection until it's suddenly closed serverside by the operator killing the coordinator pod.

In it's current state, removing the use of `native_transport_port_ssl` column probably breaks some DSE stuff. We should revisit that earlier patch to figure out a solution that works for both backends.

hhughes:
Closing out this experiment. Fixes included will be tracked under JAVA-3126 & JAVA-3125

PR #1734, opened 2023-10-04
Prioritize a-records with fewest connections when reconnecting to SNI endpoints

hhughes:


hhughes:
This version of the change didn't work because the initial endpoint objects are created before the driver context. Fixing that is much more involved so closing out this PR for now as it's unlikely to make it past the experiment stage.

PR #1735, opened 2023-10-20
JAVA-3131: Add #retrieve method to EndPoint for when caller does not …

hhughes:
…need the endpoint to be proactively resolved

Refactor existing usages of EndPoint#resolve to use retrieve when resolved ip addresses are not needed.

### DIFF
--- a/core/src/main/java/com/datastax/oss/driver/api/core/metadata/EndPoint.java
+++ b/core/src/main/java/com/datastax/oss/driver/api/core/metadata/EndPoint.java
@@ -41,6 +41,17 @@ public interface EndPoint {
   SocketAddress resolve();
 
   /**
+   * Returns a possibly unresolved instance to a socket address.
+   *
+   * <p>This should be called when the address does not need to be proactively resolved. For example
+   * if the node hostname or port number is needed.
+   */
+  @NonNull
+  default SocketAddress retrieve() {
+    return resolve();
+  }
+
+  /**
    * Returns an alternate string representation for use in node-level metric names.
    *
    * <p>Because metrics names are path-like, dot-separated strings, raw IP addresses don't make very
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultEndPoint.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultEndPoint.java
@@ -38,6 +38,11 @@ public class DefaultEndPoint implements EndPoint, Serializable {
   @NonNull
   @Override
   public InetSocketAddress resolve() {
+    return retrieve();
+  }
+
+  @Override
+  public InetSocketAddress retrieve() {
     return address;
   }
 
--- a/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/SniEndPoint.java
+++ b/core/src/main/java/com/datastax/oss/driver/internal/core/metadata/SniEndPoint.java
@@ -18,6 +18,7 @@
 package com.datastax.oss.driver.internal.core.metadata;
 
 import com.datastax.oss.driver.api.core.metadata.EndPoint;
+import com.datastax.oss.driver.shaded.guava.common.annotations.VisibleForTesting;
 import com.datastax.oss.driver.shaded.guava.common.primitives.UnsignedBytes;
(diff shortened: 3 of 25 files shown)


core/src/main/java/com/datastax/oss/driver/api/core/metadata/EndPoint.java
```diff
@@ -40,6 +40,17 @@ public interface EndPoint {
...
+  /**
+   * Returns a possibly unresolved instance to a socket address.
+   *
+   * <p>This should be called when the address does not need to be proactively resolved. For example
+   * if the node hostname or port number is needed.
+   */
+  @NonNull
+  default SocketAddress retrieve() {
+    return resolve();
+  }
```
absurdfarce, line 52: I'm wondering if it makes sense to provide a default implementation for the new method.  It's unlikely we'll be adding a new endpoint type anytime soon but failing to distinguish between "look up the addresses again" and "give me the addresses you got last time you looked them up" is what got us into this situation in the first place.  I kinda feel like having this as a default might lend itself to landing back in exactly that situation.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DefaultEndPoint.java
```diff
@@ -38,6 +38,11 @@ public DefaultEndPoint(InetSocketAddress address) {
   @NonNull
   @Override
   public InetSocketAddress resolve() {
+    return retrieve();
```
absurdfarce, line 41: Nit: save yourself the method call overhead here and just return address directly... ?

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/SniEndPointTest.java
```diff
@@ -0,0 +1,142 @@
...
+  @Spy private SniEndPoint sniEndPoint = new SniEndPoint(SNI_ADDRESS, SERVER_NAME);
+
+  private static InetAddress[] createAddresses(String... addrs) {
+    return Stream.of(addrs)
+        .map(
+            addr -> {
+              try {
+                int[] comp = Arrays.stream(addr.split("\\.")).mapToInt(Integer::parseInt).toArray();
+                return InetAddress.getByAddress(
+                    new byte[] {(byte) comp[0], (byte) comp[1], (byte) comp[2], (byte) comp[3]});
```
absurdfarce, line 51: Is there a particular reason to jump through the splitting/to Int/toArray ops here rather than just using InetAddress.getByName()?  I don't think it does a reverse lookup if you just give it a string containing an IP address but I could be wrong.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/SniEndPoint.java
```diff
@@ -64,14 +69,32 @@ public InetSocketAddress resolve() {
...
 
+  @VisibleForTesting
+  InetAddress[] resolveARecords() throws UnknownHostException {
+    // moving static call to method to allow mocking in tests
+    return InetAddress.getAllByName(proxyAddress.getHostName());
+  }
+
+  @Override
+  public InetSocketAddress retrieve() {
+    return proxyAddress;
```
absurdfarce, line 95: Don't we want to cache the last value returned by resolve() and return _that_ here instead?

absurdfarce (commented):
A few minor things but overall I very much like where this change is headed!

core/src/test/java/com/datastax/oss/driver/internal/core/metadata/SniEndPointTest.java
SiyaoIsHiding: Sorry, I am confused. Could you pls explain more?
`InetAddress.getByAddress` does not do reverse lookup. `InetAddress.getByName` will look it up. We don't want it to look up. Therefore, I think `InetAddress.getByAddress` is the one we should use. :)

SiyaoIsHiding:
Work taken over to https://github.com/apache/cassandra-java-driver/pull/1919

absurdfarce:
Replaced by https://github.com/apache/cassandra-java-driver/pull/1919.  Review comments will be handled there.

PR #1738, opened 2023-10-30
Update CcmBridge#getCurrentJvmMajorVersion to use regular expressions

hhughes:


test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
```diff
@@ -430,17 +433,15 @@ private static File createTempStore(String storePath) {
    *
    * @return major version of current JVM
    */
-  private static int getCurrentJvmMajorVersion() {
+  @VisibleForTesting
+  static int getCurrentJvmMajorVersion() {
```
absurdfarce, line 440: Nit: I'm kinda wondering if instead of making this visible for testing we don't just change things so that all the logic lives in a getCurrentJvmMajorVersion() impl which takes a version string as an arg:

```java
private static int getCurrentJvmMajorVersion() {
   return getJvmMajorVersion(System.getProperty("java.version"));
}

static int getJvmMajorVersion(String versionSysprop) {
   // Same logic in here (without the sysprop access)
}
```

Would probably simplify your tests quite a bit.

test-infra/src/main/java/com/datastax/oss/driver/api/testinfra/ccm/CcmBridge.java
```diff
@@ -430,17 +433,15 @@ private static File createTempStore(String storePath) {
...
     String version = System.getProperty("java.version");
-    if (version.startsWith("1.")) {
-      version = version.substring(2, 3);
-    } else {
-      int dot = version.indexOf(".");
-      if (dot != -1) {
-        version = version.substring(0, dot);
-      }
+    Pattern pattern = Pattern.compile("^(1\\.)?(?<major>[0-9]+)\\..*$");
```
absurdfarce: Nit: I don't think the "$" is doing much for you here

absurdfarce (approved):
I'm the first guy to applaud the use of regexes for... well, pretty much anything really.  And I like the use of named capture groups here.  But I do kinda wonder if we can't do the same kinda thing with just the following:

- Splitting based on dot
- Dropping a leading "1" if if's the first item in the result
- Returning the first item after the drop
- Throwing an exception if we don't find any dots

Am I missing an obvious case that makes that fall down?

absurdfarce:
Note: Jenkins failures are entirely derived from the switch to the ASF repo

PR #1739, opened 2023-10-30
JAVA-3101: CqlVector#from subtypeCodec parameter missing javadoc description

SiyaoIsHiding:


java-driver-parent.ipr
```diff
@@ -0,0 +1,120 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
absurdfarce: In general we don't check in internal IntelliJ files.  Both this .ipr file and the .iws file below fall into this category so my suggestion would be to remove these from your PR!

core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
```diff
@@ -84,7 +84,7 @@ public static <V extends Number> CqlVector<V> newInstance(List<V> list) {
    * </code> call on some CqlVector should return a CqlVector that is equal to the origin instance.
    *
    * @param str a String representation of a CqlVector
-   * @param subtypeCodec
+   * @param subtypeCodec the TypeCodec to parse the String representation
```
absurdfarce: This is true but it doesn't really tell us which string representation we're talking about :)

Vectors in CQL have a form that looks something like this:

```
org.apache.cassandra.db.marshal.VectorType(org.apache.cassandra.db.marshal.FloatType,3)
```

The first parameter there is the subtype of this vector, that is the type of values it contains.  The second is it's size parameter.

The subtypeCodec in CqlVector is intended to parse the string representation of the subtype so that we can build new instances of that subtype for the values contained by the vector.  So you probably want to say something like "the codec to parse the string representation of subtype values" here (or something similar) in order to make it very clear what values we're talking about.

Make sense?

absurdfarce (commented):
Looks good @SiyaoIsHiding !  A few small things to take a look at but you're off to a great start!

java-driver-parent.ipr
SiyaoIsHiding: I can't believe I checked these and didn't notice that :man_facepalming: . So embarrassing. Just reverted.

core/src/main/java/com/datastax/oss/driver/api/core/data/CqlVector.java
SiyaoIsHiding: Thank you for your detailed explanation! Just changed to `the TypeCodec to parse the string representation of subtype values`.

PR #1743, opened 2023-11-02
JAVA-3051: Memory leak

SiyaoIsHiding:


core/src/main/java/com/datastax/oss/driver/internal/core/metadata/LoadBalancingPolicyWrapper.java
```diff
@@ -31,13 +31,7 @@
...
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Queue;
-import java.util.Set;
+import java.util.*;
```
hhughes: please remove wildcard imports (this should result in a warning that fails the build)

SiyaoIsHiding:
We explored several alternatives for unit testing this memory leak fix but decided to give up testing it, as we could not find a reliable way to test the garbage collector's behavior.

### The unit test we wrote and considered:
https://github.com/apache/cassandra-java-driver/blob/cb61eb8c34d1dd7417c03fec95d6e13e9185d537/core/src/test/java/com/datastax/oss/driver/internal/core/metadata/LoadBalancingPolicyWrapperMemoryLeakTest.java#L118-L142

This test:
1. creates two `DefaultNode`
2. creates two `WeakReference` pointing to the nodes, just to poke their existence later
3. initializes the policy
4. clear all the strong references
5. requests for garbage collection
6. verify the node is collected

### We checked that
1. In my local environment (Zulu 8.72.0.17-CA-macos-aarch64), this test will succeed, and if I revert the changes from `WeakHashMap` to the strong `HashMap`, this test will fail.
2. Before all the strong references are cleared, poking the memory, the referring objects to the `weakNode2` are:
  a. `weakNode2` and `weakReference2`
  b. `InterceptedInvocation`s by Mockito
  c. `HashMap` of `allNodes` stored in `when(metadata.getNodes()).thenReturn(allNodes);`
  d. wanted in `Equals` statement in ` await().atMost(10, TimeUnit.SECONDS).until(() -> weakReference2.get() == null);`
  These are all expected and no reference is leaked.
3. If `evaluateDistance` of nodes does not return `IGNORED`, they will be stored in `BasicLoadBalancingPolicy.liveNodes`. But nodes there can be removed later by `onDown` or `onRemoved`. We suppose this is intended.

### We considered that
According to [this](https://stackoverflow.com/questions/11174328/testing-weakreference/36917692) post, `System.gc()` is more like a request/hint that some JVM will ignore, and there is no reliable way to force garbage collection. This means the test above may fail in other environments, but the last thing we want is a flaky test. 

We think workarounds like generating a huge amount of garbage to trigger garbage collection may not be worth it, either.

Therefore, we concluded that no test may be the best choice for now, and the checks we perform above may be sufficient.

hhughes (commented):
This change will allow entries to be be dropped from `LoadBalancingPolicyWrapper#distances` but I'm not entirely convinced there aren't other places where strong references to the `Node` object will remain.

In `DefaultLoadBalancingPolicy` there are `responseTimes` and `upTimes` maps which use `Node` as the key with a `ConcurrentHashMap` and I don't see where entries are ever removed so this likely will continue to hold these references (although with `upTimes` it doesn't look like items are ever added).

`ControlConnection` maintains two weak hash maps - `lastDistanceEvents` and `lastStateEvents` - where both the value types `DistanceEvent` and `NodeStateEvent` hold a hard reference to the `Node`, per the `WeakHashMap` [docs](https://docs.oracle.com/javase/8/docs/api/java/util/WeakHashMap.html) it looks like this will prevent the entries being cleaned up:
>  The value objects in a WeakHashMap are held by ordinary strong references. Thus care should be taken to ensure that value objects do not strongly refer to their own keys, either directly or indirectly, since that will prevent the keys from being discarded

Likely there are more places too.

I think it could be good to set up at least a one-off test which reproduces the leak from the original ticket and confirm that this change (and possibly the others mentioned above) successfully prevent the leak before marking this one as completed.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -96,14 +100,39 @@ public class DefaultLoadBalancingPolicy extends BasicLoadBalancingPolicy impleme
...
         profile.getBoolean(DefaultDriverOption.LOAD_BALANCING_POLICY_SLOW_AVOIDANCE, true);
+    CacheLoader<Node, AtomicLongArray> cacheLoader =
+        new CacheLoader<Node, AtomicLongArray>() {
+          @Override
+          public AtomicLongArray load(Node key) throws Exception {
+            // The array stores at most two timestamps, since we don't need more;
+            // the first one is always the least recent one, and hence the one to inspect.
+            long now = nanoTime();
+            AtomicLongArray array =
+                responseTimes.asMap().containsKey(key) ? responseTimes.get(key) : null;
```
hhughes: consider refactoring to `responseTimes.getIfPresent()` to avoid conversion to a map and potential race between contains and get

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -96,14 +100,39 @@ public class DefaultLoadBalancingPolicy extends BasicLoadBalancingPolicy impleme
...
+            if (array == null) {
+              array = new AtomicLongArray(1);
+              array.set(0, now);
+            } else if (array.length() == 1) {
+              long previous = array.get(0);
+              array = new AtomicLongArray(2);
+              array.set(0, previous);
+              array.set(1, now);
```
hhughes: nit: construction + sets could be replaced with `new AtomicLongArray(new long[] {previous, now})` unless there is a good reason that performs poorly

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -96,14 +100,39 @@ public class DefaultLoadBalancingPolicy extends BasicLoadBalancingPolicy impleme
...
+            } else {
+              array.set(0, array.get(1));
+              array.set(1, now);
+            }
+            return array;
+          }
+        };
+    this.responseTimes = CacheBuilder.newBuilder().weakValues().build(cacheLoader);
```
hhughes: `responseTimes` should have `weakKeys` in order to drop entries when `Node` objects are finalized (`weakValues` will result in entries being dropped once the AtomicArray references go out of scope)

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -276,38 +305,22 @@ protected boolean isBusy(@NonNull Node node, @NonNull Session session) {
...
-    if (responseTimes.containsKey(node)) {
-      AtomicLongArray array = responseTimes.get(node);
-      if (array.length() == 2) {
-        long threshold = now - RESPONSE_COUNT_RESET_INTERVAL_NANOS;
-        long leastRecent = array.get(0);
-        return leastRecent - threshold < 0;
-      }
+    AtomicLongArray array;
+    try {
+      array = responseTimes.asMap().containsKey(node) ? responseTimes.get(node) : null;
```
hhughes: consider refactoring as `responseTimes.getIfPresent(node)`

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -276,38 +305,22 @@ protected boolean isBusy(@NonNull Node node, @NonNull Session session) {
...
-            array = new AtomicLongArray(2);
-            array.set(0, previous);
-            array.set(1, now);
-          } else {
-            array.set(0, array.get(1));
-            array.set(1, now);
-          }
-          return array;
-        });
+    responseTimes.refresh(node);
```
hhughes: maybe add some comments about why this will reload the key synchronously (LoadingCache#reload says this may be async but i don't think it is here)

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
SiyaoIsHiding: it currently becomes 
```java
            AtomicLongArray array = responseTimes.getIfPresent(key);
            if (array == null) {
              array = new AtomicLongArray(new long[] {-1, now});
            } else {
              array = new AtomicLongArray(new long[] {array.get(1), now});
            }
```
Do you think this implementation is proper?

hhughes: the [documentation](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--) claims nanotime can be negative so I don't think you can use `-1` to represent a special value

SiyaoIsHiding: thx, reverted

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -276,38 +303,24 @@ protected boolean isBusy(@NonNull Node node, @NonNull Session session) {
...
-      if (array.length() == 2) {
-        long threshold = now - RESPONSE_COUNT_RESET_INTERVAL_NANOS;
-        long leastRecent = array.get(0);
-        return leastRecent - threshold < 0;
-      }
-    }
-    return true;
+    AtomicLongArray array = responseTimes.getIfPresent(node);
+    if (array == null) return true;
+    else if (array.length() == 2) {
```
hhughes: nit: consider collapsing the null check into this conditional so there is only one irregular state return value (`return true`)

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/LoadBalancingPolicyWrapper.java
```diff
@@ -180,14 +181,17 @@ private void processNodeStateEvent(NodeStateEvent event) {
         return; // ignore
       case RUNNING:
         for (LoadBalancingPolicy policy : policies) {
-          if (event.newState == NodeState.UP) {
-            policy.onUp(event.node);
+          DefaultNode node = event.node.get();
+          if (node == null) {
+            LOG.info("[{}] Node for this event was removed, ignoring: {}", logPrefix, event);
```
hhughes: Info-level log might be a bit high for this notice as there isn't really action the user should take when this happens. Consider dropping to debug/trace

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/LoadBalancingPolicyWrapper.java
```diff
@@ -180,14 +181,17 @@ private void processNodeStateEvent(NodeStateEvent event) {
         return; // ignore
       case RUNNING:
         for (LoadBalancingPolicy policy : policies) {
-          if (event.newState == NodeState.UP) {
-            policy.onUp(event.node);
+          DefaultNode node = event.node.get();
+          if (node == null) {
```
hhughes: nit: do we need to re-perform the null for every policy? Is there a good reason not to pull this out of the loop?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeStateEvent.java
```diff
@@ -53,10 +54,10 @@ public static NodeStateEvent removed(DefaultNode node) {
    */
   public final NodeState newState;
 
-  public final DefaultNode node;
+  public final WeakReference<DefaultNode> node;
```
hhughes: Nit: Since we're changing the type here I'm wondering if it might be cleaner to provide a @Nullable getter for DefaultNode, rather than exposing the weak reference directly. Same comment for `DistanceEvent.node`.

core/src/main/java/com/datastax/oss/driver/internal/core/session/DefaultSession.java
```diff
@@ -525,14 +526,18 @@ private void notifyListeners() {
 
     private void onNodeStateChanged(NodeStateEvent event) {
       assert adminExecutor.inEventLoop();
-      if (event.newState == null) {
-        context.getNodeStateListener().onRemove(event.node);
+      DefaultNode node = event.node.get();
+      if (node == null) {
+        LOG.info(
```
hhughes: Nit: I think info-level is too high here

metrics/micrometer/src/main/java/com/datastax/oss/driver/internal/metrics/micrometer/MicrometerMetricsFactory.java
```diff
@@ -119,15 +120,22 @@ public NodeMetricUpdater newNodeUpdater(Node node) {
   }
 
   protected void processNodeStateEvent(NodeStateEvent event) {
+    DefaultNode node = event.node.get();
+    if (node == null) {
+      LOG.info(
```
hhughes: Nit: I think info-level is too high here

metrics/microprofile/src/main/java/com/datastax/oss/driver/internal/metrics/microprofile/MicroProfileMetricsFactory.java
```diff
@@ -121,16 +122,22 @@ public NodeMetricUpdater newNodeUpdater(Node node) {
   }
 
   protected void processNodeStateEvent(NodeStateEvent event) {
+    DefaultNode node = event.node.get();
+    if (node == null) {
+      LOG.info(
```
hhughes: Nit: I think info-level is too high here

hhughes (changes requested):
Looks good! Couple of minor pieces of feedback, most importantly I think we want to avoid logging at `info` when ignoring events which have lost their node reference as this might end up creating a lot of unnecssary log churn

hhughes (approved):
LGTM

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/ReplayingEventFilter.java
```diff
@@ -82,6 +82,7 @@ public void markReady() {
         consumer.accept(event);
       }
     } finally {
+      recordedEvents.clear();
```
aratno, line 85: What's the reasoning for this change?

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/AbstractMetricUpdater.java
```diff
@@ -173,9 +173,8 @@ protected Timeout newTimeout() {
         .getTimer()
         .newTimeout(
             t -> {
-              if (t.isExpired()) {
-                clearMetrics();
-              }
+              clearMetrics();
+              cancelMetricsExpirationTimeout();
```
aratno, line 177: What's the reasoning for this change?

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -276,38 +303,23 @@ protected boolean isBusy(@NonNull Node node, @NonNull Session session) {
...
-        return leastRecent - threshold < 0;
-      }
-    }
-    return true;
+    AtomicLongArray array = responseTimes.getIfPresent(node);
+    if (array != null && array.length() == 2) {
+      long threshold = now - RESPONSE_COUNT_RESET_INTERVAL_NANOS;
+      long leastRecent = array.get(0);
+      return leastRecent - threshold < 0;
+    } else return true;
```
aratno: Style nit: Invert the condition and use an early-return if response rate is insufficient, so you don't have `else return true`

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -96,14 +99,38 @@ public class DefaultLoadBalancingPolicy extends BasicLoadBalancingPolicy impleme
...
-  protected final Map<Node, AtomicLongArray> responseTimes = new ConcurrentHashMap<>();
+  protected final LoadingCache<Node, AtomicLongArray> responseTimes;
   protected final Map<Node, Long> upTimes = new ConcurrentHashMap<>();
   private final boolean avoidSlowReplicas;
 
   public DefaultLoadBalancingPolicy(@NonNull DriverContext context, @NonNull String profileName) {
     super(context, profileName);
     this.avoidSlowReplicas =
         profile.getBoolean(DefaultDriverOption.LOAD_BALANCING_POLICY_SLOW_AVOIDANCE, true);
+    CacheLoader<Node, AtomicLongArray> cacheLoader =
```
aratno: Style nit: use a separate class for the cache value here, rather than using AtomicLongArray as a generic container. Seems like it can be something like `NodeResponseRateSample`, with methods like `boolean hasSufficientResponses`. I see this was present in the previous implementation, so not a required change for this PR, just something I noticed.

aratno (commented):
Jira link for my own reference: https://datastax-oss.atlassian.net/browse/JAVA-3051

It would be helpful to have logs from the repro that the user submitted, specifically the logs from `com.datastax.oss.driver.internal.core.metadata.NodeStateManager.SingleThreaded#setState` that look like "[{}] Transitioning {} {}=>{} (because {})".

This seems like a bug in AWS Keyspaces, since each node includes itself in `system.peers`, which is unexpected to the driver according to the user's report:
> [s0] Control node <ip> has an entry for itself in system.peers: this entry will be ignored. This is likely due to a misconfiguration; please verify your rpc_address configuration in cassandra.yaml on all nodes in your cluster.

I left a few comments but otherwise these seem like generally positive changes. I agree that it's difficult to write unit tests for memory leaks like these, especially without any scaffolding around heapdump capture or parsing. I'm a bit concerned that there may be paths where a node event may not have any strong reference and is then garbage-collected and ignored by handlers, rather than surviving long enough to serve its purpose.

absurdfarce:
Very much agreed that the underlying issue here appears to be an issue with AWS Keyspaces @aratno; that's being addressed in a different ticket.  The scope of this change is around preventing the (potentially indefinite) caching of Node instances within an LBP.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -96,14 +99,38 @@ public class DefaultLoadBalancingPolicy extends BasicLoadBalancingPolicy impleme
...
+              array.set(0, previous);
+              array.set(1, now);
+            } else {
+              array.set(0, array.get(1));
+              array.set(1, now);
+            }
+            return array;
+          }
+        };
+    this.responseTimes = CacheBuilder.newBuilder().weakKeys().build(cacheLoader);
```
aratno: I think we should add a [RemovalListener](https://guava.dev/releases/21.0/api/docs/com/google/common/cache/RemovalListener.html) here.

If a GC happens and response times for a Node are purged, then we'll end up treating that as "insufficient responses" in `isResponseRateInsufficient`, which can lead us to mark a node as unhealthy. I recognize that this is a bit of a pathological example, but this behavior does depend on GC timing and would be a pain to track down, so adding logging could make someone's life easier down the line.

core/src/main/java/com/datastax/oss/driver/internal/core/util/concurrent/ReplayingEventFilter.java
SiyaoIsHiding: ReplayingEventFilter works like a buffer. It holds a list (queue) of Events since its state is STARTED, and consumes all of them when its state becomes READY all at once. However, the list of the events is never cleared. They leak strong references for the nodes.

core/src/main/java/com/datastax/oss/driver/internal/core/metrics/AbstractMetricUpdater.java
SiyaoIsHiding: This is a lambda for timeout. Even after the timeout and lambda triggered, the Timer object is not collected and it still holds a reference to this, until it's canceled.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
SiyaoIsHiding: Thank you for your review. Would you please explain more about this?
If GC collects a node, that means the node is gone. If the node is gone, why do we care about whether it's treated as healthy or not?
Anyway, for `RemovalListener`, do you mean sth like this?
```java
    this.responseTimes = CacheBuilder.newBuilder().weakKeys().removalListener(
            (RemovalListener<Node, AtomicLongArray>) notification -> 
                    LOG.trace("[{}] Evicting response times for {}: {}", 
                            logPrefix, notification.getKey(), notification.getCause()))
            .build(cacheLoader);
```

SiyaoIsHiding: @aratno Hi Abe, thank you for your review. Is there any update?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeStateEvent.java
```diff
@@ -53,14 +55,19 @@ public static NodeStateEvent removed(DefaultNode node) {
    */
   public final NodeState newState;
 
-  public final DefaultNode node;
+  private final WeakReference<DefaultNode> node;
```
absurdfarce: I don't think we need this change.  Events of this type are very short-lived; they exist to communicate information between driver components that don't necessarily know about each other.  They aren't stockpiled or stored in any meaningful way.  You address the problem that was originally reported by changing the distances map in LoadBalancingPolicyWrapper to use weak refs... it's not at all clear to me that making the events use weak references buys you much on top of that.

Perhaps more importantly this change has the potential to make events a lot less useful.  The driver uses events to notify components about changes in nodes, but if the _actual node affected_ might not be present what use is the notification?  Components that receive events without node information have no choice but to ignore them which means (a) every receiver has to check for null node information and (b) if you just ignore events without node data (which all these receivers appear to do) you'll get a lot more indeterminate behaviour since the presence or absence of node data in events is basically a random value (since it's determined by GC pressure which is essentially random from the perspective of the driver).

I note that the null checks referenced in (a) above are a big chunk of what's actually in this PR.  If not for this constraint the change set would be a lot smaller.  That wouldn't be the worse thing as the presence of all those null checks obscures the more significant changes in at least a moderate way.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/DistanceEvent.java
```diff
@@ -29,11 +31,16 @@
 @Immutable
 public class DistanceEvent {
   public final NodeDistance distance;
-  public final DefaultNode node;
+  private final WeakReference<DefaultNode> node;
```
absurdfarce: Same as discussed in NodeStateEvent; I don't think we need a WeakReference here.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -96,14 +100,39 @@ public class DefaultLoadBalancingPolicy extends BasicLoadBalancingPolicy impleme
...
+        CacheBuilder.newBuilder()
+            .weakKeys()
+            .removalListener(
+                (RemovalListener<Node, NodeResponseRateSample>)
+                    notification ->
+                        LOG.trace(
+                            "[{}] Evicting response times for {}: {}",
+                            logPrefix,
+                            notification.getKey(),
+                            notification.getCause()))
```
absurdfarce: @aratno I know you asked for this RemovalListener in a [previous comment](https://github.com/apache/cassandra-java-driver/pull/1743#discussion_r1479160472) under the assumption that it would help identify cases in which a node was (incorrectly) marked unhealthy because data had expired from the map.  I'm going to argue instead that we shouldn't mark a node as unhealthy unless we have clear data indicating that it is so... which means the absence of response time data isn't enough to make it unhealthy.  If we take that approach would you still argue for a removal listener here?  I'm kind of two minds about it myself; I can see the benefit but it's possibly less interesting if the potentially damaging effects are mitigated.

Thoughts?

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -274,40 +303,23 @@ protected boolean isBusy(@NonNull Node node, @NonNull Session session) {
...
-      if (array.length() == 2) {
-        long threshold = now - RESPONSE_COUNT_RESET_INTERVAL_NANOS;
-        long leastRecent = array.get(0);
-        return leastRecent - threshold < 0;
-      }
+    NodeResponseRateSample sample = responseTimes.getIfPresent(node);
+    if (sample == null) {
+      return true;
+    } else {
+      return !sample.hasSufficientResponses(now);
```
absurdfarce: Style nit: the ternary operator does this a bit more cleanly:

```java
return (sample == null) ? true : !sample.hasSufficientResponses(now)
```

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -274,40 +303,23 @@ protected boolean isBusy(@NonNull Node node, @NonNull Session session) {
...
-    if (responseTimes.containsKey(node)) {
-      AtomicLongArray array = responseTimes.get(node);
-      if (array.length() == 2) {
-        long threshold = now - RESPONSE_COUNT_RESET_INTERVAL_NANOS;
-        long leastRecent = array.get(0);
-        return leastRecent - threshold < 0;
-      }
+    NodeResponseRateSample sample = responseTimes.getIfPresent(node);
+    if (sample == null) {
+      return true;
```
absurdfarce: I'm wondering if this is too aggressive now that we're not guaranteed that historical node response time data will always be present.  I'm leaning towards an argument which says that we should only say a node's response time is insufficient if we have clear, unambiguous data which indicates that.  That means that missing data (either because it hasn't been observed yet or GC pressure has removed it now that we're using weak refs) would _not_ be enough to mark a node as unhealthy.

@adutra I'm very interested in your take on this; you have considerably more context with which to evaluate this question than I do. :)

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -96,14 +100,39 @@ public class DefaultLoadBalancingPolicy extends BasicLoadBalancingPolicy impleme
...
+          public NodeResponseRateSample load(Node key) {
+            NodeResponseRateSample sample = responseTimes.getIfPresent(key);
+            if (sample == null) {
+              sample = new NodeResponseRateSample();
+            } else {
+              sample.update();
+            }
+            return sample;
+          }
+        };
```
absurdfarce: You actually don't need a LoadingCache here; a simple Cache will do.  A LoadingCache is useful when you need to create an entry for a key that's been requested but isn't in the cache yet.  That's not what's going on here; your CacheLoader is loading keys from the map, creating a new version if it isn't present and then updating it.  You're not doing something to create the NodeResponseRateSample instance beyond calling the constructor if necessary.

You can accomplish the exact same thing using just a regular Cache by changing your update logic just a bit:

```java
  protected void updateResponseTimes(@NonNull Node node) {
    try {

      responseTimes.get(node, NodeResponseRateSample::new).update();
    }
    catch (ExecutionException ee) {
      LOG.info("[{}] Exception updating node response times: {}", logPrefix, ee);
    }
  }
```

I note that this approach is even [recommended](https://javadoc.io/static/com.google.guava/guava/33.2.1-jre/com/google/common/cache/Cache.html#put(K,V)) in the Guava Javadoc.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -318,4 +330,42 @@ protected int getInFlight(@NonNull Node node, @NonNull Session session) {
...
+      if (times.length() == 1) {
+        long previous = times.get(0);
+        times = new AtomicLongArray(2);
+        times.set(0, previous);
+        times.set(1, now);
+      } else {
+        times.set(0, times.get(1));
+        times.set(1, now);
+      }
+    }
```
absurdfarce, line 342: This seems to be subject to race conditions to me, although I haven't been able to construct a specific scenario that triggers it.  The old implementation computed a result array locally and then returned that; there was no changing of state at the ultimate destination (the old responseTimes map) beyond updating an entry or not doing so.  This implementation is doing more than that; we're actually changing the types of elements (this.times changes in size) based on data we receive.

It might be okay... but then again I'm also not sure I see a harm in just making this.times be an array of size two from the beginning.

absurdfarce (changes requested):
Good work @SiyaoIsHiding!  I'd like to talk about the weak references in events a bit more.  There are a few other changes mentioned as well but the event references are really the big one for me.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
adutra: Isn't your version subject to race conditions though. In the loading cache version, `NodeResponseRateSample.update()` is called inside `CacheLoader.load()` and thus cannot be called concurrently. In your suggestion, `update()` is called _after_ the value is retrieved from the cache, and thus could be executed concurrently.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
adutra: Yes, I tend to agree that the absence of any samples should be interpreted as "we don't know".

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
adutra: Or even:

```java
return sample == null || !sample.hasSufficientResponses(now)
```

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
adutra: As I stated previously, this will race indeed, _unless_ calls to `update()` are externally synchronized. I think the loading cache approach does guarantee that.

The idea of using an array of size 2 from the beginning could save us one array allocation – why not.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -318,4 +330,42 @@ protected int getInFlight(@NonNull Node node, @NonNull Session session) {
...
+    // The array stores at most two timestamps, since we don't need more;
+    // the first one is always the least recent one, and hence the one to inspect.
+    protected AtomicLongArray times;
+
+    private NodeResponseRateSample() {
+      times = new AtomicLongArray(1);
+      times.set(0, nanoTime());
+    }
+
+    // Only for unit tests
```
adutra: Maybe annotate with `@VisibleForTesting`?

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
absurdfarce: Honestly I've gone round and round about this one over the past few days.  I agree that my impl isn't right but a LoadingCache (or at least a LoadingCache used in the way it's used here) still feels wrong to me.  I _vastly_ prefer the approach of the original code which just leverages ConcurrentHashMap.compute() to do the right thing.

I think part of the mismatch for me here is that CacheLoader.get() doesn't provide the current value (if any) as an arg so you wind up having to do this weird refresh() dance.  I'm assuming part of the reason the key wasn't provided is because that's not really the intended use case for a CacheLoader but I digress.  As mentioned ConcurrentHashMap seems to model this case more effectively, and while that class doesn't support weak references directly we can get there indirectly with Guava'a MapMaker: `new MapMaker().weakKeys.getMap()` will give us what we want.  The only thing we're missing then is the RemovalListener, but as mentioned elsewhere (a) I'm not sure if that's necessary now and (b) I'm not sure it'll do what we want anyway.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
absurdfarce: Also, a related question: do we know if RemovalListeners get called when weak keys are removed via GC?  Javadoc seems to say yes but it might be useful to confirm.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
absurdfarce: To make this more concrete: I got the relevant test case (DefaultLoadBalancingPolicyRequestTrackerTest) with the following changes:

```java
  protected final ConcurrentMap<Node, NodeResponseRateSample> responseTimes;
  protected final Map<Node, Long> upTimes = new ConcurrentHashMap<>();
  private final boolean avoidSlowReplicas;

  public DefaultLoadBalancingPolicy(@NonNull DriverContext context, @NonNull String profileName) {
    super(context, profileName);
    this.avoidSlowReplicas =
        profile.getBoolean(DefaultDriverOption.LOAD_BALANCING_POLICY_SLOW_AVOIDANCE, true);
    this.responseTimes =
            new MapMaker().weakKeys().makeMap();
  }
...
  protected void updateResponseTimes(@NonNull Node node) {
    this.responseTimes.compute(node, (k,v) -> v == null ? new NodeResponseRateSample() : v.next());
  }
...
  // Exposed as protected for unit tests
  protected class NodeResponseRateSample {

    @VisibleForTesting
    protected final long oldest;
    @VisibleForTesting
    protected final OptionalLong newest;

    private NodeResponseRateSample() {

      long now = nanoTime();
      this.oldest = now;
      this.newest = OptionalLong.empty();
    }

    private NodeResponseRateSample(long oldestSample) {
      this(oldestSample, nanoTime());
    }

    private NodeResponseRateSample(long oldestSample, long newestSample) {
      this.oldest = oldestSample;
      this.newest = OptionalLong.of(newestSample);
    }

    @VisibleForTesting
    protected NodeResponseRateSample(AtomicLongArray times) {
      assert times.length() >= 1;
      this.oldest = times.get(0);
      this.newest = (times.length() > 1) ? OptionalLong.of(times.get(1)) : OptionalLong.empty();
    }

    // Our newest sample becomes the oldest in the next generation
    private NodeResponseRateSample next() {
      return new NodeResponseRateSample(this.getNewestValidSample(), nanoTime());
    }

    // If we have a pair of values return the newest, otherwise we have just one value... so just return it
    private long getNewestValidSample() { return this.newest.orElse(this.oldest); }

    // response rate is considered insufficient when less than 2 responses were obtained in
    // the past interval delimited by RESPONSE_COUNT_RESET_INTERVAL_NANOS.
    private boolean hasSufficientResponses(long now) {
      // If we only have one sample it's an automatic failure
      if (!this.newest.isPresent())
        return false;
      long threshold = now - RESPONSE_COUNT_RESET_INTERVAL_NANOS;
      return this.oldest - threshold >= 0;
    }
  }
```

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeStateEvent.java
SiyaoIsHiding: @absurdfarce I will try to revert the changes and see whether it still leaks, but:
There are two `WeakHashMap` held by `ControlConnection`:
```java
    private final Map<Node, DistanceEvent> lastDistanceEvents = new WeakHashMap<>();
    private final Map<Node, NodeStateEvent> lastStateEvents = new WeakHashMap<>();
```
However, both DistanceEvent and NodeStateEvent hold strong references to the associated node 
From the [doc](https://docs.oracle.com/javase/8/docs/api/java/util/WeakHashMap.html) of WeakHashMap:

> Implementation note: The value objects in a WeakHashMap are held by ordinary strong references. Thus care should be taken to ensure that value objects do not strongly refer to their own keys, either directly or indirectly, since that will prevent the keys from being discarded.

In this case `DistanceEvent` and `NodeStateEvent` are the values that hold strong references to the keys. Therefore, I think it may leak references.

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
SiyaoIsHiding: I want to confirm that this means both "no oldest" and "no newest" will be considered "sufficient data". Right?

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeStateEvent.java
SiyaoIsHiding: Tested out. If `NodeStateEvent` and `DistanceEvent` hold strong references to `node`, then objects of these two kinds, together with `DefaultNode`, will not be garbage collected.

SiyaoIsHiding:
We still need to figure out how `NodeStateEvent` and `DistanceEvent` refer to their `node`. Other than that, the update PR should have resolved all the other issues.

core/src/main/java/com/datastax/oss/driver/internal/core/metadata/NodeStateEvent.java
SiyaoIsHiding: Confirmed that if we put `lastStateEvent` and `lastDistanceEvent` as two fields of `DefaultNode`, then it will work. `DefaultNode` <-> `NodeStateEvent` only point to each other, same for `DistanceEvent`. Objects of all three classes will be collected. However, I understand this is not an ideal solution.

adutra: What about this:

```java
    private final Map<Node, NodeDistance> lastDistanceEvents = new WeakHashMap<>();
    private final Map<Node, NodeState> lastStateEvents = new WeakHashMap<>();
```

SiyaoIsHiding:
Solved the `NodeState` and `NodeDistance` problem as discussed. All the issues should be resolved now. I would appreciate a review :)

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
```diff
@@ -274,40 +279,19 @@ protected boolean isBusy(@NonNull Node node, @NonNull Session session) {
...
-            array = new AtomicLongArray(2);
-            array.set(0, previous);
-            array.set(1, now);
-          } else {
-            array.set(0, array.get(1));
-            array.set(1, now);
-          }
-          return array;
-        });
+    this.responseTimes.compute(node, (k, v) -> v == null ? new NodeResponseRateSample() : v.next());
```
adutra, line 294: Not sure if this was discussed already: the new design is undeniably more elegant, but it also allocates one new `NodeResponseRateSample` each time we access the map entry (because the entry value is immutable). In the old design the value was mutable in order to avoid allocations. Are we OK with this change?

absurdfarce: My expectation is that it wouldn't matter since we're creating (and GCing) pretty small objects and the JVM usually handles that quite well but it's not unreasonable to confirm that.

@SiyaoIsHiding is this something you can test?  I suppose if we can demonstrate roughly equivalent memory usage and throughput for a decent size load that would be fine.

SiyaoIsHiding: I tested with a client app with VisualVM. 
Using this branch, `NodeResponseSample` uses up to 5000 live bytes with 170 live objects. The app finished 40000 requests in 154 seconds. 
Using the current 4.x branch, A lambda in `DefaultLoadBalancingPolicy` uses up to 3600 live bytes with 200 live objects. The only lambda I find there is in the `responseTimes.compute` [here](https://github.com/apache/cassandra-java-driver/blob/9cfb4f6712e392c1b6c87db268565fd3b27d0c5c/core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java#L291-L310). The app finished 40000 requests in 161 seconds.
So I think there isn't much performance difference.

adutra: Sorry for the late reply, and thanks for the benchmark. I think we are good here. Thanks @SiyaoIsHiding !

adutra (approved):

core/src/main/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicy.java
absurdfarce: Comments above refer to an earlier implementation of this functionality.  New version wraps everything in compute() and drives updates off of immutable NodeResponseRateSample objects so these concerns have been addressed.

core/src/test/java/com/datastax/oss/driver/internal/core/loadbalancing/DefaultLoadBalancingPolicyRequestTrackerTest.java
```diff
@@ -186,13 +186,13 @@ public void should_record_further_response_times_on_node_error() {
...
+              assertThat(value.newest.getAsLong()).isEqualTo(789);
             })
-        .hasEntrySatisfying(node2, value -> assertThat(value.get(0)).isEqualTo(789))
+        .hasEntrySatisfying(node2, value -> assertThat(value.oldest).isEqualTo(789))
         .doesNotContainKey(node3);
     assertThat(policy.isResponseRateInsufficient(node1, nextNanoTime)).isFalse();
-    assertThat(policy.isResponseRateInsufficient(node2, nextNanoTime)).isTrue();
-    assertThat(policy.isResponseRateInsufficient(node3, nextNanoTime)).isTrue();
+    assertThat(policy.isResponseRateInsufficient(node2, nextNanoTime)).isFalse();
+    assertThat(policy.isResponseRateInsufficient(node3, nextNanoTime)).isFalse();
```
absurdfarce, line 196: Semi-nit: after these changes we no longer have any test case which tests assumptions about when isResponseRateInsufficient() will return true.  We only had a few before so this isn't the end of the world... and it's not as if that method isn't reasonably well exercised by the logic here.  But I am just a bit concerned that we should have a test case to confirm that this method will return true when we create cases where it should return true.

absurdfarce (approved):
I'm quite happy with where this landed.  Had one relatively minor comment about some testing enhancements I think we could add here but it doesn't strike me as essential to do so.  This PR has been waiting a long time and it's quite clear that these changes make the situation better than it was so a side issue isn't really enough to derail this.

absurdfarce:
Great news @SiyaoIsHiding; we now have two +1 votes for this PR!  Please squash your work here down to a single commit and add the expected "patch by" line as well if you would:

```
patch by Jane He; reviewed by Alexandre Dutra and Bret McGuire for JAVA-3051
```

Once that's done I believe we're ready to merge!  Thank you _very much_ for your patience!

SiyaoIsHiding:
Thanks for the reviews and approvals! Commits squashed.

absurdfarce:
Looks great, thanks @SiyaoIsHiding !

PR #1757, opened 2023-11-03
JAVA-3055 Prevent reusing cached cancelled requests

lucboutier:
There was a critical issue in the driver that may occur when the external code cancels a request, indeed the CompletableFuture will then always throw a CancellationException.

This could occur for example when used in a reactive context with a Mono.zip failed.


adutra (approved):

absurdfarce:
I agree this is ultimately the right approach but I think I'd rather do more to completely decouple the futures returned to the user from the future that computes the PreparedStatement object.  #2003 as an example.

absurdfarce:
#2003 didn't work nearly as well as I'd hoped.  Closing that out and going with this for now; we can always refactor things later if we feel we need to make the interacting components clearer.

Thanks for the patch @lucboutier!

absurdfarce (approved):

PR #1843, opened 2023-11-14
CASSANDRA-18969: 3.1.x Move copyright notices to LICENSE, add bundled ASL dep notices to NOTICE

hhughes:


michaelsembwever:
is there nothing for LICENSE here ?

hhughes:
I don't think so. The files mentioned in LICENSE in later branches do not exist in this branch and I don't find any other licesnse headers (except the common one the project) or comments matching `copyright`, `©`, `(c)`, `&copy;` or `|copy|` using grep. Let me know if there are any other patterns I can look for.

PR #1864, opened 2023-11-17
CASSANDRA-18969: 4.x Add LICENSE and NOTICE.txt/NOTICE_binary to published jars

hhughes:


michaelsembwever:
Ok.

So what we need to do is… 

1. Copy NOTICE.txt to NOTICE_binary.txt
2. Remove the netty stuff from NOTICE.txt
3. Include NOTICE_binary.txt in the jar's META-INF
4. Add to NOTICE_binary.txt the following
```
This compiled product also includes Apache-licensed dependencies that contain the following NOTICE information: 
```

This way we are honouring 
> It is important to keep NOTICE as brief and simple as possible, as each addition places a burden on downstream consumers.
>
> Do not add anything to NOTICE which is not legally required.
> …
> …As a result, the LICENSE and NOTICE files for a binary distribution may differ from those in the source distribution it was built from.
>
> In any case, the principle remains the same: LICENSE and NOTICE must exactly represent the contents of the distribution they reside in.

ref: https://infra.apache.org/licensing-howto.html#mod-notice

michaelsembwever (approved):

PR #1865, opened 2023-11-17
Update README.md

emeliawilkinson24:
Typo carried over from old docs, needed closing parenthesis.


michaelsembwever:
@absurdfarce what should we do about this page ? 

documentation on "how to connect to different cloud providers" we should definitely embrace.

but the SCB code won't last in the driver forever…

absurdfarce:
@michaelsembwever we've been fixing a few things incrementally around the docs; my general feeling is that's probably fine for now.  It's not at all clear to me what the future of the Astra code is in the driver itself; the few times the topic has come up other devs working on the driver have been unexpectedly enthusiastic about keeping it in.  Regardless, it's not getting yanked immediately, so it seems eminently reasonable to me to have docs for our current state of affairs (which includes support for the SCB) which may have to change at some point in the future.

absurdfarce:
Looks good, thanks @emeliawilkinson24 !

PR #1894, opened 2023-11-28
CASSANDRA-19051: Test case DescribeIT.describe_output_should_match_creation_script fails

SiyaoIsHiding:


hhughes (approved):
I think it should be OK to remove these licenses as they supporting data-files which don't contain any interesting IP.

SiyaoIsHiding:
Closed because it's resolved by [CASSANDRA-18970](https://issues.apache.org/jira/browse/CASSANDRA-18970).

